PR #113, opened 2020-02-28
Add two CLI options to help with the installation inside of conda

qwfy:
Two CLI options are added to `install.sh`:

- `--no_sudo` runs all commands without `sudo`
- `--pip_mirror` allows the user to set the pip mirror

with these two options, user can install deepspeed to their conda environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,6 +106,6 @@ RUN git clone https://github.com/microsoft/DeepSpeed.git ${STAGE_DIR}/DeepSpeed
 RUN cd ${STAGE_DIR}/DeepSpeed && \
     git checkout . && \
     git checkout master && \
-    ./install.sh
+    ./install.sh --pip_sudo
 RUN rm -rf ${STAGE_DIR}/DeepSpeed
 RUN python -c "import deepspeed; print(deepspeed.__version__)"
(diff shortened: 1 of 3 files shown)


ShadenSmith:
Hello! Thank you very much for your contribution to DeepSpeed.

For most users, we are thinking that the no-sudo case is the wanted one. Would you mind inverting the default behavior and instead using a `--sudo` flag?

qwfy:
@ShadenSmith Done. Except that the option is named `--pip_sudo`, which is only effect for pip commands, not `--sudo`, considering some other installation procedures (e.g. write some config to `/etc/`) may also requires `sudo`, and the user may want to run pip without sudo, but allow writing to `/etc/` with sudo.

jeffra:
This looks good, we’ll have to update a few parts of our tests to get this to pass. Will do that tomorrow and then we can get this merged in.

jeffra:
Can you add `--pip_sudo` to this line? This should allow the tests to pass

https://github.com/microsoft/DeepSpeed/blob/master/azure-pipelines.yml#L31

ShadenSmith:
Hi @qwfy , thank you very much for the contribution.

We have a now have a conda package uploaded and we appreciate any feedback :-).

We have versions compiled for `cudatoolkit` versions 10.0 and 10.1 To install along with pytorch and other dependencies that are in the `conda-forge` channel:
```bash
conda install deepspeed cudatoolkit=10.1 -c deepspeed -c pytorch -c conda-forge
```

PR #117, opened 2020-03-03
Add deepspeed env file to pass custom env values

jeffra:
Allow users to specify their own environment variables that they would like to pass to worker nodes


ShadenSmith:
Where is the best place to document this behavior?

jeffra:
Also fixes an open issue related to default deepspeed config

PR #124, opened 2020-03-09
Make lr schedulers support fp16 optimizers

tjruwase:
Support fp16 optimizers by using the wrapped optimizer.

### DIFF
--- a/tests/unit/test_fp16.py
+++ b/tests/unit/test_fp16.py
@@ -142,3 +142,105 @@ def test_adamw_fp16_empty_grad(tmpdir):
             model.step()
 
     _test_adamw_fp16_empty_grad(args=args, model=model, hidden_dim=hidden_dim)
+
+
+def test_adam_fp16_onecycle(tmpdir):
+    config_dict = {
+        "train_batch_size": 1,
+        "steps_per_print": 1,
+        "optimizer": {
+            "type": "Adam",
+            "params": {
+                "lr": 0.00015
+            }
+        },
+        "scheduler": {
+            "type": "OneCycle",
+            "params": {
+                "cycle_first_step_size": 16000,
+                "cycle_first_stair_count": 8000,
+                "decay_step_size": 16000,
+                "cycle_min_lr": 1e-06,
+                "cycle_max_lr": 3e-05,
+                "decay_lr_rate": 1e-07,
+                "cycle_min_mom": 0.85,
+                "cycle_max_mom": 0.99,
+                "decay_mom_rate": 0.0
+            }
+        },
+        "fp16": {
+            "enabled": True
+        },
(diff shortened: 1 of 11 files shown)


tests/unit/test_fp16.py
```diff
@@ -142,3 +142,105 @@ def _test_adamw_fp16_empty_grad(args, model, hidden_dim):
...
+            "enabled": True
+        },
+        "zero_optimization": False
+    }
+    args = args_from_dict(tmpdir, config_dict)
+    hidden_dim = 10
+
+    model = SimpleModel(hidden_dim, empty_grad=True)
+
+    @distributed_test(world_size=[1])
```
ShadenSmith: You only need the `distributed_test` decorator when you want the unit test to run in multiple processes.

tests/unit/test_fp16.py
```diff
@@ -142,3 +142,105 @@ def _test_adamw_fp16_empty_grad(args, model, hidden_dim):
...
+    def _test_adam_fp16_onecycle(args, model, hidden_dim):
+        model, _, _,_ = deepspeed.initialize(args=args,
+                                             model=model,
+                                             model_parameters=model.parameters())
+        data_loader = random_dataloader(model=model,
+                                        total_samples=50,
+                                        hidden_dim=hidden_dim,
+                                        device=model.device)
+        for n, batch in enumerate(data_loader):
```
ShadenSmith: We need some sort of evaluation here or else the unit test will always pass unless it crashes. Should we do something like inspect the scheduler to ensure it's following the schedule correctly?

@jeffra had the idea that we could test LR schedulers without actually training a model, since we really only care about scheduler for most of these tests.

We should also test the scheduler's interaction with an optimizer during an actual training, too...maybe that would be a better fit for the optimizer test suite, though.

ShadenSmith: I get that this PR is focused on enabling fp16, so perhaps not crashing is sufficient for the scope of the PR. We can move my other comments to an issue.

tjruwase: Yes, this PR is focused on enabling fp16 optimizers in lr schedulers. Will it suffice to rename the test as follows tests/unit/lr_scheduler/fp16_optimizer_compatbility.py?

ShadenSmith: I think the filename is fine since it's fp16 specific. Maybe rename the specific test method to be something like `test_adam_fp16_zero_onecycle_compatibility()` and a comment about its purpose?

tjruwase: Sounds good. 

By the way here is the related [issue ](https://github.com/microsoft/DeepSpeed/issues/127)to the points you raised:

PR #139, opened 2020-03-15
Fix global_steps checkpoint loading.

ShadenSmith:


ShadenSmith:
The PR tests aren't triggering for some reason. Investigating.

ShadenSmith:
Finally got around to flushing out some tests. We now explicitly test for those deepspeed states like `global_steps`. I also added tests for LR schedulers.

PR #161, opened 2020-03-20
removed restrictions for custom optimizer

CalogeroZarbo:
Removed the restriction to be able to use any optimizer of choice. A warning message has been added.

### DIFF
--- a/deepspeed/pt/deepspeed_light.py
+++ b/deepspeed/pt/deepspeed_light.py
@@ -445,7 +445,7 @@ class DeepSpeedLight(Module):
             if self.optimizer_name != ADAM_OPTIMIZER:
                 logging.warning(
                     "**** You are using ZeRO with an untested optimizer, proceed with caution *****"
-                    )
+                )
             self.optimizer = self._configure_zero_optimizer(basic_optimizer)
         elif self.fp16_enabled():
             self.optimizer = self._configure_fp16_optimizer(basic_optimizer)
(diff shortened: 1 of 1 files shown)


tjruwase:
Looks good

ShadenSmith:
Thanks for your contribution to DeepSpeed!

CalogeroZarbo:
Thank you too for this amazing lib!

ShadenSmith:
I was thinking this over some more and I'm worried that the lone warning is easy to miss in a typically-long logfile. That could lead to some "silent" divergence errors.

What if we keep the warning, but require an additional config in the JSON such as `"zero_allow_untested_optimizer" : true` in order to "unlock" the untested optimizer? If the flag is not set, we raise an error and provide instructions and the appropriate warnings.

CalogeroZarbo:
Yep @ShadenSmith that seems to me as a very good approach to avoid problems difficult do debug in a complex training system. I'll fork again and make another pull-request if that's ok for you. Let me know.

Cheers,
Cal

tjruwase:
@CalogeroZarbo this sounds good. Thanks!

ShadenSmith:
Perfect, thanks @CalogeroZarbo !

FWkey:
what are the possible errors if i use custom optimizer? or how to ensure its correctness, should i do some tests?

tjruwase:
@FWkey, this is a tricky question. You could compare training loss for a smaller model of pytorch DDP vs ZeRO

PR #162, opened 2020-03-22
removed session_params from deepspeed_constants.py

kouml:
removed session_params from deepspeed_constants.py.

- removed session_params
- add reference of config params at README.md


ShadenSmith:
Thank you very much for your contribution to DeepSpeed!

PR #164, opened 2020-03-22
Fix incorrect tensorboard loss calculation

wildermuthn:
The loss is scaled by the gradient_accumulation_steps value unnecessarily. Probably a leftover from when gradient scaling was done in the backward pass?

### DIFF
--- a/deepspeed/pt/deepspeed_light.py
+++ b/deepspeed/pt/deepspeed_light.py
@@ -630,8 +630,7 @@ class DeepSpeedLight(Module):
             self.summary_events = [
                 (f'Train/Samples/train_loss',
                  loss.mean().item(),
-                 self.sample_count)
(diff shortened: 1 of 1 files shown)


ShadenSmith:
Hi @wildermuthn, it looks like the code change isn't getting past our formatter. I made a minor tweak but realized there's a second one needed.

We have some [info here](https://www.deepspeed.ai/contributing/#prerequisites) about running our pre-commit formatting checks. Can you do that locally and push the tweaked code? Thanks!

tjruwase:
@wildermuthn  Thanks for your contribution to DeepSpeed. I am closing this PR because it is subsumed by another [commit](https://github.com/microsoft/DeepSpeed/pull/170). Please check it out to see if it addresses your issue.

Issue #167, opened 2020-03-24
DeepSpeedDataLoader's len is not caculated properly


tomekrut:
DeepSpeedDataLoader len should be divided by the batch_size

instead

`self.len = len(self.data_sampler)`

we should have

```
import math
self.len = int(math.ceil(len(self.data_sampler) * 1.0 / self.batch_size))
```

tjruwase:
Closing because of #176

PR #168, opened 2020-03-24
fix dataloader len for tqdm and others

tomekrut:
I found a small issue see below - the fix created for it.

https://github.com/microsoft/DeepSpeed/issues/167

### DIFF
--- a/deepspeed/pt/deepspeed_dataloader.py
+++ b/deepspeed/pt/deepspeed_dataloader.py
@@ -18,6 +18,7 @@ class DeepSpeedDataLoader(object):
                  collate_fn=None,
                  num_local_io_workers=None,
                  data_sampler=None):
+
         self.tput_timer = tput_timer
         self.batch_size = batch_size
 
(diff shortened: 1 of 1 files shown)


tjruwase:
Thanks for contributing to DeepSpeed. We really appreciate it. 

For documentation purposes, could you please explain more what the problem is and why this is fix?

tomekrut:
Just few more comments to the above fix:
in case anyone wants to use tqdm for example... the number of batches to complete a single epoch is not correct. it happens because the DeepSpeed's DataLoader len is not calculated correctly. See more  details in the https://github.com/microsoft/DeepSpeed/issues/167

tjruwase:
@tomekrut  Thanks for contributing to DeepSpeed. Apologies for the delay in getting back to you. We need time to investigate and fully understand the underlying issue that you have found. Yes, you found a subtle bug in DeepSpeed, and thanks so much for that. However, this PR does not fully address the underlying issue; it fixes one scenario but breaks others as I will explain below. 

First I will provide some context to the origin of DeepSpeedDataLoader. The goal is for it to be wrapper around [torch DataLoader](https://github.com/pytorch/pytorch/blob/c8bd5ac7e9f31dc66f3bbb21bfed6fb3aa20a190/torch/utils/data/dataloader.py#L60) to enable us conveniently add data loading optimizations in the future.  As such we strive to maintain similar semantics as torch DataLoader. 

I observe that torch DataLoader supports two modes: [batching](https://pytorch.org/docs/stable/data.html#automatic-batching-default), which returns batched samples, and [non-batching](https://pytorch.org/docs/stable/data.html#disable-automatic-batching), which returns individual samples. This PR fixes the bug for batching mode but introduces a new bug for non-batching mode. It seems to me that more extensive code changes and unit tests are required to support both modes and address other issues in DeepSpeedDataLoader. I have now created #176 to track this bug.

tjruwase:
Closing because it is now replaced by #178

PR #170, opened 2020-03-24
Support multi-output models

tjruwase:
Correctly handle multi-output models by performing loss scaling in backward() instead of forward()

### DIFF
--- a/deepspeed/pt/deepspeed_light.py
+++ b/deepspeed/pt/deepspeed_light.py
@@ -564,20 +564,25 @@ class DeepSpeedLight(Module):
         self.warn_unscaled_loss = True
         self.module.train(False)
 
-    def _scale_loss(self, loss):
-        if isinstance(loss, torch.Tensor):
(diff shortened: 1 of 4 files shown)


ShadenSmith:
@tjruwase I think you will need to tell DeepSpeed to update the DeepSpeedExamples submodule in order to incorporate your fix to Squad last night.

You can do that from your branch:
```bash
git submodule update --remote DeepSpeedExamples
```

Then commit that update.

ShadenSmith:
@tjruwase would you do me a favor and also enable the Megatron tests in your branch for now? Megatron is disabled by default until I get the nightly tests going. Those lines are just commented out in `tests/model/run_sanity_check.py`. I'm wondering if we need to account for the new loss scaling in those tests too.

tjruwase:
@ShadenSmith Thanks for the guidance. I have updated the DeepSpeedExamples submodule and enabled Megatron in model tests. Now run_sanity_checks passes. For some reason, github has not received notification of the model tests passing.

PR #171, opened 2020-03-25
Fix ThroughputTimer with hybrid parallelism.

ShadenSmith:
`ThroughputTimer` computes the number of samples as `batch_size * num_workers`. The global world size will cause the timer to over-count samples if model parallelism is used, and so we should use `self.dp_world_size` in case `mpu` was supplied.

### DIFF
--- a/deepspeed/pt/deepspeed_light.py
+++ b/deepspeed/pt/deepspeed_light.py
@@ -142,17 +142,17 @@ class DeepSpeedLight(Module):
 
         self._init_distributed(dist_init_required)
 
+        # Configure distributed model
+        self._configure_distributed_model(model)
+
         # Throughput timer
         self.tput_timer = ThroughputTimer(
             batch_size=self.train_micro_batch_size_per_gpu(),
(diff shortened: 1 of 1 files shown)


deepspeed/pt/deepspeed_light.py
```diff
@@ -879,22 +885,22 @@ def csr_all_gather(self, value):
         if value.dim() == 1:
             if fill_size > 0:
                 value = torch.cat([value, value.new_zeros(fill_size)])
-            tensor_list = [
-                value.new_zeros(max_size) for _ in range(dist.get_world_size())
-            ]
+            tensor_list = [value.new_zeros(max_size) for _ in range(self.dp_world_size)]
```
ShadenSmith: I made an educated guess that this should be done over the data-parallel instead of global world. @jeffra could you please confirm?

ShadenSmith:
@tjruwase  I force-pushed over your comment and can't resolve it. I think I got the rest of them, though. Please look over for any issues.

PR #173, opened 2020-03-26
Add "zero_allow_untested_optimizer" option in conf file

CalogeroZarbo:
Dear all,
I've added the flag, in the same flavor as all the other flags have been addressed, by changing the DeepSpeedConfig class. Also, the assertion has been added to raise an error message if the flag is not set, but the user wants to use the custom optimizer. If both `zero_allow_untested_optimizer` and `zero_optimization` are set to `true` a warning message appears and the computation starts normally.

### DIFF
--- a/deepspeed/pt/deepspeed_config.py
+++ b/deepspeed/pt/deepspeed_config.py
@@ -165,6 +165,12 @@ def get_optimizer_legacy_fusion(param_dict):
         return LEGACY_FUSION_DEFAULT
 
 
+def get_zero_allow_untested_optimizer(param_dict):
+    return get_scalar_param(param_dict,
+                            ZERO_ALLOW_UNTESTED_OPTIMIZER,
+                            ZERO_ALLOW_UNTESTED_OPTIMIZER_DEFAULT)
+
+
 def get_scheduler_name(param_dict):
     if SCHEDULER in param_dict.keys() and \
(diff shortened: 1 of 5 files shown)


ShadenSmith:
Thanks so much for putting this together @CalogeroZarbo !

Would you mind adding a unit test that should trigger the assertion? Here's a similar test it could be based off of: https://github.com/microsoft/DeepSpeed/blob/master/tests/unit/test_fp16.py#L251

You could catch the assertion error with pytest:
```python
with pytest.raises(AssertionError):
    model, optim, _,_ = deepspeed.initialize(args=args,
                                             model=model,
```
and end the test after that point.

CalogeroZarbo:
@ShadenSmith Sure, on it! May I add the test to the same `test_fp16.py` file?

Thank you!
Cal

ShadenSmith:
That would be perfect. Just right under the other ZeRO tests.

CalogeroZarbo:
@ShadenSmith `pytests` are done. I added a helper class to define a custom optimizer on the fly. Thank you for the suggestion. 
Let me know if I can help further on.

Cheers,
Cal

tjruwase:
@CalogeroZarbo  Thanks so much for doing this.

ShadenSmith:
Thanks so much @CalogeroZarbo !

PR #181, opened 2020-04-03
add representation to optimizer

### LINKED ISSUE #180: Add representation method to optimizer like pytorch optimizer
Just an Idea, so Don't hesitate to close if you disagree and priority is so low.

I suggest that optimizer should have `__repr__`  method like[ pytorch's optimizer](https://github.com/pytorch/pytorch/blob/master/torch/optim/optimizer.py#L63) to see the parameters of optimizer human-readable.

kouml:
Fixes #180 .

add representation to DeepSpeed optimizer like apex optimizer which repr method is came from pytorch optimizer.


tjruwase:
Thanks for the PR!

Issue #129, opened 2020-03-11
Zero Optimizer crashes if a param_group has fewer number of elements than the data parallel size


samyam:
If a param_group has fewer parameter elements than the size of data parallelism, then certain GPUs will have empty partitions causing it to crash

PR #202, opened 2020-04-23
Fix index out of range error in ZeRO optimizer

tjruwase:
Fix index out of range error when parameter count is not multiple of ranks. 
Closing #129


ShadenSmith:
Changes look fine to me. If there's a feasible way to add a reproducing unit test, I think that is a good idea too.

tjruwase:
Thanks for the suggestion. I will work on that.

ShadenSmith:
Awesome, thanks Tunji!

Issue #99, opened 2020-02-21
DeepSpeed using DistributedSampler with model parallelism


ShadenSmith:
DeepSpeed's data loader will use `DistributedSampler` by default unless another is provided:

https://github.com/microsoft/DeepSpeed/blob/001abe2362d9edba062070fb05df40925f54cb3e/deepspeed/pt/deepspeed_dataloader.py#L43

If DeepSpeed is configured with model parallelism, or called from a library with a sub-group of the world processes, the default behavior of `DistributedSampler` is incorrect because it queries the global world size and rank information. We should specify `num_replicas` and `rank` when creating the sampler.

If `mpu` is provided to `deepspeed.initialize()`, we should query `mpu.get_data_parallel_world_size()` and `mpu.get_data_parallel_rank()` and forward that information to the sampler.

PR #216, opened 2020-05-15
Specify num_replicas and rank when creating sampler

vidushv:
Addressing issue #99. If `mpu` is provided to `deepspeed.initialize()`, then pass on the world size and rank information to `DistributedSampler` to ensure multiple processes will load the appropriate subset of the original dataset.


deepspeed/pt/deepspeed_dataloader.py
```diff
@@ -16,13 +16,15 @@ def __init__(self,
...
+                 data_sampler=None,
+                 data_parallel_world_size=None,
+                 data_parallel_rank=None):
         self.tput_timer = tput_timer
         self.batch_size = batch_size
 
         if local_rank >= 0:
             if data_sampler is None:
-                data_sampler = DistributedSampler(dataset)
+                data_sampler = DistributedSampler(dataset, data_parallel_world_size, data_parallel_rank)
```
ShadenSmith, line 27: Would you mind expanding the positional arguments into named ones for safety/clarity?

```python
DistributedSampler(dataset=dataset,
                   num_replicas=data_parallel_world_size,
                   rank=data_parallel_rank)
```
The formatter might have something to say about the line wrapping after that :-).

vidushv: Good call, just added the new commit!

ShadenSmith:
Thanks @vidushv! This looks good. I'll merge after the tests pass.

PR #228, opened 2020-05-25
fix redundant reduces

chunyang-wen:


ShadenSmith:
Thank you for your contribution to DeepSpeed!

PR #230, opened 2020-05-26
Add log util

chunyang-wen:
+ Add a logger for `DeepSpeed`. It is used to replace `print` clauses. Be more informative.
+ Doc for `pprint`. Actually I think it is confusing because there is python system package called pprint.

image


ShadenSmith:
Hi @chunyang-wen, thanks so much for putting this together. This is absolutely an improvement.

Would you mind updating other parts of the codebase to use your new logger? I believe there are some other raw `print()` and `logging.foo()` statements in `deepspeed_light.py` and elsewhere.

Second, definitely agree about `pprint()`. One thought is to move the function to `log_utils.py` and instead accept an optional parameter that specifies the ranks to print. Something like this:

```python
def log_dist(msg,
             print_rank=0,
             level=logging.INFO):
    pass
```

It would then follow the behavior of `pprint` except we can configure which rank to print. Perhaps `-1` to specify all ranks? For bonus points we could also accept a list of ranks to print (this is useful when there are hybrid forms of parallelism. Lastly, since it's not hardcoded to rank 0 I think it would be nice to prepend something like "[RANK={dist.get_rank()}]" to `msg` to improve logging?

What do you think?

chunyang-wen:
Sure. But I think it is better we replace `print` step by step (For example a single or a couple of files each time). As you know, code is consistently changing. It is easy to conflict.

chunyang-wen:
Oops. Conflict shows up.

ShadenSmith:
Thanks @chunyang-wen! I fixed the conflict and will merge once the tests pass.

deepspeed/pt/deepspeed_light.py
```diff
@@ -83,10 +85,10 @@ def _initialize_parameter_parallel_groups(parameter_parallel_size=None):
 
 
 def print_configuration(args, name):
-    print('{}:'.format(name), flush=True)
+    logger.info('{}:'.format(name))
```
ShadenSmith, line 88: I think these logs are problematic and the logging prefix destroys formatting.

deepspeed/pt/deepspeed_light.py
```diff
@@ -83,10 +85,10 @@ def _initialize_parameter_parallel_groups(parameter_parallel_size=None):
...
     for arg in sorted(vars(args)):
         dots = '.' * (29 - len(arg))
-        print('  {} {} {}'.format(arg, dots, getattr(args, arg)), flush=True)
+        logger.info('  {} {} {}'.format(arg, dots, getattr(args, arg)))
```
ShadenSmith, line 91: I think these logs are problematic and the logging prefix destroys formatting.

deepspeed/pt/deepspeed_config.py
```diff
@@ -421,13 +421,13 @@ def _do_sanity_check(self):
         self._do_warning_check()
 
     def print(self, name):
-        print('{}:'.format(name), flush=True)
+        logger.info('{}:'.format(name))
         for arg in sorted(vars(self)):
             if arg != '_param_dict':
                 dots = '.' * (29 - len(arg))
-                print('  {} {} {}'.format(arg, dots, getattr(self, arg)), flush=True)
+                logger.info('  {} {} {}'.format(arg, dots, getattr(self, arg)))
```
ShadenSmith, line 428: I think these logs are problematic and the logging prefix destroys formatting.

ShadenSmith:
Sorry @chunyang-wen, we spotted a formatting issue in two spots. Here's a sample of what the logging prefix does to some of the configuration. It also looks like some of the log statements are printing twice?

image

chunyang-wen:
> Sorry @chunyang-wen, we spotted a formatting issue in two spots. Here's a sample of what the logging prefix does to some of the configuration. It also looks like some of the log statements are printing twice?
> 
> **Edit:** it looks like the real issue here is double-logging? I see those logged statements in two different formats.
> 
> image

I will try to fix that.

deepspeed/pt/deepspeed_config.py
chunyang-wen: As long as we append the same prefix, the style is still consistent?

deepspeed/pt/deepspeed_light.py
chunyang-wen: I will check the log about the output.

jeffra:
> Would you please help me to find the logs? I cannot see other logs in the azure pipeline besides the command line output.

@chunyang-wen in the Azure Pipeline for a complete run (such as this one: https://dev.azure.com/DeepSpeedMSFT/DeepSpeed/_build/results?buildId=1381&view=results) we upload the log files for our model tests as a published artifact. You can find this at the earlier link by clicking on "1 published". Also here's a direct link to download the logs from your most recent completed run: https://artprodcus3.artifacts.visualstudio.com/A7dfb693e-97bd-4877-8c5e-3b853dae0fad/aa0b43b4-be2e-45da-91df-45b48d6530be/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL0RlZXBTcGVlZE1TRlQvcHJvamVjdElkL2FhMGI0M2I0LWJlMmUtNDVkYS05MWRmLTQ1YjQ4ZDY1MzBiZS9idWlsZElkLzEzODEvYXJ0aWZhY3ROYW1lL0JpbmdCZXJ0U3F1YWRfbG9ncw2/content?format=zip

deepspeed/pt/deepspeed_config.py
ShadenSmith: Agreed, the issue was that it was double-printing (now fixed I see) and since they had different prefixes, it was no longer readable. Here's the new result:

image

ShadenSmith:
Thanks @chunyang-wen !

chunyang-wen:
> https://dev.azure.com/DeepSpeedMSFT/DeepSpeed/_build/results?buildId=1381&view=results

Thanks.  New skills learned.

PR #232, opened 2020-05-26
Support fp32 grad clipping and fix max_grad_norm confusion

jeffra:


tests/unit/test_fp16.py
```diff
@@ -8,17 +8,51 @@
...
+                                             model_parameters=model.parameters())
+        data_loader = random_dataloader(model=model,
+                                        total_samples=50,
+                                        hidden_dim=hidden_dim,
+                                        device=model.device)
+        for n, batch in enumerate(data_loader):
+            loss = model(batch[0].float(), batch[1])
+            model.backward(loss)
+            model.step()
+
```
samyam, line 41: Dont we need to compare the loss with a reference? or is the test to just ensure that the parameters can be passed?

jeffra: That sort of test is done in our model tests, this will just test to make sure everything is train-able. So should be fine :) Would be great to add loss checking in our unit tests though.

PR #234, opened 2020-05-27
remove redundant init code

chunyang-wen:
Remove redundant initialization code.

By the way, the default value of `consecutive_hysteres`  is different in the `__init__` and `update_scale` function. But `__init__` takes precedency.


ShadenSmith:
Great, I'll merge once the tests from the updated `master` complete.

PR #236, opened 2020-05-27
if using lamb force legacy_fusion=True

jeffra:


deepspeed/pt/deepspeed_light.py
```diff
@@ -549,8 +549,7 @@ def _configure_fp16_optimizer(self, optimizer):
                 dynamic_loss_args=dynamic_loss_args,
                 mpu=self.mpu,
                 clip_grad=clip_grad,
-                fused_lamb_legacy=self.optimizer_legacy_fusion()
-                if self.optimizer_name() == LAMB_OPTIMIZER else False)
+                fused_lamb_legacy=self.optimizer_name() == LAMB_OPTIMIZER)
```
samyam, line 552: Are we getting rid of the optimizer_legacy_fusion option id the ds_config? Which is probably fine given no one would really know what it means.

jeffra: I think the only time someone would need it is if they want to use DeepSpeed with an older version of apex (which would be a non-standard install I'd say). We can probably get rid of it then? @tjruwase what do you think?

tjruwase: Yes, this is only there to support old apex FusedAdam. Since we have updated the apex requirements of DeepSpeed, it is fine for us to remove it.

jeffra: Sounds good :) I'll do that in a future PR, will merge this for now.

PR #240, opened 2020-05-27
Support loading and saving ZeRO checkpoints with changing DP degree

tjruwase:
Support loading/saving ZeRO checkpoints with changing data parallelism degree, GPU counts. This will be useful for finetuning and continued training on different number of GPUs.


jeffra:
This is just for ZeRO stage 2 right? Should we add an assert somewhere if that it's not supported for ZeRO stage 1?

tjruwase:
@jeffra There is nothing specific to ZeRO-2 in these changes, and so we could easily add them to ZeRO-1. But I thought we wanted to retire the old ZeRO-1 implementation, which is why I left it out. Is that no longer the case?

jeffra:
@tjruwase, I think we're planning to support ZeRO-1 still? We should probably have some discussions to figure out a good plan here though. Perhaps merging the implementations so that as we add features like this it's easier to support both stage 1 and 2. One example where ZeRO-1 might still be useful are cases where we need gradient accumulation, ZeRO-2 doesn't support this.

deepspeed/pt/deepspeed_zero_optimizer.py
```diff
@@ -81,11 +81,21 @@ def flatten_dense_tensors_aligned(tensor_list, alignment, pg):
...
+    return (alignment - remainder) if remainder else remainder
+
+
 def move_to_cpu(tensor_list):
     for tensor in tensor_list:
         tensor.data = tensor.data.cpu()
 
 
+def print_rank_msg(msg):
+    print(f"rank {dist.get_rank()} - {msg}")
```
chunyang-wen, line 96: a lot of similar print-related definitions.

samyam:
> @tjruwase, I think we're planning to support ZeRO-1 still? We should probably have some discussions to figure out a good plan here though. Perhaps merging the implementations so that as we add features like this it's easier to support both stage 1 and 2. One example where ZeRO-1 might still be useful are cases where we need gradient accumulation, ZeRO-2 doesn't support this.

I agreed. We still want ZeRO-1 where we want to use gradient accumulation. This is specially true for the pipeline parallelism implementation that @ShadenSmith  is working on.

deepspeed/pt/deepspeed_zero_optimizer.py
tjruwase: Will take care of this in a different PR.

tests/model/Megatron_GPT2/run_checkpoint_test.py
```diff
@@ -216,9 +319,13 @@ def check_parity(self, base_file, test_file, r_tol):
...
 def checkpoint_suite():
     suite = unittest.TestSuite()
-    suite.addTest(GPT2CheckpointTestCase('test_mp4_gpu16_node1_with_zero1'))
-    suite.addTest(GPT2CheckpointTestCase('test_mp4_gpu16_node1_with_zero2'))
-    suite.addTest(GPT2CheckpointTestCase('test_mp4_gpu16_node1_without_zero'))
+    suite.addTest(GPT2CheckpointTestCase('test_mp2_gpu4_node1_with_zero1'))
+    suite.addTest(GPT2CheckpointTestCase('test_mp2_gpu4_node1_with_zero2'))
+    suite.addTest(GPT2CheckpointTestCase('test_mp1_gpu2_load_gpu1_node1_with_zero1'))
+    suite.addTest(GPT2CheckpointTestCase('test_mp1_gpu2_load_gpu4_node1_with_zero1'))
+    suite.addTest(GPT2CheckpointTestCase('test_mp1_gpu2_load_gpu1_node1_with_zero2'))
```
samyam, line 326: Can we add one test case with mp=2, dp=2 loaded to mp=2, dp=4 or 1? Seems like we are not testing any cases with mp > 1 when the dp changes?

tjruwase: Good point. Will do;

deepspeed/pt/zero_optimizer_stage1.py
```diff
@@ -775,12 +876,136 @@ def state_dict(self):
...
+        for group_idx in range(len(self.optimizer.param_groups)):
+            all_partition_group_states = [
+                sd['base_optimizer_state'][group_idx] for sd in state_dict_list
+            ]
+            group_optimizer_states = self._retrieve_group_optimizer_states(
+                all_partition_group_states)
+            base_optimizer_group_states.append(group_optimizer_states)
+
+        for group_idx, group in enumerate(self.optimizer.param_groups):
+            for param_idx, param in enumerate(group['params']):
```
samyam, line 1000: Why do we loop over group['params'] is stage 1 and just update group['params'][0] only in stage 2?

kisseternity:
Hello, is this feature still effective in the latest deepspeed 0.7.0+ versions? I think it's useful in the cloud training, where
gpu numbers as well as dp degree might change quite often. As I tried in 0.7.0 version with ZeRO1 from 8 GPUs to 16 GPUs, there was error suggesting that the other 8 gpus couldn't find the corresponding zero checkpoint.

tjruwase:
@kisseternity, you are correct. Unfortunately, we disabled this feature because it turned out not be robust. We are working on a more general replacement.

kkteru:
Hi, is any replacement available for this now? Would appreciate any pointers!

PR #249, opened 2020-06-02
Support migration to FP16 optimizer 

tjruwase:
1) Enable initializing fp32 params from non-DeepSpeed checkpoint
2) Avoid repeated gradient norm computation


deepspeed/pt/deepspeed_light.py
```diff
@@ -164,6 +164,9 @@ def __init__(self,
         self.training_dataloader = self.deepspeed_io(
             training_data) if training_data else None
 
+        # Configure wall clock timer
+        self.timers = SynchronizedWallClockTimer()
```
eltonzheng, line 168: self.timers already exists, maybe need a rebase?

deepspeed/pt/fp16_optimizer.py
```diff
@@ -314,6 +351,10 @@ def state_dict(self):
         state_dict['clip_grad'] = self.clip_grad
         return state_dict
 
+    def refresh_fp32_params(self):
```
eltonzheng, line 354: where is this function called?

deepspeed/pt/fp16_optimizer.py
```diff
@@ -182,9 +205,14 @@ def step(self, closure=None):
 
             self.fp32_groups_flat[i].grad = grads_groups_flat[i]
 
-            norm_groups.append(get_grad_norm(self.fp32_groups_flat, mpu=self.mpu))
+        self.start_timers([COMPUTE_NORM])
+        all_groups_norm = get_grad_norm(self.fp32_groups_flat, mpu=self.mpu)
```
eltonzheng, line 209: good catch!

deepspeed/pt/fp16_optimizer.py
tjruwase: This will be called from the client function in the case where non-DeepSpeed checkpoint is loaded to initialize the model.

PR #250, opened 2020-06-03
Update paths; include information about submodule.

awan-10:
@jeffra -- please review the small changes and merge this request.

### DIFF
--- a/docs/_tutorials/bert-finetuning.md
+++ b/docs/_tutorials/bert-finetuning.md
@@ -7,7 +7,19 @@ In this tutorial we will be adding DeepSpeed to the BingBert model for the SQuAD
 
 ## Overview
 
-Please clone the repository and go to the `examples/BingBertSquad` folder to follow along.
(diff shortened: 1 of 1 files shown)


docs/_tutorials/bert-finetuning.md
```diff
@@ -7,7 +7,19 @@ In this tutorial we will be adding DeepSpeed to the BingBert model for the SQuAD
...
+
+`git clone https://github.com/microsoft/deepspeed`
+
+`cd deepspeed`
+
+The DeepSpeedExamples are submodules so you need to initialize and update them using the following commands
+
+`git submodule init`
+
+`git submodule update`
```
jeffra, line 20: We could also do this in a single command (this is what we do in our install script also): `git submodule update --init --recursive`.

awan-10: I agree. Both methods are fine so I guess I will keep the ones I have and mark this as resolved.

PR #258, opened 2020-06-08
install update: no-sudo + clean build files

jeffra:


install.sh
```diff
@@ -18,7 +18,9 @@ hostfile (hostfile: /job/hostfile). If no hostfile exists, will only install loc
     -d, --deepspeed_only    Install only deepspeed and no third party dependencies
     -t, --third_party_only  Install only third party dependencies and not deepspeed
     -l, --local_only        Install only on local machine
-    -s, --pip_sudo          Run pip with sudo (default: no sudo)
+    -s, --pip_sudo          Run pip install with sudo (default: no sudo)
+    -r, --allow_sudo        Force all sudo for entire script (probably don't want this, instead use --pip_sudo)
```
ShadenSmith, line 22: Maybe change the documentation to "Allow script to be run by root", since on its own the flag does not actually take action.

PR #263, opened 2020-06-12
remove transformer layer ID from the top module

RezaYazdaniAminabadi:


deepspeed/pt/deepspeed_cuda.py
```diff
@@ -403,11 +403,14 @@ class DeepSpeedTransformerLayer(nn.Module):
 
             initial_biases: Optional: Only used for unit test
     """
-    def __init__(self, layer_id, config, initial_weights=None, initial_biases=None):
+    layer_id = 0
+
```
chunyang-wen, line 407: Docstring should be updated accordingly.

RezaYazdaniAminabadi: sure, I will do that. thanks

deepspeed/pt/deepspeed_cuda.py
```diff
@@ -393,21 +393,24 @@ def backward(ctx, grad_output):
 class DeepSpeedTransformerLayer(nn.Module):
     """Initialize the DeepSpeed Transformer Layer.
 
+        Static variable:
```
eltonzheng, line 396: The API doc (https://deepspeed.readthedocs.io/en/latest/kernel.html) will auto retrieve the strings here, if "layer_id" is invisible to users,  shouldn't add it here.

RezaYazdaniAminabadi: good point. I will remove it then

PR #268, opened 2020-06-17
new transformer pre-ln image

ShadenSmith:
This is the new figure provided by @RezaYazdaniAminabadi


ShadenSmith:
Administrative merging as Reza approved yesterday.

PR #276, opened 2020-06-25
Support distributed init outside runtime init

jeffra:
Allow users to initialize torch distributed outside of deepspeed.initialize. Specifically needed for MPI discovery (e.g., AML) and other models. This also removes the need for --deepspeed_mpi flag, we will automatically attempt to discover MPI world info if we are not launched with a torch.distributed or deepspeed launcher.

TODO: Need to update documentation, remove --deepspeed_mpi flag, and add unit tests


jeffra:
Still need to update documentation and add unit tests

andrewchoi5 (approved):

jeffra:
Closing and moving to #608

PR #277, opened 2020-06-25
Module replacement support

jeffra:
In this PR, we add a feature inside deepspeed to help automate the use of our different ops at the modeling side. Instead of changing the modeling file to add the deepspeed ops, we inject it from the run-script by replacing the module that deepspeed-op intends to accelerate. Here, we show how we do this for the deepspeed-transformer kernel, but this can be used for the new ops that we will add to deepspeed. For how we use the injection method, I will create an example branch that does the injection for both Nvidia and HuggingFace bert-modeling.


deepspeed/pt/deepspeed_cuda.py
```diff
@@ -95,7 +98,8 @@ def __init__(self,
                  gelu_checkpoint=False,
                  adjust_init_range=True,
                  attn_dropout_checkpoint=False,
-                 stochastic_mode=False):
+                 stochastic_mode=False,
+                 huggingface=False):
```
ShadenSmith, line 102: I am a little nervous using `huggingface` as a boolean flag because the API gets worse after we support another target replacement.

Since this flag only changes `forward()`'s return type -- `(output, )` instead of `output` -- we could name this flag more specifically like `return_tuple` ? Then in the docstring we state it's for HF compatibility.

Alternatively, we should plan for variations in each transformer implementation and treat this as more of a base class that specific implementations specialize around. From above, the HF specialization could be something like this:

```python
class HFTransformerLayer(TransformerLayer):
    def forward(...):
        output = super().forward(..) # this applies our transformer function
        return (output, ) # wrap in a tuple
```

@tjruwase has some great ideas on formalizing these replacement interfaces in order to simplify adding new types.

deepspeed/pt/deepspeed_cuda.py
```diff
@@ -393,21 +401,24 @@ def backward(ctx, grad_output):
 class DeepSpeedTransformerLayer(nn.Module):
     """Initialize the DeepSpeed Transformer Layer.
 
+        Static variable:
+            layer_id: The layer-index counter starting from 0 and incrementing by 1 every time a layer object is instantiated,
```
ShadenSmith, line 405: Will users need to use `layer_id`? If it's just an implementation detail, this might be better as a comment instead of the public docstring.

deepspeed/pt/deepspeed_cuda.py
```diff
@@ -393,21 +401,24 @@ def backward(ctx, grad_output):
...
             config: An object of DeepSpeedTransformerConfig
 
             initial_weights: Optional: Only used for unit test
 
             initial_biases: Optional: Only used for unit test
     """
-    def __init__(self, layer_id, config, initial_weights=None, initial_biases=None):
+    layer_id = 0
+
+    def __init__(self, config, initial_weights=None, initial_biases=None):
```
ShadenSmith, line 416: Can the unit tests instead allocate the layer and then replace the weights/biases? I would prefer to keep these details from complicating our API. They also result in a big branch in the code just below.

deepspeed/pt/deepspeed_cuda.py
jeffra: @RezaYazdaniAminabadi thoughts here?

deepspeed/pt/replace_module.py
```diff
@@ -0,0 +1,199 @@
+import copy
+import torch
+import deepspeed
+
+
+def replace_transformer_layer(orig_layer_impl,
```
ShadenSmith, line 6: Same comment as above. When `huggingface=False`, which transformer implementation are we assuming?

This seems like the place where @tjruwase 's idea could be implemented. Totally brainstormed and untested:

```python
from abc import ABC
class TransformerReplacementPolicy(ABC):
    @classmethod
    def replace(other):
        pass
        # the main replacement body here, but use methods like self.query(other)

    @abstractmethod    
    def query(self, source):
        """"Should return a tuple of the weight and bias for the query. """
        pass

class HFTransformerReplacement(TransformerReplacementPolicy):
    def query(self, source):
        qw = source.self.query.weight
        qb = source.self.query.bias
        return (qw, qb)
```

deepspeed/pt/deepspeed_cuda.py
RezaYazdaniAminabadi: Yes, I agree with Shaden. That can just a comment. We can remove the detail from public docstring.

deepspeed/pt/deepspeed_cuda.py
RezaYazdaniAminabadi: The issue here is that we need to compare our implementation with baseline, and if we don't get it from upper level, there is no way that we have the same weights/biases. I remember we tried it in the beginning with @eltonzheng and we could not come up with any other solution of having similar weight as PyTorch. We even tried the same random seed but still the weights between our case and PyTorch was different!

deepspeed/ops/transformer/transformer.py
```diff
@@ -399,21 +432,25 @@ def backward(ctx, grad_output):
...
-    def __init__(self, layer_id, config, initial_weights=None, initial_biases=None):
+    layer_id = 0
+
+    def __init__(self, config, initial_weights=None, initial_biases=None):
         super(DeepSpeedTransformerLayer, self).__init__()
 
         self.config = config
-        self.config.layer_id = layer_id
+        self.config.layer_id = DeepSpeedTransformerLayer.layer_id
+        self.config.layer_id = DeepSpeedTransformerLayer.layer_id
```
eltonzheng, line 452: dup line?

deepspeed/module_inject/replace_module.py
```diff
@@ -0,0 +1,191 @@
+import copy
+import torch
+import deepspeed
+
+
+def replace_transformer_layer(orig_layer_impl,
```
eltonzheng, line 6: can we add a test case for this feature, to avoid being broken by future changes?

RezaYazdaniAminabadi: This means have a test-case that initially uses the bert-modeling and the inject the kernel using this function. Can you please write a test-bench to do this? I have already tested this with BingBertSquad and see similar results between baseline and when using the injection-code.
Thanks Elton.

deepspeed/ops/transformer/transformer.py
RezaYazdaniAminabadi: thanks I will remove this.

rocm-mici:

PR #290, opened 2020-07-15
updates to amp to support grad clip and grad accumulation

jeffra:
Fixes two potential issues with AMP backend as outlined here:
* https://nvidia.github.io/apex/advanced.html#gradient-accumulation-across-iterations
* https://nvidia.github.io/apex/advanced.html#gradient-clipping

h/t @Liyang90

### DIFF
--- a/deepspeed/pt/deepspeed_light.py
+++ b/deepspeed/pt/deepspeed_light.py
@@ -835,8 +835,8 @@ class DeepSpeedLight(Module):
                 if not self.fp16_enabled():
                     self.clip_fp32_gradients()
                 elif self.amp_enabled():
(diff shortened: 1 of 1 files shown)


deepspeed/pt/deepspeed_light.py
```diff
@@ -826,14 +831,22 @@ def step(self):
 
         if self.is_gradient_accumulation_boundary():
 
-            if not self.fp16_enabled() and self.gradient_clipping() > 0.0:
-                self.clip_fp32_gradients()
+            if self.gradient_clipping() > 0.0:
+                if not self.fp16_enabled():
```
Liyang90, line 835: amp mode would enter this condition as well.

PR #291, opened 2020-07-15
Empty grad fix

jeffra:
This fixes a case where there is an imbalance between empty gradients across ranks. Example: rank 0 has 2 parameters with gradients and rank 1 has 1 parameter with gradients and 1 parameter where grads are None. This caused the all-reduce to be imbalanced in size since we we're previously ignoring all grads that were None. Instead now we pad all None gradients with zero tensors. Unfortunately this imbalance did not cause a crash it would cause hanging issues in the first place that tried to access the reduced gradient data itself.

### DIFF
--- a/deepspeed/pt/deepspeed_light.py
+++ b/deepspeed/pt/deepspeed_light.py
@@ -980,6 +980,11 @@ class DeepSpeedLight(Module):
         grads = []
         for param_name, param in self.module.named_parameters():
             if param.grad is None:
+                # In cases where there is an imbalance of empty grads across
+                # ranks we must create empty grads, this will ensure that every
+                # rank is reducing the same size. In some cases it may make
+                # sense in the future to support the ability to average not
+                # w.r.t. world size but with a different value.
                 grads.append(
                     torch.zeros(param.size(),
                                 dtype=param.dtype,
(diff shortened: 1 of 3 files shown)


deepspeed/pt/deepspeed_light.py
```diff
@@ -979,7 +979,12 @@ def allreduce_no_retain(self, bucket, numel_per_bucket=500000000):
     def buffered_allreduce_fallback(self, grads=None, elements_per_buffer=500000000):
         grads = []
         for param_name, param in self.module.named_parameters():
-            if param.grad is not None:
+            if param.grad is None:
```
frankseide, line 982: May I suggest to add a comment here to explain why this change was necessary? It may be useful for future readers of the code who may not have had the experience we ran into. Thanks!

samyam, line 982: I wonder if averaging these gradients with world_size still makes sense when only some of the gradients are non-zero and others are none. Maybe something to just add a comment here ?

tests/unit/test_fp16.py
```diff
@@ -97,13 +98,14 @@ def test_lamb_fp16_empty_grad(tmpdir):
...
-    model = SimpleModel(hidden_dim, empty_grad=True)
+    model = SimpleModel(hidden_dim, empty_grad=True, rank=args.local_rank)
 
-    @distributed_test(world_size=[1])
+    @distributed_test(world_size=[2])
     def _test_lamb_fp16_empty_grad(args, model, hidden_dim):
         model, _, _,_ = deepspeed.initialize(args=args,
                                              model=model,
                                              model_parameters=model.parameters())
+        model.module.rank = args.local_rank
```
tjruwase, line 108: Why is this needed?

jeffra: good catch, it's not needed but forgot to remove it in my commit.

PR #296, opened 2020-07-21
only global rank 0 can log tensorboard data; avoid multi gpu/node race for the log directory

arashashari:


deepspeed/pt/deepspeed_light.py
```diff
@@ -144,12 +144,12 @@ def __init__(self,
         self._configure_with_arguments(args, mpu)
         self._do_sanity_check()
 
+        self._init_distributed(dist_init_required)
+
         self.sample_count = 0
-        if self.tensorboard_enabled():
+        if self.tensorboard_enabled() and self.global_rank == 0:
```
ShadenSmith, line 150: Perhaps initialize this to `None` on other ranks?

arashashari: I was thinking about resolving it [here](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/pt/deepspeed_light.py#L148) but was not sure if it is used anywhere else or not.

PR #305, opened 2020-08-01
NameError: name 'mpu' is not defined

MannyKayy:
mpu object is bound to the class instance.. 

the if statement uses  `self.mpu` but just `mpu` is called in the following lines.. 

This raises a NameError


tjruwase:
@MannyKayy  Thanks for this bug fix.

PR #307, opened 2020-08-07
Removing () from assertion in ZeRO-1.

ShadenSmith:
The parenthesis alter the evaluation of the assert() and it will always evaluate to `True`.

image


tjruwase:
Nice catch!

PR #313, opened 2020-08-11
ZeRO-2 enable CPU offload

jren73:


deepspeed/pt/deepspeed_light.py
```diff
@@ -533,8 +542,11 @@ def _configure_basic_optimizer(self, model_parameters):
                 "'max_grad_norm' is not supported as an optimizer parameter, please switch to using the deepspeed parameter 'gradient_clipping' see: https://www.deepspeed.ai/docs/config-json/#gradient-clipping for more details"
             )
         if self.optimizer_name() == ADAM_OPTIMIZER:
-            from apex.optimizers.fused_adam import FusedAdam
-            optimizer = FusedAdam(model_parameters, **optimizer_parameters)
+            if self.zero_cpu_offload():
```
ShadenSmith, line 545: I don't believe this code path can be reached because we would have already triggered the `zero_cpu_offload` condition in `_configure_optimizer()`.

jeffra:
It appears the formatting tests are failing, @jren73 can you install `pre-commit` and run the formatter outlined here: https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites

deepspeed/pt/deepspeed_zero_optimizer.py
```diff
@@ -1087,10 +1199,160 @@ def free_grad_in_param_list(self, param_list):
...
+            else:
+                single_grad_partition = _flatten_dense_tensors(
+                    self.averaged_gradients[i]).to(
+                        self.single_partition_of_fp32_groups[i].dtype)
+            assert single_grad_partition.numel() == self.partition_size[i], \
+                "averaged gradients have different number of elements that partition size {} {} {} {}".format(single_grad_partition.numel(), self.partition_size[i], i, partition_id)
+
+            #jie: transform the grad back to CPU
+            start, end = self.get_partition_fp32_group_index(i)
+            with torch.cuda.stream(self.migration_stream):
```
samyam, line 1266: I doubt this is overlapping with anything. This needs to be done directly during the backward pass if we want to overlap it with computation. We may want to revisit this to understand the performance implication. 

It probably ok for this PR though. Please just mark it as a TODO

deepspeed/pt/deepspeed_zero_optimizer.py
```diff
@@ -1087,10 +1199,160 @@ def free_grad_in_param_list(self, param_list):
         for p in param_list:
             p.grad = None
 
+    def get_partition_fp32_group_index(self, group_index):
+        start = self.first_offset[group_index]
+        end = start + self.partition_size[group_index]
+        return [int(start), int(end)]
+
+    #jie:
+    def step_with_cpuoffload(self, closure=None):
```
samyam, line 1208: Does it make sense to reuse most of the code in the step? It seems like the code all the way till Line 1261 is common and the code after line 1305 is common across both steps, so there is only about 40 lines of code that is different between the two step. Does it make sense to just put this part in two methods and differentiate them with 'if cpu_offload' ?

deepspeed/pt/deepspeed_zero_optimizer.py
```diff
@@ -95,10 +96,45 @@ def move_to_cpu(tensor_list):
...
+        return obj
+
+
+def async_copy_to(obj, dev, main_stream=None):
+    if torch.is_tensor(obj):
+        obj = Variable(obj)
+    if isinstance(obj, Variable):
+        target = torch.empty_like(obj, device=dev).copy_(obj)
+        if main_stream is not None:
+            target.data.record_stream(main_stream)
```
samyam, line 130: This is just a curiosity question: Do you need record_stream if you are doing explicity wait on this stream before accessing or deleting the target tensor?

deepspeed/pt/deepspeed_light.py
jren73: yes the zero_cpu_offload condition will in configure_optimizer(). I can delete the check in _configure_basic_optimizer

deepspeed/pt/deepspeed_zero_optimizer.py
jren73: Thank you for pointing this out! I was planning to make line 128 as
target = torch.empty_like(obj, device=dev).copy_(obj, non_blocking=True)
for async data copy

deepspeed/pt/deepspeed_zero_optimizer.py
```diff
@@ -252,15 +308,27 @@ def __init__(self,
...
-                [partition_id].clone().float().detach())
+            #that this process will update
+            if self.cpu_offload:
+                start = self.single_partition_of_fp32_groups_index
+                self.single_partition_of_fp32_groups_index += self.parallel_partitioned_fp16_groups[
+                    i][partition_id].numel()
+                end = self.single_partition_of_fp32_groups_index
+                #self.single_partition_of_fp32_groups[start:end] = torch.tensor(self.parallel_partitioned_fp16_groups[i][partition_id].detach(),
+                #                                                   dtype=self.single_partition_of_fp32_groups[i].dtype)
+                self.single_partition_of_fp32_groups[
```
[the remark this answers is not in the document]
jren73: We do not creating an fp32 copy since we do not want to waste GPU memory space. I test such usage offline with torch.cuda.memory_allocated() and torch.cuda.max_memory_allocated(), and it can convert fp16 to fp32 without extra memory usage

deepspeed/pt/deepspeed_cpu_adam.py
```diff
@@ -0,0 +1,200 @@
+import math
+import torch
+
+
+class CPUAdam(torch.optim.Optimizer):
```
ShadenSmith, line 5: Is it possible to implement this in a way that does not require reimplementing the optimizer? I know that it does not matter much at the moment since ZeRO is only implemented for Adam, but I think it's worthwhile future proofing if feasible. And avoiding code duplication is would be ideal.

To my understanding, the changes that this makes to Adam are concerned with getting flattened buffers on the CPU from ZeRO. Could we instead do this work on the ZeRO side so that `self.param_groups` and `self.state` appear normal to the optimizer? Then I think Adam could work unmodified? I'm not very close to this part of DeepSpeed; please tell me if this is nonsense :-).

Knowing that not every optimizer may play nicely and someone may want to modify something specifically for an optimizer, it would still be wonderful to avoid the reimplementation issue. I wonder if it is possible to implement this as a subclass and use `super()` to get extensibility:

```python
# base class
class CPUOptimizer(torch.optim.Optimizer):
    def step(data_from_zero):
        # sneaky replacement of CPU tensors
        cpu_stuff(data_from_zero)

        # now do the real step() work
        super().step()

# We can provide deepspeed.CPUAdam at first for convenience
# step() will first call the CPUOptimizer step() and then Adam step()"""
class CPUAdam(deepspeed.CPUOptimizer, torch.optim.Adam):
    pass

# potential client code:
class CPUFancyOptimizer(deepspeed.CPUOptimizer, FancyOptimizer):
    def step(data_from_zero):
        custom_copy_stuff()

        # do the rest. This will call CPUOptimizer step() and then FancyOptimizer's step()
        super().step(data_from_zero)
```

deepspeed/pt/deepspeed_zero_optimizer.py
```diff
@@ -182,10 +222,26 @@ def __init__(self,
...
 
+        self.total_params = 0
+        self.single_partition_of_fp32_groups_index = 0
+
+        if self.cpu_offload:
+            logger.info(f"cpu_offload enabled")
+            for i, param_group in enumerate(self.optimizer.param_groups):
+                self.total_params += sum([t.numel() for t in param_group['params']])
+            logger.info("Total # of params is {}".format(self.total_params))
+            self.single_partition_of_fp32_groups = torch.zeros(
```
ShadenSmith, line 237: I'm not close to the ZeRO codebase at all, so please correct me whenever applicable :-).

To my understanding, a lot of the ways that the CPU branches of the code diverge from the regular code path comes from `self.single_partition_of_fp32_groups` being a flattened buffer instead of a list of tensors.

I think it may be tricky to maintain/test/debug some of this design because the variable can be a different data structure depending on our configuration.

I get the performance advantages of keeping things in a nice flattened buffer. Would it be possible to instead allocate the flattened buffer and then make `self.single_partition_of_fp32_groups` a list of tensors that view the single buffer? Then there would be fewer code changes because it's the same data structure end-to-end and we're just changing the backing storage. I find this stuff super tricky and there may be some computation graph or other major consequences that I am missing.

deepspeed/pt/deepspeed_zero_optimizer.py
```diff
@@ -95,10 +96,45 @@ def move_to_cpu(tensor_list):
...
+def move_to_cuda(tensor_list, cuda_device):
+    for tensor in tensor_list:
+        tensor.data = tensor.data.to(cuda_device)
+
+
 def print_rank_msg(msg):
     print(f"rank {dist.get_rank()} - {msg}")
 
 
+def async_migrate_to(obj, dev, main_stream=None):
```
ShadenSmith, line 108: Can we have standalone unit tests for `async_migrate_to()` and `async_copy_to()`?

deepspeed/pt/deepspeed_zero_optimizer.py
```diff
@@ -1230,6 +1492,21 @@ def unscale_and_clip_grads(self, grad_groups_flat, norm_groups):
             else:
                 grad.data.mul_(1. / combined_scale)
 
+    def unscale_and_clip_grads_on_cpu(self, grad_groups_flat, norm_groups):
```
ShadenSmith, line 1495: I think this method is equivalent to calling
```python
self.unscale_and_clip_grads((self.single_partition_of_fp32_groups.grad, ), norm_groups)
```

By wrapping the single grad in a tuple of size 1 the original code will do the same thing.

deepspeed/pt/deepspeed_zero_optimizer.py
tjruwase: @samyam, I think we need the record_stream here. According to the [documentation](https://pytorch.org/docs/1.5.1/tensors.html?highlight=record_stream#torch.Tensor.record_stream), record_stream is the safe way to pass tensors from an allocating stream to a consuming stream. Jie and I agree that the `self.cpu_computation_stream` should passed in as the consuming stream.
On the other hand, the [wait_stream](https://pytorch.org/docs/1.5.1/cuda.html?highlight=wait_stream#torch.cuda.Stream.wait_stream) is needed to synchronize two streams in general.

tests/unit/test_checkpointing.py
```diff
@@ -254,8 +263,9 @@ def test_checkpoint_zero_optimizer(tmpdir, zero_stage):
             "enabled": True
         },
         "zero_optimization": {
-            "stage": zero_stage
-        },
+            "stage": zero_stage,
+            "zero_cpu_offload": use_cpu_offload
```
jeffra, line 267: I think the json config key is named `cpu_offload` above. Can we update them across the tests? This probably means the tests aren't being run with cpu offload turned on currently.

tests/unit/test_checkpointing.py
```diff
@@ -339,7 +370,8 @@ def test_checkpoint_lr_scheduler(tmpdir, zero_stage):
             "enabled": True
         },
         "zero_optimization": {
-            "stage": zero_stage
+            "stage": zero_stage,
+            "zero_cpu_offload": use_cpu_offload
```
jeffra, line 374: zero_cpu_offload -> cpu_offload

tests/unit/test_checkpointing.py
```diff
@@ -391,7 +434,8 @@ def test_checkpoint_no_lr_scheduler(tmpdir, zero_stage):
...
```
jeffra, line 438: zero_cpu_offload -> cpu_offload

deepspeed/pt/deepspeed_cpu_adam.py
tjruwase: Thanks for this feedback. We now use torch.optim.Adam directly.

tjruwase:
> Can we add an assert that if cpu_offload=True and zero stage=1 is set that we throw an assert? Similar to this: https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/pt/deepspeed_light.py#L588

Done by if cpu_offload==True then assert zero_stage==2

deepspeed/pt/deepspeed_zero_optimizer.py
tjruwase: Done

deepspeed/pt/deepspeed_zero_optimizer.py
tjruwase: This is TODO for a follow on PR.

PR #315, opened 2020-08-13
Update fan out flag for pdsh

jeffra:
The default max fan out of pdsh is 32 nodes, we've run into this in the past and manually patched but never propagated to the repo. In the future when supporting different launcher backends we will have support for passing launcher args which would help improve this. However, currently this flag is hardcoded which is not ideal.


ShadenSmith:
Looks good, as you said we can deal with configurability soon when we're revisiting the launcher backend anyway.

Do I understand correctly that this PR increases the limit to 1024 _nodes_ (not devices), since we launch pdsh workers on a per-node basis (which then spawn local processes)? A comment in the code would be nice to specify what exactly pdsh fanout means to DeepSpeed.

PR #318, opened 2020-08-14
Fix zero-1 bug related to stitching reduced grads across comm partitions

jeffra:
Has been confirmed from 1p user to have fixed their issue. Also our original repro of the diverged loss shows the issue has fixed as well.

I believe the issue was related to the old copy gradients code was not stitching the sub-partitions together correctly. After simplifying this part of the code substantially this issue appears to be fixed.

### DIFF
--- a/deepspeed/runtime/zero/stage1.py
+++ b/deepspeed/runtime/zero/stage1.py
@@ -249,8 +249,7 @@ class FP16_DeepSpeedZeroOptimizer_Stage1(object):
 
             # RS: divide up the sub-partitions and keep track of offsets for each param
             # partition_size = len(self.fp16_groups_flat[i]) / dist.get_world_size(group=self.dp_process_group)
-            params_in_rank_sub_partition, params_in_rank_sub_partitions_offsets, \
-            params_not_local = self.get_all_sub_partition_info(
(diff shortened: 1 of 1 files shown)


deepspeed/pt/zero_optimizer_stage1.py
```diff
@@ -598,28 +597,20 @@ def reduce_scatter_gradients(self,
...
-                    p.grad.copy_(updated_grad[0])
+            # stitch together all rank sub partitions for each comm idx
+            flat_comm_grads = []
+            for comm_idx, rank_partitions in enumerate(all_comm_partitions):
+                flat_comm_grads.append(torch.cat(rank_partitions))
+
+            flat_all_grads = torch.cat(flat_comm_grads)
+
+            # copy back reduced gradients but only those needed for this local rank
+            for param, updated_grad in zip(self.fp16_groups[i], _unflatten_dense_tensors(flat_all_grads, self.fp16_groups[i])):
```
samyam, line 608: By flattening the gradients aren't we creating an extra copy of the gradients? At this point are we going to have 3 (or 4) copies of the gradients? i) in param.grad, ii) in all_comm_partitions, iii) flattented gradients, (iv) gradients unflattened from the flattened gradients if _unflatten_dense_tensors is creating a new copy. ) 

We should probably release all param.grad before flattening, then delete all_comm_partitions after flattening. That way we will have at most one copy (or two) of the gradients alive at any given time?

PR #336, opened 2020-08-27
Update deepspeed_checkpointing.py

samyam:
DeepSpeedConfig in DeepSpeed Checkpointing was missing mpu during in initialization resulting in buggy world_size, micro_batch. train_batch_size and gradient accumulation step calculations when used with model parallelism and activation checkpointing.


jeffra:
Approved (after formatting fixes and tests pass)

PR #340, opened 2020-09-01
Improved testing matrix.

ShadenSmith:
This is a rehash of my older testing infrastructure ["upgrade"](https://github.com/microsoft/DeepSpeed/pull/213) that we reverted soon after. I think this is actually an upgrade this time :-).

This is allows us to spin up new testing nodes quickly when our queue times get long.

I left placeholders for temporarily disabled CUDA 10.1+ tests until we can specify constraints to farm tasks based on the available hardware and drivers.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -65,7 +65,7 @@ jobs:
 
   - script: |
       source activate $(conda_env)
-      pytest --forked --verbose tests/unit/
+      pytest --durations=0 --forked --verbose tests/unit/
     displayName: 'Unit tests'
 
   - script: |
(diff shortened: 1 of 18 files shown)


azure-pipelines.yml
```diff
@@ -52,35 +85,29 @@ jobs:
...
-  #  displayName: 'Megatron GPT2 log uploads'
-  #  condition: always()
 
-  #- task: PublishPipelineArtifact@1
-  #  inputs:
-  #    targetPath: '$(Build.SourcesDirectory)/tests/model/Megatron_GPT2/checkpoint_test_logs/'
-  #    artifactName: Megatron_GPT2_checkpoint_logs
-  #  displayName: 'Megatron GPT2 checkpoint log uploads'
-  #  condition: always()
+- job: Code_Quality_Checks
```
jeffra, line 91: Can we move this up to run before the actual tests so we can fail fast?

ShadenSmith:
This was brought in through the sparse attention release #342 . Closing.

PR #341, opened 2020-09-01
Switches BBS example to use mbsize=3 and gas=2 to fit in 16GB of memory.

ShadenSmith:


tjruwase:
What issue is this PR related to?

ShadenSmith:
We are ramping up additional testing machines with 16GB GPUs. BBS is the only one of our tests that requires >16GB of memory to complete. As an added bonus this now adds model test coverage for ZeRO-2 with gradient accumulation too :-).

Do you see any issues with decreasing the microbatch size for these model tests? The effective batch size is untouched.

tjruwase:
Thanks for the clarification. I don't see any issues with decreasing micro batch size. I was just curious about the motivation.

ShadenSmith:
Sure sure, was just wondering too :-). I haven't worked with this test much in the past.

PR #344, opened 2020-09-02
Add ZeRO-Offload checkpointing model tests

tjruwase:
Fix ZeRO-Offload checkpointing bug when changing gpu count
Add checkpointing model tests
Remove optimizer key in Megatron model tests

### DIFF
--- a/tests/model/Megatron_GPT2/ds_config_func_bs4_zero1.json
+++ b/tests/model/Megatron_GPT2/ds_config_func_bs4_zero1.json
@@ -3,14 +3,9 @@
   "gradient_accumulation_steps": 1,
   "steps_per_print": 1,
   "zero_optimization": {
-    "stage":1
+    "stage": 1
   },
-  "optimizer": {
-    "type": "Adam",
-    "params": {
-      "lr": 0.00015
-    }
-  },
+  "zero_allow_untested_optimizer": true,
   "gradient_clipping": 1.0,
   "fp16": {
     "enabled": true,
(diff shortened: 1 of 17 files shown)


tests/model/Megatron_GPT2/ds_config_func_bs4_zero1.json
```diff
@@ -3,14 +3,9 @@
...
-    "stage":1
-  },
-  "optimizer": {
-    "type": "Adam",
-    "params": {
-      "lr": 0.00015
-    }
+    "stage": 1
   },
+  "zero_allow_untested_optimizer": true,
```
jeffra, line 8: I assume that since we're given a client provided optimizer we don't know that it's an Adam-based optimizer to know that it's safe for zero. Could we change the logic in the engine to run `isinstance(client_optimizer, apex.optimizers.fused_adam.FusedAdam)`? Then the Megatron json wouldn't need this untested optimizer flag. Also I think ZeRO would work with `torch.optim.Adam` as well or maybe not, I don't recall.

Also /cc @ShadenSmith, he mentioned to me having issues with our optimizer logic recently with PP.

tests/model/Megatron_GPT2/ds_gpt2_test.sh
```diff
@@ -91,9 +91,9 @@ gpt_options=" \
        ${ds_opt} \
        ${zero_opt} \
 "
-
+DEEPSPEED_PORT=29600
```
jeffra, line 94: The default master_port is 29500, just curious why we're changing it here? Not really a big deal, but just curious haha :)

tjruwase: Yes, this was to run model tests concurrently with unit tests :).

PR #347, opened 2020-09-03
Assert ZeRO-Offload incompatible with gradient accumulation

tjruwase:
Assert if gradient accumulation configured with ZeRO-Offload
Refactor ZeRO configuration and constants


jeffra (approved):
Looks good, thanks for doing this clean-up also!

PR #351, opened 2020-09-03
Jekyll installation instructions

ShadenSmith:
Filling in some more info to steamline our Jekyll setup.


samyam (approved):

cli99 (commented):
Got "Downloading jekyll-3.9.0 revealed dependencies not in the API or the lockfile (kramdown (< 3, >= 1.17)). Either installing with `--full-index` or running `bundle update jekyll` should fix the problem." when running bundle install. `bundle update jekyll` did fix it. Other than this, the instruction works fine.

PR #358, opened 2020-09-04
Import path fixes + conditional imports

jeffra:
We discovered very strange behavior where modules that were installed were not accessible from python. It seems this was related to circular imports. After switching to relative imports this seems to have fixed the problem.

In addition to this also added the ability for deepspeed to see what ops were installed from setup.py by exchanging state in the git_version_info module that is populated and install time.

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -15,7 +15,7 @@ from .ops.transformer import DeepSpeedTransformerLayer, DeepSpeedTransformerConf
 from .utils import logger
 
 try:
-    from deepspeed.git_version_info import version, git_hash, git_branch
+    from .git_version_info import version, git_hash, git_branch
 except ImportError:
     version = "0.0.0+unknown"
     git_hash = None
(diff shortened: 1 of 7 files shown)


RezaYazdaniAminabadi (approved):

deepspeed/__init__.py
```diff
@@ -4,13 +4,15 @@
...
+from .runtime.engine import DeepSpeedEngine
+from .runtime.engine import ADAM_OPTIMIZER, LAMB_OPTIMIZER
+from .runtime.lr_schedules import add_tuning_arguments
+from .runtime.config import DeepSpeedConfig
+from .runtime.activation_checkpointing import checkpointing
+from .ops.transformer import DeepSpeedTransformerLayer, DeepSpeedTransformerConfig
+from .utils import logger
 
 try:
     from deepspeed.git_version_info import version, git_hash, git_branch
```
ShadenSmith, line 18: Does this need to be updated to `.git_version_info` ?

ShadenSmith (approved):

PR #374, opened 2020-09-08
ZeRO-Offload passing model tests

tjruwase:
Add ZeRO-Offload model tests
Restrict optimizer update+copy to DeepSpeedCPUAdam

### DIFF
--- a/deepspeed/runtime/zero/stage2.py
+++ b/deepspeed/runtime/zero/stage2.py
@@ -158,7 +158,8 @@ class FP16_DeepSpeedZeroOptimizer(object):
 
         self.cpu_offload = cpu_offload
 
-        self.deepspeed_adam_offload = (cpu_offload and type(init_optimizer) == DeepSpeedCPUAdam)
+        self.deepspeed_adam_offload = (cpu_offload
+                                       and type(init_optimizer) == DeepSpeedCPUAdam)
 
         self.device = torch.cuda.current_device() if not self.cpu_offload else 'cpu'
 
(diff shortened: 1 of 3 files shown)


deepspeed/runtime/zero/stage2.py
```diff
@@ -15,6 +15,7 @@
 from deepspeed.runtime.fp16.loss_scaler import LossScaler, DynamicLossScaler
 from deepspeed.runtime.utils import see_memory_usage, is_model_parallel_parameter
 from deepspeed.runtime.zero.config import ZERO_OPTIMIZATION_GRADIENTS
+from deepspeed.ops.adam import DeepSpeedCPUAdam
```
RezaYazdaniAminabadi, line 18: I wonder if we need to control this import with a flag, since we may not install CPU-ADAM! We may just import it when cpu_offload is true. What do you think?

deepspeed/runtime/zero/stage2.py
```diff
@@ -1574,14 +1588,14 @@ def backward(self, loss, retain_graph=False):
...
-            buf_0 = torch.empty(int(self.reduce_bucket_size * 4.5),
+            buf_0 = torch.empty(int(self.reduce_bucket_size * 2.5),
                                 dtype=torch.half,
                                 device=torch.cuda.current_device())
             self.ipg_buffer.append(buf_0)
 
             # Use double buffers to avoid data access conflict when overlap_comm is enabled.
             if self.overlap_comm:
-                buf_1 = torch.empty(int(self.reduce_bucket_size * 4.5),
+                buf_1 = torch.empty(int(self.reduce_bucket_size * 2.5),
```
RezaYazdaniAminabadi: I think we ran this with 4.5 to get the best performance.

RezaYazdaniAminabadi (approved):

deepspeed/runtime/zero/stage2.py
tjruwase: Yes, I reverted this to 4.5, Also, it crashes on larger models with 2.5

deepspeed/runtime/zero/stage2.py
tjruwase: Sure, we can do that.

PR #383, opened 2020-09-09
support avx2 by default

RezaYazdaniAminabadi:
add DS_BUILD_AVX512 flag and update the feature part accordingly


jeffra (approved):
Looks great, thanks Reza!!

PR #409, opened 2020-09-14
support different intermediate size other than 4*hidden

RezaYazdaniAminabadi:


deepspeed/ops/transformer/transformer.py
```diff
@@ -26,6 +27,7 @@ def __init__(self,
         self.layer_id = -1
         self.batch_size = batch_size
         self.hidden_size = hidden_size
+        self.intermediate_size = intermediate_size if intermediate_size > 0 else 4 * hidden_size
```
tjruwase, line 30: Is there a benefit to allowing users to specify intermediate_size <= 0? I feel this could lead to confusion.

RezaYazdaniAminabadi: I add this condition for the cases that intermediate size is not specified and has the default value (-1). I just wanted that this works with the previous scripts which do not specify the intermediate_size. If that is still confusing, I can also remove it.

tjruwase (approved):

arashashari (approved):
@ShadenSmith @jeffra I wonder if the failing test is known and not related to the PR.

PR #411, opened 2020-09-15
pytest skips for tests requiring certain ops

jeffra:
This ensures users can run our tests even if they don't install any of our ops (e.g., `DS_BUILD_CUDA=0`)


ShadenSmith (approved):

arashashari (approved):
LGTM, thanks for handling it.

PR #419, opened 2020-09-17
Remove pip --use-feature

jeffra:
Several users have reported issues with this for some reason :(

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -154,9 +154,9 @@ else
 fi
 
 if [ "$pip_mirror" != "" ]; then
-    PIP_INSTALL="pip install --use-feature=2020-resolver -v -i $pip_mirror"
+    PIP_INSTALL="pip install -v -i $pip_mirror"
 else
-    PIP_INSTALL="pip install --use-feature=2020-resolver -v"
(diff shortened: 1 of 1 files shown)


arashashari (approved):

g-karthik:
@jeffra yeah `--use-feature` was causing some issues, but it seems they went away after upgrading `pip`.

Maybe you could enforce `pip` version constraints in case this flag is needed again in future?

ShadenSmith:
Hi @g-karthik, fortunately this feature just enabled what will become pip's dependency resolution mode next month. I enabled it here thinking it was present in more pip versions, but I can now see it isn't lol. We work with it enabled and disabled, which is the important thing. We shouldn't need to enable it again.

I'm in the middle of moving to a new place and will get back to your build update thread asap. @jeffra is also working through it as you can see 🙂. Thanks so much for the detailed feedback!

jeffra:
Hi @g-karthik, thanks so much for all your excellent feedback! We for some reason saw some internal users having issues with the flag also. I agree, ideally we could branch on pip version to enable the flag though.

PR #420, opened 2020-09-18
Activation checkpointing bugfix and unit tests

ShadenSmith:


jeffra (approved):

tests/unit/test_activation_checkpointing.py
```diff
@@ -0,0 +1,147 @@
+# TODO: add tests with model parallelism for activation partitioning and other features.
+
+from copy import deepcopy
+
+import pytest
+
+import torch
+
+import deepspeed
+ckpt = deepspeed.checkpointing.checkpoint
```
tjruwase, line 10: Out of curiosity, is there a difference to 
`import deepseed.checkpointing.checkpoint as ckpt`

tests/unit/test_activation_checkpointing.py
```diff
@@ -0,0 +1,147 @@
...
+class MaskedLinearSeq(MaskedLinear):
+    """Tests pipeline modules by also returning the mask."""
+    def forward(self, x, mask):
+        return super().forward(x, mask), mask
+
+
+class MaskedLinearSeqDup(MaskedLinearSeq):
+    """MaskedLinearSeq, but with more outputs than inputs and in a different order."""
+    def forward(self, x, mask):
+        dup = x.clone().detach() * 1000 * mask
```
tjruwase, line 79: What is 1000?

tests/unit/test_activation_checkpointing.py
ShadenSmith: I started with that (which is way nicer), but found that we can't actually do that today:

image

It's due to how we create `deepspeed.checkpoint` by importing from runtime in `deepspeed/__init__.py` versus having a `deepspeed/checkpoint/__init__.py` that imports from runtime.

I think it would be nice to change this. There's an alternative implementation with `deepspeed.pipe`, but the tradeoff is we have an extra `deepspeed/pipe/` directory hanging around.

tests/unit/test_activation_checkpointing.py
ShadenSmith: Just an arbitrary scaling to ensure the values are distinct from any others like x or some old outputs.

tests/unit/test_activation_checkpointing.py
tjruwase: Thanks for sharing this tip. I am glad I asked.

Issue #127, opened 2020-03-10
lr_scheduler unit tests


tjruwase:
We need unit tests to cover the following cases
1) scheduler is generating expected lr and mom values corresponding to global steps
2) scheduler is updating optimizer correctly

tjruwase:
Unit tests for expected lr values is handled by PR #339

PR #429, opened 2020-09-21
LR scheduler unit tests

tjruwase:
Add unit tests for LR Range Test and 1Cycle
Closing #127


tests/unit/test_lr_schedulers.py
```diff
@@ -205,3 +224,345 @@ def _test_lr_warmup_decay_schedule(args,
...
+                      model=model,
+                      hidden_dim=hidden_dim,
+                      min_lr=[min_lr],
+                      max_lr=[max_lr],
+                      step_size=step_size,
+                      decay_rate=decay_rate)
+
+
+@pytest.mark.parametrize("min_mom, max_mom, decay_rate, step_size",
+                         [
```
ShadenSmith, line 474: The code formatter is really wrecking your parametrization, sorry :-(. We should probably revisit its config as its line breaks are often pretty bad.

You may want to append ` # yapf: disable` to the expression to skip formatting. You can see the same thing in `test_config.py` and other places.

ShadenSmith (approved):
Thanks!!

tests/unit/test_lr_schedulers.py
tjruwase: Done. Thanks!

andrewchoi5 (approved):

PR #433, opened 2020-09-24
Removes cpufeature and defaults to AVX-2

ShadenSmith:


ShadenSmith:
@RezaYazdaniAminabadi can you take a look? This will use AVX-2 by default and users can opt into AVX-512 with the the environment variable `DS_BUILD_AVX512=1` at installation time. I'm thinking it may be better to use something like `DS_BUILD_SIMD={scalar,avx2,avx512}`? That leaves room to expand to other instruction sets in the future if needed.

One concern is defaulting to AVX-2 and it not being present at runtime. We could catch that and provide a useful error message?

RezaYazdaniAminabadi:
@ShadenSmith yes, I agree that having the three would even work better. I think this already should be fine in the code: https://github.com/microsoft/DeepSpeed/blob/master/csrc/adam/cpu_adam.cpp#L38. For the setup, we may wanna add the scalar one too, or even pass empty flag when neither of them is used. What do you think?

Issue #428, opened 2020-09-21
azure/readme.md  megatron tutorial link is broken


ivalmian:
Current link points to

https://github.com/microsoft/DeepSpeed/blob/master/tutorials/megatron

Which is broken

conglongli:
Fixed in https://github.com/microsoft/DeepSpeed/pull/436

PR #436, opened 2020-09-24
Fix urls in tutorial

conglongli:
Fix issues such as #428. The cause is that if we don't use the absolute url path, the links only work when accessed via deepspeed.ai. Now I changed the urls (including images) to absolute path for all tutorial pages.

### DIFF
--- a/docs/_tutorials/zero-offload.md
+++ b/docs/_tutorials/zero-offload.md
@@ -1,7 +1,7 @@
 ---
 title: "ZeRO-Offload"
 ---
-We recommend that you read the tutorials on [Getting Started](/getting-started/)  and [ZeRO](/zero/) before stepping through this tutorial.
+We recommend that you read the tutorials on [Getting Started](https://www.deepspeed.ai/getting-started/)  and [ZeRO](https://www.deepspeed.ai/tutorials/zero/) before stepping through this tutorial.
 
 ZeRO-Offload is a ZeRO optimization that offloads the optimizer memory and computation from the GPU to the host CPU. ZeRO-Offload enables large models with up to 13 billion parameters to be efficiently trained on a single GPU. In this tutorial we will use ZeRO-Offload to train a 10-billion parameter GPT-2 model in DeepSpeed. Furthermore, *using ZeRO-Offload in a DeepSpeed model is quick and easy because all you need is to change a few configurations in the DeepSpeed configuration json*. No code changes are needed.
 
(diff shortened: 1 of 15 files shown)


arashashari (approved):

RezaYazdaniAminabadi (approved):

conglongli:
After discussing with @ShadenSmith and @tjruwase, we decided that files in docs should have relative path in order to support local edits. As an alternative solution, we added a real readme under azure to just point to our tutorial on website. On the other hand, I fixed some other real path mistakes in tutorial.

docs/_tutorials/zero-offload.md
```diff
@@ -1,15 +1,15 @@
 ---
 title: "ZeRO-Offload"
 ---
-We recommend that you read the tutorials on [Getting Started](/getting-started/)  and [ZeRO](/zero/) before stepping through this tutorial.
+We recommend that you read the tutorials on [Getting Started](/getting-started/)  and [ZeRO](/tutorials/zero/) before stepping through this tutorial.
```
tjruwase, line 4: Thanks for fixing this.

docs/_tutorials/zero-offload.md
```diff
@@ -1,15 +1,15 @@
...
 
 ## Training Environment
-For this tutorial, we will configure a 10 billion parameter GPT-2 model using the DeepSpeed [Megatron-LM](https://github.com/microsoft/DeepSpeedExamples/tree/master/Megatron-LM) GPT-2 code. We advise stepping through the Megatron-LM [tutorial](/megatron/) if you have not previously done so. We will use a single [NVIDIA Tesla V100-SXM3 Tensor Core GPU](https://www.nvidia.com/en-us/data-center/v100/) with 32GB RAM for this exercise.
+For this tutorial, we will configure a 10 billion parameter GPT-2 model using the DeepSpeed [Megatron-LM](https://github.com/microsoft/DeepSpeedExamples/tree/master/Megatron-LM) GPT-2 code. We advise stepping through the Megatron-LM [tutorial](/tutorials/megatron/) if you have not previously done so. We will use a single [NVIDIA Tesla V100-SXM3 Tensor Core GPU](https://www.nvidia.com/en-us/data-center/v100/) with 32GB RAM for this exercise.
```
tjruwase, line 12: Thanks for fixing this

docs/_tutorials/zero-offload.md
```diff
@@ -28,7 +28,7 @@ We need to apply two changes to the launch script for the DeepSpeed Megatron-LM
        --cpu_optimizer \
 ```
 
-Most of the flags in the changes above should be familiar if you have stepped through the Megatron-LM [tutorial](/megatron/), except for the **_--cpu_optimizer_**. This flag informs the model script to pass a CPU-based Adam optimizer, rather than a GPU-based one, to DeepSpeed as the client optimizer. It is very important that this flag be used when training with ZeRO-Offload to ensure correct operation of the DeepSpeed engine.  
+Most of the flags in the changes above should be familiar if you have stepped through the Megatron-LM [tutorial](/tutorials/megatron/), except for the **_--cpu_optimizer_**. This flag informs the model script to pass a CPU-based Adam optimizer, rather than a GPU-based one, to DeepSpeed as the client optimizer. It is very important that this flag be used when training with ZeRO-Offload to ensure correct operation of the DeepSpeed engine.
```
tjruwase, line 31: Nice catch!

ShadenSmith (approved):

tjruwase (approved):
Thanks for catching the ZeRO/ZeRO-Offload errors.

PR #449, opened 2020-09-28
Use parentesis around min and max to enable Windows build

bratao:
Hello,

Visual C++ defines min and max as macros somewhere in windows.h, and they interfere the use of standard functions min and max.

With those changes I was able to build transformers and sparse_attention on Python 3.8 with a standard Visual Studio Build Tools 2019. I´m still working on changes to setup.py to make it work out of box, but this will be another PR.

### DIFF
--- a/csrc/includes/context.h
+++ b/csrc/includes/context.h
@@ -33,8 +33,8 @@
 
 inline int DS_GET_BLOCKS(const int N)
 {
-    return std::max(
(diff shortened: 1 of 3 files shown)


RezaYazdaniAminabadi (approved):

RezaYazdaniAminabadi:
Thanks Bruno! 👍

PR #450, opened 2020-09-29
Disable default installation of CPU Adam

tjruwase:
Disable default installation of CPU Adam

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,6 @@ import torch
 import shutil
 import subprocess
 import warnings
-import cpufeature
 from setuptools import setup, find_packages
 from torch.utils.cpp_extension import CUDAExtension, BuildExtension, CppExtension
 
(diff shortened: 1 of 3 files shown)


jeffra (approved):

setup.py
```diff
@@ -25,6 +24,19 @@ def fetch_requirements(path):
         return [r.strip() for r in fd.readlines()]
 
 
+def available_vector_instructions():
+    cpu_vector_instructions = {}
+    try:
+        import cpufeature
+        cpu_vector_instructions = cpufeature.CPUFeature
+    except _:
```
ShadenSmith, line 32: Could we first `except ImportError` to catch the case of `cpufeature` not being installed? It will only be available if users manually install our requirements first.

setup.py
```diff
@@ -25,6 +24,19 @@ def fetch_requirements(path):
...
+        warnings.warn(
```
ShadenSmith, line 33: Any idea if this will catch a segfault in the import? Otherwise we may need a signal handler instead :-(.

ShadenSmith (approved):

setup.py
tjruwase: Unfortunately, I don't know the answer to this question, especially since we can't repro the segfault. I guess I could create a nullptr deference in a C++ extension to test it.

tjruwase: I can't figure out how a signal handler in python can proceed after catching a segfault. Right now it simply replays the faulting instruction in the extension. I will punt on this for now.

setup.py
tjruwase: Done!

PR #458, opened 2020-10-01
Do not load optimizer states for fp32 if flag set to False

Zixxy:
Do not load optimizer states for fp32 if flag set to False.


ShadenSmith:
Hi @Zixxy, great catch! Thanks for the PR.

ShadenSmith (approved):

PR #462, opened 2020-10-06
gan tutorial

niumanar:


jeffra:
Can we add a pointer to the example code in https://github.com/microsoft/DeepSpeedExamples/tree/master/gan?

jeffra (approved):

PR #476, opened 2020-10-20
DeepSpeed op JIT support

jeffra:
* Support for both JIT and pre-install ops
* `pip install deepspeed` support
* Converts necessary apex ops into JIT (FusedAdam and flatten/unflatten) and removed apex dependency all together
* Streamlined ability to add new ops into DeepSpeed
* Allow for python only install
* Environment report `ds_report` reports op install status and compatibility


tests/unit/test_checkpointing.py
```diff
@@ -152,8 +152,8 @@ def checkpoint_correctness_verification(args,
         compare_lr_scheduler_states(trained_model, loaded_model)
 
 
-@pytest.mark.skipif(not deepspeed.ops.__installed_ops__['lamb'],
-                    reason="lamb is not installed")
+#@pytest.mark.skipif(not deepspeed.ops.__installed_ops__['lamb'],
```
ShadenSmith, line 155: Should these now be deleted? Or switch to a compatibility check? Same comment for other unit test changes.

ShadenSmith (commented):
We will also definitely need some documentation on the new changes :-).

csrc/adam/fused_adam_frontend.cpp
```diff
@@ -0,0 +1,20 @@
+#include <torch/extension.h>
```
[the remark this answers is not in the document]
jeffra: This file just exposes the front-end for `multi_tensor_adam` but doesn't exactly exist on the apex side, however it relates to this one: https://github.com/NVIDIA/apex/blob/master/csrc/amp_C_frontend.cpp

tests/unit/modelingpreln.py
```diff
@@ -363,10 +363,18 @@ def __init__(self, hidden_size, eps=1e-12):
             self.variance_epsilon = eps
 
         def forward(self, x):
+            pdtype = x.dtype
```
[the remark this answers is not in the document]
jeffra: This comes from @RezaYazdaniAminabadi, related to FusedLayerNorm. We currently don't have a deepspeed version of this op but we (previously) use it in our unit tests and in our bert example model. This fixes a bug we didn't realize we had (since we always previously had apex installed) that we need to also port to our bert example code.

deepspeed/runtime/zero/stage2.py
```diff
@@ -132,6 +125,12 @@ def __init__(self,
                  gradient_predivide_factor=1.0,
                  gradient_accumulation_steps=1):
 
+        global flatten, unflatten
+        if flatten is None or unflatten is None:
+            utils_op = UtilsBuilder().load()
```
ShadenSmith, line 130: Since this flatten/unflatten op logic is replicated in `engine.py` and `stage2.py`, maybe we should package it up and expose it as a util?

tests/unit/test_checkpointing.py
```diff
@@ -267,8 +269,8 @@ def _test_checkpoint_fused_optimizer(args, model, hidden_dim, load_optimizer_sta
                               'deepspeed_adam'),
                          ])
 def test_checkpoint_zero_optimizer(tmpdir, zero_stage, use_cpu_offload, adam_optimizer):
-    if use_cpu_offload and not deepspeed.ops.__installed_ops__['cpu-adam']:
-        pytest.skip("cpu-adam is not installed")
+    #if use_cpu_offload and not deepspeed.ops.__installed_ops__['cpu-adam']:
+    #    pytest.skip("cpu-adam is not installed")
```
ShadenSmith, line 273: these might be TODOs, just commenting so we remember there is some commented out skip logic.

op_builder/builder.py
```diff
@@ -0,0 +1,206 @@
...
+WARNING = f"{YELLOW} [WARNING] {END}"
+
+DEFAULT_TORCH_EXTENSION_PATH = "/tmp/torch_extensions"
+
+
+def assert_no_cuda_mismatch():
+    import torch.utils.cpp_extension
+    cuda_home = torch.utils.cpp_extension.CUDA_HOME
+    # Ensure there is not a cuda version mismatch between torch and nvcc compiler
+    output = subprocess.check_output([cuda_home + "/bin/nvcc",
```
ShadenSmith, line 20: This crashes on systems with out CUDA installed.

```
DS_BUILD_OPS=0
  File "setup.py", line 74, in <module>
    assert_no_cuda_mismatch()
  File "/home/shsmit/src/deepspeed/op_builder/builder.py", line 20, in assert_no_cuda_mismatch
    output = subprocess.check_output([cuda_home + "/bin/nvcc",
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

op_builder/builder.py
```diff
@@ -0,0 +1,206 @@
...
+        version_ge_1_3 = []
+        if (TORCH_MAJOR > 1) or (TORCH_MAJOR == 1 and TORCH_MINOR > 2):
+            version_ge_1_3 = ['-DVERSION_GE_1_3']
+        version_ge_1_5 = []
+        if (TORCH_MAJOR > 1) or (TORCH_MAJOR == 1 and TORCH_MINOR > 4):
+            version_ge_1_5 = ['-DVERSION_GE_1_5']
+        return version_ge_1_1 + version_ge_1_3 + version_ge_1_5
+
+    def is_compatible(self):
+        assert_no_cuda_mismatch()
```
ShadenSmith, line 195: I totally agree that a version mismatch should be a showstopper when we go to build an op. However, asserting here in the compatibility check seems very aggressive and stops us installing on non-CUDA systems.

jeffra:
Closed and replaced with #496 which is a squashed and rebased version of this PR.

PR #481, opened 2020-10-22
new version string formatting

ShadenSmith:


deepspeed/git_version_info.py
```diff
@@ -3,7 +3,7 @@
     from .git_version_info_installed import *
 except ModuleNotFoundError:
     # Will be missing from checkouts that haven't been installed (e.g., readthedocs)
-    version = '0.3.0+[none]'
+    version = '0.3.0'
```
jeffra, line 6: In this case we don't actually know what version was installed, maybe we should use a version like 0.0.0 here?

ShadenSmith: Yeah, good plan.

ShadenSmith: Actually, maybe we should parse `version.txt` here. This is the version string that will be displayed on our readthedocs page.

jeffra: We'll need to add it to our package data manifest I think for it to be accessible at runtime, but we could do that.

PR #482, opened 2020-10-22
add adamW to CPU-ADAM implementation

RezaYazdaniAminabadi:
I modified DeepSpeed CPU-Adam implementation to add the Adam-W option. I also add a new parameter for DeepSpeedCPUAdam, called adamw_mode, which is set to true by default to fix the zero-offload convergence issue. This is also consistent with the apex fused-adam implementation.

### DIFF
--- a/csrc/adam/cpu_adam.cpp
+++ b/csrc/adam/cpu_adam.cpp
@@ -32,7 +32,7 @@ void Adam_Optimizer::Step(float* _params,
     float bias_correction2 = 1 / sqrt(1 - _betta2_t);
 
     float step_size = -1 * _alpha / bias_correction1;
-
+    float w_decay = -1 * _alpha * _weight_decay;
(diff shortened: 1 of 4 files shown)


samyam (approved):

jeffra (approved):

jeffra:
Do we expect users will want to switch between adamw modes? If so, we should somehow expose the flag in our config json.

RezaYazdaniAminabadi:
That's a good point. Right now, there is the adamw_mode defined as the parameter that can be specified when instantiating the DeepSpeedCPUAdam (https://github.com/microsoft/DeepSpeed/blob/reyazda/add-cpu-adamW/deepspeed/ops/adam/cpu_adam.py#L20). However, it that can be passed in through the deepspeed config. And this can be used for all optimizers.
Thanks Jeff

PR #484, opened 2020-10-23
Add CPUAdam optimizer for zero-offload in deepspeed engine

RezaYazdaniAminabadi:


jeffra (approved):

deepspeed/runtime/engine.py
```diff
@@ -548,12 +548,22 @@ def _configure_basic_optimizer(self, model_parameters):
...
-                optimizer = torch.optim.Adam(model_parameters, **optimizer_parameters)
+        if self.zero_cpu_offload():
+            if self.optimizer_name() == DEEPSPEED_ADAMW:
+                from deepspeed.ops.adam import DeepSpeedCPUAdam
+                optimizer = DeepSpeedCPUAdam(model_parameters, **optimizer_parameters)
+            elif self.optimizer_name() == DEEPSPEED_ADAM:
+                from deepspeed.ops.adam import DeepSpeedCPUAdam
+                optimizer = DeepSpeedCPUAdam(model_parameters,
+                                             **optimizer_parameters,
+                                             adamw_mode=False)
```
jeffra, line 559: I think this should be set to adamw = true based on our recent conversation.

deepspeed/ops/adam/cpu_adam.py
```diff
@@ -16,7 +16,8 @@ def __init__(self,
                         0.999),
                  eps=1e-8,
                  weight_decay=0,
-                 amsgrad=False):
+                 amsgrad=False,
+                 adamw_mode=True):
```
jeffra, line 20: Can we add a docstring to this class describing when and how it's used? I believe we only want to use it combined with zero-offload but we are getting users who are trying to use it solo which isn't supported.

PR #501, opened 2020-11-04
Two CPU <--> GPU data copy performance issue fix suggestions

davidmin7:
This PR is consist of two parts:

1. DeepSpeed has 'overlap_comm' option to overlap data transfer and computation during the backward step, but this does not actually work as intended as there is an hidden cuda synchronization. To be more specific, in the `_has_inf_or_nan()` function in `stage2.py` file, the operations related to the variable `cpu_sum` require a 4-byte DtoH data transfer and this is a blocking call. To solve this, I created a pinned memory storage at the initialization step and use this storage as an asynchronous cudaMemcpy target.

2. During the forward pass, partial activations are copied to the host memory through blocking cudaMemcpy calls. Potentially this can be improved with pinned memory allocations, but anyway currently not implemented in such way. Assuming there is a good enough amount of reason to not use the pinned memory allocations for this step, we can at least try to populate the pages of the target tensors ahead of the actual cudaMemcpy call. I have added a simple code which 'touches' the unpopulated pages of the tensor. This approach makes the page population to happen asynchronously to the forward computation time. Without the pre-population step like this, the pages need to be populated during the blocking cudaMemcpy calls, which are already in the critical path of the forward step. The modification is added in `checkpointing.py` file.

In a simple GPT-2 training benchmark, I see the following improvements,
Unmodified: ~20 seconds per iteration
Unmodified + overlap_comm: ~20 seconds per iteration
`_async_has_inf_or_nan()` change + overlap_comm: ~18.5 seconds per iteration
`_async_has_inf_or_nan()` change + overlap_comm + pre-page-population: ~16 seconds per iteration

I think the performance improvement by the second modification can vary significantly depending on the system used to test.


deepspeed/runtime/activation_checkpointing/checkpointing.py
```diff
@@ -402,6 +402,11 @@ def forward(ctx, run_function, *args):
                         contiguous_data_buffers[i] = tensor_list
                         data_offsets[i] = 0
 
+                    contiguous_data_buffers[i][data_offsets[i]].data[range(
```
samyam, line 405: Why do we need this for the overlap_comm ? Could you explain what is hppening here?

deepspeed/runtime/zero/stage2.py
```diff
@@ -1575,6 +1583,27 @@ def _has_inf_or_nan(x, j=None):
                 return True
             return False
 
+    @staticmethod
+    def _async_has_inf_or_nan(x, async_buff, j=None):
+        try:
+            # if x is half, the .float() incurs an additional deep copy, but it's necessary if
+            # Pytorch's .sum() creates a one-element tensor of the same type as x
+            # (which is true for some recent version of pytorch).
+            cpu_sum = float(async_buff.copy_(x.float().sum(), non_blocking=True))
```
samyam, line 1595: This is a really good catch. One question I have here is, can we guarantee that the value of cpu_sum is the value of async_buff after the copy_ and not before? Since copy_ is non-blocking, wouldn't this cause a race condition, where float(async_buff) might get executed before copy_ finishes, in which case it will just get the value that was already in async_buff?

tjruwase: I have a general question here since I have recently encoutered `non_blocking=True` a few times. What is the prescribed method for synchronizing this copy operation?

shhssdm: Hi @samyam, thanks for catching the bug. Yeah, I thought there is an implicit stream synchronization in this case but realized Pytorch doesn't enforce anything. So as you said, it is getting a stale data in this case. I can enforce a stream synchronization by using `torch.cuda.Stream.synchronize(<stream>)` and this `<stream>` should be passed from all above the call stack where it is defined from. Of course, in this case the situation pretty much becomes similar to the case where there was no communication overlap at all (only slightly better than the baseline). If this `local_overflow` does not need to be updated immediately (sorry, I don't know the exact flow of the entire code yet), I think it's better to keep the reduce result as a `gpu_sum` and read it really when it's needed.

deepspeed/runtime/activation_checkpointing/checkpointing.py
shhssdm: Actually this is irrelevant for the overlap_comm. This is for the second optimization mentioned in the post basically 'touching' the pages before the cudaMemcpy.

deepspeed/runtime/activation_checkpointing/checkpointing.py
```diff
@@ -402,6 +402,11 @@ def forward(ctx, run_function, *args):
                         contiguous_data_buffers[i] = tensor_list
                         data_offsets[i] = 0
 
+                    contiguous_data_buffers[i][data_offsets[i]].data[range(
+                        0,
+                        contiguous_data_buffers[i][data_offsets[i]].data.shape[0],
+                        int(4096 / contiguous_data_buffers[i][
```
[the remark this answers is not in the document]
shhssdm: If DeepSpeed is willing to have some constant defining the page size (4KB), that can be done.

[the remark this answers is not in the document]
shhssdm: Actually there was a easy way by using `mmap` module and using the `mmap.PAGESIZE`. I have updated the code.

deepspeed/runtime/zero/stage2.py
shhssdm: Or a simpler way would be to pull out `update_overflow_tracker_for_param_grad` from the `copy_grads_in_partition` function and to create another for loop between line# 970 & 971 to simple pre-run `update_overflow_tracker_for_param_grad`. In this case `update_overflow_tracker_for_param_grad` is not located in between the copy operations and it will not hinder the asynchronous behavior. If someone can verify if the `update_overflow_tracker_for_param_grad` call is dependent to the previous loop iteration it would be nice.

samyam: @tjruwase can synchronize  either with cuda.synchronize() or doing a wait_stream on the cuda stream. @shhssdm We cannot remove the update_overflow_tracker_from_param_grad from where it is because that is the only place where the grads we are using in the function are ready to be consumed, and the next statement removes those grads. But, can do something like self.gpu_sum = self.gpu_sum + x.float().sum(). Then we can use self.gpu_sum to calculate self.local_overflow in the has_overflow directly. This should be an equivalent way of figuring out self.local_overflow, and it would remove the synchronous copy_ in the middle of the communication overlap.

deepspeed/runtime/activation_checkpointing/checkpointing.py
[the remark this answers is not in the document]
tjruwase: Thanks!

deepspeed/runtime/activation_checkpointing/checkpointing.py
samyam: Got it. This is a clever way to page in the memory. I think its fine to pin this memory directly though which might be cleaner?

shhssdm: I think pinning the memory makes more sense, but I wasn't sure if there is any reason not to pin in the first place. And to take the full advantage of the pinned memory and some asynchronous behavior, there should be some more structural changes.

deepspeed/runtime/zero/stage2.py
shhssdm: @samyam Makes sense. I changed the code to cumulatively check the overflow instead if overlap_comm is enabled.

deepspeed/runtime/activation_checkpointing/checkpointing.py
samyam: Sure. The current solution is fine too. Could you please just add a comment there though on why that line exists?

deepspeed/runtime/zero/stage2.py
```diff
@@ -775,8 +778,11 @@ def get_grad_position(self, group_id, tensor_list, first_offset, partition_size)
             current_offset += num_elements
 
     def update_overflow_tracker_for_param_grad(self, param):
-        if param.grad is not None and self._has_inf_or_nan(param.grad.data):
-            self.local_overflow = True
+        if param.grad is not None:
+            if self.overlap_comm:
+                self.gpu_sum = self.gpu_sum + param.grad.data.float().sum()
+            elif self._has_inf_or_nan(param.grad.data):
+                self.overflow = True
```
samyam, line 785: This should be self.local_overflow not self.overflow?

deepspeed/runtime/zero/stage2.py
```diff
@@ -1532,6 +1538,9 @@ def has_overflow_partitioned_grads_serial(self):
 
     def has_overflow(self, partition_gradients=True):
         if partition_gradients:
+            if self.overlap_comm:
```
samyam, line 1541: if self.overlap_comm and self.cpu_offload

ShadenSmith:
Thanks a ton for putting this together, @shhssdm! Our testing infrastructure needs a bit of work in order to test this confidently. I've been working on this for a bit already, but need a bit of time to prepare some tests before merging this.

shhssdm:
> Thanks a ton for putting this together, @shhssdm! Our testing infrastructure needs a bit of work in order to test this confidently. I've been working on this for a bit already, but need a bit of time to prepare some tests before merging this.

shhssdm:
> Thanks a ton for putting this together, @shhssdm! Our testing infrastructure needs a bit of work in order to test this confidently. I've been working on this for a bit already, but need a bit of time to prepare some tests before merging this.

Sorry while I was trying to reply you I mis-clicked and closed the request but reopened it now. Yeah I really have no problem with it taking longer time for this merge request. I initially thought it would be a very simple request but as @samyam pointed out it required some more structural changes.

deepspeed/runtime/zero/stage2.py
shhssdm: It should be fixed now.

deepspeed/runtime/zero/stage2.py
shhssdm: Right, I fixed this.

deepspeed/runtime/activation_checkpointing/checkpointing.py
shhssdm: I added some comments for this step.

PR #524, opened 2020-11-12
ds_report bug fix on cpu and guard torch import in setup.py

jeffra:


ShadenSmith (approved):

deepspeed/env_report.py
```diff
@@ -62,6 +62,8 @@ def ninja_installed():
 def nvcc_version():
     import torch.utils.cpp_extension
     cuda_home = torch.utils.cpp_extension.CUDA_HOME
+    if cuda_home is None:
+        return f"{RED} [FAIL] cannot find CUDA_HOME via torch.utils.cpp_extension.CUDA_HOME={torch.utils.cpp_extension.CUDA_HOME} {END}"
```
tjruwase, line 66: Do we need to guard against torch import error?

tjruwase (approved):

deepspeed/env_report.py
jeffra: I think this is okay, the torch guard in setup.py is more related to the `pip install deepspeed` assumes now that torch is installed already. But we won't even install deepspeed at all if torch isn't installed.

PR #534, opened 2020-11-18
Fix potential random layout inconsistency issues in sparse attention modules

Justin1904:
There are two changes made to the SparseSelfAttention module in this PR:

1) Now SparseSelfAttention module will create a "master_layout" upfront and register it as a buffer, this saves us the need to create new layout on-the-fly later (which can cause inconsistency if there's randomness in layout creation) and also makes it easy for us to save & load the layout from checkpoint;
2) Add a broadcast of layout at the beginning to ensure different processes in distributed training will have consistent layout.

@arashashari


deepspeed/ops/sparse_attention/sparse_self_attention.py
```diff
@@ -22,7 +23,8 @@ def __init__(
         # SparsityConfig parameters needs to be set accordingly
         sparsity_config=SparsityConfig(num_heads=4),
         key_padding_mask_mode='add',
-        attn_mask_mode='mul'):
+        attn_mask_mode='mul',
+        max_seq_length=2048):
```
arashashari, line 27: Could you please add docstring for the new parameter as well?

arashashari (approved):

deepspeed/ops/sparse_attention/sparse_self_attention.py
Justin1904: Sure, just added.

deepspeed/ops/sparse_attention/sparse_self_attention.py
```diff
@@ -22,29 +23,50 @@ def __init__(
...
         # mask modes
         self.key_padding_mask_mode = key_padding_mask_mode
         self.attn_mask_mode = attn_mask_mode
 
     ops = dict()
 
+    def get_layout(self, L):
+        # if layout is never synchronized across GPUs, broadcast the layout from global rank 0
+        if self._need_layout_synchronization and dist.is_initialized():
+            dist.broadcast(self.master_layout, src=0)
```
jeffra, line 54: This might break with model parallelism (e.g., megatron-style or pipeline parallelism). However, it might be tricky to get the correct process group and rank inside the op since we can't easily communicate with the deepspeed engine here. /cc @ShadenSmith, @samyam

ShadenSmith: That's a good point @jeffra . I think we want to only broadcast along the data parallel group, similar to our [weight initialization](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L503)? But getting the group is tricky as you pointed out. We could add a `data_parallel_group=None` parameter to the constructor, and if present broadcast along that `torch.distributed` group? It'll be up to the modeling side of things to ensure that the data parallel group is created/provided. Alternatively, I think we'd need a reference to the training engine.

jeffra: Yep that makes sense. The training engine hasn't been created yet in this timeline, so that's a bit tricky. However, for now let's just the `data_parallel_group` passed into the constructor and use if if it's not None in this broadcast. Then it allows the option for this at least.

Justin1904: If we have different data_parallel_groups (i.e. in model parallel scenario), does that mean we would also require passing in the source rank to broadcast from within that process group? Do you think we would also need an optional argument for broadcast_src_rank in the constructor?

jeffra: Yep, we could also add the broadcast_src_rank parameter. This just means the caller has to do this translation instead of us, which sounds fine.

Issue #537, opened 2020-11-18
Cannot install DeepSpeed without CUDA


harrydrippin:
## Problem

I tried to install DeepSpeed above torch-preinstalled `virtualenv`, and saw this error:

```text
Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/04/wv3vnq8n3ns_3h3b5f8rn23r0000gn/T/pip-install-z1zw2ddb/deepspeed/setup.py", line 147, in <module>
        cuda_version = ".".join(torch.version.cuda.split('.')[:2])
AttributeError: 'NoneType' object has no attribute 'split'
[WARNING] Torch did not find cuda available, if cross-compling or running with cpu only you can ignore this message. Adding compute capability for Pascal, Volta, and Turing (compute capabilities 6.0, 6.1, 6.2)
```

## Why?

I found that warning message at [Line 60 of setup.py](https://github.com/microsoft/DeepSpeed/blob/ca9ab1201ff56815652b7b85496d262c2d89d6c1/setup.py#L60):

```python
if not torch.cuda.is_available():
    # Fix to allow docker buils, similar to https://github.com/NVIDIA/apex/issues/486
    print(
        "[WARNING] Torch did not find cuda available, if cross-compling or running with cpu only "
        "you can ignore this message. Adding compute capability for Pascal, Volta, and Turing "
        "(compute capabilities 6.0, 6.1, 6.2)")
```

So we can confirm that it was intentional to allow installation above no-CUDA environment to support cross-compiling or run framework with CPU only (e.g. CI environment). But at [Line 147 of setup.py](https://github.com/microsoft/DeepSpeed/blob/ca9ab1201ff56815652b7b85496d262c2d89d6c1/setup.py#L147), you can see the code trying to obtain CUDA version from torch without logical branch:

```python
torch_version = ".".join([TORCH_MAJOR, TORCH_MINOR])
cuda_version = ".".join(torch.version.cuda.split('.')[:2])
torch_info = {"version": torch_version, "cuda_version": cuda_version}
```

In this case, `torch.version.cuda` will produce `None` and the installation fails.

## Possible Fix

I think we have to add a guard for that line to prevent installation failure. I can send a PR for this if it is a bug.

Thank you for your support!

jeffra:
[...]

PR #538, opened 2020-11-18
Fix setup.py for cpu-only environment installation

harrydrippin:
This PR contains following simple suggestions:

* Add a guard to not using `torch.version.cuda` above no-CUDA environment.
* Fix several typos on setup.py.

This PR is related with #537.


jeffra (approved):

jeffra:
Thanks for your contribution! This looks good, will merge after tests pass :)

PR #542, opened 2020-11-19
Discover variables for NCCL backend on AML without mpi4py

awan-10:
This PR will allow us to discover that we are running on AzureML and discover the required flags to initialize the NCCL backend in PyTorch without relying on mpi4py.


deepspeed/runtime/engine.py
```diff
@@ -199,6 +202,45 @@ def __init__(self,
...
+        # read and environment variable to detect if we are using an Azure ML environment
+        if 'AZUREML_EXPERIMENT_ID' in os.environ:
+            return True
+        else:
+            return False
+
+    def _set_environment_variables_for_nccl_backend(self,
+                                                    args,
+                                                    master_port=6105,
+                                                    verbose=True):
```
ShadenSmith, line 215: It might be good to have a docstring here outlining what assumptions we make about the environment and the modifications that we make.

jeffra: We're essentially taking this code from AML (https://azure.github.io/azureml-web/docs/cheatsheet/distributed-training#environment-variables-from-openmpi) to do this. Ideally in the future AML will provide this logic for us via the AML SDK so we don't have to maintain this piece when it inevitably changes. In the docstring maybe let's link to this cheatsheet as a reference?

jeffra (approved):
Looks good, thanks Ammar! After addressing Shaden's comment I think we're good to merge (after tests pass).

deepspeed/runtime/engine.py
awan-10: Thanks for the feedback, Shaden and Jeff. Adding the link to the cheat sheet sounds good. I will add it as a docstring.

awan-10: I have added the docstring.

ShadenSmith: Awesome!

ShadenSmith (approved):

PR #553, opened 2020-11-24
Simplify dist init and only init if needed.

awan-10:
Fix the issue reported by Jiacheng/Ivo on AML/ITP.


deepspeed/runtime/engine.py
```diff
@@ -125,21 +125,21 @@ def __init__(self,
...
             dist_init_required = not dist.is_initialized()
 
-        if self._in_aml():
-            self._set_environment_variables_for_nccl_backend(args)
-        else:
-            self._mpi_check(args, dist_init_required)
-
         self.dist_backend = "nccl"
+
         if dist_init_required:
```
jeffra, line 130: If `dist_init_required=False` we should ensure that `dist.is_initialized()==True` maybe via an assert?

deepspeed/runtime/engine.py
```diff
@@ -125,21 +125,21 @@ def __init__(self,
...
-                dist.init_process_group(backend=self.dist_backend)
+            # discover using mpi4py if user specifies the flag
+            if hasattr(args, 'deepspeed_mpi') and args.deepspeed_mpi:
+                self._mpi_check(args, dist_init_required)
             else:
-                logger.warning(
-                    "Was given dist_init_required=True but detected that torch"
-                    "distributed was already initialized, cannot initialize twice.")
+                # detect if we are in Azure ML environment
+                if self._in_aml():
```
jeffra, line 136: I wonder if we could just do something like the following (see below). This would work I think as long as the ITP/AML environment made sure to init torch distributed before running deepspeed.initialize() which as far as I understand they do?

```python
        if dist_init_required is None:
            dist_init_required = not dist.is_initialized()

        if dist_init_required:
            if self._in_aml():
                self._set_environment_variables_for_nccl_backend(args)
            else:
                self._mpi_check(args, dist_init_required)

            self.dist_backend = "nccl"
            logger.info("Initializing torch distributed with backend: {}".format(self.dist_backend))
            dist.init_process_group(backend=self.dist_backend)
        else:
            assert dist.is_initialized(), 'dist_init_required=False but torch.distributed is not initialized, please init manually or set dist_init_required=True'
```

If for some reason they don't init the backend but they do patch the environment vars we could also add some code that checks to see if all the required environment variables are already patched to get dist.init_process_group(..) to work. If so, we init the backend. (code snippet below comes from #276)

```python
    required_env = ["RANK", "WORLD_SIZE", "MASTER_ADDR", "MASTER_PORT", "LOCAL_RANK"]
    if all(map(lambda v: v in os.environ, required_env)):
        print('safe to init torch distributed')
```

awan-10: But what if there is another user of AML/ITP and they use deepspeed without intializing ddp, then it will throw the same error as Jiacheng saw. I think it is safer to check for the MPI flag first so we don't take the AML path if the user is specifying it already.

jeffra: Maybe let’s clarify Jiacheng’s error. I think they were already initializing ddp but our code currently does the mpi and aml check code even if the backend is initialized which I think is a bug on our side.

deepspeed/runtime/engine.py
awan-10: Got it. How about doing the following as there are only three cases, 1) None, 2) False, 3) True, rigth?

```python
        if dist_init_required is None:
            dist_init_required = not dist.is_initialized()

        if dist_init_required is False:
            assert (dist.is_initialized()==True), "Torch distributed not initialized. Please set dist_init_required to True or initialize before calling deepspeed.initialize()"

        if dist_init_required:

```

deepspeed/runtime/engine.py
awan-10: Agreed. Jiacheng's error will be fixed because we moved the init part inside the if dist_init_required so we are good on that side.

deepspeed/runtime/engine.py
```diff
@@ -121,25 +121,31 @@ def __init__(self,
...
+        if dist_init_required is False:
+            assert (dist.is_initialized()==True), "Torch distributed not initialized. Please set dist_init_required to True or initialize before calling deepspeed.initialize()"
+
+        # DeepSpeed will initialize torch distributed only if the user has not already intialized it.
+        if dist_init_required and not dist.is_initialized():
+            # discover using mpi4py if user specifies the flag
+            if hasattr(args, 'deepspeed_mpi') and args.deepspeed_mpi:
+                # if in Azure ML environment and user specified this flag, notify the user to remove the flag.
+                if self._in_aml():
+                    logger.info(
```
jeffra, line 138: Can we change this to a warning? since they probably don't want to do this?

jeffra (approved):

deepspeed/runtime/engine.py
awan-10: Sounds good. I have pushed the change to make it logger.warning instead of info.

PR #559, opened 2020-11-25
supporting different hidden dimensions

RezaYazdaniAminabadi:
This PR removes the previous restriction on hidden dimension for the DeepSpeed transformer layer, so that we can train different models with any hidden dimension, as small as 32 and as large as 64K. To enable this feature, we modify some of the CUDA kernels inside a transformer layer, such as LayerNorm, Dropout, and Transform.


jeffra (commented):
Can we add a description to the PR and a short motivation for the update? I imagine this is motivated by a 1P request but can we describe it for 3p users?

csrc/transformer/dropout_kernels.cu
```diff
@@ -670,12 +672,12 @@ __global__ void dropout_kernel(const int dim,
...
 
-        out_cast[idx] = result_f;
-
-        mask[i] = m[0];
-        mask[i + 1] = m[1];
-        mask[i + 2] = m[2];
-        mask[i + 3] = m[3];
+        out_cast[j] = result_f;
+        mask_32[j] = m_32;
+        // mask[i] = m[0];
```
tjruwase, line 677: Please remove this dead code.

csrc/transformer/dropout_kernels.cu
```diff
@@ -690,14 +692,15 @@ void launch_dropout(T* out,
                     float ratio,
                     cudaStream_t stream)
 {
-    dim3 grid_dim(batch);     // DS_GET_BLOCKS(total_count/4);
-    dim3 block_dim(dim / 4);  // DS_CUDA_NUM_THREADS;
+    int total_count = batch * dim / 4;
```
tjruwase, line 695: Can you use a constant variable explaining what 4 is?

csrc/transformer/transform_kernels.cu
```diff
@@ -123,10 +130,13 @@ void launch_transform_0213<float>(float* output,
                                   int heads,
                                   cudaStream_t stream)
 {
-    dim3 block_dim(hidden_dim / heads / 4, heads);
-    dim3 grid_dim(batch_size, seq_length);
-    transform_0213<float>
-        <<<grid_dim, block_dim, 0, stream>>>(output, vals, hidden_dim, seq_length, heads);
+    hidden_dim >>= 2;
+    int head_ext = (hidden_dim - 1) / 1024;
```
tjruwase, line 134: Can you replace constant 1024 with a variable?

tjruwase (approved):

csrc/transformer/transform_kernels.cu
RezaYazdaniAminabadi: Of course, this is the maximum number of threads allowed at each block.

csrc/transformer/dropout_kernels.cu
RezaYazdaniAminabadi: sure, 4 is the loop-unrolling degree.

jeffra (changes requested):
Before merge we need confirmation of extensive convergence testing with the new changes. @RezaYazdaniAminabadi you mentioned you were in the process of finishing some runs right?

RezaYazdaniAminabadi:
That's a great point, Jeff. There are a few people working with this branch, such as @tjruwase and @arashashari. I will also do some more tests with this to make sure everything is fine.

csrc/transformer/dropout_kernels.cu
```diff
@@ -1,5 +1,7 @@
 #include "custom_cuda_layers.h"
 
+const int unroll_factor = 4;
```
samyam, line 3: assert that unroll_factor is 4 since the code is hardcoded with 4 way unrolling.

csrc/transformer/dropout_kernels.cu
```diff
@@ -13,17 +15,17 @@ __global__ void dropout_kernel(const int N,
     curandStatePhilox4_32_10_t state;
     curand_init(seed.first, idx, seed.second, &state);
 
-    CUDA_1D_KERNEL_LOOP(j, N / 4)
+    CUDA_1D_KERNEL_LOOP(j, N / unroll_factor)
```
samyam, line 18: What if N is not divisible by the unroll factor? Should we take care of the boundary condition?

csrc/transformer/dropout_kernels.cu
RezaYazdaniAminabadi: great point, I will do that

csrc/transformer/dropout_kernels.cu
RezaYazdaniAminabadi: yes, I can do that, however, I need to access the last part that is indivisible one by one. I will add this part to the code too.

csrc/includes/custom_cuda_layers.h
```diff
@@ -22,6 +22,8 @@
 #define MAX_THREAD_ITERATIONS 8  // Maximum 8K
 #define MAX_WARP_NUM 32
 
+#define MAX_REGISTERS 64
```
eltonzheng, line 25: just curious why set the max registers as 64, is it for some special kernels or a common setting? and it looks too small as a max count of register.

csrc/transformer/normalize_kernels.cu
```diff
@@ -92,15 +93,21 @@ __global__ void fused_bias_residual_layer_norm(float* vals,
     if (training)
         if (g.thread_rank() == 0) vars[row] = variance;
 
+    iterations = row_stride / iteration_stride;
```
eltonzheng, line 96: is the re-computation needed?

csrc/transformer/normalize_kernels.cu
```diff
@@ -29,26 +19,36 @@ __global__ void fused_bias_residual_layer_norm(float* vals,
...
+    vals += (row * row_stride);
 
     float sum = 0.f;
 #pragma unroll
     for (int i = 0; i < iterations; i++) {
-        vals_arr[i] = residual[row * row_stride + i * iteration_stride + id];
+        vals_arr[i] = residual[i * iteration_stride + id];
         sum += vals_arr[i];
     }
+    if ((iterations * iteration_stride + id) < row_stride) {
```
eltonzheng, line 47: it may be better to assign "iterations * iteration_stride + id" to a variable, and reuse it instead of repeat this expression.

csrc/transformer/normalize_kernels.cu
```diff
@@ -363,26 +377,35 @@ __global__ void fused_bias_residual_layer_norm(float* vals,
     if (training)
         if (g.thread_rank() == 0) vars[row] = variance;
 
+    iterations = row_stride / iteration_stride;
```
eltonzheng, line 380: same comments as previous function.

csrc/includes/custom_cuda_layers.h
RezaYazdaniAminabadi: Interesting question. I think I can push it up, however it is also multiplied by 4, since there is always the loop-unrolling factor of 4. So, it is actually 256. Also, I am worried that if more blocks are assigned to an SM, there won't be enough registers to be allocated and there will be more memory movement.

csrc/transformer/normalize_kernels.cu
RezaYazdaniAminabadi: yes, because I changed it in the middle! just wanted to use less registers! :-)

csrc/transformer/normalize_kernels.cu
RezaYazdaniAminabadi: I agree. I will do that. Thanks Elton

eltonzheng (approved):

jeffra (approved):

PR #563, opened 2020-11-28
[doc] typo fix and clarification

stas00:
This PR:
* fixes a misspelled method name
* also `( () )` doesn't read too well, until one reads the code and understands that it's not a formatting bug. I propose to simply say that it's a callable object.


tjruwase (approved):

tjruwase:
@stas00  Thanks for these fixes!

PR #564, opened 2020-11-28
tracking optimizer step in cpu-adam when loading checkpoint

RezaYazdaniAminabadi:


tjruwase (approved):

csrc/includes/cpu_adam.h
```diff
@@ -91,11 +91,8 @@ class Adam_Optimizer {
                 __half* dev_params = nullptr);
     inline void IncrementStep(size_t step)
     {
-        if (_step < step) {
+        while (_step < step) {
             _step++;
-            if (_step != step) {
```
jeffra, line 96: Could there be non-checkpoint loading scenarios where we would still want to raise this error? Totally makes sense in checkpoint loading we need to align the step count. Just concerned that we’re silently aligning them in all cases now.

RezaYazdaniAminabadi: That's a great point, but since this is coming from Python (https://github.com/microsoft/DeepSpeed/blob/cpu-adam/track-optimizer-step/deepspeed/ops/adam/cpu_adam.py#L119) and is controlled by user, I think we are safe to change this! However, I think this makes sense to print a warning so that the user knows of this update. What do you think?

RezaYazdaniAminabadi: By the way, the current step cannot stay behind the step coming in the cpu-adam step function! I think we should throw a runtime error for that.

csrc/includes/cpu_adam.h
```diff
@@ -95,20 +95,44 @@ class Adam_Optimizer {
...
         _betta2_t *= _betta2;
         if (_step < step) {
-            printf("Warning: updating optimizer's state to match the step count at %d\n", step);
-            while (_step < step) {
-                _step++;
-                _betta1_t *= _betta1;
-                _betta2_t *= _betta2;
-            }
+            printf("Warning: updating optimizer's state to match the step count at %lu\n", step);
+            _betta1_t = std::pow(_betta1, step);
```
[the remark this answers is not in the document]
RezaYazdaniAminabadi: I think in that case, the _step has become zero (https://github.com/microsoft/DeepSpeed/blob/059f6aab5bfe2abb72106af29ced1b8575ff4fd9/csrc/includes/cpu_adam.h#L62), since this means a new training process. Is in it right?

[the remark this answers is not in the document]
RezaYazdaniAminabadi: When loading a new checkpoint, the optimizer will initialize, am I right?

tjruwase (approved):

PR #567, opened 2020-12-02
[cifar tutorial] improve readability

stas00:
This PR:
* fixes a typo `parametersnd`
* fixes `__` to `_`
* removes incorrect param in a description (`optimizer` isn't passed to `initialize` in the code snippet)
* rewrites text to be more readable (some of it wasn't quite parseable)

Thanks.

### DIFF
--- a/docs/_tutorials/cifar-10.md
+++ b/docs/_tutorials/cifar-10.md
@@ -132,7 +132,7 @@ Here we we initialize DeepSpeed with CIFAR-10 model (`net`), `args`, `parameters
  # 1) Distributed model
  # 2) Distributed data loader
  # 3) DeepSpeed optimizer
- model_engine, optimizer, trainloader, __ = deepspeed.initialize(args=args, model=net, model_parameters=parameters, training_data=trainset)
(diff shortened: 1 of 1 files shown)


docs/_tutorials/cifar-10.md
```diff
@@ -122,7 +122,7 @@ def initialize(args,
                collate_fn=None):
 ```
 
-For CIFAR-10 model, we initialize DeepSpeed its model (net) is created as below, to pass the raw `model`, `optimizer`, `args`, `parametersnd` and `trainset`.
+Here we we initialize DeepSpeed with CIFAR-10 model (`net`), `args`, `parameters` and `trainset`:
```
jeffra, line 125: Loving the edits, thanks for the help here. I think you accidently added an extra 'we' here.

jeffra (approved):

docs/_tutorials/cifar-10.md
stas00: oops, thank you for catching that - fixed.

Issue #95, opened 2020-02-20
Conda Environment Install Issue


kleingeo:
Trying to get DeepSpeed installed for local use with a Conda environment, but it seems that DeepSpeed in not installing to the environment itself. After building the wheel DeepSpeed is not installing into the proper Conda conda environment location. Apex is installing in the proper environment location. Unclear why DeepSpeed is not working but Apex is.

ShadenSmith:
Hi there! We're in the process of rewriting our installation scripts (that were previously only used within Docker containers) and hoping to also release a conda package in short time. These sorts of issues should be fixed at that point.

Michiel29:
The installation scripts perform sudo -H pip installs, which install system wide. I replaced those with normal pip installs and it installed into the current environment without problems.

kleingeo:
When I installed it I did that in the install script (for both the deep speed, apex and requirements). However, there were still issues in that DeepSpeed would not install to the right environment location. Looking at the installation a little more, this seemed more likely an issue with the wheel created for DeepSpeed in the install.sh file. I was able to get it working by forcing pip to install DeepSpeed into the correct location (the same location that Apex was correctly installed to).

ShadenSmith:
We have a now have a conda package uploaded and we appreciate any feedback!

We have versions compiled for `cudatoolkit` versions 10.0 and 10.1 To install along with pytorch and other dependencies that are in the conda-forge channel:

```bash
conda install deepspeed cudatoolkit=10.1 -c deepspeed -c pytorch -c conda-forge
```

ShadenSmith:
The repo's `install.sh` should respect the environment by default now (`sudo` is opt-in). Please let me know if the issue persists.

kleingeo:
Using the conda install, deepspeed shows up when I run `conda list` but it is not available when trying to import in python.

ShadenSmith:
Hi @kleingeo, thanks for the report. I can see that on my end now as well. Not sure what happened...I'm looking into it.

Interestingly, the `deepspeed` entry point looks fine and is found in my `$PATH` after installation. And I can see the DeepSpeed library installed under `~/miniconda3/envs/test/lib/python3.7/site-packages/deepspeed/` (where `test` is my conda environment name), and also see the expected `~/miniconda3/lib/python3.7/site-packages` in my `sys.path`...so I'm not sure why the deepspeed library is not importable.

kleingeo:
[...]

PR #578, opened 2020-12-05
[build] make builder smarter and configurable wrt compute capabilities + docs

stas00:
This PR fixes:

```
RuntimeError: cuda runtime error (209) : no kernel image is available for execution on the device at ...
```
error, resulting from this build:
```
DS_BUILD_OPS=1 pip install --no-clean --no-cache -v --disable-pip-version-check -e . 
```

`multi_tensor_adam.cu` and `fused_lamb_cuda_kernel.cu` were getting only `-gencode=arch=compute_80,code=sm_80` flags and missing all the rest `-gencode arch=compute_60,code=compute_60 -gencode arch=compute_61,code=compute_61 -gencode arch=compute_70,code=compute_70 -gencode arch=compute_80,code=compute_80 -gencode arch=compute_86,code=compute_86`

The lone `-gencode=arch=compute_80,code=sm_80` comes from `CUDAExtension` - which currently checks the capacity of the 1st card only and assumes that the other cards and the same. Moreover it clamps down the number to the minimum of the same first digit in:
```
python -c "import torch; print(torch.cuda.get_arch_list())" 
['sm_37', 'sm_50', 'sm_60', 'sm_70', 'sm_75', 'sm_80']
```
so, for example, `sm_86` becomes `sm_80`.

I'm pretty sure it's wrong though since it's the card with `compute_61` that was getting this error, so these 2 libs weren't built to support `compute_61`. This is with pytorch-nightly.

The 2 cards I have right now are:
```
$ CUDA_VISIBLE_DEVICES=0 python -c "import torch; print(torch.cuda.get_device_capability())"
(8, 6)
$ CUDA_VISIBLE_DEVICES=1 python -c "import torch; print(torch.cuda.get_device_capability())"
(6, 1)
```

Note: A PR has been proposed to fix this problem transparently to `CUDAExtension` users, but it won't be available until future versions of pytorch if it's accepted. https://github.com/pytorch/pytorch/pull/48891

The cost of this deepspeed PR is that the build process is now slightly slower as it now has to build 4-5 kernels x 2 instead of just 2 (assuming all features are enabled to be compiled). So perhaps down the road we can fix that by conditioning on pytorch version and build less kernels. Alternatively you could copy the loop to get just the required archs: https://github.com/pytorch/pytorch/blob/b8f90d778d4c0739c7c07fe2b2fb0aef5e7c77e7/torch/utils/cpp_extension.py#L1531-L1545

------------------

**edit**: After understanding how `CUDAExtension`  sorts out its archs I have made further improvements to this PR

* [x] by default build the cuda extension for all supported archs (discussion above)
* [x] changed jit_mode to check archs of all visible cards and not just the first one (similar to the PR I submitted to pytorch)
* [x] add support for `TORCH_CUDA_ARCH_LIST`, in exact same manner as `CUDAExtension` does it. So now I can build deepspeed much faster by specifying only the archs that I need: `TORCH_CUDA_ARCH_LIST="6.1;7.5;8.6" DS_BUILD_OPS=1 pip install --no-clean --no-cache -v --disable-pip-version-check -e .`
   `TORCH_CUDA_ARCH_LIST` overrides the `CUDAOpBuilder`'s `cross_compile_archs` arg
* [x] refactor the code
* [x] document `CUDAOpBuilder` nuances
* [x] extend the advanced installation tutorial to document the arch mismatch error, its cause and how to build the most efficient version of deepspeed for the exact desired archs.

A related PR: proposed support for compute_86 in https://github.com/microsoft/DeepSpeed/pull/577 to include the full capabilities of the rtx-30* cards.

This might be related to https://github.com/microsoft/DeepSpeed/issues/95

### DIFF
--- a/docs/_tutorials/advanced-install.md
+++ b/docs/_tutorials/advanced-install.md
@@ -84,6 +84,29 @@ the nodes listed in your hostfile (either given via --hostfile, or defaults to
 /job/hostfile).
 
 
+## Building for the correct architectures
+
+If you're getting the following error:
+
+```python
+RuntimeError: CUDA error: no kernel image is available for execution on the device
+```
+when running deepspeed that means that the cuda extensions weren't built for the card you're trying to use it for.
+
+When building from source deepspeed will try to support a wide range of architectures, but under jit-mode it'll only support the archs visible at the time of building.
+
+You can build specifically for a desired range of architectures by setting a `TORCH_CUDA_ARCH_LIST` env variable, like so:
+
+```bash
+TORCH_CUDA_ARCH_LIST="6.1;7.5;8.6" pip install ...
+```
+
+It will also make the build faster when you only build for a few architectures.
+
+This is also recommended to do to ensure your exact architecture is used. Due to a variety of technical reasons a distributed pytorch binary isn't built to fully support all architectures, skipping binary compatible ones, at a potential cost of underutilizing your full card's compute capabilities. To see which archs get included during the deepspeed build from source - save the log and grep for `-gencode` arguments.
+
+The full list of nvidia gpus and their compute capabilities can be found [here](https://developer.nvidia.com/cuda-gpus).
+
 ## Feature specific dependencies
 
 Some DeepSpeed features require specific dependencies outside of the general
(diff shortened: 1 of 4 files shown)


jeffra (approved):
Thank you for this PR, this is really a great contribution!

PR #588, opened 2020-12-08
Pipeline warnings and checkpoint portability

ShadenSmith:
This PR does two things:

1. Drops deprecated `allreduce_gradients` argument to `backward()`
2. Fixes a portability bug in pipeline checkpointing. Previously, checkpoints could not be renamed due to this.


jeffra (approved):
Looks good, but can we add a test case that covers the checkpoint bug?

ShadenSmith:
Good point. Our pipeline checkpoint suite already covers this in a way. [This change](https://github.com/microsoft/DeepSpeed/pull/588/files#diff-26611f6be759237464a03bb1328cbc16555888836b3504dc3703e2e25d2a3ca3R1113) that returns `None` instead of a file path would trigger the same bug; we'd just fail to open `None` instead of a non-existent (or incorrect) file. Does that seem like a reasonable test?

PR #595, opened 2020-12-11
implement missing get_last_lr

stas00:
This PR fixes:
```
AttributeError: 'WarmupLR' object has no attribute 'get_last_lr'
```
by adding `get_last_lr` as it is implemented in [pytorch](https://github.com/pytorch/pytorch/blob/159f258415364f73a20ecd2b15b9f2e52ef1b922/torch/optim/lr_scheduler.py#L96)
I think it was added around pytorch-1.4.

I added `get_last_lr` to the rest of lr classes in that file, not sure if there are others.

### DIFF
--- a/deepspeed/runtime/lr_schedules.py
+++ b/deepspeed/runtime/lr_schedules.py
@@ -381,6 +381,11 @@ class LRRangeTest(object):
             lr_range_test_min_lr * lr_increase for lr_range_test_min_lr in self.min_lr
         ]
 
+    def get_last_lr(self):
+        """ Return last computed learning rate by current scheduler.
+        """
+        return self._last_lr
+
     def _update_optimizer(self, group_lrs):
         for param_group, lr in zip(self.optimizer.param_groups, group_lrs):
             param_group['lr'] = lr
@@ -390,6 +395,7 @@ class LRRangeTest(object):
             batch_iteration = self.last_batch_iteration + 1
         self.last_batch_iteration = batch_iteration
         self._update_optimizer(self.get_lr())
+        self._last_lr = [group['lr'] for group in self.optimizer.param_groups]
 
     def state_dict(self):
         return {'last_batch_iteration': self.last_batch_iteration}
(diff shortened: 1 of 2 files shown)


stas00:
It looks like your CI is stuck - not doing the build since 12 hours...

jeffra:
> It looks like your CI is stuck - not doing the build since 12 hours...

Super strange, not sure what the issue was. Our CI nodes were idle but the job didn't submit to our queue for some reason. After updating your branch with master it seems to have forced it to kick off though.

deepspeed/runtime/lr_schedules.py
```diff
@@ -381,6 +381,11 @@ def get_lr(self):
             lr_range_test_min_lr * lr_increase for lr_range_test_min_lr in self.min_lr
         ]
 
+    def get_last_lr(self):
+        """ Return last computed learning rate by current scheduler.
+        """
+        return self._last_lr
```
tjruwase, line 387: Will this not fail with attribute error if the get_last_lr() is called right after object construction. I think self._last_lr needs to initialized in the constructor.

stas00: while it shouldn't happen, technically yes, it'd fail, but to your suggestion - what would you assign to it as a default?

as I said I only mimicked how pytorch implemented it.

Probably the most clean implementation would be to assert inside `get_last_lr` if it's undefined. Should I do that?

tjruwase: It seems to work in pytorch because LRScheduler, the base class, calls [step](https://github.com/pytorch/pytorch/blob/159f258415364f73a20ecd2b15b9f2e52ef1b922/torch/optim/lr_scheduler.py#L77) in the constructor, which sets self._last_lr. I don't think DeepSpeed lr schedulers are well setup to inherit from such a base class. So, I think an assert is fine to at least inform users of the failure reason.

tjruwase (approved):

deepspeed/runtime/lr_schedules.py
tjruwase: This is great. Thanks.

tjruwase (approved):

deepspeed/runtime/lr_schedules.py
```diff
@@ -381,6 +381,12 @@ def get_lr(self):
             lr_range_test_min_lr * lr_increase for lr_range_test_min_lr in self.min_lr
         ]
 
+    def get_last_lr(self):
+        """ Return last computed learning rate by current scheduler.
+        """
+        assert getattr(self, '_last_lr', None), "need to call step() first"
```
stas00, line 387: Actually, I think this is still weak, since if last lr is 0 it will fail. Probably should change it:
```
assert getattr(self, '_last_lr', False), "need to call step() first"
```

stas00: I changed it to the latter.

tjruwase: Agreed. Good catch.

tjruwase (approved):

PR #602, opened 2020-12-14
Elastic training support

jeffra:
Supports scaling up/down training to compatible GPU counts. Adds new 'elasticity' key to our config json. Users indicate their max acceptable batch size and acceptable micro batch sizes. DeepSpeed will find a batch size that is usable with the largest list of compatible GPU counts. The intended consumers of this API and JSON addition are both the user training code and *also* the infrastructure scheduler.

```json
    "elasticity": {
        "enabled": true,
        "max_train_batch_size": 2000,
        "micro_batch_sizes": [2,4,6],
        "min_gpus": 1,
        "max_gpus" : 10000,
        "min_time": 20,
        "version": 0.1
    }
```


g-karthik:
@jeffra I haven't looked at this closely, but am I right to assume this requires the user to also use the `training_data` argument of `deepspeed.initialize()`? Also, how does the infrastructure scheduler tie to this config?

Issue #461, opened 2020-10-04
AssertionError: MPI world size 16 does not match torch world size 8


szhengac:
When I use mpirun for Megatron-LM bert pretraining, I received 
`AssertionError: MPI world size 16 does not match torch world size 8"`

For mpirun args, I use `--allow-run-as-root -np 16 -N 8 --hostfile ${hostfile}`. Is there something I miss for setting torch world size appropriately?

tjruwase:
@szhengac, can you share the number of nodes, gpu per node, and hostfile?

szhengac:
There are 2 nodes with 8 gpus per node. Hostfile is:
 ip1 slots=8
 ip2 slots=8

szhengac:
@tjruwase I wonder if there is any update on this issue?

jeffra:
Hi @szhengac, I have an idea what might be going on however want a bit more info. There should be another print out before this assertion such as "Discovered MPI settings of world_rank=". This comes from our mpi discovery code here: https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L280-L291

Also how many prints like this do you see? Are you seeing 16 of them or only 8?

jeffra:
@szhengac: I see the issue now. It's with our Megatron example, it does not support running with MPI because it's initializing torch distributed itself and not with deepspeed, but this is required since it uses model parallelism. Let me update this and I'll update this thread once it's fixed.

szhengac:
Thanks the quick fix! I wonder if I use deepspeed launcher, is there any easy way to enable numactl such as https://github.com/mlperf/training_results_v0.6/blob/master/NVIDIA/benchmarks/minigo/implementations/tensorflow/minigo/ompi_bind_DGX1.sh? It is convenient to attach this script with mpirun.

szhengac:
@jeffra I just had a try with the pr. But I got one new error when I use mpirun on a single node:

```
Traceback (most recent call last):  
  File "deepspeed_train.py", line 539, in <module>
    main()
  File "deepspeed_train.py", line 529, in main
    model, optimizer = prepare_model_optimizer(args)
  File "deepspeed_train.py", line 442, in prepare_model_optimizer
    model_parameters=optimizer_grouped_parameters)
  File "/home/ec2-user/.local/lib/python3.7/site-packages/deepspeed/__init__.py", line 119, in initialize
    config_params=config_params)
[...]

PR #608, opened 2020-12-17
Ability to initialize distributed backend outside deepspeed runtime

jeffra:
* Refactor mpi and aml patching into deepspeed.utils.distributed
* Support calling `deepspeed.init_distributed()` outside deepspeed.initialize runtime. This is required to support building models that require torch distributed (e.g., model-parallelism). Currently running a model like megatron will break if using an mpi launcher or on aml.
* Should fix issues in #461


tjruwase (approved):

deepspeed/utils/distributed.py
```diff
@@ -0,0 +1,122 @@
...
+    Initialize torch.distributed backend, potentially performing MPI discovery if needed
+    Arguments:
+        dist_backend (str): torch distributed backend, e.g., nccl, mpi, gloo
+        auto_mpi_discovery (bool): if distributed environment variables are not set, attempt to discover them from MPI
+        distributed_port (int, optional): torch distributed backend port
+    """
+
+    required_env = ["RANK", "WORLD_SIZE", "MASTER_ADDR", "MASTER_PORT", "LOCAL_RANK"]
+    if auto_mpi_discovery and not all(map(lambda v: v in os.environ, required_env)):
+        logger.info(
```
ShadenSmith, line 24: It might be nice to guard this log and the one below with a `verbose` flag defaulting to `False`? Or restrict it to one rank printing to reduce verbosity for large jobs.

deepspeed/utils/distributed.py
```diff
@@ -0,0 +1,122 @@
...
+    if not single_node:
+        master_node_params = os.environ["AZ_BATCH_MASTER_NODE"].split(":")
+        os.environ["MASTER_ADDR"] = master_node_params[0]
+        # Do not overwrite master port with that defined in AZ_BATCH_MASTER_NODE
+        if "MASTER_PORT" not in os.environ:
+            os.environ["MASTER_PORT"] = str(master_port)
+    else:
+        os.environ["MASTER_ADDR"] = os.environ["AZ_BATCHAI_MPI_MASTER_NODE"]
+        os.environ["MASTER_PORT"] = "54965"
+    logger.info("NCCL_SOCKET_IFNAME original value = {}".format(
```
ShadenSmith, line 109: Maybe guard this with `verbose` that defaults to `False` ?

tests/unit/common.py
```diff
@@ -33,6 +33,11 @@ def dist_init(local_rank, num_procs, *func_args, **func_kwargs):
             """Initialize torch.distributed and execute the user function. """
             os.environ['MASTER_ADDR'] = '127.0.0.1'
             os.environ['MASTER_PORT'] = '29503'
+            os.environ['LOCAL_RANK'] = str(local_rank)
+            # NOTE: unit tests don't support multi-node so local_rank == global rank
+            os.environ['RANK'] = str(local_rank)
+            os.environ['WORLD_SIZE'] = str(num_procs)
+
             dist.init_process_group(backend=backend,
```
ShadenSmith, line 41: Should we use `deepspeed.init_distributed()` here?

ShadenSmith (approved):

PR #638, opened 2021-01-05
Change dist to torch.distributed to fix a typo/bug in assert statements

awan-10:
The refactor in #608 introduced a small typo so this PR fixes that. @jeffra - please review and merge :)


jeffra (approved):
Great catch, thank you!

PR #640, opened 2021-01-05
Fix checkpointing docstring format

### LINKED ISSUE #636: how to restore a client_state from a saved checkpoint?
According to this [doc](https://deepspeed.readthedocs.io/en/latest/model-checkpointing.html)

I can save a custom client_state:
```
 deepspeed.DeepSpeedEngine.save_checkpoint(self, save_dir, tag=None, client_state={}, save_latest=True)
```
but I don't see how to get it back:
```
deepspeed.DeepSpeedEngine.load_checkpoint(self, load_dir, tag=None, load_module_strict=True, load_optimizer_states=True, load_lr_scheduler_states=True)
```
What am I missing?

Unless it just calls `torch.save` behind the walls and then I just have to `torch.load` it manually. I guess the loader can't do anything with the client state since where would it put it. 

I trust I will see it in action tomorrow and it'll all become clear, I was just trying to understand the docs first before integrating it and was missing the counterpart.

Thank you!

tjruwase:
Fix #636.


jeffra (approved):

stas00:
Much better - thank you!

PR #651, opened 2021-01-09
doc borked table fix

stas00:
At https://www.deepspeed.ai/docs/config-json/#optimizer-parameters there is a borked table:

snapshot_3

It looks like a valid MD, and renders fine with my own tools, but if I look at the source it's different from all other tables in that file by not having a leading new line. So this PR attempts to fix that. I'm not sure how to validate I actually fixed it.


stas00:
Yay, it renders correctly now.

At some point you may want to add to your doc validator a rule that requires at least one new line before a table. MD doesn't require it but your doc formatter does.

PR #659, opened 2021-01-11
[doc] pipeline doc typos/improvements

stas00:
This PR fixes a few typos, rewrites a few sentences to hopefully be easier to understand.

And after reading this doc I have a questions:

1. It's not clear from the tutorial why `TransformerBlockPipe` needs to return `mask`.  And if there is some hidden reason, shouldn't it then return `hidden` too?

2. Why do you need the vars to be a single tensor or a tuple? This surely could be a problematic limitation since some layers could have forward args which aren't tensors. Perhaps a more dynamic approach could be used instead? e.g. with the automatic data remapper I posted here: https://github.com/pytorch/pytorch/issues/49961#issuecomment-753441248

3. wrt Tied Layers - could you please give an example? or at least a pointer to where it can be seen in action?

Please let me know if it'd be better to separate the questions into an Issue and not mix it with the PR. I was just thinking that perhaps the answers could be integrated into the tutorials, hence posted them here.

Thank you!


docs/_tutorials/pipeline.md
```diff
@@ -269,17 +269,17 @@ by DeepSpeed:
...
-scalability issues for massive models. Starting from a `Sequential` allocates
-the model in CPU memory redundantly by every worker. A machine with 16 GPUs
+scalability issues for massive models. In this approach each worker replicates
+the whole model in CPU memory. For example, a machine with 16 GPUs
 must have as much local CPU memory as 16 times the model size.
 
 DeepSpeed provides a `LayerSpec` class that delays the construction of
 modules until the model layers have been partitioned across workers. Then,
-the modules are built on the GPU that owns the layer.
+each GPU allocates only the modules assigned to it.
```
stas00, line 280: Is this supposed to communicate that when `LayerSpec` is used, not only that each GPU allocates only the layers it's assigned to or are you also trying to say that no CPU allocation happens at all?

Or should I have rewritten it to say:

> Then each worker will allocate only the layers it's assigned to. So continuing the example from the previous paragraph, a machine with 16 GPUs will need to allocate a total 1x model size on its CPU.

I think the two paras have GPU and CPU mixed up.

docs/_tutorials/pipeline.md
```diff
@@ -269,17 +269,17 @@ by DeepSpeed:
 * `partition_method="uniform"` balances the number of layers per stage.
 
 ### Memory-Efficient Model Construction
-Building a `Sequential` and providing it `PipelineModule` is a convenient way
+Building a `Sequential` container and providing it to a `PipelineModule` is a convenient way
 of specifying a pipeline parallel model. However, this approach encounters
-scalability issues for massive models. Starting from a `Sequential` allocates
-the model in CPU memory redundantly by every worker. A machine with 16 GPUs
+scalability issues for massive models. In this approach each worker replicates
+the whole model in CPU memory. For example, a machine with 16 GPUs
```
stas00, line 275: Please verify that I have rewritten that without breaking the meaning. The current sentence "Starting from a `Sequential` allocates..." doesn't make much sense, so I dropped it altogether, but perhaps you can fix the original, so it'd be easier to understand why this approach does so.

stas00: oh, I noticed I ended up with "this approach" in 2 subsequent sentences. Probably should refine that. But will wait for your feedback first.

ShadenSmith: Good catch! What about combining the latter sentences into something like "However, this approach encounters scalability issues for massive models because each worker replicates the whole model in CPU memory." ?

docs/_tutorials/pipeline.md
ShadenSmith: That's correct, `LayerSpec` does not allocate modules until the partitioning is complete. At that point, each rank allocates only the layers assigned to it and they are then moved to the GPU. This lets us build distributed models that wouldn't fit in CPU memory.

stas00: So is it better to leave the original modification of this PR, or replace with the more exemplifying:

> Then each worker will allocate only the layers it's assigned to. So continuing the example from the previous paragraph, a machine with 16 GPUs will need to allocate a total of 1x model size on its CPU, compared to 16x in the LayerSpec example.

stas00: OK, I went ahead with the longer version. Please double check that it's still correct.

docs/_tutorials/pipeline.md
stas00: Perfect. Done.

ShadenSmith:
Thanks so much for your ongoing contributions to DeepSpeed!

Good questions! 

> It's not clear from the tutorial why TransformerBlockPipe needs to return mask. And if there is some hidden reason, shouldn't it then return hidden too?

Pipeline parallelism is somewhat rigid in that the input to each layer's `forward()` is always the output of the previous layer. When you have a stack of transformers, we need to return the mask in order to provide it to the next layer.  We don't return `hidden` because it's not needed by the next layer. The next layer uses `output` as its `hidden` input.

From a usability perspective, I'd love to be able to specify that some inputs are static and globally available (e.g., `mask`) so that we wouldn't need to forward it down the pipeline. No concrete plans for it yet, though.

> Why do you need the vars to be a single tensor or a tuple?

Agreed that this is a limitation. As a starting place, we wanted to match `torch.nn.Sequential` functionality and that container has the same restriction. I'd like to be more flexible. The primary limitation is the communication code at stage boundaries. I've been working on a more flexible communication manager that would allow something like this and also remove some hacky code in the pipeline engine.

> wrt Tied Layers - could you please give an example?

Definitely. I'm trying to release our Megatron GPT implementation asap. The word embeddings are tied because they are used on both the first and last stages.

stas00:
Thank you for all the other answers in the previous comment, @ShadenSmith.

> > It's not clear from the tutorial why TransformerBlockPipe needs to return mask. And if there is some hidden reason, shouldn't it then return hidden too?
> 
> Pipeline parallelism is somewhat rigid in that the input to each layer's `forward()` is always the output of the previous layer. When you have a stack of transformers, we need to return the mask in order to provide it to the next layer. We don't return `hidden` because it's not needed by the next layer. The next layer uses `output` as its `hidden` input.

OK, can then this example be expanded to show how the `mask` is fed to `forward` on subsequent calls in the pipeline?
https://www.deepspeed.ai/tutorials/pipeline/#inputs-and-outputs
The issue here is that there could be other inputs and it's not clear at all how the user should handle those. i.e. you're saying "The next layer uses `output` as its `hidden` input." but this is not generic enough. 

I think all that code should be explicit in the open so that the user can see what's going on and then expand it to support other inputs.

I hope you are seeing how the provided example isn't helping with a general case.

stas00:
After studying this more I finally wrapped my head around this and I propose the following to make it crystal clear from the get going:

1. Rewrite the code as:

before:

```
class TransformerBlock(nn.Module)
    ...
    def forward(self, hidden, mask):
        hidden = self.compute(hidden, mask)
        return hidden
```
after:

```
class TransformerBlockPipe(TransformerBlock)
    def forward(self, inputs):
        hidden, mask = inputs
        hidden = super().forward(hidden, mask)
        return (hidden, mask)
```
and adding a note that:

"It's clear to see how an output of one stage becomes an input to the next stage, since the previous stage in the pipeline now returns a tuple of `(hidden, mask)` and the next stage expects besides `self`, a tuple of `(hidden, mask)`.

2. One more piece is missing from that section of the tutorial. Since the original code expected just one non-tuple variable in return from `TransformerBlock.forward` you have to modify the caller from:

```
output = self.transformer_block(hidden, mask)
```
to:
```
output, _ = self.transformer_block_pipe(hidden, mask)
```
because otherwise `output` will contain a tuple, and not `output`

Please let me know if you feel this makes things easier to understand and I will use that in this PR.

stas00:
And as I'm trying to port t5 transformers to `Pipe` I see why there is a requirement of all inputs to be Tensors - because they have to be spliced into micro-batches. 

1. If I got it right - they shouldn't just be tensors, they should be tensors with first dimension of batch size. I actually had a tensor that was tracking things, but it's a no-go, since it's of the wrong shape. So this pre-requisite can and probably should be clarified.

2. Surely, there should be a way to handle booleans here. So the input parts that are of batch size will be spliced into micro-batches, but parts that are on/off switches can be probably left alone - hmm, but then it might be difficult to put them back into `outputs` correctly - perhaps the `inputs` should have an optional `leave_me_alone_inputs` which should be replicated as is? So the sig would be:
```
def forward(self, inputs, leave_me_alone_inputs):
    ....
    return (inputs, leave_me_alone_inputs)
```
?
This would also allow passing random other inputs that don't need to be micro-batch-sliced.

3. Same issue happens to optional input parts, which could be `None` or a tensor. The Pipe needs to slice the tensor, but leave None alone.

I think I have an idea how to work around this by creating a Pipe wrapper at run time and initialize it with a late `__init__` so overcome this. I will keep you posted.

Currently in transformers t5 I have 15 distinct input parts, some tensors of BS, some `None`, some `Bool`, some tensors that aren't of BS.

ShadenSmith:
Hi @stas00 , thanks a ton for your contibutions and comments. Unfortunately I lost my first reply in a power outage. Here are a few quick comments and clarifications and I'll double back over the day.

DeepSpeed does not slice inputs into micro-batches, we just rely on the data loader to do that. We already have the concept of micro-batches for gradient accumulation and build off of that instead. Instead, we just `next()` the provided data iterator `gradient_accumulation_steps` number of times.

The `forward()` input limitations are partly a trade-off of pipeline parallelism, and partly engineering limitations. Pipeline modules use this forward implementation implicitly:
```python
def forward(self, inputs):
    x = inputs
    for layer in self.layers:
        x = layer(x)
    return x
```
This contract lets us split `self.layers` and place them on different machines. There are two consequences here:

- Inputs like booleans, etc. still need to "flow" into the layer unless they are default valued keyward args. That's where something like the `TransformerBlockPipe` subclass comes in, we can wrap the functionality of an existing module and just specialize the call into forward with any additional args. Within the pipeline, there's no call site since it's done implicitly by the engine.
- Successive layers could reside on different machines, and so any inputs must be communicated by NCCL. This is why we are limited to tensor inputs/outputs today. I agree that your recursive `to()` would help in a non-distributed environment, but we need a bit more in order to send/recv (also, the receiving processes do not know the shape/type of data to receive without an initial handshake).  Maybe we could something like pickle more general data structures and send them as `ByteTensor`s ?

stas00:
> * Inputs like booleans, etc. still need to "flow" into the layer unless they are default valued keyward args. That's where something like the `TransformerBlockPipe` subclass comes in, we can wrap the functionality of an existing module and just specialize the call into forward with any additional args. Within the pipeline, there's no call site since it's done implicitly by the engine.

Yes, this one I have already figured out and basically created a wrapper class to which I pass the flow True/False args via its `__init__`. So we are good here.

> * Successive layers could reside on different machines, and so any inputs must be communicated by NCCL. This is why we are limited to tensor inputs/outputs today. I agree that your recursive `to()` would help in a non-distributed environment, but we need a bit more in order to send/recv (also, the receiving processes do not know the shape/type of data to receive without an initial handshake).  Maybe we could something like pickle more general data structures and send them as `ByteTensor`s ?

I think what would help you to help me is to show you exactly what I'm dealing with:

1. I have a few simple things like: 

    `None`, which may become a tensor - but it shouldn't be sliced or reconstructed on the way out.

   There really should be a `None` type of tensor

2. Then I have a medium complexity things like:

   `()`. which gets filled out by each stage - e.g. it accumulates all hidden_states, attentions, etc - I have 5 of these in t5

   This one is difficult since now instead of accumulating say 6 values (for a stack of 6 blocks), I end up with 12, 18 or more depending on the number of chunks and so I'm attempting to re-construct them once we are out of the pipeline. Not simple but doable.

  Same here - it shouldn't be sliced or reconstructed on the way out.

  What I'm attempting at the moment is a closure, that I pass to the wrapper's `__init__` and grow these accumulators via the closure.

3. And the one I'm fighting with at the moment is this:

   `past_key_values` which is tuple of tuples of tensors, and the batch dimension is not even on the top level.

   This one needs to be sliced but on the inside of a tuple, but since it's a tuple of a tuple I can't pass it with `input` - I'm trying to pass it to wrapper's `__init__` instead and to do my own slicing from inside the wrapper, by deriving the size of the microbatch and counting each invocation (and hoping they are not async).
   
   I'm thinking of adding some encoding logic where I convert the tuple of a tuple into a new stack of tensors, pushing the batch dimension to the front so that it get sliced correctly and then inside `forward` pick out the section that is relevant to that block. pretty nuts. 
   
  The original logic runs a simple loop, picks the right segment for each block in the loop from the first tuple, then takes chunks of that segment via the nested tuple and eventually gets a 2 tensor slice which it adds to hidden_state in self-attention. here is the structure I need to pass around:
snapshot_3

Hope this helps to understand what I'm dealing with and I hope you have some practical ideas.

Thank you!

stas00:
let's merge this as is?

ShadenSmith (approved):

docs/_tutorials/pipeline.md
```diff
@@ -269,17 +269,18 @@ by DeepSpeed:
...
+for massive models because each worker replicates the whole model in CPU memory.
+For example, a machine with 16 GPUs must have as much local CPU memory as 16 times the model size.
 
 DeepSpeed provides a `LayerSpec` class that delays the construction of
-modules until the model layers have been partitioned across workers. Then,
-the modules are built on the GPU that owns the layer.
+modules until the model layers have been partitioned across workers.
+Then each worker will allocate only the layers it's assigned to. So, continuing the
+example from the previous paragraph, a machine with 16 GPUs will need to allocate a
+total of 1x model size on its CPU, compared to 16x in the LayerSpec example.
```
g-karthik, line 281: @stas00 @ShadenSmith I think there's a typo in this sentence, it should be "compared to 16x in the Sequential example"

stas00: Good catch, @g-karthik 

How about:

> So, comparing to the
example from the previous paragraph, a machine with 16 GPUs will need to allocate a
total of 1x model size on its CPU and not 16x.

g-karthik: Sounds fine to me!

stas00: https://github.com/microsoft/DeepSpeed/pull/888

PR #664, opened 2021-01-13
Add the flops profiler which measures the time, number of estimated flops and parameters of each module in a model 

cli99:
The original pull request is at https://github.com/microsoft/DeepSpeed/pull/544


jeffra (approved):

jeffra:
Next let's be sure to add a tutorial on deepspeed.ai and also updated json documentation here (https://www.deepspeed.ai/docs/config-json/). But we can do that in a follow-up PR.

PR #676, opened 2021-01-16
[tutorials] typos

stas00:
This PR fixes several typos, incomplete sentences and several incorrect statements.

As you can tell no human can read 50000000 and so the author made a mistake of explaining it as 50M, when it should be 500M. (fixed). Unless you switch to hjson as we have been discussing elsewhere, perhaps it'd be better to just replace `50000000` with `5e8` in the documentation.

p.s. my editor is setup to clean trailing space, so a small part of the diff is that - please let me know if that's inefficient and I will make sure future PRs will avoid whitespace changes.


jeffra:
Thanks @stas00! I pushed one update to switch from 500000000 -> 5e8 across both the zero and config docs. I think even if we switch to hjson it doesn't support 500_000_000 style so we're probably best using 5e8 style going forward for numbers. I think we will probably switch to hjson soon, it's a backwards compatible change so should be relatively easy/painless.

jeffra (approved):

stas00:
Fantastic! 

This is definitely a welcome change!

Thank you, @jeffra!

PR #677, opened 2021-01-16
Fix ZeRO 2 + Pipelining

leogao2:
ZeRO 2 + Pipelining is broken. After hunting down the issue here (https://github.com/EleutherAI/gpt-neox/issues/62), I've made a patch that seems to solve the issue but might create bad inefficiencies elsewhere.

Since this class turns `enable_backward_allreduce` off, `self.overlapping_partition_gradients_reduce_epilogue()` defined in the `DeepSpeedEngine` class never actually runs. I suspect this is because of efficiency problems; `get_flat_partition` in stage2.py might do something expensive memorywise (a quick glance at the code certainly gives off that impression); someone will have to look into that later. But in the meantime, this fixes ZeRO2 + Pipelining enough to run a demo.


ShadenSmith:
This is great, thank you!

ShadenSmith (approved):

ShadenSmith:
Hi @leogao2 , looks like our auto-formatter wants to make some changes. Can you setup [pre-commit](https://www.deepspeed.ai/contributing/#prerequisites) with updated formatting! Thanks!

ShadenSmith:
Merged, thanks @leogao2 !

PR #681, opened 2021-01-20
model parallel norm and overflow tests

ShadenSmith:


rocm-mici:

PR #687, opened 2021-01-23
Fix all Pipeline Module Parameters being sent to cuda:0

sdtblck:
so, it seems that internally, PipelineModule calls `self.to('cuda')` (L189). 

This appears to send the model parameters to GPU0, for *every visible gpu* running the deepspeed script, and makes scaling up difficult. 

Instead, we can send the parameters to the current device with `self.to(f'cuda:{self.global_rank}')`.

I'm actually unsure as to why the entire model's parameters are getting sent to one single GPU at this stage at all - aren't they meant to be partitioned equally across all GPUs?

Would appreciate anyone clearing this up for me.

### DIFF
--- a/deepspeed/runtime/pipe/module.py
+++ b/deepspeed/runtime/pipe/module.py
@@ -186,7 +186,7 @@ class PipelineModule(nn.Module):
 
         #with torch.random.fork_rng(devices=[torch.cuda.current_device()]):
         self._build()
-        self.to('cuda')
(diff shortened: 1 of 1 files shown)


sdtblck:
Actually it should be sending to local_rank, have made the changes

ShadenSmith (approved):
Thanks for catching this @sdtblck . The environment I've been testing this in had the default CUDA device already set, so I didn't catch it being broken for more general settings.

PR #695, opened 2021-01-26
Use pure Python to determine IP address instead of nonportable `hostname -I`

eyolfson:
The main moviation behind this change is that [coreutils][1] states, for
`hostname`, "portable scripts should not rely on its existence". In fact, the
[inetutils][2] version of `hostname` does not support the `-I` flag. The
[man page][3] from the expected version of `hostname` says to "not make any
assumptions about the order of the output".

This change uses pure Python to grab the first IP address that will work with
MPI. This was the location that caused problems for me, I haven't checked if
`hostname` appears anywhere else in the code. I used the first IPv4 address
that supported streams, since order should not matter.

[1]: https://www.gnu.org/software/coreutils/manual/html_node/hostname-invocation.html
[2]: https://www.gnu.org/software/inetutils/manual/html_node/hostname-invocation.html
[3]: https://linux.die.net/man/1/hostname

### DIFF
--- a/deepspeed/utils/distributed.py
+++ b/deepspeed/utils/distributed.py
@@ -55,7 +55,8 @@ def mpi_discovery(distributed_port=TORCH_DISTRIBUTED_DEFAULT_PORT, verbose=True)
     master_addr = None
     if rank == 0:
         ips = [
-            addrinfo[4][0] for addrinfo in socket.getaddrinfo(socket.gethostname(), None)
+            addrinfo[4][0] for addrinfo in socket.getaddrinfo(socket.gethostname(),
+                                                              None)
             if addrinfo[0] == socket.AF_INET and addrinfo[1] == socket.SOCK_STREAM
         ]
         master_addr = ips[0]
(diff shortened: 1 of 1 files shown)


jeffra:
Hi @eyolfson, thanks for your contribution and also for pointing out our hostname portability issue. Let's figure out a way to get this resolved. I do have some reservations on relying on hostname address resolution like this though. 

In our case (and I think many others) the hostname is a DNS resolvable name, and this mapping will be to an externally visible IP address. In addition to performance issues, this address also comes with many restrictions in terms of which ports are open, etc. What we need is a way to get the local/private IP address on the host that might be behind NAT. Here's a concrete example on one of our hosts. I imagine other setups using Docker might also have similar issues though.

image

image

eyolfson:
Hi @jeffra, thanks for the reply!

There's a [Stack Overflow][1] post that has a discussion on how to do this without any external dependencies. The solution may not be robust for your use case. A better solution may be to use [netifaces][2], but I'm not sure if you want to add another dependency. Let me know which direction would be better to go down, thanks!

[1]: https://stackoverflow.com/questions/166506/finding-local-ip-addresses-using-pythons-stdlib
[2]: https://pypi.org/project/netifaces/

PR #696, opened 2021-01-26
Improve starred expressions

eyolfson:
`deepspeed/profiling/flops_profiler/profiler.py` uses starred expressions
that are no longer valid with [PEP 617][1]. The new Python parser is in 3.9,
and this change allows DeepSpeed to run with the newest Python version. I have
not checked all locations that has this issue. However, this change allows me
to run simple examples.

[1]: https://www.python.org/dev/peps/pep-0617/


jeffra:
@cli99 can you take a look at this?

eyolfson:
Yeah, to quickly check, just grab Python 3.9 and run `python deepspeed/profiling/flops_profiler/profiler.py`. You should get:

```  File "deepspeed/profiling/flops_profiler/profiler.py", line 730
    (*input_res),
     ^
SyntaxError: can't use starred expression here```

Just for completeness: `python --version`:

```Python 3.9.1```

cli99 (approved):
Good catch. Thanks.

PR #708, opened 2021-01-28
Dist testing backend fixes, etc.

jeffra:
Misc. distributed changes, outlined below.

- [x] Update distributed testing backend to not remove args.local_rank patching. Not all unit tests were actually setting the local rank properly which led to many tests running only on one cuda device even if we had world size > 1.
- [x] Previous fix exposed hanging issue with zero-2 tests and `SimpleModel(empty_grads=True)`, empty_grads should only be true for tests where we want to test gradient imbalance issues. Only a few unit tests do this so I removed a lot of unneccisary empty_grads=True calls.
- [x] Add support for env["LOCAL_RANK"] in deepspeed.initialize, this means we (technically) no longer require args.local_rank to be passed into deepspeed.initialize via args. Currently our launcher will still pass `--local_rank <n>` to each sub-process so for now the user node will still need support for this for now.
- [x] Expose the `init_method` param for torch distributed initialization (this was motivated by user request)
- [x] Remove some unit test cases for transformer kernel that were failing on v100, @RezaYazdaniAminabadi is investigating this separately and will have an updated PR soon.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -464,7 +464,7 @@ class DeepSpeedEngine(Module):
                                                   -1))
         env_local_rank = int(os.environ.get("LOCAL_RANK", -1))
         if env_local_rank >= 0:
-            assert self.local_rank == int(env_local_rank), "Mismatch in local rank setting, args.local_rank={self.local_rank} but env['LOCAL_RANK']={env_local_rank}."
+            assert self.local_rank == int(env_local_rank), f"Mismatch in local rank setting, args.local_rank={self.local_rank} but env['LOCAL_RANK']={env_local_rank}."
 
         config_file = args.deepspeed_config if hasattr(args,
                                                        'deepspeed_config') else None
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -480,9 +480,6 @@ class DeepSpeedEngine(Module):
                 assert args.deepspeed_config is None, "Not sure how to proceed, we were given both a deepscale_config and deepspeed_config"
             args.deepspeed_config = args.deepscale_config
 
-        assert hasattr(args, 'local_rank') and type(args.local_rank) == int, \
-            'DeepSpeed requires integer command line parameter --local_rank'
-
         local_rank_err = "DeepSpeed requires a command line parameter of --local_rank [int] and/or setting the LOCAL_RANK environment variable."
         if hasattr(args, 'local_rank'):
             assert type(args.local_rank) == int, local_rank_err
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -458,11 +458,11 @@ class DeepSpeedEngine(Module):
 
     # Configure based on command line arguments
     def _configure_with_arguments(self, args, mpu):
-        self.local_rank = args.local_rank if hasattr(args,
-                                                     'local_rank') else os.environ.get(
-                                                         "LOCAL_RANK",
-                                                         -1)
-        env_local_rank = os.environ.get("LOCAL_RANK", -1)
+        self.local_rank = args.local_rank if hasattr(
+            args,
+            'local_rank') else int(os.environ.get("LOCAL_RANK",
+                                                  -1))
+        env_local_rank = int(os.environ.get("LOCAL_RANK", -1))
         if env_local_rank >= 0:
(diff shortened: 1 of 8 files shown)


deepspeed/runtime/engine.py
```diff
@@ -458,7 +458,14 @@ def _set_distributed_vars(self):
 
     # Configure based on command line arguments
     def _configure_with_arguments(self, args, mpu):
-        self.local_rank = args.local_rank if hasattr(args, 'local_rank') else 0
+        self.local_rank = args.local_rank if hasattr(
+            args,
+            'local_rank') else int(os.environ.get("LOCAL_RANK",
+                                                  -1))
+        env_local_rank = int(os.environ.get("LOCAL_RANK", -1))
+        if env_local_rank >= 0:
```
tjruwase, line 466: I think it would be nice if this validation logic was moved into _do_args_sanity_check().

tjruwase (approved):

deepspeed/runtime/engine.py
```diff
@@ -473,8 +477,15 @@ def _do_args_sanity_check(self, args):
...
             args.deepspeed_config = args.deepscale_config
 
-        assert hasattr(args, 'local_rank') and type(args.local_rank) == int, \
-            'DeepSpeed requires integer command line parameter --local_rank'
+        local_rank_err = "DeepSpeed requires a command line parameter of --local_rank [int] and/or setting the LOCAL_RANK environment variable."
+        if hasattr(args, 'local_rank'):
+            assert type(args.local_rank) == int, local_rank_err
+            if "LOCAL_RANK" in os.environ:
+                env_local_rank = int(os.environ.get("LOCAL_RANK", -1))
+                assert env_local_rank == args.local_rank, \
```
g-karthik, line 485: @jeffra SOS! the addition of this assertion is causing my jobs to fail now!

Why was this assert added? A downstream script can have `args.local_rank` existing in its argument parser and yet be totally stale until DeepSpeed overrides its value via MPI discovery, etc.

For instance, a downstream script may be written in a manner where it can be used for both single-GPU (without DeepSpeed) and multi-GPU training (with as well as without DeepSpeed), wherein single-GPU training is equivalent to defaulting to `args.local_rank = -1`.

g-karthik (commented):
@jeffra please take a look at my inline comment! Your addition of that specific assert is causing my jobs to fail now.

EDIT: Oh, hmm, looks like you addressed this in a follow-up PR https://github.com/microsoft/DeepSpeed/pull/720. Perhaps my build doesn't contain this fix. Will re-build and test.

g-karthik:
@jeffra Okay so with the latest build which includes https://github.com/microsoft/DeepSpeed/pull/720, the assertion error I was getting earlier went away.

Having said that, I see with this latest build of DeepSpeed that it is no longer updating `args.local_rank` to contain the right value of `local_rank`. My client script's `args.local_rank` remains -1 even after the `deepspeed.initialize()` call. This change in behavior, in turn, is affecting my downstream logic such as ensuring that only the 0th world rank process writes to disk, etc.

Earlier, the behavior was such that if the client script has the command-line arg, then deepspeed would overwrite the arg to contain the right values as identified by `mpi4py` (via the setting of the environment variables in the `init_distributed()` method call prior to args sanity checks and configurations in the `DeepSpeedEngine`).

Would appreciate any assistance here!

jeffra:
Hi @g-karthik, sorry for my delayed response on this blocking issue :( Glad to hear that #720 fixed the issue. Are you installing via PyPI or from source? PyPI is a bit out of date with master right now.

I believe #764 should fix the issue you are now seeing. Can you give the PR branch a try though?

g-karthik:
Hi @jeffra I'm installing from source.

Looks like you've merged it to master already, so I'll try the master branch!

I'm a bit confused though - which commit caused this regression? I'm unable to find it.

g-karthik:
@jeffra the fix in https://github.com/microsoft/DeepSpeed/pull/764 seems to work, thanks for this!

But for my sanity, can you point me to the commit where this was originally "undone", i.e., `args.local_rank` was no longer being set properly within deepspeed?

I'm trying to reconcile this particular behavior with past commits and pretty much all I see is `self.local_rank` assignments, not `args.local_rank` assignments.

Without changing a single line of code in my scripts, I see that without your https://github.com/microsoft/DeepSpeed/pull/764 fix, every rank tries to write a model checkpoint directory. But that was not happening 2-3 weeks ago, i.e., only one rank used to write a model checkpoint directory.

jeffra:
I think i may have added this regression in #608

Issue #710, opened 2021-01-29
stuck in trying to combine PP with DeepSpeed


stas00:
I have been trying many different ways, but I'm still stuck trying to combine a Pipeline (from pytorch-nightly) with DeepSpeed.

I adapted MPU from https://github.com/microsoft/DeepSpeedExamples/tree/master/Megatron-LM/mpu, which by running some tests on seems to produce what I need. 

But I can't figure out how to launch DeepSpeed to make things work.

## Setup 4 gpus

2 model parallel groups:  [g0, g1], [g2, g3]
2 data parallel groups:  [g0, g2], [g1, g3]

Inside the pipeline I correctly switch model layers to either [g0, g1] or [g2, g3].

I pass mpu to deepspeed init as suggested in the docs:
```
    model, optimizer, _, lr_scheduler = deepspeed.initialize(
        args=SimpleNamespace(**ds_args),  # expects an obj
        model=model,
        model_parameters=model_parameters,
        config_params=config,
        mpu = mpu,
    )
```

Now how do I launch deepspeep to use my PP and to achive 2D Parallelism? Do I tell it about 4 gpus or just gpus 0 and 2?

1. if I tell it about 4 gpus (well, letting it see all gpus) - it launches 4 processes which I think is wrong, and it crashes with:

```
CUDA_VISIBLE_DEVICES=0,1,2,3,  deepspeed  program
[...]
  File "/home/stas/hf/transformers/src/transformers/models/t5/modeling_t5.py", line 256, in forward
    return self.weight * hidden_states
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:2 and cuda:1!
```
gpus 2 and 1 shouldn't end up together in the pipeline - they are in the wrong groups.

2. if I tell it only about the gpus 0+2 it launches 2 processes, but then crashes with:

```
CUDA_VISIBLE_DEVICES=0,1,2,3,  deepspeed --include localhost:0,2 program

  0%|                                                                                                                                  | 0/2 [00:00<?, ?it/s]Traceback (most recent call last):
  File "./finetune_trainer.py", line 373, in <module>
    main()
  File "./finetune_trainer.py", line 303, in main
    train_result = trainer.train(
  File "/home/stas/hf/transformers/src/transformers/trainer.py", line 919, in train
    tr_loss += self.training_step(model, inputs)
  File "/home/stas/hf/transformers/src/transformers/trainer.py", line 1286, in training_step
[...]

PR #714, opened 2021-01-30
[launcher] look ma, no more zombies

stas00:
Almost identical to the solution merged into pytorch https://github.com/pytorch/pytorch/pull/49305, this PR automatically takes care of killing zombies, should one of the processes fail (or get killed) or the parent process gets manually terminated by the operator.

As I already explained the 2 problems and the solution in full details here: https://github.com/pytorch/pytorch/pull/49305 I hope it's ok if I don't repeat it again.

Thank you.


jeffra:
Thank you @stas! This has been on our backlog to investigate for a long time and a huge annoyance. Can’t wait to try this out soon.

stas00:
Let's trade - I will fix annoying issues and someone give me a bit of guidance with the 2D integration - as it looks really not going anywhere - says supports non-DeepSpeed pipeline but it doesn't look it is the case :( https://github.com/microsoft/DeepSpeed/issues/710 really could use help with someone who has the deepspeed internals understanding.

jeffra (approved):

jeffra:
Thanks @stas00 for the help here! We've tested on our side and this is awesome. One remaining note though, this will fix the zombie problem on a single-node/multi-process setup however the issue currently still remains on a multi-node/multi-process setup. I think it shouldn't be too bad to extend this to work on multi-node though now that we have the single-node/multi-process setup working.

PR #727, opened 2021-02-04
enable apex.optimizers.FusedAdam and deepspeed fused fp16 optimizer

ShadenSmith:
If a client provides an optimizer of type `apex.optimizers.FusedAdam`, it won't be given a fused FP16 optimizer. The `isinstance()` query here only checks for DeepSpeed's `FusedAdam`.

The implementation here was just hacked together to test the logic. We could also test for `apex.optimizers.FusedAdam` specifically, and just be careful not handle missing `apex`?

Would also love to put together a list of requirements for compatibility!

Major question: can we merge this without breaking existing checkpoints? We may need to handle loading from an unfused fp16 optimizer checkpoint.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -16,7 +16,7 @@ from deepspeed.runtime.zero.stage2 import FP16_DeepSpeedZeroOptimizer
 from deepspeed.runtime.zero.stage1 import FP16_DeepSpeedZeroOptimizer_Stage1
 from deepspeed.runtime.zero.utils import is_zero_supported_optimizer
 from deepspeed.runtime.activation_checkpointing import checkpointing as activation_checkpointing
-from deepspeed.runtime.fp16.fused_optimizer import FP16_Optimizer
+from deepspeed.runtime.fp16.fused_optimizer import FP16_Optimizer, FP16_FUSED_SUPPORTED_OPTIMIZERS, is_fp16_fused_supported_optimizer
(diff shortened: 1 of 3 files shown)


deepspeed/runtime/engine.py
```diff
@@ -632,7 +632,10 @@ def _configure_fp16_optimizer(self, optimizer):
         dynamic_loss_args = self.dynamic_loss_scale_args()
         clip_grad = self.gradient_clipping()
         if isinstance(optimizer,
-                      FusedAdam) or self.optimizer_name() == ONEBIT_ADAM_OPTIMIZER:
+                      FusedAdam) or self.optimizer_name() in [
+                          'FusedAdam',
```
jeffra, line 636: Maybe we can do this in a similar way as what we're doing with zero compatible optimizers. As in, attempt to import apex and append apex's FusedAdam to a list. I see that `isinstance` accepts a tuple as a 2nd arg which should make it a bit easier to check across multiple classes. 

The annoying part is we still have to check 1bit-adam via string since if we import the class it currently requires MPI, hopefully we can clean that up in the upcoming changes using NCCL so there are no dependency issues simply from the import (cc: @awan-10, @conglongli).

https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/zero/utils.py#L26-L38

ShadenSmith: Sounds good, I pushed an update.

rocm-mici:

PR #732, opened 2021-02-06
Replace timer print rank 0 with logging

SeanNaren:
Codebase is awesome, thanks for the great work guys :)

To reduce the logging stdout when training it would be nice to swap to using logging rather than print to configure verbosity.

Let me know if theres any thoughts/issues!

### DIFF
--- a/deepspeed/utils/timer.py
+++ b/deepspeed/utils/timer.py
@@ -4,6 +4,7 @@ Copyright 2019 The Microsoft DeepSpeed Team
 
 import time
 import torch
+from deepspeed.utils.logging import log_dist
 
 from deepspeed.utils import logger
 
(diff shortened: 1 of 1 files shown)


jeffra (approved):

ShadenSmith (approved):

deepspeed/utils/timer.py
```diff
@@ -98,9 +91,9 @@ def log(self, names, normalizer=1.0, reset=True, memory_breakdown=False):
                     reset=reset) * 1000.0 / normalizer
                 string += ' | {}: {:.2f}'.format(name, elapsed_time)
 
-        # TODO: use our logging utilitied to selectively print. Useful for model
-        # parallelism because rank=0 is too restrictive.
-        print_rank_0(string)
+        # TODO: expose ranks as rank=0 is too restrictive
+        # useful for model parallelism
+        log_dist(string, ranks=[0])
```
ShadenSmith, line 96: Can we add an optional `ranks` kwarg to `log()`? We will need more printing flexibility if using forms of model parallelism.

SeanNaren: Done! let me know if it looks good :)

PR #734, opened 2021-02-08
Only call Init Distributed if required

SeanNaren:
Adds an else statement to make sure we only call `init_distributed` if required. 

I added this due to running into an `mpi4py` missing requirement issue, since we do not expose the `auto_mpi_discovery` flag within the `init_distributed` function. As a result by ensuring we use the check first to see if distributed has already been initialized, we can omit calling this check.


jeffra (approved):
Good catch, thank you!

PR #741, opened 2021-02-10
Activation checkpointing for non-tensor arguments and return values

tjruwase:
Improve generality of activation checkpointing to support non-tensor input and output types

### DIFF
--- a/deepspeed/runtime/activation_checkpointing/checkpointing.py
+++ b/deepspeed/runtime/activation_checkpointing/checkpointing.py
@@ -23,6 +23,7 @@ from torch.cuda import _lazy_call, device as device_ctx_manager
 
 from deepspeed.runtime.config import DeepSpeedConfig
 from deepspeed.utils import logger
+from deepspeed.runtime.utils import move_to_device
 from deepspeed.utils.timer import SynchronizedWallClockTimer as Timers
 
 #DeepSpeed Checkpointing Enabled or Disabled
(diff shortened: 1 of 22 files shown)


deepspeed/runtime/activation_checkpointing/checkpointing.py
```diff
@@ -311,6 +311,44 @@ def get_full_inputs(tensors, device=None):
...
+        return [move_to_device(v, device) for v in item]
+    elif isinstance(item, tuple):
+        return tuple([move_to_device(v, device) for v in item])
+    elif isinstance(item, dict):
+        return {k: move_to_device(v, device) for k, v in item.items()}
+    else:
+        return item
+
+
+def filter_tensor_values(all_values):
```
ShadenSmith, line 327: Super minor: the `value` in the names of the filter/merge routines first made me think we're filtering based on elements of the tensors (e.g., remove NaNs).  Maybe rename to something along the lines of `extract_tensors()` ?

deepspeed/runtime/activation_checkpointing/checkpointing.py
```diff
@@ -311,6 +311,44 @@ def get_full_inputs(tensors, device=None):
...
+    tensor_values = [v for v in all_values if torch.is_tensor(v)]
+    non_tensor_values = [v for v in all_values if not torch.is_tensor(v)]
+    tensor_flags = [torch.is_tensor(v) for v in all_values]
+    if type(all_values) is tuple:
+        return tuple(tensor_values), tuple(non_tensor_values), tuple(tensor_flags)
+    return tensor_values, non_tensor_values, tensor_flags
+
+
+def merge_tensor_values(tensor_values, non_tensor_values, tensor_flags):
```
ShadenSmith, line 336: Not relevant to this PR, but this seems like a pattern that could be reused in other places? I think in the future these merge/filter routines could be generalized and reused to do things like split/merge mixed-type lists of tensors for bucketing.

ShadenSmith (approved):

deepspeed/runtime/activation_checkpointing/checkpointing.py
```diff
@@ -311,6 +311,44 @@ def get_full_inputs(tensors, device=None):
     return tuple(inputs)
 
 
+def move_to_device(item, device):
```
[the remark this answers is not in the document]
tjruwase: Will do.

PR #742, opened 2021-02-10
[doc] 1bit_adam dependencies

stas00:
This PR adds instructions to install 1bit_adam dependencies

I think there are further fixes that are needed.

1. the config at https://www.deepspeed.ai/tutorials/onebit-adam/#32-configuration-for-bingbertsquad-with-deepspeed-and-1-bit-adam-enabled
suggests the user to use:
```
"max_grad_norm": 1.0,
```
but the program crashes when I add it:

```
  File "/mnt/nvme1/code/github/00optimize/DeepSpeed-zombies/deepspeed/__init__.py", line 110, in initialize
        engine = DeepSpeedEngine(args=args,engine = DeepSpeedEngine(args=args,

  File "/mnt/nvme1/code/github/00optimize/DeepSpeed-zombies/deepspeed/runtime/engine.py", line 174, in __init__
    self._configure_optimizer(optimizer, model_parameters)
  File "/mnt/nvme1/code/github/00optimize/DeepSpeed-zombies/deepspeed/runtime/engine.py", line 549, in _configure_optimizer
    basic_optimizer = self._configure_basic_optimizer(model_parameters)
  File "/mnt/nvme1/code/github/00optimize/DeepSpeed-zombies/deepspeed/runtime/engine.py", line 592, in _configure_basic_optimizer
    basic_optimizer = self._configure_basic_optimizer(model_parameters)
  File "/mnt/nvme1/code/github/00optimize/DeepSpeed-zombies/deepspeed/runtime/engine.py", line 592, in 
    raise ValueError(
ValueError: 'max_grad_norm' is not supported as an optimizer parameter, please switch to using the deepspeed parameter 'gradient_clipping' see: https://www.deepspeed.ai/docs/config-json/#gradient-clipping for more details
```

2. Here https://www.deepspeed.ai/tutorials/onebit-adam/#pre-requisites-for-1-bit-adam it says one has to use the special launcher but it works just fine without the launcher, that is I was able to run it with just:

```
deepspeed script.py
```
The config I used was (logger dump):
```
    "fp16":{
        "enabled":true,
        "hysteresis":2,
        "initial_scale_power":16,
        "loss_scale":0,
        "loss_scale_window":1000,
        "min_loss_scale":1
    },
    "gradient_accumulation_steps":1,
    "gradient_clipping":1.0,
    "optimizer":{
        "params":{
            "bias_correction":false,
            "cuda_aware":true,
            "freeze_step":400,
            "lr":0.0002,
            "weight_decay":0.01
        },
        "type":"OneBitAdam"
    },
    "scheduler":{
        "params":{
            "warmup_max_lr":3e-05,
            "warmup_min_lr":0,
            "warmup_num_steps":500
        },
        "type":"WarmupLR"
    },
    "steps_per_print":2000,
    "train_micro_batch_size_per_gpu":4,
    "wall_clock_breakdown":false,
    "zero_allow_untested_optimizer":true,
    "zero_optimization":{
        "allgather_bucket_size":200000000.0,
        "allgather_partitions":true,
        "contiguous_gradients":true,
        "cpu_offload":true,
        "overlap_comm":true,
        "reduce_bucket_size":200000000.0,
        "reduce_scatter":true,
        "stage":2
    }
}
```

### DIFF
--- a/docs/_tutorials/onebit-adam.md
+++ b/docs/_tutorials/onebit-adam.md
@@ -29,7 +29,13 @@ cd DeepSpeedExamples/
 
 1-bit Adam uses advanced communication schemes that are not yet supported by PyTorch distributed and NCCL. We rely on Message Passing Interface (MPI) for these advanced communication primitives.
 
-We package the necessary dependencies in the DeepSpeed docker images. However, if you are using a different build system, please install MPI and mpi4py on your system. We have tested CUDA-Aware MPI communication using the [MVAPICH2-GDR](http://mvapich.cse.ohio-state.edu/userguide/gdr/) library. However, any CUDA-Aware communication library including [OpenMPI](https://www.open-mpi.org/) should work fine with these examples.
+We package the necessary dependencies in the DeepSpeed docker images. However, if you are using a different build system, please install MPI and mpi4py on your system. To install the prerequisites run:
+
+```shell
+pip install deepspeed[1bit_adam]
+```
+
+We have tested CUDA-Aware MPI communication using the [MVAPICH2-GDR](http://mvapich.cse.ohio-state.edu/userguide/gdr/) library. However, any CUDA-Aware communication library including [OpenMPI](https://www.open-mpi.org/) should work fine with these examples.
 
 An example launch command for 1-bit Adam using the `deepspeed` launcher is as follows:
 


conglongli:
@stas00 Thanks for the PR and the questions! For your first question, yes it is a typo in our tutorial, let me make a commit to your PR to fix it. For your second question, because currently 1-bit Adam uses MPI backend to communicate during the compression stage (after "freeze_step" steps), so without the mvapich or openmpi launcher it may crash after reaching the compression stage, or you may not be able to achieve the expected throughput gain. We recently wrote an arxiv paper about 1-bit Adam so feel free to take a read if interested in the details https://arxiv.org/abs/2102.02888. If you have more questions about 1-bit Adam, feel free to let me and @awan-10 know. Thanks!

stas00:
Thank you, @conglongli!

I was just running a quick does-it-work-test trying to rule our an error a user reported in transformers, but it should really be submitted here, since the crash is somewhere in the DeepSpeed code. So I asked her to file an Issue here.

> For your second question, because currently 1-bit Adam uses MPI backend to communicate during the compression stage (after "freeze_step" steps), so without the mvapich or openmpi launcher it may crash after reaching the compression stage, or you may not be able to achieve the expected throughput gain. 

Perhaps it'd be useful to add your comment as a note, so that unwary user like myself will not conclude that all is good based on a quick test when seeing the training complete without a hitch?

conglongli:
> Perhaps it'd be useful to add your comment as a note in the section where the launcher is discussed, so that unwary user like myself will not conclude that all is good based on a quick test when seeing the training complete without a hitch?

@stas00 Yes I agree. Will do. I noticed that your PR is from your fork. So I will just approve and merge your PR first, and then create another PR to fix your two questions. And let us know when you have more bandwidth/interests/questions about 1-bit Adam :)

conglongli (approved):
LGTM

Issue #675, opened 2021-01-16
[zero tutorial] unreadable image


stas00:
This tutorial https://www.deepspeed.ai/tutorials/zero/ has this image which gets scaled to fit the narrow width and as a result it gets rendered unreadable, here is a snapshot:
![snapshot_1](https://user-images.githubusercontent.com/10676103/104793947-9ba63900-5759-11eb-9f67-96921665e071.png)

The image is fine if I find a way to view its full size, which is 
https://www.deepspeed.ai/assets/images/oom_dp8_1.5B_log.png
but it's not clickable.

So either it needs to be made clickable, or perhaps the error can be just shown as code quote.

edit: it happens again later in the same tutorial.

-----------------

edit: very bottom of https://www.deepspeed.ai/tutorials/zero/ - prev/next navigation - there are more tutorials but [Next] is linking nowhere... you have added 3 more since then that can be seen in the menu on the left.

tjruwase:
@stas00 Finally got around to this: #746. Can you please check if the images are better now?

stas00:
It's impossible to tell until this is rendered on your website - but I see that you made them clickable, so that should do the trick I think.

Thank you for fixing that, @tjruwase

tjruwase:
@stas00, I am reopening this to address the following issues

I have created PR #761 to address both issues. But below are the details. 

1) I presume you see an issue with the ZeRO-Offload [tutorial](https://www.deepspeed.ai/tutorials/zero-offload/)? Sorry, that was because I overlooked this tutorial and had only fixed the ZeRO [tutorial ](https://www.deepspeed.ai/tutorials/zero/). 

2) For the question below, the issue is a mismatch between the order of the menu on the left and the prev/next navigation. While the menu on the left is in arbitrary order (loosely creation order), the prev/next navigation is in alphabetical order. And so the ZeRO tutorials are rightfully the left ones on the prev/next navigation. I have tweaked the naming on the menu on the left to mitigate the mismatch. 
> edit: very bottom of https://www.deepspeed.ai/tutorials/zero/ - prev/next navigation - there are more tutorials but [Next] is linking nowhere... you have added 3 more since then that can be seen in the menu on the left.

stas00:
[...]

PR #746, opened 2021-02-10
Clickable screenshots

tjruwase:
Make screenshots clickable for easier viewing to address #675.


jeffra (approved):

stas00:
@tjruwase, could you please rebuild the website? Your change is not showing up.

Also there is a small typo:

> we also need to set cpu_offload flag to true enable ZeRO-Offload optimizations

it's missing a 2nd "to" should be "to true to enable"

might be easier for your to push it directly :)

PR #773, opened 2021-02-20
[doc] fix incorrect param name

stas00:
Invalid param name

Thanks.

### DIFF
--- a/docs/_pages/config-json.md
+++ b/docs/_pages/config-json.md
@@ -165,7 +165,7 @@ Example of ***scheduler***
 
 | Description                                                                                                                                                                                                   | Default |
 | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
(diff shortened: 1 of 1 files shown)


jeffra (approved):
Good catch

stas00:
Is there are reason why you use this weird markdown:
```
***loss\_scale\_window***
```
and not just:
```
**loss_scale_window**
```

I'm not sure why you had to do that in first place, but it also prevents `grep` from working and finding such bugs as fixes in this PR.

stas00:
Here is a little perl magic to quickly rewrite your docs, e.g. on:
```
perl -pi -e 'BEGIN{ sub a{$_=$_[0]; ~s#\\##g; $_}} s|\*\*\*(\S+)\*\*\*|"<i>**".a($1)."**</i>"|msge' docs/_pages/config-json.md
```

To change all .mds:
```
find . -name "*md" -exec perl -pi -e 'BEGIN{ sub a{$_=$_[0]; ~s#\\##g; $_}} s|\*\*\*(\S+)\*\*\*|"<i>**".a($1)."**</i>"|msge' {} \;
```

it seems to change just 2 files.

PR #775, opened 2021-02-20
[docs] make bold+italic work without escaping _

stas00:
one can't `grep` for params in the docs because of escaped `_` as in ` ***loss\_scale\_window***`, which makes it
1. difficult to find docs
2. allows for mistakes as in https://github.com/microsoft/DeepSpeed/pull/773 where a param is renamed but the search-n-replace misses the escaped versions.

so this PR uses an alternative bold+italics syntan which doesn't require escaping. `<i>**A_B**</i>` test: <i>**A_B**</i>

Also looking at diff some `***A_B***` weren't escaped in first place.

This was done with:

```
find . -name "*md" -exec perl -pi -e 'BEGIN{ sub a{$_=$_[0]; ~s|\\||g; $_}} s|\*\*\*(\S+)\*\*\*|"<i>**".a($1)."**</i>"|msge' {} \;
```

I used it for all `***`'s inputs, but can change to have only for the ones containing `_` - I just thought it'd be more consistent that way. Especially when new entries are added and copy-n-paste is done, so `***foo***` won't be copied and edited as `***foo_bar***` (i.e. escape missed because it wasn't in the original) as it's in a few places in the docs now.

But it's your call.


tjruwase (approved):

tjruwase:
@stas00  Thanks so much for these significant docs improvements. They are greatly appreciated.

stas00:
Actually, I'm not sure you needed to escape `_` in the first place, if you look at `reduce_bucket_size` prior to this PR - it's not escaped:
```
***reduce_bucket_size***: [boolean]

| Description                                                                                                         | Default |
| ------------------------------------------------------------------------------------------------------------------- | ------- |
| Number of elements reduced/allreduced at a time. Limits the memory required for the allgather for large model sizes | `5e8`   |
```

Yet, it renders just fine:
snapshot_19

Is it possible that just removing `\` is enough and leaving `***` is just fine?

In any case why do you even need bold+italic - italic fonts almost never aid readability. Most projects use fixed font for variables without bold or italic, so it reads even better.

PR #791, opened 2021-02-24
Fix the bias-add and add the layer-norm-eps parameter

RezaYazdaniAminabadi:
Fix the bias-add precision and indexing and also adding the layer-norm-eps as a configurable parameter for transformer


csrc/transformer/transform_kernels.cu
```diff
@@ -259,11 +259,21 @@ __global__ void bias_add_transform_0213<__half>(__half* output,
...
+        vals_arr_f[l].x += bias_arr_f[l].x;
+        vals_arr_f[l].y += bias_arr_f[l].y;
+        output_half[l] = __float22half2_rn(vals_arr_f[l]);
+    }
 
-    output_half[0] = vals_half[0] + bias_half[0];
-    output_half[1] = vals_half[1] + bias_half[1];
-    output_half[2] = vals_half[2] + bias_half[2];
-    output_half[3] = vals_half[3] + bias_half[3];
+    // output_half[0] = vals_half[0] + bias_half[0];
```
tjruwase, line 273: Please remove these lines.

tjruwase (approved):

csrc/transformer/transform_kernels.cu
RezaYazdaniAminabadi: I was thinking to use this part of the code in another PR and make the half-addition configurable. I am however not sure if that would impact the performance a lot! What do you think?

RezaYazdaniAminabadi: I have add the __ACC_HALF__ in the kernel code, which we can set it later at compile time. Is this fine?

PR #804, opened 2021-02-28
issue with the implementation of column_sum_reduce

zmx19951103:
hi, i take a look at the code of column_sum_reduce, i have 2 questions:
   1. the goal of column_sum_reduce is to get the column sum of inp matrix with shape[rows, width] and the result shape should be [width],right ? It seems that the judgment condition of pos is not suitable
   2. the implementation of cuda kernel based on the asumption that, the thread with same threadIdx.y will group into a thread_block_tile, the blockDim is (32,32), i read the nvidia document https://on-demand.gputechconf.com/gtc/2017/presentation/s7622-Kyrylo-perelygin-robust-and-scalable-cuda.pdf, THREAD BLOCK TILE is a subset of threads of a thread block, divided into tiles in row-major order. doesn't it mean thread with the same threadIdx.x will group into a thread_block_tile ?
thanks !!!!

### DIFF
--- a/csrc/transformer/general_kernels.cu
+++ b/csrc/transformer/general_kernels.cu
@@ -43,7 +43,7 @@ __global__ void column_sum_reduce(const T* __restrict__ inp,
 
     if (threadIdx.x == 0) {
         int pos = blockIdx.x * TILE_DIM + threadIdx.y;
-        if (pos < (rows * width)) out[pos] = sum;
+        if (pos < width) out[pos] = sum;
     }
 }
 


csrc/transformer/general_kernels.cu
```diff
@@ -43,7 +43,7 @@ __global__ void column_sum_reduce(const T* __restrict__ inp,
 
     if (threadIdx.x == 0) {
         int pos = blockIdx.x * TILE_DIM + threadIdx.y;
-        if (pos < (rows * width)) out[pos] = sum;
+        if (pos < width) out[pos] = sum;
```
RezaYazdaniAminabadi, line 46: Thanks for fixing this! I would say it still was working when the hidden dimension was dividable by 32, however, it would have a caused a memory leak for when the hidden is not dividable by 32!

RezaYazdaniAminabadi:
> hi, i take a look at the code of column_sum_reduce, i have 2 questions:
> 
> 1. the goal of column_sum_reduce is to get the column sum of inp matrix with shape[rows, width] and the result shape should be [width],right ? It seems that the judgment condition of pos is not suitable
> 2. the implementation of cuda kernel based on the asumption that, the thread with same threadIdx.y will group into a thread_block_tile, the blockDim is (32,32), i read the nvidia document https://on-demand.gputechconf.com/gtc/2017/presentation/s7622-Kyrylo-perelygin-robust-and-scalable-cuda.pdf, THREAD BLOCK TILE is a subset of threads of a thread block, divided into tiles in row-major order. doesn't it mean thread with the same threadIdx.x will group into a thread_block_tile ?
>    thanks !!!!

Hi @zmx19951103 

Thanks for fixing this bug. Regarding your second question, I think both x and y dimensions are assigned to different thread_block tiles, however, since this is a 2-dimensional tile, we are just using the the threadIx.y for saving the output after all is reduced across each tile ([here ](https://github.com/microsoft/DeepSpeed/pull/804/files#diff-4d9c9a0e5761838a7574e74b1ced66057684ca7817627fa59073deed1b0b79a5R42)) whose got the same y index and x index changes from 0 to 31. So, what you are saying is also true, and this is also our assumption when reducing the elements in a row!
I hope this answered your question.
Thanks, 
Reza

RezaYazdaniAminabadi (approved):

csrc/transformer/general_kernels.cu
zmx19951103: yes! thanks for your approval!!

PR #812, opened 2021-03-02
Fixing gelu_checkpointing memory issue

RezaYazdaniAminabadi:
This PR solves the memory error when enabling gelu-checkpoint. I also checked this modification for the rest of memory optimization flags inside transformer layer, and they all passed the unit tests.

@owmohamm, can you please verify if this is working on your side?
Thanks


owmohamm:
Tested out the changes locally and are working fine. With my configuration which is analogous to Bert-Large memory savings are as follows

Flags Set| Memory Used
-- | --
None | 30252MiB
gelu_checkpoint | 27308MiB
normalize_invertible | 27650MiB
normalize_invertible + gelu_checkpoint | 23184MiB
normalize_invertible + gelu_checkpoint + attn_dropout_checkpoint | 21686MiB

eltonzheng (approved):

PR #825, opened 2021-03-05
Allow args to be optional in deepspeed.initialize

jeffra:
Before this PR `args` passed into deepspeed.initialize must be an object with [get|set]attribute support. This PR allows users to initialize deepspeed via `deepspeed.initialize(model=model, config_params=config_dict)`. This assumes that local_rank is set as an environment variable (which is already done via deepspeed/pytorch.distributed launchers and/or from deepspeed.init_distributed).

Can only do the above call with kwargs, positional args need to be backwards compatible to support `deepspeed.initialize(args, model)`. NOTE: this change does not require changes to existing code that use deepspeed.initialize with args.

This PR allows hugging face and lightning to not have to pass a `SimpleNamespace` object to deepspeed.initialize which was previously a hack to get deepspeed working.

For hugging face integration this means they can (optionally) remove the `SimpleNamespace` object passed to local rank, since it will be set since `deepspeed.init_distributed` is already called prior to deepspeed.initialize which ensures the local rank is set properly. https://github.com/huggingface/transformers/blob/256482ac9285c467fb97ca3b1b693a4de1d0ac60/src/transformers/integrations.py#L409-L414  /cc @stas00

For lightning integration I don't believe it is using `deepspeed.init_distributed` for launching but if they assume torch.distributed launcher then this is fine, see: https://github.com/pytorch/pytorch/blob/master/torch/distributed/launch.py#L253. They can also remove `SimpleNamespace` here as well then. https://github.com/PyTorchLightning/pytorch-lightning/blob/efda48faab666b78bb5f71bff4a0838f9b82dee5/pytorch_lightning/plugins/training_type/deepspeed.py#L243 /cc @SeanNaren 

In either case hf/lightning could set or assert `os.environ['LOCAL_RANK'] = self.local_rank` before calling deepspeed.initialize.


stas00:
That is indeed much simpler. Thank you, @jeffra!

We will have to wait for your new release to implement that so that we can require that version number.

But probably need to wait for @cli99 to sort out the issue w/ non-DeepSpeed optim/scheduler before the new release if it's not too far.

SeanNaren:
Nice changes @jeffra! Regarding lightning not using `deepspeed.init_distributed`, I think because this uses auto_mpi discovery as default, which required mpi4py to be installed, hence we just do a torch.distributed init process group ourselves. I may be incorrect now, will check :)

jeffra:
> Nice changes @jeffra! Regarding lightning not using `deepspeed.init_distributed`, I think because this uses auto_mpi discovery as default, which required mpi4py to be installed, hence we just do a torch.distributed init process group ourselves. I may be incorrect now, will check :)

Oh I see, interesting. `deepspeed.init_distributed` will only use the auto_mpi discovery code path (which requires mpi4py) if the training process(es) were launched without the deepspeed launcher or the [torch distributed launcher](https://github.com/pytorch/pytorch/blob/master/torch/distributed/launch.py).

deepspeed/runtime/engine.py
```diff
@@ -496,8 +496,9 @@ def _configure_with_arguments(self, args, mpu):
         # environment variable is set. We must align args.local_rank to this value for
         # backwards compatability with scripts relying on [args|self].local_rank containing
         # the correct local rank info.
-        args.local_rank = int(os.environ['LOCAL_RANK'])
-        self.local_rank = args.local_rank
+        self.local_rank = int(os.environ['LOCAL_RANK'])
```
tjruwase, line 499: Could we add a sanity check that 'LOCAL_RANK' is in os.environ?

tjruwase (approved):

deepspeed/runtime/engine.py
stas00: `int(os.environ.get('LOCAL_RANK', '-1'))` ?

jeffra: If we set local rank to -1 in this section of the code we'll attempt to send the model to device -1 which would error out later. I'll add an assert with a useful error message here.

stas00: well, I only meant that -1 can be a default for no-local-rank, which is what we use and of course check later - but you're right deepspeed always requires `local_rank> -1`  so the fallback I proposed makes no sense.

PR #827, opened 2021-03-06
Bug fix: Remove client optimizer param_group list item that does not have 'params'

cli99:
This PR solves the problem where the client optimizer created by users might contain param_group that does not have 'params'. This problem used to cause crashes in zero stage 1 and 2 where 'params' value is needed.


deepspeed/runtime/engine.py
```diff
@@ -544,6 +544,13 @@ def _configure_distributed_model(self, model):
     def _configure_optimizer(self, client_optimizer, model_parameters):
 
         if client_optimizer is not None:
+            for i, param_group in enumerate(client_optimizer.param_groups):
```
samyam, line 547: You are changing an iterable variable during the iteration. Doesn't the indices changes once you pop items out inside the loop? I am not sure if the indices will remain consistent when you have multiple groups that are empty. Can you please test for such a scenario?

cli99: The indices would change after poping items. I changed the logging to not print the indices.

samyam: I'm actually not concerned about the indices in the print but rather during the pop.  If you have popped index 0, then what would be the value of i at the next iteration?, would it still be 0 or 1? If its 1, and you need to pop it as well, then you would be popping the wrong element since what was index 1 at the beginning is now index 0 since you have already popped the original index 0

cli99: It would be 1 and actually, the item originally at 1 (now at 1 after the first item was poped) would be skipped for checking. I changed that. Thanks.

samyam (approved):

PR #833, opened 2021-03-08
less scary overflow notice

stas00:
This all-caps OVERFLOW in:
```
[deepspeed] OVERFLOW! Skipping step. Attempted loss
```
is quite intimidating for the users and makes them feel that something is wrong. 

This PR is suggesting a more Info-style no-caps version:
```
[deepspeed] Overflow! Skipping step. Attempted loss
```

The biggest problem here is that the "overflow" message lacks context. It'd be even more informative and less confusing to specifically say what it is an overflow of. Could it say perhaps:
```
[deepspeed] fp16 dynamic loss scale overflow! Skipping step. Attempted loss
```
? Now users can act on it, since they know where to look for information to adjusting this if they prefer to get the optimizer to work from step 1.

It would have been even better, IMHO, if only 1 message were printed e.g. at the point when the first step is finally taken, e.g. as in:

```
First 22 steps were skipped due to fp16 dynamic loss scale overflow, starting stepping from step 23.
```
Otherwise there is a huge flurry of these messages.


jeffra (approved):

stas00:
ok, as discussed changed it to:
```
[deepspeed] fp16 dynamic loss scale overflow! Skipping step. Attempted loss 
```

also fixed `[deepscale]` - must be another project one you is dreaming about ;)

PR #848, opened 2021-03-10
Fix misaligned-grad

samyam:
When a parameter is not divisible by world size, the partitioned gradients are mis-aligned due to incorrect padding handling. This PR should fix for that.


ShadenSmith (commented):
Thanks @samyam! Can we add a unit test to cover this case in our CI?

samyam:
> Thanks @samyam! Can we add a unit test to cover this case in our CI?
       --> @ShadenSmith Done

ShadenSmith (approved):

jeffra:
these changes are included in #851

PR #855, opened 2021-03-12
fix pipeline engine allreduce bug supporting zero optimizer

gongjingcs:
Hello, I found that when the pipeline engine synchronizes the gradient, the communication function in the zero stage will not be called when the zero optimizer is setted.
I think this should be a bug, thanks


loadams:
Thanks @gongjingcs - we apologize for not getting to this sooner, but seems this has been fixed in the latest versions of DeepSpeed, so I'll close this PR.

PR #867, opened 2021-03-18
Samyamr/stage 3 skip modules without parameters

samyam:


deepspeed/runtime/zero/stage3.py
```diff
@@ -1066,8 +1066,9 @@ def _register_hooks_recursively(self, module, count=[0]):
         #print(f"{module.__class__} : {module.id}")
 
         for child in module.children():
-            count[0] = count[0] + 1
-            self._register_hooks_recursively(child, count=count)
+            if len(list(get_all_parameters(child, recurse=True))) > 0:
```
samyam, line 1069: TODO handle an unlikely corner case that a module or its children does not have any parameters but does have an external parameter in one of the children module

PR #870, opened 2021-03-18
[Fix] Ensure we remove existing hooks before assigning new hooks using ZeRO 3

SeanNaren:
We were seeing an issue in PL where we wrap the engine twice, once in training and once for testing. This ensures we remove the previously added hooks to the module after wrapping the model again!


SeanNaren:
The test passes without `remove_module_hooks`, which means something else in PL is causing this issue, not the requirement to re-wrap the module. Will be investigating further in the PL so the test case reflects the correct case

deepspeed/runtime/zero/stage3.py
```diff
@@ -2845,3 +2848,17 @@ def _handle_overflow(cpu_sum, x, i):
...
+
+def remove_module_hooks(module, hook_location='ds_hook_handles', recurse=True):
+    """
+    Function to remove all module forward/backward hooks. This is useful in cases
+    where the model may be wrapped more than once and the module hooks should
+    be removed first.
+    """
+    if hasattr(module, hook_location):
+        for k, v in getattr(module, hook_location).items():
+            v.remove()
```
samyam, line 2861: @SeanNaren We should also remove the hook handle from the dictionary after removing the hook itself no?

SeanNaren: Good catch! thanks @samyam :) I'll make the change. 

Haven't had time to dig into what this actually fixes on the Lightning side, but will get to it as soon as I can, not a high priority issue as the fix exists upstream in lightning for now as well!

rocm-mici:

SeanNaren:
I'm closing this PR as I could not reproduce this issue (i also think I saw a PR that addressed this already?)

PR #871, opened 2021-03-18
[ignore] Check if fp16_partitions called

sshleifer:


rocm-mici:

PR #873, opened 2021-03-18
zero-3 dynamic tracing support

jeffra:
Previously ZeRO-3's prefetcher was essentially disabled if the sub-module access pattern changed after the first iteration. This PR adds support for detecting and adapting our sub-module trace to more dynamic access patterns.

In initial testing with a modified version of Megatron that adds a new sub-module into forward after a few steps we're seeing ~10% boost in performance.


rocm-mici:

PR #881, opened 2021-03-19
Fix for fragmented linear inputs in ZeRO 3 Linear layers where reshap…

samyam:
…e is needed instead of view. Should not affect performance for cases which only requires view since reshape will just do view when possible


stas00:
This made a huuuuuuuuuuuuuuggggggggeeeeeeeeee difference! Thank you!

jeffra (approved):

Issue #872, opened 2021-03-18
[zero3] how to get the model reconstructed for saving?


stas00:
Under zero2 `self.model.state_dict()` returns a fp16 version of the model, under zero3 it returns some placeholder: with all the weights being just `tensor([1.],`, so how can we get the trained model out of deepspeed? 

This is related to https://github.com/microsoft/DeepSpeed/issues/800 but there under zero2 we at least had the fp16 version save-able, now there is no way at all. This is a total lock-in, unless I'm missing on some API that was added with zero3.

Ideally it'd be awesome if it were to reconstruct it directly on the disk, since that will ensure that there will be enough memory to do so.

To summarize different requests so far - users have 3 different needs:

1. being able to leave DeepSpeed after training with it. So they should be able to take a DeepSpeed checkpoint and recover the full consolidated fp32 model in a single file
2. same as (1) but fp16. Surely once can go from (1) to (2) easily, but I think it'd be much faster to get to (2) directly - I could be wrong. If I'm wrong than (1) is enough.
3. being able to call `deepspeed.consolidate_weights()` in the rank0 process which would give users full non-partitioned weights back (perhaps with a bool arg of whether they want the fp16 or fp32 version). So now they can just save the model as they do with any other pytorch tools. This would only be practical for small-ish models. The key here is that while this would be somewhat costly they will be able to use their code almost w/o any change if they train in various ways and not just with deepspeed. I think this must happen on cpu, since it's unlikely gpus will have the memory for that. It probably will have to return a copy of the model with the consolidated weights, so that the user can continue training the original model. So probably something along the lines of https://github.com/microsoft/DeepSpeed/pull/882 but in addition the partitioning will need to be removed too. The result of this call will provide users an equivalent of what they have under zero2 at this moment (if it's fp16).

I think all 3 would be more or less the same code, with just different ways of using it - (3) using the existing deepspeed engine and not needing to access the filesystem, (1) and (2) w/o needing an engine and relying exclusively on the filesystem.

Thank you!

@jeffra, @tjruwase

tjruwase:
@stas00, do you have repro steps?

stas00:
If you have a huggingface repo already from its root:
```
[...]

PR #882, opened 2021-03-19
Save ZeRO3 (partitioned) fp16 weights

tjruwase:
Save ZeRO3 (partitioned) fp16 weights. This is a first step to using ZeRO3 weights outside DeepSpeed, #872.

### DIFF
--- a/deepspeed/runtime/zero/stage3.py
+++ b/deepspeed/runtime/zero/stage3.py
@@ -2832,6 +2832,17 @@ class FP16_DeepSpeedZeroOptimizer_Stage3(object):
     def save_checkpoint_epilogue(self):
         self.persistent_parameters[0].all_gather(self.persistent_parameters)
 
+    def save_partitioned_weights(self, state_dict):
+        for name, param in self.module.named_parameters():
+            if name in state_dict.keys():
+                state_dict[name] = param.ds_tensor
+        return state_dict
+
+    def load_partitioned_weights(self, state_dict):
+        for name, param in self.module.named_parameters():
+            if name in state_dict.keys():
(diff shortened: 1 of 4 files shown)


stas00:
Perfect. Now I can save the fp16 weights! thank you!

deepspeed/runtime/zero/stage3.py
```diff
@@ -2832,6 +2832,17 @@ def save_checkpoint_prologue(self):
     def save_checkpoint_epilogue(self):
         self.persistent_parameters[0].all_gather(self.persistent_parameters)
 
+    def save_partitioned_weights(self, state_dict):
+        for name, param in self.module.named_parameters():
+            if name in state_dict.keys():
+                state_dict[name] = param.ds_tensor
```
stas00, line 2838: Found an issue here: `param.ds_tensor` is a flattened buffer. So `state_dicts` ends up having 1D vectors.

stas00: but we can't shape it back to the original since we only have a part of the tensor, so doing something like `narrow(0, 0, param.ds_numel).view(param.ds_shape)` from `_allgather_param()` won't work and the shape has no meaning here anyway.

So this line of logic is useful when it's used to load the `param.ds_tensor` directly by each gpu, as coded in the rest of this PR.

I just tried to use it to get the partitioned fp16 weights, but now I understand this is not possible using this approach.

Bottom line - there is no problem here, just needed to understand that this is not a real `state_dict` that is being saved but something like `flattened_params_state_dict`.

All is good!

tjruwase:
Redundant by #892 and #893

PR #884, opened 2021-03-23
[zero3] GatheredParameters can now handle a list of params

stas00:
With Samyam's guidance making `GatheredParameters` context manager support a list of params to handle whole layers easier.
e.g. when loading pre-trained weights. It's simpler to use for the user and it's faster than doing it a param at a time.

1.  switch to list of params, while still supporting a single param / update doc
2.  add a full example of how to do a Pretrained model loading
3.  fix the prose for modifier_rank

@samyam, @jeffra


deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -897,41 +898,75 @@ def forward(self, input):
...
-            register_external_parameter(self.fwd_module, self.param)
+            for p in self.params:
+                register_external_parameter(self.fwd_module, p)
 
     def __enter__(self):
         if not self.enabled:
             return
-        self.param.all_gather()
+        for p in self.params:
+            p.all_gather()
```
samyam, line 961: I know I said lets not worry about optimizing, but now its bugging me a bit can we do 

self.params[0].all_gather(param_list =self.params)

instead of the loop. This will fuse all the all_gather into a single one

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -897,41 +898,75 @@ def forward(self, input):
...
 
     def __exit__(self, *exc):
         if not self.enabled:
             return
-        if self.src_rank is not None:
-            torch.distributed.broadcast(self.param,
-                                        self.src_rank,
-                                        group=self.param.ds_process_group)
-        self.param.partition(has_been_updated=self.src_rank is not None)
+        for p in self.params:
```
samyam, line 966: We can optimize it here as well if we want by making the broadcast non-blocking when putting into the cuda queue:
handles = []
if self.src_rank is not None:
      for p in self.params:
           handles.append(torch.distributed.broadcast(p,self.src_rank, group=self[0].ds_process_group,async_op=True)
      for h in handles:
            h.wait()

samyam (approved):
Looks great! Thank you @stas00.

jeffra (approved):

PR #891, opened 2021-03-25
Avoid key error for skipped gradients

tjruwase:
Avoid key error for skipped gradients in dynamic networks


tjruwase:
Made redundant by #861

Issue #800, opened 2021-02-26
[request] new method to save the fp32 model


stas00:
Currently checkpoint saves the original fp32 weights in `zero_pp_rank_*` files, which makes it inaccessible to users who may need the final result to be used elsewhere w/o Deepspeed (e.g. inference or a different system).

So we need `deepspeed.save_model()` which will reconstruct the master weights and save them as `model.bin`.

For context for this need please see: https://github.com/microsoft/DeepSpeed/issues/797

Thank you.

stas00:
@ShadenSmith shared elsewhere that each fp32 param group is stored in a flattened buffer in the fp16 optimizer [here](https://github.com/microsoft/DeepSpeed/blob/7eb083c22112b67ee8ea43036479bb7680136349/deepspeed/runtime/fp16/fused_optimizer.py#L48) and needs to be exposed as a method.

stas00:
Any chances this new method to save the fp32 model can be added?

Users are reporting struggles when they customize their saving/loading of the checkpoints
https://github.com/huggingface/transformers/issues/10789

I think there is also a problem to switch from zero2 checkpoint to zero3, correct? Which means that we really need to salvage the fp32 model weights from the zero2 checkpoint before it can be loaded in zero3.

@ShadenSmith, @jeffra

stas00:
PR fix solution: https://github.com/microsoft/DeepSpeed/pull/892

PR #892, opened 2021-03-26
offline full fp32 weights reconstruction for zero 2+3 checkpoints

stas00:
This PR adds a magical script that extracts and reconstructs full fp32 weights from either zero 2 or 3 checkpoint. This consolidated state dict can then be used anywhere w/o DeepSpeed.

* I had to change the saving checkpoint code to also add the param names and shapes to the checkpoint optim files
* The script doesn't require live engine, everything is done in completely standalone logic - unfortunately I couldn't fully remove the dependency on `deepspeed` since the pickled data refers to `DeepSpeedEngine` and can't be unpickled w/o `import deepspeed`, but otherwise it's completely standalone.
* When a new checkpoint is saved this script gets copied automatically to the top checkpoint folder so that it's trivial for the user to get the weights out, but of course the API can be used directly - it just will be a really bad idea to do it live, since for large models it'd be very slow and memory consuming to mix it into the normal training. But if someone wants it - it's 
```
deepspeed.save_checkpoint(output_dir)
dist.barrier()
    convert_zero_chkpt_to_fp32_consolid_state_dict(f"{output_dir}/global_step1"), output_file):
```

I guess the only non-smooth part is figuring out that "global_stepXX" part - probably should expose some of that deepspeed API.

To run the script, from the top level checkpoint dir, is just:

```
$ cd /path/to/checkpoints_dir
$ ./zero_to_fp32.py global_step1 pytorch_model.bin
Processing zero checkpoint at global_step1
Detected checkpoint of type zero stage 3, world_size: 2
Saving fp32 state dict to pytorch_model.bin (total_numel=60506624)
```

Future improvements:
- currently the script uses 2x memory of the final checkpoint - eventually it'd be good to find a smart way to avoid that requirement
- currently the script relies on sufficient RAM to do the work - for huge models some kind of memory mapping will need to be used

TODO:

* [ ] Where would be a good place to document this? 

Lot's of thanks to @tjruwase for the guidance. Once I understood how zero handles its data, it was surprisingly easy to do the rest. Kudos for building an easy to handle system!

p.s. PR with fp16 live consolidator for zero3 addressing https://github.com/microsoft/DeepSpeed/issues/872 is coming soon 

@tjruwase 

Fixes: https://github.com/microsoft/DeepSpeed/issues/800

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -3,10 +3,13 @@ Copyright 2019 The Microsoft DeepSpeed Team
 '''
 
 import os
+import stat
 import torch
 import warnings
 import hashlib
 import torch.distributed as dist
+from collections import OrderedDict
+from shutil import copyfile
 
 from torch.nn.modules import Module
 from torch.distributed.distributed_c10d import _get_global_rank
@@ -1684,8 +1687,30 @@ class DeepSpeedEngine(Module):
         torch.save(state, save_path)
         self._curr_save_path = None
 
+    def _get_param_shapes(self):
+        param_shapes = OrderedDict()
+        for name, param in self.module.named_parameters():
+            param_shapes[name] = param.ds_shape if hasattr(param,
+                                                           "ds_shape") else param.shape
+            # print(f"saving param {name} {param_shapes[name]}")
+        return param_shapes
+
+    def _copy_recovery_script(self, save_path):
+        base_dir = os.path.dirname(os.path.dirname(__file__))
+        script = "zero_to_fp32.py"
+        src = os.path.join(base_dir, "utils", script)
+        dst = os.path.join(save_path, script)
+        logger.info(f"creating recovery script {dst}")
+        copyfile(src, dst)
+        # make executable
+        os.chmod(dst, os.stat(dst).st_mode | stat.S_IEXEC)
+
     def _save_zero_checkpoint(self, save_path, tag):
         zero_checkpoint_name = self._get_zero_ckpt_name(save_path, tag)
-        zero_sd = {'optimizer_state_dict': self.optimizer.state_dict()}
+        zero_sd = dict(
+            optimizer_state_dict=self.optimizer.state_dict(),
+            param_shapes=self._get_param_shapes(),
+        )
         torch.save(zero_sd, zero_checkpoint_name)
+        self._copy_recovery_script(save_path)
         logger.info('zero checkpoint saved {}'.format(zero_checkpoint_name))
--- a/deepspeed/utils/zero_to_fp32.py
+++ b/deepspeed/utils/zero_to_fp32.py
@@ -1,0 +1,151 @@
+#!/usr/bin/env python
+
+# This script extracts fp32 consolidated weights from a zero 2 and 3 DeepSpeed checkpoints. It gets
+# copied into the top level checkpoint dir, so the user can easily do the conversion at any point in
+# the future. Once extracted, the weights don't require DeepSpeed and can be used in any
+# application.
+#
+# example: python zero_to_fp32.py global_step1 pytorch_model.bin
+
+import argparse
+import torch
+import glob
+import os
+from collections import OrderedDict
+import deepspeed
+
+# while this script doesn't use deepspeed to recover data, since the checkpoints are pickled with
+# DeepSpeed data structures it has to be available in the current python environment.
+
+
+def get_optim_files(checkpoint_dir):
+
+    if not os.path.isdir(checkpoint_dir):
+        raise FileNotFoundError(f"Directory '{checkpoint_dir}' doesn't exist")
+
+    # XXX: need to test that this simple glob rule works for multi-node setup too
+    optim_files = sorted(glob.glob(f"{checkpoint_dir}/*_optim_states.pt"))
+
+    if len(optim_files) == 0:
+        raise FileNotFoundError(
+            f"can't find '*_optim_states.pt' files in directory '{checkpoint_dir}'")
+
+    return optim_files
+
+
+def parse_optim_states(files):
+    state_dicts = []
+    for f in files:
+        state_dicts.append(torch.load(f))
+
+    if not "zero_stage" in state_dicts[0]['optimizer_state_dict']:
+        raise ValueError(f"non zero checkpoint")
+    zero_stage = state_dicts[0]['optimizer_state_dict']["zero_stage"]
+
+    # the groups are named differently in each stage
+    if zero_stage == 2:
+        fp32_groups_key = "single_partition_of_fp32_groups"
+    elif zero_stage == 3:
+        fp32_groups_key = "fp32_flat_groups"
+    else:
+        raise ValueError(f"unknown zero stage {zero_stage}")
+
+    param_shapes = state_dicts[0]["param_shapes"]
+    fp32_flat_groups = [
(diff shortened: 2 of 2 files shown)


tjruwase (approved):

humza909:
@stas00 I think this offline reconstruction is only casting fp16 weights to fp32? 

Then how we can use stage3 with fp32 weights. I could not see any method to save 32bit partitioned weights. 
This means if a model was originally trained with fp32 converting it to fp16 will cause a precision loss, and if someone wants to avoid this loss there is no way to train and save with stage3. 

Please share your thoughts. 

Thanks

stas00:
> @stas00 I think this offline reconstruction is only casting fp16 weights to fp32?

no, it's extracting fp32 weight shards and reconstructs full fp32 weights from those. fp16 is not involved anywhere in this process.

humza909:
@stas00 Thanks for your response. 

Two/Three questions regarding stage3:

1. If a model is trained in fp16 setting and consolidated by `gather_16bit_weights_on_model_save` then how the script `zero_to_fp32.py` will extract fp32 weights? 
2. If I am training with only fp32 enabled then what method will consolidate weights? Do we still use `gather_16bit_weights_on_model_save`? There is no other config option available.  
3. Why not save the weights directly to the enabled fp precision in config either fp16 or fp32 without going for additional steps. I know going from fp16 to fp32 will take additional RAM, but if fp32 is enabled save the model in fp32. 
4. Below the model parameters are saved in optim_states.pt? 
```

ls -lh global_step100/
total 15G
-rw-rw-r-- 1 snxds snxds  97M Apr  4 08:07 zero_pp_rank_0_mp_rank_00_model_states.pt
-rw-rw-r-- 1 snxds snxds 7.4G Apr  4 08:07 zero_pp_rank_0_mp_rank_00_optim_states.pt
-rw-rw-r-- 1 snxds snxds  97M Apr  4 08:07 zero_pp_rank_1_mp_rank_00_model_states.pt
-rw-rw-r-- 1 snxds snxds 7.4G Apr  4 08:07 zero_pp_rank_1_mp_rank_00_optim_states.pt
```

Thanks

stas00:
> @stas00 Thanks for your response.
> 
> Two/Three questions regarding stage3:
> 
>     1. If a model is trained in fp16 setting and consolidated by `gather_16bit_weights_on_model_save` then how the script `zero_to_fp32.py` will extract fp32 weights?

If you did `save_checkpoint` it'll extract fp32 weights. 

`gather_16bit_weights_on_model_save` has nothing to do with the normal Deepspeed save/load sharded weight cycles.

>     2. If I am training with only fp32 enabled then what method will consolidate weights? Do we still use `gather_16bit_weights_on_model_save`? There is no other config option available. If it is still saving in 16 bit then where are the remaining 16 bit saved? How bits are split? Why the model was originally converted to 16-bit and used for forward and backward pass?

the fp32 weights are saved as parts of `save_checkpoint` and `zero_to_fp32.py` extracts those.

you're asking a good question about `gather_16bit_weights_on_model_save` - I asked for it specifically to deal with half-precision training. You simply don't call this method if you're not doing half-precision. Perhaps if there is a need you can ask for `gather_32bit_weights_on_model_save`, if you need it.

>     3. Why not save the weights directly to the enabled fp precision in config either fp16 or fp32 without going for additional steps. I know going from fp16 to fp32 will take additional RAM, but if fp32 is enabled save the model in fp32.

Because it's super-expensive both cpu RAM and time-wise. When you do serious training you must save checkpoints frequently due to hardware crashes, and if this process isn't fast you're wasting gpu idling time. So the helper method was added to aid the beginner user to get things working. Once you have everything else working you don't want to use gathering ever again for any largish model size serious training.

>     4. Below the model parameters are saved in optim_states.pt?


what is question 4?

humza909:
@stas00 Thanks for your response.

Ignore question 04. Likely last series of questions.

1. I am still confused why everything in stage3 is in 16-bit? Even I am using full-precision/fp32. 

2. Your answer to my first question is `save_checkpoint` will extract fp32 weights, but how? My model was converted to fp16 before the start of training, and fp32 weights were discarded at the beginning. Is it only the type casting now? 

Any recommended links where I can get the implementation details would be very helpful. 

Thanks for you time :)

stas00:
> @stas00 Thanks for your response.
> 
> Ignore question 04. Likely last series of questions.
> 
>     1. I am still confused why everything in stage3 is in 16-bit? Even I am using full-precision/fp32.

It's not. If you're not using half-precision. Nothing is in fp16. Everything is in fp32.

You should never convert your model to any format when you use deepspeed. It'll do it for you on the need basis.

>     2. Your answer to my first question is `save_checkpoint` will extract fp32 weights, but how? My model was converted to fp16 before the start of training, and fp32 weights were discarded at the beginning. Is it only the type casting now?

See the previous answer. It'll convert it back to fp32 if you pre-converted to fp16.

To give you a hint deepspeed performs the same as mixed precision protocol, but in reverse - it converts the model to fp16 and it keeps a copy of fp32 weights in its optimizer. So you always have both weights. `save_checkpoint` saves only fp32 weights.

So if you're not doing mixed precision, you only have fp32 weights and they are saved with `save_checkpoint`.

> Any recommended links where I can get the implementation details would be very helpful.

I think for this particular topic there isn't much documentation, you can read the papers here:
https://huggingface.co/docs/transformers/main/main_classes/deepspeed#main-deepspeed-resources

You can make a request to document these nuances.

khulaifi95:
@stas00 Sorry but I find this issue related to my problem here. To make an extension on the last question from @humza909, for zero stage-3 training across 2 nodes, is it ever possible to reconstrcut the model weights from the contents that saved in checkpoints from one node?

zcakzhuu:
> > 1. I am still confused why everything in stage3 is in 16-bit? Even I am using full-precision/fp32.
> 
> It's not. If you're not using half-precision. Nothing is in fp16. Everything is in fp32.
> 
> You should never convert your model to any format when you use deepspeed. It'll do it for you on the need basis.
> 
> > 2. Your answer to my first question is `save_checkpoint` will extract fp32 weights, but how? My model was converted to fp16 before the start of training, and fp32 weights were discarded at the beginning. Is it only the type casting now?
> 
> See the previous answer. When not using half precision it'll upcast it into fp32 if you pre-converted to fp16.
> 
> To give you a hint deepspeed performs the same as mixed precision protocol, but in reverse - it converts the model to fp16 and it keeps a copy of fp32 weights in its optimizer. So you always have both weights. `save_checkpoint` saves only fp32 weights.
> 
> So if you're not doing mixed precision, you only have fp32 weights and they are saved with `save_checkpoint`.
> 
> > Any recommended links where I can get the implementation details would be very helpful.
> 
> I think for this particular topic there isn't much documentation, you can read the papers here: https://huggingface.co/docs/transformers/main/main_classes/deepspeed#main-deepspeed-resources the rest of the doc might be useful as well as a general reading, but it's specific to the `transformers` integration.
> 
> You can make a request to document these nuances. These are all excellent questions and are important to understand.

Hi, thanks for the great answers! They are really helpful. I have a question about fine-tuning in fp32 using peft. Regarding question 2, if I am checkpointing a model that is fine-tuned using peft in fp32 with a transformers trainer, do I have to run zero_to_fp32.py before resuming training from checkpoint using the trainer? Aren't the weights in .bin file fp32 already?

stas00:
@khulaifi95, I didn't understand your question, could you please try to reframe it?

Issue #879, opened 2021-03-19
memory leak in an environment like notebook


stas00:
In HF/DS tests I use a notebook-like environment for some of the tests, which means I don't fork a new process for each deepspeed run, but emulate the distributed env in the process and repeatedly run deepspeed. This makes it much easier/faster to test the values of the weights.

There must be either some global variable that holds onto memory, or the circular reference happening, since memory doesn't get released after deepspeed has done its work and I even explicitly deleted the engine/scheduler/optimizer variables.

I tested that if I remove deepspeed from the equation there is no leak.

Now I have a few dozens of those tests and I get some 10GB extra RAM used per deepspeed invocation. So it quickly grows to 100GBs.
 
This is with a test model of 2x 1 value weights y = wx+b - i.e. the model's memory foot print is ~0. 

Any idea where the objects might be held and not destroyed? If I repeat the same re-creation of the trainer in a loop the old one should get its memory freed up.

So this is an example of a cell in a jupyter notebook:
```
if trainer.deepspeed:
    print("reloading")
    trainer.deepspeed = None
    trainer.optimizer = None
    trainer.lr_scheduler = None
trainer = get_regression_trainer(output_dir=output_dir, deepspeed=ds_config_dict, skip_memory_metrics=True)
trainer.train()
```
of course the explicit `None`s should be needed as they should get overwritten by the new trainer, I was just doing a sanity check.

So there must be a circular reference where 2 or more internal variables refer to each other and thus the memory doesn't get freed.

I emulate a dist env with just:
```
dist_env_1_gpu = dict(
            MASTER_ADDR="localhost", MASTER_PORT="10999", RANK="0", LOCAL_RANK="0", WORLD_SIZE="1"
        )
for k,v in dist_env_1_gpu.items():
    os.environ[k]=v
```

I'd be happy to try to investigate this on my own if you could help me with pointers at potential suspects.

Thank you!

@jeffra, @samyam

PR #896, opened 2021-03-27
fix cpu_adam memory leak on deepspeed re-use in the same process

stas00:
As reported in https://github.com/microsoft/DeepSpeed/issues/879 currently deepspeed leaks some 8GB general RAM when re-running `deepspeed.initialize` in the same process (notebook or pytest worker). There is no problem with one  `deepspeed.initialize`  per process, since the memory is reclaimed on process exit.

With the help from @tjruwase we narrowed it down to cpu_adam creating objects, but not freeing them. It was happening only with zero2 or 3 with `cpu_offload` enabled.

This PR:
* extends the `cuda_adam.cpp` extension to add `destroy_adam_optimizer`
* adds `DeepSpeedCPUAdam.__del__` that calls the new cpp method

I haven't checked if other deepspeed optimizers have the same issue.

There is another small leak  of ~50MB general RAM remaining somewhere still, when re-running `deepspeed.initialize` in the same process, but it's far less of a problem than the current 8GB.

Fixes: https://github.com/microsoft/DeepSpeed/issues/879

@RezaYazdaniAminabadi


csrc/adam/cpu_adam.cpp
```diff
@@ -672,11 +672,21 @@ int ds_adam_step_plus_copy(int optimizer_id,
     return 0;
 }
 
+int destroy_adam_optimizer(int optimizer_id)
+{
+    // std::cout << "Adam Optimizer #" << optimizer_id
+    //          << " is destroyed." << std::endl;
+    s_optimizers.erase(optimizer_id);
```
RezaYazdaniAminabadi, line 679: I think we need to destroy all optimizers here, when the __del__ function is called on the Python side, right? Then, I would use the .clear() rather than erase one element.

csrc/adam/cpu_adam.cpp
```diff
@@ -672,11 +672,21 @@ int ds_adam_step_plus_copy(int optimizer_id,
     return 0;
 }
 
+int destroy_adam_optimizer(int optimizer_id)
+{
+    // std::cout << "Adam Optimizer #" << optimizer_id
+    //          << " is destroyed." << std::endl;
```
RezaYazdaniAminabadi, line 678: Do we want to give this warning? or just simply remove it

RezaYazdaniAminabadi (approved):

RezaYazdaniAminabadi (approved):

deepspeed/ops/adam/cpu_adam.py
```diff
@@ -96,7 +96,7 @@ def __setstate__(self, state):
             group.setdefault('amsgrad', False)
 
     @torch.no_grad()
-    def step(self, closure=None, fp16_param_groups=None):
+    def step(self, optId=0, closure=None, fp16_param_groups=None):
```
stas00, line 99: this is python land, so camel case is for typically reserved for class names only.

csrc/adam/cpu_adam.cpp
stas00: `DeepSpeedCPUAdam.__init__` creates only one, so its corresponding ``DeepSpeedCPUAdam.__del__` should free only that one. 

It surely could delete the whole thing if it's empty to be super-clean.

Perhaps you don't need an array of optimizers in the first place if it's always one record anyway.

I just don't know the intricacies of why it was implemented this way in first place (i.e. to support multiple optimizers), so it's hard for me to know the right approach here.

csrc/adam/cpu_adam.cpp
stas00: We don't really need it, other than for debug, I thought to leave it while we are sorting out if I proposed the right approach and will delete it just before this is approved and then clean it up.

deepspeed/ops/adam/cpu_adam.py
stas00: so, again, if this case is reduced to only ever having id 0, this is now becoming a bit of a mess. Why do we then create `self.opt_id`, but not use it? i.e. you removed its usage in most places. 

Where does `step()` gets the id now from? It looks like you're relying on the default `0`.

Clearly you guys need to have a bit of a design session around this.

If I may suggestion, let's revert the change you added, so that this PR is just adding a proper cleanup. And then if you decided to kill the counter, this is a different change I'd say.

csrc/adam/cpu_adam.cpp
stas00: I removed it.

csrc/adam/cpu_adam.cpp
RezaYazdaniAminabadi: I think the way we have it currently sounds fine! thanks Stas :)

csrc/adam/cpu_adam.cpp
RezaYazdaniAminabadi: Thanks

deepspeed/ops/adam/cpu_adam.py
RezaYazdaniAminabadi: I agree. I revert back my changes

jeffra (approved):

Issue #877, opened 2021-03-19
[zero3] apex was installed without --cpp_ext.  Falling back to Python flatten and unflatten.


stas00:
Running on one setup the current master I get:

```
[2021-03-18 22:12:42,472] [WARNING] [stage3.py:34:<module>] apex was installed without --cpp_ext.  
Falling back to Python flatten and unflatten.
```

Is it using a previously installed apex, or an apex comes with DeepSpeed and what needs to be done to make it fast?

I don't see apex already installed on that setup. And I'm not using it in the ds config file.

Thank you.

stas00:
It's because `try` doesn't check if apex is even installed before checking for `apex_C`. So it needs to be fixed.

However we are discussing removing this code altogether 
https://github.com/microsoft/DeepSpeed/issues/877 which would close this issue.

PR #910, opened 2021-04-01
[zero] faster flatten/unflatten (cpp version) 

stas00:
(this PR has evolved and has been edited from its original version which proposed using apex's functions)

This PR is trying to switch to the fastest version of `flatten`/`unflatten` and consistently use it through all zero stages.

as discovered in later comments of this PR https://github.com/microsoft/DeepSpeed/pull/910#issuecomment-812150852 `UtilsBuilder().load().flatten` is on par with  `apex_C.flatten` speed-wise and 2-3x faster than  `torch._utils._flatten_dense_tensors`
(about the same for `unflatten`)

So this PR was modified to switch to `UtilsBuilder().load().flatten` and `UtilsBuilder().load().unflatten` 

Let's try first if it may work loading it at `import`-level since it will trigger build the first time for JIT. If it doesn't work we can fold it into each class's `__init__` like `zero2.py` had it prior to this PR.

Fixes: https://github.com/microsoft/DeepSpeed/issues/877

@jeffra


jeffra:
Good point @stas00, curious on @samyam's thoughts here too but we do have apex's (un)flatten ops ported into our JIT-compatible ops. We use them in zero-2, take a look here:

https://github.com/microsoft/DeepSpeed/blob/7fcc8911cbced68a045063a4edc0507688fc78e6/deepspeed/runtime/zero/stage2.py#L124-L127

We might be able to use these in z3 instead so we don't have to depend on apex.

jeffra:
We should be able to use it as a drop in replacement I suspect. The (un)flatten code in our op is taken directly from apex, which was taken from pytorch haha take a look, it's not doing anything fancy:

https://github.com/microsoft/DeepSpeed/blob/master/csrc/utils/flatten_unflatten.cpp

stas00:
This is totally weird then because they don't use the c++ implementation in the core: 
https://github.com/pytorch/pytorch/blob/33b95c6bac5cf9c52fa646bf7335664b0556263d/torch/_utils.py#L248

stas00:
Also there is a related issue: https://github.com/microsoft/DeepSpeed/issues/877 if this apex code is removed it'll close this issue as well.

jeffra:
> This is totally weird then because they don't use the c++ implementation in the core:
> https://github.com/pytorch/pytorch/blob/33b95c6bac5cf9c52fa646bf7335664b0556263d/torch/_utils.py#L248

haha yep, I've also dug into this and don't understand why torch doesn't just use these ops directly. Seems odd...

I believe @RezaYazdaniAminabadi did some small benchmarking at one point between the torch (un)flatten functions vs the cpp op and the cpp op performed significantly better. But I think we sometimes forget to use it in practice.

stas00:
It looks like this is an identical code, c++

https://github.com/pytorch/pytorch/blob/33b95c6bac5cf9c52fa646bf7335664b0556263d/torch/_utils.py#L248

```
inline at::Tensor flatten_dense_tensors(at::TensorList tensors) {
  static auto flatten = [](const at::Tensor &t) { return t.contiguous().view({-1}); };
  if (tensors.size() == 1)
    return flatten(tensors[0]);
  return at::cat(fmap(tensors, flatten));
}
```
python: 

https://github.com/pytorch/pytorch/blob/547346d66350b4ac325941752f53fa92f756f6a9/torch/csrc/utils/tensor_flatten.h#L10

```
def _flatten_dense_tensors(tensors):
    if len(tensors) == 1:
        return tensors[0].contiguous().view(-1)
    flat = torch.cat([t.contiguous().view(-1) for t in tensors], dim=0)
    return flat
```

stas00:
> I believe @RezaYazdaniAminabadi did some small benchmarking at one point between the torch (un)flatten functions vs the cpp op and the cpp op performed significantly better. But I think we sometimes forget to use it in practice.

@RezaYazdaniAminabadi, do you by chance have the benchmark still? If it's faster we should notify the pytorch devs to rectify this omission - perhaps this function pair is not widely used.

stas00:
I did some benchmarking and the results are very strange. If I use cProfile or `line_profiler` I get the cpp deepspeed version of flatten is 32x slower than pytorch python version! if I use `timeit`, the speed is the same.

Setup:
```
pip install line_profiler
```
Here is the little benchmark I wrote:
```
#!/usr/bin/env python

import argparse

import torch
from torch._utils import _flatten_dense_tensors, _unflatten_dense_tensors
from deepspeed.ops.op_builder import UtilsBuilder

util_ops = UtilsBuilder().load()
flatten = util_ops.flatten
unflatten = util_ops.unflatten

torch.manual_seed(0)
# emulate a small typical model weights
x = [torch.rand((512,512)).cuda(), torch.rand((512,1024)).cuda(), torch.rand((512,30000)).cuda()]
t = x * 30

# warm up and check that the same output is produced
flat_py = _flatten_dense_tensors(t)
flat_cpp = flatten(t)
numel = flat_cpp.numel()
assert torch.eq(flat_py, flat_cpp).all(), "both produce the same tensor"

# the programs being tested
def py():
    for i in range(1000):
        flat = _flatten_dense_tensors(t)

def cpp():
    for i in range(1000):
        flat = flatten(t)

#### cProfile ####

import cProfile

def cprofileme():
    print("--------------- cProfile -----------------")
    cProfile.run("py()", sort=-1)
    cProfile.run("cpp()", sort=-1)

#### timeit ####

import timeit
import time

def timeme():
    print("--------------- timeit -----------------")
    print(f'py ={timeit.Timer("py()", globals=globals()).timeit(number=1)}')
    print(f'cpp={timeit.Timer("cpp()", globals=globals()).timeit(number=1)}')

#### line_profiler ####

def line_profileme():
    print("--------------- line_profier -----------------")
    profile(py)()
    profile(cpp)()

if __name__ == "__main__":
    parser.add_argument("-l", action='store_true')
    args = parser.parse_args()
    if args.l:
        line_profileme()
    else:
        cprofileme()
        timeme()

```

Output:
```
./flatten_bench.py       
Using /home/stas/.cache/torch_extensions as PyTorch extensions root...
Emitting ninja build file /home/stas/.cache/torch_extensions/utils/build.ninja...
Building extension module utils...
Allowing ninja to set a default number of workers... (overridable by setting the environment variable MAX_JOBS=N)
ninja: no work to do.
Loading extension module utils...
Time to load utils op: 0.06846189498901367 seconds
--------------- cProfile -----------------
         184004 function calls in 0.130 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.130    0.130 <string>:1(<module>)
     1000    0.010    0.000    0.129    0.000 _utils.py:248(_flatten_dense_tensors)
     1000    0.017    0.000    0.095    0.000 _utils.py:264(<listcomp>)
        1    0.001    0.001    0.130    0.130 flatten_bench.py:25(py)
        1    0.000    0.000    0.130    0.130 {built-in method builtins.exec}
     1000    0.000    0.000    0.000    0.000 {built-in method builtins.len}
     1000    0.024    0.000    0.024    0.000 {built-in method cat}
    90000    0.013    0.000    0.013    0.000 {method 'contiguous' of 'torch._C._TensorBase' objects}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
    90000    0.065    0.000    0.065    0.000 {method 'view' of 'torch._C._TensorBase' objects}


         1004 function calls in 4.413 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    4.413    4.413 <string>:1(<module>)
        1    0.003    0.003    4.413    4.413 flatten_bench.py:29(cpp)
        1    0.000    0.000    4.413    4.413 {built-in method builtins.exec}
     1000    4.410    0.004    4.410    0.004 {built-in method utils.flatten}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


--------------- timeit -----------------
py =4.643545062281191
cpp=4.6434575244784355

```
part 2:
```
kernprof -l flatten_bench.py -l; python -m line_profiler  flatten_bench.py.lprof
Using /home/stas/.cache/torch_extensions as PyTorch extensions root...
Emitting ninja build file /home/stas/.cache/torch_extensions/utils/build.ninja...
Building extension module utils...
Allowing ninja to set a default number of workers... (overridable by setting the environment variable MAX_JOBS=N)
ninja: no work to do.
Loading extension module utils...
Time to load utils op: 0.06914520263671875 seconds
--------------- line_profier -----------------
Wrote profile results to flatten_bench.py.lprof
Timer unit: 1e-06 s

Total time: 0.122491 s
File: flatten_bench.py
Function: py at line 25

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    25                                           def py():
    26      1001        522.0      0.5      0.4      for i in range(1000):
    27      1000     121969.0    122.0     99.6          flat = _flatten_dense_tensors(t)

Total time: 4.41902 s
File: flatten_bench.py
Function: cpp at line 29

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    29                                           def cpp():
    30      1001       1072.0      1.1      0.0      for i in range(1000):
    31      1000    4417952.0   4418.0    100.0          flat = flatten(t)
```

Is my benchmark somehow wrong? I don't understand that huge discrepancy between timeit and profilers

stas00:
for `unflatten` the 3 ways are much more closely.

```
py =0.12716494593769312
cpp=0.07942950166761875
apex=0.0548561904579401
```
The speed is:

1. `apex_C.flatten`
2. `UtilsBuilder().load().flatten`
3. `torch._utils import _flatten_dense_tensors`

the modified benchmark is:

```
!/usr/bin/env python

import argparse

import torch
from torch._utils import _flatten_dense_tensors, _unflatten_dense_tensors
from deepspeed.ops.op_builder import UtilsBuilder

from apex_C import flatten as flatten_apex
from apex_C import unflatten as unflatten_apex

util_ops = UtilsBuilder().load()
flatten = util_ops.flatten
unflatten = util_ops.unflatten

torch.manual_seed(0)
# emulate a small typical model weights
x = [torch.rand((512,512)).cuda(), torch.rand((512,1024)).cuda(), torch.rand((512,30000)).cuda()]
unflat_t = x * 20

# warm up and check that the same output is produced
flat_py = _flatten_dense_tensors(unflat_t)
flat_cpp = flatten(unflat_t)
flat_apex = flatten_apex(unflat_t)
#numel = flat_cpp.numel()
assert torch.eq(flat_py, flat_cpp).all(), "both produce the same tensor"
assert torch.eq(flat_py, flat_apex).all(), "both produce the same tensor"

flat_t = flat_py
unflat_py = _unflatten_dense_tensors(flat_py, unflat_t)
for i in range(len(unflat_t)): assert torch.eq(unflat_t[i], unflat_py[i]).all()
unflat_cpp = _unflatten_dense_tensors(flat_cpp, unflat_t)
for i in range(len(unflat_t)): assert torch.eq(unflat_t[i], unflat_cpp[i]).all()
unflat_apex = _unflatten_dense_tensors(flat_apex, unflat_t)
for i in range(len(unflat_t)): assert torch.eq(unflat_t[i], unflat_apex[i]).all()

# the programs being tested
def py():
    for i in range(1000):
        unflat = _unflatten_dense_tensors(flat_t, unflat_t)

def cpp():
    for i in range(1000):
        unflat = unflatten(flat_t, unflat_t)

def apex():
    for i in range(1000):
        unflat = unflatten_apex(flat_t, unflat_t)

#### cProfile ####

import cProfile

def cprofileme():
    print("--------------- cProfile -----------------")
    cProfile.run("py()", sort=-1)
    cProfile.run("cpp()", sort=-1)
    cProfile.run("apex()", sort=-1)

#### timeit ####

import timeit

def timeme():
    print("--------------- timeit -----------------")
    print(f'py ={timeit.Timer("py()", globals=globals()).timeit(number=1)}')
    print(f'cpp={timeit.Timer("cpp()", globals=globals()).timeit(number=1)}')
    print(f'apex={timeit.Timer("apex()", globals=globals()).timeit(number=1)}')

#### line_profiler ####
# this one requires a special way to be called
# pip install line_profiler
# kernprof -l flatten_bench.py -l; python -m line_profiler  flatten_bench.py.lprof

def line_profileme():
    print("--------------- line_profier -----------------")
    profile(py)()
    profile(cpp)()
    profile(apex)()

if __name__ == "__main__":
    parser.add_argument("-l", action='store_true')
    parser.add_argument("-c", action='store_true')
    parser.add_argument("-t", action='store_true')
    args = parser.parse_args()
    if args.l:
        line_profileme()
    elif args.c:
        cprofileme()
    elif args.t:
        timeme()
```

stas00:
Proposed to pytorch to switch to the cpp version https://github.com/pytorch/pytorch/issues/55240 so down the road we won't need the workaround.

stas00:
It looks good now. As I had to turn some functions into methods to make this work, I'm not sure if I placed them in the most intuitive place between other existing methods - please feel free to move them or tell me where to move them. Either way works.

tjruwase (approved):

jeffra (approved):

stas00:
FYI, pytorch has now replaced the slower python version with the cpp version https://github.com/pytorch/pytorch/pull/58006

PR #941, opened 2021-04-10
Delete check of pdsh

tma15:
Checking whether or not `pdsh` exists seems to be redundant because it is checked by `runner.backend_exists()` when `args.launcher` is `pdsh`. Furthermore, this check is always done regardless of `args.launcher`. If I understand correctly, it is not necessary for `openmpi` and `mvapich`.

Sorry if I misunderstand. Thanks.


jeffra (approved):
Good catch, thank you for your contribution!

PR #945, opened 2021-04-12
Add find_unused_parameters option to DeepSpeedEngine

ghosthamlet:
As unused parameters in modules may not be expected sometimes, 
add an explicit error msg when it occurred and an option to avoid the error: microsoft#707


ghosthamlet:
I can't find where was the yapf error in my commit code, also no error when run yapf local, it is strange.

ghosthamlet:
Ok, Fixed it.

tjruwase:
@ghosthamlet, thanks for this feature. Rather than modifying DeepSpeed engine, can you please add this configuration to the [zero options](https://www.deepspeed.ai/docs/config-json/#zero-optimizations-for-fp16-training)? That is a more appropriate place since the issue only affects zero stage 2. Thanks.

ghosthamlet:
@tjruwase Thanks for the review, i changed it, please review again.

deepspeed/runtime/zero/constants.py
```diff
@@ -29,7 +29,8 @@
     "cpu_offload_use_pin_memory": [true|false] (deprecated),
     "sub_group_size" : 1000000000000,
     "offload_param": {...},
-    "offload_optimizer": {...}
+    "offload_optimizer": {...},
+    "stage2_find_unused_parameters": [true|false]
```
tjruwase, line 33: Can you please just name this "find_unused_parameters" so it can be easily extended to other stages in future.

tjruwase:
Thanks for making the changes. One final request can you also add a unit test?

deepspeed/runtime/zero/constants.py
ghosthamlet: No problem, now it changed.

ghosthamlet:
> Thanks for making the changes. One final request can you also add a unit test?

Ok, let me find out how the test works and add it.

ghosthamlet:
@tjruwase Test added, please review again.

tjruwase (approved):

Issue #937, opened 2021-04-08
[tests] integrating transformers deepspeed tests into deepspeed tests


stas00:
@jeffra, here is how you can integrate our tests:

```
git clone https://github.com/huggingface/transformers
cd transformers
pip install -e .[testing]
RUN_SLOW=1 pytest tests/deepspeed
```

awan-10:
This is resolved in the linked PR already.

PR #958, opened 2021-04-14
[CI] Add HF transformers tests

awan-10:
Adding the tests from here: https://github.com/microsoft/DeepSpeed/issues/937


.github/workflows/main.yml
```diff
@@ -49,3 +49,11 @@ jobs:
         run: |
           if [[ -d ./torch-extensions ]]; then rm -rf ./torch-extensions; fi
           TORCH_EXTENSIONS_DIR=./torch-extensions pytest --durations=0 --forked --verbose tests/unit/
+
+      - name: Transformer tests
```
stas00, line 53:
```suggestion
      - name: HF transformers tests
```

.github/workflows/main.yml
```diff
@@ -49,3 +49,11 @@ jobs:
...
+        run: |
+          if [[ -d ./torch-extensions ]]; then rm -rf ./torch-extensions; fi
+          git clone https://github.com/huggingface/transformers
+          cd transformers
+          pip install -e .[testing]
```
stas00, line 58:
```suggestion
          pip install -e .
          pip install sacrebleu rouge-score nltk parameterized psutil datasets
```

As `faiss-cpu` won't build, let's hardcode just the packages we want for now.

.github/workflows/main.yml
```diff
@@ -63,13 +63,34 @@ jobs:
...
+          ds_report
       - name: HF transformers tests
         run: |
           if [[ -d ./torch-extensions ]]; then rm -rf ./torch-extensions; fi
           git clone https://github.com/huggingface/transformers
+          git rev-parse --short HEAD
           cd transformers
-          pip install -e .
+          pip install .
           pip install sacrebleu rouge-score nltk parameterized psutil datasets
```
stas00, line 95: Let's see if this work, but alternatively we could do:
```
find examples/pytorch -name requirements.txt -exec pip install -r {} \;
```
which will be automatically finding all the dependencies. As each example has its own `requirements.txt` now.

jeffra (approved):

PR #970, opened 2021-04-17
1-bit LAMB optimizer

conglongli:
1-bit LAMB: Communication Efficient Large-Scale Large-Batch Training with LAMB's Convergence Speed.

Author: @conglongli, @awan-10, @samyam, Hanlin Tang, Yuxiong He
Paper: https://arxiv.org/abs/2104.06069


docs/_pages/features.md
```diff
@@ -172,15 +172,17 @@ Please see the [core API doc](https://deepspeed.readthedocs.io/) for more detail
 
 ## Training Optimizers
 
-### 1-bit Adam optimizer with up to 5x less communication
+### 1-bit Adam and 1-bit LAMB optimizers with up to 5x less communication
 
-DeepSpeed has an efficient implementation of a novel algorithm called 1-bit Adam.
-It offers the same convergence as Adam, incurs up to 5x less communication that enables
+DeepSpeed has two communication-efficient optimizers called 1-bit Adam and 1-bit LAMB.
+They offer the same convergence as Adam/LAMB, incurs up to 5x less communication that enables
```
awan-10, line 178: minor typo: incurs--> incur

awan-10 (approved):

awan-10:
Thank you @conglongli for this PR. It looks good to me. I just added a minor comment and approved all the other changes as is.

docs/_pages/features.md
conglongli: Fixed. Thanks Ammar.

docs/_pages/config-json.md
```diff
@@ -88,6 +88,40 @@ The 1-bit Adam optimizer supports the following three params keys/values in addi
...
+
+```json
+"optimizer": {
+    "type": "OneBitLamb",
+    "params": {
+      "lr": 11e-3,
+      "weight_decay": 0.01,
+      "bias_correction": false,
+      "max_coeff": 0.3,
+      "min_coeff": 0.01,
```
jeffra, line 101: should we add descriptions for max/min coeff in the table below?

jeffra (approved):
Can we use `self.deepspeed.enable_backward_allreduce` instead of creating `self.deepspeed.pipeline_enable_backward_allreduce`?

Otherwise looks good to me!

deepspeed/runtime/utils.py
```diff
@@ -125,9 +130,17 @@ def has_overflow(self, params):
                                          op=torch.distributed.ReduceOp.MAX,
                                          group=torch.distributed.group.WORLD)
         elif self.mpu is not None:
+            if self.deepspeed.pipeline_enable_backward_allreduce is False:
```
jeffra, line 133: This will crash if `self.deepspeed = None` right?

conglongli: Yes you are right. I added checks about whether self.deepspeed exist, and whether it is PipelineEngine or not.

conglongli:
> Can we use `self.deepspeed.enable_backward_allreduce` instead of creating `self.deepspeed.pipeline_enable_backward_allreduce`? /cc @ShadenSmith , @sdtblck
> 
> Otherwise looks good to me!

@jeffra @ShadenSmith We do need to add self.deepspeed.pipeline_enable_backward_allreduce in order to prevent PipelineEngine to do allreduce, because 1-bit Adam/LAMB it will handle the allreduce communication during the compression stage. On the other hand when using PipelineEngine we need to keep self.deepspeed.enable_backward_allreduce to always false since at 1-bit Adam/LAMB warmup stage we still need PipelineEngine to handle the allreduce. I did some further fix in 1-bit Adam/LAMB to make sure that when we use PipelineEngine, it will not touch self.deepspeed.enable_backward_allreduce, so that 1-bit Adam/LAMB shall not affect normal use of pipeline parallelism during warmup stage.

jeffra (approved):

docs/_pages/config-json.md
conglongli: Just added.

PR #971, opened 2021-04-17
Make reduce scatter optional for ZeRO-1 as workaround

tjruwase:
Reduce-scatter seems to be broken in ZeRO stage 1, so make it optional as a workaround. This will enable allreduce instead by default which could impact throughput for some models.


jianzi123:
@tjruwase Why was scatter replaced from zero1 to allreduce

tjruwase:
@jianzi123, this was done a long time ago as a temporary workaround for a bug in reduce_scatter. The reduce_scatter bug has since been fixed and so this workaround is no longer needed.

PR #985, opened 2021-04-20
Update unfused_optimizer.py

awan-10:


tjruwase (approved):

deepspeed/runtime/fp16/fused_optimizer.py
```diff
@@ -397,17 +397,18 @@ def load_state_dict(self, state_dict, load_optimizer_states=True):
...
-        self.clip_grad = state_dict['clip_grad']
+            # I think it should actually be ok to reload the optimizer before the model.
+            self.dynamic_loss_scale = state_dict['dynamic_loss_scale']
+            self.cur_scale = state_dict['cur_scale']
+            self.cur_iter = state_dict['cur_iter']
+            if state_dict['dynamic_loss_scale']:
+                self.last_overflow_iter = state_dict['last_overflow_iter']
+                self.scale_factor = state_dict['scale_factor']
+                self.scale_window = state_dict['scale_window']
+                self.optimizer.load_state_dict(state_dict['optimizer_state_dict'])
```
Liyang90, line 409: The indent for this line should not be changed, it should not be under the condition 
`if state_dict['dynamic_loss_scale']:`

awan-10: Thanks for catching this. I am pushing a fix soon.

awan-10:
With many discussions across multiple teams, we figured out that a new flag like "load_module_only" will be a good idea to handle this situation. Interestingly, I wanted to write a helper function to copy fp16 params to fp32 in this case and found out that Tunji has it already for fused optimizer in https://github.com/microsoft/DeepSpeed/pull/273. I will reuse that logic and do it for the unfused optimizer. For now, we can close this PR and reopen it when we bring the internal dev branch changes to master.

PR #994, opened 2021-04-22
Fix overflow with pipe parallel / onebit optimizers

sdtblck:
I forgot to include this in my PR, but if there's an overflow using onebitadam in one pipeline parallel stage, but not another, it can silently hang.

This should fix that.

### DIFF
--- a/deepspeed/runtime/fp16/fused_optimizer.py
+++ b/deepspeed/runtime/fp16/fused_optimizer.py
@@ -8,10 +8,12 @@ This file is adapted from FP16_Optimizer in NVIDIA/apex
 import torch
 import math
 from torch._utils import _flatten_dense_tensors, _unflatten_dense_tensors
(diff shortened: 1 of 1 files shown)


conglongli:
@sdtblck yes we actually also detected this problem when preparing the 1-bit LAMB PR, and fixed it by extending the self.overflow_checker.has_overflow() function to check if enable_backward_allreduce or pipeline_enable_backward_allreduce is set to false: https://github.com/microsoft/DeepSpeed/blob/cf5ea8912afed2e0734d90a75b6b384a6ec9a2cf/deepspeed/runtime/utils.py#L120

Could you check if this solution works for your case (we did test it with unit test)? If so I think there is no need to patch again and add another allreduce communication :)

sdtblck:
Ah yep, looks like that's doing exactly the same thing so I'll just trust it works for our case too.

PR #999, opened 2021-04-23
[doc] add missing pin_memory entry

stas00:
This PR:

- `offload_param` was missing `pin_memory` - fixed
- also moved the entry in `offload_optimizer` to have it in the same place, before numerical params.


ShadenSmith (approved):
Thanks!

PR #1001, opened 2021-04-23
Minor usability tweaks/fixes

tjruwase:


csrc/aio/py_test/parse_aio_stats.py
```diff
@@ -9,31 +9,12 @@
...
-    COMPLETE_LATENCY: 'aggr: complete',
-    READ_SPEED: 'E2E Read Speed',
-    WRITE_SPEED: 'E2E Write Speed'
-}
+PERF_METRICS = [READ_SPEED, WRITE_SPEED]
+
+METRIC_SEARCH = {READ_SPEED: 'E2E Read Speed', WRITE_SPEED: 'E2E Write Speed'}
 
 NUM_BYTES = (400 * 1024 * 1024)
 NUM_GIGA_BYTES = (1024 * 1024 * 1024)
```
ShadenSmith, line 20: Could we rename to something like `BYTES_PER_GB` or just `GIGA` ? I think `NUM_GIGA_BYTES` might be misleading since its used to compute the number of GB.

csrc/aio/py_test/parse_aio_stats.py
```diff
@@ -107,12 +85,7 @@ def get_metric(file, metric):
     with open(file) as f:
         for line in f.readlines():
             if line.startswith(METRIC_SEARCH[metric]):
-                if metric == RAW_RATE:
-                    fields = line.split()
-                    raw_time_sec = float(fields[2]) / 1e06
-                    raw_rate = (thread_count * num_giga_bytes * 1.0) / raw_time_sec
-                    return raw_rate
-                elif metric in [READ_SPEED, WRITE_SPEED]:
+                if metric in [READ_SPEED, WRITE_SPEED]:
```
ShadenSmith, line 88: should the `PERF_METRICS` list created above be used instead? Or maybe there are more possible perf metrics that shouldn't trigger this branch.

csrc/aio/py_test/parse_aio_stats.py
```diff
@@ -107,12 +85,7 @@ def get_metric(file, metric):
     with open(file) as f:
```
ShadenSmith, line 85: It might be useful to document this parsing method with the assumptions of the log format and some sample lines.

csrc/aio/py_test/parse_aio_stats.py
```diff
@@ -47,13 +28,10 @@ def parse_arguments():
...
-        '--metric',
-        type=str,
-        required=True,
-        help=
-        'Performance metric to report: [raw_rate|e2e_rate|submit_latency|complete_latency]'
-    )
+    parser.add_argument('--metric',
+                        type=str,
+                        required=True,
+                        help='Performance metric to report: [read_speed|write_speed]')
```
ShadenSmith, line 34: Does "speed" here mean bandwidth or latency?

ShadenSmith (approved):

csrc/aio/py_test/parse_aio_stats.py
tjruwase: bandwidth, good point.

csrc/aio/py_test/parse_aio_stats.py
tjruwase: Yes, there could perf metrics we want to ignore or handle differently.

csrc/aio/py_test/parse_aio_stats.py
tjruwase: Turns out this is dead code anyways :)

csrc/aio/py_test/parse_aio_stats.py
tjruwase: Done.

PR #1004, opened 2021-04-23
Samyamr/full precision for ZeRO Stage2 and Stage3

samyam:


deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -154,7 +154,7 @@ class ZeroParamStatus(Enum):
 _orig_torch_empty = torch.empty
 
 
-def empty_cuda_tensor(*size, **kwargs):
+def empty_cuda_tensor_half(*size, **kwargs):
     if not 'device' in kwargs.keys():
         kwargs['device'] = torch.device('cuda:{}'.format(os.environ["LOCAL_RANK"]))
     tensor = _orig_torch_empty(*size, **kwargs)
```
tjruwase, line 160: Is this missing .half()?

tjruwase (approved):

deepspeed/runtime/zero/partition_parameters.py
samyam: The .half() happen in return

deepspeed/runtime/zero/stage2.py
```diff
@@ -187,6 +187,7 @@ def __init__(self,
         partition_id = dist.get_rank(group=self.dp_process_group)
 
         self.all_reduce_print = False
+        self.dtype = self.optimizer.param_groups[0]['params'][0].dtype
```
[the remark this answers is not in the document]
samyam: The assert in the engine will now only get triggered for stage 1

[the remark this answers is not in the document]
tjruwase: Aaah, right I misread the assert.

deepspeed/runtime/engine.py
```diff
@@ -574,6 +574,8 @@ def _configure_distributed_model(self, model):
         self.module = model
         if self.fp16_enabled():
             self.module.half()
```
samyam, line 576: Need to add an assert here to prevent case where a Z3 module is initialized as fp32 with zero.init. Simply doing a self.module.half is not going to work here for zero.init initialized module.

deepspeed/runtime/zero/partition_parameters.py
tjruwase: This is interesting. I had to view the file in order to see return with half. I don't know why the diff does not show the return at all.

deepspeed/runtime/engine.py
samyam: Also need to avoid clipping for fp32 in engine for stage 2 and stage 3

stas00:
And just to indicate priority to this PR, we have all those bfloat16 models that won't train under fp16/mixed precision, and users want to use DeepSpeed to overcome GPU memory limitations, so they badly need this. Thank you!

deepspeed/runtime/engine.py
tjruwase: Clipping for fp32 is now avoided in engine for ZeRO

stas00:
When you feel this looks good enough to test please let me know and I will start testing this branch on the `transformers` side. Thank you.

jeffra (approved):

stas00:
This is awesome - thank you!

Issues:

As I am writing HF transformers tests for fp32, here is what I find is that `zero.Init` doesn't get `dtype` from the config file, I have to explicitly do:

``` 
           ds_config = deepspeed_config()
            # XXX: Fixme - we shouldn't need to figure dtype out, it should be in the config file
            dtype = torch.float16 if ds_config.get("fp16", {}).get("enabled", True) else torch.float
            with deepspeed.zero.Init(dtype=dtype, config=ds_config):
                model = cls(config, *model_args, **model_kwargs)
```

I thought the whole point of passing `config` to `zero.Init` is so that we don't need to manually parse the file in multiple places, I was expecting this to work:

``` 
           ds_config = deepspeed_config()
            with deepspeed.zero.Init( config=ds_config):
                model = cls(config, *model_args, **model_kwargs)
```

PR #1008, opened 2021-04-27
Refactor param_dict to config

SeanNaren:
As discussed in #983, allows `deepspeed_config` to be a file path or a dictionary when passed into the `Init` function or `initialize` function. This will be a breaking change for people using `config_params` however should allow cleaner parsing in the future of all the parameters.

cc @stas00 @tjruwase

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -58,7 +58,8 @@ def initialize(args=None,
                mpu=None,
                dist_init_required=None,
                collate_fn=None,
-               deepspeed_config=None):
+               config=None,
+               config_params=None):
     """Initialize the DeepSpeed Engine.
 
     Arguments:
@@ -88,8 +89,10 @@ def initialize(args=None,
             mini-batch of Tensor(s).  Used when using batched loading from a
             map-style dataset.
 
-        deepspeed_config: Optional: Instead of requiring args.deepspeed_config you can pass your deepspeed config
+        config: Optional: Instead of requiring args.deepspeed_config you can pass your deepspeed config
             as an argument instead, as a path or a dictionary.
+
+        config_params: Optional: Same as `config`, kept for backwards compatibility.
 
     Returns:
         A tuple of ``engine``, ``optimizer``, ``training_dataloader``, ``lr_scheduler``
@@ -123,7 +126,8 @@ def initialize(args=None,
                                  mpu=mpu,
                                  dist_init_required=dist_init_required,
                                  collate_fn=collate_fn,
-                                 deepspeed_config=deepspeed_config)
+                                 config=config,
+                                 config_params=config_params)
     else:
         assert mpu is None, "mpu must be None with pipeline parallelism"
         engine = PipelineEngine(args=args,
@@ -135,7 +139,8 @@ def initialize(args=None,
                                 mpu=model.mpu(),
                                 dist_init_required=dist_init_required,
                                 collate_fn=collate_fn,
-                                deepspeed_config=deepspeed_config)
+                                config=config,
+                                config_params=config_params)
 
     return_items = [
         engine,
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -64,7 +64,7 @@ def initialize(args=None,
 
     Arguments:
         args: an object containing local_rank and deepspeed_config fields.
-            This is optional if `deepspeed_config` is passed.
+            This is optional if `config` is passed.
 
         model: Required: nn.module class before apply any wrappers
 
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -58,11 +58,12 @@ def initialize(args=None,
                mpu=None,
                dist_init_required=None,
                collate_fn=None,
-               config_params=None):
+               deepspeed_config=None):
     """Initialize the DeepSpeed Engine.
 
     Arguments:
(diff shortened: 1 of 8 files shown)


deepspeed/__init__.py
```diff
@@ -58,11 +58,12 @@ def initialize(args=None,
                mpu=None,
                dist_init_required=None,
                collate_fn=None,
-               config_params=None):
+               deepspeed_config=None):
```
stas00, line 61: Sorry, I think we have had a bit of a miscommunication in https://github.com/microsoft/DeepSpeed/pull/983

This would be a breaking change.

I think it'd be OK for `transformers` since we constantly raise the minimal requirement for deepspeed, but is this the case for other DeepSpeed users?

deepspeed/runtime/config.py
```diff
@@ -521,16 +523,14 @@ def write_config(self, filename):
 
 
 class DeepSpeedConfig(object):
-    def __init__(self, json_file, mpu=None, param_dict=None):
+    def __init__(self, ds_config: Union[str, dict], mpu=None):
         super(DeepSpeedConfig, self).__init__()
-
-        if param_dict is None:
+        self._param_dict = ds_config
+        if not isinstance(ds_config, dict) and os.path.exists(ds_config):
```
stas00, line 529: this would break if the path doesn't exist, since json won't be loaded and we are now missing the dict completely.

so this needs `else: raise ValueError(...)` if the check fails.

deepspeed/__init__.py
SeanNaren: Had a feeling, lots of concepts were discussed :)

If we like the change but mind BW compat, I can re-add the `config_params` parameter, setting it to `deepspeed_config` (optionally I'll print a warning message, if we like).

deepspeed/runtime/config.py
SeanNaren: Let me know if it look good now!

deepspeed/__init__.py
stas00: My feeling is that they haven't used `deepspeed_` prefix in the first place, because it's all deepspeed and thus doesn't require self-reference ;)

SeanNaren: haha that's a great point! maybe `config_params` should be kept as is then? Just internally we've renamed?

I'd prefer just `config` if I had to choose though

stas00: Yes, config params repeats itself twice.

I agree `config` would be the cleanest.

Well, since the deepspeed trusts us to mess with their API, I hope it won't be overstepping if you made it `config` and left `config_params` as a back-compat deprecated arg name. 

But personally I'm totally fine with `config_params` since it's only one place where it's used.

Internally there is no back-compat issue I guess, so if it makes the code more intuitive then it's probably a good choice.

deepspeed/runtime/config.py
stas00: It's perfect, thank you for adding the check, @SeanNaren

deepspeed/__init__.py
SeanNaren: so with that said, `config` will be the new name, and `config_params` will remain as BW compat, let's see if @tjruwase is cool with it!

tjruwase: Thanks guys for driving this conversation. Yes, we have avoided using `deepspeed_` because of the self-reference. I like the proposal of `config` for new name and `config_params` for BW compat. I will also ping the rest of the folks here for awareness of the changes.

ShadenSmith: I am also a fan of `config` for the new name!

jeffra: Sounds good to me as well :)

deepspeed/__init__.py
```diff
@@ -58,11 +58,13 @@ def initialize(args=None,
...
                dist_init_required=None,
                collate_fn=None,
+               config=None,
                config_params=None):
     """Initialize the DeepSpeed Engine.
 
     Arguments:
-        args: an object containing local_rank and deepspeed_config fields. This is optional if `config_params` is passed.
+        args: an object containing local_rank and deepspeed_config fields.
+            This is optional if `deepspeed_config` is passed.
```
SeanNaren, line 67:
```suggestion
            This is optional if `config` is passed.
```

SeanNaren:
Thanks everyone for pitching in :) should be good to go, let me know if there is anything that needs addressing!

deepspeed/__init__.py
```diff
@@ -122,6 +126,7 @@ def initialize(args=None,
                                  mpu=mpu,
                                  dist_init_required=dist_init_required,
                                  collate_fn=collate_fn,
+                                 config=config,
```
stas00, line 129: Do `DeepSpeedEngine` and `PipelineEngine` still need `config_params` though? Are those public APIs or internal? If they are internal then perhaps it's enough to alias `config_params` to `config` in `initialize` and adjust those 2 classes to receive just `config` ?

SeanNaren: Was contemplating this, if preferred I can move the logic to do this into the `initialize` function instead of doing it internally! Wanted to keep `initialize` as clean as possible, but didn't know which would be preferred

stas00: Oh, I see, this is perfectly fine then. This is good. Thank you!

jeffra (approved):

PR #1014, opened 2021-04-28
Add function `to_sequential` to PipelineModule

sdtblck:
In https://github.com/EleutherAI/gpt-neox we were previously maintaining two separate models - one if the user wanted to use pipeline parallel, and one if they didn't.

The more straightforward solution was to add a `to_sequential` function to export the PipelineModule as an nn.Sequential model, so we could train with deepspeed features that aren't compatible with pipe parallel (i.e ZeRO 2+).

Figure this might be a useful addition to the base module, too. I'm not 100% sure if the support for tied layers here is as flexible as it could / should be, since their capabilities are not very well documented, but it works at least for our purposes (with tied Embeddings as the output layer).

### DIFF
--- a/deepspeed/runtime/pipe/module.py
+++ b/deepspeed/runtime/pipe/module.py
@@ -19,6 +19,15 @@ from .topology import PipeDataParallelTopology, PipelineParallelGrid
 class PipelineError(Exception):
     """Errors related to the use of deepspeed.PipelineModule """
 
+class Lambda(torch.nn.Module):
+    def __init__(self, func):
+        super().__init__()
+        self.func = func
+
+    def forward(self, x):
+        return self.func(x)
+
+
 
 class LayerSpec:
     """Building block for specifying pipeline-parallel modules.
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/pipe/module.py
```diff
@@ -578,3 +587,31 @@ def _is_checkpointable(self, funcs):
...
+                    # owner
+                    module = spec.build(log=False)
+                    layers.append(module)
+                    tied_layers[spec.key].append(module)
+            elif isinstance(spec, LayerSpec):
+                layers.append(spec.build(log=False))
+            else:
+                # check that it's a lambda function
+                LAMBDA = lambda:0
+                if isinstance(spec, type(LAMBDA)) and spec.__name__ == LAMBDA.__name__:
```
ShadenSmith, line 612: `PipelineModule` should work with any callable object, and I think the `Lambda` module above will too.  Maybe the filtering condition could be `hasattr(spec, '__call__')` to support things like named methods?

ShadenSmith (commented):
This is a great idea, thanks @sdtblck !

One caveat is that we lose the activation checkpointing that the `PipelineModule`'s forward can be configured to use. But users can instead use torch's `checkpoint_sequential()` if they want checkpointing. Or we could wrap the layers in a similar way as `Lambda` if we really want to mirror functionality. What are your thoughts?

ShadenSmith:
In addition to `to_sequential`, there may be another way we could accomplish this while keeping the normal `PipelineModule`, if that would be useful. 

If we short-circuit this condition and use the regular training engine, I *think* that `PipelineModule` should behave as a normal `torch.nn.Module` and you can use ZeRO-2, etc.  I intended for that to be the case, but not tested these days.

https://github.com/microsoft/DeepSpeed/blob/dad26428e3f28898b8d0f5ace1b3df3e6db8f8e8/deepspeed/__init__.py#L119-L120

sdtblck:
> In addition to `to_sequential`, there may be another way we could accomplish this while keeping the normal `PipelineModule`, if that would be useful.
> 
> If we short-circuit this condition and use the regular training engine, I _think_ that `PipelineModule` should behave as a normal `torch.nn.Module` and you can use ZeRO-2, etc. I intended for that to be the case, but not tested these days.
> 
> https://github.com/microsoft/DeepSpeed/blob/dad26428e3f28898b8d0f5ace1b3df3e6db8f8e8/deepspeed/__init__.py#L119-L120

Hi @ShadenSmith , I actually tried this as well - and it seems this way of doing things drops any tied modules (since it seems the pipe engine handles them specially.) So for example, if we used this with a model with tied embeddings, the to_logits function that uses the word embedding weights would just get silently dropped.

sdtblck:
> This is a great idea, thanks @sdtblck !
> 
> One caveat is that we lose the activation checkpointing that the `PipelineModule`'s forward can be configured to use. But users can instead use torch's `checkpoint_sequential()` if they want checkpointing. Or we could wrap the layers in a similar way as `Lambda` if we really want to mirror functionality. What are your thoughts?

Hm. Yeah this is a good point that I had overlooked. I'll spend some time looking into the best way to get this working today.

deepspeed/runtime/pipe/module.py
sdtblck: good point yes! I'll make that change too

sdtblck:
Hi @ShadenSmith 

I think the two latest commits should fix both the above requirements. There is maybe some repeated code between `SequentialModel` and `PipelineModule` that could be slimmed down - but I have tested with gpt-neox and it works well.

rocm-mici:

loadams:
@sdtblck - just fixed some formatting issues that were preventing this - if the tests pass, would this be good to merge now?

Issue #1009, opened 2021-04-27
Reconstruction of fp32 weights on stage3 doesn't work


exelents:
In #892 @stas00 proposed a new script which can consolidate fp32 weights from fp16 model checkpoint on stage 3 training.
Unfortunately I have found? that t5-11b model can't be consolidated due to some error:
```
└──>$ ./zero_to_fp32.py global_step3250/ pytorch_model1.bin
Processing zero checkpoint 'global_step3250/'
Detected checkpoint of type zero stage 3, world_size: 1
Traceback (most recent call last):
  File "./zero_to_fp32.py", line 151, in <module>
    convert_zero_chkpt_to_fp32_consolid_state_dict(args.checkpoint_dir, args.output_file)
  File "./zero_to_fp32.py", line 122, in convert_zero_chkpt_to_fp32_consolid_state_dict
    tuple(fp32_flat_groups[i].narrow(0,
  File "./zero_to_fp32.py", line 122, in <genexpr>
    tuple(fp32_flat_groups[i].narrow(0,
RuntimeError: start (32899072) + length (16777216) exceeds dimension size (32899072).
```

Maybe @stas00 could say what is the problem, and how it can be fixed?

exelents:
Here is steps to reproduce my error with model reconstruction:

- Clone this repo:
    https://github.com/exelents/try_t5_siamese

- Extract folder "qasc" from this archive:
    https://drive.google.com/file/d/1gwvFiPzWW0JLr0XLS25PuG2Br5S4fPbR/view?usp=sharing

- Go to clonned repo folder and run ./create-siamese-template.sh - it will create siamese NN from two t5-small encoders in folder ./templates/siamese-t5-small-template

- then you can run ./run-siamese-small.sh and wait until any checkpoint would be saved. If you don't want to wait for 1625 steps you can change checkpoint save period in ./run-siamese-small.sh in parameter --save_steps

- go to ./output_dir/<your_latest_checkpoint> and run 
    ./zero_to_fp32.py ./<global_step_path> pytorch_model1.bin

stas00:
Could you please save one checkpoint and share it - on my dev machine I won't be able to train 45GB model.

Alternatively, can you reproduce the same problem with a much smaller model? say t5-small? Or do you get it only with t5-11b?

exelents:
Yes, I can reproduce it on t5-small
```
└──>$ ./zero_to_fp32.py global_step10/ pytorch_model1.bin 
[...]

PR #1017, opened 2021-04-29
[zero_to_fp32.py] support param groups

stas00:
In my original version I happened to use a model with 1 param group so I wasn't aware that there could be multiple flattened tensors - one per group, so my reconstruction script [was breaking]( https://github.com/microsoft/DeepSpeed/issues/1009) when it run into non-one-single-flat-tensor.

This PR tries to fix that.

There might be a more efficient way to do it, but for now just trying to make sure that it functions correctly.

I also left some disabled debug code for now while it's new and likely to need debug still. We can remove it later if we feel it's solid.

While I tested this on a few live models, it'd be great to have a functional test for zero2 and zero3 for this code. But I'm not quite familiar with how your test suite is done and unfortunately don't really have time right now to sort it out.

The simplest conceptual test would be:
```
# original_fp32_state_dict_path
model = (multi-param-group model).from_pretrained(original_fp32_state_dict_path)
engine = deepspeed.initialize(model, ...)
engine.save_checkpoint()
! ./zero_to_fp32.py global_step1 pytorch_model.bin
! diff pytorch_model.bin original_fp32_state_dict_path # should be identical
```

Fixes: https://github.com/microsoft/DeepSpeed/issues/1009

@exelents, please check that this PR solves the problem for you.

### DIFF
--- a/deepspeed/utils/zero_to_fp32.py
+++ b/deepspeed/utils/zero_to_fp32.py
@@ -17,6 +17,8 @@ import deepspeed
 # while this script doesn't use deepspeed to recover data, since the checkpoints are pickled with
 # DeepSpeed data structures it has to be available in the current python environment.
 
+debug = 0
+
 
 def get_optim_files(checkpoint_dir):
 
@@ -41,6 +43,8 @@ def parse_optim_states(files):
     if not "zero_stage" in state_dicts[0]['optimizer_state_dict']:
         raise ValueError(f"non zero checkpoint")
     zero_stage = state_dicts[0]['optimizer_state_dict']["zero_stage"]
+    world_size = state_dicts[0]['optimizer_state_dict']["partition_count"]
+    param_shapes = state_dicts[0]["param_shapes"]
 
     # the groups are named differently in each stage
     if zero_stage == 2:
@@ -50,12 +54,15 @@ def parse_optim_states(files):
     else:
         raise ValueError(f"unknown zero stage {zero_stage}")
 
-    param_shapes = state_dicts[0]["param_shapes"]
+    # if there is more than one param group, there will be multiple flattened tensors - one
+    # flattened tensor per group - for simplicity merge them into a single tensor
+    #
+    # XXX: could make the script more memory efficient for when there are multiple groups - it
+    # will require matching the sub-lists of param_shapes for each param group flattened tensor
     fp32_flat_groups = [
-        state_dicts[i]['optimizer_state_dict'][fp32_groups_key][0]
+        torch.cat(state_dicts[i]['optimizer_state_dict'][fp32_groups_key], 0)
         for i in range(len(state_dicts))
     ]
-    world_size = state_dicts[0]['optimizer_state_dict']["partition_count"]
 
     return zero_stage, world_size, param_shapes, fp32_flat_groups
 
@@ -93,6 +100,10 @@ def convert_zero_chkpt_to_fp32_consolid_state_dict(checkpoint_dir, output_file):
     # - for zero3 we need to zip the partitions together at boundary of each param, re-consolidating
     # each param, while dealing with padding if any
 
+    if debug:
+        for i in range(world_size):
+            print(f"fp32_flat_groups[i].shape={fp32_flat_groups[i].shape}")
+
     if zero_stage == 2:
         # XXX: memory usage doubles here (zero2)
         full_single_fp32_vector = torch.cat(fp32_flat_groups, 0)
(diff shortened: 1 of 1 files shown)


deepspeed/utils/zero_to_fp32.py
```diff
@@ -50,12 +54,15 @@ def parse_optim_states(files):
...
+    # if there is more than one param group, there will be multiple flattened tensors - one
+    # flattened tensor per group - for simplicity merge them into a single tensor
+    #
+    # XXX: could make the script more memory efficient for when there are multiple groups - it
+    # will require matching the sub-lists of param_shapes for each param group flattened tensor
     fp32_flat_groups = [
-        state_dicts[i]['optimizer_state_dict'][fp32_groups_key][0]
-        for i in range(len(state_dicts))
+        torch.cat(state_dicts[i]['optimizer_state_dict'][fp32_groups_key],
+                  0) for i in range(len(state_dicts))
```
stas00, line 64: This is the only functional change in this PR. Instead using just the first element, it now uses them all.

exelents:
Converting of "siamese" model based on t5-11b encoders is done successfully. But when I load it into CPU memory I gone a stange message:
```
Some weights of T5Siamese were not initialized from the model checkpoint at ./siamese_train_deepspeed/models/siamese-t5-11b-fp16/checkpoint-1625 and are newly initialized: ['encoder_left.encoder.embed_tokens.weight', 'encoder_right.encoder.embed_tokens.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
```

deepspeed/utils/zero_to_fp32.py
tjruwase: This seems fine for now. I agree we have to revisit, especially for very large models that could cause CPU OOM.

tjruwase (approved):

stas00:
Thank you for running the checks, @exelents 

> ./siamese_train_deepspeed/models/siamese-t5-11b-fp16/checkpoint-1625 and are newly initialized: 
> ['encoder_left.encoder.embed_tokens.weight', 'encoder_right.encoder.embed_tokens.weight']

You have `encoder_right.shared.weight` - aren't those tied/aliased? (Same for left)

exelents:
> You have `encoder_right.shared.weight` - aren't those tied/aliased? (Same for left)

Yes, I have these weights returned from named_parameters() function of my model, so I suppose they should exist in the checkpoint.

```
dd = list(model_engine.module.named_parameters())
dd = list(filter(lambda x: 'shared' in x[0], dd))
dd
```
```
[('encoder_left.shared.weight',
  Parameter containing:
  tensor([1.], device='cuda:0', dtype=torch.float16, requires_grad=True)),
 ('encoder_right.shared.weight',
  Parameter containing:
  tensor([1.], device='cuda:0', dtype=torch.float16, requires_grad=True))]
```

stas00:
But that's what I'm saying: the checkpoint does have `encoder_left.shared.weight` and `encoder_right.shared.weight' and those are restored. 

The loader complains about `encoder_left.encoder.embed_tokens.weight` and `encoder_right.encoder.embed_tokens.weight`.

In your `dd` dump above they don't exist.

BTW, the new version of `zero_to_fp32.py` has a `debug` flag - turn it on and when you run it you will see each weight as it gets loaded.

exelents:
I turned on debug flag and it show me all weights in checkpoint including:
└──>$ cat debug.txt | grep shared
encoder_left.shared.weight full shape: torch.Size([32128, 512]) partition0 numel=16449536 partitioned_padding_numel=0
encoder_right.shared.weight full shape: torch.Size([32128, 512]) partition0 numel=16449536 partitioned_padding_numel=0

stas00:
We have already established that.

Please review https://github.com/microsoft/DeepSpeed/pull/1017#issuecomment-829120612 the warning is for 2 other names.

exelents:
Ah, okay. I understood. In the code of T5EncoderModel I see that variable encoder_right.encoder.embed_tokens is initialized from external variable encoder_left.shared while T5Stack is created, and it isn't needed to be saved.
Thank you.

stas00:
Yes, probably could clean that up so that it doesn't produce a misleading warning.

The key is to please check that the resumed checkpoint scores well for you. I did only a quick 100 or so steps and the loss looked correct. Also please re-check with zero2. I did test it as well, but a second pair of eyes is always better.

I was just concerned that perhaps somehow the saved weights weren't in the same order as the `param_names` dict, so it'd always reshape it correctly, since it's the same number of elements once it's all flattened into a single tensor, but I checked that the order appears to be correct.

PR #1026, opened 2021-04-30
[Stage][Fix] Add additional conditions when checking types of output from the model

SeanNaren:
Fixes an upstream issue in Lightning that I found when testing the latest DeepSpeed version on our CI.

From the forward pass in Lightning, we offer a few different potential output types. Due to the additional type checking of the output, some of these failed (`None` and `Dict`). This adds additional checks to ensure we handle those cases within the hook.

If possible I'd like to add a test (even though Lightning CI caught this), any guidance on where would be the right place?

### DIFF
--- a/tests/unit/test_zero_context.py
+++ b/tests/unit/test_zero_context.py
@@ -243,3 +243,42 @@ def test_ext_param_returnobj():
         assert len(net.dangler._external_params) == 0
         engine.backward(loss)
         engine.step()
+
+
+class ModelContainerVariableOutputType(ModelContainer):
+    def __init__(self, dim=16, output_type=dict):
+        super().__init__()
+        self.output_type = output_type
+        self.dim = dim
+        self.linear1 = torch.nn.Linear(dim, dim)
+
+    def forward(self, input):
+        act1 = self.linear1(input)
+        if self.output_type is None:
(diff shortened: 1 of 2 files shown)


tests/unit/test_zero_context.py
```diff
@@ -243,3 +243,42 @@ def test_ext_param_returnobj():
...
+        self.dim = dim
+        self.linear1 = torch.nn.Linear(dim, dim)
+
+    def forward(self, input):
+        act1 = self.linear1(input)
+        if self.output_type is None:
+            return None
+        if self.output_type is dict:
+            return {'loss': act1.sum()}
+        return act1.sum()
```
tjruwase, line 261: It seems this code path will not be executed since the output types are restricted to dict and None?

SeanNaren: Yep totally right, I can remove this (wanted to keep it general)

tjruwase: It is good to keep it general, then perhaps create a test case with a different output type?

SeanNaren: Done @tjruwase let me know what you think :)

tjruwase (approved):

Issue #1033, opened 2021-05-03
AssertionError: num_elems 513024> buffer 513000.


Vbansal21:
My System:
Ubuntu groovy 20.10
NVIDIA-SMI 460.73.01    Driver Version: 460.73.01    CUDA Version: 11.2
torch 1.8.1
deepspeed 0.3.16

For some combination it works, while for others it doesn't, it is very random.

code:
`import deepspeed, torch

model = torch.nn.Sequential(
    torch.nn.Linear(512,1000)
)

deepspeed_args = {
  "train_batch_size": 1,
  "gradient_accumulation_steps": 1,
  "fp16": {
    "enabled": True,
    "loss_scale": 0,
    "initial_scale_power": 3,
    "loss_scale_window": 1000,
    "hysteresis": 1,
    "min_loss_scale": 1
    },
  "gradient_clipping":1.0,
  "zero_optimization": {
    "stage": 3,
    "offload_param":{
        "device": "nvme",
        "nvme_path":"/mnt/nvme0n1p3/"
        },
    "offload_optimizer": {
        "device": "nvme",
        "nvme_path": "/mnt/nvme0n1p3/"
        },
    }
    
}

optimizer = torch.optim.Adam(model.parameters(), lr= 0.001,betas=[0.8,0.99],eps=1e-8,weight_decay=3e-7)
scheduler = torch.optim.lr_scheduler.StepLR(optimizer, 1.0, gamma=0.95)

model, optimizer,_,scheduler = deepspeed.initialize(model=model,optimizer=optimizer,lr_scheduler=scheduler,config_params=deepspeed_args)`

error:
`[2021-05-03 12:50:20,837] [INFO] [logging.py:60:log_dist] [Rank -1] DeepSpeed info: version=0.3.16, git-hash=unknown, git-branch=unknown
[2021-05-03 12:50:20,838] [INFO] [distributed.py:36:init_distributed] Not using the DeepSpeed or torch.distributed launchers, attempting to detect MPI environment...
[2021-05-03 12:50:21,196] [INFO] [distributed.py:83:mpi_discovery] Discovered MPI settings of world_rank=0, local_rank=0, world_size=1, master_addr=192.168.29.88, master_port=29500
[2021-05-03 12:50:21,197] [INFO] [distributed.py:46:init_distributed] Initializing torch distributed with backend: nccl
[...]

PR #1036, opened 2021-05-03
Align optimizer swap buffer sizes

tjruwase:
Optimizer swap buffer sizes should be aligned, fix for #1033. 
NVMe intra-request validation should be on entire file.


tjruwase:
Added fix for #1037.

samyam (approved):

jeffra (approved):

stas00:
This doesn't quite fix https://github.com/microsoft/DeepSpeed/issues/1037 though. The problem is that the warning is not matching the traceback. and very difficult to see and make a correlation with the JIT crash. 

This needs to be an assert.

tjruwase:
@stas00, good point. I have re-opened the issue. I think we require a more general solution because assert is heavy-weight in some cases. But I agree we need to make the failure messages more meaningful.

PR #1039, opened 2021-05-03
Avoid unused parameters assert by default

tjruwase:
Unused parameters assert should be disabled by default

### DIFF
--- a/deepspeed/runtime/zero/stage2.py
+++ b/deepspeed/runtime/zero/stage2.py
@@ -898,9 +898,9 @@ class FP16_DeepSpeedZeroOptimizer(object):
                     # avoid the error
                     # Error msg adapted from torch.nn.parallel.DistributedDataParallel
                     assert not self.find_unused_parameters, """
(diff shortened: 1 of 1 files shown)


jeffra (approved):

ghosthamlet:
@tjruwase Thanks for notification. I noticed you changed two things, I not very sure why that is better, can you explain?
1. disabled unused parameters assert. Why not just follow `torch.nn.parallel.DistributedDataParallel`'s practice to enable it for easy debug? If you disabled it default, then most time it is hard for users to notice the unsed parameters in their models, then this option may be not very useful for debugging.
2. find_unused_parameters changed to ignore_unused_parameters. Why not follow `torch.nn.parallel.DistributedDataParallel`'s naming practice?

ghosthamlet:
Besides, I found you forgot to change the doc: https://github.com/microsoft/DeepSpeed/blob/master/docs/_pages/config-json.md#zero-optimizations-for-fp16-training.

PR #1053, opened 2021-05-07
[configure_distributed_model] improve assert

stas00:
This PR changes the 2 asserts to actually print the names of the params that are wrong. e.g.:
```
fp16 is enabled but the following parameters have dtype that is not fp16: wav2vec2.masked_spec_embed
```


stas00:
Honestly, the way this autoformatter sometimes formats the code - it makes absolutely no sense readability-wise 

```
                   names = [
                        n for n,
                        p in self.module.named_parameters() if p.dtype != torch.half
                    ]
```

surely there are at least slightly better necessarily evil autoformatters out there...

tjruwase (approved):

Issue #967, opened 2021-04-16
ZeRO Stage1 behaviour changed between v0.3.2 and v0.3.3


e-sha:
Hello,
I found significant difference in BERT training curves between versions v0.3.2 and v0.3.3. The [log](https://github.com/microsoft/DeepSpeed/commit/08c96a1bc6d9c0345b31889bba66d73f20d0b0b5#) doesn't help me to understand the reason. Is it an expected behaviour?

![Screenshot from 2021-04-16 17-08-02](https://user-images.githubusercontent.com/6618622/115036618-5a4b3400-9ed6-11eb-8cf3-84dd3946f53f.png)
![Screenshot from 2021-04-16 17-10-17](https://user-images.githubusercontent.com/6618622/115036990-ae561880-9ed6-11eb-9bcf-e0eae6aaa430.png)

I tried pretraining of a shallow BERT model. My deepspeed config:
```
{
  "train_batch_size": 384,
  "train_micro_batch_size_per_gpu": 48,
  "steps_per_print": 1000,
  "prescale_gradients": false,
  "zero_allow_untested_optimizer": true,
  "optimizer": {
    "type": "adam",
    "params": {
      "lr": 0.00152,
      "weight_decay": 0.01,
      "bias_correction": false
    }
  },
  "gradient_clipping": 1.0,

  "wall_clock_breakdown": false,

  "fp16": {
    "enabled": true,
    "loss_scale": 0,
    "initial_scale_power": 20
  },
  "zero_optimization": {
      "stage": 1
  }
}
```
My BERT config:
```
{
  "attention_probs_dropout_prob": 0.1,
  "hidden_act": "gelu",
  "hidden_dropout_prob": 0.1,
  "hidden_size": 2048,
  "initializer_range": 0.02,
  "intermediate_size": 8192,
  "max_position_embeddings": 512,
  "num_attention_heads": 32,
  "num_hidden_layers": 2,
  "type_vocab_size": 10,
  "vocab_size_or_config_json_file": 103296
}
```

aced125:
uh oh

tjruwase:
@e-sha, can you please try this [branch](https://github.com/microsoft/DeepSpeed/tree/olruwase/zero_optional_reduce_scatter)?

e-sha:
Thank you for the reply. I will check it next week and will send you the results.

e-sha:
Yes it fixes the problem. Please explain the problem in more details.
[...]

PR #1058, opened 2021-05-08
fix ZERO_OPTIMIZATION_REDUCE_SCATTER_DEFAULT default value 

mpjlu:
There is more than 20% performance improvement when set ZERO_OPTIMIZATION_REDUCE_SCATTER_DEFAULT = true.

The default value in the docs is also true. 
https://www.deepspeed.ai/docs/config-json/#zero-optimizations-for-fp16-training
image

Most people will not set reduce_scatter in the config file,  to avoid the performance issue, we propose to do this change.


mpjlu:
@tjruwase

e-sha:
There is an open issue with reduce-scatter and ZeRO stage 1
https://github.com/microsoft/DeepSpeed/issues/967

mpjlu:
> There is an open issue with reduce-scatter and ZeRO stage 1
> #967
reduce-scatter not only used for stage 1, but also stage 2. And the current stage 1 implementation will be deprecated, so we propose to revert   ZERO_OPTIMIZATION_REDUCE_SCATTER_DEFAULT default value to true.

mpjlu:
> #1042 should fix the perf and correctness issues in zero stage 1. It will be merged early this week. Feel free to give it an early try though. Hopefully you see better performance :)

Yes, I think the perf and correctness issues in zero stage 1 will be fixed. Will we revert ZERO_OPTIMIZATION_REDUCE_SCATTER_DEFAULT default value to true?

jeffra:
> > #1042 should fix the perf and correctness issues in zero stage 1. It will be merged early this week. Feel free to give it an early try though. Hopefully you see better performance :)
> 
> Yes, I think the perf and correctness issues in zero stage 1 will be fixed. Will we revert ZERO_OPTIMIZATION_REDUCE_SCATTER_DEFAULT default value to true?

I’ll double check tomorrow but I believe the new PR does this.

mpjlu (commented):
no newline at end of file

jeffra (approved):

Issue #998, opened 2021-04-23
[doc] compile NVMe profiling docs


stas00:
Let's use this issue to gather instructions on how to profile one's NVMe.

Instructions from @tjruwase so far:
```
# read benchmark
cd csrc/aio/py_test
dd if=/udev/urandom of=input.file count=400 bs=1M
# create a new dir, otherwise the script deletes its contents
mkdir read-logs
./run_read_sweep.sh input.file read-logs
python parse_aio_stats.py --logdir read-logs --metric read_speed | sort -k9 -n
```
this assumes the cwd is on the NVMe drive.

the tail of the list should show the fastest speeds.

note that it tests the real setup, NVMe<->CPU<->GPU so make sure to test this on the actual system that you intend to use it one.

So for writing it'd be:
```
# write benchmark
cd csrc/aio/py_test
mkdir write-test-data
mkdir write-logs
./run_write_sweep.sh 100 write-test-data write-logs
python parse_aio_stats.py --logdir write-logs --metric write_speed | sort -k9 -n
```

@tjruwase, what else should we document here? 

e.g. how to read the reports?

what are the good numbers? So that the users will know if their NVMe is fast enough?

----------------------

Some stats: on my xpg gammix s11 pro 2tb NVMe drive with published specs of:
- max read speed of 3500 MB/s
- max write speed of up to 3000 MB/s

I get:
- read between 1-3.15GB/s
- write ???

I tried my 860 Evo SSD and getting about 0.5GB/s read, but I only run a small portion of the benchmark.

-------------

tjruwase:
Benchmark results from DGX-2 node, which has 8 Micron 9200 NVMe raided into a single volume. 

  | Peak | 1-process | multi-process
-- | -- | -- | --
Read | 28 | 25.3 | 25.6
Write | 24.8 | 19.2 | 21.7

tjruwase:
The sweep results suggest that zero-infinity can be configured to do offloading at 3GB/sec read and 2.6GB/sec write. So you want to configure the [asynchronous I/O module](https://www.deepspeed.ai/docs/config-json/#asynchronous-io) similar to configurations that achieve these numbers. Specifically, you want to add the following to deepspeed config: 

```json
aio: {
 "block_size": 262144,
 "queue_depth": 32,
 "thread_count": 1,
"single_submit": false,
"overlap_events": true
}
````

tjruwase:
[...]

PR #1059, opened 2021-05-10
Auto-tuning aio parameters

tjruwase:
Performance scripts for auto-tuning/auto-generating aio parameters in deepspeed config. 
Closing #998.


tjruwase:
@stas00, thanks for the feedback. 

1. --log_dir, I agree with using something other than `.` as default value. How about something more specific, like `aio_bench_logs`?
2. --io_dir, I think users do need to specify this so that the script can be run from a different folder than the one being tested. However, the script does create and delete a subfolder that contains the read and write files.

stas00:
> @stas00, thanks for the feedback.
> 
>     1. --log_dir, I agree with using something other than `.` as default value. How about something more specific, like `_aio_bench_logs`?

That's perfect.

>     2. --io_dir, I think users do need to specify this so that the script can be run from a different folder than the one being tested. However, the script does create and delete a subfolder that contains the read and write files.

Ah, yes! we discussed that already - I forgot!

Should we rename it to match mnemonics better? e.g. `nvme_dir`

Also I'm not sure the description is clear:

```
  --io_dir IO_DIR       Directory in which to perform I/O tests. Normally 
                        where (NVMe) device is mounted.
```

That would suggest to me to use the mount point of the drive. Perhaps:

> a writable sub-dir on an NVMe drive

? Not liking it 100% either...

conglongli (approved):
General code style/logic looks good to me.

PR #1065, opened 2021-05-12
Improve flops profiler functionality

cli99:
This PR makes the following improvements to the flops profiler:
1. supports pytroch > 1.8
2. use timing information from wall clock timers to compute FLOPS for bwd pass, and overall iteration
3. prints data parallelism, model parallelism, samples/second, etc.
4. change the aggregated profile to print from top to the specified module depth.
5. adds bert-large example
6. adds the option to print to file rather than stdout


tjruwase (approved):

deepspeed/runtime/engine.py
```diff
@@ -160,6 +160,14 @@ def __init__(self,
             steps_per_output=self.steps_per_print(),
             monitor_memory=False)
 
+        if dist.get_rank() == 0:
+            logger.info(
+                f"DeepSpeed Flops Profiler Enabled: {self.flops_profiler_enabled()}")
+
+        if self.flops_profiler_enabled():
+            self.flops_profiler = FlopsProfiler(self.module, self)
+            self.wall_clock_breakdown()
```
tjruwase, line 169: What does this line do?

deepspeed/utils/timer.py
```diff
@@ -37,7 +37,7 @@ def stop(self):
             """Stop the timer."""
             assert self.started_, 'timer is not started'
             torch.cuda.synchronize()
-            self.elapsed_ += (time.time() - self.start_time)
+            self.elapsed_ = (time.time() - self.start_time)
```
tjruwase, line 40: Is there a reason for changing this from accumulation? There is a reset api for explicitly resetting `self.eplased_`.

deepspeed/profiling/flops_profiler/README.md
```diff
@@ -3,79 +3,152 @@
...
+params of model = params per GPU * mp_size:                   336.23 M
+fwd MACs per GPU:                                             3139.93 G
+fwd flops per GPU = 2 * fwd MACs per GPU:                     6279.86 G
+fwd flops of model = fwd flops per GPU * mp_size:             6279.86 G
+fwd latency:                                                  76.67 ms
+bwd latency:                                                  108.02 ms
+fwd FLOPS per GPU = fwd flops per GPU / fwd latency:          81.9 TFLOPS
+bwd FLOPS per GPU = 2 * fwd flops per GPU / bwd latency:      116.27 TFLOPS
+fwd+bwd FLOPS per GPU = 3 * fwd flops per GPU / (fwd+bwd latency):   102.0 TFLOPS
+step latency:                                                 34.09 us
```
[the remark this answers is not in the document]
cli99: for the overall FLOPs per GPU, shall we use the equation: 3 * fwd flops per GPU / (fwd latency + bwd latency + step latency)?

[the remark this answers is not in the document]
tjruwase: I think so. This is our current practice with ZeRO + Megatron numbers.

deepspeed/utils/timer.py
cli99: I would like to have the previous captured elapsed time saved somewhere and it seems the accumulation of self.elapsed_ is not used anywhere and the timers get reset when a new iteration starts. I can also use a new field in the timer to record past timings.

deepspeed/runtime/engine.py
cli99: wall_clock_breakdown is enabled when flops profiler is enabled. this is done elsewhere and this line is a typo.

tjruwase: Got it, so line 169 will be removed.

deepspeed/utils/timer.py
tjruwase: I see. The nvme offloading uses accumulation property to time asynchronous operations such as [here](https://github.com/microsoft/DeepSpeed/blob/1f82ab78c2206b97f2d27041da8b7b127180e8e0/deepspeed/runtime/swap_tensor/async_swapper.py#L140).  If you need a reset on stop(), perhaps you can add a reset_flag (default false) to the api?

tjruwase (approved):

deepspeed/utils/timer.py
cli99: That works well. I reverted the change to the elapsed_ accumulation and added the reset flag to timer stop() https://github.com/cli99/DeepSpeed/blob/fix-flops-profiler/deepspeed/utils/timer.py#L36.

g-karthik:
@cli99 this is great, thank you for this feature @cli99!

I was just wondering last week why my DeepSpeed jobs suddenly stopped printing the flops profile!

PR #1066, opened 2021-05-12
[docs] Rename train_step_batch_size to train_micro_batch_size_per_gpu 

wbuchwalter:
Config file doc uses both `train_step_batch_size` and  `train_micro_batch_size_per_gpu` to refer to the same thing, this is a bit confusing.
Renamed everything to `train_micro_batch_size_per_gpu` which is the actual key name in the config json.


docs/_pages/config-json.md
```diff
@@ -24,7 +24,7 @@ title: "DeepSpeed Configuration JSON"
...
-| Number of training steps to accumulate gradients before averaging and applying them. This feature is sometimes useful to improve scalability since it results in less frequent communication of gradients between steps. Another impact of this feature is the ability to train with larger batch sizes per GPU. When specified, <i>**train_step_batch_size**</i> is automatically calculated using <i>**train_batch_size**</i> and number of GPUs. Should not be concurrently specified with <i>**train_step_batch_size**</i> in the configuration JSON. | `1`     |
+| Number of training steps to accumulate gradients before averaging and applying them. This feature is sometimes useful to improve scalability since it results in less frequent communication of gradients between steps. Another impact of this feature is the ability to train with larger batch sizes per GPU. When specified, <i>**train_micro_batch_size_per_gpu**</i> is automatically calculated using <i>**train_batch_size**</i> and number of GPUs. Should not be concurrently specified with <i>**train_micro_batch_size_per_gpu**</i> in the configuration JSON. | `1`     |
```
tjruwase, line 27: Please can you adjust to read as:

When specified, **train_step_batch_size** is automatically calculated using **train_micro_batch_size_per_gpu** ...

wbuchwalter: Hi @tjruwase,

Not sure I follow, `train_step_batch_size` doesn't exist anywhere else in the doc and is not a valid config key as far as I can tell.
Only `train_batch_size` (total batch size for an optimizer step) and `train_micro_batch_size_per_gpu` (batch size for a single gpu at a single micro step), are valid config keys.

That specific sentence is saying that if `gradient_accumulation` is specified then `train_micro_batch_size_per_gpu` will be automatically calculated (`train_micro_batch_size_per_gpu = train_batch_size/gradient_accumulation/num_gpus`).

So it seems correct?
Let me know if I got that wrong.

tjruwase: @wbuchwalter, sorry for not being clear. By the way, thanks for tackling this tricky portion of docs. 

You are correct that `train_step_batch_size` does not exist. My suggestion was only about the sentence you highlighted. Your fix only partially fixes the broken documentation in term of what the [code ](https://github.com/microsoft/DeepSpeed/blob/6124eb37c07dae57f6da01eb07f708fd6e8d4d25/deepspeed/runtime/config.py#L689) does. Given any two out of these three parameters, we can compute the third as follows:

1. Given `train_batch_size` and `train_micro_batch_size_per_gpu`, compute `gradient_accumulation_steps`: [here](https://github.com/microsoft/DeepSpeed/blob/6124eb37c07dae57f6da01eb07f708fd6e8d4d25/deepspeed/runtime/config.py#L702)

2. Given `train_batch_size` and `gradient_accumumation_steps`, compute `train_micro_batch_size_per_gpu`: [here](https://github.com/microsoft/DeepSpeed/blob/6124eb37c07dae57f6da01eb07f708fd6e8d4d25/deepspeed/runtime/config.py#L709)

3. Given train_micro_batch_size_per_gpu and gradient_accumulation_steps, compute train_batch_size: [here](https://github.com/microsoft/DeepSpeed/blob/6124eb37c07dae57f6da01eb07f708fd6e8d4d25/deepspeed/runtime/config.py#L715)

My suggestion was to adjust to emphasize (3) rather and (2) above.

wbuchwalter: I see, so then the docs needs more changes than I thought: The doc is saying that `train_batch_size` is required, which is not true given (3) as you highlighted.
It is also saying that `train_micro_batch_size_per_gpu` and `gradient_accumulation` should not be specified together, which is also not true.

I will update to reflect that and explain (1) (2) and (3)

wbuchwalter:
@tjruwase Please check updated version.
Also I'm not quite sure how to test that I didn't break the Jekyll formatting.

tjruwase (approved):

tjruwase:
> @tjruwase Please check updated version.
> Also I'm not quite sure how to test that I didn't break the Jekyll formatting.

Thanks so much for doing this. The Jekyll formatting looks fine. I will merge.

PR #1067, opened 2021-05-12
ZeRO2-Offload: Load balance gradient copying to CPU

tjruwase:
During backward in ZeRO-2 Offload, reduced gradients are accumulated into CPU memory for later optimizer step. Due to the previous model partitioning scheme, gradient copying to CPU occurs one rank at a time, which slows down backward and under-utilizes PCIe. This PR introduces a new model partitioning scheme that spreads gradient copying evenly among all (most) ranks at any point in time. This improves backward time and PCIe utilization.

### DIFF
--- a/deepspeed/runtime/zero/stage2.py
+++ b/deepspeed/runtime/zero/stage2.py
@@ -191,6 +191,9 @@ class FP16_DeepSpeedZeroOptimizer(object):
         self.all_reduce_print = False
         self.dtype = self.optimizer.param_groups[0]['params'][0].dtype
 
+        self.round_robin_fp16_groups = []
+        self.round_robin_fp6_indices = []
+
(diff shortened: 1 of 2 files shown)


eltonzheng (approved):

samyam (commented):
This is quite clever. :) 

Just documenting what we discussed on the potential simplification by just re-ordering the parameters in a round robin fashion directly in the self.fp16_groups to avoid all the book-keeping.

tjruwase:
> This is quite clever. :)
> 
> Just documenting what we discussed on the potential simplification by just re-ordering the parameters in a round robin fashion directly in the self.fp16_groups to avoid all the book-keeping.

My initial attempt at doing this hurt performance. So I am making it a TODO for when I have more time to investigate.

PR #1068, opened 2021-05-13
Add memory breakdown to training stages

cli99:
This PR uses the memory_breakdown field read from the deepspeed config file to enable memory profiling at different training stages.


cli99:
shall we add a field to enable/disenable the memory profiling in the zero implementations as well?

PR #1072, opened 2021-05-13
Seeded unit tests

ShadenSmith:
Use `pytest-randomly` to seed our unit tests in a deterministic way. By default we'll use the system time and the seed is included in each testing log. We can use a fixed seed by supplying `--randomly-seed=SEED` to `pytest`


jeffra (approved):

ShadenSmith:
Seeding works! Merging.

image

Issue #1070, opened 2021-05-13
AsyncIO Error with Stage 3 + NVME offload


stephenrawls:
Hi,

When trying to use ZERO Stage 3 with NVME offloading, which is required for fitting large models into memory, I am seeing the following error:
```
/nvme/zero_stage_3/fp16params/rank24/0_param.tensor.swp: buffer nbytes != file bytes 28824000 != 28311552
python: /usr/local/lib/python3.6/dist-packages/deepspeed/ops/csrc/aio/py_lib/deepspeed_py_aio_handle.cpp:223: int deepspeed_aio_handle_t::pread(const at::Tensor&, const char*, bool, bool): \
Assertion `static_cast<long long int>(buffer.nbytes()) == num_file_bytes' failed.
```
I have inserted some debug `print()` statements at both the write and read python call sites, here: https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/swap_tensor/utils.py#L19-L26

And I observe that when I am writing, the tensor really is trying to write `28824000` bytes (because it has `14412000` elements). However when I do an `ls -l /nvme/zero_stage_3/fp16params/rank24/0_param.tensor.swp` I observe that the file only has `28311552` bytes as mentioned in the error message. So it seems that somehow the async write command is failing to properly write the full contents of the tensor.

Any idea why this would happen? Or suggestions for how to debug further?

I have tried looking at kernel logs via `dmesg`, but nothing turns up. I have also tried running program with `strace -e trace=io_submit,io_getevents,io_setup`, but I only see the `io_setup` syscall and not the `io_submit` or `io_getevents` syscalls.

I do have libaio_dev installed.

My deepspeed config looks like this:
```
zero_optimization:
  stage: 3
  stage3_prefetch_bucket_size: 1e9
  stage3_param_persistence_threshold: 1e6
  stage3_max_live_parameters: 1e9
  overlap_comm: true
  contiguous_gradients: true
  offload_param:
      device: nvme
      nvme_path: /nvme
      pin_memory: false
      max_in_cpu: 1e9
      buffer_size: 1e9
      buffer_count: 5
    offload_optimizer:
      device: nvme
      nvme_path: /nvme
      pin_memory: false
```

Thanks,
Stephen

tjruwase:
@stephenrawls , thanks for reporting this issue.  Can you share details for us to repro this issue?

tjruwase:
[...]

PR #1086, opened 2021-05-20
ZeRO-Infinity: Swap into unaligned fp16 buffer

tjruwase:
Support nvme swap into unaligned fp16 param buffer. Fix for #1070.

Significant credit to @stephenrawls for bug discovery and triaging.


deepspeed/runtime/swap_tensor/partitioned_param_swapper.py
```diff
@@ -294,21 +305,43 @@ def swap_in(self, params, async_op=True, swap_in_buffers=None):
...
+            inflight_numel = dest_buffer.numel()
+
+        swap_in_paths = self._get_swap_paths([param])
+
+        swap_in_tensors(self.aio_read_handle, swap_in_buffers, swap_in_paths)
+        self._update_inflight_swap_in([param], swap_in_buffers, inflight_numel)
+        self.synchronize_reads()
+
+        if require_swap_buffer:
+            dest_buffer.data.copy_(param.ds_tensor.data)
```
eltonzheng, line 341: if dest_buffer is not pinned in memory, should we pin it here or doesn't matter?

eltonzheng (approved):

deepspeed/runtime/swap_tensor/partitioned_param_swapper.py
tjruwase: Good question. Since we don't own `dest_buffer`, we leave it to the caller to decide if they want to pin or not.

PR #1100, opened 2021-05-24
delay imports for replace policies and fix missing req

jeffra:
- [x] Add requirement for dev on importlib-metadata > 4, we were seeing pytest issues with importlib-metadata 2.0
- [x] Delay import of static class variables, this was causing our unit test framework to crash since importing megatron seems to initialize cuda which causes us python/pytorch multi-process issues :(
- [x] fix incorrect import path for ParallelTransformerLayer

### DIFF
--- a/deepspeed/module_inject/replace_module.py
+++ b/deepspeed/module_inject/replace_module.py
@@ -387,6 +387,9 @@ def replace_module(model, orig_class, replace_fn, _replace_policy):
         policy.update({orig_class: (replace_fn, _replace_policy)})
     else:
         for plcy in replace_policies:
+            # instantiate a throw-away policy in order to populate the _orig_layer_class
+            _ = plcy(None)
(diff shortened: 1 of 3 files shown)


RezaYazdaniAminabadi (approved):

deepspeed/module_inject/replace_module.py
```diff
@@ -387,6 +387,9 @@ def replace_module(model, orig_class, replace_fn, _replace_policy):
         policy.update({orig_class: (replace_fn, _replace_policy)})
     else:
         for plcy in replace_policies:
+            # instantiate a throw-away policy in order to populate the _orig_layer_class
+            _ = plcy(None)
+            assert plcy._orig_layer_class != None
```
RezaYazdaniAminabadi, line 392: that's great! Thanks Jeff :)

RezaYazdaniAminabadi (approved):

PR #1102, opened 2021-05-25
Refactor to use local vars for rank, world_size, group.

awan-10:
This is a simple refactor for the parameter partitioning code to use local vars instead of self.* variants.


deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -885,6 +894,10 @@ def _allgather_params(self, param_list, hierarchy=0):
         return None
 
     def _reduce_scatter_gradients(self, param_list):
+        rank = self.rank
```
samyam, line 897: This one is fine for now but will probably need to be changed and placed inside the loop on 905, since this is going over a list of params

samyam (approved):

deepspeed/runtime/zero/partition_parameters.py
awan-10: Changed and fixed in the new commit (1f4a749)

Microsoft-code (approved):

raghavthind2005 (approved):

rocm-mici:

PR #1135, opened 2021-06-05
[zero] fix missed subclasses partitioning bug

stas00:
Currently any params/buffers created in  `__init__` of a subclass that is not a subclass of `nn.Module` are missed by the partitioning logic and so any model having such classes breaks under deepspeed. I think this is primarily `zero.Init` case. 

This PR recursively gets all classes and now the afflicted models work.

@samyam


samyam (approved):
Thank you for identifying and fixing this bug @stas00

PR #1150, opened 2021-06-09
Samyamr/largest partitioned params calculation fix

samyam:
largest_partiitoned_params is calculated incorrectly making it much larger than it has to be. The PR  fixes the calculation.


deepspeed/runtime/zero/stage3.py
```diff
@@ -852,7 +852,7 @@ def __init__(self,
                 count = count + 1
 
         #Largest partitioned param
-        largest_partitioned_param_numel = max(self.fp16_partitioned_groups_flat_numel)
+        largest_partitioned_param_numel = max([max([tensor.numel() for tensor in fp16_partitioned_group]) for fp16_partitioned_group in self.fp16_partitioned_groups])
```
tjruwase, line 855: Does self.fp16_partitioned_groups_flat_numel need to be updated?

tjruwase (approved):

deepspeed/runtime/zero/stage3.py
tjruwase: Do references like [this ](https://github.com/microsoft/DeepSpeed/blob/5f16b17297525b321208b493e89c665e68a04163/deepspeed/runtime/zero/stage3.py#L931) need updating?

mrgomdev:
The change uses tensor.numel() for retrieving the sizes. However, if we already built the model in the context of deepspeed.zero.Init(), tensor.numel() or tensor.data get pointing only the placeholder, after partitioning and offloading the original tensor into ds_tensor and ds_numel, accordingly. In some cases, this can cause some problems, including create an deepspeed engine after deepspeed.zero.Init(). Please check out this situation and confirm the change is valid.

PR #1151, opened 2021-06-09
Add Windows support in README, use c++17 on Windows  to support latest VC & cuda build tool

eltonzheng:


README.md
```diff
@@ -96,6 +96,12 @@ If you would like to pre-install any of the DeepSpeed extensions/ops (instead
 of JIT compiling) or install pre-compiled ops via PyPI please see our [advanced
 installation instructions](https://www.deepspeed.ai/tutorials/advanced-install/).
 
+On Windows you can build wheel with following steps, currently only inference mode is supported.
+1. Install pytorch, such as pytorch 1.8 + cuda 11.1
+2. Install latest visual cpp build tool
```
jeffra, line 101: Is there a version that we've used for this? "latest" will probably change over time so it might be nice to tell folks which version we know works.

jeffra (approved):

README.md
eltonzheng: cool, will add detailed version info here.

PR #1154, opened 2021-06-11
Avoid partitioning small activations

tjruwase:
Avoid partitioning activations that are smaller than MP size.


deepspeed/runtime/activation_checkpointing/checkpointing.py
```diff
@@ -416,7 +416,7 @@ def save_args_for_backward(*all_args):
 
             inputs = []
             for i, item in enumerate(args[:-1]):
-                if not torch.is_tensor(item):
+                if not torch.is_tensor(item) or mp_size > item.numel():
```
ShadenSmith, line 419: The logic looks good to me :-). We may want to revisit sometime with a partitioning threshold, similar to the Z3 engine.

ShadenSmith (approved):
LGTM. We may want to revisit sometime with a partitioning threshold, similar to the Z3 engine.

Issue #663, opened 2021-01-12
Issue with deepspeed installation through pip requirements


baijumeswani:
I am trying to install deepspeed using pip through a requirements.txt file.

Contents of the requirements.txt file:
```
--pre
-f https://download.pytorch.org/whl/torch_stable.html
torch==1.6.0+cu101
torchvision==0.7.0+cu101
torchtext==0.7.0
deepspeed
```

When I try to run ```pip install -r requirements.txt```, I get the error:
```
ERROR: Command errored out with exit status 1:
     command: /home/bmeswani/anaconda3/envs/test/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-mr3spwnm/deepspeed_f25fb2385d494aa3a282ec5481159c40/setup.py'"'"'; __file__='"'"'/tmp/pip-install-mr3spwnm/deepspeed_f25fb2385d494aa3a282ec5481159c40/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-v85pfl_r
         cwd: /tmp/pip-install-mr3spwnm/deepspeed_f25fb2385d494aa3a282ec5481159c40/
    Complete output (12 lines):
    Traceback (most recent call last):
      File "/tmp/pip-install-mr3spwnm/deepspeed_f25fb2385d494aa3a282ec5481159c40/setup.py", line 18, in <module>
        import torch
    ModuleNotFoundError: No module named 'torch'
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-mr3spwnm/deepspeed_f25fb2385d494aa3a282ec5481159c40/setup.py", line 21, in <module>
        raise ImportError('Unable to import torch, please visit https://pytorch.org/ '
    ImportError: Unable to import torch, please visit https://pytorch.org/ to see how to properly install torch on your system.
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

It would appear that ```deepspeed``` setup.py depends on ```torch``` which prevents me from installing ```torch``` and ```deepspeed``` through a single ```requirements.txt``` file.

[...]

PR #1166, opened 2021-06-16
Remove hard torch dependency at install

jeffra:
In order to support pre-compiling our ops we require torch already be installed. However, when projects want to depend on deepspeed in a requirements file in many cases they are not able to do so (see issue #663). This PR allows changes this requirement to only be enforced if the user explictly tries to pre-compile any of the deepspeed ops.


ShadenSmith (approved):
I think we can also update the docs here? https://github.com/microsoft/DeepSpeed/blame/master/docs/_tutorials/advanced-install.md#L14

PR #1167, opened 2021-06-17
fix checkpoint loading

conglongli:
When user calls load_checkpoint with load_optimizer_states/load_lr_scheduler_states as False, we shall not include the optimzer/lr_scheduler as deepspeed_states, and instead return them to user as part of client_state.


deepspeed/runtime/engine.py
```diff
@@ -1721,14 +1721,16 @@ def _load_checkpoint(self,
...
             'module',
-            'optimizer',
-            'lr_scheduler',
             'csr_tensor_module_names',
             'skipped_steps',
             'global_steps',
             'dp_world_size',
             'mp_world_size'
         ]
+        if load_lr_scheduler_states:
```
tjruwase, line 1730: Can you clarify in which scenarios would this be correct. Our contract is that client_state would contain exactly what the user passed in `save_checkpoint(...)`.

ShadenSmith: Hi @tjruwase , this change was useful for a Megatron variant in which the LR scheduler's `load_state_dict()` was modified to require additional arguments from the client. We don't have a way to provide those args and this gives us a way to defer loading to the client.

jeffra:
Already merged in different PR

/cc @conglongli

PR #1169, opened 2021-06-17
add assert to make sure zero isn't enabled with 1bit-*

jeffra:


conglongli (approved):

jromagn:
Is compatibility planned/in the works?

Issue #1069, opened 2021-05-13
Performance Degradation with ZERO Stage 3


stephenrawls:
Hi,

I am trying to benchmark a 10B parameter Huggingface `RobertaForMaskedLM` model with both ZERO Stage 2 and ZERO Stage 3 to compare the latency impact of parameter partitioning.

I am seeing _much worse_ performance with Stage 3 than expected however, so want to check if something looks wrong.

```
|--------------------+-------+--------+-------+---------+--------|
| Description        | Model | # p3dn | batch | Samples | TFLOPS |
|                    | Size  |  hosts |  size | Per Sec |  / GPU |
|--------------------+-------+--------+-------+---------+--------|
| Baseline (stage2)  | 10B   |     16 |     8 |     178 |   44.7 |
| Stage3, no offload | 10B   |     16 |     8 |      77 |   19.5 |
| Stage3, no offload | 10B   |      8 |     8 |      41 |   21.9 |
| Stage3, no offload | 10B   |      4 |     8 |      23 |   23.5 |
| Stage3, no offload | 10B   |      2 |     8 |    11.6 |   23.5 |
| Stage3, no offload | 10B   |      1 |     8 |     OOM |    OOM |
|--------------------+-------+--------+-------+---------+--------|
```
The problem does not seem to be related to network bandwidth, because when I move to p4d machines, which have 4x the bandwidth of p3dn machines (400 Gbps vs 100 Gbps) I see similar degradation:
```
|--------------------+-------+--------+-------+---------+--------|
| Description        | Model | # p4dn | batch | Samples | TFLOPS |
|                    | Size  |  hosts |  size | Per Sec |  / GPU |
|--------------------+-------+--------+-------+---------+--------|
| Baseline (stage2)  | 10B   |     16 |     8 |     432 |    109 |
| Stage3, no offload | 10B   |      4 |     8 |      44 |   44.5 |
|--------------------+-------+--------+-------+---------+--------|
```

I tried increasing `stage3_max_live_parameters` from 1e9 → 2e9, and `stage3_prefetch_bucket_size` from 5e8 → 10e8, but neither change impacted performance.

In addition, I ended up adding some `time.time()` statements before + after:
  a. The blocking `fetch()` call: https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/zero/stage3.py#L1520
[...]

PR #1170, opened 2021-06-17
ZeRO-3 Slowdown

tjruwase:
Fix significant zero-3 slowdown compared to zero-2, based on #1069.
Two key parts of this PR:
1) Fix hook registration problem that created off-by-one tracing issue
2) Move tracing book-keeping to parameter partition logic.


deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -1043,6 +1083,34 @@ def _partition_gradient(self, param, partition_buffer=None, accumulate=False):
...
+        if new_status == ZeroParamStatus.AVAILABLE:
+            self.avaliable_parameter_numel += numel
+
+    def _get_available_parameter_numel(self):
+        return self.avaliable_parameter_numel
+
+    def _synchronize_communication(self, param_list, handle_list):
+        for param, handle in zip(param_list, handle_list):
+            if handle is not None:
+                handle.wait()
```
zarzen, line 1110: `handle.wait()` ensures the communication task is enqueued, but not neccessarily completed. 
Based on the section "Synchronous and asynchronous collective operations" of [torch.distributed](https://pytorch.org/docs/stable/distributed.html) 

So we probably need other sync mechanism.

deepspeed/runtime/zero/stage3.py
```diff
@@ -520,11 +556,25 @@ def _all_gather(self, partitioned_params, async_op=False):
...
-            if handle is not None:
-                with torch.cuda.stream(self.comm_stream):
-                    handle.wait()
-            param.ds_status = ZeroParamStatus.AVAILABLE
+
+        if len(self.params_in_flight) > 0:
+            avail_params = [param.ds_id for param in self.params_in_flight]
+            status_params = [param.ds_status for param in self.params_in_flight]
+
+            with torch.cuda.stream(self.comm_stream):
```
zarzen, line 564: The `with torch.cuda.stream(self.comm_stream)` statement is not going to take effect on the underlying *communication* operations, as far as i know. The reason is that the `torch.distributed.all_gather` uses its own cuda stream. 

Code reference point:
https://github.com/pytorch/pytorch/blob/v1.9.0/torch/lib/c10d/ProcessGroupNCCL.cpp#L1111

deepspeed/runtime/zero/partition_parameters.py
jfc4050: there's been a lot of discussion about this here https://github.com/microsoft/DeepSpeed/issues/1069#issuecomment-870185380

the `handle.wait()` will make sure no more work happens on current stream until the collective has completed so for this to be safe it either needs to be consumed in the same stream or synchronized.

@zarzen calls out below that the call is made from `self.comm_stream`, i think if we get rid of that (do the all gather from default stream so that kernel launches on default stream dont progress until the allgather completes) we won't need the additional synchronize. Like he mentions below it seems to independently decide to use its own stream so the actual allgather work isn't being done in `comm_stream` anyways

tjruwase: @jfc4050 and @zarzen, thanks for your deep investigation of this issue. I am closely following your updates on #1069. 

Thanks for your explanation, I now see the value of eliminating `self.comm_stream` and doing the `handle.wait()` in default stream. I will try this out.

deepspeed/runtime/zero/stage3.py
tjruwase: @zarzen, thanks. I will try this. Also, I see some great progress in your PR #1188. Can you see our we can merge both efforts?

tjruwase: @zarzen, that sounds good. I was not suggesting merging into a single PR. What I meant was to avoid conflicts and duplications in both efforts. For now, we don't need to do anything but just be aware of the relationship of both efforts. Thanks.

deepspeed/runtime/zero/partition_parameters.py
jfc4050: np! thinking about this a little more, there could be a small amount of value to doing allgather in the comm_stream, but not for parallelizing communication - there's some memcpy operations that get done as part of allgather that could potentially be overlapped with forward pass computation and parameter gathering communication.

Doing this means that the streams would have to be carefully synchronized where appropriate though, so might not be worth the extra complexity, hopefully we can find lower hanging fruit

zarzen:
Hi @tjruwase , 
I am recently merging your modifications with the customized allgather I introduced in another PR #1188 . 
I placed the code in my personal repo at here:  [zhen/zero3_debug](https://github.com/zarzen/DeepSpeed/tree/zhen/zero3_debug)

Basically, I replaced the torch allgather with my own allgather. https://github.com/zarzen/DeepSpeed/blob/zhen/zero3_debug/deepspeed/runtime/zero/partition_parameters.py#L677_L683

I can see performance improvements, especially for the backward part. 

While after analyzing the timeline I found two issues. let me bring the timeline at here:
image

Two issues: 
1. the prefetching only partially overlaps with computation. 
It would be nice to see the next-next module also prefetched. (As denoted in green circle)
2. there is a pretty large gap between the computation of current module and next-fetch-module operation
Which I denoted in red box. 

For the first issue, I think the it is mostly due to the condition check in the `prefetch_next_sub_modules`:
[def prefetch_next_sub_modules(self,...](https://github.com/zarzen/DeepSpeed/blob/21d6c36ee69f49ec4ddafbed7c3740131ee5bb48/deepspeed/runtime/zero/stage3.py#L349-L351 )

Do you have idea the timeline gap of second issue?
And for the first issue, do you have some suggestions for launching the next-next prefetching earlier?

tjruwase:
> 1. the prefetching only partially overlaps with computation.
>    It would be nice to see the next-next module also prefetched. (As denoted in green circle)

I think we to consider that prefetching increases memory consumption, at the extreme we lose the benefit of parameter partitioning. This is my guess for why we restrict next-next module prefetching. Of course this depends on the model size and on the exposed latency of parameter gather on the critical path. Perhaps @samyam can chime in.

tjruwase:
> 2\. there is a pretty large gap between the computation of current module and next-fetch-module operation
>     Which I denoted in red box.

Nothing comes to mind right away for why this is happening. Is this timeline showing for the entire forward or for individual module computation?

zarzen:
> > 1. the prefetching only partially overlaps with computation.
> >    It would be nice to see the next-next module also prefetched. (As denoted in green circle)
> 
> I think we should consider that prefetching increases memory consumption, at the extreme we lose the benefit of parameter partitioning. This is my guess for why we restrict next-next module prefetching. Of course this depends on the model size and on the exposed latency of parameter gather on the critical path. Perhaps @samyam can chime in.

agree, memory consumption is definitely an important consideration. while I saw there is a inflight size checking for param gathering. that param suppose to limit the memory usage. but for now the size of a module is smaller than the `stage3_prefetch_bucket_size` is not going to work. 

> > 2. there is a pretty large gap between the computation of current module and next-fetch-module operation
> > Which I denoted in red box.
> 
> Nothing comes to mind right away for why this is happening. Is this timeline showing for the entire forward or for individual module computation?

this timeline is for two consecutive modules, not the entire timeline for forward. 
I can share the profiling file if you would like to look into details.

zarzen:
Hi @tjruwase  @samyam 
After detailed analysis, I found the source of small cudaMemcpy operations. 
Every time the `release_sub_module` is invoked, the `_partition_param` function will be called, which contains a memory operation at here:
https://github.com/microsoft/DeepSpeed/blob/3c4a94924ec1474afa2b78a9ccfcfb5690dc4f49/deepspeed/runtime/zero/partition_parameters.py#L729

I would recommend to replace it with the following:
```python
param.data = torch.empty(1, dtype=self.dtype, device=param.device)
```
Thus, it can avoid the memory movement, besides, the `element_wise_kernel` won't be launched for setting the memory. It would reduce the forward time from ~180ms to ~150ms in my case. 

For the pipelining issue, I found the `prefetch_next_sub_modules` takes about ~500us, in my setup. 
I found even though the param gathering operation is launched in async way. The other parts take significant amount of time. 
I would like to place the `prefetch_next_sub_modules` function call into a separate thread. But I am not sure whether it is going to break exist deisgn. Could you comment a little bit?

Here is the current time:
image

You can see the computation kernels of next module is delayed by the `prefetch_next_sub_modules` function call.

tjruwase:
> I would recommend to replace it with the following:
> 
> ```python
> param.data = torch.empty(1, dtype=self.dtype, device=param.device)
> ```

Please go ahead with this change. Thanks!

tjruwase:
> For the pipelining issue, I found the `prefetch_next_sub_modules` takes about ~500us, in my setup.
> I found even though the param gathering operation is launched in async way. The other parts take significant amount of time.
> I would like to place the `prefetch_next_sub_modules` function call into a separate thread. But I am not sure whether it is going to break exist deisgn. Could you comment a little bit?

Can you clarify if you mean separate stream or python thread? Regardless, it seems to me that making `prefetch_next_sub_modules` more asynchronous should be fine. We just need to be careful to properly coordinate with the next `fetch_sub_module`. We can also think of unit tests for the corner cases as I don't think our existing unit tests do this.

zarzen:
> > For the pipelining issue, I found the `prefetch_next_sub_modules` takes about ~500us, in my setup.
> > I found even though the param gathering operation is launched in async way. The other parts take significant amount of time.
> > I would like to place the `prefetch_next_sub_modules` function call into a separate thread. But I am not sure whether it is going to break exist deisgn. Could you comment a little bit?
> 
> Can you clarify if you mean separate stream or python thread? Regardless, it seems to me that making `prefetch_next_sub_modules` more asynchronous should be fine. We just need to be careful to properly coordinate with the next `fetch_sub_module`. We can also think of unit tests for the corner cases as I don't think our existing unit tests do this.
> 
> @samyam what do you think?

A separate python threads. because I didn't see any computation or communication on CUDA is required for `release_sub_module`. Even there are memory operations, they should probably happen on the default cuda stream, maintained by PyTorch.

tjruwase:
> Which way do you think is most convienient for the change? (since it is a very small modification, so I didn't plan for a PR at this moment. )

I think these changes can wait until your PR is ready because they won't give much benefits without your PR. What do you think?

zarzen:
> > Which way do you think is most convienient for the change? (since it is a very small modification, so I didn't plan for a PR at this moment. )
> 
> I think these changes can wait until your PR is ready because they won't give much benefits without your PR. What do you think?

Sure. we can leave it as it is for now, and it is just a one line change. Thanks!

zarzen:
Hi, is this PR and #1188 still under the plan for merging to the master?

tjruwase:
@zarzen, apologies for the delay on this. The merge was delayed because of our latest release a few weeks ago. We will now focus on reviewing and merging both PRs soon. Thanks for your patience.

deepspeed/runtime/zero/partition_parameters.py
tjruwase: That makes sense, the low benefits does not justify the complexity.

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -667,15 +695,25 @@ def _all_gather(self, param_list, async_op=False, hierarchy=None):
                     handle = self._allgather_param(param,
                                                    async_op=async_op,
                                                    hierarchy=hierarchy)
-                    param.ds_status = ZeroParamStatus.INFLIGHT  # if async_op else ZeroParamStatus.AVAILABLE
+                    param.update_status(ZeroParamStatus.INFLIGHT
+                                        if async_op else ZeroParamStatus.AVAILABLE)
```
samyam, line 699: async_op is always true in this branch. We do not need the if condition.

tjruwase:
@zarzen, my teammates have reviewed this PR and it looks good. The next step is to rerun correctness and perf tests, especially since significant code was merged into the master branch for our last release. Can you also run perf tests on your side to check that there are no regressions. I will be doing the same. Although this is a bit of work, but hopefully it would make merging #1188 easier.

zarzen:
> @zarzen, my teammates have reviewed this PR and it looks good. The next step is to rerun correctness and perf tests, especially since significant code was merged into the master branch for our last release. Can you also run perf tests on your side to check that there are no regressions. I will be doing the same. Although this is a bit of work, but hopefully it would make merging #1188 easier.

sure, I have seen you merged the master into my branch, Thanks!
I will start the regression tests ASAP, possibly complete before Friday.

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -479,8 +479,11 @@ def get_model():
 
         self._validate_remote_device(remote_device, _ds_config)
 
+        self.avaliable_parameter_numel = 0
```
stas00, line 482:
```suggestion
        self.available_parameter_numel = 0
```

and everywhere else :)

tjruwase:
Subsumed by #1453

PR #1185, opened 2021-06-24
switch bin files to use python3 instead of python

jeffra:
Up until now our launcher assumes that `env python` is py3.x. However some systems that rely on older packages that still require py2.7 need `env python` to resolve to py2.x. This PR makes sure deepspeed binaries point to py3.


ShadenSmith (approved):

tjruwase:
@jeffra, do we want a similar fix in install.sh for [python ](https://github.com/microsoft/DeepSpeed/blob/5127b2fa256209ddadaab857bba8359cee6a7c3c/install.sh#L155) and [pip](https://github.com/microsoft/DeepSpeed/blob/5127b2fa256209ddadaab857bba8359cee6a7c3c/install.sh#L143)?

PR #1188, opened 2021-06-25
ZeRO3, improved parameter all-gather operation

zarzen:
For `bing_bert` model with following configuration (about 5.1B params), forward computation improved from ~470ms to ~270ms, backward time improved from ~900ms to ~670ms. (hardware setup: 1x ec2-p4d.24xlarge instance)

``` json
    "bert_model_config": {
        "vocab_size_or_config_json_file": 32003,
        "hidden_size": 2560,
        "num_hidden_layers": 64,
        "num_attention_heads": 40,
        "intermediate_size": 10240,
        "hidden_act": "gelu",
        "hidden_dropout_prob": 0.1,
        "attention_probs_dropout_prob": 0.1,
        "max_position_embeddings": 512,
        "initializer_range": 0.02
    },
```

1) Removing the norm computation in debug printing
2) Changing _all_gather to be sync op in fetch_sub_module
    Reason: the async version is not async at all, because each
    all_gather calls torch.cuda.synchronize() to guarantee previous
    communication op to be completed
3) Adding new function `_allgather_params_coalesced`
    the existing _allgather_params has explicit memcpy after the
    all-gather op. We can avoid the explicit memory copy at
    python side, to improve the performance.
4) changed `_partition_param` with `torch.empty` function. 

Notes:
Using most recent updates on PyTorch `_all_gather_base` function could get further performance boost. 
As the `_all_gather_base` function avoids the redundant memory copy. Refer to https://github.com/pytorch/pytorch/pull/56315


zarzen:
@jfc4050 @tjruwase 
I have pushed the customized all_gather operation. 
The op uses the cuda stream specified by `torch.cuda.stream`, and it returns a handle with wrapped cuda event. so you can query/synchronize/wait of the communication on the given stream. 

The performance of micro-benchmark shows about 1.1ms time reduction for allgathering a transformer layer with 9.8M params (half-precision) on each partition on a p4d.24xl instance. 

In end to end training, the forward time could be further reduced to 200ms for 5.1B bing-bert model. (previously forward time around 280ms-300ms)

zarzen:
Hi,
I would like to remove the customized c++ extension for all-gather operation, due to the hanging issue I observed using customized all-gather wrapper. and left the python part.

The reason for hanging: the customized all-gather is running on a separate cuda-stream, compared to `torch.distributed.` APIs. Thus, it is easy to have two nccl-communicators working, which cause potential hanging.

tjruwase:
@zarzen, thanks for the update. Do you have an idea of the perf impact? Have you pushed the changes into this PR? Also, is this PR ready for review or are you still testing?

tjruwase:
@zarzen, another thought is whether the in-place allgather should be PR'd into pytorch?

zarzen:
@tjruwase for the performance impact, the end-to-end training probably drops ~80-100ms in forward and backward, in my setup. I haven't benchmarked yet with all other optimizations I can do. I was working on fixing the hanging issue. 
The hanging issue only shows up when I moved the prefetching function to a separate Python thread, so I thought the hanging might due to the size mismatching of collective operations. 

I haven't pushed the change yet. I will do the change probably by the end of this week. 

I agree the `in place all-gather` implementation is better to go Pytorch. 
Thus, I submitted an issue here https://github.com/pytorch/pytorch/issues/61781 and drafted an implementation. 

The master branch of PyTorch has an in-place all-gather, called `_all_gather_base`, while it only supports a single tensor as output. That's why I am suggesting having a list of tensors as outputs tensor for the `_all_gather_base` function.

zarzen:
Hi, I have removed the c++ extension, this is ready for review.

zarzen:
Their are 15 failed unit tests
```
FAILED tests/unit/test_checkpointing.py::test_checkpoint_unfused_optimizer
FAILED tests/unit/test_config.py::test_batch_config[2-32-16-1-True]
FAILED tests/unit/test_config.py::test_batch_config[2-32-18-1-False]
FAILED tests/unit/test_configurable_parallel.py::TestConfigurableMP::test_gpt2_basic
FAILED tests/unit/test_configurable_parallel.py::TestConfigurableMP::test_gpt2_mp2_no_resize
FAILED tests/unit/test_configurable_parallel.py::TestConfigurableMP::test_gpt2_mp_2to1
FAILED tests/unit/test_configurable_parallel.py::TestConfigurableMP::test_gpt2_mp_2to4
FAILED tests/unit/test_configurable_parallel.py::TestConfigurablePP::test_pp_basic
FAILED tests/unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_mp1_pp_2to1
FAILED tests/unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_mp1_pp_2to4
FAILED tests/unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_mp2_pp_2to1
FAILED tests/unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_mp2_pp_1to2
FAILED tests/unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_pp_2to1_mp_2to1
FAILED tests/unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_pp_1to2_mp_1to2
FAILED tests/unit/test_topology.py::test_stage_to_global
```

Failures in `test_configurable_parallel.py` all due to `megatron` import error. 
Seems not relevant to this PR

tjruwase:
Yes, you can ignore those unit tests. We need to skip them.

zarzen:
I have did perf test, but on V100 GPU rather than A100s, due to the aws capacity issue. 
I can still see about ~300ms reduction in forward pass and ~250ms reduction in backward.

tjruwase:
What model size?

zarzen:
> What model size?
~5.1B params
```
    "train_batch_size": 512, 
    "train_micro_batch_size_per_gpu": 8, 
    "steps_per_print": 100, 
    "prescale_gradients": false, 
    "bert_token_file": "bert-base-uncased", 
    "bert_model_config": {
        "vocab_size_or_config_json_file": 3.200300e+04, 
        "hidden_size": 2.560000e+03, 
        "num_hidden_layers": 64, 
        "num_attention_heads": 40, 
        "intermediate_size": 1.024000e+04, 
        "hidden_act": "gelu", 
        "hidden_dropout_prob": 0.1, 
        "attention_probs_dropout_prob": 0.1, 
        "max_position_embeddings": 512, 
        "initializer_range": 0.02
    }, 
    "zero_optimization": {
        "stage": 3, 
        "overlap_comm": true, 
        "reduce_scatter": true, 
        "contiguous_gradients": true
    }, 
    "optimizer": {
        "type": "Adam", 
        "params": {
            "lr": 0.0001, 
            "weight_decay": 0.01, 
            "bias_correction": true, 
            "eps": 1e-06
        }
    }, 
    "gradient_clipping": 1.0, 
    "wall_clock_breakdown": true
```

zarzen:
Confirmed the effectiveness of groupped all-gather ops. 

## timeline of master branch in forward pass
image

## timeline of this PR in forward pass
image

zarzen:
Hey @tjruwase I saw the code failed due to `stderr: RuntimeError: Tensors must be CUDA and dense`. is there anything I need to do for the fix?

jeffra:
> Hey @tjruwase I saw the code failed due to `stderr: RuntimeError: Tensors must be CUDA and dense`, in `nv-transformers-v100` section. is there anything I need to do for the fix?

I do see another error message in this test before this dense related one:

```
RuntimeError: Invalid function argument. Expected parameter `tensor_list` to be of type List[torch.Tensor].
```

Are any of your new `all_gather` calls passing a single tensor instead of a list of tensors? I haven't looked over all the proposed code here but maybe the line I added a comment to.

tjruwase:
@zarzen, thanks for following up and sorry that I forgot to update you. Yes, I was able to fix the issue. The problem was that 2 zero context objects were constructed along the way and a parameter that was gathered by one context was partitioned in the other. The fix is to avoid multiple zero contexts, but rather to reuse the existing context to register any newly discovered parameter. The core of the fix is [here](https://github.com/microsoft/DeepSpeed/blob/f4cebc0aa81041654895c16ac7b796ad3306f36e/deepspeed/runtime/zero/stage3.py#L956). 

We are fortunate that the HF unit tests was able to expose this issue. I will take a closer look at your unit test failures as well.

zarzen:
> @zarzen, thanks for following up and sorry that I forgot to update you. Yes, I was able to fix the issue. The problem was that 2 zero context objects were constructed along the way and a parameter that was gathered by one context was partitioned in the other. The fix is to avoid multiple zero contexts, but rather to reuse the existing context to register any newly discovered parameter. The core of the fix is [here](https://github.com/microsoft/DeepSpeed/blob/f4cebc0aa81041654895c16ac7b796ad3306f36e/deepspeed/runtime/zero/stage3.py#L956).
> 
> We are fortunate that the HF unit tests was able to expose this issue. I will take a closer look at your unit test failures as well.

Are you refer to this commit, [a75e46](https://github.com/microsoft/DeepSpeed/pull/1170/commits/a75e465cab75653a5bfad1c70d72f464f9df9335), for fixing the multi-context issue? 

does that mean I can wait #1170 get merged first?

tjruwase (approved):

PR #1202, opened 2021-07-02
[zero.Init] post_init partitining is to be run only by a child module

stas00:
Originally the zero partitioning `post_init` was run only by the top `super` class, missing out on the children. Then we fixed it to run on all classes. Which was still insufficient. A parent creating weights, and child wanting to modify these weights now had to gather the weights. Moreover `post_init` was run multiple times on the same object.

This PR changes the wrapper logic to run `post_init` only once per object and do it only when child's `__init__` has finished. Thus no longer requiring the child to do any gathering.

Thanks to @samyam for suggesting how to tell the child from its parents.


samyam (approved):
Thank you @stas00 for this fix

PR #1181, opened 2021-06-23
[model weights] zero_to_fp32 multiple improvements

stas00:
As requested at https://github.com/microsoft/DeepSpeed/issues/1062#issuecomment-865994154 for small models it's too awkward to go hunt for the checkpoint folder, run commands manually and go back to the software to use it.

So this PR:

- [x] provides a similar API to do it on the fly:
```
from deepspeed.utils.zero_to_fp32 import get_fp32_state_dict_from_zero_checkpoint
state_dict = get_fp32_state_dict_from_zero_checkpoint(checkpoint_dir) # already on cpu
model = model.cpu()
model.load_state_dict(state_dict)
#submit to model hub or save the model to share with others
```
so all you need here is the path to the pytorch checkpoint dir.

The only possible culprit here is that the model will no longer be useable in the deepspeed context of the same application. i.e. you will need to re-init the deepspeed engine, since `model.load_state_dict(state_dict)` will remove all the deepspeed magic from it.

And of course the main concern here is that the application will not have enough memory to accommodate this process.

- [x] provides an even simpler API to do it on the fly:

```
from deepspeed.utils.zero_to_fp32 import load_state_dict_from_zero_checkpoint
model = load_state_dict_from_zero_checkpoint(trainer.model, checkpoint_dir)
# now save it and push to hub
# note the model is no longer deepspeed-usable and requires a new deepspeed initialize if you want to continue using it right away
```

- [x] Also note, I changed the API to find the deepspeed sub-dir automatically so all you need is the path to the best or last normal checkpoint. i.e. you don't need the `global_step2` in the path.

Here are a few additional unrelated fixes:
- [x] in stage3 added a detector for params that didn't get partitioned
- [x] rewrote `_get_zero_param_shapes` to rely on `optimizer.fp16_groups` as `module.named_parameters` was completely unreliable
- [x] added a fix to copy `zero_to_fp32.py` only once
- [x] added a check that `len(optim_files)==world_size` - had a user who was missing one optim file.
- [x] multiple other sanity checks and validations and easier debug 

@piegu, please give it a try and let me know if it makes things easy for you.

I'm totally open to any suggestions to make it better.

@samyam, @jeffra, @tjruwase

PR #1212, opened 2021-07-08
only create recovery script from rank 0

adammoody:
While saving a zero checkpoint, it seems that every process copies the ``zero_to_fp32.py`` recovery script into the checkpoint directory.  One then gets the following message from every process:
```
[2021-07-08 14:59:35,558] [INFO] [engine.py:1938:_copy_recovery_script] creating recovery script /path/to/checkpoint/zero_to_fp32.py
```
This tweaks things so that only global rank 0 does that task.


tjruwase:
Turns out this fix is already in #1181. Thanks!

PR #1224, opened 2021-07-13
enable async io op on powerpc architectures

adammoody:
Similar to the changes made for the CPU Adam op, this updates the asynchronous I/O op to build on PowerPC systems by using ``-mcpu=native`` instead of ``-march=native`` and by guarding the includes for ``cpuid.h`` and ``x86intrin.h``.

This additionally moves the ``simd_width`` and ``cpu_arch`` methods from ``op_builder/cpu_adam.py`` to ``op_builder/builder.py``, where both are also called from ``op_builder/async_io.py``.

Note that the ``simd_width`` method in ``builder.py`` has changed slightly in that it returns ``-D__SCALAR__`` as a fallback instead of an empty string.  This was to copy the behavior used in ``cpu_adam.py``.  I think that change should be safe, but please double check me.

A couple of open items for ``simd_width``:
- It doesn't seem like anything actually uses ``-D__SCALAR__``.
- It looks like an empty string is still returned if ``lscpu`` cannot be found.  Should that also be ``-D__SCALAR__`` to be consistent?

### DIFF
--- a/csrc/aio/py_lib/deepspeed_py_copy.h
+++ b/csrc/aio/py_lib/deepspeed_py_copy.h
@@ -6,8 +6,10 @@ Licensed under the MIT license.
 Functionality for swapping optimizer tensors to/from (NVMe) storage devices.
 */
 
+#if (__x86_64__ || __i386__)
 #include <cpuid.h>
 #include <x86intrin.h>
+#endif
 
 #include <deepspeed_aio_common.h>
 #include <stdlib.h>
--- a/op_builder/async_io.py
+++ b/op_builder/async_io.py
@@ -31,8 +31,8 @@ class AsyncIOBuilder(OpBuilder):
 
     def cxx_args(self):
         CPU_ARCH = self.cpu_arch()
-
-        args = [
+        SIMD_WIDTH = self.simd_width()
+        return [
             '-g',
             '-Wall',
             '-O0',
@@ -42,14 +42,9 @@ class AsyncIOBuilder(OpBuilder):
             '-Wno-reorder',
             CPU_ARCH,
             '-fopenmp',
+            SIMD_WIDTH,
             '-laio',
         ]
-
-        simd_width = self.simd_width()
-        if len(simd_width) > 0:
-            args.append(simd_width)
-
-        return args
 
     def extra_ldflags(self):
         return ['-laio']
--- a/op_builder/async_io.py
+++ b/op_builder/async_io.py
@@ -30,6 +30,8 @@ class AsyncIOBuilder(OpBuilder):
         return ['csrc/aio/py_lib', 'csrc/aio/common']
 
     def cxx_args(self):
+        CPU_ARCH = self.cpu_arch()
+
         args = [
             '-g',
             '-Wall',
@@ -38,7 +40,7 @@ class AsyncIOBuilder(OpBuilder):
             '-shared',
             '-fPIC',
             '-Wno-reorder',
-            '-march=native',
+            CPU_ARCH,
             '-fopenmp',
             '-laio',
         ]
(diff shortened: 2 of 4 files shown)


tjruwase:
@adammoody, thanks. 

@stas00, FYI. I recall there was an issue with empty string on some hardware.

stas00:
Yes, except it was AVX.... which was `""` on some hardware

A better solution is to change the builder to filter out any empty string options as these guarantee a problem and difficult debug time.

adammoody:
I see the async_io.py avoids appending empty strings here:

https://github.com/microsoft/DeepSpeed/blob/54bed327f55325dd517fe34c369d4ea9311c6547/op_builder/async_io.py#L46-L48

Perhaps that logic should be copied to cpu_adam, as well?

And it could be done for both CPU_ARCH and SIMD_WIDTH.

tjruwase:
@adammoody, are you able to apply the empty string filtering to this PR?

adammoody:
Yes, I'll rework this later today to filter out any empty strings.

stas00:
Probably the best long term solution would be to do it here for all sub-classes at once:
https://github.com/microsoft/DeepSpeed/blob/54bed327f55325dd517fe34c369d4ea9311c6547/op_builder/builder.py#L221

it will make the function a bit bloated, but can also wrap each of these calls to remove empty string elements in the list.

e.g. by adding a wrapper `strip_empty_entries(self.cxx_args())` (for each of these calls)

adammoody:
I pushed those changes as separate commits.  I took a first pass before I saw the later comments.  The second commit is to align with those directions.  I can rebase all of these into one commit later, but I thought it may be easier to review the incremental changes.

I found a three places where ``cxx_args`` is invoked from the builder classes.  Please let me know if I missed any or got too many.

I also rearranged the order of flags in cpu_adam to align better with async_io.  Also, please let me know if you'd like me to roll back that change to restore the original order.  While doing that, I did notice that the async op is built with ``-O0`` while cpu_adam uses ``-O3``.

adammoody:
I decided to restore the option order and rebase the last few commits to reduce the number of changes.

tjruwase (approved):

tjruwase:
@adammoody, looks good. Right now the CI is failing because of formatting issues. Please see [here](https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md) for simple steps to fix these issues. Thanks!

stas00:
What @tjruwase said, I use the manual:
```
pre-commit run --all-files
```
---------------

I'd include the wrapper for all  4 calls to future proof the buillder code:

https://github.com/microsoft/DeepSpeed/blob/54bed327f55325dd517fe34c369d4ea9311c6547/op_builder/builder.py#L224-L227

but of course this can be done as the need arises.

-----------------

> While doing that, I did notice that the async op is built with -O0 while cpu_adam uses -O3.

Yes, I raises this issue a while back - I think we have these flags repeated more than twice too in the final gcc call - I forget which one of -OX takes precedence when repeated.

Basically need to look at the final gcc sequence and see if it needs to be cleaned up - remember we have these coming from pytorch too. Perhaps a separate issue/PR?

adammoody:
> 
> 
> @adammoody, looks good. Right now the CI is failing because of formatting issues. Please see [here](https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md) for simple steps to fix these issues. Thanks!

Yes, thanks.  I haven't integrated the format checking into my development environment yet.  I'll do that next.

adammoody:
@stas00 , I just pushed a commit to also strip the other flags you mentioned.  I also see some for cuda flags passed to nvcc.  I'm guessing we should probably also include those.  That right?

stas00:
I'd say Yes! since the specific builders override it

adammoody:
This strips empty strings from the nvcc flags now, as well.

PR #1239, opened 2021-07-19
Fix cudaErrorInvalidConfiguration in attn_softmax() for large values of sequence_length*num_heads

dfyz:
When testing DeepSpeed internally, we stumbled upon a strange error when training transformer models with 32 heads and max sequence length of 2048. We pinned it down to the forward softmax kernel in `softmax_kernels.cu` using a 2D grid instead of a 1D one, as in other kernels. The CUDA documentation [states](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#features-and-technical-specifications) that the `maximum y- or z-dimension of a grid of thread blocks` is `65535`. You're supposed to get a `cudaErrorInvalidConfiguration` error when you try to launch a kernel exceeding this limit, and this is exactly what we're seeing — `heads * sequence_length / block_compute_size` is `32 * 2048 / 1 = 65536` in our case.

This PR fixes this by swapping the `x` and `y` dimensions of the grid. I believe this is the least error-prone approach, and it should work because the current `x` dimension is simply `batch_size`, which, AFAIK, is very unlikely to exceed 65535 anywhere.

Switching the forward softmax kernel to a 1D grid remains an option that I can implement if the current fix has some drawbacks I didn't think of.

Note 1: this kind of error is unfortunately quite hard to debug right now, since there is no error checking after launching custom kernels in [BertTransformerLayer<T>::Forward()](https://github.com/microsoft/DeepSpeed/blob/32e85eda58c560e5d5a596f71fce3682ac2ef80a/csrc/transformer/ds_transformer_cuda.cpp#L147). I solved this locally by copiously applying [this macro](https://github.com/microsoft/DeepSpeed/blob/32e85eda58c560e5d5a596f71fce3682ac2ef80a/csrc/transformer/inference/csrc/softmax.cu#L23) from another part of DeepSpeed in `BertTransformerLayer<T>::Forward()`, but decided not to include this change as part of this PR as it is not directly related to the kernel fix proper.

Note 2: I only tested this very lightly — the new test in `test_cuda_forward.py`, a transformer with a very small hidden size but large `num_heads*seq_length`, passes with this fix and fails on master. Since the fix is very straightforward, I decided that should be enough. I can add more tests if you want me to.

### DIFF
--- a/csrc/transformer/softmax_kernels.cu
+++ b/csrc/transformer/softmax_kernels.cu
@@ -4,6 +4,20 @@
 
 namespace cg = cooperative_groups;
 
+dim3 get_attn_softmax_grid(int batch_size, int heads, int sequence_length, int threads)
+{
+    int seq_length4 = sequence_length / 4;
+    int block_compute_size =
+        (seq_length4 < threads ? (int)pow(2.0, floor(log2((float)(threads / seq_length4)))) : 1);
+    // Note that the Y and Z dimensions are limited to 65535, while X is basically unlimited:
+    // https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#features-and-technical-specifications
+    // The batch size is typically relatively small, while the sequence length could potentially be
+    // arbitrarily large. We therefore place the batch size second to avoid hitting the Y limit.
+    unsigned x = heads * sequence_length / block_compute_size;
+    unsigned y = batch_size;
+    return {x, y};
+}
+
 // Fused attention + softmax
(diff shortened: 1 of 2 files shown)


RezaYazdaniAminabadi:
Hi @dfyz 

Thanks for the PR and fixing this bug.
I will take a closer look and we can merge this soon 👍 

Best,
Reza

RezaYazdaniAminabadi (approved):

dfyz:
Thank you! 🥳

Issue #1209, opened 2021-07-08
GPT-NEO model parallelism not work issue


switiz:
@RezaYazdaniAminabadi 
When I inference using gpt-neo 1.3b model with tutorial code. may be model parallelism is not work
I tested several times using multiple A100 gpus. In my opinion, if i increase the number of gpus, memory usage per gpu should be decrease. But rather it is increasing.

And if the number of gpu is increased, host ram consumption increases to process * num_gpus. 
So, if the num of gpu is increase continuously, I met  host OOM.

What should I do?

### code
```
local_rank = int(os.getenv('LOCAL_RANK', '0'))
world_size = int(os.getenv('WORLD_SIZE', '1'))

tokenizer = AutoTokenizer.from_pretrained("EleutherAI/gpt-neo-1.3B")
model = AutoModelForCausalLM.from_pretrained('/workspace/home/hugging_gpt/gpt_torch/')
generator = pipeline('text-generation', model=model, tokenizer=tokenizer, device=local_rank)
generator.model = deepspeed.init_inference(generator.model,
                                    mp_size=world_size,
                                    dtype=torch.float,
                                    replace_method='auto')
prompt = "Q: why not work? ? A:"
start_time = time.time()
sample_outputs = generator(prompt, do_sample=True, top_k=50,
                                max_length=300, top_p=0.95, temperature=1.5, num_return_sequences=1)

if world_size > 0 and torch.distributed.get_rank() == 0:
    print(f'spend:{time.time()-start_time}')
    for i, sample_output in enumerate(sample_outputs):
        print("[index]: {} \n[decode]: {}".format(i, sample_outputs))
```

### test result
1gpu
[0] A100-SXM4-40GB   | 38'C,   0 % |  7966 / 40537 MB |

2gpu
[0] A100-SXM4-40GB   | 40'C,  33 % |  8704 / 40537 MB |
[1] A100-SXM4-40GB   | 38'C,  87 % |  8706 / 40537 MB |
 
4 gpu
[0] A100-SXM4-40GB   | 39'C,  50 % |  8318 / 40537 MB |
[1] A100-SXM4-40GB   | 37'C,  45 % |  8654 / 40537 MB |
[2] A100-SXM4-40GB   | 39'C,  79 % |  8654 / 40537 MB |
[3] A100-SXM4-40GB   | 40'C,  34 % |  8352 / 40537 MB |

###logs
```
[...]

Issue #1161, opened 2021-06-15
Questions about implementing model parallelism in the inference engine


hyunwoongko:
Hello. I would like to ask about the model parallelism feature in the inference engine. In general, the model parallelism that I can think of is inter-layer model parallelism like GPipe (only partitioning part, not pipelining) and intra-layer model parallelism like Megatron-LM.

https://github.com/microsoft/DeepSpeed/blob/aa16828c267dc7a790a5cc2f43f93a67617a2364/deepspeed/inference/engine.py#L102

But, in the case of the current implementation of inference engine, it seems to broadcast all parameters from device 0 if mpu is not input. But, as far as I know, these operations don't seem parallelizing way that models are sliced.  

https://github.com/microsoft/DeepSpeed/blob/aa16828c267dc7a790a5cc2f43f93a67617a2364/deepspeed/inference/engine.py#L213

In this part of the code, it seems that input is also broadcasting to all devices. So, I wonder why deepspeed is broadcasting in all parts when mpu isn't inputted.

hyunwoongko:
- num_gpus = 2
![image](https://user-images.githubusercontent.com/38183241/122065262-c2fa5280-ce2c-11eb-927b-9a35cdc7a6c7.png)

- num_gpus = 1
![image](https://user-images.githubusercontent.com/38183241/122065368-da394000-ce2c-11eb-929d-40d75e8d5e77.png)

hyunwoongko:
If model parallelism in the form of tensor slicing for GPT-neo is not currently supported, the paragraph below may need to be slightly modified :) If I'm mistaken, please help me to do model parallelization!

![image](https://user-images.githubusercontent.com/38183241/122068124-166da000-ce2f-11eb-8764-b8b3b8bc8653.png)

RezaYazdaniAminabadi:
For reducing the results across different GPUs, we use the all_reduce in the [inference-api](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/ops/transformer/inference/transformer_inference.py#L311). The model-parallel group is also created in the [inference-engine](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/inference/engine.py#L88) and passed to the inference-module.

hyunwoongko:
Thanks for your kind reply. :)

[...]

Issue #1192, opened 2021-06-28
Inference with `replace_method='auto'` fails for `gpt2-xl` generation


aphedges:
I modified the code at "[End-to-End GPT NEO 2.7B Inference](https://github.com/microsoft/DeepSpeed/blob/aa16828c267dc7a790a5cc2f43f93a67617a2364/docs/_tutorials/inference-tutorial.md#end-to-end-gpt-neo-27b-inference)" by replacing `EleutherAI/gpt-neo-2.7B` with `gpt2-xl`, and I get the following stack trace when running on 2 GPUs:

```console
[2021-06-28 13:56:01,937] [ERROR] Uncaught Exception
Traceback (most recent call last):
  File "gpt2-generation.py", line 20, in <module>
    string = generator("DeepSpeed is", do_sample=True, min_length=50)
  File "venv/lib/python3.7/site-packages/transformers/pipelines/text_generation.py", line 148, in __call__
    output_sequences = self.model.generate(input_ids=input_ids, **generate_kwargs)  # BS x SL
  File "venv/lib/python3.7/site-packages/torch/autograd/grad_mode.py", line 27, in decorate_context
    return func(*args, **kwargs)
  File "venv/lib/python3.7/site-packages/transformers/generation_utils.py", line 1026, in generate
    **model_kwargs,
  File "venv/lib/python3.7/site-packages/transformers/generation_utils.py", line 1535, in sample
    output_hidden_states=output_hidden_states,
  File "venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
  File "venv/lib/python3.7/site-packages/deepspeed/inference/engine.py", line 229, in forward
    return self.model_orig_fwd(*inputs, **kwargs)
  File "venv/lib/python3.7/site-packages/transformers/models/gpt2/modeling_gpt2.py", line 956, in forward
    return_dict=return_dict,
  File "venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
  File "venv/lib/python3.7/site-packages/transformers/models/gpt2/modeling_gpt2.py", line 799, in forward
    output_attentions=output_attentions,
  File "venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
[...]

PR #1244, opened 2021-07-20
Reducing the memory-overhead of creating large-models for multi-GPU run

RezaYazdaniAminabadi:
This PR addresses the issue of creating large models for inference. We first partition the parallelable tensors on CPU based on the number of GPUs to run the model and then move each partition to the device. 

This fixes https://github.com/microsoft/DeepSpeed/issues/1209 and https://github.com/microsoft/DeepSpeed/issues/1161

TODO: I will change the inference test scripts to reflect these changes.

### DIFF
--- a/deepspeed/inference/engine.py
+++ b/deepspeed/inference/engine.py
@@ -44,15 +44,6 @@ class InferenceEngine(Module):
         self.quantize_merge_count = 1
         self.quantization_scales = None
 
-        if self.mpu:
-            self.mp_world_size = dist.get_world_size(
-                group=self.mpu.get_model_parallel_group())
-            self.mp_group = self.mpu.get_model_parallel_group()
-        elif self.mp_world_size > 1 and not dist.is_initialized():
-            self._create_model_parallel_group()
-        else:
-            self.module.to(torch.cuda.current_device())
-
         self._check_quantize_setting(quantization_setting)
 
         if self.checkpoint:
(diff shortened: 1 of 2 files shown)


jeffra (approved):

acostin1:
I just tested this branch and I can confirm it splits the NeoGPT125M models equally between the 8 A100 GPUs - but the performance is ~10% slower.

Configuration before with 0.4.3 - load model with .cuda(), deepspeed zero2, batch size 100
- memory of first GPU 39GB (20 GB taken on model load)
- time to finish my training - 73h
image


Configuration with 0.4.4 - load model with .cpu(), batch size 240 - now it occupies all memory equally 
image

But estimated time to finish 85h
image

hyunwoongko:
```python
self.module.to(torch.cuda.current_device())
```

Does this work fine as we want it to?
I implemented the code so that only parameters in the CPU are uploaded to the gpu.
https://github.com/tunib-ai/parallelformers/blob/main/parallelformers/parallel/engine.py#L86

Is it the same implementation as this one?

switiz:
When I test this PR. It's works for me.
Test 13B model with half precision 

before
1 - 8 gpus : OOM

after
1 gpus : 30G
2 gpus : 16.5G
4 gpus : 8.5G
8 gpus : 6G

It is not completely divisible by number of gpus, but it is acceptable

thank you
BR

aphedges:
From looking at the code, this PR also seems to solve #1192.

PR #1246, opened 2021-07-23
[zero-1] ignore overlap/contiguous_gradients flags

jeffra:
1)  Overlap and contiguous grads are meaningless in stage 1 and should be ignored.
2) Found typo'd mpi variable that would cause a crash (I guess it's not a frequently used code path).


tjruwase (approved):

samyam (approved):
LGTM and makes sense to do this. Since none of the backward hooks are triggered by stage 1, there is no overlapping or contiguous gradients.

PR #1247, opened 2021-07-23
test link against libaio using distutils

adammoody:
Work in progress...

This implements more robust testing for libaio based on the discussion threads in https://github.com/microsoft/DeepSpeed/issues/1126 and https://github.com/microsoft/DeepSpeed/pull/1213

This first commit implements the fallback solution which lets the user declare where libaio is installed.  This successfully builds against the libaio installed in my conda environment after doing something like:
```
# install libaio into conda environment
conda install libaio

# tip the compiler on where to find libaio.h and libaio.so
export CFLAGS="-I${CONDA_PREFIX}/include $CFLAGS"
export LDFLAGS="-L${CONDA_PREFIX}/lib $LDFLAGS"

export DS_BUILD_OPS=1
python setup.py bdist_wheel
```

I'll work on the package manager detection bit described in https://github.com/microsoft/DeepSpeed/issues/1126#issuecomment-854101593 and submit that as part of this same PR.

### DIFF
--- a/op_builder/async_io.py
+++ b/op_builder/async_io.py
@@ -50,10 +50,14 @@ class AsyncIOBuilder(OpBuilder):
         return ['-laio']
 
     def is_compatible(self):
-        aio_libraries = ['libaio-dev']
-        aio_compatible = self.libraries_installed(aio_libraries)
+        # Check for the existence of libaio by using distutils
+        # to compile and link a test program that calls io_submit,
+        # which is a function provided by libaio that is used in the async_io op.
+        # If needed, one can define -I and -L entries in CFLAGS and LDFLAGS
+        # respectively to specify the directories for libaio.h and libaio.so.
+        aio_compatible = self.has_function('io_submit', ('aio', ))
         if not aio_compatible:
             self.warning(
-                f"{self.NAME} requires the libraries: {aio_libraries} but are missing. Can be fixed by: `apt install libaio-dev`."
+                f"{self.NAME} requires libaio but it is missing. Can be fixed by: `apt install libaio-dev`."
             )
         return super().is_compatible() and aio_compatible
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -7,6 +7,7 @@ import time
 import importlib
 from pathlib import Path
 import subprocess
+import distutils.ccompiler
 from abc import ABC, abstractmethod
 
 YELLOW = '\033[93m'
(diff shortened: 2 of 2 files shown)


adammoody:
Hmm, the first commit by itself leads to the following error on while testing a build on the CI system:
```
AttributeError: module 'distutils' has no attribute 'ccompiler'
```
That might be happening because ``numpy`` or some other package perhaps has a ``distutils`` so that it's looking at the wrong ``distutils`` package for ``ccompiler`` as described in: https://github.com/AICoE/log-anomaly-detector/issues/376

That page lists a work around that seems to help here, as well.

op_builder/async_io.py
```diff
@@ -50,10 +50,9 @@ def extra_ldflags(self):
         return ['-laio']
 
     def is_compatible(self):
-        aio_libraries = ['libaio-dev']
-        aio_compatible = self.libraries_installed(aio_libraries)
+        aio_compatible = self.has_function('io_submit', ('aio', ))
```
tjruwase, line 53: Can you please a brief explanation for `io_submit`?

adammoody: If that's too much or not quite right, just let me know.

tjruwase (approved):

op_builder/async_io.py
```diff
@@ -50,10 +50,14 @@ def extra_ldflags(self):
...
+        # Check for the existence of libaio by using distutils
+        # to compile and link a test program that calls io_submit,
+        # which is a function provided by libaio that is used in the async_io op.
+        # If needed, one can define -I and -L entries in CFLAGS and LDFLAGS
+        # respectively to specify the directories for libaio.h and libaio.so.
+        aio_compatible = self.has_function('io_submit', ('aio', ))
         if not aio_compatible:
             self.warning(
-                f"{self.NAME} requires the libraries: {aio_libraries} but are missing. Can be fixed by: `apt install libaio-dev`."
+                f"{self.NAME} requires libaio but it is missing. Can be fixed by: `apt install libaio-dev`."
```
stas00, line 61: But this is incorrect. It requires some variation of `libaio` dev package and not `libaio`.

stas00: I guess we can address it in https://github.com/microsoft/DeepSpeed/pull/1250

op_builder/builder.py
```diff
@@ -160,6 +161,10 @@ def libraries_installed(self, libraries):
             valid = valid or result.wait() == 0
         return valid
 
+    def has_function(self, funcname, libraries):
+        compiler = distutils.ccompiler.new_compiler()
+        return compiler.has_function(funcname, libraries=libraries)
```
stas00, line 166: what if the user doesn't have the compiler installed?

PR #1249, opened 2021-07-24
aio: enable -O3 optimization

adammoody:
This updates the aio op to build with ``-O3`` rather than ``-O0`` optimization.

It looks like all ops either build with ``-O2`` or ``-O3``, but aio uses ``-O0`` for some reason.  Is that intentional?

### DIFF
--- a/op_builder/async_io.py
+++ b/op_builder/async_io.py
@@ -30,6 +30,7 @@ class AsyncIOBuilder(OpBuilder):
         return ['csrc/aio/py_lib', 'csrc/aio/common']
 
     def cxx_args(self):
+        # -O0 for improved debugging, since performance is bound by I/O
         CPU_ARCH = self.cpu_arch()
         SIMD_WIDTH = self.simd_width()
         return [


tjruwase:
@adammoody, aio op is IO-bound and thus gets little benefit from compiler optimizations. So we choose `-O0` for better debugging experience.

adammoody:
Ok, thanks.  I had noticed that some of aio includes some SIMD code like here:

https://github.com/microsoft/DeepSpeed/blob/6ba96289702afc56948aa9f6cff55d7df678ccc0/csrc/aio/py_lib/deepspeed_py_copy.cpp#L107-L109

On PowerPC, those SIMD-tuned algorithms aren't available, so it falls back to the non-vectorized copy method here:

https://github.com/microsoft/DeepSpeed/blob/6ba96289702afc56948aa9f6cff55d7df678ccc0/csrc/aio/py_lib/deepspeed_py_copy.cpp#L46-L49

I thought the compiler might help in that case, and that maybe that would be useful since someone took the time to create the SIMD versions.

I do understand performance being bound by much slower I/O, though.

Mind if I replace this PR with one that adds a comment to note why ``-O0`` is used?

tjruwase:
> Ok, thanks. I had noticed that some of aio includes some SIMD code like here:

Yes, you are correct. That was a failed attempt to create fast copy routines, but was abandoned due to poor progress. The routines are currently not used, and these codes should be ideally be deleted. Sorry for the confusion.

tjruwase:
> Mind if I replace this PR with one that adds a comment to note why `-O0` is used?

Thanks, that would be very much appreciated.

tjruwase (approved):

PR #1250, opened 2021-07-24
query for libaio package using known package managers

adammoody:
This depends on PR https://github.com/microsoft/DeepSpeed/pull/1247

It extends that PR to add package manager detection and corresponding error messages on which package to install.  This is related to https://github.com/microsoft/DeepSpeed/issues/1126#issuecomment-854101593, though it has modified things to only print messages based on the package manager.

### DIFF
--- a/op_builder/async_io.py
+++ b/op_builder/async_io.py
@@ -1,6 +1,9 @@
 """
 Copyright 2020 The Microsoft DeepSpeed Team
 """
+import distutils.spawn
+import subprocess
+
 from .builder import OpBuilder
 
 
@@ -50,6 +53,34 @@ class AsyncIOBuilder(OpBuilder):
     def extra_ldflags(self):
         return ['-laio']
 
+    def check_for_libaio_pkg(self):
+        libs = dict(
+            dpkg=["-l",
+                  "libaio-dev"],
+            pacman=["-Q",
+                    "libaio"],
+            rpm=["-q",
+                 "libaio-devel"],
+        )
+
+        found = False
+        for pkgmgr, data in libs.items():
+            flag, lib = data
+            path = distutils.spawn.find_executable(pkgmgr)
+            if path is not None:
+                cmd = f"{pkgmgr} {flag} {lib}"
+                result = subprocess.Popen(cmd,
+                                          stdout=subprocess.PIPE,
+                                          stderr=subprocess.PIPE,
+                                          shell=True)
+                if result.wait() == 0:
+                    found = True
+                else:
+                    self.warning(
+                        f"{self.NAME}: please install the {lib} package with {pkgmgr}")
+                break
+        return found
+
     def is_compatible(self):
         # Check for the existence of libaio by using distutils
         # to compile and link a test program that calls io_submit,
@@ -59,6 +90,14 @@ class AsyncIOBuilder(OpBuilder):
         aio_compatible = self.has_function('io_submit', ('aio', ))
         if not aio_compatible:
             self.warning(
-                f"{self.NAME} requires libaio but it is missing. Can be fixed by: `apt install libaio-dev`."
+                f"{self.NAME} requires the dev libaio .so object and headers but these were not found."
+            )
+
+            # Check for the libaio package via known package managers
+            # to print suggestions on which package to install.
+            self.check_for_libaio_pkg()
+
+            self.warning(
+                "If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found."
             )
         return super().is_compatible() and aio_compatible
--- a/op_builder/async_io.py
+++ b/op_builder/async_io.py
@@ -56,16 +56,19 @@ class AsyncIOBuilder(OpBuilder):
     def check_for_libaio_pkg(self):
         libs = dict(
             dpkg=["-l",
-                  "libaio-dev"],
+                  "libaio-dev",
+                  "apt"],
             pacman=["-Q",
-                    "libaio"],
+                    "libaio",
+                    "pacman"],
             rpm=["-q",
-                 "libaio-devel"],
+                 "libaio-devel",
+                 "yum"],
         )
 
         found = False
         for pkgmgr, data in libs.items():
-            flag, lib = data
+            flag, lib, tool = data
             path = distutils.spawn.find_executable(pkgmgr)
             if path is not None:
                 cmd = f"{pkgmgr} {flag} {lib}"
@@ -77,7 +80,7 @@ class AsyncIOBuilder(OpBuilder):
                     found = True
                 else:
                     self.warning(
-                        f"{self.NAME}: please install the {lib} package with {pkgmgr}")
+                        f"{self.NAME}: please install the {lib} package with {tool}")
                 break
         return found
 
--- a/csrc/transformer/softmax_kernels.cu
+++ b/csrc/transformer/softmax_kernels.cu
@@ -4,6 +4,20 @@
 
 namespace cg = cooperative_groups;
 
+dim3 get_attn_softmax_grid(int batch_size, int heads, int sequence_length, int threads)
+{
+    int seq_length4 = sequence_length / 4;
+    int block_compute_size =
+        (seq_length4 < threads ? (int)pow(2.0, floor(log2((float)(threads / seq_length4)))) : 1);
+    // Note that the Y and Z dimensions are limited to 65535, while X is basically unlimited:
+    // https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#features-and-technical-specifications
+    // The batch size is typically relatively small, while the sequence length could potentially be
+    // arbitrarily large. We therefore place the batch size second to avoid hitting the Y limit.
+    unsigned x = heads * sequence_length / block_compute_size;
+    unsigned y = batch_size;
+    return {x, y};
+}
+
 // Fused attention + softmax
 template <int tbSize, int blockStride, int tbSeq>
 __global__ void attn_softmax(float* vals,
@@ -22,12 +36,12 @@ __global__ void attn_softmax(float* vals,
     cg::thread_block b = cg::this_thread_block();
     cg::thread_block_tile<tbSize> g = cg::tiled_partition<tbSize>(b);
 
-    int batch = blockIdx.x;
-    int row = blockIdx.y;
+    int batch = blockIdx.y;
+    int row = blockIdx.x;
     int max_threads_in_sequence = std::max(seq_length, tbSeq);
     int seq_lane = threadIdx.x % max_threads_in_sequence;
 
-    int data_offset = batch * (gridDim.y * block_width) + row * block_width +
+    int data_offset = batch * (gridDim.x * block_width) + row * block_width +
                       (threadIdx.x / max_threads_in_sequence) * seq_length;
     int mask_offset = batch * seq_length;
 
@@ -153,12 +167,12 @@ __global__ void attn_softmax(__half* vals,
     cg::thread_block b = cg::this_thread_block();
     cg::thread_block_tile<tbSize> g = cg::tiled_partition<tbSize>(b);
 
-    int batch = blockIdx.x;
-    int row = blockIdx.y;
+    int batch = blockIdx.y;
+    int row = blockIdx.x;
     int max_threads_in_sequence = std::max(seq_length, tbSeq);
     int seq_lane = threadIdx.x % max_threads_in_sequence;
 
-    int data_offset = batch * (gridDim.y * block_width) + row * block_width +
+    int data_offset = batch * (gridDim.x * block_width) + row * block_width +
                       (threadIdx.x / max_threads_in_sequence) * seq_length;
     int mask_offset = batch * seq_length;
 
@@ -300,9 +314,7 @@ void launch_attn_softmax<float>(float* vals,
     const int threads = 128;
     int seq_length4 = sequence_length / 4;
 
-    int block_compute_size =
-        (seq_length4 < threads ? (int)pow(2.0, floor(log2((float)(threads / seq_length4)))) : 1);
-    dim3 grid_dim(batch_size, heads * sequence_length / block_compute_size);
+    dim3 grid_dim = get_attn_softmax_grid(batch_size, heads, sequence_length, threads);
 
     int subblock_max_workload = MAX_THREAD_ITERATIONS * 4 * threads;
 
@@ -333,10 +345,7 @@ void launch_attn_softmax<float>(float* vals,
             <<<grid_dim, block_dim, 0, stream>>>(vals, attn_mask, heads, seq_length4, iterations);
     else {
         const int threads = 256;
-        block_compute_size =
-            (seq_length4 < threads ? (int)pow(2.0, floor(log2((float)(threads / seq_length4))))
-                                   : 1);
-        dim3 grid_dim(batch_size, heads * sequence_length / block_compute_size);
+        dim3 grid_dim = get_attn_softmax_grid(batch_size, heads, sequence_length, threads);
 
         int subblock_max_workload = MAX_THREAD_ITERATIONS * 4 * threads;
 
@@ -370,9 +379,7 @@ void launch_attn_softmax<__half>(__half* vals,
     const int threads = 128;
     int seq_length4 = sequence_length / 4;
 
-    int block_compute_size =
-        (seq_length4 < threads ? (int)pow(2.0, floor(log2((float)(threads / seq_length4)))) : 1);
-    dim3 grid_dim(batch_size, heads * sequence_length / block_compute_size);
+    dim3 grid_dim = get_attn_softmax_grid(batch_size, heads, sequence_length, threads);
 
     int subblock_max_workload = MAX_THREAD_ITERATIONS * 4 * threads;
 
@@ -404,10 +411,7 @@ void launch_attn_softmax<__half>(__half* vals,
             <<<grid_dim, block_dim, 0, stream>>>(vals, attn_mask, heads, seq_length4, iterations);
     else {
         const int threads = 256;
-        block_compute_size =
-            (seq_length4 < threads ? (int)pow(2.0, floor(log2((float)(threads / seq_length4))))
-                                   : 1);
-        dim3 grid_dim(batch_size, heads * sequence_length / block_compute_size);
+        dim3 grid_dim = get_attn_softmax_grid(batch_size, heads, sequence_length, threads);
 
         int subblock_max_workload = MAX_THREAD_ITERATIONS * 4 * threads;
 
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -482,6 +482,9 @@ class DeepSpeedEngine(Module):
     def zero_allgather_partitions(self):
         return self._config.zero_config.allgather_partitions
 
+    def zero_round_robin_gradients(self):
+        return self._config.zero_config.round_robin_gradients
+
     def dump_state(self):
         return self._config.dump_state
 
@@ -908,11 +911,13 @@ class DeepSpeedEngine(Module):
         elif zero_stage <= ZERO_OPTIMIZATION_GRADIENTS:
             overlap_comm = self.zero_overlap_comm()
             contiguous_gradients = self.zero_contiguous_gradients()
+            round_robin_gradients = self.zero_round_robin_gradients()
 
             # Overlap and contiguous grads are meaningless in stage 1 and are ignored
             if zero_stage == ZERO_OPTIMIZATION_OPTIMIZER_STATES:
                 overlap_comm = False
                 contiguous_gradients = False
+                round_robin_gradients = False
 
             if isinstance(self.module, PipelineModule):
                 if overlap_comm:
@@ -940,7 +945,8 @@ class DeepSpeedEngine(Module):
                 gradient_predivide_factor=self.gradient_predivide_factor(),
                 gradient_accumulation_steps=self.gradient_accumulation_steps(),
                 ignore_unused_parameters=self.zero_ignore_unused_parameters(),
-                partition_grads=zero_stage == ZERO_OPTIMIZATION_GRADIENTS)
+                partition_grads=zero_stage == ZERO_OPTIMIZATION_GRADIENTS,
+                round_robin_gradients=round_robin_gradients)
         elif zero_stage == ZERO_OPTIMIZATION_WEIGHTS:
             logger.info("Initializing ZeRO Stage 3") if dist.get_rank() == 0 else None
             from deepspeed.runtime.zero.stage3 import FP16_DeepSpeedZeroOptimizer_Stage3
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -573,7 +573,7 @@ class DeepSpeedEngine(Module):
         if "OMPI_COMM_WORLD_LOCAL_RANK" in os.environ:
             ompi_local_rank = os.environ.get("OMPI_COMM_WORLD_LOCAL_RANK")
             local_rank = os.environ.get('LOCAL_RANK', ompi_local_rank)
-            assert ompi_local_rank == local_rank, f"LOCAL_RANK ({local_rank}) != OMPI_COMM_WORLD_LOCAL_RANK ({mpi_local_rank}), " \
+            assert ompi_local_rank == local_rank, f"LOCAL_RANK ({local_rank}) != OMPI_COMM_WORLD_LOCAL_RANK ({ompi_local_rank}), " \
                 "not sure how to proceed as we're seeing conficting local rank info."
             os.environ['LOCAL_RANK'] = local_rank
 
@@ -907,6 +907,13 @@ class DeepSpeedEngine(Module):
                 gradient_predivide=self.gradient_predivide)
         elif zero_stage <= ZERO_OPTIMIZATION_GRADIENTS:
             overlap_comm = self.zero_overlap_comm()
+            contiguous_gradients = self.zero_contiguous_gradients()
+
+            # Overlap and contiguous grads are meaningless in stage 1 and are ignored
+            if zero_stage == ZERO_OPTIMIZATION_OPTIMIZER_STATES:
+                overlap_comm = False
+                contiguous_gradients = False
+
             if isinstance(self.module, PipelineModule):
                 if overlap_comm:
                     logger.warning(
@@ -921,7 +928,7 @@ class DeepSpeedEngine(Module):
                 dynamic_loss_scale=self.dynamic_loss_scale(),
                 dynamic_loss_args=self.dynamic_loss_scale_args(),
                 clip_grad=self.gradient_clipping(),
-                contiguous_gradients=self.zero_contiguous_gradients(),
+                contiguous_gradients=contiguous_gradients,
                 reduce_bucket_size=self.zero_reduce_bucket_size(),
                 allgather_bucket_size=self.zero_allgather_bucket_size(),
                 dp_process_group=self.data_parallel_group,
--- a/deepspeed/runtime/zero/config.py
+++ b/deepspeed/runtime/zero/config.py
@@ -189,4 +189,4 @@ class DeepSpeedZeroConfig(DeepSpeedConfigObject):
         self.round_robin_gradients = get_scalar_param(
             zero_config_dict,
             ZERO_OPTIMIZATION_ROUND_ROBIN_GRADIENTS,
-            ZERO3_OPTIMIZATION_CONTIGUOUS_GRADIENTS_DEFAULT)
+            ZERO_OPTIMIZATION_ROUND_ROBIN_GRADIENTS_DEFAULT)
--- a/deepspeed/runtime/zero/config.py
+++ b/deepspeed/runtime/zero/config.py
@@ -39,6 +39,7 @@ class DeepSpeedZeroConfig(DeepSpeedConfigObject):
         self.gather_fp16_weights_on_model_save = None
 
         self.ignore_unused_parameters = None
+        self.round_robin_gradients = None
 
         if ZERO_OPTIMIZATION in param_dict.keys():
             zero_config_dict = param_dict[ZERO_OPTIMIZATION]
@@ -184,3 +185,8 @@ class DeepSpeedZeroConfig(DeepSpeedConfigObject):
         self.legacy_stage1 = get_scalar_param(zero_config_dict,
                                               ZERO_OPTIMIZATION_LEGACY_STAGE1,
                                               ZERO_OPTIMIZATION_LEGACY_STAGE1_DEFAULT)
+
+        self.round_robin_gradients = get_scalar_param(
+            zero_config_dict,
+            ZERO_OPTIMIZATION_ROUND_ROBIN_GRADIENTS,
+            ZERO3_OPTIMIZATION_CONTIGUOUS_GRADIENTS_DEFAULT)
--- a/deepspeed/runtime/zero/constants.py
+++ b/deepspeed/runtime/zero/constants.py
@@ -30,7 +30,8 @@ ZeRO optimization should be enabled as:
     "sub_group_size" : 1000000000000,
     "offload_param": {...},
     "offload_optimizer": {...},
-    "ignore_unused_parameters": [true|false]
+    "ignore_unused_parameters": [true|false],
+    "round_robin_gradients": [true|false]
     }
 }
 '''
@@ -124,6 +125,10 @@ ZERO_OPTIMIZATION_IGNORE_UNUSED_PARAMETERS_DEFAULT = True
 ZERO_OPTIMIZATION_LEGACY_STAGE1 = "legacy_stage1"
 ZERO_OPTIMIZATION_LEGACY_STAGE1_DEFAULT = False
 
+# Stage 2 - partition gradients in a round robin fashsion to load-balance reduction and offload copying
+ZERO_OPTIMIZATION_ROUND_ROBIN_GRADIENTS = 'round_robin_gradients'
+ZERO_OPTIMIZATION_ROUND_ROBIN_GRADIENTS_DEFAULT = False
+
 #yapf: disable
 ZERO_OPTIMIZATION_DEFAULT = {
     ZERO_OPTIMIZATION_STAGE:
@@ -161,5 +166,7 @@ ZERO_OPTIMIZATION_DEFAULT = {
     ZERO_OPTIMIZATION_IGNORE_UNUSED_PARAMETERS:
     ZERO_OPTIMIZATION_IGNORE_UNUSED_PARAMETERS_DEFAULT,
     ZERO_OPTIMIZATION_LEGACY_STAGE1:
-    ZERO_OPTIMIZATION_LEGACY_STAGE1_DEFAULT
+    ZERO_OPTIMIZATION_LEGACY_STAGE1_DEFAULT,
+    ZERO_OPTIMIZATION_ROUND_ROBIN_GRADIENTS:
+    ZERO_OPTIMIZATION_ROUND_ROBIN_GRADIENTS_DEFAULT
 }
--- a/deepspeed/runtime/zero/stage2.py
+++ b/deepspeed/runtime/zero/stage2.py
@@ -99,7 +99,8 @@ class FP16_DeepSpeedZeroOptimizer(object):
                  gradient_predivide_factor=1.0,
                  gradient_accumulation_steps=1,
                  ignore_unused_parameters=True,
-                 partition_grads=True):
+                 partition_grads=True,
+                 round_robin_gradients=False):
 
         if dist.get_rank() == 0:
             logger.info(f"Reduce bucket size {reduce_bucket_size}")
@@ -159,6 +160,7 @@ class FP16_DeepSpeedZeroOptimizer(object):
         self.gradient_accumulation_steps = gradient_accumulation_steps
         self.micro_step_id = 0
         self.ignore_unused_parameters = ignore_unused_parameters
+        self.round_robin_gradients = round_robin_gradients
 
         self.extra_large_param_to_reduce = None
 
@@ -232,10 +234,15 @@ class FP16_DeepSpeedZeroOptimizer(object):
             # This ensures that gradients are reduced in a fashion such that ownership round robins among the ranks.
             # For example, rather than 3 gradients (g_n+2, g_n+1, g_n) that are reduced consecutively belonging
             # to the same rank, instead they will belong to 3 ranks (r_m+2, r_m+1, r_m).
-            round_robin_tensors, round_robin_indices = self._round_robin_reorder(
-                self.fp16_groups[i],
-                dist.get_world_size(group=self.dp_process_group)
-            )
+            if self.round_robin_gradients:
+                round_robin_tensors, round_robin_indices = self._round_robin_reorder(
+                    self.fp16_groups[i],
+                    dist.get_world_size(group=self.dp_process_group)
+                )
+            else:
+                round_robin_tensors = self.fp16_groups[i]
+                round_robin_indices = list(range(len(self.fp16_groups[i])))
+
             self.round_robin_fp16_groups.append(round_robin_tensors)
             self.round_robin_fp6_indices.append(round_robin_indices)
 
--- a/deepspeed/runtime/zero/stage2.py
+++ b/deepspeed/runtime/zero/stage2.py
@@ -106,6 +106,7 @@ class FP16_DeepSpeedZeroOptimizer(object):
             logger.info(f"Reduce bucket size {reduce_bucket_size}")
             logger.info(f"Allgather bucket size {allgather_bucket_size}")
             logger.info(f"CPU Offload: {cpu_offload}")
+            logger.info(f'Round robin gradient partitioning: {round_robin_gradients}')
         # The fused optimizer does all the work. We need this layer for two reason:
         # 1. maintain same user API from apex.fp16_utils
         # 2. keep common stuff here in case we need to add ne552w fused optimizer later
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -7,7 +7,13 @@ import time
 import importlib
 from pathlib import Path
 import subprocess
+import shlex
+import shutil
+import tempfile
 import distutils.ccompiler
+import distutils.log
+import distutils.sysconfig
+from distutils.errors import CompileError, LinkError
 from abc import ABC, abstractmethod
 
 YELLOW = '\033[93m'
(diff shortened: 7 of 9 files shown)


adammoody:
@stas00 , I have these two PRs for the libaio search.  This is a bit different from the original plan we discussed.

The ``distutils.ccompiler`` test seemed to work well for both my conda install and a system install where libaio is available in the default search paths, which should be the case for most package managers.  So I opted to use that as the first search method.

This adds the package manager query step, though now it's used to help print a message to the user which may help them resolve the issue rather than a check as to whether the build will actually work.

Please let me know if you still think we should flip this.

op_builder/async_io.py
```diff
@@ -59,6 +90,11 @@ def is_compatible(self):
         aio_compatible = self.has_function('io_submit', ('aio', ))
         if not aio_compatible:
             self.warning(
-                f"{self.NAME} requires libaio but it is missing. Can be fixed by: `apt install libaio-dev`."
+                f"{self.NAME} requires libaio but it was not found.  If libaio is installed, try setting CFLAGS and LDFLAGS."
             )
+
+            # Check for the libaio package via known package managers
+            # to print suggestions on which package to install.
+            self.check_for_libaio_pkg()
```
stas00, line 98:
```suggestion
                f"{self.NAME} requires the dev libaio .so object and headers but these were not found. 
            )

            # Check for the libaio package via known package managers
            # to print suggestions on which package to install.
            self.check_for_libaio_pkg()
            
            self.warning("If libaio is already installed, (perhaps from source), try setting CFLAGS and LDFLAGS environment variables to where it can be found.")
```

I'd suggest to first give users a possible easy solution via their package manager and only then the fallback to the complicated solution.

But I also don't fully understand the logic here - if the right package is found already to be installed why would it fail to build `io_submit`. User messing up their global `ld.conf`?

If we checked that it's already found to be installed, should we not say that? rather than "if libaio is already installed"

stas00 (changes requested):
This is really good to check that the functionality is actually there - with the compile check! Much much better.

Just a few suggestions to perhaps to make the troubleshooting easier for the user.

(and as I commented in the other PR, probably need to gracefully report if the compiler is not found, rather than possibly sending a cryptic error message).

Thank you!

op_builder/async_io.py
adammoody: Thanks for reviewing this, @stas00 .  If libaio was installed from source or through something like conda, it might actually be installed, but just not in a path where the compiler looks by default.  The CFLAGS/LDFLAGS tip is trying to help users who are in that situation.

The combination of the failed package manager check and this warning just after would produce a pair of messages like:
```
async_io: please install the libaio-devel package with rpm
If libaio is already installed, (perhaps from source), try setting CFLAGS and LDFLAGS environment variables to where it can be found.
```

stas00: I totally get the ` CFLAGS/LDFLAGS ` case - I probably wasn't clear if it sounded to you that I thought it was redundant. That's why I thought you're in the best position to complete this, since yours is that special case, Adam.

stas00: but actually it'll produce 3 lines:
```
async_io requires libaio but it was not found
please install the libaio-devel package with rpm
If libaio is already installed, (perhaps from source), try setting CFLAGS and LDFLAGS environment variables to where it can be found.

```

adammoody: Yes, I just realized that I missed that third line you suggested.  Those changes should be in there now.

adammoody: I still need to add a check for the C compiler.

op_builder/async_io.py
```diff
@@ -59,6 +90,15 @@ def is_compatible(self):
         aio_compatible = self.has_function('io_submit', ('aio', ))
```
tjruwase, line 90: I get a compile error when I build DeepSpeed from source using install.sh. I don't fully understand what this function is doing, but it generates and fails to compile the following source.
```c
main (int argc, char **argv) {
    io_submit();
}
```
The error message is reproducible by running gcc directly on the file:
```bash
/tmp/io_submitikjqrmn5.c:1:1: warning: return type defaults to ‘int’ [-Wimplicit-int]
 main (int argc, char **argv) {
 ^~~~
/tmp/io_submitikjqrmn5.c: In function ‘main’:
/tmp/io_submitikjqrmn5.c:2:5: warning: implicit declaration of function ‘io_submit’ [-Wimplicit-function-declaration]
     io_submit();
     ^~~~~~~~~
/tmp/ccFfhIAq.o: In function `main':
io_submitikjqrmn5.c:(.text+0x15): undefined reference to `io_submit'
collect2: error: ld returned 1 exit status
```

So  I am curious how this is working for you (and in our CI). Can you help? Thanks.

tjruwase (approved):

op_builder/async_io.py
tjruwase: I accidentally hit approve, sorry for the confusion

tjruwase (changes requested):

op_builder/async_io.py
adammoody: The ``has_function`` method is creating a simple C program that uses a named function, which it then attempts to compile and link.  In this case, the link is failing with ``undefined reference to `io_submit'``.  That function should be provided in ``libaio.so``, so one problem might be that's failing to link with libaio.

Does this cause your whole build fail, or just the test to detect whether libaio is available?

I'll see if I can reproduce it.  Can you share the ``install.sh`` command you're using?

tjruwase: Yes, it is failing the entire build. Sorry for not linking [install.sh](https://github.com/microsoft/DeepSpeed/blob/master/install.sh) earlier. It is the standard script for building from source.

Thanks for the explanation. As far as I can see that c program is not well-formed since it is missing the `#include<libaio.h>` at the minimum. Also, `io_submit` takes arguments so I don't expect this call `io_submit()` to succeed. Thanks.

adammoody: Yes, that C program is not exactly well-formed, but something like that is typically enough to test whether a certain function symbol exists.  There are two problems I'd like to figure out.

First, I can imagine one gets this error if the test program fails to link with libaio or the libaio that it finds does not provide the ``io_submit`` symbol for some reason.

Second, I'd like to figure out why the failure of that test link kills the whole build rather than just disables the libaio extension.

I can force a similar type of error message if I replace ``io_submit`` with some bogus function name that I know does not exist.  However, in my case, the build seems to just disable the aio op and keep going rather than aborting the whole build.  I'm running a ``python setup.py build_dist`` command directly rather than using the ``install.sh`` script.  I still need to try that myself.

As for that second item where it kills your build, do you see any other errors messages in the ``install.sh`` build output?

stas00: add `try/except`?

stas00: as I mentioned earlier, this program may crash as well if the user doesn't cc tools installed.

adammoody: The ``ccompiler.has_function`` method doesn't seem to raise an exception here, even when the link fails due to an undefined symbol.  It just seems to return ``False`` in that case.

I'm reworking this to use ``compile`` and ``link_executable`` instead, which do seem to raise exceptions.  That should also let us catch compile errors (e.g., no compiler) separately from link errors (``io_submit`` is undefined).  That requires one to put together the test program manually.

I think that will help resolve the missing compiler problem.  However, I'm unsure whether this will help with either of the other two problems above, namely, why is ``io_submit`` missing, and why is that causing the whole build to abort rather than just disabling the aio op.  We'll keep hunting those.

stas00: You're trying to solve a very complex problem, Adam. This is super challenging, and we haven't even tested it yet on a bunch of different platforms. I'm very grateful you were inspired to work on this.

If you find a way for this compile test to gracefully fail, I think it'd be a huge help - that's to have it optional and not getting in the way of the build.

adammoody:
The ``./install.sh`` build also works in my environment, so I'm not able to reproduce the same failure mode.  This has been reworked to include try/catch statements using different ``ccompiler`` methods.  I've also added some print statements to perhaps shed more light on things.

@tjruwase , when you get a chance, would you please try your build again this lastest version?

op_builder/async_io.py
tjruwase: > Yes, it is failing the entire build. Sorry for not linking [install.sh](https://github.com/microsoft/DeepSpeed/blob/master/install.sh) earlier. It is the standard script for building from source.
> 
> Thanks for the explanation. As far as I can see that c program is not well-formed since it is missing the `#include<libaio.h>` at the minimum. Also, `io_submit` takes arguments so I don't expect this call `io_submit()` to succeed. Thanks.

Apologies for the delayed response. Below is the error I see with master (without your latest changes).  I will share the results of your branch asap. Also, to echo @stas00, thanks for doggedly taking on a very challenging problem.

```bash
worker-0: Installing collected packages: deepspeed
worker-0: Successfully installed deepspeed-0.4.4+6ae756c
worker-0: Warning: Permanently added '[192.168.0.52]:44896' (ECDSA) to the list of known hosts.
worker-0: /tmp/io_submitlnvp9xtf.c:1:1: warning: return type defaults to 'int' [-Wimplicit-int]
worker-0:  main (int argc, char **argv) {
worker-0:  ^~~~
worker-0: /tmp/io_submitlnvp9xtf.c: In function 'main':
worker-0: /tmp/io_submitlnvp9xtf.c:2:5: warning: implicit declaration of function 'io_submit' [-Wimplicit-function-declaration]
worker-0:      io_submit();
worker-0:      ^~~~~~~~~
worker-0: --------------------------------------------------
worker-0: DeepSpeed C++/CUDA extension op report
worker-0: --------------------------------------------------
worker-0: NOTE: Ops not installed will be just-in-time (JIT) compiled at
worker-0:       runtime if needed. Op compatibility means that your system
worker-0:       meet the required dependencies to JIT install the op.
worker-0: --------------------------------------------------
worker-0: JIT compiled ops requires ninja
worker-0: ninja .................. [OKAY]
worker-0: --------------------------------------------------
worker-0: op name ................ installed .. compatible
worker-0: --------------------------------------------------
worker-0: cpu_adam ............... [NO] ....... [OKAY]
worker-0: fused_adam ............. [NO] ....... [OKAY]
worker-0: fused_lamb ............. [NO] ....... [OKAY]
worker-0: sparse_attn ............ [NO] ....... [OKAY]
worker-0: transformer ............ [NO] ....... [OKAY]
worker-0: stochastic_transformer . [NO] ....... [OKAY]
worker-0: async_io ............... [NO] ....... [OKAY]
worker-0: transformer_inference .. [NO] ....... [OKAY]
worker-0: utils .................. [NO] ....... [OKAY]
worker-0: quantizer .............. [NO] ....... [OKAY]
worker-0: --------------------------------------------------
worker-0: DeepSpeed general environment info:
worker-0: torch install path ............... ['/opt/conda/lib/python3.6/site-packages/torch']
worker-0: torch version .................... 1.8.0a0+17f8c32
worker-0: torch cuda version ............... 11.1
worker-0: nvcc version ..................... 11.1
worker-0: deepspeed install path ........... ['/opt/conda/lib/python3.6/site-packages/deepspeed']
worker-0: deepspeed info ................... 0.4.4+6ae756c, 6ae756c, master
worker-0: deepspeed wheel compiled w. ...... torch 1.8, cuda 11.1
worker-0: Warning: Permanently added '[192.168.0.52]:44896' (ECDSA) to the list of known hosts.
```

adammoody: Thanks @tjruwase .  In the output about, it looks like it's printing the compiler warnings about the implicit return types, but I don't see the linker error about ``io_submit`` being undefined.  Actually, the bottom part of the output suggests that it left the async_io op enabled, which probably means it must have also linked successfully.

Do you also see the linker error in your ``install.sh`` output?

I've seen people suggest ways to silence the compile/link output from distutils to hide those warnings by redirecting stderr with ``os.dup2``.  For now, I'm not hiding that output since it should help us debug the problem.

stas00:
when it works for Tunji, please ping me and I will run a test on my box too before merging.

tjruwase:
@adammoody, I am so sorry! I previously shared a number of incorrect feedback on this PR as I totally misread my logs. Apologies for the unnecessary alarm and confusion from me and the extra work on you. Let me try to set the record straight.

1. What `self.has_function()` was reporting in the build log are compiler warnings, not compiler errors. I recognize that there is a big difference between this two. Moreover, given that the purpose of `self.has_function()`, as you explained, is simply to use linker errors to detect the presence/absence of a function symbol, then I think it is okay to ignore preceding compiler warnings since they don't prevent the linker step from succeeding.  
2. These compiler warnings did not fail the entire build, as I previously claimed. I simply did not read the previous log carefully.  This line from the previous log showed that the `async_io` is OKAY, i.e. deemed compatible by the build process. 
```bash
worker-0: async_io ............... [NO] ....... [OKAY]
```
3. Still on the previous build, the `libaio `extension (i.e., `deepspeed.ops.aio`) was not disabled and is infact usable. I verified this by successfully running the associated unit test as follows: 
```bash
pytest tests/unit/test_aio.py
======================================================= test session starts ========================================================
platform linux -- Python 3.6.10, pytest-6.1.2, py-1.9.0, pluggy-0.13.1
rootdir: 
plugins: forked-1.3.0, pythonpath-0.7.3, cov-2.10.1, hypothesis-4.50.8
collected 28 items

tests/unit/test_aio.py ............................                                                                          [100%]

================================================== 28 passed in 361.58s (0:06:01) ==================================================
```
4. With the latest updates to this branch, the main difference I see are the extra log messages showing the entire failing compiler command line. 
```bash
worker-0: Compiler default opts: ['gcc', '-pthread', '-B', '/opt/conda/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes']
worker-0: Compiler (so) default opts: ['gcc', '-pthread', '-B', '/opt/conda/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes', '-fPIC']
worker-0: Linker default opts: ['gcc', '-pthread', '-B', '/opt/conda/compiler_compat', '-Wl,--sysroot=/']
worker-0: Linker (so) default opts: ['gcc', '-pthread', '-shared', '-B', '/opt/conda/compiler_compat', '-L/opt/conda/lib', '-Wl,-rpath=/opt/conda/lib', '-Wl,--no-as-needed', '-Wl,--sysroot=/']
worker-0: Tempfile: /tmp/tmpbs6r3xbl/test.c
worker-0: Test program: int main(int argc, char** argv) { io_submit(); return 0; }
worker-0: CFLAGS: []
worker-0: creating tmp/tmpbs6r3xbl
worker-0: gcc -pthread -B /opt/conda/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -c /tmp/tmpbs6r3xbl/test.c -o tmp/tmpbs6r3xbl/test.o
worker-0: /tmp/tmpbs6r3xbl/test.c: In function 'main':
worker-0: /tmp/tmpbs6r3xbl/test.c:1:35: warning: implicit declaration of function 'io_submit' [-Wimplicit-function-declaration]
worker-0:  int main(int argc, char** argv) { io_submit(); return 0; }
worker-0:                                    ^~~~~~~~~
worker-0: LDFLAGS: []
worker-0: gcc -pthread -B /opt/conda/compiler_compat -Wl,--sysroot=/ tmp/tmpbs6r3xbl/test.o -laio -o /tmp/tmpbs6r3xbl/a.out
worker-0: Cleaning up temp directory for the test C program
```

@adammoody, once again I do apologize for the unnecessary alarm and extra work.

@stas00, please go ahead with testing on your side. 

Going forward, I wonder if we should suppress the **expected** compiler warnings (i.e., `-Wimplicit-int`, `-Wimplicit-function-declaration`) in the build process since they are expected behavior of `self.has_function()`.  What do you guys think?

tjruwase:
So I went ahead to test my proposal of suppressing these specific compiler warnings using the following change to this branch:
```bash
index 033a3b6..db0cfef 100644
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -204,7 +204,7 @@ class OpBuilder(ABC):
             #os.dup2(filestderr.fileno(), sys.stderr.fileno())

             # Attempt to compile the C program into an object file.
-            cflags = []
+            cflags = ['-Wno-implicit-int', '-Wno-implicit-function-declaration']
             if 'CFLAGS' in os.environ:
                 cflags = os.environ['CFLAGS'].split()
             print('CFLAGS:', cflags)
````

And now the corresponding build log snippet looks like the following. What do you think?
```bash
worker-0: Testing for io_submit
worker-0: Compiler default opts: ['gcc', '-pthread', '-B', '/opt/conda/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes']
worker-0: Compiler (so) default opts: ['gcc', '-pthread', '-B', '/opt/conda/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes', '-fPIC']
worker-0: Linker default opts: ['gcc', '-pthread', '-B', '/opt/conda/compiler_compat', '-Wl,--sysroot=/']
worker-0: Linker (so) default opts: ['gcc', '-pthread', '-shared', '-B', '/opt/conda/compiler_compat', '-L/opt/conda/lib', '-Wl,-rpath=/opt/conda/lib', '-Wl,--no-as-needed', '-Wl,--sysroot=/']
worker-0: Tempfile: /tmp/tmp871com4z/test.c
worker-0: Test program: int main(int argc, char** argv) { io_submit(); return 0; }
worker-0: CFLAGS: ['-Wno-implicit-int', '-Wno-implicit-function-declaration']
worker-0: creating tmp/tmp871com4z
worker-0: gcc -pthread -B /opt/conda/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -Wno-implicit-int -Wno-implicit-function-declaration -c /tmp/tmp871com4z/test.c -o tmp/tmp871com4z/test.o
worker-0: LDFLAGS: []
worker-0: gcc -pthread -B /opt/conda/compiler_compat -Wl,--sysroot=/ tmp/tmp871com4z/test.o -laio -o /tmp/tmp871com4z/a.out
worker-0: Cleaning up temp directory for the test C program
worker-0: async_io ............... [NO] ....... [OKAY]
```

stas00:
with libaio preinstalled I get:
```
    Running setup.py (path:/mnt/nvme1/code/github/00optimize/deepspeed/setup.py) egg_info for package from file:///mnt/nvme1/code/github/00optimize/deepspeed
    Created temporary directory: /tmp/pip-pip-egg-info-xsatlyf7
    Running command python setup.py egg_info
    DS_BUILD_OPS=0
    Testing for io_submit
    Compiler default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes']
    Compiler (so) default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes', '-fPIC']
    Linker default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/']
    Linker (so) default opts: ['gcc', '-pthread', '-shared', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-L/home/stas/anaconda3/envs/py38-pt19/lib', '-Wl,-rpath=/home/stas/anaconda3/envs/py38-pt19/lib', '-Wl,--no-as-needed', '-Wl,--sysroot=/']
    Tempfile: /tmp/tmp7b212fef/test.c
    Test program: int main(int argc, char** argv) { io_submit(); return 0; }
    CFLAGS: []
    creating tmp
    creating tmp/tmp7b212fef
    gcc -pthread -B /home/stas/anaconda3/envs/py38-pt19/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -c /tmp/tmp7b212fef/test.c -o tmp/tmp7b212fef/test.o
    /tmp/tmp7b212fef/test.c: In function ‘main’:
    /tmp/tmp7b212fef/test.c:1:35: warning: implicit declaration of function ‘io_submit’ [-Wimplicit-function-declaration]
        1 | int main(int argc, char** argv) { io_submit(); return 0; }
          |                                   ^~~~~~~~~
    LDFLAGS: []
    gcc -pthread -B /home/stas/anaconda3/envs/py38-pt19/compiler_compat -Wl,--sysroot=/ tmp/tmp7b212fef/test.o -laio -o /tmp/tmp7b212fef/a.out
    Cleaning up temp directory for the test C program
    Testing for io_submit
    Compiler default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes']
    Compiler (so) default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes', '-fPIC']
    Linker default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/']
    Linker (so) default opts: ['gcc', '-pthread', '-shared', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-L/home/stas/anaconda3/envs/py38-pt19/lib', '-Wl,-rpath=/home/stas/anaconda3/envs/py38-pt19/lib', '-Wl,--no-as-needed', '-Wl,--sysroot=/']
    Tempfile: /tmp/tmp24l1d9ti/test.c
    Test program: int main(int argc, char** argv) { io_submit(); return 0; }
    CFLAGS: []
    creating tmp/tmp24l1d9ti
    gcc -pthread -B /home/stas/anaconda3/envs/py38-pt19/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -c /tmp/tmp24l1d9ti/test.c -o tmp/tmp24l1d9ti/test.o
    /tmp/tmp24l1d9ti/test.c: In function ‘main’:
    /tmp/tmp24l1d9ti/test.c:1:35: warning: implicit declaration of function ‘io_submit’ [-Wimplicit-function-declaration]
        1 | int main(int argc, char** argv) { io_submit(); return 0; }
          |                                   ^~~~~~~~~
    LDFLAGS: []
    gcc -pthread -B /home/stas/anaconda3/envs/py38-pt19/compiler_compat -Wl,--sysroot=/ tmp/tmp24l1d9ti/test.o -laio -o /tmp/tmp24l1d9ti/a.out
    Cleaning up temp directory for the test C program
```

stas00:
If I uninstall libaio-dev:
```
    Running setup.py (path:/mnt/nvme1/code/github/00optimize/deepspeed/setup.py) egg_info for package from file:///mnt/nvme1/code/github/00optimize/deepspeed
    Created temporary directory: /tmp/pip-pip-egg-info-femeyzk_
    Running command python setup.py egg_info
    DS_BUILD_OPS=0
    Testing for io_submit
    Compiler default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes']
    Compiler (so) default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes', '-fPIC']
    Linker default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/']
    Linker (so) default opts: ['gcc', '-pthread', '-shared', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-L/home/stas/anaconda3/envs/py38-pt19/lib', '-Wl,-rpath=/home/stas/anaconda3/envs/py38-pt19/lib', '-Wl,--no-as-needed', '-Wl,--sysroot=/']
    Tempfile: /tmp/tmpkmtnddn5/test.c
    Test program: int main(int argc, char** argv) { io_submit(); return 0; }
    CFLAGS: []
    creating tmp/tmpkmtnddn5
    gcc -pthread -B /home/stas/anaconda3/envs/py38-pt19/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -c /tmp/tmpkmtnddn5/test.c -o tmp/tmpkmtnddn5/test.o
    /tmp/tmpkmtnddn5/test.c: In function ‘main’:
    /tmp/tmpkmtnddn5/test.c:1:35: warning: implicit declaration of function ‘io_submit’ [-Wimplicit-function-declaration]
        1 | int main(int argc, char** argv) { io_submit(); return 0; }
          |                                   ^~~~~~~~~
    LDFLAGS: []
    gcc -pthread -B /home/stas/anaconda3/envs/py38-pt19/compiler_compat -Wl,--sysroot=/ tmp/tmpkmtnddn5/test.o -laio -o /tmp/tmpkmtnddn5/a.out
    /home/stas/anaconda3/envs/py38-pt19/compiler_compat/ld: cannot find -laio
    collect2: error: ld returned 1 exit status
    Failed to link a test C program
    Cleaning up temp directory for the test C program
     [WARNING]  async_io: please install the libaio-dev package with dpkg
    Testing for io_submit
    Compiler default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes']
    Compiler (so) default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/', '-Wsign-compare', '-DNDEBUG', '-g', '-fwrapv', '-O3', '-Wall', '-Wstrict-prototypes', '-fPIC']
    Linker default opts: ['gcc', '-pthread', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-Wl,--sysroot=/']
    Linker (so) default opts: ['gcc', '-pthread', '-shared', '-B', '/home/stas/anaconda3/envs/py38-pt19/compiler_compat', '-L/home/stas/anaconda3/envs/py38-pt19/lib', '-Wl,-rpath=/home/stas/anaconda3/envs/py38-pt19/lib', '-Wl,--no-as-needed', '-Wl,--sysroot=/']
    Tempfile: /tmp/tmpbjngta83/test.c
    Test program: int main(int argc, char** argv) { io_submit(); return 0; }
    CFLAGS: []
    creating tmp/tmpbjngta83
    gcc -pthread -B /home/stas/anaconda3/envs/py38-pt19/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -c /tmp/tmpbjngta83/test.c -o tmp/tmpbjngta83/test.o
    /tmp/tmpbjngta83/test.c: In function ‘main’:
    /tmp/tmpbjngta83/test.c:1:35: warning: implicit declaration of function ‘io_submit’ [-Wimplicit-function-declaration]
        1 | int main(int argc, char** argv) { io_submit(); return 0; }
          |                                   ^~~~~~~~~
    LDFLAGS: []
    gcc -pthread -B /home/stas/anaconda3/envs/py38-pt19/compiler_compat -Wl,--sysroot=/ tmp/tmpbjngta83/test.o -laio -o /tmp/tmpbjngta83/a.out
    /home/stas/anaconda3/envs/py38-pt19/compiler_compat/ld: cannot find -laio
    collect2: error: ld returned 1 exit status
    Failed to link a test C program
    Cleaning up temp directory for the test C program
     [WARNING]  async_io: please install the libaio-dev package with dpkg
```

stas00:
So the only nuance from my side is that many Ubuntu users 
1. may not know what `dpkg` is, as `apt` is the package manager
2. `dpgk` doesn't fetch packages, `apt` does

Could we expand that lookup table to have an entry for the program that we tell users to use? e.g. on Ubuntu that would be:
```
```

alternatively just dropping the package manager name?
```
 [WARNING]  async_io: please install the libaio-dev system package
```
note I changed the wording to system package, so users won't try to do `pip install`

stas00:
But while we are at it, as I raised this issue in the past. Why are we issuing a warning and not an exception? We know DS_BUILD_AIO=1 will fail to build, and the warning is much harder to see.

Any reason why not to make it an exception during the build?

this shouldn't succeed w/o libaio, yet it does:
```
DS_BUILD_AIO=1 pip install -e . --global-option="build_ext" --global-option="-j8" --no-cache -v --disable-pip-version-check
```
and then will fail at runtime.

tjruwase:
@stas00, so should this case fail because we are building only one extension? I believe by default, we test compatibility of all extensions even though user might not need all of them.

stas00:
Yes, this is what I was suggesting. So if `DS_BUILD_AIO=1` (or BUILD_ALL) and the setup is incompatible it should assert.

So this is an additional suggestion.

And if it's just checking  for compatibility then warning is right.

tjruwase:
> Yes, this is what I was suggesting. So if `DS_BUILD_AIO=1` (or BUILD_ALL) and the setup is incompatible it should assert.
> 
> So this is an additional suggestion.
> 
> And if it's just checking for compatibility then warning is right.

Agreed. @adammoody, is this something that is easy to add to this PR or do you think it should be a separate PR? Thanks!

adammoody:
Thanks for running those tests, @tjruwase and @stas00 .

I've disabled the prints and also enabled the mechanism that redirects ``stderr`` to a file, which is a bigger hammer to silence compile and linker output.

I added a ``verbose=False`` argument to the ``has_function`` interface that lets one turn on some messages.  In particular, I think this should print the actual compile and link lines that are used, and it does not redirect ``stderr``.  So that should help in debugging, though there is no easy way to enable it as a user just yet.

@stas00 , you can see there are also except statements that can catch compile and link errors separately, in case we want to add more specific messages there.

tjruwase (approved):

adammoody:
Another potential problem with my existing code is that if someone happens to include an argument with an embedded space in ``CFLAGS`` or ``LDFLAGS``, like ``CFLAGS=-DVAL='HELLO WORLD'``, then the string ``split()`` in there will not do the right thing.  I found I had to manually add those to pass them through to the compile and link steps.  In which case, the test might report a failure when the library does actually exist.

I know Python and/or PyTorch must be doing something similar, because I know I had to set those variables to point to my ``libaio``.  With some digging, we could likely figure that out.

adammoody:
@stas00 , I'm happy to change the missing package messages about installing ``libaio-dev``.  Do you have a preference between the two you suggested?

I can also change things to throw an exception if the aio op was requested.  Though I might need some more direction to do that correctly.

stas00:
As you commented elsewhere, here is the solution: https://stackoverflow.com/a/79985/9201239

```
import shlex
shlex.split(os.environ['CFLAGS'])
```

you can also recode the above multline into:

```
cflags = shlex.split(os.environ.get('CFLAGS', ""))
```

no need for if

stas00:
> @stas00 , I'm happy to change the missing package messages about installing libaio-dev. Do you have a preference between the two you suggested?

I think the first approach would be the least misleading, the 2nd is just faster to implement but it's not as clear.

stas00:
may I suggest to use a normal push and not force push? This way all the previous experiments get preserved and can be re-used / rolled back to, or seen what has been tried and abandoned. At the end the PR gets squashed into one commit anyway, so there is no harm in having a 100 commits while working on the PR.

Also I'm not sure it preserves the code comments if the code gets wiped out via force-push, if I'm not mistaken those get lost.

Just a suggestion of course, it's your PR :)

adammoody:
> 
> 
> may I suggest to use a normal push and not force push? This way all the previous experiments get preserved and can be re-used / rolled back to, or seen what has been tried and abandoned. At the end the PR gets squashed into one commit anyway, so there is no harm in having a 100 commits while working on the PR.

Oh, sure.  No problem.  I was worried about all of those intermediate commits getting merged in, not knowing they'd be squashed.  Good to know I can keep those in.

adammoody:
> 
> 
> > @stas00 , I'm happy to change the missing package messages about installing libaio-dev. Do you have a preference between the two you suggested?
> 
> I think the first approach would be the most directive, the 2nd is just faster to implement but it's not as clear.
> 
> Thank you.

Added this naming ``apt``, ``pacman``, and ``yum``.  Is ``pacman`` the right tool to name for archlinux?

stas00:
re: archlinux - it looks like `pacman` through and through: https://wiki.archlinux.org/title/pacman#Installing_packages

re: redhat, is `yum` the standard package manager these days? I haven't used redhat in decades, but from what I remember `rpm` was a standalone package manager.

This is different with `apt`, where `dpkg` is low-level tool and `apt` is the package manager.

tjruwase:
> I can also change things to throw an exception if the aio op was requested. Though I might need some more direction to do that correctly.

@adammoody, thanks for offering to help with this. On second thought, I think it might make sense as a separate PR as we need that support for all deepspeed ops. Also, I feel this PR is probably heavy enough as is. 
```bash
grep -r DS_BUILD_ op_builder/*.py
op_builder/async_io.py:    BUILD_VAR = "DS_BUILD_AIO"
op_builder/cpu_adam.py:    BUILD_VAR = "DS_BUILD_CPU_ADAM"
op_builder/fused_adam.py:    BUILD_VAR = "DS_BUILD_FUSED_ADAM"
op_builder/fused_lamb.py:    BUILD_VAR = 'DS_BUILD_FUSED_LAMB'
op_builder/quantizer.py:    BUILD_VAR = "DS_BUILD_QUANTIZER"
op_builder/sparse_attn.py:    BUILD_VAR = "DS_BUILD_SPARSE_ATTN"
op_builder/stochastic_transformer.py:    BUILD_VAR = "DS_BUILD_STOCHASTIC_TRANSFORMER"
op_builder/transformer.py:    BUILD_VAR = "DS_BUILD_TRANSFORMER"
op_builder/transformer_inference.py:    BUILD_VAR = "DS_BUILD_TRANSFORMER_INFERENCE"
op_builder/utils.py:    BUILD_VAR = "DS_BUILD_UTILS"
```

What do you think?

adammoody:
@tjruwase , I just added a change to ``setup.py`` that might do the trick before I saw your note.  If we think it'll take some iterations, we can do that in a different PR.

adammoody:
> 
> 
> re: redhat, is `yum` the standard package manager these days? I haven't used redhat in decades, but from what I remember `rpm` was a standalone package manager.
> 
> This is different with `apt`, where `dpkg` is low-level tool and `apt` is the package manager.

I believe we use ``yum`` to install ``rpm`` packages on our systems.  It plays a role like ``apt`` where it can fetch packages from remote repositories and then install them via ``rpm``.

tjruwase:
> @tjruwase , I just added a change to `setup.py` that might do the trick before I saw your note.
> 
> [7bf5c0f](https://github.com/microsoft/DeepSpeed/commit/7bf5c0fefc290044a2cb198e436ffe7276f5d8f4)
> 
> If we think it'll take some iterations, we can do that in a different PR.

Awesome. This worked for me in that without libaio-dev this now breaks, `sudo  DS_BUILD_AIO=1 pip install -e . --global-option="build_ext" --global-option="-j8" --no-cache -v --disable-pip-version-check`

```bash
      File "/data/users/olruwase/deepseed/public/findlibaio2/setup.py", line 122, in <module>
        assert False, f"Unable to pre-compile {op_name}, perhaps disable with {env_var}=0"
    AssertionError: Unable to pre-compile async_io, perhaps disable with DS_BUILD_AIO=0
    DS_BUILD_OPS=0
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
Exception information:
  File "/opt/conda/lib/python3.6/site-packages/pip/_internal/cli/base_command.py", line 188, in _main
    status = self.run(options, args)
  File "/opt/conda/lib/python3.6/site-packages/pip/_internal/cli/req_command.py", line 185, in wrapper
    return func(self, options, args)
  File "/opt/conda/lib/python3.6/site-packages/pip/_internal/commands/install.py", line 333, in run
    reqs, check_supported_wheels=not options.target_dir
  File "/opt/conda/lib/python3.6/site-packages/pip/_internal/resolution/legacy/resolver.py", line 179, in resolve
    discovered_reqs.extend(self._resolve_one(requirement_set, req))
```

tjruwase:
@adammoody, this looks good to merge. 

@stas00, what say ye?

adammoody:
@tjruwase , squeezed in one more commit that I think should just run each ``is_compatible`` test once.  I'll put down the pencil now.

stas00:
Let me re-test

tjruwase:
> @tjruwase , squeezed in one more commit that I think should just run each `is_compatible` test once. I'll put down the pencil now.

And all the surviving DeepSpeed bugs let out a collective sigh of relief 🥇

stas00:
with the lib it's good, w/o it, I don't think the user can make any sense out of it:

```
DS_BUILD_AIO=1 pip install -e . --global-option="build_ext" --global-option="-j8" --no-cache -v --disable-pip-version-check

    Running setup.py (path:/mnt/nvme1/code/github/00optimize/deepspeed/setup.py) egg_info for package from file:///mnt/nvme1/code/github/00optimize/deepspeed
    Created temporary directory: /tmp/pip-pip-egg-info-nx64tvmv
    Running command python setup.py egg_info
      File "/mnt/nvme1/code/github/00optimize/deepspeed/setup.py", line 124, in <module>
        assert False, f"Unable to pre-compile {op_name}, perhaps disable with {env_var}=0"
    AssertionError: Unable to pre-compile async_io, perhaps disable with DS_BUILD_AIO=0
    DS_BUILD_OPS=0
WARNING: Discarding file:///mnt/nvme1/code/github/00optimize/deepspeed. Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.                                                                                                                          
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
Exception information:
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/cli/base_command.py", line 180, in _main
    status = self.run(options, args)
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/cli/req_command.py", line 205, in wrapper
    return func(self, options, args)
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/commands/install.py", line 318, in run
    requirement_set = resolver.resolve(
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 103, in resolve
    r = self.factory.make_requirement_from_install_req(
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 444, in make_requirement_from_install_req
    raise self._build_failures[ireq.link]
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 185, in _make_candidate_from_link
    self._editable_candidate_cache[link] = EditableCandidate(
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 334, in __init__
    super().__init__(
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 151, in __init__
    self.dist = self._prepare()
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 234, in _prepare
    dist = self._prepare_distribution()
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 345, in _prepare_distribution
    return self._factory.preparer.prepare_editable_requirement(self._ireq)
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/operations/prepare.py", line 622, in prepare_editable_requirement
    dist = _get_prepared_distribution(
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/operations/prepare.py", line 60, in _get_prepared_distribution
    abstract_dist.prepare_distribution_metadata(finder, build_isolation)
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/distributions/sdist.py", line 36, in prepare_distribution_metadata
    self.req.prepare_metadata()
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/req/req_install.py", line 549, in prepare_metadata
    self.metadata_directory = self._generate_metadata()
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/req/req_install.py", line 524, in _generate_metadata
    return generate_metadata_legacy(
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/operations/build/metadata_legacy.py", line 67, in generate_metadata
    call_subprocess(
  File "/home/stas/anaconda3/envs/py38-pt19/lib/python3.8/site-packages/pip/_internal/utils/subprocess.py", line 244, in call_subprocess
    raise InstallationSubprocessError(proc.returncode, command_desc)
pip._internal.exceptions.InstallationSubprocessError: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
Removed file:///mnt/nvme1/code/github/00optimize/deepspeed from build tracker '/tmp/pip-req-tracker-3wthsjb4'
Removed build tracker: '/tmp/pip-req-tracker-3wthsjb4'
```

After a very close study I finally found the assert message way before the normal traceback.
```
   AssertionError: Unable to pre-compile async_io, perhaps disable with DS_BUILD_AIO=0
       DS_BUILD_OPS=0
```
and then warnings follow, which is very odd.

My vote is at least to kill "perhaps disable with DS_BUILD_AIO=0 DS_BUILD_OPS=0" - as this doesn't suggest anything useful. I enabled these manually because I needed those. And definitely not `DS_BUILD_OPS=0` because it defeats the purpose of pre-buillding completely.

What this needs is a rethinking of why warnings are printed separately from the error and after it. Since it's illogical to the user.
But perhaps some other time.

For now, perhaps just changing the assert message to:
```
AssertionError: Unable to pre-compile async_io - can't find libaio .so and/or header files. See other messages for more details.
```
will be a good enough first step.

adammoody:
Hmm, yeah, the messages are being displayed in a different order than how they are printed, probably due to a mixing of ``stdout`` and ``stderr`` streams.  Let me take another look.  One of those is coming from ``assert`` and the others are ``self.warnings()``.  Let me take another stab at it.

adammoody:
@stas00 , how about this change?  I modified ``setup.py`` to only print the suggestion to disable an op if it sees that the environment variable for that op is not already defined.  The assert message just says that it failed to build, but offers no suggestion.

The order of those messages might still come out mixed (``stderr`` vs ``stdout``), though maybe it's less confusing.

To fix the order, we probably should have ``builder.warning()`` write to ``stderr`` or define a ``builder.error()`` function that prints to ``stdout`` like ``builder.warning()`` does.

stas00:
>how about this change? I modified setup.py to only print the suggestion to disable an op if it sees that the environment variable for that op is not already defined. The assert message just says that it failed to build, but offers no suggestion.

Looks good. thank you!

Trying to sync std streams can be a madness, when I mentioned re-think I was thinking that different components do their work, and print nothing and return success/failure and any messages instead to the caller. Then the manager that runs these components has the freedom to print any messages it collected from the components in the order it needs.

But again, it's probably a job for another time.

adammoody:
It looks like ``setup.py`` only uses ``assert`` to bail out when it detects a fatal error.  I'm guessing that ``assert`` is what generates the (noisy) python stack trace.  We could avoid that if there is a better way to exit, but I don't know the convention for bailing with an error from a setup script offhand.

stas00:
adding a few empty lines before and after the assert?

e.g. pytorch recently started doing logs with banners that stand out from the other noise, examples:


```
******************************************
  /home/ubuntu/tmp/test_script.py FAILED
==========================================
Root Cause:
[0]:
  time: 2021-07-23_02:22:07
  rank: 0 (local_rank: 0)
  exitcode: 1 (pid: 43258)
  error_file: <N/A>
  msg: "Process failed with exitcode 1"
==========================================
Other Failures:
[1]:
  time: 2021-07-23_02:22:07
  rank: 1 (local_rank: 1)
  exitcode: 1 (pid: 43259)
  error_file: <N/A>
  msg: "Process failed with exitcode 1"
[2]:
  time: 2021-07-23_02:22:07
  rank: 2 (local_rank: 2)
  exitcode: 1 (pid: 43260)
  error_file: <N/A>
  msg: "Process failed with exitcode 1"
[3]:
  time: 2021-07-23_02:22:07
  rank: 3 (local_rank: 3)
  exitcode: 1 (pid: 43261)
  error_file: <N/A>
  msg: "Process failed with exitcode 1"
******************************************

**********************************************************************
               CHILD PROCESS FAILED WITH NO ERROR_FILE                
**********************************************************************
CHILD PROCESS FAILED WITH NO ERROR_FILE
Child process 3094556 (local_rank 0) FAILED (exitcode 1)
Error msg: Process failed with exitcode 1
Without writing an error file to <N/A>.
While this DOES NOT affect the correctness of your application,
no trace information about the error will be available for inspection.
Consider decorating your top level entrypoint function with
torch.distributed.elastic.multiprocessing.errors.record. Example:

  from torch.distributed.elastic.multiprocessing.errors import record

  @record
  def trainer_main(args):
     # do train
**********************************************************************

```

adammoody:
I can revert that with a new commit if needed, but this last commit adds a ``abort`` function that prints and error message in red and then calls ``sys.exit(1)`` rather than ``assert``.

stas00:
1. I'm not sure if all terminals support color codes - typically I don't see those being used other than in some small projects. It's up to @tjruwase to approve or not.

2. `sys.exit(1)` can break the "super"-system that calls `setup.py` - e.g. if installing some other package via `pip` and `deepspeed` is just a dependency of some package, I think this would be a very unwelcome behavior.

adammoody:
Ah, I see.  Great points, @stas00 .

A call to ``sys.exit(1)`` definitely seems like a bad idea.

I borrowed the color code idea from the ``builder.warning()`` code:
https://github.com/microsoft/DeepSpeed/blob/97f7ed9e9836f2894387c3a7e38e7db2a8e5dcfc/op_builder/builder.py#L13-L15

I will revert this and go back to the ``assert``.

tjruwase:
I am fine with color code actually. Yes, `sys.exit()` seems an overkill.

adammoody:
From the previous commit https://github.com/microsoft/DeepSpeed/pull/1250/commits/0e4cde6b0a650591c3fafface7e27b4efd9aad4f, we could replace the ``sys.exit(1)`` with an ``assert False, msg``.

adammoody:
Ok, I'll stop now, for real this time :-)

Thanks for all of your help with this, @tjruwase and @stas00 .

stas00:
I think your PR got hit by never starting CI - happened a few times to my PRs, you probably need to push an empty commit.
```
git commit --allow-empty -m "Trigger CI"
git push
```

tjruwase:
I just started the CI.

PR #1254, opened 2021-07-28
Activation checkpointing improvements

tjruwase:
1) Attempt to checkpoint all activations
2) Make cpu checkpoint independent of partitioned activations
3) Refactoring and clean ups


deepspeed/runtime/utils.py
```diff
@@ -59,24 +59,52 @@ def set_random_seed(seed):
...
+        item: tensor to move or (possibly nested) container of tensors to move.
+        device: target device
+        criterion_func: Function to restrict move operation to items meet criterion
+
+    Returns:
+        None
+    """
+    if criterion_func(item):
+        device_copy = item.to(device)
+        item.data = device_copy.data
```
eltonzheng, line 100: not sure why need to overwrite item.data to remove it instead of letting python/pytorch to decide by reference count, and not sure if this is absolutely safe, that item's other properties are always consistent with new "data" after overwriting.

deepspeed/runtime/utils.py
```diff
@@ -59,24 +59,52 @@ def set_random_seed(seed):
...
+        return tuple([copy_to_device(v, device, criterion_func) for v in item])
+    elif isinstance(item, dict):
+        return {k: copy_to_device(v, device, criterion_func) for k, v in item.items()}
+    else:
+        return item
+
+
+def move_to_device(item, device, criterion_func):
+    """
+    Move tensor on to specified device by changing the storage.
```
eltonzheng, line 88: should remove "on"?

deepspeed/runtime/activation_checkpointing/checkpointing.py
```diff
@@ -348,6 +353,132 @@ def merge_tensors(tensor_objects, non_tensor_objects, tensor_flags):
     return tuple(merged_objects)
 
 
+def is_activation_to_checkpoint(item):
+    """
+        Is an activation to be checkpointed
+    """
+    global mp_size
+    return torch.is_tensor(item) and item.is_floating_point() and item.numel() >= mp_size
```
eltonzheng, line 361: don't know why using condition of "numel() > mp_size", could you add some comments here?

eltonzheng (approved):

PR #1261, opened 2021-07-29
[Doc] round_robin_gradients

tjruwase:
Document ZeRO-2 offload optimization: `round_robin_gradients`.


stas00:
Looks good - thank you for adding it, @tjruwase!

Did you want to add why is it off by default - i.e. may fail at times? Or is it just a matter of time before it's always working?

tjruwase:
We plan to make it work always. I don't see any fundamental reason why it shouldn't, it just needs testing in more scenarios.

PR #1263, opened 2021-07-29
Correctness fix PP+ZeRO for gradient accumulation + updates from master

jeffra:
* Correctness fix PP+ZeRO for gradient accumulation
* Cherry picked round robin grad partitioning fixes from master
* Cherry picked ignore overlap/contiguous grad settings for ZeRO-1 from master

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1176,6 +1176,9 @@ class DeepSpeedEngine(Module):
         return loss
 
     def allreduce_gradients(self, bucket_size=MEMORY_OPT_ALLREDUCE_SIZE):
+        # Pass (PP) gas boundary flag to optimizer (required for zero)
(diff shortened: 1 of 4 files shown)


ShadenSmith (approved):

deepspeed/runtime/engine.py
```diff
@@ -1163,6 +1176,10 @@ def forward(self, *inputs, **kwargs):
         return loss
 
     def allreduce_gradients(self, bucket_size=MEMORY_OPT_ALLREDUCE_SIZE):
+        # Pass (PP) gas boundary flag to optimizer (required for zero)
+        self.optimizer.is_gradient_accumulation_boundary = self.is_gradient_accumulation_boundary(
```
tjruwase, line 1180: Is `self.optimizer` guaranteed to have `is_gradient_accumulation_boundary` attribute?

tjruwase: We traced the break to this line. Crash went away after commenting this line.

PR #1271, opened 2021-08-02
Use mpu in DeepSpeedConfig() call

tjruwase:
Calling DeepSpeedConfig() without the available mpu can lead to avoidable assertions concerning batch size and gradient accumulation steps. This avoidable assertions occur because mpu determines world size.

### DIFF
--- a/deepspeed/runtime/zero/partition_parameters.py
+++ b/deepspeed/runtime/zero/partition_parameters.py
@@ -325,6 +325,7 @@ class Init(InsertPostInitMethodToModuleSubClasses):
                  mem_efficient_linear=True,
                  remote_device=None,
                  pin_memory=False,
+                 config_dict_or_path=None,
                  config=None,
                  enabled=True,
                  dtype=None,
@@ -348,8 +349,9 @@ class Init(InsertPostInitMethodToModuleSubClasses):
             pin_memory (bool, optional): Potentially increase performance by
                 using pinned memory for model weights. ``remote_device`` must be
                 ``"cpu"``. Defaults to ``False``.
-            config (``json file`` or dict, optional): If provided, provides configuration
+            config_dict_or_path (dict or ``json file``, optional): If provided, provides configuration
                 for swapping fp16 params to NVMe.
+            config (dict or ``json file``, optional): Deprecated, use config_dict_or_path instead.
             enabled (bool, optional): If ``False``, this context has no
                 effect. Defaults to ``True``.
             dtype (``dtype``, optional): Can be used to change the data type of the parameters.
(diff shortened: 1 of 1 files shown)


tjruwase:
@stas00 FYI

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -215,11 +215,16 @@ def recurse(cl):
 # Inserts _post_init_method at the end of init method
 # for all sub classes of torch.nn.Module
 class InsertPostInitMethodToModuleSubClasses(object):
-    def __init__(self, enabled=True, mem_efficient_linear=True, config=None, dtype=None):
+    def __init__(self,
+                 enabled=True,
+                 mem_efficient_linear=True,
+                 ds_config=None,
```
stas00, line 221: Won't this be a non-back-compat change? i.e. you're proposing to rename a key in a method of a public API.

tjruwase: I don't think this is public. Are you referring to [this](https://deepspeed.readthedocs.io/en/latest/zero3.html#deepspeed.zero.Init)?

stas00: Ah, you're right, this is its super-class's init that you're proposing to change. My mistake.

Should we rename it in `zero.Init` too, and alias the old config name so that the 2 APIs are consistent?

tjruwase: They are not the same type: `ds_config = DeepSpeedConfig(config, ...)`. Is this what you are referring to?

stas00: I hear you.

This is very confusing since `zero.Init` takes config arg which is typically called `ds_config` - the file that is.

Perhaps it'd help to rename things so that they are more intuitive. I can't tell a difference between `config` and `ds_config` and `_ds_config` - they all look the same to me.

tjruwase: Fair point. Will think a bit about renaming. Thanks.

tjruwase: How about

- `config`, `ds_config`, _and `ds_config `=> configuration json or dict from client world into DeepSpeed API

- `engine_config `=> output of DeepSpeedConfig(`ds_config`), internal to DeepSpeed, represents deepspeed engine configuration

stas00: But the 2 are essentially the same, except one is source and the other is the resulting object.

Proposal:

- `config_dict_or_path`: ``json file`` or dict,
- `config`: DS config object

or if `ds_` prefix is even more intuitive then:

- `ds_config_dict_or_path`: ``json file`` or dict,
- `ds_config`: DS config object

-----

as an example we use `model` and `model_name_or_path` in HF `transformers`. the former is the object, the latter is the source the latter is built from.

tjruwase: I prefer the `ds_`  prefix versions, thanks. Also, thanks for sharing the HF approach, that is helpful. 

Since the `zero.Init()` API is affected, you suggested we simply alias the current `config` argument, right?

stas00: alias and probably deprecate the old name.

Issue #1279, opened 2021-08-05
Suggestion: Support CPU-only environments/remove CUDA requirement


peterukk:
Hi,

I'm training neural networks on a small CPU-only workstation, and I wanted to use the DeepSpeedCPUAdam optimizer to accelerate the training, but noticed that CUDA is required:

```
 File "...python3.9/site-packages/deepspeed/ops/op_builder/builder.py", line 37, in installed_cuda_version
    assert cuda_home is not None, "CUDA_HOME does not exist, unable to compile CUDA op(s)"
```

Making CUDA optional would be a great feature. In my own field (climate modeling and weather forecasting), machine learning is an increasingly popular topic but GPU's and deep learning are often overkill. (Right now, I'm training neural nets with 1-2 hidden layers and 10-20 neurons, which is quite efficient on CPU's). Furthermore, support for non-NVIDIA GPU's should be relevant in the longer term. 

I realize this is presented as a deep learning library, but dropping the CUDA requirement for features which do not really require it might still be useful. Thanks!

tjruwase:
@peterukk, are you able to help test this PR #1280?

peterukk:
Wow that's fast action! I'm still getting the same error. 

`(py39) [puk@3ymlvp2:~]$ pip3 install git+https://github.com/microsoft/DeepSpeed.git@cpu-adam/optional_CUDA-copy`
...
```
import torch
from torch import nn
import pytorch_lightning as pl
from pytorch_lightning.plugins import DeepSpeedPlugin
from deepspeed.ops.adam import DeepSpeedCPUAdam
from torch.utils.data import DataLoader, TensorDataset

class MLP_cpu(pl.LightningModule):    
      
    def __init__(self, nx, ny, nneur=8, learning_rate=None, SequentialModel=None):
      super().__init__()
      if SequentialModel==None:
          self.layers = nn.Sequential(
            nn.Linear(nx, nneur),
            nn.ReLU(),
            nn.Linear(nneur, nneur),
            nn.ReLU(),
            nn.Linear(nneur, ny)
          )
      else:
          self.layers = SequentialModel
      self.mse = nn.MSELoss()
      # Hyperparameters
      self.learning_rate = learning_rate
      self.save_hyperparameters()
      
    def forward(self, x):
      return self.layers(x)
    
    def training_step(self, batch, batch_idx):
      ...
      return loss
  
    def validation_step(self, batch, batch_idx):
        ...
        return loss
      
[...]

PR #1280, opened 2021-08-05
CPU-Adam: add compile-flag to enable param-copy from CPU to GPU

RezaYazdaniAminabadi:
This addresses https://github.com/microsoft/DeepSpeed/issues/1279


csrc/adam/cpu_adam.cpp
```diff
@@ -102,11 +103,16 @@ void Adam_Optimizer::Step(float* _params,
             SIMD_STORE(_exp_avg + i, momentum_4.data);
             SIMD_STORE(_exp_avg_sq + i, variance_4.data);
         }
+#if defined(__ENABLE_CUDA__)
```
tjruwase, line 106: I believe all the cuda and cublas header files should be guarded as well.  We would probably need to test the build in a CPU-only environment.

RezaYazdaniAminabadi: I agree, I will add more guards for them too.

op_builder/builder.py
```diff
@@ -267,6 +273,12 @@ def cpu_arch(self):
             return '-mcpu=native'
         return '-march=native'
 
+    def is_cuda_available(self):
+        if torch.cuda.is_available():
```
tjruwase, line 277: Will this not throw an import exception if torch.cuda is not available? Should this be in a try/except?

RezaYazdaniAminabadi: good point 👍 I will add that. Thanks

molly-smith:
Closing since https://github.com/microsoft/DeepSpeed/pull/2507 should address this.

PR #1284, opened 2021-08-06
Use correct input size for splits

### LINKED ISSUE #1255: Runtime error when enable memory-centric-tiled-linear
Hi team,

I'm running the [example](https://github.com/microsoft/DeepSpeedExamples/blob/25d73cf73fb3dc66faefa141b7319526555be9fc/Megatron-LM-v1.1.5-ZeRO3/examples/ds_pretrain_gpt2-zero3.sh) on 8 * 40GB A100 Node,

the configurations are 

Screen Shot 2021-07-28 at 8 48 35 PM

and the stacktrace is

Screen Shot 2021-07-28 at 8 41 53 PM

it seems that the args --memory-centric-tiled-linear and --tile-factor are the root cause. So Anyone can help me to fix this problems? Thanks.

tjruwase:
Compute tile size on-the-fly from input. Fix #1255.


samyam (approved):
Looks good. Thank you Tunji.

PR #1286, opened 2021-08-07
FP16 fused and unfused grad norm query.

ShadenSmith:


ShadenSmith:
This still needs FP32 and ZeRO. And unit tests :-).

deepspeed/runtime/engine.py
```diff
@@ -256,6 +258,30 @@ def set_train_batch_size(self, train_batch_size):
         self._config.train_batch_size = train_batch_size
         self._config.gradient_accumulation_steps = new_gas
 
+    def _compute_global_grad_norm(self):
+        params = [p for p in self.module.parameters() if p.grad is not None]
```
stas00, line 262: would it be a bit more efficient if tied params not be not calculated more than once? Probably, something like:
```
params = dict((p.data_ptr(), p) for p in self.module.parameters() if p.grad is not None).values()
```
or it might help to have a wrapper to do that, since this is a handy util.

But please verify that I got it write. Thanks.

jeffra:
This ended up being added in a later PR

Issue #1265, opened 2021-07-30
Blocking issue when using deepspeed inference(maybe mutex or nccl issue)


switiz:
### Description
Dear Deepspeed

I have some issue when using model parallel(inference engine)
sometimes gpu utilization is fixed to 100% and code is hanged
so i made test code and test deepspeed engine. here's my test code. 

### TestCode
```
import os
import deepspeed
import torch
import transformers
from transformers import pipeline, AutoTokenizer

def init():
    local_rank = int(os.getenv('LOCAL_RANK', '0'))
    world_size = int(os.getenv('WORLD_SIZE', '1'))
    generator = pipeline(
        'text-generation', model='EleutherAI/gpt-neo-2.7B', device=local_rank)
    generator.model = deepspeed.init_inference(generator.model,
                                            mp_size=world_size,
                                            dtype=torch.float,
                                            replace_method='auto')
    return generator

def predict(text, max_len):
    top_k = 50
    temperature = 1.0
    top_p = 1.0
    return_seq = 1
    string = generator(text, do_sample=True, min_length=50, max_length=max_len, top_k=top_k, temperature=temperature, top_p=top_p, num_return_sequences=return_seq,
                            pad_token_id=3)
    if torch.distributed.get_rank() == 0:
        print(string)


if __name__ == '__main__':
    generator = init()
    text = 'a'
    seq = 2023
    for i in range(2, seq):
        print(f'##### max_len: {i}')
        predict(text, i)

```

### DS_Report
```
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
cpu_adam ............... [NO] ....... [OKAY]
fused_adam ............. [NO] ....... [OKAY]
fused_lamb ............. [NO] ....... [OKAY]
sparse_attn ............ [NO] ....... [OKAY]
transformer ............ [NO] ....... [OKAY]
stochastic_transformer . [NO] ....... [OKAY]
async_io ............... [NO] ....... [OKAY]
transformer_inference .. [NO] ....... [OKAY]
utils .................. [NO] ....... [OKAY]
[...]

PR #1309, opened 2021-08-16
Add more synchronizations and barriers for the multi-gpu inference case

RezaYazdaniAminabadi:
Attempting to resolve gpu-halt issue: https://github.com/microsoft/DeepSpeed/issues/1265

### DIFF
--- a/deepspeed/inference/engine.py
+++ b/deepspeed/inference/engine.py
@@ -206,7 +206,6 @@ class InferenceEngine(Module):
                         input = input.to(torch.cuda.current_device())
                         if not input.is_contiguous():
                             input = input.contiguous()
-
                 for k in kwargs:
                     if torch.is_tensor(kwargs[k]):
                         kwargs[k] = kwargs[k].to(torch.cuda.current_device())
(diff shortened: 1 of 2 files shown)


deepspeed/inference/engine.py
```diff
@@ -213,18 +204,16 @@ def forward(self, *inputs, **kwargs):
...
                 for k in kwargs:
                     if torch.is_tensor(kwargs[k]):
                         kwargs[k] = kwargs[k].to(torch.cuda.current_device())
-                        if self.mp_world_size > 1:
-                            if not kwargs[k].is_contiguous():
-                                kwargs[k] = kwargs[k].contiguous()
-                            dist.broadcast(kwargs[k], 0)
+                        if not kwargs[k].is_contiguous():
+                            kwargs[k] = kwargs[k].contiguous()
+                        dist.broadcast(kwargs[k], 0)
```
hyunwoongko, line 214: Why didn't you remove this line too?
I think we don't need to use broadcast operation.

hyunwoongko: similar with https://github.com/NVIDIA/Megatron-LM/blob/68797d90935354cf8510dd84711ab6ef5db095f1/megatron/mpu/mappings.py#L85

RezaYazdaniAminabadi: We need to broadcast the inputs all across the model-parallel ranks since after the first token being generated, they might not be synched. You can check this, by removing the broadcasts and run the GPT-Neo test, you will see that the two GPUs print different results. I think this is different than the example you are bringing from the Megatron, they use it mainly to inject all-reduce in the backward.

Issue #326, opened 2020-08-21
Regarding lack of drop_last in Deepspeed Trainloader


rsn870:
Hi ,

After encountering problems during training with my custom dataset using Deepspeed Initialize , I have narrowed the problem to the fact that since the number of images in my dataset is not divisible by the batch size , Deepspeed is throwing an error. This is easily resolvable in torch Dataloader by setting drop_last = True , the Deepspeed.pt.Dataloader source code however does not use this argument at all . 

Could this be added to the source please ?

If not , please suggest an alternative means for training

tjruwase:
Hi, thanks for reporting this issue. We will work on this as soon as possible. However, if you can prepare a PR, that would be great as well.

rsn870:
Sure , thanks for your prompt response . I will try preparing a PR asap

PR #1321, opened 2021-08-24
Added drop_last to DeepSpeedDataLoader

hxri:
This solves issue #326

The feature to drop the last incomplete batch while using the DeepSpeedDataLoader was not available. This raised an issue when the dataset size is not divisible by the batch size. The DataLoader in torch.utils.data does have the function hence adding this to the DeepSpeed DataLoader will now enable the drop_last parameter.

Default behaviour - 
drop_last = False

To drop the last incomplete batch - 
drop_last = True


hxri:
@jeffra Please review

deepspeed/runtime/dataloader.py
```diff
@@ -41,7 +41,8 @@ def __init__(self,
                  num_local_io_workers=None,
                  data_sampler=None,
                  data_parallel_world_size=None,
-                 data_parallel_rank=None):
+                 data_parallel_rank=None,
+                 drop_last=False):
```
tjruwase, line 45: Thanks for fixing this issue. But I have a comment. 

If this is hardcoded to False, does it not retain the original problem in #326? I think it would be better to make this a ds_config option that could be passed into DeepSpeedDataLoader [here](https://github.com/microsoft/DeepSpeed/blob/aa1212910c8a067e4014e157627fa80b696e824b/deepspeed/runtime/engine.py#L1153)?

hxri: I left it default to "False" because in the torch.utils.data, it's "False" by default. Again if anyone encounters an error due to  "dataset size is not divisible by the batch size", and needs to use the feature they can always pass drop_last=True in [here](https://github.com/microsoft/DeepSpeed/blob/aa1212910c8a067e4014e157627fa80b696e824b/deepspeed/runtime/engine.py#L1153).

Is this okay or should I implement it as a ds_config?

hxri:
@tjruwase 

Kindly review the changes that I have made to [test_data.py](https://github.com/microsoft/DeepSpeed/blob/a15bc27b63a58d11c6d2ef2f3bae046852f6bc0c/deepspeed/runtime/config.py), in order to test the dataloader_drop_last functionality. 

Thanks!

tests/unit/test_data.py
```diff
@@ -9,3 +14,42 @@ def test_repeating_loader():
...
+
+    model = SimpleModel(hidden_dim)
+
+    def random_dataloader(model, total_samples, hidden_dim, device, drop_last, dtype=torch.half):
+        batch_size = model.train_micro_batch_size_per_gpu()
+        train_data = torch.randn(total_samples, hidden_dim, device=device, dtype=dtype)
+        train_label = torch.empty(total_samples,
+                                dtype=torch.long,
+                                device=device).random_(hidden_dim)
+        train_dataset = torch.utils.data.TensorDataset(train_data, train_label)
```
tjruwase, line 35: Can you please refactor this into a `random_dataset()` in `test/unit/simple_model.py`?

tjruwase (approved):

tjruwase:
Thanks @hxri. It looks like yapf is still failing.

tests/unit/test_data.py
hxri: @tjruwase I am sorry but I couldn't understand your comment.

The [random_dataloader()](https://github.com/microsoft/DeepSpeed/blob/18d89c62f26c428953090aebc84c858b77fcaebd/tests/unit/test_data.py#L29-L37) in tests/unit/test_data.py is actually refactored from [random_dataloader()](https://github.com/microsoft/DeepSpeed/blob/274c375c87f35d1c7509dfc6a14f8a84d5efab84/tests/unit/simple_model.py#L181-L189) in test/unit/simple_model.py. 

The changes I made there is to add the `dataloader_drop_last = True` Argument

or is it that you need me to use 
```
train_loader = torch.utils.data.DataLoader(train_dataset, batch_size=batch_size) 
```
instead of 
```
train_loader = DeepSpeedDataLoader(train_dataset,
                                           batch_size=batch_size,
                                           dataloader_drop_last=drop_last)
```

Please correct me if I am wrong.

tjruwase: Sorry for not being clear. What I meant is not to reimplement [random_dataloader()](https://github.com/microsoft/DeepSpeed/blob/6eb13a5245599cc5f76fec72aee0d88edfaeeb04/tests/unit/test_data.py#L29) in test_data.py, which essentially duplicates [random_dataloader()](https://github.com/microsoft/DeepSpeed/blob/336dd089e5e11d46daf3abfad048d71b72912504/tests/unit/simple_model.py#L181) in simple_model.py. Instead, why not do the following

1. In simple_model.py, implement random_dataset() to provide [these](https://github.com/microsoft/DeepSpeed/blob/336dd089e5e11d46daf3abfad048d71b72912504/tests/unit/simple_model.py#L182-L187) core logic of random_dataloader().
2. In simple_model.py, use the random_dataset() in random_dataloader() to avoid duplication.
3. In test_data.py, delete random_dataloader(), and use random_dataset() directly in _test_dataloader_drop_last() such as below:
```python
train_dataset = random_dataset(model, ...)
data_loader = DeepSpeedDataLoader(train_dataset, ...)
```
Will that work? Thanks.

hxri: @tjruwase 

Thanks a lot for the detailed explanation. I have made the changes. Please review.

hxri:
> Thanks @hxri. It looks like yapf is still failing.

I have corrected all formatting issues. But I am not able to see the exact problem in yapf test [logs](https://github.com/microsoft/DeepSpeed/pull/1321/checks?check_run_id=3453614909).

hxri:
@tjruwase 

I have corrected the yapf issue by running pre-commit with clang installed (had to try different ways to install clang in my machine). Now it seems to be passing all tests locally.

Thanks!

bin/deepspeed
```diff
@@ -1 +1 @@
-ds
\ No newline at end of file
+ds
```
tjruwase, line 1: I am sure you didn't modify these 4 files. Can you please remove them from the PR?

bin/deepspeed.pt
```diff
@@ -1 +1 @@
-ds
\ No newline at end of file
+ds
```
tjruwase, line 1: Same here

deepspeed/ops/csrc
```diff
@@ -1 +1 @@
-../../csrc
\ No newline at end of file
+../../csrc
```
tjruwase, line 1: Same here

deepspeed/ops/op_builder
```diff
@@ -1 +1 @@
-../../op_builder
\ No newline at end of file
+../../op_builder
```
tjruwase, line 1: Same here.

tests/unit/simple_model.py
```diff
@@ -178,13 +178,18 @@ def forward(self, x, y, **kwargs):
         return hidden_dim
 
 
-def random_dataloader(model, total_samples, hidden_dim, device, dtype=torch.half):
-    batch_size = model.train_micro_batch_size_per_gpu()
+def random_dataset(total_samples, hidden_dim, device, dtype=torch.half):
```
tjruwase, line 181: Thanks for this refactoring.

tests/unit/test_data.py
```diff
@@ -9,3 +14,35 @@ def test_repeating_loader():
         assert next(loader) == 1
         assert next(loader) == 2
         assert next(loader) == 3
+
+
+def test_dataloader_drop_last(tmpdir):
```
tjruwase, line 19: I think the coverage should be improved using [parameterization](https://github.com/microsoft/DeepSpeed/blob/336dd089e5e11d46daf3abfad048d71b72912504/tests/unit/test_fp16.py#L609). In particular, I think at least these four cases should be covered where drop_last=[true,False] and total_samples is (not) divisible by train_batch_size. For example, assuming total_samples=50:

1. drop_last=True, train_batch_size=1
2. drop_last=True, train_batch_size=4
3. drop_last=False, train_batch_size=1
4. drop_last=False, train_batch_size=4

There might be other corner cases (e.g., world_size=2), but I think those 4 are required. Parametrization is heavily used in other unit tests, in case you are not familiar with it. Thanks.

hxri: Sure I will add that too. Thanks!

hxri: Done. Can you please review @tjruwase Thanks!

tests/unit/test_data.py
```diff
@@ -9,3 +15,44 @@ def test_repeating_loader():
...
+    hidden_dim = 10
+
+    model = SimpleModel(hidden_dim)
+
+    @distributed_test(world_size=[1])
+    def _test_dataloader_drop_last(args, model, hidden_dim):
+        optimizer = torch.optim.AdamW(params=model.parameters())
+        batch_size = model.train_micro_batch_size_per_gpu()
+        model, _, _, _ = deepspeed.initialize(args=args,
+                                              train_batch_size=train_batch_size,
```
tjruwase, line 43: This will unfortunately break since `train_batch_size` is not an argument of `deepspeed.intialize()`.

hxri: My bad. Corrected it.

tests/unit/test_data.py
```diff
@@ -9,3 +15,44 @@ def test_repeating_loader():
...
+                         [(1,
+                           True),
+                          (4,
+                           True),
+                          (1,
+                           False),
+                          (4,
+                           False)])
+def test_dataloader_drop_last(tmpdir, train_batch_size, drop_last):
+    config_dict = {
```
tjruwase, line 30: I am so sorry that I did not communicate clearly previously. I am sorry for the extra work. 
Let me try again and hopefully. 

1. `config_dict `should be like 
```python
{ 'train_batch_size': train_batch_size, 'dataloader_drop_last': drop_last, ....}
```
This will allow `train_batch_size `and `dataloader_drop_last `to pass into `deepspeed.initialize()` normally via `args`. 

2. `train_dataset `should be passed directly to deepspeed.initialize() as [training_data](https://github.com/microsoft/DeepSpeed/blob/336dd089e5e11d46daf3abfad048d71b72912504/deepspeed/__init__.py#L66) arg. This will validate that all the changes you made in the library are working correctly. The returned [training_data_loader](https://github.com/microsoft/DeepSpeed/blob/336dd089e5e11d46daf3abfad048d71b72912504/deepspeed/__init__.py#L159) value should be used in the rest of unit test. Apologies that I previously suggested that you should use `DeepSpeedDataLoader` directly in the unit test, it is not necessary. 

Please let me know of anything that is not clear. Thanks.

tjruwase: Great question. [args_from_dict()](https://github.com/microsoft/DeepSpeed/blob/ddffbae021fef24bf64945d45ace9bac7e08508d/tests/unit/simple_model.py#L230) converts the `config_dict` to a json file and sets `args.deepspeed_config` to the file path. This is how fields in `config_dict` get passed into the DeepSpeed engine. 

The final bit of code looks good.

tjruwase: One more thing, it seems you will need to enable fp16 in config_dict, like [this](https://github.com/microsoft/DeepSpeed/blob/ddffbae021fef24bf64945d45ace9bac7e08508d/tests/unit/test_fp16.py#L317-L319).

tjruwase: Also, is it possible to run the unit tests on your box following these [guide](https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#unit-tests)?

hxri: @tjruwase 

Done!

tjruwase: Cool. Thanks. I will launch the CI again on this side.

tests/unit/test_data.py
```diff
@@ -9,3 +15,46 @@ def test_repeating_loader():
...
+    }
+    args = args_from_dict(tmpdir, config_dict)
+    hidden_dim = 10
+
+    model = SimpleModel(hidden_dim)
+
+    @distributed_test(world_size=[1])
+    def _test_dataloader_drop_last(args, model, hidden_dim):
+        optimizer = torch.optim.AdamW(params=model.parameters())
+        batch_size = model.train_micro_batch_size_per_gpu()
```
tjruwase, line 46: This line is failing unit tests because SimpleModel does not have this attribute.
image

A solution is to pass batch into `_test_dataloader_drop_last()`.

tjruwase: On a second look, it seems you don't need `batch_size` at all, am I correct? I think you can just delete that line.

hxri: @tjruwase 

I have a few doubts.

1. Should I declare the `batch_size = N` or should I use the `model.train_micro_batch_size_per_gpu()`, which seems to be not working since `SimpleModel.py` doesn't have that function.
2.  If I statically declare the `batch_size` and pass It to `_test_dataloader_drop_last()`, where do I have to use it inside the function? since all the functions inside it don't have such a parameter.

Thanks!

hxri: > On a second look, it seems you don't need `batch_size` at all, am I correct? I think you can just delete that line.

Sure! I was wondering the same. My bad.

Thank you!

tjruwase: I don't think you need to use `batch_size` at all since you are not constructing a dataloader, DeepSpeed will do that for you. DeepSpeed knows how to extract `batch_size` from the json config as you see [here](https://github.com/microsoft/DeepSpeed/blob/ddffbae021fef24bf64945d45ace9bac7e08508d/deepspeed/runtime/engine.py#L1179).

hxri: Got it! I am learning a lot. Thank you for your patience and support,

tests/unit/simple_model.py
```diff
@@ -178,13 +178,18 @@ def forward(self, x, y, **kwargs):
...
     train_data = torch.randn(total_samples, hidden_dim, device=device, dtype=dtype)
     train_label = torch.empty(total_samples,
                               dtype=torch.long,
                               device=device).random_(hidden_dim)
     train_dataset = torch.utils.data.TensorDataset(train_data, train_label)
+    return train_dataset
+
+
+def random_dataloader(model, total_samples, hidden_dim, device):
```
tjruwase, line 190: Looks like `dtype=torch.half` was dropped from `random_dataloader()` signature. Please restore.

hxri: @tjruwase Made the changes. Thanks!

hxri: @tjruwase The unit test fails again. Says `SimpleModel` has no attribute `device`.

Should I pass in the config, `device = 'cpu'`?

tjruwase: Good catch! I just noticed this myself. Please `torch.cuda.curent_device()`.

hxri: Done!

tjruwase: So, it turns out that torch.cuda.current_device() will not work because of a pre-existing problem with our unit tests unrelated to this PR. So in the meantime, I have pushed a change using torch.device('cpu'), as your earlier suggested, to unblock this PR. I will address the other issue in our unit tests later.

In case you are curious, our unit test problem has to do with cuda initialization like this [one](https://github.com/pytorch/pytorch/issues/21092).

tjruwase: Okay, CI has completed and all looks good. I am ready to merge whenever you are. Please let me know.

hxri: @tjruwase Thanks a lot! I am ready to merge.

PR #1322, opened 2021-08-24
[zero] restore fp16 params if no zero ckpts available

jeffra:
In fine-tuning scenarios the user often wants to only load the model parameter checkpoints and may not have the zero optimizer states on disk since they are not needed. This fixes a bug where the weights are not properly restored from the ckpt if the zero optimizer states are not present on disk.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1884,9 +1884,11 @@ class DeepSpeedEngine(Module):
                                                          load_module_only=load_module_only)
 
         if self.zero_optimization() and load_path is not None:
-            self._load_zero_checkpoint(load_dir,
+            success = self._load_zero_checkpoint(load_dir,
                                        tag,
                                        load_optimizer_states=load_optimizer_states)
+            if not success:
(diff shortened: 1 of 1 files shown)


stas00:
I confirm that it solved the problem! Thank you, Jeff!

Good to go to merge this into the big-science branch!

stas00:
BTW, the checkpoint still requires `mp_rank_*_model_states.pt` and the program crashes without those.

```
AssertionError
        sd_loader = SDLoaderFactory.get_sd_loader(ckpt_list)sd_loader = SDLoaderFactory.get_sd_loader(ckpt_list)
  File "/gpfsssd/worksf/projects/rech/six/ura81os/code/deepspeed-big-science/deepspeed/runtime/state_dict_factory.py", line 162, in check_ckpt_list
    assert len(self.ckpt_list) > 0
```

Do we really need those? Or is this another legacy check?

tjruwase (approved):

stas00:
So now this needs to be replayed to the `big-science` branch. Thank you!

PR #1324, opened 2021-08-25
Fix PP checkpoint bloat

tjruwase:
Use tensor cloning to avoid checkpoint bloat resulting from bad interaction of `torch.save()` and PP.

See [here](https://pytorch.org/docs/stable/notes/serialization.html#preserve-storage-sharing) for more details of the relevant `torch.save()` behavior. 

@stas00, FYI


ShadenSmith (approved):

jeffra (approved):
This is great, thanks Tunji!

PR #1328, opened 2021-08-26
Added 4-byte alignment on NCCL/RCCL

amathews-amd:
Fixes https://ontrack.amd.com/browse/MSRCHA-137

- Adds 4-byte alignment on NCCL/RCCL workloads to speed up workloads. 
  - The start location of all data partitions (across worldsize) is aligned to 4-byte boundary.


deepspeed/runtime/zero/stage2.py
```diff
@@ -230,6 +230,11 @@ def __init__(self,
         # number of elements per partition in each group
         self.partition_size = []
 
+        #align nccl all-gather send buffers to 4-bye boundary
+        self.nccl_start_alignment_factor = 2  # 4-byte alignment/sizeof(fp16) = 2
+
+        assert (allgather_bucket_size % self.nccl_start_alignment_factor == 0), "allgather_bucket_size must be a multiple of nccl_start_alignment_factor"
```
tjruwase, line 236: It would be useful for this assert message to include the `nccl_start_alignment_factor` value so that users can adjust the `allgather_bucket_size` appropriately

jeffra:
> > @amathews-amd, can you add unit tests for this?
> 
> Let me see what I can do.
> 
> Can you point me to instructions on how to setup a development environment and environment to test (num-gpus, installed package dependencies) ? I am currently using https://github.com/microsoft/DeepSpeed/blob/master/docker/Dockerfile and running pytest, but there seems to be other dependencies.

This should include all the packages needed to run our tests, we currently assume the machine that is running the tests has 4 gpus available.

https://github.com/microsoft/DeepSpeed/blob/master/requirements/requirements-dev.txt

deepspeed/runtime/zero/stage2.py
```diff
@@ -1376,22 +1383,25 @@ def buffered_reduce_fallback(self,
...
+        base_size = ( total_num_elements // divisor ) \
+            + ( total_num_elements % divisor > 0)
+        base_size *= alignment_factor
 
         start = 0
         for id in range(dp):
             partition_size = base_size
-            if id < remaining:
-                partition_size = partition_size + 1
+            if id == (dp - 1):
```
tjruwase, line 1403: It seems to me that the new code is not functionality equivalent even if alignment_factor = 1 
Consider tne=6, dp=4, where tne is total_num_elements
In the old code, we have 
tne/dp = 1
tne%dp = 2
partition_size = [2, 2, 1, 1]

In the new code, we have partition_size = [2,2,2,0]

Am I correct?

amathews-amd: Yes. That would be correct. 

The tne is a big number usually. 
The particular test we were looking for in MSRCHA-137 (run_glue) had tne=1,566,913,544. Original base_size was 195864193, across 8-gpus. This particular change would make base_size = 195864194. 
The partitions then are [195864194, 195864194, ... , (195864194-8)]. The last partition can go up to (base_size-15), where 15 = (alignment_factor*num_gpus-1).

tjruwase: Thanks for the response and context. We need this to work correctly even for small sizes of tne, and I suspect you will run into such cases in our unit tests. I think it is fine to pad total_num_elements for cases where total_num_elements < alignment_factor * dp, and your change [here](https://github.com/amathews-amd/DeepSpeed/blob/a0d99dbdcefc7fd8d41435a21f7d3ac084b038ad/deepspeed/runtime/zero/stage2.py#L291-L292) seems to do that already.  However, rather than align to (alignment_factor * dp_size) which could lead to excessive padding, why not align to LCM(alignment_factor, dp_size)?

amathews-amd: LCM(alignment_factor, dp_size) would not work. We split the partitions based on dp_size already in both the original code and new case. After this division, we need the first element of each partition to have a 4-byte position. For FP16, this means the address of starting element of each partition has to be multiple of 2.

tjruwase: Yes, you are correct. I was wrong. Thanks for the explanation.

tjruwase:
@amathews-amd, can you please share some perf improvement numbers that you see with these changes? Thanks!

amathews-amd:
@tjruwase , on MI100 gpu's, the improvement for the dataset that trips this issue was 2.05x in training samples per second. The core developers of RCCL believe that this will improve NCCL performance also. When I tested on V100, I saw that this have not improved performance improvement significantly (1.02x), but that might be due to hardware that I tested on.  

I believe @suffiank has also verified improvement internally to MSFT.

tjruwase:
> @tjruwase , on MI100 gpu's, the improvement for the dataset that trips this issue was 2.05x in training samples per second. The

That is awesome. Thanks. 
Regarding this PR, it seems only the unit tests are outstanding. The changes LGTM.

amathews-amd:
> > @tjruwase , on MI100 gpu's, the improvement for the dataset that trips this issue was 2.05x in training samples per second. The
> 
> That is awesome. Thanks.
> Regarding this PR, it seems only the unit tests are outstanding. The changes LGTM.

Done. Please let me know if the tests are not in the right place.

tjruwase:
@amathews-amd, this looks great. Waiting for CI to complete and this can be merged.

deepspeed/runtime/zero/stage2.py
```diff
@@ -300,7 +306,8 @@ def __init__(self,
             # each process will compute on a different part of the partition
             data_parallel_partitions = self.get_data_parallel_partitions(
                 self.fp16_groups_flat[i],
-                i)
+                i,
+                self.nccl_start_alignment_factor)
```
samyam, line 310: @amathews-amd since you have already aligned self.fp16_groups_flat, do we need this alignment on 310 again? We should be able to leave the get_data_parallel_parititons method unmodified  no?

samyam, line 310: @amathews-amd since you have already aligned self.fp16_groups_flat, do we need this alignment on 310 again? We should be able to leave the get_data_parallel_parititons method unmodified  no?

deepspeed/runtime/zero/stage2.py
```diff
@@ -318,8 +325,8 @@ def __init__(self,
                 i].requires_grad = True  # keep this in case internal optimizer uses it
             param_group['params'] = [self.single_partition_of_fp32_groups[i]]
 
-            partition_size = len(self.fp16_groups_flat[i]) / dist.get_world_size(
-                group=self.real_dp_process_group[i])
+            #partition_size is same across all partitions
+            partition_size = data_parallel_partitions[0].numel()
```
samyam, line 329: Can we leave this part as it was before as both are doing the same thing here? I do not see the benefit of the change but I think we should add an assertion partition_size % alignment == 0 here.

deepspeed/runtime/zero/stage2.py
```diff
@@ -1376,22 +1383,25 @@ def buffered_reduce_fallback(self,
 
     # views the tensor as multiple partitions and returns
     # those partitions
-    def get_data_parallel_partitions(self, tensor, group_id):
+    def get_data_parallel_partitions(self, tensor, group_id, alignment_factor):
```
samyam, line 1386: I think this function does not need to be modified.

deepspeed/runtime/zero/stage2.py
```diff
@@ -1993,8 +2003,12 @@ def _restore_from_fp32_weights(self, all_state_dict):
             ]
             flat_merged_partitions = self.flatten_dense_tensors_aligned(
                 merged_partitions,
+                self.nccl_start_alignment_factor *
                 dist.get_world_size(group=self.real_dp_process_group[i]))
-            dp_partitions = self.get_data_parallel_partitions(flat_merged_partitions, i)
+            dp_partitions = self.get_data_parallel_partitions(
+                flat_merged_partitions,
+                i,
+                self.nccl_start_alignment_factor)
```
samyam, line 2011: Again, since flat_merged_partitions is aligned, we not need the alignment here but we can add an assertion that dp_partitions are all aligned to nccl_start_aligned_factor

deepspeed/runtime/zero/stage2.py
amathews-amd: len(self.fp16_groups_flat[i]) / world_size != data_parallel_partitions[0].numel()
They are not doing the same thing; The new code uses ceil() and therefore, can be len(self.fp16_groups_flat[i]) / world_size **+ 1** in some cases.

deepspeed/runtime/zero/stage2.py
amathews-amd: flatten_dense_tensors_aligned() is not a good name (part of the original code). All it actually does is padding at the end.

deepspeed/runtime/zero/stage2.py
amathews-amd: Same as my comment above.

To be specific, the nccl alignment we are concerned about here is the alignment of the start location of each partition. See the newly added testcases.

deepspeed/runtime/zero/stage2.py
amathews-amd: This function is what does the 'alignment'. 

To be specific, the nccl alignment we are concerned about here is the alignment of the start location of each partition. See the newly added testcases.

deepspeed/runtime/zero/stage2.py
samyam: The padding at the end is enough to keep things aligned

deepspeed/runtime/zero/stage2.py
samyam: I think in this case they will always be the same because  self.fp16_groups_flat[i].numel() % (world_size*nccl_alignment) == 0 due to the padding in flatten_dense_tensor_aligned.

deepspeed/runtime/zero/stage2.py
samyam: Yes I get that but isn't it redundant since the alignment is already done by flatten_dense_tensor_aligned?

deepspeed/runtime/zero/stage2.py
samyam: please see comment above

deepspeed/runtime/zero/stage2.py
samyam: Thanks. I agree its not fully explicit but its not completely implicit either since you are making the alignment when creating the buffer. I think we can also add explicit assertions once we get the partitions from get_data_parallel_parititons  to ensure directly in Stage 2 that the alignment is satisfied instead in addition to having the unit tests. Since the assertion is only in initialization, this will only be done once at the beginning.

amathews-amd: Done.

amathews-amd: I have made the modifications as recommended.

samyam (approved):

deepspeed/runtime/zero/stage2.py
samyam: Thank you. Looks great to me..

PR #1329, opened 2021-08-26
update for cuda-11.4

stas00:
cuda-11.4 is out.

might choose to not list explicit versions in the future. For now just add the 11.4 entry.

@tjruwase 

(and `big-science` please)


jeffra:
@stas, I cherry-picked to `big-science` as well

stas00:
Great! Thanks a lot, @jeffra!

PR #1330, opened 2021-08-27
Added both CPU support for lamb optimizer!!

harishsg99:
I was facing issue while building model using deepspeed on CPU , As lamb optimizer does not support CPU.So I decided to add cpu support for lamb optimizer.


rocm-mici:

mrwyattii:
This PR is incomplete and does not fully implement `DeepSpeedLambCPU`. For example, the GPU op builder is still used in this implementation: `self.fused_lamb_cpu = FusedLambBuilder().load()`

PR #1333, opened 2021-08-30
Remove duplicate clip grad function in deepspeed

awan-10:


awan-10:
@conglongli @samyam, please review this PR and approve so it can get merged. This was a post-release pending item. A similar fix already exists and works fine for the fused fp16 optimizer.

conglongli (approved):
LGTM

PR #1337, opened 2021-09-01
Support client lr schedulers that are not subclass of torch _LRScheduler

tjruwase:
Not all lr schedulers are subclass from torch _LRScheduler.


conglongli (approved):
Tested and works well. Thanks @tjruwase for the quick fix!

PR #1354, opened 2021-09-09
Remove dropout as client code can do it independently.

awan-10:
We might do double dropout if the client code does not force it so it is better to let the client do dropout instead of inside the moe layer.


conglongli (commented):
LGTM, @yaozhewei please also review and then approve

yaozhewei (approved):
Looks all good to me. Thanks a lot

PR #1368, opened 2021-09-15
Improve handling of `Init`'s config arguments

aphedges:
## Changes

- Remove remaining usage of `config` instead of `config_dict_or_path`
- Add note in docstring that `config` is deprecated
- Fix typo in docstring

## Explanation

I figured out that when using `transformers`, as of DeepSpeed commit e08c239, I was having issues loading models for training as fp32. I figured out that `transformers` is still using the `config` parameter, even though it does not function the way it did back before the commit. This causes `InsertPostInitMethodToModuleSubClasses.dtype` to default to `torch.half`, even though the class should take `ds_config.fp16_enabled` into account.

My fix here is to treat `config` as a deprecated alias for the new `config_dict_or_path` parameter. This seems reasonable given a comment on another PR: <https://github.com/microsoft/DeepSpeed/pull/1271#discussion_r683837140>. In addition, from #1008 and #1271, it seems that there only needs to be one instance of `DeepSpeedConfig` in the method, so I removed the second one.

I added a runtime warning so users can see that the parameter that they were using should be changed. I will be putting a PR out to transfer to the new parameter in `transformers` after this PR is merged and released. I'm guessing that the old `config` parameter should be removed at some point in the future, but I have no clue about the deprecation policy for this library.

### DIFF
--- a/deepspeed/runtime/zero/partition_parameters.py
+++ b/deepspeed/runtime/zero/partition_parameters.py
@@ -374,12 +374,12 @@ class Init(InsertPostInitMethodToModuleSubClasses):
                 ``"cpu"``. Defaults to ``False``.
             config_dict_or_path (dict or ``json file``, optional): If provided, provides configuration.
             config (``json file`` or dict, optional): If provided, provides configuration
-                for swapping fp16 params to NVMe.
+                for swapping fp16 params to NVMe. Deprecated, use ``config_dict_or_path`` instead.
             enabled (bool, optional): If ``False``, this context has no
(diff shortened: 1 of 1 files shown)


aphedges:
It seems that #1373 was released and merged after this, so one of the core problems this PR fixes no longer exists. However, it still fixes a related bug, so I think that it is still relevant. I've rebased to incorporate the changes from the merged PR.

aphedges:
The bug and the main documentation fix in this PR were solved when #1407 was merged. No need to include this now, then.

PR #1370, opened 2021-09-15
Introduce a device rank when setting device

SeanNaren:
When users pass specific device IDs to the Trainer in Lightning, this caused DeepSpeed to have a reference to the incorrect rank when setting device. This was due to the assumption that `local_rank` pointed to the correct CUDA device in the DeepSpeed engine.

Associated issue: https://github.com/PyTorchLightning/pytorch-lightning/issues/9521

This PR introduces a new variable in the args `device_rank`, which if present will be used to set the device (else default to local rank). I would like to add a test for this, but any assistance on where/what format would be appreciated :)

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -593,7 +593,9 @@ class DeepSpeedEngine(Module):
             return None
 
     def _set_distributed_vars(self, args):
-        device_rank = args.device_rank if args is not None and args.device_rank else self.local_rank
+        device_rank = args.device_rank if args is not None and hasattr(
+            args,
+            'device_rank') else self.local_rank
         if device_rank >= 0:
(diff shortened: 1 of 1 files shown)


jeffra:
Hi @SeanNaren, thanks for your PR :) 

A couple comments, 
1) I think we need a `hasattr` guard for the case when args does not have `device_rank`. I believe this will fix the current unit tests that are failing on our side.
2) I assume the intent is that we always defer to `args.device_rank` for the device placement if args.device_rank exists, correct? In that case I think there's a small edge case in the current code, if `args.device_rank == 0` then it will currently defer to `self.local_rank`.

SeanNaren:
Thanks @jeffra! Hopefully the change fixes the tests :)

jeffra (approved):

PR #1388, opened 2021-09-22
fix file-system race condition in saving/loading to disk on some test nodes

jeffra:
For some reason we were seeing corrupt checkpoints in our unit tests where we immediately save/restore checkpoints. These changes seem to fix the issue by bypassing the `torch.save` file descriptor and forcing a flush at the end of the same.

Also, this commit bumps DSE to latest commit


jeffra:
will wait for amd changes

PR #1391, opened 2021-09-23
fix dataloder len

### LINKED ISSUE #1390: [BUG] Dataloader length is wrong when batch size > 1
**Describe the bug**
The dataloader length is set to length of data sampler here: https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/dataloader.py#L71

This is wrong when batch size is > 1
Data loader actual length should be sampler length divided by batch size.

This caused me quite a confusion when debug a training problem when I try to print the number of steps trained.

**To Reproduce**
print the dataset length, data sampler length and data loader length of any training example

**Expected behavior**
Training a model on 2 nodes with 8 GPUs each.
Given 11788 training images, batch size per gpu is 3.
Distributed data sampler should shared the data on 16 gpus, len is 737
when batch size is 3, then data loader length is 246
```
[1,0]<stdout>:data loader len:  246
[1,0]<stdout>:sampler len:  737
[1,0]<stdout>:data set len:  11788
```
**ds_report output**

wrong output:
```
[1,0]<stdout>:data loader len:  737
[1,0]<stdout>:sampler len:  737
[1,0]<stdout>:data set len:  11788
```

I will try to raise a PR soon.
Thanks

roywei:
fix #1390


deepspeed/runtime/dataloader.py
```diff
@@ -68,9 +68,10 @@ def __init__(self,
         self.device_count = device_count
         self.batch_size = batch_size
         self.pin_memory = pin_memory
-        self.len = len(self.data_sampler)
-        self.data = None
         self.dataloader_drop_last = dataloader_drop_last
+        # create self.dataloader and self.data
+        self._create_dataloader()
```
tjruwase, line 73: Why call create_dataloader() here? Our current approach is to create in __iter__. If I remember correctly, this is needed to handle multiple epoch. Did you encounter a problem with the current approach?

tjruwase:
@roywei, thanks for this PR. Can you please add some unit tests? An appropriate location would be `tests/unit/test_data.py`

tjruwase:
@roywei, can you please address formatting issues using instructions [here](https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites)? Thanks.

roywei:
@tjruwase  Thanks for the review! will add test and reformat later today.

Microsoft-code (approved):

deepspeed/runtime/dataloader.py
```diff
@@ -68,7 +69,10 @@ def __init__(self,
         self.device_count = device_count
         self.batch_size = batch_size
         self.pin_memory = pin_memory
-        self.len = len(self.data_sampler)
+        if dataloader_drop_last:
+            self.len = len(self.data_sampler) // self.batch_size
```
chunyang-wen, line 73: Why `self.len` is divided by `batch_size`?

tjruwase:
@roywei, just wanted to check if you able to finish this PR? Thank!

roywei:
@tjruwase Sorry I totally forgot about this. I merged some commits from @junxu with test (Thanks to @junxu !!)
It should be good now.

tests/unit/test_data.py
```diff
@@ -56,3 +57,45 @@ def _test_dataloader_drop_last(args, model, hidden_dim):
...
 
     _test_dataloader_drop_last(args=args, model=model, hidden_dim=hidden_dim)
+
+
+@pytest.mark.parametrize('world_size, total_samples, batch_size, drop_last, result',
+                         [(16,
+                           11788,
+                           3,
+                           True,
+                           245),
```
tjruwase, line 67: I think it would be better to compute `result` value in the function based on `drop_last` rather than pass as an argument.

tests/unit/test_data.py
```diff
@@ -56,3 +57,45 @@ def _test_dataloader_drop_last(args, model, hidden_dim):
...
+                           246)])
+def test_dataloader_len(world_size, total_samples, batch_size, drop_last, result):
+    args = {
+        "world_size": world_size,
+        "total_samples": total_samples,
+        "batch_size": batch_size,
+        "drop_last": drop_last,
+        "result": result
+    }
+    #args = args_from_dict(tmpdir, config_dict)
```
tjruwase, line 81: Please remove this line

tests/unit/test_data.py
```diff
@@ -56,3 +57,45 @@ def _test_dataloader_drop_last(args, model, hidden_dim):
...
+                         [(16,
+                           11788,
+                           3,
+                           True,
+                           245),
+                          (16,
+                           11788,
+                           3,
+                           False,
+                           246)])
```
tjruwase, line 72: In general, values like `world_size`, `total_samples`, `batch_size`, which are not changing should be set in the function rather than passed as args.

jeffra:
Stale, feel free to re-open if this is still required.

PR #1396, opened 2021-09-24
[CI] Add HF transformers integration tests

jeffra:
Related to #937 and is a refresh of #958 from @awan-10


jeffra:
fixed issue in #958, closing this in favor of other

Issue #1347, opened 2021-09-06
[BUG]  Module names GPT2ModelPipe & ParallelTransformerLayerPipe is hardcoded in deepspeed


DanielHesslow:
In multiple places in the deepspeed code there exist hardcoded checks such as:
`if self.module.__class__.__name__ == 'GPT2ModelPipe':`
or 
`all('ParallelTransformerLayerPipe' in f.__class__.__name__ for f in funcs)`
In order to make the examples work.

This means that it is not possible to integrate deepspeed into an existing codebase by looking at the examples or documentation, one needs to thoroughly read the deepspeed code and then name the modules exactly as stated above to get the desired behavior.
To my knowledge this peculiarity is not documented anywhere.

This has been previously noted for checkpointing here: #688 

But it is also necessary in order to send the boolean attention masks across pipeline stages. If one attempts to send the attention masks as described in the tutorial but having the wrong module name, deepspeed will deadlock on the backward pass. I believe this is due to NACL not correctly handing boolean tensor but such workaround should happen in deepspeed, or in the very least there should be warnings. 

This should either be _clearly_ documented, especially in the examples, or better yet, fixed. 

Thanks

hyunwoongko:
I think this is a problem. We need to provide a more flexible API.

hyunwoongko:
https://github.com/microsoft/DeepSpeed/pull/1399 I opened PR.

hyunwoongko:
https://github.com/microsoft/DeepSpeed/pull/139 This PR is merged. could you check?

DanielHesslow:
Thanks, from looking at the code this appear to solve the issue. Though the conversion of boolean tensors is seemingly only done on the last tensor (as that is the attention mask in GPT2ModelPipe). With current implementation I think has_bool_tensor should probably be renamed to last_tensor_is_bool or smth. A perhaps better alternatives is to let the user provide a list of which tensors are boolean.

Unfortunately I don't have the time atm to properly test the fix, I'll close the isssue for now and and I'll reopen it later if there are any remaining problems.

And again, thanks for the quick fix :)

PR #1376, opened 2021-09-16
Fix non-fp tensor bugs of contiguous activation checkpointing

hyunwoongko:
refer https://github.com/microsoft/DeepSpeed/issues/1375

It works well.

PR #1399, opened 2021-09-25
Add flexibility of pipeline parallel module and engine

hyunwoongko:
## What for?
Add flexibility of pipeline parallel module and engine (https://github.com/microsoft/DeepSpeed/issues/1347)

## Request review
@ShadenSmith @jeffra @sdtblck


stas00:
Heh, @thomasw21 was just dealing with it - https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/107

You can see from that PR @ShadenSmith is looking at fixing bool tensor support and removing this hack altogether.

thomasw21:
> Heh, @thomasw21 was just dealing with it - [bigscience-workshop/Megatron-DeepSpeed#107](https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/107)
> 
> You can see from that PR @ShadenSmith is looking at fixing bool tensor support and removing this hack altogether.

Hey thanks @stas00 . Yeah I've been working to fix some issues I've been having with current pipeline parallelism implementation. I've just open a draft PR if you want to check it out. https://github.com/microsoft/DeepSpeed/pull/1400

tjruwase:
@hyunwoongko, thanks for your hard work on this PR. I will take a look today so we can wrap it up soon. Thanks for your patience.

deepspeed/runtime/activation_checkpointing/checkpointing.py
```diff
@@ -365,11 +365,14 @@ def partition_activations(args, cpu_checkpoint, contiguous_checkpoint):
...
 
     inputs = []
+    num_non_fp_tensors = 0
     for i, item in enumerate(args):
         if not is_activation_to_checkpoint(item):
             inputs.append(item)
+            num_non_fp_tensors += 1
             continue
 
+        i -= num_non_fp_tensors
```
tjruwase, line 375: Can we avoid modifying the index variable `i`? This pattern makes the code brittle. Will the following pattern work?

```python
i = 0
for item in args:
  # loop body
  i+1
```

deepspeed/runtime/pipe/engine.py
```diff
@@ -49,14 +49,15 @@ class PipelineEngine(DeepSpeedEngine):
     This engine is created by ``deepspeed.initialize()`` when a :class:`PipelineModule`
     is provided.
     """
-    def __init__(self, *super_args, **super_kwargs):
+    def __init__(self, has_bool_tensors=False, *super_args, **super_kwargs):
```
tjruwase, line 52: I would prefer for this PP changes to be a separate PR, as it seems fairly independent of the activation checkpointing changes. Also, I am more comfortable reviewing the activation checkpointing changes but will need other team members to review the PP changes. Thanks.

tjruwase:
@hyunwoongko, I left some comments. Thanks.

deepspeed/runtime/activation_checkpointing/checkpointing.py
hyunwoongko: At the bottom, `i` is used, and that's where the problem arises.

hyunwoongko: I know it's a bad idea to change `i`, but it's the best way to maintain the overall code structure.

hyunwoongko: `i` is used very often. Are you sure you want to change all these `i`?

```python
if contiguous_checkpoint:
            if i >= len(contiguous_data_buffers):
                tensor_list = [
                    torch.tensor(()).new_empty([partition_size],
                                               dtype=partition.dtype,
                                               device=buffer_device)
                    for i in range(num_layers)
                ]
                contiguous_data_buffers.append(tensor_list)
                data_offsets.append(0)
            elif contiguous_data_buffers[i] is None:
                tensor_list = [
                    torch.tensor(()).new_empty([partition_size],
                                               dtype=partition.dtype,
                                               device=buffer_device)
                    for i in range(num_layers)
                ]
                contiguous_data_buffers[i] = tensor_list
                data_offsets[i] = 0

            # Because the 'new_empty' returns uninitialized pages,
            # the pages need to be populated during the cudaMemcpy time
            # which increases the data copy time. To avoid this, we
            # pre-populate these pages by simply writing 0 ahead of
            # the actual cudaMemcpy operation time. Due to the
            # previously launched GPU kernels, there is a small
            # window of time here for CPUs to populate pages asynchronously.
            contiguous_data_buffers[i][data_offsets[i]].data[range(
                0,
                contiguous_data_buffers[i][data_offsets[i]].data.shape[0],
                int(mmap.PAGESIZE /
                    contiguous_data_buffers[i][data_offsets[i]].data.element_size())
            )] = 0

            contiguous_partition = contiguous_data_buffers[i][
                data_offsets[i]].data.copy_(partition.data)
            data_offsets[i] = data_offsets[i] + 1
            inputs.append(contiguous_partition)
        else:
            partition = partition.cpu() if CPU_CHECKPOINT else partition
            inputs.append(partition)
```

hyunwoongko: Also, the `i` used in the outer loop is used again in the inner loop. It seems good to change the `i` of the inner loop to `j`.

```python
    for i, item in enumerate(args):
        if not is_activation_to_checkpoint(item):
            inputs.append(item)
            num_non_fp_tensors += 1
            continue

        i -= num_non_fp_tensors
        partition_size = get_partition_size(item)
        partition = item.detach().contiguous().view(-1).narrow(
            0,
            get_partition_start(item),
            partition_size).clone()

        buffer_device = torch.device('cpu') if cpu_checkpoint else partition.device

        if contiguous_checkpoint:
            if i >= len(contiguous_data_buffers):
                tensor_list = [
                    torch.tensor(()).new_empty([partition_size],
                                               dtype=partition.dtype,
                                               device=buffer_device)
                    for i in range(num_layers)
                ]
                contiguous_data_buffers.append(tensor_list)
                data_offsets.append(0)
```

deepspeed/runtime/pipe/engine.py
hyunwoongko: good idea. Should I close this PR and open a new one?

hyunwoongko: I re-opend https://github.com/microsoft/DeepSpeed/pull/1376 . 
I will remove contiguous checkpointing parts in this PR. 
Let's discuss above PR.

hyunwoongko:
I have separated the two PRs separately.

deepspeed/runtime/activation_checkpointing/checkpointing.py
tjruwase: I agree that `i` is used often and it would be too painful to change it. This is why I proposed initializing `i` outside the loop and incrementing appropriately in the loop. That way the rest of the code does not have to change. Does that make sense?

hyunwoongko: yes i think it's good By the way, as you suggested, I split the PR, so I think it would be good to discuss it #1376.

deepspeed/runtime/pipe/engine.py
```diff
@@ -854,7 +855,7 @@ def _exec_send_activations(self, buffer_id):
         # NCCL does not like to send torch.BoolTensor types, so cast the mask to half().
         # We could do char, but with half() we can eventually flatten with other fp16
         # messages (TODO)
-        if self.module.__class__.__name__ == 'GPT2ModelPipe':
+        if self.module.__class__.__name__ == 'GPT2ModelPipe' or self.has_bool_tensors:
```
tjruwase, line 858: Do we need to keep testing against GPT2ModelPipe here? Wont self.has_bool_tensors also be true? Or do we need to modify the GPT2ModelPipe codes?

hyunwoongko: I considered backward compatibility.

hyunwoongko: > do we need to modify the GPT2ModelPipe codes?

Yes I want to use the pipeline engine using another class. Supporting only one class name is inflexible.

hyunwoongko: The reason I kept `GPT2ModelPipe` as is is for backwards compatibility.

tjruwase: That makes sense. Thanks.

tjruwase (approved):

PR #1400, opened 2021-09-26
WIP: Big science fix passing multiple tensors

thomasw21:
Fixes necessary for implementation of prefix lm:
 - Passing tuple of tensors between stages
 - Allowing models to pass bool tensors. (This seems fixed in `torch.distributed` for pytorch > 1.7)

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -850,7 +850,7 @@ class PipelineEngine(DeepSpeedEngine):
             for idx in range(num_tensors):
                 recv_dtype = torch.LongTensor(data=[0]).to(self.device)
                 p2p.recv(recv_dtype, send_stage)
-                recv_dtype = recv_dtype.item()
+                recv_dtype = self.ID_TO_DTYPE[recv_dtype.item()]
                 recv_ndims = torch.LongTensor(data=[0]).to(self.device)
                 p2p.recv(recv_ndims, send_stage)
                 recv_ndims = recv_ndims.item()
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -790,7 +790,7 @@ class PipelineEngine(DeepSpeedEngine):
                 assert isinstance(tensor, torch.Tensor)
                 send_shape = torch.LongTensor(data=tensor.size()).to(self.device)
                 send_ndims = torch.LongTensor(data=[len(tensor.size())]).to(self.device)
-                send_dtype = torch.LongTensor(data=self.DTYPE_TO_ID[tensor.dtype]).to(self.device)
+                send_dtype = torch.LongTensor(data=[self.DTYPE_TO_ID[tensor.dtype]]).to(self.device)
                 p2p.send(send_dtype, recv_stage)
                 p2p.send(send_ndims, recv_stage)
                 p2p.send(send_shape, recv_stage)
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -63,7 +63,7 @@ class PipelineEngine(DeepSpeedEngine):
         torch.int64,
         torch.bool
     ]
-    DTYPE_TO_ID = {id_: dtype for id_, dtype in enumerate(ID_TO_DTYPE)}
+    DTYPE_TO_ID = {dtype: id_ for id_, dtype in enumerate(ID_TO_DTYPE)}
 
     def __init__(self, *super_args, **super_kwargs):
         super().__init__(*super_args, **super_kwargs)
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -896,7 +896,7 @@ class PipelineEngine(DeepSpeedEngine):
             # Inject the partitoned tensor into the output before sending
 
             # XXX Hack
-            inputs = tuple([part.to_meta(), part.data(), *inputs[1:]])
+            inputs = ([part.to_meta(), part.data(), *[elt.grad for elt in inputs[1:]]])
 
         # XXX Terrible hack
         # Drop the attention mask from the input buffer here. It does not have
@@ -990,7 +990,7 @@ class PipelineEngine(DeepSpeedEngine):
                 local_part=outputs[1],
                 group=self.grid.get_slice_parallel_group())
             outputs[0].data = part_output.full()
-            outputs = tuple([outputs[0], *outputs[2:]])
+            outputs = ([outputs[0], *outputs[2:]])
             # save for backward
             self.pipe_buffers['outputs'][buffer_id] = outputs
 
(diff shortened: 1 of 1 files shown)


hyunwoongko:
@thomasw21 Nice work. BTW what does `Passing tuple of tensors between stages` mean? It seems like it's already possible to send a tuple. Could you explain in detail?

thomasw21:
Hey @hyunwoongko ! Thanks for taking an interest ! My bad I had added a quick description in order to open this draft PR to share my views, and it might be inaccurate. According to https://github.com/microsoft/DeepSpeed/pull/1400/files#diff-26611f6be759237464a03bb1328cbc16555888836b3504dc3703e2e25d2a3ca3L892 passing tuple was not possible. I'll admit I have little experience in DeepSpeed and so I'm unclear if the fix is correct. 

Also I seem to be hitting a deadlock somewhere, so this PR is not ready. If you feel you have an idea of what is going on, feel free to share :D

stas00:
FYI, `big-science` has been finally merged into `master`, so we should do all fixes in `master` and switch BigScience's Meg-DS to use deepspeed@master for all the new trainings.

Going forward we plan to do all fixes in `deepspeed@master`, and then we can freeze a specific sha for production use on our side.

Therefore if it's not too much trouble could you please re-create this PR on deepspeed@master? Thank you, Thomas.

deepspeed/runtime/pipe/engine.py
```diff
@@ -588,14 +598,23 @@ def _exec_forward_pass(self, buffer_id):
...
+            if isinstance(outputs, tuple):
+                first_output = outputs[0]
+                # TODO: Improve pipe partitioning to pass multiple tensors that require grads
+                assert all([torch.is_tensor(elt) and elt.requires_grad is False for elt in outputs[1:]])
+                outputs_tail = outputs[1:]
+            elif torch.is_tensor(outputs):
+                first_output = outputs
+                outputs_tail = []
+            else:
+                raise NotImplementedError
```
stas00, line 610:
```suggestion
                raise ValueError("expecting a tensor or a tuple of tensors")
```

Let's be user-friendly

deepspeed/runtime/pipe/engine.py
```diff
@@ -889,10 +913,20 @@ def _exec_send_grads(self, buffer_id):
...
-            part = PartitionedTensor(tensor=inputs.grad,
+            if isinstance(inputs, tuple):
+                first_input = inputs[0]
+                assert all([torch.is_tensor(elt) for elt in inputs[1:]])
+                inputs_grad_tail = [elt.grad for elt in inputs[1:] if elt.grad is not None]
+            elif torch.is_tensor(inputs):
+                first_input = inputs
+                inputs_grad_tail = []
+            else:
+                raise NotImplementedError
```
stas00, line 924: same as above

deepspeed/runtime/pipe/engine.py
```diff
@@ -889,10 +913,20 @@ def _exec_send_grads(self, buffer_id):
...
+            assert torch.is_tensor(first_input)
+            part = PartitionedTensor(tensor=first_input.grad,
                                      group=self.grid.get_slice_parallel_group())
-            inputs = tuple([part.to_meta(), part.data()])
+
+            inputs = ([part.to_meta(), part.data(), *inputs_grad_tail])
```
stas00, line 929: should this not be a tuple?

stas00 (approved):
Great work, Thomas!

I left a few small suggestions.

I think the only issue here is that this new code doesn't have a test, so unless we add one it's likely to break down the road.

I'm not an expert on DS test suite, I added a few tests following monkey-see-monkey-do but this one might be not so easy to add. Let's first see if this is in general acceptable by the DS team and then we can work on a test.

stas00:
@ShadenSmith, @jeffra - could you please have a look at the proposed changes - this is currently a blocker for our adding of prefix-lm to Meg-DS. Thank you!

hyunwoongko:
스크린샷 2021-10-06 오후 2 03 06

@thomasw21 I reproduced this issue. I was talking about being able to transfer in the middle stage. But, as you said, it was impossible to send a tuple in the last stage. Is this problem solvable?

hyunwoongko:
@thomasw21 I'll test your branch now. If it works, this will be a very great work.

hyunwoongko:
I found this line has error.

```
assert all([torch.is_tensor(elt) and elt.requires_grad is False for elt in outputs[1:]])
```

Why output tensor of last stage must not be grad tensor? 
Wouldn't that make training impossible?

thomasw21:
Hey @hyunwoongko! Concerning the grad issue, i had two reasons:
 - it doesn't fit our current need
 - I'm not super sure on how we'd want to handle that, typically with `PartitionedTensor`

hyunwoongko:
https://github.com/microsoft/DeepSpeed/issues/1432

I made un issue yo make it more accessible for others.
I think someone may have similar problems later on as we do.

deepspeed/runtime/pipe/engine.py
thomasw21: This is a tuple, in python you don't need to specifcy `tuple`, parenthesis work just fine.

thomasw21: ah well played I didn't notice, which now begs the question why it works ...

stas00:
> @stas00 Thanks ! I'll wait to get first general approval of the approach before writing tests I guess?
> 
> Something important to note is that passing bool tensors works because we use torch 1.7+ (there was an issue with bool tensors not working in distributed setting using nccl)
> 
>     * Discussion: [Fix deepspeed prefix-lm bigscience-workshop/Megatron-DeepSpeed#107 (comment)](https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/107#discussion_r714321658)
> 
>     * Related issue (fixed): [[distributed] NCCL Backend doesn't support torch.bool data type pytorch/pytorch#24137](https://github.com/pytorch/pytorch/issues/24137)

perhaps then it should check that `pytorch>=1.7` - I don't know if we could simply require that for the pipeline files?

@jeffra, what do you think? Do you have internal needs to support torch<1.7 for PP?

thomasw21:
I should have fixed all the failing tests:

`pytest --forked unit/test_checkpointing.py::test_checkpoint_moe unit/test_checkpointing.py::test_checkpoint_zero_no_optimizer unit/test_checkpointing.py::test_checkpoint_unique_tag unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_mp2_pp_2to1 unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_mp1_pp_2to1 unit/test_configurable_parallel.py::TestConfigurablePP::test_pp_basic unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_pp_1to2_mp_1to2 unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_mp2_pp_1to2 unit/test_configurable_parallel.py::TestConfigurablePP::test_gpt2_pp_2to1_mp_2to1 unit/test_configurable_parallel.py::TestConfigurableMP::test_gpt2_mp2_no_resize`

image

stas00:
@jeffra, @tjruwase - could you please start CI? Thank you!

PR #1421, opened 2021-09-30
Improve inference documentation

aphedges:
There was a problem where one of the examples was not working when running DeepSpeed with only a single GPU, so I fixed the code. While I was at it, I made some other small fixes to the documentation file.


jeffra (approved):
Thank you! :)

Issue #1317, opened 2021-08-19
zero_to_fp32.py still imports a wrong model after fix


xycforgithub:
Same as #1165; the issue is closed now so I'm opening a new one. 
It looks like the zero_to_fp32.py still cannot load the correct weights under multi-gpu and ZERO-2 setting.

Loading 'mp_rank_00_model_states.pt':

states['module']['deberta.encoder.layer.0.output.dense.weight']
tensor([[-0.0211, 0.0068, 0.0206, ..., 0.0057, 0.0316, 0.0256],
[ 0.0273, 0.0141, 0.0118, ..., -0.0122, 0.0054, 0.0010],
[ 0.0479, -0.0237, -0.0604, ..., -0.0340, -0.0183, 0.0691],
...,
[ 0.0270, -0.0231, 0.0218, ..., 0.0563, 0.0641, -0.0094],
[-0.0563, -0.0837, -0.0427, ..., 0.0242, -0.0132, -0.0512],
[-0.0012, 0.0064, 0.0465, ..., 0.0219, 0.0259, -0.0281]],
device='cuda:0', dtype=torch.float16)

Loading the exported weights using `load_state_dict_from_zero_checkpoint`:
(Pdb) self.deberta.encoder.layer[0].output.dense.weight
Parameter containing:
tensor([[ 0.0207, -0.0448, 0.0022, ..., 0.0406, -0.0338, -0.0174],
[-0.0577, -0.0648, 0.0404, ..., 0.0108, -0.0167, -0.0100],
[ 0.0548, 0.0063, 0.0024, ..., 0.0311, 0.0249, 0.0167],
...,
[-0.0081, 0.0194, -0.0266, ..., -0.0269, -0.0002, 0.0257],
[ 0.0202, -0.0002, 0.0831, ..., -0.0008, -0.0094, 0.0258],
[-0.0320, 0.0529, -0.0259, ..., 0.0117, -0.0292, -0.0064]],

I'm using Deepspeed 0.4.5. I can confirm the problem happens on 2gpus, and not on 1gpu.

stas00:
@tjruwase, any chance you guys can take over the maintenance of zero_to_fp32, as I can't keep up with the changes in the core. It's the best that any changes you do get synced with the weights re-consolidation code in tandem.

I created the initial tests here:
https://github.com/microsoft/DeepSpeed/blob/10b48405ab0e69a1caf455f445526c4b39b0dbb8/tests/unit/test_zero.py#L135

[...]

PR #1422, opened 2021-10-01
[zero_to_fp32] fix to handle world_size

stas00:
This PR is trying to address specifically https://github.com/microsoft/DeepSpeed/issues/1317#issuecomment-926998777
as my original fix used a hardcoded alignment to 4, but the correct one is `2*world_size`

While at it I simplified `flatten_dense_tensors_aligned` a bit. hope it's ok.

-------------------------

Trying to unravel the padding in zero2 w/o the optimizer object looks impossible.

e.g. if I use a tiny model from the test with 57 params, and 2 gpus, 

1. zero2 pads with 3 ending up with 60 params and then splits it into 2x30 so this aligns to 2*world_size
2. but then before saving it, it looks up a different type of padding `self.groups_padding` which is totally different from nccl padding, so when it saves the `single_partition_of_fp32_groups` partitions it actually saves these 2 sizes `[30, 29]`

So when trying to reconstruct I end up with wanting 57 params and having 59 - so that gap is impossible to work out.

I suspect there is a bug somewhere there, as I think it should save `[30, 27]` instead, but it sort of works when DS reconstructs it using the complex logic.

The workaround I use is to basically align both numbers to `2*world_size` and only then compare that they are the same.

Please correct me if I'm wrong but I didn't see that the saved `single_partition_of_fp32_groups` actually ever include any padding. If they do that would explain why other users sometimes have an issue using this script with zero2. Because the padding would completely mess up the weights. I tried hard to write a test that would break, but I couldn't, I always get the exact same model back.

in Z2 I simply concatenate all `single_partition_of_fp32_groups` and then re-shape into params. (which apparently is only correct for single param group and will be fixed shortly).

@tjruwase, am I missing something here? The partitioning plus double padding code is so complex.

### DIFF
--- a/deepspeed/runtime/zero/stage2.py
+++ b/deepspeed/runtime/zero/stage2.py
@@ -770,10 +770,7 @@ class FP16_DeepSpeedZeroOptimizer(object):
 
     # create a flat tensor aligned at the alignment boundary
     def flatten_dense_tensors_aligned(self, tensor_list, alignment):
-        num_elements = 0
-        for tensor in tensor_list:
-            num_elements = num_elements + tensor.numel()
-
+        num_elements = sum(t.numel() for t in tensor_list)
         remaining = num_elements % alignment
 
         if remaining:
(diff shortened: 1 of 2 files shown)


tjruwase:
@stas00, thanks for doing the grueling work. No, you are not missing anything. The two padding logics evolved independently and we were not careful to avoid their conflicts. I will look into simplifying the padding logic by for starters padding exactly once, rather than multiple times for different constraints.

tjruwase (approved):

PR #1423, opened 2021-10-01
Fix many typos

aphedges:
After I came across some typos in the documentation and code while debugging, I decided to just run PyCharm's spellcheck on the entire repository. I was generally conservative, only making changes if I had a high confidence it was wrong. I probably missed a lot of typos, but this is still an improvement.

I tried to split the changes up into several commits for easier reviewing, but it's still a large diff. Let me know if I should make it more granular or if I should split it up into multiple commits.

There are also a couple of changes that modify identifiers like variable and function names. If you think they can cause compatibility issues, I will remove them. It's just hard to tell what code is part of the official API and what is not.

### DIFF
--- a/csrc/aio/py_lib/deepspeed_py_aio_handle.cpp
+++ b/csrc/aio/py_lib/deepspeed_py_aio_handle.cpp
@@ -196,7 +196,7 @@ bool deepspeed_aio_handle_t::_is_valid_parallel_aio_op(const bool read_op,
 {
     const auto op_string = read_op ? "Read" : "Write";
     if (num_bytes % get_thread_count()) {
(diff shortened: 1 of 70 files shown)


tjruwase:
@aphedges, Oh wow!!! This is amazing. Thanks so much for taking the trouble to make this incredibly valuable contribution.

tjruwase (approved):

hyunwoongko:
awesome.

aphedges:
@tjruwase, you're welcome! It surprisingly only took around 2 hours.

I don't think the failing test is my fault from reading the error message, but I'm not entirely sure.

PR #1428, opened 2021-10-05
Add basic MoE timing breakdown

awan-10:
This PR will add some basic breakdown of time spent in moe forward, alltoall, and gating functions. 

The output looks like this:

image


Microsoft-code (approved):

awan-10:
> Lines 184, 185 need same changes. thx

Thanks for catching the second instance. Pushed.

PR #1434, opened 2021-10-06
Fix typo (ValueError)

rpengms:
ValueError: not enough values to unpack (expected 3, got 2)


tjruwase:
@rpengms, thanks for this PR. It is quite strange that we have not run into this issue previously. Can you please share a stack trace for the failure that is PR fixes?

PR #1443, opened 2021-10-09
fix cl for pp/big science support

conglongli:
Need to also check and reset buffers in eval_batch.


stas00:
I guess this is just now waiting for one Approval - probably one of @jeffra, @tjruwase, @ShadenSmith on Monday.

PR #1446, opened 2021-10-11
Fix sparse attention imports

olegklimov:
Trivial, I don't know how it works for other people. But this fixes it for me (tested again just now).


jeffra:
Thank you for your contribution :)

PR #1398, opened 2021-09-25
Bfloat16 zero2

raamjad:
Enabling support for Bfloat16 with ZeRO-2 optimizer, partially resolving the issue#974
https://github.com/microsoft/DeepSpeed/issues/974

Bfloat16 collective communication requires NCCL 2.10.3.

PR #1453, opened 2021-10-12
Various ZeRO Stage3 Optimizations + Improvements (including bfloat16 support)

jfc4050:
the addition of bfloat16 support depends on another change that is in PR (https://github.com/microsoft/DeepSpeed/pull/1398). for now including the commit in here. will rebase once that PR goes through

# Results

**results on 8 p4d cluster (64 A100 GPUs)**
Roberta-10B:  54.86 -> 92.07 TFLOPS/GPU
Roberta-50B: 67.61 -> 116.06 TFLOPS/GPU

**results on 32 p4d cluster (256 A100 GPUs)**
Roberta-50B: ??? (TODO. due to hardware availability) -> 108 TFLOPS/GPU

note that these TFLOPS/GPU calculations ignore gradient checkpointing (which we do layer-wise), to account for it you can multiply by 4/3. These numbers are roughly at parity with ZeRO stage 2, but we unfortunately haven't quite gotten to linear weak scaling on our (worse than yours) hardware.

some notes about the conditions used to get the above numbers:
* p4d nodes w/ 8 A100 GPUs each and 400Gbps inter-node bandwidth
* batch size/GPU of 16
* sequence length 512
* 52B parameter Roberta model directly from huggingface, not using any custom fused kernels. vocab size ~53k
* activations are checkpointed layer-wise
* using bfloat16, we've observed similar throughput as fp16
* ZeRO Stage3, not using any CPU/disk offload, and contiguous gradients enabled.

# Testing
* 3 day training run using a BERT based model to verify equivalent loss curve as an unpartitioned model
* added several unit tests to verify activations/gradients

# Changelist
**Optimizations:**
- batching allgather calls to amortize fixed overhead and improve
  bandwidth utilization (credit: @zarzen)
- batching reduce_scatter calls to amortize fixed overhead and
  improve bandwidth utilization
- using *_base variants of allgather and reduce scatter to reduce memory
  allocations and data movement (credit: @zarzen)
- more fine grained synchronization for communication that allows
  blocking on less work
- asynchronous parameter deallocations (remains correct due to record_stream calls)
- precomputation of fetching code/caching of fetching+partitioning decisions - using a fetch queue rather than deciding what to (pre)fetch at each iteration
- prefetch hook fix (credit: @tjruwase)
- more aggressive initialization-time defragmentation of parameter partitions
- optimization of grad-norm computation - moving some math operations out of for loops and delaying synchronization
- limiting queued collectives to reduce memory pressure on pytorch cuda caching allocator (not elegant solution, real solution is perhaps best handled by pytorch)
- removing debug logs that were including the result of tensor operations, these were causing host-device transfers and synchronization
- [for optimizer CPU offload] made some host-device tensor copies async

**New features:**
- bfloat16 support for ZeRO 3 (based off changes in https://github.com/microsoft/DeepSpeed/pull/1398)
- there were some features missing for reduce_scatter that were available for allreduce, added them in so they are both at parity

**QOL improvements:**
- speed up model initialization by moving weights to GPU before weight
  initialization. Before was taking ~20 minutes for large models, now takes a minute or two.
- relative unit test imports so that unit tests can be run from any directory
- change performance logging to include memory consumption
- add logging w/ model size when done partitioning model
- added tracking for pytorch CUDA caching allocator flushes (bad for performance) and a warning log for when this is happening
- added NVTX ranges to high impact operations for easier profiling

**Bug fixes:**
- fix init context method when parent modules modify child weights. In cases where this was happening the parent module's modifications would not take effect on child module's weights since they were already partitioned, causing discrepancy in results when the context manager was used
- noticed that on master with our model we were getting different loss curves than unpartitioned model (was getting several unexpected gradient overflows), verified that this has been fixed
- parameter prefetching fix (stolen from @tjruwase)


tjruwase:
@jfc4050, Wow!!! This is a massive accomplishment. Thanks doing this. 

Can you please address the the formatting checks failure by following these [instructions](https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites)?

tjruwase:
@samyam, @stas00, and @jeffra FYI

tjruwase:
@jfc4050, is it possible to separate the bfloat support from this PR? To my understanding, these are quite independent changes, right?

stas00:
Amazing!

jfc4050:
> > > @jfc4050, is it possible to separate the bfloat support from this PR? To my understanding, these are quite independent changes, right?
> > 
> > 
> > sure will be some effort to detangle from the partitioning initializer fixes but i think we can do that. Note that once this PR #1398 goes in and we rebase, the bfloat16 specific changes in this PR will be quite small
> > seems like all the test failures are due to _all_gather_base and torch.cuda.nvtx.range not being present in the older pytorch version used in the tests. can we consider a pytorch version bump or should we go back and make them backwards compatible? there will be some performance penalty associated with using all_gather over _all_gather_base if users decide to use an older pytorch version that doesnt support it
> 
> Got it. Is it possible to support both here? Even if we bumped version on our CI, it would still break backwards-compatibility for existing users. We could document the perf benefits of the a version. It might also be worth further cluttering the output by issuing a (one-time) warning for this. What do you think?

yep that sounds reasonable. will make that change soon

tjruwase:
@jfc4050, We need another guard for torch version below:
image

jfc4050:
i think what @zarzen said sounds reasonable

tjruwase:
@zarzen and @jfc4050, thanks for the feedback. #1462 is now merged. #1188 and #1398 are both updated, and I resolved the conflict with #1453. I will suspend activity on #1170 for now and focus on merging #1188. @zarzen can you please rerun your correctness tests and we can both focus on any remaining CI issues (e.g., with HF unit tests). Does this sound reasonable?

tjruwase:
@jfc4050, @zarzen, @raamjad, #1188 merged over the weekend, but created conflicts for #1453. I expect #1398 to also merge today or tomorrow, which could create more conflicts for #1453. So, @jfc4050 please let me know if you need help resolving these conflicts. Thanks!

tjruwase:
@jfc4050, @raamjad, @zarzen. Update #1398 is now merged.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1,3448 +1,3399 @@
...
+
+def input(msg):
+    return
+
+
+def split_half_float_double(tensors):
+    dtypes = [
+        "torch.cuda.HalfTensor",
+        "torch.cuda.FloatTensor",
+        "torch.cuda.DoubleTensor"
```
szhengac, line 64: bfloat16 is missing.

jfc4050: thanks for the catch. i think this might be dead code now will remove it if so

jfc4050: was dead code, removed it

deepspeed/runtime/config.py
```diff
@@ -805,8 +805,6 @@ def _initialize_params(self, param_dict):
         self.gradient_clipping = get_gradient_clipping(param_dict)
         self.fp16_enabled = get_fp16_enabled(param_dict)
         self.bfloat16_enabled = get_bfloat16_enabled(param_dict)
-        assert not (self.fp16_enabled and self.bfloat16_enabled), 'bfloat16 and fp16 modes cannot be simultaneously enabled'
```
manuelciosici, line 808: Are you sure this line should be deleted? I can't imagine how one would use `bfloat16` and `fp16` at the same time.

jfc4050: these lines were duplicated somehow during the merge/conflict resolution. the check still remains below

manuelciosici: I see that [now](https://github.com/microsoft/DeepSpeed/blob/bf0dd663cab9080a24c1972a40634aec7606f45e/deepspeed/runtime/config.py#L811). Thank you for the clarification.

jfc4050:
@tjruwase would you mind approving the workflow run? i think everything should pass now

tests/unit/test_aio.py
```diff
@@ -4,8 +4,8 @@
 import torch
 import deepspeed
 import torch.distributed as dist
-from common import distributed_test
 from deepspeed.ops.aio import AsyncIOBuilder
+from .common import distributed_test
```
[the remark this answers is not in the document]
jfc4050: not necessary, might have happened on accident when resolving merge conflicts. i will put it back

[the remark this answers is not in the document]
tjruwase: Thanks. It is not a big deal, I was just curious if something subtle was going on. No need to change.

tjruwase:
@jfc4050, thanks for this PR. It is passing all tests and so all that is left for merging is code review, which I will prioritize to complete by Monday. Also, we will run some correctness tests over here.

tjruwase:
@jfc4050, in the meantime could you rerun your perf experiments to check for any changes in the gains? Thanks.

jfc4050:
> @jfc4050, thanks for this PR. It is passing all tests and so all that is left for merging is code review, which I will prioritize to complete by Monday. Also, we will run some correctness tests over here.

> @jfc4050, in the meantime could you rerun your perf experiments to check for any changes in the gains? Thanks.

thanks! i will run those tests as soon as possible (been meaning to anyways), we are a little constrained by hardware availability right now so just giving a heads up that there will be some delays

jfc4050:
hi @tjruwase just to give you an update on performance testing, i'm still waiting on machines - is it ok if we do performance re-testing after addressing major review feedback?

also looks like some of the merge conflicts are showing whole file conflicts, should i be using CRLF or LF line endings?

jfc4050:
also do you have a rough idea of when you will be able to take a look?

tjruwase:
> hi @tjruwase just to give you an update on performance testing, i'm still waiting on machines - is it ok if we do performance re-testing after addressing major review feedback?

This sounds reasonable to me. 

> also looks like some of the merge conflicts are showing whole file conflicts, should i be using CRLF or LF line endings?
I don't think CRLF vs LF is the issue here. I think git is just acting up because I also saw similar issues recently. Let me take a shot at resolving the conflicts.

tjruwase:
> also do you have a rough idea of when you will be able to take a look?

I should complete by Monday EOD.

jeffra:
> hi @tjruwase just to give you an update on performance testing, i'm still waiting on machines - is it ok if we do performance re-testing after addressing major review feedback?
> 
> also looks like some of the merge conflicts are showing whole file conflicts, should i be using CRLF or LF line endings?

One issue is solved I think we're using CRLF across the codebase so pushed a fix for that with `stage3.py`. The other conflicts @tjruwase and I will look at shortly.

deepspeed/runtime/zero/stage3.py
```diff
@@ -165,383 +176,329 @@ def _inject_parameters(module, cls):
...
-        self.step_id += 1
+    def record_trace(self, sub_module: Module) -> None:
+        """adds sub module to trace"""
+        if self.trace_complete:
+            raise RuntimeError(
+                "attemted to record trace when trace was already complete")
+
+        self.__submodule_order.append(sub_module)
+        for param in sorted(set(iter_params(sub_module)), key=lambda p: p.ds_id):
+            self.__param_order.append(
```
bazpasha, line 263: Hi! This doesn't seem to take dangling external parameters into account. If I got it right, in case there's a parameter requested outside of its module, it will be assigned as an external parameter to the parent of the module that requested it, so the order of the parameters might change after tracing. It causes errors later on in `fetch_sub_module`

deepspeed/runtime/zero/stage3.py
```diff
@@ -165,383 +176,329 @@ def _inject_parameters(module, cls):
...
+                filter(
+                    lambda p: self.__most_recent_step_id_param_fetched_for[p] < self.
+                    __step_id,
+                    params_to_fetch))
+            while self.__param_queue and len(discarded_from_prefetch_queue) < len(
+                    params_not_already_fetched):
+                param_in_trace = self.__param_queue.popleft()
+                self.__most_recent_step_id_param_fetched_for[
+                    param_in_trace.param] = param_in_trace.step_id_last_used_at
+                discarded_from_prefetch_queue.add(param_in_trace.param)
```
bazpasha, line 363: This can be tricky with external parameters and the `if` below might trigger the exception. I've changed the code as suggested (i.e. I just skip external parameters), but it's a hack
```suggestion
            params_not_already_fetched = set(
                filter(
                    lambda p: self.__most_recent_step_id_param_fetched_for[p] < self.
                    __step_id and not p.is_external_param,
                    params_to_fetch))
            while self.__param_queue and len(discarded_from_prefetch_queue) < len(
                    params_not_already_fetched):
                param_in_trace = self.__param_queue.popleft()
                self.__most_recent_step_id_param_fetched_for[
                    param_in_trace.param] = param_in_trace.step_id_last_used_at
                if not param_in_trace.param.is_external_param:
                    discarded_from_prefetch_queue.add(param_in_trace.param)

```

deepspeed/runtime/zero/stage3.py
jfc4050: hey, my understanding of the way this should work is that external parameters are registered beforehand, then the sequence of events would be
1. external param gathered as part of the module that is passing it
2. because the parameter is registered as external, it doesn't get partitioned after the module has finished - this way it remains available for downstream modules that use it.

if that is the case, this prefetching code doesn't need to care about external parameters

deepspeed/runtime/zero/stage3.py
jfc4050: have some questions about this, see this comment https://github.com/microsoft/DeepSpeed/pull/1453#discussion_r753365239

deepspeed/runtime/comm/coalesced_collectives.py
```diff
@@ -0,0 +1,116 @@
...
+
+    TODO. see if PyTorch team wants a c++ verson of this for ProcessGroupNCCL
+    """
+    this_rank = torch.distributed.get_rank(group)
+    world_sz = torch.distributed.get_world_size(group)
+
+    partition_lst_for_each_tensor = [None] * len(tensors)
+    for tensor_idx, tensor in enumerate(tensors):
+        flattened_tensor = tensor.view(-1)
+        chunk_sz = math.ceil(tensor.numel() / world_sz)
```
tjruwase, line 58: It seems this partitioning logic works well if tensor.numel() is multiple of world size, otherwise it breaks and creates empty partitions. Can you please double check corner-cases like tensor.numel() == 7 and world_size == 6?

szhengac: The empty tensor will be padded in https://github.com/microsoft/DeepSpeed/blob/2b5f6ea2bba91ea1a5e5b9d2febedf041926d810/deepspeed/runtime/comm/coalesced_collectives.py#L83-L89 and skipped in https://github.com/microsoft/DeepSpeed/blob/2b5f6ea2bba91ea1a5e5b9d2febedf041926d810/deepspeed/runtime/comm/coalesced_collectives.py#L109-L114. I previously have tested the case where tensor.numel() = 4 and world_size = 128.

tjruwase: Awesome, thanks @szhengac!!!

deepspeed/runtime/engine.py
```diff
@@ -729,6 +729,9 @@ def bfloat16_enabled(self):
     def fp16_master_weights_and_gradients(self):
         return self._config.fp16_master_weights_and_gradients
 
+    def bfloat16_enabled(self):
```
tjruwase, line 732: Now redundant because of [this](https://github.com/microsoft/DeepSpeed/blob/a637cc2cd508381d79a522c98fef0d4e747ef28f/deepspeed/runtime/engine.py#L726-L727).

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -3,29 +3,68 @@
...
+from deepspeed.utils.logging import logger
 
 from ..swap_tensor.partitioned_param_swapper import AsyncPartitionedParameterSwapper, PartitionedParamStatus
 from ..config import DeepSpeedConfig
 
 param_count = 0
-partitioned_param_data_shape = [1]
+partitioned_param_data_shape = [0]
+
+if hasattr(torch.distributed, "_all_gather_base"):
```
tjruwase, line 38: Could we avoid duplication by import/reuse this logic from [here](https://github.com/jfc4050/DeepSpeed/blob/2b5f6ea2bba91ea1a5e5b9d2febedf041926d810/deepspeed/runtime/comm/coalesced_collectives.py#L16-L38)?

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -3,29 +3,68 @@
...
+from deepspeed.utils import init_distributed, instrument_w_nvtx, logger
 from deepspeed.utils.debug import debug_param2name_id_shape, debug_param2name_id_shape_device, debug_module2name, debug_param2name, debug_param2name_id_shape_status, printflock, log_rank_file
+from deepspeed.utils.logging import logger
 
 from ..swap_tensor.partitioned_param_swapper import AsyncPartitionedParameterSwapper, PartitionedParamStatus
 from ..config import DeepSpeedConfig
 
 param_count = 0
-partitioned_param_data_shape = [1]
+partitioned_param_data_shape = [0]
```
tjruwase, line 36: Looks like `partitioned_param_data_shape` is not originally used. Is there a reason for this change?

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -332,11 +466,79 @@ def _post_init_method(self, module):
...
+                self.dtype = torch.float
         else:
-            self.dtype = dtype
+            self.dtype = dtype or torch.half
+
+
+class AllGatherHandle:
+    def __init__(self, handle, param: Parameter) -> None:
+        self.__handle = handle
+        self.__param = param
```
tjruwase, line 485: Is it useful to assert that ```param.ds_status == ZeRoParamStatus.INFLIGHT```, like is done in `AllGatherCoalescedHandle`?
```

File: deepspeed/runtime/zero/partition_parameters.py [hunk 11]
```diff
@@ -548,6 +750,14 @@ def _post_init_method(self, module):
                 print_rank_0(
                     f"Partitioning param {debug_param2name_id_shape(param)} module={debug_module2name(module)}"
                 )
+
+                if param.is_cuda:
+                    torch.distributed.broadcast(param, 0, self.ds_process_group)
```
tjruwase, line 755: Why is broadcast needed here? Did you discover a scenario where the normal partition/gather sequence is not sufficient to handle?

deepspeed/runtime/zero/stage3.py
```diff
@@ -1,3474 +1,3351 @@
...
-                    zero_init=zero_init,
-                    additional_buffer_factor=additional_buffer_factor
-                )
-
-                options_str = format_options(cpu_offload=cpu_offload,
-                                             cpu_offload_params=cpu_offload_params,
-                                             zero_init=zero_init)
-                print(
-                    f" {cpu_mem/2**30:7.2f}GB | {gpu_mem/2**30:6.2f}GB | {options_str}")
+"""
```
tjruwase, line 1: @jfc4050, the changes in stage3.py are significant. Thanks for this massive contribution. Unfortunately, it is going to take a bit more time and help from teammates to complete. Sorry about that.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1,3474 +1,3351 @@
...
-    # printflock(f"[{rank}] {message}")
-    # - print to log file per rank
-    # log_rank_file(rank, message)
-
-
-def input(msg):
-    return
-
-
-def split_half_float_double(tensors):
```
tjruwase, line 53: I am unable to find how this dtype-based communication is handled in the new code. Is it no longer needed?

deepspeed/runtime/zero/partition_parameters.py
szhengac: We need to make sure each rank receives the respective chunk from the same weight tensor. This is important if some pytorch module uses the tensor based initialization method such as orthogonal initialization. This is also why we first gather tensor before apply the user defined init function in https://github.com/jfc4050/DeepSpeed/blob/s3-pr/deepspeed/runtime/zero/partition_parameters.py#L331-L332. It also helps unit-test for the correctness.

deepspeed/runtime/zero/stage3.py
szhengac: I think the commit does not handle the case where they are multiple dtypes. By default, we assume only a single dtype is used and all the partitioned gradient are fused into a large buffer. @jfc4050 pls correct me if I am wrong.

deepspeed/runtime/zero/partition_parameters.py
tjruwase: @szhengac, thanks for the explanation. I am not familiar with this orthogonal initialization approach, but that is not a big deal. What is not clear to me is the following: 
1. Will the all-gather of a given tensor happen on all ranks?
2. If so, then won't the tensor value be the same on all the ranks?
3. If so, what will the broadcast achieve?

Can you please clarify which of the 3 steps I am getting wrong? Thanks.

jfc4050: 1. yes
2. no, because after the allgather we perform some initialization operation that will give different results on each rank unless seeding is carefully controlled
3. the broadcast ensures that rank 0 is used as the "master" set of initialized weights - this is important if elements are not initialized independently - a simple example is lets say we have an initialization operation that results in the parameter having norm 1, two ranks may individually initialize in a way that satisfies this property, but if you had each rank partition and then did an allgather the resulting tensor is not guaranteed to satisfy this property

deepspeed/runtime/zero/partition_parameters.py
jfc4050: sure thats worth adding

deepspeed/runtime/zero/partition_parameters.py
jfc4050: hm, no might just be vestigial from something i was doing earlier. will revert this

deepspeed/runtime/zero/partition_parameters.py
jfc4050: sorry i dont fully understand, can you clarify?

deepspeed/runtime/zero/stage3.py
jfc4050: yeah thats correct, it seems to me like `stage3.py` assumes that everything is a single dtype in a few places though, for example https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/zero/stage3.py#L1190-L1193

deepspeed/runtime/zero/stage3.py
jfc4050: no problem! thanks for all the help reviewing. please let me know if anything is unclear and i will make an effort to document better

deepspeed/runtime/zero/partition_parameters.py
tjruwase: Got it. We have always assumed seeding is carefully controlled, which is of course is not full-proof :). This looks like a bug fix to me, thanks!

deepspeed/runtime/zero/stage3.py
tjruwase: That is fair. To my knowledge we are yet to experience a multi-dtype use case in ZeRO-3. Need to watch for possible conflict with #1487.

deepspeed/runtime/zero/partition_parameters.py
tjruwase: I mean these codes in [partition_parameters.py](https://github.com/jfc4050/DeepSpeed/blob/2b5f6ea2bba91ea1a5e5b9d2febedf041926d810/deepspeed/runtime/zero/partition_parameters.py#L38-L67) and [coalesced_collectives.py](https://github.com/jfc4050/DeepSpeed/blob/2b5f6ea2bba91ea1a5e5b9d2febedf041926d810/deepspeed/runtime/comm/coalesced_collectives.py#L16-L38) are quite similar, or perhaps I am missing the subtle difference. So I was wondering if we could avoid the code duplication.

deepspeed/runtime/zero/stage3.py
```diff
@@ -73,16 +65,34 @@ def lcm(x, y):
     return x * y // gcd(x, y)
 
 
+def debug_rank0(message: str) -> None:
+    if dist.get_rank() == 0:
+        logger.debug(message)
+
+
+def get_cuda_mem_allocated_str() -> str:
```
tjruwase, line 73: This looks like a subset of [see_memory_usage](https://github.com/microsoft/DeepSpeed/blob/c3f1d82b07b7541ee03961b07959b9f38894de5e/deepspeed/runtime/utils.py#L812). Is it needed?

jfc4050: i think we can get rid of it

deepspeed/runtime/zero/partition_parameters.py
jfc4050: there's some subtle differences, i'm not sure how if i know how to reuse without complicating/obfuscating things. can you suggest a change?

deepspeed/runtime/zero/partition_parameters.py
jfc4050: since its not used do you mind if i remove it to avoid confusing readers? this PR changes the shape of partitioned parameters to (0,) instead of (1,)

tjruwase: Sure, please remove

deepspeed/runtime/zero/partition_parameters.py
tjruwase: In that case, don't worry about it for now. We can always revisit later.

deepspeed/runtime/zero/partition_parameters.py
jfc4050: oh nvm, its actually used but only in unit tests. so its actually required to have
`partitioned_param_data_shape = [0]` otherwise the tests fail

deepspeed/runtime/zero/partition_parameters.py
jfc4050: added

deepspeed/runtime/zero/stage3.py
jfc4050: removed

deepspeed/runtime/engine.py
jfc4050: removed

deepspeed/runtime/zero/partition_parameters.py
jfc4050: yw! i'll add some comments here, if you found it unintuitive i bet others will as well

stas00:
2 questions:

1. is this PR stripping all the windows \r line endings and thus all files are rewritten? It makes it very difficult to see what actually changed - and while it's possible to tell github to ignore whitespace and then it's better, this is going to impact all future pushes since those will continue coming from the windows land with \r newlines, so it's probably the best to stick to the original style.

If I'm not mistaken: `pre-commit run --all-files` should fix them all? If not `unix2dos`?

2. I'm working on the preparing for the integration of this into HF Transformers, and I'd like to know what replaced `stage3_gather_fp16_weights_on_model_save` in z3?

Are there any other bf16-specific config params that were added or removed?

Thank you!

stas00:
Additionally, why the inconsistent naming? 
```
        "fp16": {
            "enabled": False
        },
        "bfloat16": {
            "enabled": True
        },
```

Could we please use `bf16` to match `fp16`? 

Thank you!

stas00:
One more thing - how can I test this PR if it asserts?
```
AssertionError: bfloat16 mode is only enabled for Zero2 currently
```

jfc4050:
> is this PR stripping all the windows \r line endings and thus all files are rewritten?

we fixed this earlier but i may have accidentally broken it, will take a look

> I integrated bf16 into HF Transformers, [WIP] [Deepspeed] add support for bf16 mode huggingface/transformers#14569 and I'd like to know what replaced stage3_gather_fp16_weights_on_model_save in z3?

i don't see anything matching `stage3_gather_fp16_weights_on_model_save` from a search through DS code - could have been removed by a prior change

> Are there any other bf16-specific config params that were added or removed?

no, at least not with this PR

> One more thing - how can I test this PR with zero3 if it asserts?

this might have gotten broken by merge conflict resolution - will take a look

stas00:
> > I integrated bf16 into HF Transformers, [WIP] [Deepspeed] add support for bf16 mode [huggingface/transformers#14569](https://github.com/huggingface/transformers/pull/14569) and I'd like to know what replaced stage3_gather_fp16_weights_on_model_save in z3?
> 
> i don't see anything matching `stage3_gather_fp16_weights_on_model_save` from a search through DS code - could have been removed by a prior change

```
grep -Ir stage3_gather_fp16_weights_on_model_save deepspeed
deepspeed/runtime/engine.py:                    f"Did not save the model {path} because `stage3_gather_fp16_weights_on_model_save` is False"
deepspeed/runtime/zero/constants.py:ZERO_OPTIMIZATION_GATHER_FP16_WEIGHTS_ON_MODEL_SAVE = 'stage3_gather_fp16_weights_on_model_save'
deepspeed/autotuning/config_templates/template_zero3.json:    "stage3_gather_fp16_weights_on_model_save": false,
```

so it's there. Can you see it?

This flag was added to handle the reconstruction of the weights on model save, so perhaps a similar one needs for bf16 or to rename it for a generic ` stage3_gather_half_precision_weights_on_model_save` - with a back-compat deprecated old name 

Thank you!

jfc4050:
oh whoops messed up my search before. i see it now. `stage3_gather_half_precision_weights_on_model_save` or similar sounds good to me

deepspeed/runtime/zero/stage3.py
```diff
@@ -2984,42 +2867,49 @@ def backward(self, loss, retain_graph=False):
...
-        self.micro_step_id += 1
-        print_rank_0(
-            f"Total fully available parameters {self.param_coordinator.total_available_parameter_numel}"
-        )
-
         if self.swap_optimizer:
             self.optimizer_swapper.pre_backward()
 
         see_memory_usage(f"Before backward", force=False)
-        if self.contiguous_gradients:
```
tjruwase, line 2996: @jfc4050, it seems the contiguous_gradients and overlap_communication logic is removed, Is that correct?

jfc4050: contiguous gradients and communication overlap are still supported, its just that this piece of code doesnt need to be aware of either anymore.

the IPG buffer (there's one of them now rather than two) is now allocated once during initialization and kept for the length of training

jfc4050:
> One more thing - how can I test this PR with zero3 if it asserts?
@stas00 just pushed a fix to this, should work now. looks like it got broken by merge conflict resolutions

tjruwase:
> > One more thing - how can I test this PR with zero3 if it asserts?
> > @stas00 just pushed a fix to this, should work now. looks like it got broken by merge conflict resolutions

@jfc4050, incidentally I am also tinkering with your branch. I just resolved a bunch of conflicts. So I hope we are not stomping over each others changes. I will try sync with your latest branch before pushing again.

jfc4050:
>@jfc4050, incidentally I am also tinkering with your branch. I just resolved a bunch of conflicts. So I hope we are not stomping over each others changes. I will try sync with your latest branch before pushing again.

i just made a one-liner change to the config.py file - if you do stomp on my changes it should be very easy to un-stomp :)

stas00:
Great work, I have enabled both z2 and z3 bf16 in my functional tests and they all pass: https://github.com/huggingface/transformers/pull/14569

Thank you for fixing the white-space, now the PR is readable.

So for me 2 outstanding issues are:

1. what we do with `stage3_gather_fp16_weights_on_model_save`
https://github.com/microsoft/DeepSpeed/pull/1453#issuecomment-986174591
follow up was here: https://github.com/microsoft/DeepSpeed/pull/1453#issuecomment-987372525

2. asking to rename `bfloat16` to `fp16` in the config file:
https://github.com/microsoft/DeepSpeed/pull/1453#issuecomment-986177071

jeffra:
> A big question: Are you planning on having some Ampere GPUs on CI? We don't at HF, so it'd be great if you did, otherwise bf16 won't be tested. And through our tests you will have a great coverage in addition to your tests.

I don't think we are opposed to having Ampere GPU(s) in our CI. However, I think we should be able to test some aspects of bf16 even without the hardware support for it (just like we can run fp16 on P40 even though it doesn't have TCs).

stas00:
How would that work if the underlying software asserts when the hardware is wrong?

```
CUDA_VISIBLE_DEVICES=1 python -c "import torch; torch.cuda.amp.autocast(dtype=torch.bfloat16)"
  File "/home/stas/anaconda3/envs/py38-pt111/lib/python3.8/site-packages/torch/cuda/amp/autocast_mode.py", line 25, in __init__
    super().__init__("cuda", enabled=enabled, dtype=dtype, cache_enabled=cache_enabled)
  File "/home/stas/anaconda3/envs/py38-pt111/lib/python3.8/site-packages/torch/autocast_mode.py", line 178, in __init__
    raise RuntimeError('Current CUDA Device does not support bfloat16. Please switch dtype to float16.')
RuntimeError: Current CUDA Device does not support bfloat16. Please switch dtype to float16.
```

jeffra:
> How would that work if the underlying software asserts when the hardware is wrong?
> 
> ```
> CUDA_VISIBLE_DEVICES=1 python -c "import torch; torch.cuda.amp.autocast(dtype=torch.bfloat16)"
>   File "<string>", line 1, in <module>
>   File "/home/stas/anaconda3/envs/py38-pt111/lib/python3.8/site-packages/torch/cuda/amp/autocast_mode.py", line 25, in __init__
>     super().__init__("cuda", enabled=enabled, dtype=dtype, cache_enabled=cache_enabled)
>   File "/home/stas/anaconda3/envs/py38-pt111/lib/python3.8/site-packages/torch/autocast_mode.py", line 178, in __init__
>     raise RuntimeError('Current CUDA Device does not support bfloat16. Please switch dtype to float16.')
> RuntimeError: Current CUDA Device does not support bfloat16. Please switch dtype to float16.
> ```
> 
> as you're saying it may work for some aspects of bf16.

Haha nevermind :p you are right, I think we'll need Ampere in our CI... Didn't realize torch raises an exception like this in some cases.

stas00:
and so additionally perhaps Deepspeed should mimic torch and also assert if bf16 is not supported and a user is trying to use it? Otherwise they are going to have a really terrible performance with bf16 emulation, no?

jfc4050:
hey @tjruwase, any progress on reviews?

deepspeed/runtime/zero/stage3.py
```diff
@@ -165,383 +168,328 @@ def _inject_parameters(module, cls):
...
-            self.sub_module_trace.append(sub_module.id)
-            self.id_to_sub_module_map[sub_module.id] = sub_module
+    def record_trace(self, sub_module: Module) -> None:
+        """adds sub module to trace"""
+        if self.trace_complete:
+            raise RuntimeError(
+                "attemted to record trace when trace was already complete")
+
+        self.__submodule_order.append(sub_module)
+        for param in sorted(set(iter_params(sub_module)), key=lambda p: p.ds_id):
```
tjruwase, line 254: Is the sorting here on ds_id important? Does it assume that ds_id represents the order in which the parameters are encountered in forward?

deepspeed/runtime/zero/stage3.py
```diff
@@ -165,383 +168,328 @@ def _inject_parameters(module, cls):
...
 
-    def reset_step(self):
-        self.step_id = 0
+        # kick off all gather for params in the immediately required submodule
+        for param in params_to_fetch:
+            debug_rank0(f"-fetch: {param.ds_summary()}")
+        self.__all_gather_params(params_to_fetch)
+
+        # wait for parameters in the immediately needed submodule to become available
+        for param in iter_params(current_submodule):
```
tjruwase, line 314: Could this not iterate over `params_to_fetch`? Or do we expect to find a different list of parameters here?

deepspeed/runtime/zero/stage3.py
```diff
@@ -165,383 +168,328 @@ def _inject_parameters(module, cls):
...
+            param.ds_active_sub_modules.add(current_submodule.id)
+            debug_rank0(f"-wait: {param.ds_summary()}")
+            if param in self.__inflight_param_registry:
```
tjruwase, line 317: Should this not always be true since we called `record_trace()` on this submdule before calling `fetch_sub_module()`? I am curious of the situations where an assert here would be incorrect.

tjruwase:
> hey @tjruwase, any progress on reviews?

Happy New Year! Apologies for the delay, but the review is progress. I think I should finish today. I am leaving comments/questions as I go along stage3.py, which is the last file. 

In the meantime, can you please address the comments from @stas00:  https://github.com/microsoft/DeepSpeed/pull/1453#issuecomment-989400089? 

Thanks!

deepspeed/runtime/zero/stage3.py
```diff
@@ -812,12 +771,40 @@ def __init__(self,
...
+
+        self.__params_in_ipg_bucket: List[Parameter] = []
+        self.is_gradient_accumulation_boundary: bool = True
 
-        self.reduction_stream = torch.cuda.Stream(
-        ) if self.overlap_comm else torch.cuda.current_stream()
-        self.callback_queued = False
-        self.copy_grad_stream = torch.cuda.Stream()
+        self.__param_reduce_events: Deque[Event] = collections.deque()
+        self.__max_param_reduce_events: int = 2
```
tjruwase, line 804: Does it make sense to make this a json configurable value?

deepspeed/runtime/zero/stage3.py
```diff
@@ -912,6 +892,43 @@ def __init__(self,
         if dist.get_rank(group=self.dp_process_group) == 0:
             see_memory_usage(f"After initializing ZeRO optimizer", force=False)
 
+    @staticmethod
+    def defragment(tensors: List[Tensor]) -> Tensor:
```
tjruwase, line 896: Awesome!

deepspeed/runtime/zero/stage3.py
```diff
@@ -2385,15 +2223,20 @@ def get_partition_info(self, tensor_list, partition_size, partition_id):
...
         """
+        self.micro_step_id = 0
+
         # FP32 grad should never exist.
         # For speed, set model fp16 grad to None by default
         for group in self.fp16_groups:
             for p in group:
                 if set_grads_to_None:
+                    if p.grad is not None and p.grad.is_cuda:
+                        p.grad.record_stream(torch.cuda.current_stream())
```
tjruwase, line 2239: Nice catch!

deepspeed/runtime/zero/stage3.py
```diff
@@ -2828,6 +2699,8 @@ def unscale_and_clip_grads(self, sub_group_id, total_norm):
             clip = ((total_norm / self.loss_scale) + 1e-6) / self.clip_grad
             if clip > 1:
                 combined_scale = clip * self.loss_scale
+        # to maintain behavior of averaging over accumulation steps
+        combined_scale *= self.micro_step_id + 1
```
tjruwase, line 2703: Won't self.micro_step_id be zero at this point because of [this](https://github.com/microsoft/DeepSpeed/blob/e236293bd610ff793a6a1a3bf294b2b6f4b254a3/deepspeed/runtime/zero/stage3.py#L2430) and [this](https://github.com/microsoft/DeepSpeed/blob/e236293bd610ff793a6a1a3bf294b2b6f4b254a3/deepspeed/runtime/zero/stage3.py#L2392)?

deepspeed/runtime/zero/stage3.py
```diff
@@ -2916,42 +2790,49 @@ def backward(self, loss, retain_graph=False):
...
-        Usually if parameters of modules whose input parameters do not require
-        grad computation do not trigger post call and will therefore will remain unpartitioned '''
-        self._partition_all_parameters()
+
+        self.param_coordinator.reset_step()
 
         if self.swap_optimizer:
             self.optimizer_swapper.post_backward()
 
+    def get_fp32_grad_partitions(self) -> Dict[int, Dict[int, Tensor]]:
```
tjruwase, line 2805: Is there use of this function? Also, do it handle nvme swapping?

deepspeed/runtime/zero/stage3.py
jfc4050: i don't think it matters in most cases but its just a cheap way to get additional safety. under the hood, `iter_params` gets the parameters by using `module.named_parameters()` and afaict PyTorch has not explicitly guaranteed any ordering of the returned parameters

to DeepSpeed, we encounter all parameters for a given (leaf) module at once, since we need them all before executing the module. the parameters for a single module are allgathered as a batch. The sorting is to avoid the possibility of parameters for a module being in a different order for different ranks in the batch call. This would cause subtle correctness issues from mixing of data between parameters

deepspeed/runtime/zero/stage3.py
jfc4050: yep, youre right there's no need to make another iterator. will change it to use `params_to_fetch`

deepspeed/runtime/zero/stage3.py
jfc4050: which assert? or are you talking about `if param in self.__inflight_param_registry:`

jfc4050:
> > hey @tjruwase, any progress on reviews?
> 
> Happy New Year! Apologies for the delay, but the review is progress. I think I should finish today. I am leaving comments/questions as I go along stage3.py, which is the last file.
> 
> In the meantime, can you please address the comments from @stas00: [#1453 (comment)](https://github.com/microsoft/DeepSpeed/pull/1453#issuecomment-989400089)?
> 
> Thanks!

Oops that was rude of me :) happy new year! thanks for all the help so far, i know this is a large PR.

i'll start working my way through yours and @stas00's comments

deepspeed/runtime/zero/stage3.py
jfc4050: i wanted to avoid putting them in the public API since they won't be necessary anymore after PyTorch starts using stream ordered `cudaMallocAsync` and `cudaFreeAsync` in their caching memory allocator. 

Maybe we could have it read from env variable or something so savvy users can mess with it, but knowing that DeepSpeed team has the discretion to remove this at any time?

deepspeed/runtime/zero/stage3.py
stas00: Also please be aware of needing to know the exact order of params in flattened tensors, since later we need to be able to reconstruct the flattened partitions into a normal model via this map:

https://github.com/microsoft/DeepSpeed/blob/fead387f7837200fefbaba3a7b14709072d8d2cb/deepspeed/runtime/engine.py#L2886-L2926

(it's then used in zero_to_fp32.py to cold extra fp32 weights).

It was already recently updated to not use `module.named_parameters()` but just want to make sure that this function remains in sync with any new sorting. 

Perhaps it's already working just fine, so I'm only bringing it up to make sure it remains so. Thanks.

deepspeed/runtime/zero/stage3.py
jfc4050: its mostly just to defragment tensor data thats created during initialization and kept throughout the training process (parameter partitions). We found that this helped with memory pressure a bit. 

defragmentation of more transient memory allocations was already handled before i started making changes, and i didn't see much opportunity to further reduce fragmentation without creating a lot of overhead

deepspeed/runtime/zero/stage3.py
stas00: but you will have users still using old pytorch for at least a year or even longer. Deepspeed currently still support pt-1.2, is it not?

it shouldn't be a problem to make things configurable and indicate in the docs that a param x is experimental or a subject to change - as long as the disclaimer is added from the beginning - backward compat doesn't need to be maintained and no "social" agreements will be broken.

deepspeed/runtime/zero/stage3.py
jfc4050: yep i think youre right. will fix this and add a unit test

deepspeed/runtime/zero/stage3.py
stas00: Could we make this accessible to users down the road at training time? I can surely see situations where a user will want to run this occasionally at their own schedule. So that the default run doesn't do it, but it can be configured to defrag periodically or say via a low mem counter trigger?

This is an awesome feature!

deepspeed/runtime/zero/stage3.py
jfc4050: the only place its used is in unit testing, i just wanted a method to use that would give a unified interface for getting grad partitions, regardless of optimizer offloading or not.

For swapping, i believe it will work when tensors are swapped in (i.e this has been called https://github.com/jfc4050/DeepSpeed/blob/master/deepspeed/runtime/zero/stage3.py#L2634-L2652) but i haven't tested that

jfc4050:
> I will repeat my admiration of this incredible engineering work and contribution. Thanks so much.
> 
> Overall, the code looks good to me. I left a few minor comments and there are the two comments by @stas00 to take care of for the final approval.

Thanks! that's very nice of you to say. DeepSpeed has been hugely useful for us, glad to be able to give back.

Working my way through the comments now. as we discussed earlier, after that will do a final performance test to make sure we haven't lost any gains during the PR.

deepspeed/runtime/zero/stage3.py
tjruwase: @jfc4050, thanks for the explanation. I concur with @stas00 that this should ideally be promoted to a utility outside ZeRO. So can you please annotate with a TODO so that we don't forget.

deepspeed/runtime/zero/stage3.py
tjruwase: Sounds good.

deepspeed/runtime/zero/stage3.py
tjruwase: @jfc4050, I think you can add a TODO to make this configurable.

jfc4050: sure that works for me

deepspeed/runtime/zero/stage3.py
jfc4050: works for me, added TODO

deepspeed/runtime/zero/stage3.py
jfc4050: added TODOs.

deepspeed/runtime/zero/stage3.py
tjruwase: Yes, I am curious whether that condition could ever be false.

jfc4050: yes, its possible. the registry is for parameters that are currently in flight - a parameter we need could not be in flight if it was fetched earlier and not partitioned (for example if its persistent or we determined the reuse distance was small enough to keep it around)

regardless of if it was already available or fetched in current step, we expect that the end result is that the parameter is available, which is why we end with
```
assert param.ds_status == ZeroParamStatus.AVAILABLE, param.ds_summary()
```
for each parameter that is currently needed

deepspeed/runtime/zero/config.py
```diff
@@ -171,10 +171,16 @@ def _initialize(self, zero_config_dict):
...
+        # config key has been renamed to use "16bit" instead of "fp16." falling back
+        # to old config name in order to preserve backwards compatibility
+        self.gather_16bit_weights_on_model_save = ZERO_OPTIMIZATION_GATHER_16BIT_WEIGHTS_ON_MODEL_SAVE_DEFAULT
+        for key in [
+                ZERO_OPTIMIZATION_GATHER_16BIT_WEIGHTS_ON_MODEL_SAVE,
+                ZERO_OPTIMIZATION_GATHER_FP16_WEIGHTS_ON_MODEL_SAVE
+        ]:
+            if key in zero_config_dict:
+                self.gather_16bit_weights_on_model_save = zero_config_dict[key]
+                break
```
stas00, line 183: Unless I missed it, I think this change didn't preserve backward compatibility for the original: `gather_fp16_weights_on_model_save` config entry. And given that DS's config checker simply ignores entries it doesn't know, now the users with the old config that still has `gather_fp16_weights_on_model_save` will have a broken behavior.

Thanks.

jfc4050: hey, i believe this specific piece of code will work. added a test to confirm (which i should have done to begin with) https://github.com/microsoft/DeepSpeed/pull/1453/commits/e04dc6a24f9d587f9d45cbd3b6873022951f49ee

stas00: you're absolutely right! I did miss it! Thank you very much for adding a test, Justin.

jfc4050: sure!

jfc4050:
@tjruwase @stas00 i think the outstanding comments have been addressed now, lmk if there's any other changes we need. otherwise, I'll start trying to get some hardware to rerun performance tests

stas00:
@jfc4050, all my requests have been addressed and I have already updated the HF/DS integration https://github.com/huggingface/transformers/pull/14569 and all the many tests pass. All my tests are functional though, none are qualitative as I rely on deespeed to have those.

tjruwase (approved):

tjruwase:
@jfc4050, I will merge latest Friday just to give some space to the PR-MoE release.

jfc4050:
@tjruwase here's results from rerun. they actually got a little better, i'm not sure if its because i've upgraded pytorch/HF or anything - i forgot to record the exact library versions i used to get the performance results reported originally.

on 8 p4d cluster (64 A100 GPUs)
Roberta-10B: 59.23 -> 111.31 TFLOPS/GPU
Roberta-50B: 71.38 -> 127.85 TFLOPS/GPU

---------------------

also i took a look at the failing tests.
```
FAILED unit/test_zero_tiled.py::test_tiled_backward[23-29-2-2-False]
FAILED unit/test_zero_tiled.py::test_tiled_returnbias_backward[23-29-2-2-False]
FAILED unit/test_zero_tiled.py::test_tiled_forward[23-29-2-2-False]
```
 looks like they are for the `TiledLinear` module, and the outputs are off by a small amount - maybe a numerical issue/flaky test. in any case i dont think they are related to any changes made here

tjruwase:
@jfc4050, thanks for sharing the updated perf results. They look great. 

Yes, those unit tests are incredibly flaky. Not to worry about them. Everything looks good for merge Friday.

deepspeed/runtime/zero/stage3.py
jfc4050: yep that is unaffected. this sorting is for parameter ordering within modules in allgather batches, the ordering within the flattened buffers you describe remains the same

stas00:
Wow! Amazing work all who participated!

Please let me know when a new DS release is made and we will roll out the HF/DS integration of bf16 in HF Transformers.

deepspeed/runtime/zero/constants.py
```diff
@@ -116,7 +116,8 @@
 
 # gathers params for saving a model - inefficient but is required in certain situations
```
bliu3650, line 117: Could someone kindly help explain which situations were required by this 16bit parameters gathering (infeeicient) feature, given that there is zero_to_fp32.py script which can help save the parameters?  Thanks.

stas00: This is for those who can't be bothered with running `zero_to_fp32.py` and want the 16-bit model extracted on the fly - which is fine for tiny to small models but very slow for large models.

It's also the default in the HF Trainer integration of Deepspeed to make it easy for users to start and have things work transparently. But the documentation explains how to improve upon this default.
https://huggingface.co/docs/transformers/main/main_classes/deepspeed#getting-the-model-weights-out

PR #1456, opened 2021-10-14
Switch to use or not einsum op. Needed for ORT

ganik:
Replace einsum op as its not supported by onnxruntime.
Branch by USE_EINSUM flag.
Example:

...
deepspeed.moe.sharded_moe.USE_EINSUM = False
...
self.fc3 = deepspeed.moe.layer.MoE(..)
...


ykim362:
Can ORT support einsum? Einsum is becoming more universal especially in the context of MoE.

PR #1487, opened 2021-10-27
allreduce_always_fp16

Dipet:
The same logic as `allreduce_always_fp32` but using fp16 data type. It can be used when we have a very large model and PCIe bandwidth is insufficient.
Also changed `new_zeros` to `new_empty` in `sparse_all_gather` because we do not need zeros arrays there.


deepspeed/runtime/engine.py
```diff
@@ -1754,9 +1759,13 @@ def allreduce_bucket(self, bucket, dp_group):
                     tensor_to_allreduce.mul_(self.gradient_predivide_factor() /
                                              dist.get_world_size(group=dp_group))
         else:
-            tensor_to_allreduce.div_(dist.get_world_size(group=dp_group))
+            tensor_to_allreduce.mul_(1. / dist.get_world_size(group=dp_group))
```
tjruwase, line 1762: Can you please explain the benefit of this change?

deepspeed/runtime/utils.py
```diff
@@ -266,7 +266,8 @@ def _has_inf_or_nan(x, i):
             # if x is half, the .float() incurs an additional deep copy, but it's necessary if
             # Pytorch's .sum() creates a one-element tensor of the same type as x
             # (which is true for some recent version of pytorch).
-            cpu_sum = float(x.float().sum())
+            cpu_sum = float(torch.sparse.sum(x.float())) if x.is_sparse else float(
```
tjruwase, line 269: This logic seems to be unrelated to fp16 reduce. I think it should be a separate PR and appropriate documented.

deepspeed/runtime/engine.py
```diff
@@ -1742,6 +1745,8 @@ def allreduce_bucket(self, bucket, dp_group):
 
         if self.allreduce_always_fp32():
             tensor_to_allreduce = tensor.float()
+        elif self.allreduce_always_fp16():
+            tensor_to_allreduce = tensor.half()
```
tjruwase, line 1749: Does this mean that it is possible to configure fp32 training with fp16 reduction? I can't think of scenarios where that might be useful? Can you please share more details on the motivation for this PR?

tjruwase:
> The same logic as `allreduce_always_fp32` but using fp16 data type. It can be used when we have a very large model and PCIe bandwidth is insufficient.

Thanks for the PR. As I mentioned elsewhere, can you please share more motivation for this PR? I am also curious as to how PCIe bandwidth affects gradient reduction.

deepspeed/runtime/engine.py
Dipet: Multiplication is 2-3 times faster than division. There we divide 1 number and after that multiply array to this value. This is faster than dividing the whole array to value.

```python
import time
import numpy as np

arr = np.random.randn(1024, 1024, 3).astype(np.float32)
val = 15648931
val_inverse = 1.0 / val
N = 1000
print(arr.shape, arr.min(), arr.max())

s = time.time()
for _ in range(N):
    res = arr / val
dt_div = time.time() - s

s = time.time()
for _ in range(N):
    res = arr * val_inverse
dt_mul = time.time() - s

print(f"dt_div: {dt_div:.3f} dt_mul: {dt_mul:.3f}")
# dt_div: 3.730 dt_mul: 1.734
```

deepspeed/runtime/engine.py
Dipet: I have a very large model (5-10 GB of weights) on the backward step when we try to synchronize gradients we need to transfer whole this weights between GPUs. If we have no NVLINK and we want to use fp32 weights to train we will be limited by PCIe bandwidth (5-10 GB/s for PCIe3 and 10-20 GB/s for PCIe4). Transfering gradients in fp16 and after that applying all arithmetic operations using fp32 we do not lose a lot of quality.

deepspeed/runtime/engine.py
tjruwase: Thanks for the explanation. My recollection is that modern compilers automatically perform such optimizations :).

deepspeed/runtime/engine.py
tjruwase: That makes sense. I was not aware of such scenarios, thanks for expanding my knowledge. The only concern I have is that we now have two conflicting configuration flags for specifying reduction data type. It would be preferable to consolidate the reduction data type into a single configuration flag:
```json
{
  "reduction_type": "fp16 | fp32 | bfloat16 | int8, ..."
}
```
What do you think?

Dipet: Sounds logical. I will rewrite the code to use a single flag.

deepspeed/runtime/engine.py
Dipet: > My recollection is that modern compilers automatically perform such optimizations

No, at first np.ndarray is complex data type and compiler do not know how to optimize overloaded operation.
And another point that in some cases division and multiplication to inverse number is not fully equivalent because float data type has some rounding errors.

tjruwase: Fair enough. Thanks.

tjruwase:
@Dipet, I am seeing your frequent updates on this PR. Can you please ping me when you think it is ready to review? Thanks so much for this massive contribution to DeepSpeed.

Dipet:
@tjruwase I fixed errors. And replaced  `allreduce_always_fp16` and `allreduce_always_fp32` by single argument - `communication_data_type`

deepspeed/runtime/engine.py
```diff
@@ -1763,10 +1764,10 @@ def allreduce_bucket(self, bucket, dp_group):
                     tensor_to_allreduce.mul_(self.gradient_predivide_factor() /
                                              dist.get_world_size(group=dp_group))
         else:
-            tensor_to_allreduce.div_(dist.get_world_size(group=dp_group))
+            tensor_to_allreduce.mul_(1. / dist.get_world_size(group=dp_group))
             dist.all_reduce(tensor_to_allreduce, group=dp_group)
 
-        if self.allreduce_always_fp32() and tensor is not tensor_to_allreduce:
+        if self.communication_data_type is not None and tensor is not tensor_to_allreduce:
```
tjruwase, line 1770: Do we want to copy if communication_data_type is fp16? Won't that deviate from previous behavior?

deepspeed/runtime/zero/stage2.py
```diff
@@ -199,7 +199,7 @@ def __init__(self,
             assert self.cpu_offload and type(self.optimizer) in [DeepSpeedCPUAdam], f"fp16_master_and_gradients requires optimizer to support keeping fp16 master and gradients while keeping the optimizer states in fp32. Currently only supported using ZeRO-Offload with DeepSpeedCPUAdam. But current setting is ZeRO-Offload:{self.cpu_offload} and optimizer type {type(self.optimizer)}. Either disable fp16_master_weights_and_gradients or enable ZeRO-2 Offload with DeepSpeedCPUAdam"
 
         if self.reduce_scatter:
-            assert not self.allreduce_always_fp32, "allreduce_always_fp32 is not yet supported with ZeRO-2 with reduce scatter enabled"
+            assert self.communication_data_type is None, "communication_data_type is not yet supported with ZeRO-2 with reduce scatter enabled"
```
tjruwase, line 202: Should assert not just be for `torch.float32`? In other words, we want to retain support for torch.float16.

deepspeed/runtime/zero/stage2.py
```diff
@@ -837,8 +837,8 @@ def gradient_reduction_w_predivide(self, tensor):
 
         tensor_to_allreduce = tensor
 
-        if self.allreduce_always_fp32:
-            tensor_to_allreduce = tensor.float()
+        if self.communication_data_type is not None:
```
tjruwase, line 840: == `torch.float32`

deepspeed/runtime/zero/stage2.py
```diff
@@ -852,7 +852,7 @@ def gradient_reduction_w_predivide(self, tensor):
             tensor_to_allreduce.div_(dp_world_size)
             dist.all_reduce(tensor_to_allreduce, group=self.dp_process_group)
 
-        if self.allreduce_always_fp32 and tensor is not tensor_to_allreduce:
+        if self.communication_data_type is not None and tensor is not tensor_to_allreduce:
```
tjruwase, line 855: `== torch.float32`

deepspeed/runtime/zero/stage2.py
```diff
@@ -1297,17 +1297,21 @@ def set_none_gradients_to_zero(self, i, partition_id):
...
 
         tensor_to_allreduce = tensor
 
         if pg_correctness_test:
-            allreduce_always_fp32 = True
+            communication_data_type = torch.float32
 
-        if allreduce_always_fp32:
-            tensor_to_allreduce = tensor.float()
+        if communication_data_type is not None:
```
tjruwase, line 1313: `== torch.float32`

deepspeed/runtime/zero/stage2.py
```diff
@@ -1318,7 +1322,7 @@ def allreduce_bucket(self, bucket, allreduce_always_fp32=False, rank=None, log=N
             global_rank = _get_global_rank(self.dp_process_group, rank)
             dist.reduce(tensor_to_allreduce, global_rank, group=self.dp_process_group)
 
-        if allreduce_always_fp32 and tensor is not tensor_to_allreduce:
+        if communication_data_type is not None and tensor is not tensor_to_allreduce:
```
tjruwase, line 1325: `== torch.float32`

deepspeed/runtime/zero/stage3.py
```diff
@@ -756,14 +756,14 @@ def __init__(self,
...
-        self.allreduce_always_fp32 = allreduce_always_fp32
+        self.communication_data_type = communication_data_type
         self.gradient_predivide_factor = gradient_predivide_factor
         self.postscale_gradients = postscale_gradients
         self.gradient_accumulation_steps = gradient_accumulation_steps
         self.micro_step_id = INITIAL_MICRO_STEP_ID
 
         if self.reduce_scatter:
-            assert not self.allreduce_always_fp32, "allreduce_always_fp32 is not yet supported with ZeRO-2 with reduce scatter enabled"
+            assert self.communication_data_type is None, "communication_data_type is not yet supported with ZeRO-2 with reduce scatter enabled"
```
tjruwase, line 766: `not torch.float32`

deepspeed/runtime/zero/stage3.py
```diff
@@ -1964,8 +1964,8 @@ def gradient_reduction_w_predivide(self, tensor):
 
         tensor_to_allreduce = tensor
 
-        if self.allreduce_always_fp32:
-            tensor_to_allreduce = tensor.float()
+        if self.communication_data_type is not None:
```
tjruwase, line 1967: `== torch.float32`

deepspeed/runtime/zero/stage3.py
```diff
@@ -1979,7 +1979,7 @@ def gradient_reduction_w_predivide(self, tensor):
             tensor_to_allreduce.div_(dp_world_size)
             dist.all_reduce(tensor_to_allreduce, group=self.dp_process_group)
 
-        if self.allreduce_always_fp32 and tensor is not tensor_to_allreduce:
+        if self.communication_data_type is not None and tensor is not tensor_to_allreduce:
```
tjruwase, line 1982: `== torch.float32`

deepspeed/runtime/zero/stage3.py
```diff
@@ -2304,17 +2304,21 @@ def set_none_gradients_to_zero(self, i, partition_id):
...
 
         tensor_to_allreduce = tensor
 
         if pg_correctness_test:
-            allreduce_always_fp32 = True
+            communication_data_type = torch.float32
 
-        if allreduce_always_fp32:
-            tensor_to_allreduce = tensor.float()
+        if communication_data_type:
```
tjruwase, line 2320: `== torch.float32`

deepspeed/runtime/zero/stage3.py
```diff
@@ -2325,7 +2329,7 @@ def allreduce_bucket(self, bucket, allreduce_always_fp32=False, rank=None, log=N
             global_rank = _get_global_rank(self.dp_process_group, rank)
             dist.reduce(tensor_to_allreduce, global_rank, group=self.dp_process_group)
 
-        if allreduce_always_fp32 and tensor is not tensor_to_allreduce:
+        if communication_data_type is not None and tensor is not tensor_to_allreduce:
```
tjruwase, line 2332: `== torch.float32`

docs/_pages/config-json.md
```diff
@@ -151,11 +151,11 @@ Example of <i>**scheduler**</i>
...
 
-<i>**fp32_allreduce**</i>: [boolean]
+<i>**communication_data_type**</i>: [boolean]
 
-| Description                                                    | Default |
-| -------------------------------------------------------------- | ------- |
-| During gradient averaging perform allreduce with 32 bit values | `false` |
+| Description                                                             | Default |
+| ----------------------------------------------------------------------- | ------- |
+| During gradient averaging perform communication with selected data type |  `none` |
```
tjruwase, line 158: I think it is better to make the default communication data type fp16 and completely drop the none option. What do you think?

deepspeed/runtime/config.py
```diff
@@ -211,8 +211,21 @@ def get_zero_reduce_scatter(param_dict):
...
 
 
-def get_allreduce_always_fp32(param_dict):
-    return get_scalar_param(param_dict, FP32_ALLREDUCE, FP32_ALLREDUCE_DEFAULT)
+def get_communication_data_type(param_dict):
+    val = get_scalar_param(param_dict,
+                           COMMUNICATION_DATA_TYPE,
+                           COMMUNICATION_DATA_TYPE_DEFAULT)
+    val = val.lower()
+    if val == "none":
```
tjruwase, line 219: I can't think of a situation where a none value for communication_data_type is useful. I think we should have fp16 as the default value. Can you share your thoughts here?

Dipet: In this case, `none` means do not change the model's data type. Perhaps we could rename `none` to `default` for more readability. I think it is a bad idea to forcefully set fp16 data type for communication because it is can reduce the quality of the model if we train the model in fp32 mode. I think it's better for the user to set these values himself.

tjruwase: Yes, you are absolutely correct; I was wrong. Using the model data type as the default communication data type is the right thing to do. However, I am unsure that `default` actually improves readability since this option is meant to help users override the default communication data type. So I am wondering if a better approach is for users to only use this option when they want to override since deepspeed can internally figure out the model data type in the default case. In other words the only valid values for this configuration option will be override data types, e.g., fp16, fp32, etc. What do you think?

deepspeed/runtime/engine.py
```diff
@@ -1076,7 +1077,7 @@ def _configure_fp16_optimizer(self, optimizer):
     def _configure_zero_optimizer(self, optimizer):
         zero_stage = self.zero_optimization_stage()
         log_dist('Creating fp16 ZeRO stage {} optimizer'.format(zero_stage), ranks=[0])
-        assert not self.allreduce_always_fp32(), "ZeRO does not support 'fp32_allreduce': true"
+        assert self.communication_data_type is None, "ZeRO supports only 'communication_data_type': 'none'"
```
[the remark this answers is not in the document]
tjruwase: The assert is specifically to reflect that fp32 all-reduce is not supported in ZeRO and is independent of considerations about valid configuration values for `communication_data_type`. In future, when ZeRO starts to support fp32 all-reduce it will much easier to remove this assert without worrying of other side effects. Does that make sense?

deepspeed/runtime/engine.py
tjruwase: Sorry, I was not very clear with my comment, so let me restate it. I think this condition should be 
`if self.condition_data_type == torch.float32`, so that it does not include when `condition_data_type == torch.float16`.

deepspeed/runtime/zero/stage2.py
tjruwase: Got it. So this is the case where we need the override. I would prefer to avoid having self.communication_data_type being None since that really means the model data type. How about the following?
```python
      if self.communication_data_type != tensor.dtype:
          tensor_to_allreduce = tensor.to(self.communication_data_type)
```
Also if self.communication_data_type is always a valid torch data type, then we could even eliminate the if condition and it will be a noop if `self.communication_data_type == tensor.dtype`.

deepspeed/runtime/config.py
Dipet: So your suggestion is to communicate between GPUs with the `float16` as the default data type?
I think this is based on how the project owners see the main purpose of this project. If the main purpose is the training speed that's ok. If the main purpose is speed without reducing the model quality, I think this is not the best option.
In any case, if we decide to use fp16 as the default communication data type, we need to properly describe this in the documentation. This is where I need your help to point me to where I need to add information about these changes to the documentation.
Another problem with these changes is that when users upgrade the package to the version with these changes, their results may differ from their old results.
So if your think using fp16 as a default communication data type is a good idea - I will remove `default` and will only use torch data types as valid arguments.

tjruwase: No, I don't think fp16 should be default communication data type. I want the model training data type (e.g., fp32, fp16, bfloat16) to be the default communication data type, and the configuration option `communication_data_type` should be a way for users to override the communication data type. 

I am sorry that I am not explaining clearly. Please don't make any more changes yet, as you have already done a lot of work. I will send a better explanation when I have a bit more time. Sorry about this.

Dipet: ok, no problem. I think I understand what you mean. Just remove `default` as a valid argument and set `COMMUNICATION_DATA_TYPE_DEFAULT=None`. And when`communication_data_type=None` set it is to a valid torch data type in `engine.py` when model and all arguments will be set.

tjruwase: Yes, we are aligned. Thanks so much for your help.

tjruwase (approved):

deepspeed/runtime/zero/stage2.py
```diff
@@ -200,7 +200,7 @@ def __init__(self,
             assert self.cpu_offload and type(self.optimizer) in [DeepSpeedCPUAdam], f"fp16_master_and_gradients requires optimizer to support keeping fp16 master and gradients while keeping the optimizer states in fp32. Currently only supported using ZeRO-Offload with DeepSpeedCPUAdam. But current setting is ZeRO-Offload:{self.cpu_offload} and optimizer type {type(self.optimizer)}. Either disable fp16_master_weights_and_gradients or enable ZeRO-2 Offload with DeepSpeedCPUAdam"
 
         if self.reduce_scatter:
-            assert not self.allreduce_always_fp32, "allreduce_always_fp32 is not yet supported with ZeRO-2 with reduce scatter enabled"
+            assert self.communication_data_type in (torch.float16, torch.bfloat16), f"ZeRO-2 supports only float16 or bfloat16 communication_data_type. Got: '{self.communication_data_type}'"
```
tjruwase, line 203: Assert message is missing "... with reduce scatter enabled".

tjruwase (approved):

PR #1512, opened 2021-11-02
Tensor-Parallelism general support

RezaYazdaniAminabadi:
This PR provides support for model parallelism during inference without the need for injecting the kernels. 

Add the [PR](https://github.com/microsoft/DeepSpeedExamples/pull/144) in DeepSpeed-Example branch and verify the tensor-parallelism functionality on different model architectures:

 - Bert
 - Roberta
 - GPT2
 - GPT-Neo
 - GPT-J
 - Wav2vec2 
 - T5

cc: @stas00 @hyunwoongko


jeffra (approved):

hyunwoongko:
I will analyze this today and apply it to oslo. thank you.

cc @stas00 @RezaYazdaniAminabadi

deepspeed/module_inject/replace_module.py
```diff
@@ -299,18 +338,126 @@ def transpose(data):
...
+                                      child.weight.shape[1] // mp_size),
+                                     device=child.weight.device,
+                                     dtype=child.weight.dtype)
+            data = mp_replace.copy(new_weight, child.weight.data)
+            new_embedding = nn.Embedding(child.weight.shape[0],
+                                         child.weight.shape[1] // mp_size)
+            new_embedding.weight.data.copy_(data)
+            return new_embedding
+
+        def update_mp_params(child):
```
hyunwoongko, line 399: @RezaYazdaniAminabadi @stas00 This code works well for a few cases, but I don't think it's a good structure to scale to 70 models. Is there any more efficient way?

deepspeed/module_inject/replace_module.py
```diff
@@ -299,18 +338,126 @@ def transpose(data):
             new_module.output_b.data = _4hh_b
         return new_module
 
+    def replace_wo_policy(module, all_reduce_linears):
```
hyunwoongko, line 341: The strategy of this method is usually to apply column slice, and if the name of a specific layer is input, does that layer apply row slice?

deepspeed/module_inject/replace_module.py
```diff
@@ -299,18 +338,126 @@ def transpose(data):
...
+            if hasattr(child, 'embed_dim'):
+                child.embed_dim = child.embed_dim // mp_size
+
+        conv_linear_layer = False
+        if linear_layer_setting is not None:
+            linear_policies = {linear_layer_setting[0]: _replace}
+            if len(linear_layer_setting) == 2:
+                linear_policies.update({linear_layer_setting[1]: _slice_embedding})
+        else:
+            if orig_layer_impl is HFGPT2LayerPolicy._orig_layer_class:
```
hyunwoongko, line 419: Rather than doing this, how about loading that module object and checking that it is a `Conv1D`? In the future, models using Conv1D modules other than GPT2 may exist.

deepspeed/module_inject/replace_module.py
```diff
@@ -299,18 +338,126 @@ def transpose(data):
...
+                try:
+                    import transformers
+                    conv_linear_layer = True
+                    linear_policies = {transformers.model_utils.Conv1D: _replace}
```
hyunwoongko, line 423: Why did you set GPT2 not to slice embeddings?

deepspeed/module_inject/replace_module.py
hyunwoongko: Can't we automate this a bit more? It would be nice to have a strategy that doesn't require parameter names at all.

hyunwoongko: @RezaYazdaniAminabadi You probably thought more when you are making this than me. I'm curious about your opinion.

deepspeed/module_inject/replace_module.py
hyunwoongko: Any special reason?

hyunwoongko: Please tell me if I misunderstood.

deepspeed/module_inject/replace_module.py
hyunwoongko: cc @jaketae

deepspeed/module_inject/replace_module.py
hyunwoongko: I've been thinking briefly, what about profiling strategy? First, forward the module to the profiling stage, and all layers forwarded are considered as columns, and the last forwarded are considered as rows.

hyunwoongko: In general, the row sliced layer must be forwarded later because the row sliced layer takes the output of the column sliced layer. I'm going to re-implement it this way. If you have any other comments, please let me know.
@RezaYazdaniAminabadi @stas00

deepspeed/module_inject/replace_module.py
```diff
@@ -299,18 +338,126 @@ def transpose(data):
...
+                except ImportError:
+                    linear_policies = {nn.Linear: _replace}
+            else:
+                linear_policies = {nn.Linear: _replace, nn.Embedding: _slice_embedding}
```
hyunwoongko, line 427: How do you differentiate between positional embedding and token embedding?

deepspeed/module_inject/replace_module.py
hyunwoongko: It seems more flexible to just use torch.fx than this. Let's automate the whole process of tensor & pipeline parallelization using torch.fx.

hyunwoongko: I succeed implementing new tensor parallel engine.

The following is the new policy object. with torch.fx and tracing (profiling) strategy, we don't need to know the name of parameter !

```python
class T5LayerInfo(LayerInfo):
    @staticmethod
    def base():
        return T5Block

    @staticmethod
    def attention():
        return T5Attention

    @staticmethod
    def mlp():
        return T5DenseReluDense

    @staticmethod
    def reducing_required():
        return ["d_model", "n_heads", "inner_dim"]
```

hyunwoongko: we captured the last forwarded linear or conv 1d layer of the MLP or Attention layer, so we can row parallelize only the last layer, and other layers are column parallelized.

hyunwoongko: @stas00 Is it easy to integrate at this level? I'd actually like to remove of the `reducing_required()` method as well, but I can't think of a way to figure this out at the current level. In DeepSpeed, this is managed inside the engine, but since each model has a different value, it would be a better structure to manage this by model policy.

deepspeed/module_inject/replace_module.py
hyunwoongko: I used `model.get_input_embeddings()`. this is useful in this case.

deepspeed/module_inject/replace_module.py
stas00: I'm missing the full context. Do you suggest to have a policy record for each model like in the example you have shown here:
https://github.com/microsoft/DeepSpeed/pull/1512#discussion_r791247791

I'd help to see several full examples, then it's much easier to see how it can be integrated. 

For example I started integrating Deepspeed-Inference after studying a few examples here:
https://github.com/microsoft/DeepSpeedExamples/pull/144

So I can see what's common, what's unique, which code sections are the driver and need to go into into the Trainer loop.

Monkey-see, monkey-do style is the easiest w/o needing to figure out all the low-level details.

Does it make sense?

hyunwoongko: > I'm missing the full context. Do you suggest to have a policy record for each model like in the example you have shown here:

BERT:

```python
from transformers.models.bert.modeling_bert import BertAttention, BertOutput, BertLayer
from oslo.parallelism.mpu import LayerInfo


class BertLayerInfo(LayerInfo):
    @staticmethod
    def base():
        return BertLayer

    @staticmethod
    def attention():
        return BertAttention

    @staticmethod
    def mlp():
        return BertOutput

    @staticmethod
    def reducing_required():
        return ["all_head_size", "num_attention_heads"]

```

GPT2:

```python
from transformers.models.gpt2.modeling_gpt2 import GPT2Block, GPT2Attention, GPT2MLP

from oslo.parallelism.mpu import LayerInfo


class GPT2LayerInfo(LayerInfo):
    @staticmethod
    def base():
        return GPT2Block

    @staticmethod
    def attention():
        return GPT2Attention

    @staticmethod
    def mlp():
        return GPT2MLP

    @staticmethod
    def reducing_required():
        return ["embed_dim", "split_size", "num_heads"]
```

And you can use engine like this

```python
from oslo import MPU, TensorParallelEngine

mpu = MPU(tensor_parallel_size=4, pipeline_parallel_size=1)
enigne = TensorParallelEngine(model, info_obj, mpu)
engine.parallelize()

# rest code is same.
```

and this engine also supports training.

hyunwoongko: Since the current version code has not been added in OSLO yet, I will write some example code after deployment so that you can easily apply it. Does it make sense?

hyunwoongko: currently, some elements like "all_head_size", "num_attention_heads" are being managed in deepspeed engine now. but I think it's not good structure. so I pushed them towards policy. (because they are model specific parts)

stas00: Yes, please.

As I reported to you originally it didn't appear that different OSLO components can be integrated separately and require all other OSLO components.

So Deepspeed-Inference I can relatively easily integrate into the HF Trainer. We just need to figure out a few MPU quirks. With OSLO I have no idea how to do it, because what I tried didn't work.

But let's not derail this PR and discuss OSLO either on OSLO or HF Transformers Issues. This PR is about Deepspeed-Inference.

hyunwoongko: But as I said before, I wanted it to be mounted as a transformers internal solution instead of `import oslo`. what do you think?

hyunwoongko: + This has nothing to do with deep speed, so let's talk about the transformers issue.

deepspeed/module_inject/replace_module.py
RezaYazdaniAminabadi: That's a good point. Thanks @hyunwoongko

deepspeed/module_inject/replace_module.py
RezaYazdaniAminabadi: That is true that this part needs some refactoring. Please let me know if you have some ideas

RezaYazdaniAminabadi:
> I will analyze this today and apply it to oslo.
> 
> cc @stas00 @RezaYazdaniAminabadi

Thanks a lot @hyunwoongko for the thorough review on this PR. I will use some of your feedback to make this stronger.

deepspeed/module_inject/replace_module.py
hyunwoongko: @RezaYazdaniAminabadi Yes, offline chat would be better. When do you like it?

deepspeed/module_inject/replace_module.py
hyunwoongko: I left a comment on the issue. please note.
https://github.com/huggingface/transformers/issues/13690#issuecomment-1022036232

PR #1519, opened 2021-11-04
modify the quantize.py file for efficiency

yaozhewei:
1. Update the calculate from torch.split to torch.amin/torch.amax for fast computation
2. Update stochastic rounding computation logic (faster and cleaner)
    a. support both sym/asym sr in pytorch level
    b. reduce the new tensor creator from 2-->1
    c. support cpu tensor as well
3. change fp16 --> fp32 to avoid overflow issue
4. change some other logic for easy understanding

### DIFF
--- a/deepspeed/runtime/quantize.py
+++ b/deepspeed/runtime/quantize.py
@@ -33,14 +33,14 @@ class Quantizer(object):
         self.q_groups = q_groups
         self.q_mixed_fp16 = q_mixed_fp16
         self.q_change_ratio = q_change_ratio
-        self.q_type = q_type
+        self.q_type = 1#q_type
         self.qsteps = 0
         self.q_init_period = q_period
         self.quantize_real_ratio = 1.000
         self.q_verbose = q_verbose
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/quantize.py
```diff
@@ -82,41 +82,15 @@ def quantize(self,
...
-            g in zip(add_s,
-                     input_flat)
-        ]
-        input_flat = [((q + a_s) * s).clamp(-(q_range >> 1),
-                                            (q_range >> 1) - 1) / s for q,
-                      a_s,
-                      s in zip(input_flat,
-                               add_s,
-                               scale)]
+        p = input_g.new(input_g.shape).uniform_(-0.5, 0.5)
```
chunyang-wen, line 87: Not a good idea to change the signature of a function at first sight.

rocm-mici:

mrwyattii:
Stale PR. `quantize.py` is quite different now. These changes are no longer relevant, therefore closing the PR.

PR #1521, opened 2021-11-05
Use fstr in launcher

chunyang-wen:
This PR tries to use fstr in `deepspeed/launcher/launch.py`.

There is also something strange here: defining multiple times for the same function.

The only difference is the cmd captured. But only the last registered signal handler will take effect.


tjruwase (approved):

tjruwase:
@chunyang-wen, thanks for this PR. This is great!

PR #1525, opened 2021-11-05
[ZeRO] Default disable elastic ckpt in stage 1+2 and reduce CPU memory overhead during ckpt load

jeffra:
Example scenario that this PR addresses:

10B parameter model using 256 GPUs, each machine has 8 GPUs. We observe that each ZeRO rank checkpoint is 525MB. If elastic checkpointing is enabled we currently require each rank to load all ZeRO checkpoints in order to potentially re-partition the model to a new world size. If elastic checkpointing is disabled we are still loading all ZeRO checkpoints and then we end up throwing most of the state away.

If elastic checkpointing is disabled we currently attempt to load 525MB * 256 files * 8 gpus for a requirement of 1TB of CPU memory per machine just for optimizer state. On certain machines this results in OOM errors as you can imagine.

This PR loads only the required states for each rank. We still require all ZeRO partitions for fp32 master weights so this is loaded but is now enough to reduce the CPU memory requirements per node. For our 10B example this should be more along the lines of 40GB * 8 gpus = 320 GB.

Turns off elastic checkpointing support for ZeRO-2 as default, we've seen issues with this mode and advise users to turn it off.


deepspeed/runtime/engine.py
```diff
@@ -2229,8 +2229,22 @@ def _get_all_zero_checkpoints(self, load_dir, tag):
...
+            # Fully load state for current rank
+            # TODO: remove assumption on zero_sd_list[0] in zero load_state_dict and use rank instead
+            if self.zero_elastic_checkpoint() or i == 0 or dist.get_rank(
+                    group=self.optimizer.dp_process_group) == i:
+                zero_sd_list.append(torch.load(ckpt_name, map_location='cpu'))
+            else:
+                _state = torch.load(ckpt_name, map_location='cpu')
+                # Extract fp32 groups only, otherwise throw away to prevent unnecessary CPU memory overheads
+                _state = {
+                    'optimizer_state_dict': {
```
tjruwase, line 2242: Can we replace the hardcoded constants `optimizer_state_dict` and `single_partition_of_fp32_groups` with variables imported from stage2.py to ease code maintenance?

tjruwase (approved):

tests/unit/simple_model.py
```diff
@@ -161,7 +161,7 @@ def step(self, closure=None):
                 state = self.state[p]
                 if len(state) == 0:
                     state['integer_step'] = 0
-                    state['tensor_step'] = torch.zeros(1)
```
tjruwase, line 164: @jeffra, FYI.

jeffra: Really excellent find, this had been bugging me for a while :)

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2127,24 +2142,36 @@ def load_state_dict(self,
...
             required_version = pkg_version.parse("0.3.17")
-            ckpt_version = state_dict_list[0].get("ds_version", False)
             error_str = f"ZeRO stage 1 changed in {required_version} and is not backwards compatible " \
                 "with older stage 1 checkpoints. If you'd like to load an old ZeRO-1 checkpoint " \
                 "please use an older version of DeepSpeed (<= 0.5.8) and set 'legacy_stage1': true in your zero config json."
+            assert required_version <= ckpt_version, f"Old version: {ckpt_version} {error_str}"
 
-            assert ckpt_version, f"Empty ds_version! {error_str}"
-            assert required_version <= pkg_version.parse(ckpt_version), f"Old version: {ckpt_version} {error_str}"
+        if ckpt_version < pkg_version.parse("0.5.10"):
```
jeffra, line 2165: TODO: remove this block of code and also add logic to detect world size change during ckpt loading.

stas00:
This is interesting, as this PR seems to have fixed a bug where `groups['step']` in apex's `FusedAdam` wasn't getting saved restored. I was just debugging this and happened to sync my master and voila - the bug went away! Thank you, @jeffra

Issue #1298, opened 2021-08-11
Warmup Schedueler is Not Linear warmup, Increase in Log curve, inscrease Too Fast for big model


BitVoyage:
![image](https://user-images.githubusercontent.com/10386094/129052778-948a5ea5-c084-4259-9a70-37a8e9cadc63.png)

BitVoyage:
Would DeepSpeed Support Linear Warmup ? Thank You

tjruwase:
@BitVoyage, thanks for your question. Can you please give more details on what you need? We are currently finishing some changes that will make it easy to integrate existing torch learning rate schedulers into DeepSpeed training.

skpig:
@tjruwase I have encountered the same issue as well. I hope the learning rate can increase linearly in Warmup Stage.

skpig:
@tjruwase @stas00  Is the feature being implemented? If not, I'm willing to give it a try.

tjruwase:
@skpig, please go ahead with your plans. This will be greatly appreciated.

PR #1530, opened 2021-11-06
Add warmup_type arguments in WarmupLR and WarmupDecayLR

skpig:
About issue #1298 @tjruwase 
I add `warmup_type` arguments in WarmupLR and WarmupDecayLR. If `log`  is chosen(Default) , the training curve will be the same as before. If `linear`, then learning rate will increase linearly during warmup.


tjruwase:
Thanks. These looks good. Could you please add some unit tests for this [here](https://github.com/microsoft/DeepSpeed/blob/master/tests/unit/test_lr_schedulers.py)?

skpig:
When I ran `pytest --forked tests/unit/test_lr_schedulers.py`, every test failed with Error reported
`AttributeError: Can't pickle local object 'distributed_test.<locals>.dist_wrap.<locals>.dist_init'`.
I'm using MacOS with no CUDA. And I already install `pytest-forked`. Could you please help me deal with it?

tjruwase:
@skpig, not a problem for now since the CI will also run the tests. I will investigate any failures.

skpig:
Any problems during review? If so, please feel free to at me.

deepspeed/runtime/lr_schedules.py
```diff
@@ -727,6 +736,13 @@ def __init__(self,
         self.max_lrs = self._format_param(self.optimizer, warmup_max_lr, "max_lr")
         self.delta_lrs = [big - small for big, small in zip(self.max_lrs, self.min_lrs)]
         self.warmup_num_steps = max(2, warmup_num_steps)
+        # Currently only support linear and log function
+        if warmup_type not in {'log', 'linear'}:
```
tjruwase, line 740: Can we avoid hardcoded constants 'log' and 'linear' with global vars, such as WARMUP_LOG_RATE and WARMUP_LINEAR_RATE? Or any names that seem reasonable? Thanks.

skpig: @tjruwase Thanks for your suggestion. I've just replaced them with global variables

tjruwase (approved):

PR #1532, opened 2021-11-08
fstr for multnode_runner

chunyang-wen:
This PR tries to:
+ use fstr in `multinode_runner.py`. Last piece in `launch` package.
+ add a reference link for `%n` in `PDSHRunner`.
+ use `raise` instead of `assert` and move the validation early into the constructor of each runner.


jeffra (approved):
Thank you @chunyang-wen!

PR #1540, opened 2021-11-09
Fix return code

Dipet:
Fix returning child processes return code. I always get `return_code=1` without these changes

### DIFF
--- a/deepspeed/launcher/launch.py
+++ b/deepspeed/launcher/launch.py
@@ -134,7 +134,8 @@ def main():
             except Exception:
                 pass
         if last_return_code is not None:
-            raise subprocess.CalledProcessError(returncode=last_return_code, cmd=cmd)
+            logger.error(f"{cmd} exits with return code = {last_return_code}")
+            sys.exit(last_return_code)
         if signum in sig_names:
             logger.info(f"Main process received {sig_names[signum]}, exiting")
         sys.exit(1)


chunyang-wen:
Actually, `sys.exit` and raise are similar. `sys.exit` just raises a `SystemExit` exception. `os._exit(-1)` will force the program to exit directly.

Dipet:
Just simple example:

**test2.py**
```python
import subprocess
raise subprocess.CalledProcessError(returncode=42, cmd="test")
```

**test.py**
```python
import subprocess

pipe = subprocess.Popen(["python", "test2.py"])
pipe.wait()
print(pipe.returncode)
```

Output:
```
  File "test2.py", line 2, in <module>
    raise subprocess.CalledProcessError(returncode=42, cmd="test")
subprocess.CalledProcessError: Command 'test' returned non-zero exit status 42.
1
```

So we see that we always get `return_code = 1`

chunyang-wen:
The two cases both cause the program to stop. The real return code is not the same with the actual return code. But `raise subprocess.CalledProcessError(returncode=42, cmd="test")` provides more information about the command executed and the returncode in log.

So the owners maybe need to claim their design purpose.

Dipet:
Guys, could you take a look at these changes?

tjruwase:
@Dipet thanks for the PR. @Dipet and @chunyang-wen, thanks for the analysis of this issue. 

I agree with @chunyang-wen that the existing code provides valuable context in the stack trace for debugging the crash, in terms of the values of `last_return_code` and `cmd`. This extra context is lost by simply calling `sys.exit(last_return_code)`. @Dipet, can you share the output of using `sys.exit()` in your `test2.py` code? Thanks!

@jeffra, FYI.

Dipet:
Output with `raise subprocess.CalledProcessError(returncode=42, cmd="test")`:
```
  File "test2.py", line 3, in <module>
    raise subprocess.CalledProcessError(returncode=42, cmd="test")
subprocess.CalledProcessError: Command 'test' returned non-zero exit status 42.
1
```

---

Output with `sys.exit(42)`
```
42
```

---

Output with
```python
returncode = 42
try:
    raise subprocess.CalledProcessError(returncode=42, cmd="test")
except subprocess.CalledProcessError:
    traceback.print_exc()
sys.exit(returncode)
```

```
  File "test2.py", line 9, in <module>
    raise subprocess.CalledProcessError(returncode=42, cmd="test")
subprocess.CalledProcessError: Command 'test' returned non-zero exit status 42.
42
```

---

Output with 
```python
returncode = 42
print(str(subprocess.CalledProcessError(returncode=42, cmd="test")), file=sys.stderr)
sys.exit(returncode)
```

```
Command 'test' returned non-zero exit status 42.
42
```

tjruwase:
@Dipet, thanks for sharing these outputs. Like I said earlier, I find the outputs that include the subprocess return code and command more useful for debugging.

@Dipet, @chunyang-wen, @jeffra what do you think?

jeffra:
I agree with @tjruwase 

Perhaps we could go with this option?

```python
returncode = 42
try:
    raise subprocess.CalledProcessError(returncode=42, cmd="test")
except subprocess.CalledProcessError:
    traceback.print_exc()
sys.exit(returncode)
```

```
  File "test2.py", line 9, in <module>
    raise subprocess.CalledProcessError(returncode=42, cmd="test")
subprocess.CalledProcessError: Command 'test' returned non-zero exit status 42.
42
```

chunyang-wen:
> I agree with @tjruwase
> 
> Perhaps we could go with this option?
> 
> ```python
> returncode = 42
> try:
>     raise subprocess.CalledProcessError(returncode=42, cmd="test")
> except subprocess.CalledProcessError:
>     traceback.print_exc()
> sys.exit(returncode)
> ```
> 
> ```
>   File "test2.py", line 9, in <module>
>     raise subprocess.CalledProcessError(returncode=42, cmd="test")
> subprocess.CalledProcessError: Command 'test' returned non-zero exit status 42.
> 42
> ```

There is no need to raise.

```python
returncode = 42
logger.info(f"{cmd} exits with return code = {return_code}")
sys.exit(returncode)
```

deepspeed/launcher/launch.py
```diff
@@ -134,7 +135,11 @@ def sigkill_handler(signum, frame):
             except Exception:
                 pass
         if last_return_code is not None:
-            raise subprocess.CalledProcessError(returncode=last_return_code, cmd=cmd)
+            try:
+                raise subprocess.CalledProcessError(returncode=last_return_code, cmd=cmd)
```
chunyang-wen, line 139: There is no need to raise a known exception and catch it in order to print the message. Just print enough message here. @jeffra what do you think?

jeffra: I agree, printing seems fine instead. Perhaps printing to the error logger if it’s non zero might be nice?

chunyang-wen: Sure. use `logger.error` is better here.

tjruwase (approved):

tjruwase:
@Dipet, thanks so much for contributing this PR.

PR #1544, opened 2021-11-09
Refine quantizer for supporting larger hidden-dim and group size

RezaYazdaniAminabadi:


rocm-mici:

PR #1549, opened 2021-11-11
load_checkpoint backward compatibility

conglongli:
In https://github.com/microsoft/DeepSpeed/pull/1484 a checkpoint key was changed, and it generated a backward compatibility issue when loading old checkpoints. This PR fixes the issue and also helps to match the old value with new key.


tjruwase (approved):

jeffra:
Good catch @conglongli! Maybe we need to create a static checkpoint that is saved in our unit tests that we attempt to load as a unit test. This would help catch backwards compatibility issues going forward.

PR #1560, opened 2021-11-15
Replace brute force and add log

chunyang-wen:
+ Replace a brute force search with a numpy operation
+ Add log about candidate and gpu information

It is a little strange here that `lcm` can be larger than `max_acceptable_batch_size`. By not adding `lcm` to `base_list` when `lcm` is greater than `max_acceptable_batch_size` will cause unit tests to fail.


chunyang-wen:
Any update on this?

jeffra (approved):
Thank you @chunyang-wen!

PR #1562, opened 2021-11-15
Fix partial recovery of sparse_tensor_module_names and dynamically check if gradient data is sparse 

Dipet:
- correctly update `sparse_tensor_module_names` when `load_module_strict=False`
- Do not add `Embedding` weights to `sparse_tensor_module_names` because `sparse` flag is not saved in `state_dict` for module and we can not correctly restore `sparse_tensor_module_names` in that case, because we can not correctly determine sparse flag
- Dynamically check if grad is sparse. Now sparse grads will work not only for `Ebmedding` and `EmbeddingBag`, but for all sparse operations.


tjruwase:
@Dipet, thanks for this PR.

jeffra:
Thank you @Dipet, this looks good to me!

jeffra (approved):

PR #1573, opened 2021-11-18
[CI] transformers@master has been fixed

stas00:
transformers@master should work again, so rolling back to `master` checkout.

@jeffra

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,7 +92,7 @@ jobs:
           if [[ -d ./torch-extensions ]]; then rm -rf ./torch-extensions; fi
           git clone https://github.com/huggingface/transformers
           cd transformers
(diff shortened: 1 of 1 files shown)


jeffra (approved):

jeffra:
Thanks @stas00 for the quick help finding and fixing this.

stas00:
Bowing to your team for having a huge faith to run your CI with our master, which is unheard of!

jeffra:
> Bowing to your team for having a huge faith to run your CI with our master, which is unheard of!

Haha no worries, definitely happy to revert to specific hashes to quickly unblock our side if/when things happen :) Feels like a win-win for us and HF to more quickly detect issues like this. We actually heard from an internal user last night who ran into this OrderedDict issue so am really glad we caught it early!

PR #1577, opened 2021-11-19
Add documentation for TensorBoard logging

manuelciosici:
I could use some help filling in the one-liner for `Train/Samples/elapsed_time_ms_backward_inner` and `Train/Samples/elapsed_time_ms_step`.


jeffra:
Thank you @manuelciosici for this contribution, this is very helpful! We were just discussing that this feature was currently undocumented.

jeffra:
`elapsed_time_ms_step` is the optimizer step time
`elapsed_time_ms_backward_inner` is the backward time that does not include the the gradient reduction time. Only in cases where the gradient reduction is not overlapped, if it is overlapped then the inner time should be about the same as the entire backward time.

jeffra (approved):

manuelciosici:
@jeffra Thanks for the help. I copy-pasted your descriptions into the table. I also added the TensorBoard section to the navigation menu.

PR #1578, opened 2021-11-19
sharded_moe: make top1gating ONNX-exportable

garymm:
Extract code from top1gating into helper functions that are
`torch.jit.script`ed.
Otherwise during a `torch.jit.trace`, the non-Tensor
values used in the calculations get recorded as constants.
`torch.jit.script` coerces them into Tensors and preserves
their dynamic shapes. This enables ONNX export.
We can't script the entire `top1gating` function because it
includes stateful caching logic which is incompatible with ONNX.

Also clean things up:

* Remove unneeded imports.
* Move all imports to top of file.
* Move comments to more relevant locations.
* Consistently use already-imported unqualified type names
  (e.g. Tensor instead of torch.Tensor)
* Fix type annotations.


awan-10 (approved):

garymm:
@awan-10 thanks for reviewing. Could you please merge it?

PR #1588, opened 2021-11-23
allow external control of gradient accumulation boundary

jeffra:
Many integrations are made a lot easier by deferring to the client code for when we are in a gradient accumulation boundary or not. This PR allows the client code to manually set the boundary state and propagates it to the right places in the DS engine.

`self.optimizer` may optionally use the `is_gradient_accumulation_boundary` state, but no harm in passing the state there for non-zero optimizers as well.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1757,6 +1757,26 @@ class DeepSpeedEngine(Module):
             return self._is_gradient_accumulation_boundary
 
     def set_gradient_accumulation_boundary(self, is_boundary):
+        """Manually overrides the DeepSpeed engine's gradient accumulation boundary state, this is an optional
+        feature and should be used with care. The state should be set before to the intended
+        value before each forward/backward. The final fordward/backward should have the
+        boundary state set to True. See example below:
+
+        .. code-block:: python
+
+        engine.set_gradient_accumulation_boundary(False)
+        for _ in range(gradient_accumulation_steps - 1):
+            micro_batch = next(data_loader)
+            loss = engine(micro_batch)
+            engine.backward(loss)
+        engine.set_gradient_accumulation_boundary(True)
+        micro_batch = next(data_loader)
+        loss = engine(micro_batch)
+        engine.backward(loss)
+
+        Arguments:
+            is_boundary (bool): are we at a gradient accumulation boundary or not?
+        """
         self._is_gradient_accumulation_boundary = is_boundary
         self.optimizer.is_gradient_accumulation_boundary = is_boundary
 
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/engine.py
```diff
@@ -1746,11 +1747,18 @@ def is_gradient_accumulation_boundary(self):
...
             bool: if the current step is a gradient accumulation boundary.
         """
-        return (self.micro_steps + 1) % self.gradient_accumulation_steps() == 0
+        if self._is_gradient_accumulation_boundary is None:
+            return (self.micro_steps + 1) % \
+                self.gradient_accumulation_steps() == 0
+        else:
+            return self._is_gradient_accumulation_boundary
+
+    def set_gradient_accumulation_boundary(self, is_boundary):
```
tjruwase, line 1759: I think a doc string is needed here to help clients use properly. In particular, about ensuring that clients clear this flag after optimizer.step().

tjruwase: An alternative approach that comes to mind is instead of adding a new engine function, we can add a `is_gradient_accumulation_boundary` argument to `backward()` and `step()` such as: 
```python
   def backward(self, ..., is_gradient_accumulation_boundary=False):
         self._is_gradient_accumulation_boundary = is_gradient_accumulation_boundary
        ...

  def step(self, ..., is_gradient_accumulation_boundary=False):
        self._is_gradient_accumulation_boundary = is_gradient_accumulation_boundary
       ...
```
I am not sure if this will work well with the current kwargs constraints of `backward` and `step`.

deepspeed/runtime/engine.py
```diff
@@ -1757,6 +1757,26 @@ def is_gradient_accumulation_boundary(self):
...
+
+        engine.set_gradient_accumulation_boundary(False)
+        for _ in range(gradient_accumulation_steps - 1):
+            micro_batch = next(data_loader)
+            loss = engine(micro_batch)
+            engine.backward(loss)
+        engine.set_gradient_accumulation_boundary(True)
+        micro_batch = next(data_loader)
+        loss = engine(micro_batch)
+        engine.backward(loss)
```
tjruwase, line 1775: For completeness, should this snippet not include `engine.step()`?

jeffra: Oh good call, that's an important bit, in this case we only need to call `engine.step()` once at the end.

tjruwase: Also, would the following be correct usage of this api?
```python
   for i, micro_batch in enumerate(next(data_loader)):
        is_gradient_accumulation_boundary = (i+1 % gradient_accumulation_steps) == 0
        engine.set_gradient_accumulation_boundary(is_gradient_accumumation_boundary)
        loss = engine(micro_batch)
        engine.backward(loss)
        engine.step()

```

jeffra: Yep, that would work too. In the case of megatron and fairseq they both call their optimizer step in completely different code paths then the main fwd/bwd passes so it's a bit more aligned with the style in the docstring (without a final engine.step) I guess.

tjruwase (approved):

PR #1591, opened 2021-11-24
Add tensor methods in flops counting and separate macs and flops

cli99:
This PR adds support to count flops for the following operations:

1. PyTorch tensor methods such as torch.matmul, torch.mm, torch.addmm, torch.bmm, etc, and the corresponding torch.Tensor.xxx methods
2. other activations, e.g.  sigmoid, silu, and gelu
3. other norms

This PR also separates the MACs and flops counting. multiply-add operations are counted as 2 flops while others are counted as 1. FLOPS is computed using flops.

### DIFF
--- a/deepspeed/profiling/flops_profiler/README.md
+++ b/deepspeed/profiling/flops_profiler/README.md
@@ -1,6 +1,6 @@
 # DeepSpeed Flops Profiler
 
-> Measures the parameters, latency, and floating point operations of your model.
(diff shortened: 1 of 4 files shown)


jeffra (approved):

BitCalSaul:
`x @ y` is still not counted as flops. I added each line in the code of Transformer, and found the attention operation was not counted as flops.

PR #1592, opened 2021-11-26
Tweaks for lm-eval-harness

zphang:
- Importing one fix from Microsoft/DeepSpeed
- Disable `sanity_check`, which does not seem to be doing the right thing. (For some reason it checks against every possible mergeable weight key, for every layer)


rocm-mici:

mrwyattii:
This PR is quite old and it appears outdated. Looks like the relevant changes occurred here https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/212

PR #1596, opened 2021-11-29
wall_clock_breakdown disable failed in moe layer

PeiqiWang:
wall_clock_breakdown disable failed in moe layer due to incorrectly used a function as property in https://github.com/microsoft/DeepSpeed/blob/d14baad9401bc1de8c982c1b3590d91108c4bbf1/deepspeed/runtime/engine.py#L1017


jeffra:
Thank you @PeiqiWang for your contribution! It seems this combo is missed in our unit tests.

jeffra (approved):

PR #1607, opened 2021-12-02
Refactor ZeRO naming to reduce confusion

jeffra:
* Rename `stage2.py` to `stage_1_and_2.py` to make it clear that both stage 1 and 2 are implemented here
* Remove `FP16_*` prefixes on ZeRO class names since they now support multiple dtypes
* Update BF16 + ZeRO documentation
* Removes deprecated zero stage 1 code (hasn't been used since v0.3.17)


deepspeed/runtime/engine.py
```diff
@@ -1326,28 +1325,12 @@ def _configure_zero_optimizer(self, optimizer):
...
-                max_elements_per_comm=self.zero_reduce_bucket_size(),
-                dp_process_group=self.data_parallel_group,
-                elastic_checkpoint=self.zero_elastic_checkpoint(),
-                mpu=self.mpu,
-                postscale_gradients=self.postscale_gradients(),
-                gradient_predivide_factor=self.gradient_predivide_factor(),
-                gradient_predivide=self.gradient_predivide,
+        if self.zero_legacy_stage1():
+            raise Exception(
+                "If you must use the deprecated version of zero stage 1 please revert an an older version of deepspeed (<= 0.5.8)."
```
manuelciosici, line 1330: Can I suggest "The deprecated zero stage 1 is not supported in deepspeed >= 0.5.8. Downgrade to a version less than 0.5.8 if you need to the deprecated  zero stage 1."?

tjruwase (approved):

PR #1622, opened 2021-12-08
Control ds_report output

### LINKED ISSUE #1541: [BUG] async-io-related warnings at start time when nvme is not configured to be used
`async_io` warnings shouldn't be emitted when nvme is not in the ds config file.

Deepspeed under multi-node is so noisy, thousands of lines dumped to the log file, making troubleshooting the log file on crash extremely difficult. **Why does each of the 1024 processes in the ensemble need to print the same startup info, each being hundreds of lines long**?

But this report is about `async-io` which emits warnings when it shouldn't. So it's a general issue and not directly related to not log on global rank 0 only. But it then gets exacerbated by printing it 3 * 1024 times :(

```
[...]
 [WARNING]  async_io: please install the libaio-devel package with yum
```

the config I'm using isn't using nvme/`async-io`. 

these warnings shouldn't be printed at all when nvme is not in the ds config file.

This manifests at least when using Megatron-Deepspeed.

Thank you!

@tjruwase

Here is the config I'm using:

```
{
  "train_micro_batch_size_per_gpu": $MICRO_BATCH_SIZE,
  "train_batch_size": $GLOBAL_BATCH_SIZE,
  "gradient_clipping": 1.0,
  "zero_optimization": {
    "stage": 1
  },
  "fp16": {
    "enabled": true,
    "loss_scale": 0,
    "loss_scale_window": 500,
    "hysteresis": 2,
    "min_loss_scale": 1,
    "initial_scale_power": 12
  },
  "steps_per_print": 2000,
  "wall_clock_breakdown": false
}
```

tjruwase:
Add two flags to help control ds_report out:

1. --hide_operator_status
2. --hide_errors_and_warnings

Separate cli and function entry points into ds_report

Should fix #1541.

### DIFF
--- a/bin/ds_report
+++ b/bin/ds_report
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-from deepspeed.env_report import main
+from deepspeed.env_report import cli_main
 
 if __name__ == '__main__':
-    main()
+    cli_main()
--- a/deepspeed/env_report.py
+++ b/deepspeed/env_report.py
@@ -106,21 +106,17 @@ def parse_arguments():
     parser.add_argument(
         '--hide_operator_status',
         action='store_true',
-        help='Suppress display of installation and compatiblity statuses of DeepSpeed operators. '
+        help=
+        'Suppress display of installation and compatiblity statuses of DeepSpeed operators. '
     )
-    parser.add_argument(
-        '--hide_errors_and_warnings',
-        action='store_true',
-        help='Suppress warning and error messages.'
-    )
+    parser.add_argument('--hide_errors_and_warnings',
+                        action='store_true',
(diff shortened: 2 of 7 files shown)


tjruwase:
@stas00, FYI

jeffra:
@stas00 on the Megatron-DeepSpeed side (once this PR is merged) we'll need to pass the appropriate flags (`hide_operator_status` and `hide_errors_and_warnings`) here: https://github.com/microsoft/Megatron-DeepSpeed/blob/acc3e9b5fab400912107c33fbcc9e0d02809dd6e/pretrain_gpt.py#L209

jeffra (approved):

stas00:
Perfect, thank you gentleman!

PR #1634, opened 2021-12-11
Support CPU Adam and Adagrad on Windows

lostmsu:
Ensures, that CPU Adagrad and Adam can be build with MSVC++

Unit tests for them appear to pass on my machine, but I am hoping to add a CI build in the future too to ensure they don't break.


lostmsu:
@jeffra @RezaYazdaniAminabadi the CI tests seem to be passing. Is there anything I need to do for this PR to be merged?

csrc/sparse_attention/utils.cpp
```diff
@@ -94,14 +94,14 @@ ret_t sdd_segment(torch::Tensor layout, int start_width)
     // block index
     torch::Tensor idx = torch::zeros_like(layout);
     int current = 0;
-    size_t H = layout.size(0);
-    size_t M = layout.size(1);
-    size_t N = layout.size(2);
+    int64_t H = layout.size(0);
+    int64_t M = layout.size(1);
+    int64_t N = layout.size(2);
```
jeffra, line 99: Was this change required for CPU Adam+Adagrad support on windows?

jeffra: Oh I see, you're right here, this is fixing a bug I think. 

https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#_CPPv4NK2at6Tensor4sizeEN2at7DimnameE

jeffra (approved):

csrc/sparse_attention/utils.cpp
lostmsu: I, honestly, forgot what the compile error was, but torch's C++ [size](https://pytorch.org/cppdocs/api/function_namespaceat_1a9dd1a76ff523b31c90a5b3c9c7a3dab8.html?highlight=size) explicitly returns `int64_t`.

RezaYazdaniAminabadi: Can we use uint64_t instead, because if I remember correctly size_t is unsigned long int?

RezaYazdaniAminabadi (approved):

Issue #1635, opened 2021-12-13
[BUG] CPU adam too slow even with AVX512


nobellant215:
**Describe the bug**
Optimizer step is taking unexpectedly long. I am running a GPT-like model with 1.35B parameters with ZeRO-3, optimizer CPU offload enabled.

I have added extra timers to measure exact timing for optimizer step:

```
 time (ms) | update_optims: 21766.25 | unscale_grads: 506.45 | optimizer_step: 22850.66
 time (ms) | forward_microstep: 376.60 | backward_microstep: 864.56 | backward_inner_microstep: 806.86 | backward_allreduce_microstep: 57.62 | step_microstep: 22880.87
time (ms) | forward: 376.60 | backward: 864.56 | backward_inner: 806.87 | backward_allreduce: 57.62 | step: 22880.87
```
I believe update computation using cpu-adam should be much faster - less than 5 seconds but here it's taking more than 22 seconds ...

**System info**
 - OS: Ubuntu 18.04
 - One RTX3090 and  Xeon Silver 4210R(which has 1 AVX512 unit)
 - Python 3.9, Pytorch 1.7.1+cu110

Below is my deepspeed configuration
```
{
    "fp16": {
        "enabled": true,
        "loss_scale": 10,
        "loss_scale_window": 1000,
        "initial_scale_power": 16,
        "hysteresis": 2,
        "min_loss_scale": 1
    },

    "optimizer": {
        "type": "AdamW",
        "params": {
            "lr": 5e-05,
            "betas": [0.9, 0.999],
            "eps": 1e-08,
            "weight_decay": 0.0
        }
    },

    "scheduler": {
        "type": "WarmupLR",
        "params": {
            "warmup_min_lr": 0,
            "warmup_max_lr": 5e-05,
            "warmup_num_steps": 0
        }
    },

    "zero_optimization": {
        "stage": 3,
        "offload_optimizer": {
            "device": "cpu",
            "pin_memory": true
        },
        "offload_parameter": {
            "device": "cpu",
            "pin_memory": true
        },
        "overlap_comm": true,
        "contiguous_gradients": "auto",
        "sub_group_size": 1e9,
        "stage3_prefetch_bucket_size": "auto",
        "stage3_max_live_parameters": 1e9,
        "stage3_max_reuse_distance": 1e9,
        "stage3_gather_fp16_weights_on_model_save": true
    },
    "gradient_clipping": "auto",
    "steps_per_print": 2000,
[...]

PR #1637, opened 2021-12-14
Fix cpu-adam AVX performance

RezaYazdaniAminabadi:
This fixes https://github.com/microsoft/DeepSpeed/issues/1635


deepspeed/ops/transformer/inference/transformer_inference.py
```diff
@@ -146,6 +146,31 @@ def _transpose_for_context(x):
                                       (hidden_size_per_partition,)
             return x.view(*new_x_layer_shape)
 
+        def backup_attention(mixed_query, key_layer, value_layer, input_mask):
```
jeffra, line 149: @RezaYazdaniAminabadi, was this an accidental addition? seems unrelated to the PR right?

jeffra (approved):

deepspeed/ops/transformer/inference/transformer_inference.py
RezaYazdaniAminabadi: Thanks Jeff, will remove it!

PR #1657, opened 2021-12-22
remove force-multi and fix None val check in base tuner in autotuning

cli99:
This PR fixes the following bugs in autotuning:
1. the autotuning scheduler does not add force-multi to the experiment script and uses what users pass in
2. the initial best metric value could be None in the tuning and needs to be checked


deepspeed/autotuning/scheduler.py
```diff
@@ -333,7 +333,6 @@ def run_experiment(exp: dict, reservations, user_script, user_args):
     include_str = include_str[:-1]
     master_port = exp["master_port"]
     exp["launcher_args"] = [
-        "--force_multi",
```
tjruwase, line 336: What is the downside of this option?

jeffra: @cli99 and I were talking about this before the holidays. If I recall correctly, she was seeing an issue where if password-less ssh isn't setup correctly than `--force_multi` and autotuning in general will fail. Even on our ITP nodes we saw an issue where if the destination host was set to `localhost` this didn't match the ssh alias (i.e., worker-N) and hung with a password prompt. We do need `--force_multi` if we're doing auto-tuning across multiple nodes though so not sure what the best default is here? Can auto-tuning detect if we're trying to tune across multiple hosts and use `--force_multi` in that case only?

cli99: We ask the users to specify --force_multi in the training script if it's needed. The autotuner does not add it explicitly but passes what the train script uses.

jeffra: If a single candidate job requires > 1 node then yes we don't need `--force_multi`. However, if a single candidate job requires only 1 node and we have 2+ nodes available in the scheduling pool then we will require `--force_multi` for it to launch the candidate job on a remote node. We can sync offline, @tjruwase and I are discussing this now as well :)

cli99: In that case, due to line https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/launcher/runner.py#L350, multi_node_exec is also set. We don't need --force_multi explicitly?

rocm-mici:

loadams:
These changes were already applied in other PRs, closing this one.

PR #1690, opened 2022-01-12
Add additional arguments to launch routine

liamcli:
This PR adds additional arguments to the launcher to support more flexible calls to user script.  Specifically, arguments are added to support calling a module, calling without `python`, and calling without the `local_rank` argument.  There are also small formatting improvements as well as export of some additional info in `launch.py`.


jeffra (approved):
Looks good, thanks for the contribution @liamcli :)

PR #1717, opened 2022-01-22
Add codespell to pre-commit checks

aphedges:
Back in October 2021, I created #1423 to fix a large number of spelling errors. I've noticed some spelling errors in commits since then, but doing what I did on my first pass is way too time-intensive to do regularly. However, I found out about [codespell](https://github.com/codespell-project/codespell), which is used in some other open source projects. Instead of being a strict check against a dictionary, it uses rules to fix common mistakes. While it likely does not catch every error that a more complete spellcheck would, it produces much fewer false positives. Running it on the repository (which takes under a second) catches some errors I missed back in my previous pass, as well as some spelling errors that were added more recently. Adding it to the checks will make spelling errors more difficult to sneak in.

I put the configuration directly into `.pre-commit-config.yaml` for ease of development, but I notice that clang-format and yapf both have separate config files. codespell supports both `setup.cfg` and `.codespellrc` files for configuration. Should I put it in one of them instead? If so, any preference for one file over the other?

I explicitly excluded three files that have been automatically generated and not meant for human consumption anyway. In addition, I explicitly excluded the word `unsupport` from the check because it is used in 12 poorly worded error messages that can probably have their wording improved to be more grammatical. I don't think that change is important enough to make it into this PR, though.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,3 +32,16 @@ repos:
     hooks:
     -   id: clang-format  # formatter of C/C++ code based on a style guide: LLVM, Google, Chromium, Mozilla, and WebKit available
         args: []
+
+-   repo: https://github.com/codespell-project/codespell
+    rev: v2.1.0
+    hooks:
+    -   id: codespell
+        args: [
+            # Do not check files that are automatically generated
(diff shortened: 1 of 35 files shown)


tjruwase (approved):

jeffra:
Thank you @aphedges, this is super cool!

PR #1718, opened 2022-01-23
DataLoader Length Fix

Sanger2000:
DeepSpeedDataLoader currently has a length equal to the length of the dataset passed in. I made a small change such that its length is equal to the number of batches in the data loader (the same way it is done in pytorch).

An alternative change could be to set the length equal to `len(self.data)` after `self._create_dataloader()` is called.


Sanger2000:
It's a pretty minor change, but let me know if it makes sense.

jeffra (approved):
LGTM. Thank you for the contribution!

Issue #1674, opened 2022-01-05
[REQUEST] Lightning ecosystem CI


Borda:
Hello and Happy New Year 2022! :tada: 
Just wondering if you already heard about quite the new **Pytorch Lightning (PL) ecosystem CI** where we would like to invite you to... You can check out our blog post about it: [Stay Ahead of Breaking Changes with the New Lightning Ecosystem CI](https://devblog.pytorchlightning.ai/stay-ahead-of-breaking-changes-with-the-new-lightning-ecosystem-ci-b7e1cf78a6c7) :zap: 
We already have joined integration between PL and your framework, so we would like to level it up. At this moment, we are running tests with your latest version, but it may accidentally happen that your next version will be incompatible with our next release version... :confused:  But here we have a solution - ecosystem CI with testing both - your and our latest development head we can find it very early and prevent releasing eventually bad version... :+1: 

**What is needed to do?**
- have some tests, including PL integration
- add config to ecosystem CI - https://github.com/PyTorchLightning/ecosystem-ci

**What will you get?**
- scheduled nightly testing configured development/stable heads
- slack notification if something went wrong
- testing also on multi-GPU machine

SeanNaren:
This would be awesome :)

for more information, this will help ensure that DeepSpeed doesn't break functionality for users in Lightning. To do this however, there needs to be tests in DeepSpeed that rely on Lightning (which currently isn't the case).

@tjruwase @jeffra you can assign this to me; I think I can assist here! I know for a long time you've been requesting adding Lightning based tests to the CI, so definitely is overdue :)

jeffra:
Thank you @SeanNaren! This would be really wonderful to have for both lightning and deepspeed :)

PR #1726, opened 2022-01-26
Add a very simple PyTorch Lightning test!

SeanNaren:
Related #1674.

As discussed offline with @jeffra would be neat to have some simple Lightning tests to ensure we remain compatible together :)

This test is extremely simple, testing a toy model/dataset using the Lightning DeepSpeed plugin which wraps the DeepSpeed library. It's not a huge test as I didn't want to clog up CI time or start adding extensive testing before we get something rolling.

Let me know if I've done the right thing here cc @jeffra @tjruwase @borda


Borda (approved):
lgtm :rabbit:

jeffra (approved):
Excellent, thank you @SeanNaren! This is a great start in adding coverage between our two projects.

Issue #1710, opened 2022-01-19
[BUG] No matching distribution found for triton==1.0.0


stas00:
**Describe the bug**

**To Reproduce**

```
$ git clone https://github.com/microsoft/DeepSpeed
$ cd DeepSpeed
$ DS_BUILD_CPU_ADAM=1 DS_BUILD_UTILS=1 pip install -e . --global-option="build_ext" --global-option="-j8" --no-cache -v --disable-pip-version-check
[...]
ERROR: Could not find a version that satisfies the requirement triton==1.0.0 (from deepspeed) (from versions: 0.1, 0.1.1, 0.1.2, 0.1.3, 0.2.0, 0.2.1, 0.2.2, 0.2.3, 0.3.0)
ERROR: No matching distribution found for triton==1.0.0
```

Once I manually do:

```
pip install triton==1.0.0
```

it then works.

Not sure why the manual build doesn't take care of figuring out the requirements.

@jeffra

aphedges:
@stas00, I actually came across this problem myself earlier this month but for a different package. I can describe the full problem if you want, but I think the problem I encountered has a slightly different cause.

Looking at [`triton`](https://pypi.org/project/triton/), I can see that the most recent version to have a source distribution is `0.3.0`, which is also the most recent version that pip found for you. From what I can tell, pip does not appear to show versions that are unavailable for an specific platform. For example:

```console
# On x86_64 macOS 11:
$ pip index versions triton
triton (0.3.0)
Available versions: 0.3.0, 0.2.3, 0.2.2, 0.2.1, 0.2.0, 0.1.3, 0.1.2, 0.1.1, 0.1
# On x86_64 CentOS 7:
$ pip index versions triton
triton (1.1.1)
Available versions: 1.1.1, 1.1.0, 1.0.0, 0.4.2, 0.4.1, 0.3.0, 0.2.3, 0.2.2, 0.2.1, 0.2.0, 0.1.3, 0.1.2, 0.1.1, 0.1
  INSTALLED: 1.0.0
  LATEST:    1.1.1
```

Versions of Triton past `0.3.0` only have `manylinux2014` x86_64 wheels on PyPI. I'm guessing you're experiencing these errors on a different platform than that?

stas00:
pip finally added `index` - yay!

I think it's something else here, since I do see 1.0.0 on JeanZay:

```
pip index versions triton
WARNING: pip index is currently an experimental command. It may be removed/changed in a future release without prior warning.
[...]

PR #1727, opened 2022-01-26
launcher save pid + require manual triton install for sparse-attn

jeffra:
When launching deepspeed processes programmatically it's useful to save the top-level launcher's process id (pid). This allows one method for cleaning up runner processes since if we kill the top-level launcher process than the children processes will be killed as well (on that worker).

Also removing auto-install of `triton` due to several issues with the triton 1.0.0 install we've seen (e.g., #1710).

### DIFF
--- a/deepspeed/launcher/runner.py
+++ b/deepspeed/launcher/runner.py
@@ -441,7 +441,7 @@ def main(args=None):
     if args.save_pid:
         main_pid = os.getpid()
         launcher_pid = result.pid
-        pid_file = os.path.join(PID_FILE_BASEPATH, f"{main_pid}.ds")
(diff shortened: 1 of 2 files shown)


deepspeed/launcher/runner.py
```diff
@@ -428,8 +436,22 @@ def main(args=None):
 
     logger.info(f"cmd = {' '.join(cmd)}")
     result = subprocess.Popen(cmd, env=env)
+
+    pid_file = None
+    if args.save_pid:
+        main_pid = os.getpid()
+        launcher_pid = result.pid
+        pid_file = os.path.join(PID_FILE_BASEPATH, f"{main_pid}.ds")
```
tjruwase, line 444: Make deepspeed more prominent in the name, e.g. _main_deepspeed_pid.ds

tjruwase (approved):

PR #1732, opened 2022-01-29
Move param_shapes to model files

tjruwase:
Move param_shapes out of zero_* checkpoint files into mp_rank_* files. 
Symbolic constants to replace hard-coded. 
Add deepspeed.checkpoint

@stas00, FYI


stas00:
So the main question here - is do we want to reserve the backward compatibility? Since a checkpoint saved before this PR was merged will not work with the new `zero_to_fp32.py` code.

Since all the newly saved checkpoints will be OK, we surely could ask the user to downgrade deepspeed to an older version to recover the weights. since it'd be a one time thing it probably won't be too much of an annoyance. 

That's said let's start saving which deepspeed version the checkpoint was generated with, so that in a pinch a user can quickly find the matching version.

So  `zero_to_fp32.py`  could start with a preamble:
```
Parsing checkpoint created by deepspeed==0.5.7
```

And then it's loud and clear to the user which version is needed to recover the file.

How does that sound?

Of course the super friendly way is to handle the back-compat way explicitly and check both files for `param_shapes` or use the right file depending on the deepspeed version.

tjruwase:
> But I think we should still log the version the checkpoint was created with.

Done!

jeffra (approved):

PR #1747, opened 2022-02-07
Improve how runner parses env var file

liamcli:
Only split by the first "=" when parsing lines from the deepspeed environment file.


tjruwase (approved):

tjruwase:
@liamcli, thanks so much.

PR #1755, opened 2022-02-09
Fixing the inference build-path when pre-building the op

RezaYazdaniAminabadi:


stas00:
Thank you for fixing this build issue, @RezaYazdaniAminabadi 

I validated that it solves the problem.

tjruwase (approved):

stas00:
You can probably merge this anyway, as I don't think any of the tests exercise this path. If they did they would have failed when the original incorrect code was committed.

And at some point some tests would be helpful...

PR #1775, opened 2022-02-15
Modify unit test to cover more cases

RezaYazdaniAminabadi:
This PR adds two set of modification to the unit tests for better test coverage on optimizer functionality:
 1. Increase the parameters groups to catch errors better at optimizer step function.
 2. Reduce the loss scale by using "`initial_scale_power`" to reduce the likelihood of overflowing in the backward and the optimizer step will be called into eventually. 

cc: @tjruwase, @jeffra


tjruwase (approved):

rocm-mici:

PR #1778, opened 2022-02-16
Add checkpoint_comm_enabled flag

jeffra:
Optional disable communication collectives inside load/save checkpoint via a new top-level json parameter:
`checkpoint_comm_enabled: [true|false]`

This option is not tested with zero-3 or moe, it might work okay but wanted to be on the safe side adding asserts in these cases. We can test these paths if/when someone needs this option with those features.

New zero json parameter: `move_params_to_cpu_during_init: [true|false]`. Setting this to false will skip the `move_to_cpu(self.bit16_groups[i])` call inside zero 1/2 initialization. In the context of Big Science this might help them to reduce cpu memory usage when CPU memory is smaller than GPU memory per node.

/cc @stas00


rocm-mici:

PR #1790, opened 2022-02-24
01 adam optimizer

EugeneLYC:
Maximizing Communication Efficiency for Large-scale Training via 0/1 Adam

Author: @EugeneLYC, @conglongli, @minjiaz, Christopher De Sa, Yuxiong He
Paper: https://arxiv.org/abs/2202.06009


deepspeed/runtime/fp16/onebit/zoadam.py
```diff
@@ -0,0 +1,352 @@
...
+            1-bit Adam. The var_freeze_step is usually the end of the learning rate warmup
+            and thus does not require tuning. (default: 100000)
+        var_update_scaler (int, optional): The interval to update the variance. Note that
+            the update policy for variance follows an exponential rule, where var_update_scaler
+            denotes the kappa in the 0/1 Adam paper. (default: 16)
+        local_step_scaler (int, optional): The interval to scale the local steps interval
+            according to the learning rate policy. (default: 32678)
+        local_step_clipper (int, optional): The largest interval for local steps with
+            learning rate policy. This corresponds to the variable H in the 0/1 Adam paper.
+            (default: 16)
```
conglongli, line 39: For these 4 hyperparameters for 0/1 Adam, when writing the tutorial please make sure to include a more detailed description (can refer the sections or equations in paper) and a tuning strategy.

deepspeed/runtime/fp16/onebit/zoadam.py
```diff
@@ -0,0 +1,352 @@
...
+                            self.comm_backend_handle.compressed_allreduce(
+                                comm_buffer,
+                                state['worker_error'],
+                                state['server_error'],
+                                self.deepspeed.local_rank))
+                    if 'exp_avg_mask' in group:
+                        if comm_buffer.device != group['exp_avg_mask'].device:
+                            group['exp_avg_mask'] = group['exp_avg_mask'].to(
+                                device=comm_buffer.device)
+                        comm_buffer.mul_(group['exp_avg_mask'])
```
conglongli, line 248: This mask need to be applied after every compressed_allreduce that is related to the gradient (because 1-bit compression cannot handle the exact zeros). So (1) Here you can add another indentation to only apply mask when self.size >1. (2) We also need to apply the mask after the two compressed_allreduce at line 225 and 265.

conglongli (changes requested):
Thanks Yucheng for the great work! I left a few comments that should be easy to fix. I think this PR should be ready to merge after the fix and adding the doc/tutorial.

docs/_tutorials/zero-one-adam.md
```diff
@@ -0,0 +1,131 @@
...
+
+`var_update_scaler` is the interval to update the variance. Note that the update policy for variance follows an exponential rule, and `var_update_scaler` denotes the $\kappa$ factor (please refer to the [0/1 Adam paper](https://arxiv.org/abs/2202.06009) for detailed expression). In practice, we found its default value (16) is able to work well on most of the tasks, including BERT-Base/Large pretraining, GPT pretraining, and ImageNet training.
+
+`local_step_scaler` and `local_step_clipper` are two hyperparameters for learning rate based local step policy in 0/1 Adam. Informally, `local_step_scaler` denotes the number of steps 0/1 Adam takes before increasing its local step interval. `local_step_clipper` is the largest local step interval, i.e., the maximal number of communication rounds 0/1 Adam can skip before the next synchronization. The learning rate policy is the default policy used in 0/1 Adam. We can also trivially construct other policies by setting these two hyperparameters such as constant local step interval policy by setting `local_step_scaler=1` and `local_step_clipper=constant`.
```
conglongli, line 104: I believe `local_step_scaler` denotes the number of steps 0/1 Adam takes before multiplying its local step interval by 2? If so then it might be better to explicitly write so instead of "increasing its local step interval" which is a bit vague. Similary, we might want to describe explicitly how we increase the variance update interval based on var_update_scaler (I believe it's also multiplying by 2?).

docs/_tutorials/zero-one-adam.md
```diff
@@ -0,0 +1,131 @@
...
+
+### 2.1 Running Pre-training with DeepSpeed and 0/1 Adam
+
+We provide example scripts under [DeepSpeedExamples/bing_bert/01_adam/](https://github.com/microsoft/DeepSpeedExamples/tree/master/bing_bert/01_adam). There are 3 sets of scripts corresponding to NCCL-based implementation, MPI-based implementation on Ethernet systems, and MPI-based implementation on InfiniBand systems. For MPI-based implementation, we provide both example scripts when launching with deepspeed or mpirun.
+
+### 2.2 Configuration for BERT Pre-training with DeepSpeed and 0/1 Adam enabled
+
+The `deepspeed_bsz4k_01adam_config_seq128_*.json` and `deepspeed_bsz4k_01adam_config_seq512_*.json` files give the user the ability to specify DeepSpeed
+options in terms of batch size, micro batch size, optimizer, learning rate, and other parameters. In these files we include the tuned hyperparameters to reproduce experiments in our [paper](https://arxiv.org/abs/2202.06009).
+
```
conglongli, line 128: It'd be helpful to add a 2.3 GLUE fine-tuning section here, just a few sentences to point out which script to use since in DSE PR we did add the GLUE-related code.

deepspeed/runtime/fp16/onebit/zoadam.py
```diff
@@ -228,6 +228,11 @@ def step(self, closure=None, grads=None):
                                         state['server_error'],
                                         self.deepspeed.local_rank)
                                     exp_avg.mul_(beta1).add_(1 - beta1, grad_onebit)
+                                    if 'exp_avg_mask' in group:
+                                        if grad_onebit.device != group['exp_avg_mask'].device:
+                                            group['exp_avg_mask'] = group['exp_avg_mask'].to(
+                                                device=grad_onebit.device)
+                                        grad_onebit.mul_(group['exp_avg_mask'])
```
conglongli, line 235: I think this actually needs to be added before line 230 "exp_avg.mul_(beta1).add_(1 - beta1, grad_onebit)"

conglongli (changes requested):
Thanks Yucheng for applying the requested changes and adding the doc/tutorial. I reviewed and left a few minor comments.

conglongli (approved):
LGTM, thank you Yucheng!

PR #1798, opened 2022-02-28
Refactor MoE and Groups API to simplify model creation and mangement

awan-10:
This PR makes API breaking changes to the groups creation as well as MoE layer creation. 

The tutorial has been updated to reflect the changes. DeprecatedException will be raised when users try to use the old API.

The corresponding client-side changes will be coming in PRs on the Megatron-DeepSpeed side (https://github.com/microsoft/Megatron-DeepSpeed/pull/36) as well as on the CIFAR example side (https://github.com/microsoft/DeepSpeedExamples/pull/163).


jeffra (approved):
Looks great, thanks guys for all the hard work here.

PR #1801, opened 2022-03-01
bf16+pipeline parallelism

tjruwase:
bf16_optimizer implementing optimizer state sharding (a.k.a., zero stage 1)
Integration with pipeline parallelism


tjruwase:
@stas00, FYI

deepspeed/runtime/engine.py
```diff
@@ -981,6 +969,9 @@ def _configure_distributed_model(self, model):
                     hasattr(param,
                             'ds_id') for param in self.module.parameters()):
                 self.__check_params(self.module, torch.bfloat16)
+            if self.zero_optimization_stage() == 0 and not self.pipeline_parallelism:
+                raise NotImplementedError(
+                    "BF16 support is not yet implemented when not running ZeRO")
```
jeffra, line 974: Let's update this error to be more clear on what's supported/no-supported (talked offline)

jeffra:
I've added a separate issue (https://github.com/microsoft/DeepSpeed/issues/1835) to discuss the addition of a `gradient_accumulation_dtype` config parameter to help clear up confusion about modes. This is a summary based on a previous offline discussion between myself, @tjruwase, and @stas00.

ShadenSmith (approved):

deepspeed/runtime/utils.py
```diff
@@ -867,3 +871,149 @@ def get_only_unique_item(items):
...
+    if global_grad_norm is None:
+        global_grad_norm = get_grad_norm(parameters, mpu=mpu)
+    clip_coef = max_norm / (global_grad_norm + eps)
+    if clip_coef < 1:
+        for p in parameters:
+            p.grad.detach().mul_(clip_coef)
+    return global_grad_norm
+
+
+def get_global_norm_of_tensors(input_tensors, norm_type=2, mpu=None):
```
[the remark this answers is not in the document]
tjruwase: Created #1839.

deepspeed/runtime/engine.py
```diff
@@ -1605,6 +1616,9 @@ def print_forward_breakdown(self, fwd_time):
 
     @instrument_w_nvtx
     def allreduce_gradients(self, bucket_size=MEMORY_OPT_ALLREDUCE_SIZE):
```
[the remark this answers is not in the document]
tjruwase: Agree 1000X

deepspeed/runtime/engine.py
```diff
@@ -1789,7 +1807,7 @@ def _take_model_step(self, lr_kwargs, block_eigenvalue={}):
         if (not self.zero_optimization() and not self.fp16_enabled()
                 and not self.amp_enabled()):
             self.zero_grad()
-        else:
+        elif not self.bfloat16_enabled():
```
thomasw21, line 1810: Do we not zero grad when bf16 is enabled?

stas00: bf16 manages its own grads because remember it accumulates them in a special fp32 buffer which gets cleared after step, see:

https://github.com/microsoft/DeepSpeed/pull/1801/files#diff-e6635a81d2c2bf0938b5f83b1c4945e0f344e0106191a6960df8ee4aa64cc55fR1810

stas00: but I'm actually not sure if the branch above is correct, since it'd be true for bf16 I think - let me check in debugger.

stas00: it totally goes into line 1809 as I suspected.

@tjruwase, is that by design - the `elif` on 1810 will never get executed under bf16.

stas00: we rewrote that logic as it was somewhat broken, but wasn't really a bug.

stas00: https://github.com/microsoft/DeepSpeed/pull/1801/commits/144f652768a980b7a4f1e1104ae7872189742c99

deepspeed/runtime/bf16_optimizer.py
```diff
@@ -0,0 +1,582 @@
...
+        for i, (bf16_partitions, fp32_partition) in enumerate(zip(self.bf16_partitioned_groups, self.fp32_groups_flat_partition)):
+            partition_id = dist.get_rank(group=self.real_dp_process_group[i])
+            bf16_partitions[partition_id].data.copy_(fp32_partition.data)
+
+    def clear_hp_grads(self):
+        for flat_gradients in self.fp32_groups_gradients_flat:
+            flat_gradients.zero_()
+
+        for group in self.fp32_groups_has_gradients:
+            group = [False] * len(group)
```
thomasw21, line 436: Are you doing an in-place operation? Don't you have to do something like:
```python
for i in range(len(self.fp32_groups_has_gradients)):
    self.fp32_groups_has_gradients[i] = [False] * len(self.fp32_groups_has_gradients)
```

deepspeed/runtime/bf16_optimizer.py
```diff
@@ -0,0 +1,582 @@
...
+
+        see_memory_usage('end bf16_optimizer', force=True)
+
+    def _link_all_hp_params(self):
+        dp_world_size = dist.get_world_size(group=self.dp_process_group)
+        for i, param_group in enumerate(self.optimizer.param_groups):
+            # Link bf16 and fp32 params in partition
+            # TODO: Make this configurable
+            partition_id = dist.get_rank(group=self.real_dp_process_group[i])
+            partition_size = self.bf16_groups_flat[i].numel() // dp_world_size
```
thomasw21, line 221: is `self.bf16_groups_flat[i].numel()` guaranteed to be a multiple of dp_world_size?

tjruwase: Yes, because of [this](https://github.com/microsoft/DeepSpeed/blob/a9bfaee9c3b951f5a964ef69e82cfa984027fbb4/deepspeed/runtime/bf16_optimizer.py#L136-L139).

deepspeed/runtime/bf16_optimizer.py
tjruwase: Good catch! Now fixed like this
```
        for i, group in enumerate(self.fp32_groups_gradients):
            self.fp32_groups_has_gradients[i] = [False] * len(group)
```

deepspeed/runtime/engine.py
```diff
@@ -981,6 +969,10 @@ def _configure_distributed_model(self, model):
                     hasattr(param,
                             'ds_id') for param in self.module.parameters()):
                 self.__check_params(self.module, torch.bfloat16)
+            if self.zero_optimization_stage() == 0 and not self.pipeline_parallelism:
+                raise NotImplementedError(
+                    "When not running ZeRO, BF16 training support is only supported for Pipeline parallelism"
```
kisseternity, line 974: Hello, I wonder why BF16 is only supported for Pipeline parallelism or ZeRO 1 to ZeRO3, since there is not such limit in prior version.

tjruwase: @kisseternity, apologies for the confusion here. This is a new `bf16+Pipeline parallelism` [code path](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/bf16_optimizer.py) that was written in the last minute for [BLOOM ](https://huggingface.co/blog/bloom-megatron-deepspeed#bf16optimizer) model training. The existing restrictions in combining with ZeRO are temporary. We plan to harmonize these combinations and eliminate the confusions.

kisseternity: > @kisseternity, apologies for the confusion here. This is a new `bf16+Pipeline parallelism` [code path](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/bf16_optimizer.py) that was written in the last minute for [BLOOM ](https://huggingface.co/blog/bloom-megatron-deepspeed#bf16optimizer) model training. The existing restrictions in combining with ZeRO are temporary. We plan to harmonize these combinations and eliminate the confusions.

Thanks for replying. In that case, I think bf16 can be used without the  bf16 optimizer or ZeRO as before.

tjruwase: Yes, these changes do not affect the previous support for `bf16+ZeRO`.

PR #1822, opened 2022-03-10
fix file ordering

szhengac:
`get_optim_files` in `zero_to_fp32.py` does not sort files correctly: 1, 10, 2, 3, 4, 5, 6, 7, 8, 9, leading to the wrong merged pytorch checkpoint. Add natural ordering to correct it.


deepspeed/utils/zero_to_fp32.py
```diff
@@ -52,7 +61,7 @@ def get_model_state_file(checkpoint_dir, zero_stage):
 
 def get_optim_files(checkpoint_dir):
     # XXX: need to test that this simple glob rule works for multi-node setup too
-    optim_files = sorted(glob.glob(os.path.join(checkpoint_dir, "*_optim_states.pt")))
+    optim_files = sorted(glob.glob(os.path.join(checkpoint_dir, "*_optim_states.pt")), key=natural_keys)
```
tjruwase, line 64: Can you please share unit test for this case?

jeffra: Unfortunately, our CI does not currently support multi-node testing :(

Maybe a unit test that calls `get_optim_files` directly and passes a `tmpdir` path where some empty files were created with the `*_optim_states.pt` naming style.

szhengac: Makes sense to me. I will prepare one.

szhengac: @jeffra @tjruwase unittest is added

szhengac:
@tjruwase any update on this PR?

jeffra (approved):
Thank you @szhengac! Looks good to me.

Issue #966, opened 2021-04-16
ZeRO 0, 1, 2, 3 produce different results


szhengac:
Previously I observed that ZeRO 1 produces significant worse performance than ZeRO 2 in finetuning. Similar observation is also mentioned in https://github.com/microsoft/DeepSpeed/issues/757. I created a simple test to see how loss changes with different ZeRO stages. The test code snippet is pasted as follows:

```
import os
import json
import argparse
import torch
import deepspeed
from torch import nn
from torch.utils.data.distributed import DistributedSampler


class SimpleModel(torch.nn.Module):
    def __init__(self, hidden_dim, empty_grad=False, zero=0):
        super(SimpleModel, self).__init__()
        self.linear = torch.nn.Linear(hidden_dim, hidden_dim)
        mlp = [self.linear]
        mlp.append(torch.nn.Linear(hidden_dim, hidden_dim//2))
        for _ in range(6):
            l = torch.nn.Linear(hidden_dim//2, hidden_dim//2)
            mlp.append(l)
        mlp.append(torch.nn.Linear(hidden_dim//2, hidden_dim))
        l = torch.nn.Linear(hidden_dim, hidden_dim)
        l.weight = self.linear.weight
        l.bias = self.linear.bias
        mlp.append(l)
        if zero == 3:
            deepspeed.zero.register_external_parameter(self, self.linear.weight)
            deepspeed.zero.register_external_parameter(self, self.linear.bias)
        self.mlp = nn.Sequential(*mlp)
        if empty_grad:
            self.layers2 = torch.nn.ModuleList([torch.nn.Linear(hidden_dim, hidden_dim)])
        self.cross_entropy_loss = torch.nn.CrossEntropyLoss()

    def forward(self, x, y):
        hidden_dim = x
        hidden_dim = self.mlp(hidden_dim)
        return self.cross_entropy_loss(hidden_dim, y)


def create_config_from_dict(tmpdir, config_dict):
    config_path = os.path.join(tmpdir, 'temp_config.json')
    with open(config_path, 'w') as fd:
        json.dump(config_dict, fd)
    return config_path


def get_data_loader(model, total_samples, hidden_dim, device):
    batch_size = model.train_micro_batch_size_per_gpu()
    train_data = torch.randn(total_samples, hidden_dim, device=device, dtype=torch.half)
    train_label = torch.empty(total_samples,
                              dtype=torch.long,
                              device=device).random_(hidden_dim)
[...]

PR #1823, opened 2022-03-10
fix step in adam

szhengac:
deepspeed zero performs optimization step on each sub-group independently, but adam implementation uses a single step for the whole param group so different sub-groups will use different steps to compute exponential weight for first-order and second-order momentum. This pr fixes this issue.


deepspeed/ops/adam/fused_adam.py
```diff
@@ -131,6 +124,8 @@ def step(self,
                 state = self.state[p]
                 # State initialization
                 if len(state) == 0:
+                    # DeepSpeed processes each subgroup a time, so we need to keep tracking step for each tensor separately
+                    state['step'] = 0
```
tjruwase, line 128: This is only true for zero stage 3, right?

szhengac: yes, i think it only affects zero 3. zero 1 and 2 seems to only call optimization step on each param_group once.

tjruwase (approved):

tjruwase:
@szhengac, thanks!

tjruwase:
@szhengac, apologies I am reverting my approval because my colleagues just made me aware of some issues that require a bit more time to understand. Primarily, this fix changes the behavior of non zero stage 3 logic and will break existing checkpoints since it changes checkpoint state. Ideally, since only zero stage 3 exhibits a problem then the fix should probably be in stage 3 rather than in shared fused adam code. Apologies for the confusion.

tjruwase:
@szhengac, could you please provide a unit test that demonstrates the original zero stage 3 problem? Thanks!

tjruwase:
@szhengac, how about making these two changes for backwards compatibility for existing stage < 3 checkpoints.
1. Restore the group level step
2. Initialize param level step from group level if one exists, rather than initializing to zero

I will annotate the PR with these suggestions.

tjruwase:
@szhengac, how about making these two changes for backwards compatibility for existing stage < 3 checkpoints.
1. Restore the group level step
2. Initialize param level step from group level if one exists, rather than initializing to zero

I will annotate the PR with these suggestions.

deepspeed/ops/adam/fused_adam.py
```diff
@@ -109,13 +109,6 @@ def step(self,
             bias_correction = 1 if group['bias_correction'] else 0
             beta1, beta2 = group['betas']
 
-            # assume same step across group now to simplify things
```
tjruwase, line 112: Restore this for backwards compatibility of existing stage < 3 checkpoints

tjruwase, line 112: Restore this for backwards compatibility of existing stage < 3 checkpoints

deepspeed/ops/adam/fused_adam.py
tjruwase, line 128: state['step'] = group.get('step', 0)

tjruwase, line 128: state['step'] = group.get('step', 0)

szhengac: This looks like a good idea. I will also put a more detailed comment to explain it otherwise other readers may get confused.

szhengac:
@tjruwase any update on this PR?

tjruwase:
> @tjruwase any update on this PR?

This looks good to me. Is it possible for you to do some e2e comparing loss curves before and after this change for zero < 3?

tjruwase:
> @tjruwase Sure, will do it when I have capacity for the experiments. I have done some e2e finetuning with zero 3 after the fix, it improves accuracy by 6-7% when there are two param groups.

That is exciting news about the zero3 improvement :).

tjruwase (approved):

raghavthind2005 (approved):

tjruwase:
@szhengac, just checking if you had time to push this? Thanks!

szhengac:
@tjruwase the conflict has been resolved.

szhengac:
> @szhengac, thanks for resolving the conflict. Did you get a chance to measure convergence impact for zero < 3? Thanks!

I haven't got extra machine capacity to test it.

szhengac:
@tjruwase I have done some loss testing and I can obtain exactly the same loss curve.

szhengac:
But at the same time, I also realized zero 1 and 2 give different loss trajectories regardless of this PR. Similar issue was reported before: https://github.com/microsoft/DeepSpeed/issues/966

tjruwase:
> But at the same time, I also realized zero 1 and 2 give different loss trajectories regardless of this PR. Similar issue was reported before: #966

I thought for that issue we had resolved the [loss divergence](https://github.com/microsoft/DeepSpeed/issues/966#issuecomment-829516471) across the zero stages, right?

szhengac:
I thought so. But the following script gives me different results:
```
import os
import json
import argparse
import torch
import deepspeed
from torch import nn
from apex.normalization import FusedLayerNorm as BertLayerNorm
from torch.utils.data.distributed import DistributedSampler


class SimpleModel(torch.nn.Module):
    def __init__(self, hidden_dim, empty_grad=False, zero=0):
        super(SimpleModel, self).__init__()
        linear = torch.nn.Linear(hidden_dim, hidden_dim)
        FinalLayerNorm = BertLayerNorm(hidden_dim, eps=1e-12)
        mlp = [linear, FinalLayerNorm]
        mlp.append(torch.nn.Linear(hidden_dim, hidden_dim//2))
        for _ in range(6):
            l = torch.nn.Linear(hidden_dim//2, hidden_dim//2)
            mlp.append(l)
        mlp.append(torch.nn.Linear(hidden_dim//2, hidden_dim))
        l = torch.nn.Linear(hidden_dim, hidden_dim)
        l.weight = linear.weight
        l.bias = linear.bias
        mlp.append(l)
        if zero == 3:
            deepspeed.zero.register_external_parameter(self, linear.weight)
            deepspeed.zero.register_external_parameter(self, linear.bias)
        self.mlp = nn.Sequential(*mlp)
        if empty_grad:
            self.layers2 = torch.nn.ModuleList([torch.nn.Linear(hidden_dim, hidden_dim)])

    def forward(self, x, y):
        hidden_dim = x
        hidden_dim = self.mlp(hidden_dim)
        return self.cross_entropy_loss(hidden_dim, y)


def create_config_from_dict(tmpdir, config_dict):
    config_path = os.path.join(tmpdir, 'temp_config.json')
    with open(config_path, 'w') as fd:
        json.dump(config_dict, fd)
    return config_path

def get_data_loader(model, total_samples, hidden_dim, device, dtype, local_rank):
    batch_size = model.train_micro_batch_size_per_gpu()
    train_label = torch.empty(total_samples,
                              dtype=torch.long,
                              device=device).random_(hidden_dim)
    sampler = DistributedSampler(train_dataset)
    train_loader = torch.utils.data.DataLoader(train_dataset,
                                               batch_size=batch_size,
                                               sampler=sampler)
    return train_loader


def get_args(tmpdir, config_dict):
    parser.add_argument("--local_rank", type=int, default=0)
    parser.add_argument('--zero', type=int, default=0)
    args = parser.parse_args()  #args=''

    config_dict["zero_optimization"]["stage"] = args.zero
    print('config_dict["zero_optimization"]', config_dict["zero_optimization"])
    config_path = create_config_from_dict(tmpdir, config_dict)

    args.deepspeed_config = config_path
    return args


def print0(msg):
        print(msg, flush=True)


rank = int(os.environ['RANK'])
print('seed:', 2222 + rank)
torch.random.manual_seed(2222 + rank)

config_dict = {
    "train_batch_size": 16*2*8,
    "train_micro_batch_size_per_gpu": 16,
    "steps_per_print": 1,
    "optimizer": {
        "type": "Adam",
        "params": {
            "lr": 0.01,
            "weight_decay": 0.01,
            "bias_correction": True,
            "eps": 1e-6
        }
    },
    "gradient_clipping": 0.1,
    "fp16": {
        "enabled": False,
        "initial_scale_power": 10
    },
    "bfloat16": {
        "enabled": True
    },
    "zero_optimization": {
        "stage": 1,
        "overlap_comm": False,
        "contiguous_gradients": False,
        "reduce_bucket_size": 20,
        "comm_force_reduce_scatter": True,
        "comm_allgather_stream_sync": False,
        "stage3_param_persistence_threshold" : 1e6,
        "comm_small_bucket_size": 5e8,
        "stage3_param_persistence_threshold": 1e6,
        "stage3_max_reuse_distance" : 1e6,
        "stage3_max_live_parameters": 1e6,
        "stage3_prefetch_bucket_size": 5e8
    }
}

args = get_args('/tmp/', config_dict)
hidden_dim = 1027


#with deepspeed.zero.Init(config=config_dict, enabled=True):
model = SimpleModel(hidden_dim, empty_grad=False, zero=args.zero)

model, _, _,_ = deepspeed.initialize(args=args,
                                     model=model,
                                     dist_init_required=True)

data_loader = get_data_loader(model=model,
                              total_samples=50000,
                              hidden_dim=hidden_dim,
                              device=model.device,
                              dtype=torch.bfloat16,
                              local_rank=args.local_rank)


def print_params(tag, model):
        for n, p in model.named_parameters():
            print0("{} {}:{}".format(tag, n, p))

#print_params('pre-train', model)
for n, batch in enumerate(data_loader):
    #if torch.distributed.get_rank() == 0 and model.is_gradient_accumulation_boundary():
    model.backward(loss)
    model.step()
    if torch.distributed.get_rank() == 0 and model.is_gradient_accumulation_boundary():
        print("{}, LOSS: {}".format(n, loss.item()))
    #print_params('step={}'.format(n), model)
    if n == 32: break

tjruwase:
> I thought so. But the following script gives me different results:

Got it. This means we have a regression in DeepSpeed and is independent of this PR. Could you please create a new issue with your test case? Thanks so much for sharing this test case, in addition to this PR. The PR is good to merge.

szhengac:
@tjruwase I have opened an issue: https://github.com/microsoft/DeepSpeed/issues/1945

PR #1828, opened 2022-03-11
Split github action unit tests by platform/GPU

mrwyattii:
This PR makes it so `unit-tests` github action jobs can be run independently on the different platforms. This is useful for when a single platform fails the unit tests (e.g., because the nodes are down or the environment is horked) and you wish to only rerun on that platform.


jeffra (approved):
looks good to me, once formatting is fixed let's merge

PR #1836, opened 2022-03-16
[build] support cuda-11.6

stas00:
to support the new cuda during build.

Thank you.

@jeffra 

the failing amd ci job seems to be totally unrelated, you're not running cuda-11.6


jeffra (approved):
Thanks Stas, yep our amd unit tests are having an unrelated issue at the moment.

PR #1842, opened 2022-03-16
THH header fix

rraminen:
Filing this PR to avoid 'THH/THHGeneral.h' file not found error error during DeepSpeed build.

Used rocm/pytorch:rocm4.5.2_ubuntu18.04_py3.8_pytorch_1.9.0 (with an old PyTorch vesion) for testing DeepSpeed. There were no issues.

This issue occurs as THC/THCGeneral.h is deprecated. Reference: https://github.com/pytorch/pytorch/pull/66391/files

cc: @jithunnair-amd

### DIFF
--- a/csrc/lamb/fused_lamb_cuda_kernel.cu
+++ b/csrc/lamb/fused_lamb_cuda_kernel.cu
@@ -8,7 +8,6 @@
 #include "ATen/cuda/CUDAContext.h"
 #include "ATen/cuda/detail/IndexUtils.cuh"
 //#include "ATen/Type.h"
-#include <THC/THCGeneral.h>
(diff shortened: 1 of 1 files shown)


jithunnair-amd:
@jeffra For a bit more background, the `THC/THCGeneral.h` file was available till PyTorch version 1.10, but is not available in PyTorch version 1.11. But we tested it with PyTorch 1.9, just to make sure this change is backward-compatible.

jeffra (approved):
Excellent, thank you two! I'll approve/merge now but also curious for @RezaYazdaniAminabadi to take a look here. It seems our torch 1.2 tests also passed so looks good to me.

Issue #1795, opened 2022-02-26
[BUG] IndexError / Runtime Error with `torch.nn.TransformerEncoder`


floatshadow:
**Describe the bug**

**To Reproduce**
I'm fairly new to DeepSpeed, and I tried deepspeed zero-3 and offload with  pytorch transformer.
And I write a simple test code:
```python
import os
import torch
import deepspeed
import argparse

import torch.nn as nn
from torch.utils.data import DataLoader, Dataset
from torch.utils.data.distributed import DistributedSampler
from torch.nn.parallel import DistributedDataParallel

from apex.optimizers import FusedAdam
from deepspeed.ops.adam import DeepSpeedCPUAdam
from deepspeed.runtime.utils import see_memory_usage

parser = argparse.ArgumentParser()
parser.add_argument("--local_rank", type=int, default=0)
parser.add_argument("--world_size", default=2)
parser.add_argument("--init_method", default='tcp://localhost:12306')
parser.add_argument("--master_addr", default="127.0.0.1")
parser.add_argument("--master_port", default="12346")
deepspeed.add_config_arguments(parser)
args = parser.parse_args()
args.rank = int(os.getenv("RANK", "0"))
args.init_method = f"tcp://{args.master_addr}:{args.master_port}"

print(f"initializing distributed...", flush=True)
torch.distributed.init_process_group(backend='nccl', 
                           init_method=args.init_method, 
                           rank=args.local_rank, 
                           world_size=args.world_size)
#deepspeed.init_distributed(distributed_port=args.master_port, init_method=args.init_method)
args.rank = torch.distributed.get_rank()
if torch.distributed.is_initialized():
  print(f"Worker Rank {args.rank} initialized.")


class BoringDataset(Dataset):
  def __init__(self) -> None:
      super().__init__()
      self.data = torch.randn(20, 128, 1024)
  
  def __len__(self):
    return len(self.data)
  
  def __getitem__(self, index):
      return self.data[index, :, :], self.data[index, :, :]


train_set = BoringDataset()
train_sampler = DistributedSampler(train_set)
train_loader = DataLoader(train_set, batch_size=2, sampler=train_sampler)


see_memory_usage(f"Before Building Model", force=True)
encoder_layer = nn.TransformerEncoderLayer(
    d_model=1024,
    nhead=8) 
with deepspeed.zero.Init(remote_device="cpu", config_dict_or_path=args.deepspeed_config, enabled=True):
[...]

PR #1849, opened 2022-03-21
Fix for ZeRO stage 3 prefetch queue

mrwyattii:
Fix to bug found while addressing https://github.com/microsoft/DeepSpeed/issues/1795


deepspeed/runtime/zero/stage3.py
```diff
@@ -350,7 +350,7 @@ def fetch_sub_module(self, current_submodule: Module) -> None:
                     params_to_fetch))
             while self.__param_queue and len(discarded_from_prefetch_queue) < len(
                     params_not_already_fetched):
-                param_in_trace = self.__param_queue.popleft()
```
tjruwase, line 353: Are you seeing some failure with this code? Please take a look at this [description](https://github.com/microsoft/DeepSpeed/blob/0eb2c763b47344ecdacb88713a710dc2753da39f/deepspeed/runtime/zero/stage3.py#L217-L219).

mrwyattii: Thanks @tjruwase - yes I am seeing a bug introduced in #1453 that triggers this [Runtime Error](https://github.com/microsoft/DeepSpeed/blob/0eb2c763b47344ecdacb88713a710dc2753da39f/deepspeed/runtime/zero/stage3.py#L358) for some models

PR #1851, opened 2022-03-22
Import ABC from collections.abc for Python 3.10 compatibility.

tirkarthi:
Import ABC directly from `collections` was deprecated and removed in Python 3.10. Import from `collections.abc` for Python 3.10 compatibility.


tjruwase:
@tirkarthi, thanks for this PR. Can you  clarify that this is compatible with python-3.* < 3.10?

jeffra (approved):
Ah I see a deprecation warning in python 3.8.8 saying: `DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working`.

Good catch @tirkarthi, thank you.

jeffra:
@tirkarthi, if you want feel free to add python 3.10 to our setup.py here as well. I think it's good for us to start understanding places where we might break with 3.10.
https://github.com/microsoft/DeepSpeed/blob/b61d7199c42523b8359374b07cedd81396878f4b/setup.py#L285-L290

PR #1857, opened 2022-03-23
[zero-3] add bwd support for list/dict types returned in fwd

jeffra:
Currently if a forward pass returns a `list` or `dict` zero-3 will skip the backward pass on any Tensors that are embedded in the structures.

- [x] add support for list return types
- [x] add a unit test for common iterable data types (e.g., tuple, list, dict)
- [x] add support for `dict` return types
- [x] add support for objects that inherit from dict/list/tuple, e.g., https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_outputs.py
- [x] fix/test hf+ds regression


deepspeed/runtime/zero/stage3.py
```diff
@@ -87,20 +87,38 @@ def iter_params(module: Module, recurse=False) -> Iterable[Parameter]:
...
+                                                    backward_function,
+                                                    output)
+            touched_outputs[key] = touched_output
+        return touched_outputs
     elif type(outputs) is torch.Tensor:
         return functional.apply(module, backward_function, outputs)
     else:
+        if not is_builtin_type(outputs):
+            logger.warning(
+                f"A module is returning an unknown type ({type(outputs)}) at forward, backward will be skipped for any tensors embedded in this structure."
```
samyam, line 120: Maybe change the warning to: "A module has unknown inputs or outputs type ({type(output)}) and the tensors embedded in it cannot be detected. The Z3 hooks designed to trigger before or after backward pass of the module relies on knowing the input and output tensors and therefore may not get triggered properly."

deepspeed/runtime/zero/stage3.py
```diff
@@ -87,20 +87,38 @@ def iter_params(module: Module, recurse=False) -> Iterable[Parameter]:
...
         touched_outputs = []
         for output in outputs:
             touched_output = _apply_to_tensors_only(module,
                                                     functional,
                                                     backward_function,
                                                     output)
             touched_outputs.append(touched_output)
-        return tuple(touched_outputs)
+        return tuple(touched_outputs) if isinstance(outputs, tuple) else touched_outputs
+    elif isinstance(outputs, dict):
```
samyam, line 106: Oh cool, good call to add the dictionary.

samyam (approved):
Looks good. Just left a few minor comments.

tjruwase (approved):

PR #1880, opened 2022-04-05
Bug Fixes for autotuner and flops profiler

cli99:
This PR makes the following change:
1. fix the autotuner timing bug when backward is not called
2. change autotuner default to not tune zero stage if "zero_optimization" is not defined
3. fix flops profiler bug when computing flops for [torch.nn.functional.interpolate](https://pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html)


rocm-mici:

tjruwase (approved):

PR #1885, opened 2022-04-08
Bug fix for flops profilers output

VisionTheta:
When flops profiler is enabled, it will print lots of numpy `einsum_path` results in the standard output, which will whelm profilers output and waste much time on printing, especially when profiling large models.

This PR makes the following change:

1. directly remove the print line.


VisionTheta:
Hi @conglongli @cli99, could you please have a look at this PR? It's my first and simple contribution is DeepSpeed but it will help others to use profiler effectively.

cli99:
Hi @[VisionTheta](https://github.com/VisionTheta), thanks for catching this.

cli99 (approved):

PR #1907, opened 2022-04-21
Fix launcher for reading env vars

szhengac:
The system does not recognize the env setting with `f"export {key}={quote(val)}; "` I tested in on AWS cloud, and EFA is not properly setup with the latest DeepSpeed. I roll back to the previous version where `exports += "export {}={}; ".format(key, val)` was used and EFA can be recognized.

### DIFF
--- a/deepspeed/launcher/multinode_runner.py
+++ b/deepspeed/launcher/multinode_runner.py
@@ -70,7 +70,7 @@ class PDSHRunner(MultiNodeRunner):
 
         exports = ""
         for key, val in self.exports.items():
-            exports += f"export {key}={quote(val)}; "
+            exports += "export {}={}; ".format(key, val)
 
         # https://linux.die.net/man/1/pdsh
         # %n will be replaced by pdsh command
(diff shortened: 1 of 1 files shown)


tjruwase:
@szhengac, I don't understand why this should be an issue. Is it a function of python version?

szhengac:
@tjruwase, for example, if `key=LD_LIBRARY_PATH` and `val=/opt/amazon/efa/lib64:/opt/aws-ofi-nccl/lib:$LD_LIBRARY_PATH`, `f"export {key}={quote(val)}; "` gives us `"LD_LIBRARY_PATH='/opt/amazon/efa/lib64:/opt/aws-ofi-nccl/lib:$LD_LIBRARY_PATH';"` while `exports += "export {}={}; ".format(key, val)` gives us `'LD_LIBRARY_PATH=/opt/amazon/efa/lib64:/opt/aws-ofi-nccl/lib:$LD_LIBRARY_PATH;'`. The former one does not work on our system.

tjruwase:
Got it. Thanks for the explanation. I don't fully understand the nuances of single and double quotes in python. But the change looks fine.

tjruwase (approved):

PR #1908, opened 2022-04-22
Fix DeepspeedDataLoader length

lkm2835:
```len(DeepspeedDataLoader)``` is currently equal to ```len(data_sampler) = len(dataset) / data_parallel_world_size```.

But, ```len(DataLoader)``` is equal to ```len(dataloader) =  len(data_sampler) / batch_size```.


rocm-mici:

PR #1914, opened 2022-04-26
Fix: Sparse tensors not updating

Dipet:
`param.grad.data` returns new object, for this reason we do not assign updated sparse gradients to the target tensor and the optimizer works with unique gradient tensor on each device.


tjruwase:
@Dipet, thanks for sharing this context. I have not worked a lot with sparse tensors, so that is probably why I have not run into this issue. My suggestion would be to implement this logic only for the sparse tensor code path, so that the dense tensor code paths remain unchanged for backwards-compatibility. I am happy to help brain-storm on this. 

FYI @jeffra, who has worked more on sparse tensors.

Dipet:
@jeffra Tests failed, but it looks like problem with hardware, because tests failed on `nvcc` call.

Dipet:
@jeffra Can you look at the PR?

Dipet:
Guys, could you take a look at this PR?

tjruwase:
@Dipet, apologies for the delay. I had suggested restricting this change only to sparse tensors so that dense tensor code paths remain backwards compatible. Is there a problem with doing that?

Dipet:
@tjruwase I changed the logic for sparse tensors only, dense tensors should work the same as before.

tjruwase (approved):
Thanks!

PR #1915, opened 2022-04-26
Fairseq support

jeffra:
This is the first step towards deepspeed supporting fairseq. In order to use deepspeed w. fairseq we still require some changes on the fairseq trainer side. We are working on open-sourcing these changes as well so users can use deepspeed w. fairseq. Keep an eye out for updates on this line if you are interested.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -85,6 +85,12 @@ except ImportError:
     APEX_INSTALLED = False
     pass
 
+try:
+    from fairseq.optim.fairseq_optimizer import FairseqOptimizer
+    FAIRSEQ_INSTALLED = True
+except ImportError:
+    FAIRSEQ_INSTALLED = False
+
 
 def split_half_float_double_sparse(tensors):
     supported_types = [
@@ -415,6 +421,23 @@ class DeepSpeedEngine(Module):
             float: norm
         """
         return self._global_grad_norm
+
+    def __getattr__(self, name):
+        """
+        Pass through attributes defined in the model if they are not overridden by ds-engine.
+        """
+        _module = {}
+        if "module" in self.__dict__:
+            _module = self.__dict__['module']
+
+        if name in dir(self):
+            return getattr(self, name)
+        elif name in dir(_module):
+            return getattr(_module, name)
+        else:
+            raise AttributeError("'{}' object has no attribute '{}'".format(
+                type(self).__name__,
+                name))
 
     def checkpoint_tag_validation_enabled(self):
         return self._config.checkpoint_tag_validation_enabled
@@ -893,9 +916,18 @@ class DeepSpeedEngine(Module):
                            optimizer_name,
                            None) is not None)
 
+    def _supported_optims(self):
+        expected_optim_types = [Optimizer]
+        if FAIRSEQ_INSTALLED:
+            # fairseq optims are not torch.optim objects
+            expected_optim_types.append(FairseqOptimizer)
+        return expected_optim_types
+
     # Validate configuration based on command line arguments
     def _do_sanity_check(self):
-        assert isinstance(self.client_optimizer, (type(None), Optimizer, Callable)), \
+        expected_optim_types = self._supported_optims()
+        expected_optim_types += [type(None), Callable]
+        assert isinstance(self.client_optimizer, tuple(expected_optim_types)), \
             f'Client Optimizer is of unexpected type {type(self.client_optimizer)}'
 
         if not self.client_optimizer:
@@ -949,7 +981,7 @@ class DeepSpeedEngine(Module):
             )
 
     def _configure_distributed_model(self, model):
-        self.module = model
+        self.__dict__['module'] = model
         if self.fp16_enabled():
             if self.zero_optimization_partition_weights() and any(
                 [hasattr(param,
@@ -1034,7 +1066,7 @@ class DeepSpeedEngine(Module):
     # Configure optimizer
     def _configure_optimizer(self, client_optimizer, model_parameters):
         if client_optimizer is not None:
-            if isinstance(client_optimizer, Optimizer):
+            if isinstance(client_optimizer, tuple(self._supported_optims())):
                 client_optimizer.param_groups[:] = [
                     pg for pg in client_optimizer.param_groups if len(pg["params"]) != 0
                 ]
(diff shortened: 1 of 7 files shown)


mrwyattii (approved):

cli99 (approved):

deepspeed/runtime/engine.py
```diff
@@ -1091,9 +1129,10 @@ def _configure_optimizer(self, client_optimizer, model_parameters):
                 # If apex/amp is available it will be imported above
                 raise RuntimeError(
                     "Unable to import apex/amp, please make sure it is installed")
-            self.module, self.optimizer = amp.initialize(
+            model, self.optimizer = amp.initialize(
```
tjruwase, line 1132: Does self.module also need updating here?

deepspeed/runtime/engine.py
```diff
@@ -2326,8 +2365,11 @@ def load_moe_state_dict(checkpoint_path,
                         state_dict.update(expert_state_dict)
                     moe_layer_id += 1
 
-    def load_module_state_dict(self, state_dict, strict=True):
-        self.module.load_state_dict(state_dict, strict=strict)
+    def load_module_state_dict(self, state_dict, strict=True, model_f=None):
```
tjruwase, line 2368: Can `model_f` be renamed to be more descriptive. For example, could it be named `custom_load_fn`?

tjruwase (changes requested):

deepspeed/runtime/engine.py
jeffra: It gets set via `_set_client_model(model)` on the next line, we have to be careful how to assign `self.module` now so i factored it out into this helper method.

stas00:
Hi @jeffra, 

The pass-through feature is very neat!

Though it makes things very difficult to debug since it doesn't pass the engine object to the method it passed it to, so it could be a puzzle to the user which thinks a normal engine's `forward` will get called.

```
class A:
    def __init__(self): pass
    def a(self): print(f"a() was called on {type(self)}")

class B:
    def __init__(self, a):
        self.module = a

    def __getattr__(self, name):
        _module = {}
        if "module" in self.__dict__:
            _module = self.__dict__['module']

        if name in dir(self):
            return getattr(self, name)
        elif name in dir(_module):
            return getattr(_module, name)
        else:
            raise AttributeError(
                f"'{type(self).__name__}' object has no attribute '{name}'")

a = A()
b = B(a)
a.a()
b.module.a()
b.a()
```

```
$ python test-generate.py
a() was called on <class '__main__.A'>
a() was called on <class '__main__.A'>
a() was called on <class '__main__.A'>
```

so we can clearly see the `module`'s method will be called with the unwrapped model.

stas00:
So @VictorSanh found that a model's forward breaks under ZeRO3 when the pass-through `generate` is invoked and he narrowed it down to this missing code:

https://github.com/microsoft/DeepSpeed/blob/349f845b838c6992c5ea19e80fc728bef9645962/deepspeed/runtime/engine.py#L1821-L1826

if he manually adds it before calling `generate` everything works.

stas00:
This, of course, also means that any HF `transformers` model with external params will run into this issue regardless of whether they call `generate` on the unwrapped model or using the magic of `getattr` - so this PR didn't break anything directly, and perhaps we ought to open a new Issue instead. Please let me know it'd be more beneficial this way.

VictorSanh:
Thanks Stas!

Some more context about this: The initial error arose when trying to perform generation (inference) on a HF transformers like model (saying "like" because it has some custom components, that i will describe later).

The model is wrapped into a DSEngine, and calling `generate` is possible because of the inheritance described above.
The call to `generate` will then go into HF transformers generation utilities to eventually call the language model's forward (for instance, inside `greedy_search`):

https://github.com/huggingface/transformers/blob/8c5026628a29a89cf122bc1c95cff8101f78c7c0/src/transformers/generation/utils.py#L2195

at that point, `self` is the unwrapped model (no DS Engine anymore).

The error arises in the vocab projection module (i.e. at the very end of the network), which instead of a regular `nn.Linear`, is a custom Linear that allows optionally freeze a slice of it while the other slice is always trainable:
```
class DecoupledLinear(nn.Linear):
    # Derived from https://pytorch.org/docs/stable/_modules/torch/nn/modules/linear.html#Linear
    """
    Implements a decoupling of parameters to allow freezing (or not) a subset of the parameters.
    In practise, the regular `weight` can be trained or frozen (i.e. `partially_freeze=True`), and if `out_additional_features` > 0, then it will create `out_additional_features * in_features` additional parameters that are always trained.
    If `out_additional_features=0`, then the module defaults back to the regular behavior of `nn.Linear`.
    """

    def __init__(
        self,
        in_features: int,
        out_features: int,
        out_additional_features: int = 0,
        bias: bool = True,
        partially_freeze: bool = True,
        device=None,
        dtype=None,
    ) -> None:
        """
        out_additional_features: int. Number of additional trainable dimensions. Only makes sense when `partially_freeze=True`.
        partially_freeze: bool. If True, the regular `weight` will be frozen and extra parameters (if any) will be trainable. If False, default to the regular behavior of nn.Linear.
        """
        super().__init__(in_features, out_features, bias, device, dtype)
        self.out_additional_features = out_additional_features
        self.partially_freeze = partially_freeze

        self.in_features = in_features

        if partially_freeze:
            self.weight.requires_grad_(False)
            if bias:
                self.bias.requires_grad_(False)

        if out_additional_features > 0:
            self.additional_fc = nn.Linear(
                in_features=in_features,
                out_features=out_additional_features,
                bias=bias,
                device=device,
                dtype=dtype,
            )

    def forward(self, input: torch.Tensor) -> torch.Tensor:
        output = F.linear(input, self.weight, self.bias)

        if self.out_additional_features > 0:
            additional_features = F.linear(input, self.additional_fc.weight, self.additional_fc.bias)
            output = torch.cat((output, additional_features), -1)

        return output
```

The error would typically be a `RuntimeError: size mismatch, got 15, 15x2048,0` in the call to `additional_features = F.linear(input, self.additional_fc.weight, self.additional_fc.bias)`.

stas00:
ok, so perhaps we can rescue this situation in this way.

Have deepspeed provide a new method that the user will call before `generate` and it'll do everything that needs to be done, like the missing code 

https://github.com/microsoft/DeepSpeed/blob/349f845b838c6992c5ea19e80fc728bef9645962/deepspeed/runtime/engine.py#L1821-L1826

stas00:
I think this pass-through magic is actually problematic as it swipes the problem under the carpet

Instead fairseq, transformers, and other frameworks - should provide for the wrapped model arg and use it at the point of `__call__`. So the unwrapped model will be used to deal with inheritance. And at the point of `forward` the wrapped engine will be used.

This doesn't have to be deepspeed specific and could be used for any framework that does the wrapping and needs to ensure the wrapping stays during `forward`.

stas00:
**deepspeed solution b:**
A possible mad solution would be to stick the engine object somewhere in the first `forward`'s pre-hook when doing the partitioning and using that pre-hook run it during the unwrapped `forward`, with some multiple call stacking tracking in case it was already invoked directly so it won't run twice.

PR #1953, opened 2022-05-14
Checkpoint reshaping

tjruwase:
Checkpoint reshaping utilities 
Universal checkpoint


stas00:
Cross link the other half: https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/239

rocm-mici:

deepspeed/runtime/engine.py
```diff
@@ -2673,25 +2685,35 @@ def get_sparse_tensor_module_names(original_set,
...
+        if self.load_universal_checkpoint():
+            zero_sd_list = None
+            checkpoint_folder = f'{os.path.join(load_dir, tag)}'
+        else:
+            if load_optimizer_states and self.dp_world_size != self.loaded_checkpoint_dp_world_size:
+                raise ZeRORuntimeException("The checkpoint being loaded used a DP " \
+                    f"world size of {self.loaded_checkpoint_dp_world_size} but the " \
+                    f"current world size is {self.dp_world_size}. Automatic adjustment " \
+                    "of ZeRO's optimizer state partitioning with a new world size is not " \
+                    "currently supported.")
```
jeffra, line 2697: Should we recommend here to use of the universal checkpoint tool in the error message?

stas00: Absolutely, except there is no such tool in the deepspeed repo. We only have it inside Meg-DS at the moment.

stas00: such a tool at the moment is very client- or should I say model-specific as it has to do with individual param names and prescribes how each should be merged.

I think most of it can be abstracted to the deepspeed side and have the client only provide a few dicts with merging rules, so the client side script would be very thin.

stas00: eventually we also want to expand this functionality to all ZeRO stages, so that `zero_to_fp32.py` will become much much simpler than its current very difficult to follow version.

mrwyattii (approved):

jeffra (approved):

deepspeed/checkpoint/constants.py
```diff
@@ -11,15 +11,30 @@
...
+
+#########################################
+# Checkpoint naming constants
+#########################################
+MODEL_FILE_PREFIX = 'mp_rank_'
+ZERO_FILE_PREFIX = 'bf16_' + 'zero_pp_rank_'
+OPTIM_FILE_SUFFIX = '_optim_states.pt'
+MODEL_FILE_SUFFIX = '_model_states.pt'
+LAYER_FILE_PREFIX = 'layer_'
+BF16_ZERO_FILE_PREFIX = ZERO_FILE_PREFIX
```
Muennighoff, line 35: Can we not support fp16 for ckpt reshaping?
Should be `ZERO_FILE_PREFIX = 'zero_pp_rank_'` in that case

tjruwase: Apologies, this was a hack that I forgot to clean up. Yes, the goal is to support for fp16 and bf16. Will cleanup asap.

deepspeed/checkpoint/deepspeed_checkpoint.py
```diff
@@ -0,0 +1,316 @@
...
+    def _build_pp_transformer_map(self):
+        data_map = {}
+        transformer_layers = self.layer_keys[1:-1]
+        layers_per_pp = len(transformer_layers) // self.pp_degree
+        data_map = {
+            i: transformer_layers[i * layers_per_pp:(i + 1) * layers_per_pp]
+            for i in range(0,
+                           self.pp_degree)
+        }
+        return data_map
```
Muennighoff, line 245: This secretly ignores some layers, e.g.
`layer_keys = ['layer_01', 'layer_03', 'layer_04', 'layer_05', 'layer_06', 'layer_07', 'layer_08', 'layer_09', 'layer_10', 'layer_11', 'layer_12', 'layer_13', 'layer_14', 'layer_15', 'layer_16', 'layer_17', 'layer_18', 'layer_19', 'layer_20', 'layer_21', 'layer_22', 'layer_23', 'layer_24', 'layer_25', 'layer_26', 'layer_27', 'layer_28', 'layer_29', 'layer_30', 'layer_31', 'layer_32', 'layer_34']`
&
`pp_degree = 4`

yields

`{0: ['layer_03', 'layer_04', 'layer_05', 'layer_06', 'layer_07', 'layer_08', 'layer_09'], 1: ['layer_10', 'layer_11', 'layer_12', 'layer_13', 'layer_14', 'layer_15', 'layer_16'], 2: ['layer_17', 'layer_18', 'layer_19', 'layer_20', 'layer_21', 'layer_22', 'layer_23'], 3: ['layer_24', 'layer_25', 'layer_26', 'layer_27', 'layer_28', 'layer_29', 'layer_30']}`

tjruwase: Do you mean 'layer_01' and 'layer_34'?

Muennighoff: No `'layer_31', 'layer_32'`

deepspeed/checkpoint/reshape_utils.py
```diff
@@ -0,0 +1,87 @@
...
+            f'Cannot merge two states of types {type(state_a)} and type {type(state_b)}')
+
+    if type(state_a) in (dict, OrderedDict):
+        return merge_state_dict(state_a, state_b, key_list)
+    elif type(state_a) in (list, tuple):
+        return type(state_a)(merge_state_list(state_a, state_b, key_list))
+    elif torch.is_tensor(state_a):
+        return torch.cat([state_a, state_b], 0)
+    else:
+        return state_a
```
Muennighoff, line 87: Maybe it'd be nice to have an arg to pass a list of keys where always `state_a` should be taken?
Currently, this merges RNG states for me, which is undesirable

tjruwase: This is correct. I had plans to do this as I gained more understanding of the various merge semantics but never got around to it. In general, it seems there are three classes of merge: (1) concat, (2) pick any (or `state_a` in this case) , and (3) average. So, my thought was to pass in some sort of mapping of keys to one of these three merge strategies. In practice, it seems (1) and (2) only apply for Megatron-DeepSpeed. Is this useful information? 

Also, please note that we somewhat abandoned this approach in favor of the universal checkpointing implemented [here](https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/239/files#diff-6c7e4f412a4bdd7ccd977eaacfcf8eb7bedfffb0c08e65776c00df9aced4f129).  You can see the merging logic for universal approach [here](https://github.com/bigscience-workshop/Megatron-DeepSpeed/blob/a247614b2fa41326ac8482a1735df867778cec85/tools/convert_checkpoint/ds_to_universal.py#L233-L239).

deepspeed/checkpoint/deepspeed_checkpoint.py
tjruwase: Hmm, that is indeed concerning. I will take a closer look. Thanks for catching this.

deepspeed/checkpoint/reshape_utils.py
Muennighoff: Nice, that approach is a lot cleaner.
Do you think it's a big problem we're not averaging the layernorms in the above & thus [here](https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/239/files#diff-ecfdbe3a107133adc51e3f7c5d73ef006d6b06bd18bfc16ecb4708834e6fdc97)?
My assumption is that as long as we continue training afterwards it should be completely fine

tjruwase: Yes, I believe we skipped the averaging because the layernorms in the Big Science model get automatically fixed when we continue training. But we should probably average to generalize to other models, so this needs a TODO.

deepspeed/checkpoint/constants.py
Muennighoff: Do you have an estimate when there will be fp16 support?

stas00: I was away and now Tunji is away this week.

Given the lack of resources this is need driven, so first need to complete the bf16 work and then can look at fp16 next if there is a reasonable need for it and of course I can't speak on Tunji's behalf of his availability and priorities.

deepspeed/checkpoint/deepspeed_checkpoint.py
stas00: hmm, I wonder if we have missed something here. thank you for catching that.

in BS 176B we use `    --pp-partition-method 'type:transformer|embedding' `
https://github.com/bigscience-workshop/bigscience/blob/7f9bf77c2e0905d2db418684f58fc91479f17dc8/train/tr11-176B-ml/tr11-176B-ml.slurm#L104

and then 2 PP layers less, as the embed layers are about the same size as PP layers, so the PP layers aren't divisible by PP degree.

But we did quality tests which should have failed if some layers were skipped and then initted randomly.

@tjruwase, did you not run your test scripts with `--pp-partition-method 'type:transformer|embedding'`? I thought we have matched the set ups.

stas00: ah, but this is not the universal checkpoint, is it not? That would explain it - as the previous path was never finished in favor of the universal checkpoint method.

deepspeed/checkpoint/reshape_utils.py
stas00: layernorms should be identical across TPs. Except they aren't due a bug somewhere in Meg-DS which only manifests at huge models. Which we workaround by forcefully resyncing layernorms on each forward path using this branch https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/272 so it should be safe to just pick the first one.

deepspeed/checkpoint/deepspeed_checkpoint.py
```diff
@@ -0,0 +1,316 @@
...
+
+        self.old_2d_map = meg_2d_parallel_map(self.original_pp_degree,
+                                              self.original_tp_degree)
+        self.old_2d_map.simple_init()
+        self.new_2d_map = reshape_meg_2d_parallel(old_pp_degree=self.original_pp_degree,
+                                                  old_tp_degree=self.original_tp_degree,
+                                                  new_pp_degree=self.pp_degree,
+                                                  new_tp_degree=self.tp_degree)
+
+        self.zero_checkpoint = ZeROCheckpoint(dir)
```
Muennighoff, line 72: This breaks if loading a checkpoint not trained with ZeRO, as there are no ZeRO files.
Is `DeepSpeedCheckpoint` only for models trained with ZeRO?

stas00: I think at the moment it is.

So it needs to be further developed to support non-ZeRO checkpoints, but probably in another PR? Perhaps just adding an assert here saying that this feature isn't yet supported?

stas00:
If possible let's please merge this as we need to merge this PR https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/239 and it depends on the current PR. Thank you!

tests/unit/test_checkpointing.py
```diff
@@ -1260,8 +1274,8 @@ def _go2(models):
     def _go1(models):
         ds_config["zero_optimization"]["elastic_checkpoint"] = elastic_load
         model, _, _, _ = deepspeed.initialize(config=ds_config,
-                                              model=models[1],
-                                              model_parameters=models[1].parameters())
+                                                  model=models[1],
+                                                  model_parameters=models[1].parameters())
```
stas00, line 1278: this indent looks strange.

tests/unit/test_checkpointing.py
```diff
@@ -1190,13 +1190,22 @@ def _go():
...
         ds_config["zero_optimization"]["elastic_checkpoint"] = elastic_load
         model, _, _, _ = deepspeed.initialize(config=ds_config,
                                               model=models[1],
                                               model_parameters=models[1].parameters())
         model.load_checkpoint(tmpdir, load_optimizer_states=load_optim)
+
+        if load_optim:
+            saved_sd = torch.load(os.path.join(tmpdir, 'opt-state-dict'))
+            curr_sd = model.optimizer.optimizer.state_dict()
+            assert curr_sd['param_groups'] == saved_sd['param_groups']
```
stas00, line 1208: so one workflow fails here: https://github.com/microsoft/DeepSpeed/runs/7184872087?check_suite_focus=true

mismatching on the number of params:

```
   -   'params': [139739444197968],
  ?                     ^^ ^^^ -
  +   'params': [139739450210816],
  ?                     ^^^ ^^^
```

deepspeed/checkpoint/deepspeed_checkpoint.py
tjruwase: We already have an assert for this [here](https://github.com/microsoft/DeepSpeed/blob/ecb3dc8a2b8d1d173a7696026a935c978a61e2c5/deepspeed/checkpoint/zero_checkpoint.py#L26).

PR #1977, opened 2022-05-25
[tests] dump connection info if distributed init fails

jeffra:
Occasionally on AMD CI runners we are seeing random address in use errors, hoping to dump some additional context if/when this happens again. We're currently unable to trigger this again so adding this code into `master` in case it happens again.


mrwyattii (approved):

rocm-mici:

Issue #327, opened 2020-08-21
Generic Tensorboard Utility in Deepspeed 


rsn870:
Hi ,

The tensorboard config present by default in Deepspeed only provides logging for train/loss and train/lr . This feature however is quite limited , for eg in my training :
inputs = data.to(model_engine.device)
mean_style = mean_style.to(model_engine.device)
latent_rnd =  torch.randn(micro_batch_size, 512).to(model_engine.device)
losses , encode_lst = model_engine(inputs,mean_style,latent_rnd,mode='train')
total = torch.sum(torch.stack(losses))

It is total which I pass to backward as loss thus I would get graphs only for this . However I am interested in viewing the graphs for each individual component of losses list as well to finetune optimization if necessary , this however requires me to use an external tensorboard. 

Instead of using the default config then can a utility be made to insert a 'customizable' tensorboard if possible ?

ShadenSmith:
A simple first step might be to give users an access point to our `SummaryWriter` object in order to allow for arbitrary logging from client codes? We could also provide a simple `log_scalar()` interface so folks can do some simple logging without diving into TensorBoard.

I can see some confusion coming from the asymmetry of the `SummaryWriter` only being present on global rank 0.

lekurile:
Hi @rsn870,

Since the logging capabilities were expanded in GH-2013 to support TensorBoard, WandB, and CSV logging formats ([see documentation here](https://www.deepspeed.ai/docs/config-json/#monitoring-module-tensorboard-wandb-csv)), I'll close the issue for now.

Feel free to open another issue if there are additional requests for expanded logging capabilties.

Thanks,
Lev

PR #1996, opened 2022-06-06
DeepSpeed Monitor Module

Quentin-Anthony:
This PR introduces a general `Monitor` tool which manages logging with torch's TensorBoard, [WandB](https://wandb.ai/home), or CSV files. 

This would allow users to add custom logging such as https://github.com/microsoft/DeepSpeed/issues/327, and enables DeepSpeed to easily apply logging outside the engine.

TODO:

- [x] Basic functionality for each monitor
- [x] Refactor into module
- [x] Add documentation + examples
- [x] Add pydantic config support
- [x] Add monitor tests


deepspeed/runtime/engine.py
```diff
@@ -488,53 +490,8 @@ def curriculum_enabled(self):
...
-        except ImportError:
-            print(
-                'If you want to use tensorboard logging please `pip install tensorboard`'
-            )
-            raise
-        from torch.utils.tensorboard import SummaryWriter
-
-        return SummaryWriter(log_dir=log_dir)
+    def monitor_enabled(self):
+        return self._config.tensorboard_enabled or self._config.wandb_enabled or self._config.csv_monitor_enabled
```
jeffra, line 494: Can we move this check to `utils/monitor.py` somewhere? This way in the future if we add new monitoring backends we don't have to add any code to the engine and only to the monitor itself.

Quentin-Anthony: Done

deepspeed/monitor/wandb.py
```diff
@@ -0,0 +1,20 @@
+from .utils import *
+from .monitor import *
```
jeffra, line 2: Can we use explicit imports here? I think there are only two things we need to import right? `check_wandb_availability` and `Monitor`?

Quentin-Anthony: Done

rocm-mici:

Quentin-Anthony:
@jeffra -- This is ready to review again when you get time

jeffra (approved):
🚀

PR #2000, opened 2022-06-07
Graceful exit on failures for multi-node runs

jerrymannil:
* Use Popen.terminate() to stop the child processes gracefully

* The Popen.kill() command cause the child training processes to end abruptly. This may cause the child processes to become zombies without communicating properly to the launcher process about the kill signal. So the launcher process continue to wait for signals from child process, causing the ssh command to not return back to the pdsh command

Fixes microsoft#1995


jerrymannil:
Need to do some more testing

Issue #653, opened 2021-01-10
[config] no param naming validation


stas00:
I have noticed that one could make a typo in the config file and DeepSpeed silently ignores it. e.g. I did:
```
    "gradient_clipp1ng": 1
```
and got no errors.  

Not sure if it should validate the value as well, as float is expected, but I guess int is fine as it's just converted to float.

This lack of validation could lead to user errors which could be hard to notice.

Nothing urgent, but at some point it'd be good to tell the user if they added a config entry which is invalid.

Thank you.

tjruwase:
@stas00 Apologies for the radio silence and for stalling your progress.

Your point is well made regarding silent errors from typos in the config file. In the past, users have actually exploited this lack of validation as a way to convenient disable features without deleting the entries. So perhaps a reasonable solution would be to introduce yet another config option to validate the config entries and issue a warning for each unrecognized entry. What do you think?

stas00:
Oh, no problem, I'm sure you guys need some breaks too!

Your suggestion to make the validation optional would work great, @tjruwase. This is a very low priority feature request.

Since you mentioned that the lack of validation is exploited to make users' life easier,  I'm also thinking that json isn't a user-friendly format for configuration, because it has a ton of problems, e.g.:
- don't put comma here, but must put it here, 
- can't use comments, 
- can't use python 5_000_000_000 numbers - who can read 5000000000
json, IMHO, is more of a program-intermediary format, rather than designed for human manipulation.

It'd have been much better to use some other, more user-friendly format, IMHO. But it has been a while since I had to deal with config files so I'm not sure what's the standard these days.

tjruwase:
Completely agree with the pain of json, so we are very open to any suggestions.

jeffra:
Thanks @stas00, JSON validation I think is on our radar and definitely something we should add in the short term. Also agree JSON isn't the most user-friendly of formats. We'll have to look into alternatives. I have heard good things about [hjson](https://hjson.github.io/). It allows commas/comments but I don't think supports numbers in the way you describe unfortunately. YAML might also be a reasonable option but also comes with difficulties haha

stas00:
I was going to suggest YAML, but I thought that something better must have been invented since then. YAML is from around 2001 if I remember correct.

I have never used https://hjson.github.io/, but it does look like the most straightforward solution as it was created to overcome limitations of json for human use. May be that's all is needed. 

[...]

PR #2004, opened 2022-06-09
Refactor ZeRO configs to use Pydantic

mrwyattii:
Brings Pydantic into the zero config:
- Defines a base `DeepSpeedConfigModel` class which ZeRO and offload configs inherit. This class allows us to enforce extra params are not provided in the config and also implements logic for deprecated config params
- Unify the ZeRO (and offload) `constants.py` and `config.py` into a single Pydantic class that contains the default values for all parameters
- Changes in the rest of the code to remove values defined in `constants.py`


jeffra:
This PR is aiming to help with this longstanding issue https://github.com/microsoft/DeepSpeed/issues/653

/cc @stas00

stas00:
Wonderful! Thank you for the heads up, Jeff!

Also probably extending hjson-support to the rest of DS config would be great, but doesn't have to be part of this PR - just a suggestion.

jeffra:
Absolutely, I think adding hjson + pydantic support is the eventual goal. One all of our configs are managed by pydantic the goal is to also replace the [config-json](https://www.deepspeed.ai/docs/config-json/) page with a [read-the-docs style page](https://deepspeed.readthedocs.io/en/latest/) that is auto-generated from doc-strings that are easier to maintain and more accurate. However, one step at a time :)

rocm-mici:

deepspeed/runtime/config.py
```diff
@@ -864,7 +851,13 @@ def _initialize_params(self, param_dict):
         self.gradient_predivide_factor = get_gradient_predivide_factor(param_dict)
         self.sparse_gradients_enabled = get_sparse_gradients_enabled(param_dict)
 
-        self.zero_config = DeepSpeedZeroConfig(param_dict)
+        if ZERO_OPTIMIZATION in param_dict:
```
tjruwase, line 854: Refactoring this logic could make it easier to read. Something like 
`self.zero_config = get_...`

deepspeed/autotuning/autotuner.py
```diff
@@ -499,7 +498,7 @@ def tune_space(self,
                    prev_best_mbs=0,
                    prev_best_metric_val=0):
         config_zero = tuning_space.get(ZERO_OPTIMIZATION, {})
-        stage = config_zero.get(ZERO_OPTIMIZATION_STAGE, ZERO_OPTIMIZATION_STAGE_DEFAULT)
+        stage = config_zero.stage
```
tjruwase, line 501: Isn't this a bug if `config_zero` equals `{}`, and leads to evaluation of `{}.stage`?

deepspeed/runtime/config_utils.py
```diff
@@ -8,6 +8,86 @@
...
+from deepspeed.utils import logger
+
+
+class DeepSpeedConfigModel(BaseModel):
+    """
+    This class should be used as a base for all DeepSpeed configs. It extends
+    pydantic.BaseModel to allow for deprecated fields. To enable this feature,
+    add deprecated=True to pydantic.Field:
+
+    my_dep_field: int = Field(0, deprecated=True)
```
tjruwase, line 21: Can you document first argument of `Field(`, i.e., `0`.?

tjruwase (approved):

PR #2008, opened 2022-06-09
Graceful exit on failures for multi-node runs

jerrymannil:
* Use Popen.terminate() to stop the child processes gracefully; Kill them if terminate doesn't work

* The Popen.kill() command cause the training processes to end abruptly. This may cause the child processes to become zombies without communicating properly to the parent process about the kill signal. So the ssh session continue to wait for signals from the child processes, causing it to not return back to the pdsh command

Fixes microsoft#1995


jerrymannil:
@jeffra can you look at this ?

jerrymannil:
@jeffra @tjruwase Can one of you help with merging ?

jerrymannil:
@jeffra @tjruwase  Can one of you help with merging ?

tjruwase:
@jerrymannil, taking a look. Apologies for the delay.

tjruwase (approved):

jeffra (approved):
Thank you @jerrymannil , so sorry for the delay.

PR #2017, opened 2022-06-14
update CODEOWNERS

conglongli:


jeffra:
Thank you @conglongli! Can you also add @arashb @xiaoxiawu-microsoft. I think they are the only ones missing?

jeffra (approved):

PR #2024, opened 2022-06-16
Fix inference unit test import error catching

mrwyattii:
Fixes the error we see here, when not running inference tests: https://github.com/microsoft/DeepSpeed/runs/6912710983?check_suite_focus=true#step:7:1409

This change also addresses a bug where an import error when `-m "inference"` is passed to pytest and there is an import error, all tests are skipped, and the test succeeds

Also adds a check at the beginning of testing for expected torch and cuda versions (this is in response to seeing that pip will sometimes upgrade torch when setting up the environment)


tests/unit/test_inference.py
```diff
@@ -17,7 +17,7 @@
     from lm_eval.evaluator import evaluate
     from transformers import pipeline, AutoModelForCausalLM, AutoTokenizer
     from huggingface_hub import HfApi
-except ImportError:
+except (ImportError, AssertionError) as e:
```
tjruwase, line 20: Is it safe to skip every assertion error?

mrwyattii: That's a good point, actually this setup could skip all tests if there is an AssertionError and ImportError with `-m "inference"` and show passing. I think we can add another check to see if the `inference` tag was passed to pytest (otherwise, we don't care if there is an error importing here)

jeffra (approved):

PR #2028, opened 2022-06-17
Fixing several issue in API and kernels to run inference with model-parallelism

RezaYazdaniAminabadi:


tests/unit/test_inference.py
```diff
@@ -321,3 +321,38 @@ def _go():
...
+            mp_size=2,
+            dtype=dtype,
+            replace_method="auto",
+            replace_with_kernel_inject=True,
+            enable_cuda_graph=enable_cuda_graph,
+        )
+        bs_output = pipe(query, **inf_kwargs)
+        ds_output = pipe(query, **inf_kwargs)
+
+        assert assert_fn(bs_output, ds_output)
```
RezaYazdaniAminabadi, line 356: @mrwyattii, Thanks for adding the test for the MP inference. For the test in this case, we might want to do another check to see if the results between GPUs are similar.

mrwyattii: Added!

mrwyattii (approved):
LGTM

cmikeh2:
MP working on current inference, closing PR.

PR #2035, opened 2022-06-21
Share a list of weight attributes instead of a single one in TiedLayerSpec API

thomasw21:


deepspeed/runtime/pipe/module.py
```diff
@@ -419,24 +428,27 @@ def _partition_layers(self, method='uniform'):
...
 
     def _synchronize_tied_weights(self):
         for key, comm in self.tied_comms.items():
-            dist.broadcast(
-                getattr(comm['module'],
-                        comm['weight_attr']),
-                src=min(comm['ranks']),
-                group=comm['group'],
-            )
+            for weight_attr in comm['weight_attr']:
```
Muennighoff, line 445:
```suggestion
            for weight_attr in comm['weight_attrs']:
```

deepspeed/runtime/pipe/module.py
```diff
@@ -191,7 +203,7 @@ def forward(self, inputs):
         self.forward_funcs = []
         self.fwd_map = {}
         self.tied_modules = nn.ModuleDict()
-        self.tied_weight_attrs = {}
+        self.tied_weight_attrss = {}
```
tjruwase, line 206: Is the double `s` a typo or intentional?

tjruwase:
@thomasw21, thanks for the PR and apologies for the delay in reviewing. Can you please provide a clarification? It seems there are two issues being addressed in this PR: (1) supporting user-specified forward function and (2) list of weight attributes. If this is correct, I think it might be better to split into two separate PRs. What do you think? Thanks!

thomasw21:
Indeed! Sorry I've sort of dropped this as well currently, as we've been focusing on other aspects of BigScience. I'll try split the PRs when I get the chance! My bad!

jomayeri:
@tjruwase Are these changes we still want? If so I can revive them with the current develop branch.

PR #2041, opened 2022-06-22
Minimizing memory usage for Mixture of Experts + ZeRO-2

siddharth9820:
Keeping the number of GPUs and the base model fixed, I have noticed that memory usage in the gradient upscaling phase becomes a bottleneck.
Here's an example:

- Base Model = 2.7B
- GPUs = 32 40GB A100s
- Global batch size = 32
- Microbatch size = 1

image

Memory usage grows linearly with the number of experts and number of layers and quadratically with the hidden size!
Which means that it is a much bigger problem for larger base models like 6.7B which goes OOM at 16 experts.

The fix:

- Running the optimizer for a parameter group as soon as the fp-32 gradients are ready. This does away with the need to generate all fp-32 gradients.

- Tiling the expert parameter group.  It's size grows linearly with the number of experts (4ENH*H/G)

After this fix I am able to run a 6.7B base model with 16 experts on 32 GPUs.

Todo

- [x] Check if gradient norms are being calculated correctly
- [x] Check for model convergence, both for MoEs as well as dense models
- [ ] Should the decision to tile and how many tiles be left on the user or automated?
- [x] Code formatting


siddharth9820:
The logic for calculating the l2-norm of the gradients looks sounds to me. It matches the original logic in ZeRO-2.
Ran training for a few iterations on GPT-6B + 16 experts on 32 40 GB A 100 GPUs, and the curves look to be going in the right direction.
image

PR #2043, opened 2022-06-22
Add MuP optimizers

mrwyattii:
Integrating the optimizers from the [MuP](https://github.com/microsoft/mup) project.


tests/unit/test_mup_optimizers.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) Microsoft Corporation.
```
loadams, line 1: What folder should this go into inside of `tests/unit`?

mrwyattii: Probably `unit/runtime/`, but maybe in one of the subdirs. Look for where we have other tests related to optimizers!

loadams: Moved to runtime.  The other optimizers are ops, so they are in folders there.  So just moving this to runtime for now.

mrwyattii (commented):
LGTM, thanks @loadams

loadams (approved):

PR #2047, opened 2022-06-23
Fix for AMD unit tests

mrwyattii:
- Add asserts to prevent FP16 with CPUAdam on AMD CPUs (not supported)
- AMD runners now use a venv, removed sudo from pip install
- Add skips to unit tests for CPUAdam

### DIFF
--- a/deepspeed/ops/adam/cpu_adam.py
+++ b/deepspeed/ops/adam/cpu_adam.py
@@ -76,6 +76,12 @@ class DeepSpeedCPUAdam(torch.optim.Optimizer):
                             amsgrad=amsgrad)
         super(DeepSpeedCPUAdam, self).__init__(model_params, default_args)
 
+        self.cpu_vendor = get_cpu_info()["vendor_id_raw"].lower()
+        if "amd" in self.cpu_vendor:
+            for group_id, group in enumerate(self.param_groups):
+                for param_id, p in enumerate(group['params']):
+                    assert p.dtype != torch.half, "FP16 params not supported on AMD CPUs"
+
         self.opt_id = DeepSpeedCPUAdam.optimizer_id
         DeepSpeedCPUAdam.optimizer_id = DeepSpeedCPUAdam.optimizer_id + 1
         self.adam_w_mode = adamw_mode
@@ -128,6 +134,9 @@ class DeepSpeedCPUAdam(torch.optim.Optimizer):
 
         # intended device for step
         device = torch.device('cpu')
+
+        if "amd" in self.cpu_vendor:
+            assert fp16_param_groups is None, "FP16 params not supported on AMD CPUs"
 
         # converting the fp16 params to a group of parameter
         if type(fp16_param_groups) is list:
--- a/deepspeed/ops/adam/cpu_adam.py
+++ b/deepspeed/ops/adam/cpu_adam.py
@@ -8,6 +8,7 @@ import time
 from cpuinfo import get_cpu_info
 from pathlib import Path
 from ..op_builder import CPUAdamBuilder
+from deepspeed.utils import logger
 from deepspeed.utils.logging import should_log_le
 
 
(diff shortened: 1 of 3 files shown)


deepspeed/ops/adam/cpu_adam.py
```diff
@@ -129,6 +136,9 @@ def step(self, closure=None, fp16_param_groups=None):
         # intended device for step
         device = torch.device('cpu')
 
+        if "amd" in self.cpu_vendor:
+            assert fp16_param_groups is None, "FP16 params not supported on AMD CPUs"
```
RezaYazdaniAminabadi, line 140: I think we don't need to assert for this condition. The fp16_param_groups is supposed to be on GPU and it is okay on AMD.

deepspeed/ops/adam/cpu_adam.py
```diff
@@ -76,6 +78,18 @@ def __init__(self,
                             amsgrad=amsgrad)
         super(DeepSpeedCPUAdam, self).__init__(model_params, default_args)
 
+        self.cpu_vendor = get_cpu_info()["vendor_id_raw"].lower()
+        if "amd" in self.cpu_vendor:
+            for group_id, group in enumerate(self.param_groups):
+                for param_id, p in enumerate(group['params']):
+                    if p.dtype == torch.half:
+                        logger.warning(
+                            "FP16 params for CPUAdam may not work on AMD CPUs")
```
jithunnair-amd, line 87: @mrwyattii Can you please elaborate a bit more on what error/issue you observed when you added this warning?

rraminen: w.r.t https://github.com/microsoft/DeepSpeed/issues/4698
cc: @loadams

RezaYazdaniAminabadi: Hi @rraminen, Hi @jithunnair-amd 

I think CPU-Adam should be supported to some extent but some overlapping features for copying data from CPU to GPU may not be working properly. Also, I remember that I could not find the corresponding vector instructions for handling the fp16 data-type. So, we may be able to remove this and let it run safely only for restricted cases. Or, if you can help verify the functionality that we need for CPU-Adam is supported, we can have another PR to add this?
Thanks,
Reza

jithunnair-amd: @RezaYazdaniAminabadi You mentioned: "I could not find the corresponding vector instructions for handling the fp16 data-type." Would that be the ones used in SIMD_STORE2 and SIMD_LOAD2 here: https://github.com/microsoft/DeepSpeed/blob/master/csrc/includes/simd.h#L55

rraminen: @RezaYazdaniAminabadi,
Below are the only references to FP16 instructions in simd.h: 

https://github.com/microsoft/DeepSpeed/blob/a8b821535aa0b254efa681d51b4951734ca021cc/csrc/includes/simd.h#L36
https://github.com/microsoft/DeepSpeed/blob/a8b821535aa0b254efa681d51b4951734ca021cc/csrc/includes/simd.h#L58

Those conversion instructions are part of AVX512F which is supported in Zen4 as it is very basic AVX512 feature. Do you want to qualify that this code works fine on newer than AMD Zen 4 CPUs?

sfc-gh-reyazda: Hi @rraminen @jithunnair-amd 

sorry for the delay. 
@jithunnair-amd, I think except the loading part there were some casting operations used which is kind of hacky and I had to do that to go around loading the data right. Honestly, I have not tested this part of the code recently but I would be happy to change this part based on your suggestion (and remove the warning or give warning for older architectures as @rraminen mentioned). Unfortunately, I don't have AMD CPUs to try this so I would need your help to verify this.
Thanks,
Reza

Issue #1987, opened 2022-05-31
[BUG] run unit test shows 'Cannot re-initialize CUDA in forked subprocess'


delock:
**Describe the bug**
When run unit test test_zero2_reduce_scatter_off, get error message: "Cannot re-initialize CUDA in forked subprocess"

**To Reproduce**
1. Go to tests/unit
2. run command 'pytest -k test_zero2_reduce_scatter_off'
3. The test would fail and see message "RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method"
4. Most of unit test have same error message when they have @distributed_test decorator
5. Manually modify common.py to use spawn method does not solve the problem, because in spawn cannot pickle local function

**Expected behavior**
The test would not fail with error message "RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method"

**ds_report output**
```
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
cpu_adam ............... [NO] ....... [OKAY]
cpu_adagrad ............ [NO] ....... [OKAY]
fused_adam ............. [NO] ....... [OKAY]
fused_lamb ............. [NO] ....... [OKAY]
 [WARNING]  please install triton==1.0.0 if you want to use sparse attention
sparse_attn ............ [NO] ....... [NO]
transformer ............ [NO] ....... [OKAY]
stochastic_transformer . [NO] ....... [OKAY]
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
 [WARNING]  async_io: please install the libaio-dev package with apt
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
async_io ............... [NO] ....... [NO]
utils .................. [NO] ....... [OKAY]
quantizer .............. [NO] ....... [OKAY]
transformer_inference .. [NO] ....... [OKAY]
--------------------------------------------------
DeepSpeed general environment info:
[...]

PR #2049, opened 2022-06-24
Fix building package without a GPU

### LINKED ISSUE #2010: [BUG] Impossible to prebuild w/o having at least one gpu
**Describe the bug**

The goal is to cross-compile on a VM that has no GPUs for other VMs that have GPUs (preparing docker images for CIs). Currently the prebuilding fails.

**To Reproduce**

Normally when gpus are enabled all work., Now let's emulate no GPUs installed by adding `CUDA_VISIBLE_DEVICES=` and repeat the same prebuild command - and then it fails.
```
$ CUDA_VISIBLE_DEVICES= TORCH_CUDA_ARCH_LIST="6.1;8.0;8.6" DS_BUILD_CPU_ADAM=1 DS_BUILD_AIO=1  DS_BUILD_UTILS=1 pip install -e . --global-option="build_ext" --global-option="-j8" --no-cache -v --disable-pip-version-check 2>&1 
WARNING: Disabling all use of wheels due to the use of --build-option / --global-option / --install-option.
Using pip 22.0.4 from /home/stas/anaconda3/envs/py38-pt111/lib/python3.8/site-packages/pip (python 3.8)
Obtaining file:///mnt/nvme0/code/github/00optimize/deepspeed
  Preparing metadata (setup.py): started
  Running command python setup.py egg_info
    File "<string>", line 2, in <module>
    File "<pip-setuptools-caller>", line 34, in <module>
    File "/mnt/nvme0/code/github/00optimize/deepspeed/setup.py", line 238, in <module>
      bf16_support = torch.cuda.is_bf16_supported()
    File "/home/stas/anaconda3/envs/py38-pt111/lib/python3.8/site-packages/torch/cuda/__init__.py", line 92, in is_bf16_supported
      return torch.cuda.get_device_properties(torch.cuda.current_device()).major >= 8 and cuda_maj_decide
    File "/home/stas/anaconda3/envs/py38-pt111/lib/python3.8/site-packages/torch/cuda/__init__.py", line 481, in current_device
      _lazy_init()
    File "/home/stas/anaconda3/envs/py38-pt111/lib/python3.8/site-packages/torch/cuda/__init__.py", line 216, in _lazy_init
      torch._C._cuda_init()
  RuntimeError: No CUDA GPUs are available
  No CUDA runtime is found, using CUDA_HOME='/usr/local/cuda-11.6'
  [WARNING] Torch did not find cuda available, if cross-compiling or running with cpu only you can ignore this message. Adding compute capability for Pascal, Volta, and Turing (compute capabilities 6.0, 6.1, 6.2)
  DS_BUILD_OPS=0
  Installed CUDA version 11.6 does not match the version torch was compiled with 11.5 but since the APIs are compatible, accepting this combination
  Install Ops={'cpu_adam': 1, 'cpu_adagrad': False, 'fused_adam': False, 'fused_lamb': False, 'sparse_attn': False, 'transformer': False, 'stochastic_transformer': False, 'async_io': 1, 'utils': 1, 'quantizer': False, 'transformer_inference': False}
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.

```

**Expected behavior**

I wonder if perhaps we need a new flag that tells the prebuild not to check if there is an actual GPU installed?

We are using `TORCH_CUDA_ARCH_LIST` to cross-compile for the target gpus.

**ds_report output**

```
cpu_adam ............... [YES] ...... [OKAY]
sparse_attn ............ [NO] ....... [OKAY]
async_io ............... [YES] ...... [OKAY]
utils .................. [YES] ...... [OKAY]
torch install path ............... ['/home/stas/anaconda3/envs/py38-pt111/lib/python3.8/site-packages/torch']
torch version .................... 1.11.0+cu115
torch cuda version ............... 11.5
nvcc version ..................... 11.6
deepspeed install path ........... ['/mnt/nvme0/code/github/00optimize/deepspeed/deepspeed']
deepspeed info ................... 0.6.6+828ab718, 828ab718, master
deepspeed wheel compiled w. ...... torch 1.11, cuda 11.5
```

**Screenshots**

 - OS: [e.g. Ubuntu 21.10]
 - Python version 3.8

Thank you.

@jeffra 

cc: @ydshieh, who originally reported this

aphedges:
Fixes #2010.

The first commit is the minimal change needed to get DeepSpeed installed. The second commit that fully removes `bf16_support` from `torch_info`. If the second commit goes too far, it can be dropped before this PR is merged. I have tested both and confirmed that they both allow me to install DeepSpeed. See <https://github.com/microsoft/DeepSpeed/issues/2010#issuecomment-1165869012> for more information about my approach.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -235,7 +235,7 @@ if torch_available and torch.version.cuda is not None:
         nccl_version = ".".join(str(torch.cuda.nccl.version())[:2])
     else:
         nccl_version = ".".join(map(str, torch.cuda.nccl.version()[:2]))
-    if hasattr(torch.cuda, 'is_bf16_supported'):
+    if hasattr(torch.cuda, 'is_bf16_supported') and torch.cuda.is_available():
         bf16_support = torch.cuda.is_bf16_supported()
 if torch_available and hasattr(torch.version, 'hip') and torch.version.hip is not None:
     hip_version = ".".join(torch.version.hip.split('.')[:2])


stas00:
Looks great, @aphedges! Thank you for working on fixing this!

setup.py
```diff
@@ -235,13 +234,10 @@ def create_dir_symlink(src, dest):
         nccl_version = ".".join(str(torch.cuda.nccl.version())[:2])
     else:
         nccl_version = ".".join(map(str, torch.cuda.nccl.version()[:2]))
-    if hasattr(torch.cuda, 'is_bf16_supported'):
-        bf16_support = torch.cuda.is_bf16_supported()
 if torch_available and hasattr(torch.version, 'hip') and torch.version.hip is not None:
     hip_version = ".".join(torch.version.hip.split('.')[:2])
 torch_info = {
     "version": torch_version,
-    "bf16_support": bf16_support,
```
stas00, line 244: as long as there is nothing relying on this field it should be safe to remove.

aphedges: In this repository, It was only used by a test, and I've replaced the usage in the test with a runtime check.

The usage in the test was only introduced 3 weeks ago in 7fc3065074885b0a2ec59c9c49e04296dd0e0fcd, so it probably hasn't been used much by other programs.

stas00: I totally agree, but let's see what the maintainers say.

stas00: as you pointed out the key here is that the hardware environment may chance since the installation and thus a memorized bf16 state can easily be invalid at run time. 

This would be especially true for cross-platform/cross-hardware prebuilds.

stas00: @jeffra, also there is a new development with IPEX - which now supports BF16 on CPU! So I highly recommend to already split up into 2 different state checks - have_bf16_cpu and have_bf16_gpu as we have just went through in transformers. https://github.com/huggingface/transformers/pull/17738

tests/unit/util.py
```diff
@@ -20,8 +20,8 @@ def bf16_required_version_check():
 
     if (TORCH_MAJOR > 1 or
         (TORCH_MAJOR == 1 and TORCH_MINOR >= 10)) and (CUDA_MAJOR >= 11) and (
-            NCCL_MAJOR > 2 or
-            (NCCL_MAJOR == 2 and NCCL_MINOR >= 10)) and torch_info['bf16_support']:
+            NCCL_MAJOR > 2 or (NCCL_MAJOR == 2 and NCCL_MINOR >= 10)
+        ) and torch.cuda.is_available() and torch.cuda.is_bf16_supported():
```
mrwyattii, line 24: Having a call to `torch.cuda.is_bf16_supported()` causes a problem with our unit tests (specifically the `distributed_test` decorator) because this function call initializes CUDA

setup.py
```diff
@@ -235,13 +234,10 @@ def create_dir_symlink(src, dest):
         nccl_version = ".".join(str(torch.cuda.nccl.version())[:2])
     else:
         nccl_version = ".".join(map(str, torch.cuda.nccl.version()[:2]))
-    if hasattr(torch.cuda, 'is_bf16_supported'):
-        bf16_support = torch.cuda.is_bf16_supported()
```
mrwyattii, line 239: The `bf16_support` value is only used in unit testing - perhaps we could add a try...except block around this?

mrwyattii (changes requested):
It is not necessary to have `bf16_support` defined in setup.py, but we cannot have it defined during unit testing either (see the error in[ torch-latest unit tests](https://github.com/microsoft/DeepSpeed/runs/7085395930?check_suite_focus=true))

aphedges:
> It is not necessary to have `bf16_support` defined in setup.py, but we cannot have it defined during unit testing either (see the error in[ torch-latest unit tests](https://github.com/microsoft/DeepSpeed/runs/7085395930?check_suite_focus=true))

Are you sure the CI failure is related to my changes? I can see the same error in other PRs, such as <https://github.com/microsoft/DeepSpeed/runs/7096559886?check_suite_focus=true> for #2064.

setup.py
aphedges: My first commit would be sufficient to prevent the crash because it checks whether CUDA is available.

mrwyattii: The bf16 support is only utilized in unit tests - so it shouldn't affect users that build/run on different hardware

mrwyattii:
> > It is not necessary to have `bf16_support` defined in setup.py, but we cannot have it defined during unit testing either (see the error in[ torch-latest unit tests](https://github.com/microsoft/DeepSpeed/runs/7085395930?check_suite_focus=true))
> 
> Are you sure the CI failure is related to my changes? I can see the same error in other PRs, such as https://github.com/microsoft/DeepSpeed/runs/7096559886?check_suite_focus=true for #2064.

You're right the current failures we're seeing on torch-latest is due to torch releasing 1.12 recently and it doesn't play well with our `distributed_test` decorator. We've disabled this workflow for now. However, please take a look at #1987 and #1990 where we moved bf16 support checking to `setup.py` for the reason of it causing issues during unit tests. We also added the torch-latest runner in #1990 as the tests that utilize bf16 are skipped on other runners.

setup.py
aphedges: I know it's set up that way in CI, but if a developer runs the unit tests on a different machine than the build, it could still cause problems.

aphedges:
> You're right the current failures we're seeing on torch-latest is due to torch releasing 1.12 recently and it doesn't play well with our `distributed_test` decorator. We've disabled this workflow for now. However, please take a look at #1987 and #1990 where we moved bf16 support checking to `setup.py` for the reason of it causing issues during unit tests. We also added the torch-latest runner in #1990 as the tests that utilize bf16 are skipped on other runners.

I hadn't read the explanation on #1990. Thanks for pointing it out!

As much I disagree that putting it in `setup.py` is the right place, I care the most about being able to compile `deepspeed` properly. Would you be fine with this PR if I drop my second commit, or do you want any other changes?

mrwyattii:
> As much I disagree that putting it in `setup.py` is the right place, I care the most about being able to compile `deepspeed` properly. Would you be fine with this PR if I drop my second commit, or do you want any other changes?

Yes, after dropping the second commit this should be good to go. And I agree that we should check for bf16 support (and perhaps nccl version too) at runtime - but we need to first resolve the issue we're having with `distributed_test` decorator. I think given that it's only used in unit tests, it's ok for now.

aphedges:
The failure in <https://github.com/microsoft/DeepSpeed/runs/7118253963?check_suite_focus=true> seems like a flaky performance test. DeepSpeed inference was around 10% slower than expected.

aphedges:
@mrwyattii, I force-pushed, and the checks passed this time. Can this PR be approved and merged?

PR #2050, opened 2022-06-24
Fix missing import in replace_module.py

aphedges:
The lack of this import is causing inference initialization to fail for me. It seems like it should have been added in 36ad3119d5420c6b25b456626a40f4ded8c05a81 when `dist` was introduced to `replace_module.py`, but it appears to have been missed. This fix also appears to have been included in the in-progress PR #2028, but I was hoping this problem could get fixed on `master` sooner than the time it will take for that more complex PR to be merged.

### DIFF
--- a/deepspeed/module_inject/replace_module.py
+++ b/deepspeed/module_inject/replace_module.py
@@ -7,6 +7,7 @@ from .replace_policy import replace_policies
 from ..constants import INFERENCE_GENERIC_MODE, INFERENCE_SPECIALIZED_MODE
 from ..runtime.weight_quantizer import WeightQuantization
 from torch import nn
+from torch import distributed as dist
 
 
 class LinearAllreduce(nn.Module):


deepspeed/module_inject/replace_module.py
```diff
@@ -7,6 +7,7 @@
 from ..constants import INFERENCE_GENERIC_MODE, INFERENCE_SPECIALIZED_MODE
 from ..runtime.weight_quantizer import WeightQuantization
 from torch import nn
+from torch import distributed as dist
```
Quentin-Anthony, line 10: We introduced a new DS comms interface in https://github.com/microsoft/DeepSpeed/commit/36ad3119d5420c6b25b456626a40f4ded8c05a81, and are moving away from direct `torch.distributed` imports unless necessary. 

Can you change this to: `from deepspeed import comm as dist`

aphedges: Sure! I noticed both in the codebase and wasn't sure which to do. I guess I chose incorrectly. I'll push a fix shortly.

Quentin-Anthony (approved):

Quentin-Anthony:
@jeffra -- This PR looks good to me.

PR #2051, opened 2022-06-24
Add flake8 to pre-commit checks

aphedges:
Back in #1602, I proposed adding flake8 checks, and @jeffra said it was a good idea. I've put in two PRs (#1601 and #2050) that likely would have been caught by flake8, so I decided to finally get some of the work I had done locally into a better state and into a PR.

~I don't have the testing infrastructure to run the full test suite locally, so I'll be using periodic pushes to this PR to fix problems as needed. Hopefully it doesn't take too long to complete.~ All CI now passes.

I changed a docstring but couldn't render the documentation to verify I didn't break anything. I think my almost complete lack of Ruby knowledge makes trying to fix the rendering errors not worth the effort.

I tried to fix as much code as I could, but I used `# noqa: F821` to suppress some errors so the check can be used going forward. Here are the locations of my usage and when the error was introduced: 
- `unfused_mode` was recently broken in 8164ea9e6d35dfbbe62b0f15448daf05268b9e81  (2022-05-24).
- `_dim_to_name` has always broken since it was added in 6996bb015911f5b437e2a732dd7f478f15abd167 (2021-09-21).
- `q_range` has always broken since it was added in ed3de0c21b1fea330de9c1a78a23ca33f340ef20 (2021-05-24).
- `profile` has always broken since it was added in a128f34e7d6404a18f00861a9c3b700c50e22010 (2021-04-07).
- `checkpoint`, `cuda`, and `resolved_archive_file`  have always broken since it was added in 734d8991c845664c5138a9f69704fd058a51b3c1 (2020-05-29).

None of these errors, which would have shown up if the code were ever run, are reported as issues, so I suspect the code is neither tested nor used. I suggest someone with more knowledge of these components look into fixes or remove what might be dead code.


mrwyattii:
@aphedges This is great - is it ready for review?

aphedges:
@mrwyattii, not yet. I've been rebasing a local version periodically to fix new breakages, but I haven't pushed it yet. I also have a comment to finish writing up because I found some bugs that I've suppressed warnings for in favor of getting the checks implemented. I expect to finish my improvements and set it ready for review later this week.

aphedges:
@mrwyattii, it's all ready! I've pushed my latest changes, updated the PR description with some additional information, and set the PR as ready for review.

deepspeed/runtime/fp16/fused_optimizer.py
```diff
@@ -181,7 +181,7 @@ def step_fused_adam(self, closure=None):
                                                      apply_scale=False)
 
         # Stash unscaled gradient norm
-        self._global_grad_norm = scaled_global_grad_norm / self.cur_scale
+        self._global_grad_norm = scaled_grad_norm / self.cur_scale
```
mrwyattii, line 184: Is this to fix a bug? why was this changed?

op_builder/builder.py
```diff
@@ -188,7 +187,6 @@ def installed_rocm_version():
         ROCM_MAJOR = '0'
         ROCM_MINOR = '0'
         if OpBuilder.is_rocm_pytorch():
-            from torch.utils.cpp_extension import ROCM_HOME
```
mrwyattii, line 191: I have a PR that uses this import (#2093). Could we preserve this? or would that cause the precommit to fail?

tests/unit/modeling.py
```diff
@@ -216,9 +213,7 @@ def __init__(self,
         self.out_features = out_features
         self.fused_gelu = False
         self.fused_tanh = False
-        if isinstance(act,
-                      str) or (sys.version_info[0] == 2 and isinstance(act,
-                                                                       unicode)):
+        if isinstance(act, str):
```
mrwyattii, line 216: Why was the or statement removed here?

tests/unit/test_checkpointing.py
```diff
@@ -1353,7 +1345,6 @@ def _test_load_immediate_save(args, model, tmpdir):
 @pytest.mark.parametrize('zero_stage', [0, 1, 2, 3])
 def test_save_before_accum_grad_is_done(tmpdir, zero_stage):
     config_dict = {
-        "train_batch_size": 4,
```
mrwyattii, line 1356: Why was the batch size config removed here?

mrwyattii (commented):
This mostly looks good to me. I left some comments for changes that I'm unsure about if you could help me understand.

There are quite a few added `#noqa` for `F401` and `F821` - is this normal to need to add these when bringing flake into a project? I'd like to better understand why they are necessary.

deepspeed/runtime/fp16/fused_optimizer.py
aphedges: `scaled_global_grad_norm` was undefined. However, there is a `scaled_grad_norm = get_global_norm(norm_list=norm_groups)` a couple of lines up, so I assumed that's what was intended instead.

op_builder/builder.py
aphedges: I noticed it, but it would cause precommit to fail. If you get that PR merged, I can rebase and keep it without the failure.

tests/unit/modeling.py
aphedges: `unicode` is undefined when running on Python 3 (it's Python 2-only), and this project uses f-strings in `setup.py`, so it probably can't even build wheels for anything below Python 3.6. This check is therefore useless and removing it is likely better than suppressing the flake8 error.

tests/unit/test_checkpointing.py
aphedges: This violated F601, which is having a duplicate `dict` key. The value is redeclared later on, so I chose to remove this one to preserve the current behavior.

aphedges:
> There are quite a few added `#noqa` for `F401` and `F821` - is this normal to need to add these when bringing flake into a project? I'd like to better understand why they are necessary.

I personally think it depends on the codebase, and the reason for locally suppressing the two checks are different:

The usages of `# noqa: F821` ("undefined name `name`") are discussed in the main comment (<https://github.com/microsoft/DeepSpeed/pull/2051#issue-1284360635>). In short, I decided to suppress some of these flake8 errors for undeclared variables that I was unable to fix myself. I was able to fix most of them (e.g., in here and in #2050), but I'd rather someone with more knowledge of the codebase fix these. They all involve actual code that currently will fail at runtime but probably hasn't been run in months or years, and the intended behavior is not clear to me from the surrounding code or the commit history. Please see the linked comment for a more detailed explanation.

I needed to refresh myself about my usage of `# noqa: F401` ("`module` imported but unused"). The check is to remove unused imports, which in addition to being messy, can cause bugs like the one I fixed in #1601. In all but one of these cases, the import is used only to check whether a module can be imported, even though the import isn't used. There might be a better way to do this, but I'm not knowledgeable enough about the import system to do it myself. The only other case is special: <https://github.com/microsoft/DeepSpeed/blob/bb1836a2f1868441323f32c5b7cb3a2fbcb34220/deepspeed/git_version_info.py#L3> The identifiers imported from `.git_version_info_installed.py` are only used for re-exporting from `git_version_info.py`. Most of these are the sole statement in a `try` block, and the pattern is common enough in this project and others that it might be worth changing flake8 itself, but that seems like a bit much. I definitely removed many more unused imports than I needed to add `# noqa F401`.

I also suppressed some checks in the pre-commit configuration itself. Some, like the `E` and `W` families, are mostly for code formatting. They would take some more work that I don't think is worth it, given that my focus here was on fixing current and preventing future bugs. In addition, this repository already uses `yapf` for code formatting checks, so they might have been redundant or contradictory. Other checks, such as `F541` ("f-string without any placeholders") and `F841` ("local variable `name` is assigned to but never used"), might be related to bugs or cause wasted resources, but they were too much effort for me to personally fix. There is also a risk of false positives in some cases, hence why I ignored unused imports in `__init__.py`, where such a construct is often intentional.

I don't intend this to be an absolutely comprehensive PR for flake8 integration. This is just meant to be a step to fix some problems and prevent the introduction of some more in the future. Other people can improve on the work I've done here if more or better checks are desired.

aphedges:
@mrwyattii, just following up on this. This kind of PR is annoying to rebase, so I'd like it get merged before one of the larger PRs makes it in.

op_builder/builder.py
aphedges: Thank you! Please let me know when it's merged so I can rebase and push the adjusted version of this PR.

jeffra:
Hi @aphedges, thank you so much for doing this PR! I am looking it over now and overall it looks good, but just trying to make sure there aren't any subtle issues since it touches so many files. Let's try and merge it in this week but there are other PRs coming that we're working on today.

aphedges:
I completely understand that you want to carefully review this. I'll keep updating it as new PRs make it to `master`. For example, aa88137b8d093f222f6c925d70b68024ae26ef0c introduced some new breakages, and I just pushed my rebased branch with fixes in a separate commit. Part of the reason I was hoping this could get merged sooner rather than later is that I need to do some guesswork about some of the breakages that frequently happen in the very large commits. There are multiple undeclared variables in `deepspeed/module_inject/replace_module.py` that I hope I fixed correctly, but I don't really know if I did.

aphedges:
I have rebased against `master` again to fix the merge conflict and fixed the new flake8 error introduced in ee7ea3b8055283990c782f89f8672cbfbcfbe1de.

aphedges:
I have rebased against `master` again to fix the merge conflict and the new flake8 errors introduced in 0f4f2f982c3b55e61d2d7d0631e01050316c9aca. The commit also removed two errors I was suppressing, so it wasn't all bad.

aphedges:
I have rebased against master again to fix the merge conflict and the new flake8 errors introduced in 69b7c970d262bdc248d87061ceb2c3431c6dbd78. The commit also introduced an error (calling `Path.isfile()` instead of [`Path.is_file()`](https://docs.python.org/3.10/library/pathlib.html#pathlib.Path.is_file)) that wasn't caught by flake8 but was caught by PyCharm.

op_builder/builder.py
aphedges: I've rebased against your now-merged PR. See <https://github.com/microsoft/DeepSpeed/pull/2051#issuecomment-1189682379> for more information.

aphedges:
I have fixed the new merge conflict caused by 844d9f31a98165d445cbf8ae4f5ded0a82061399.

aphedges:
@jeffra, have you gotten around to reviewing yet?

aphedges:
I have rebased against `master` again to fix the merge conflict and the new flake8 errors introduced in 80d0a32f0bf2f3974a13e02ff21addabbcc37734.

aphedges:
I have rebased against `master` again to fix the merge conflict and the new flake8 errors introduced in 8413b7f83d311a0928b2f240f19f234de293478f and 5349347bb648ce9eb84b1d78d77e50968f299b62. The latter fix was more complex, so I needed to do some guessing and suppress an error I could not figure out how to fix myself.

The sooner this PR gets merged, the sooner PR authors can fix these mistakes themselves that will often cause their code to fail at runtime.

aphedges:
There was a test failure but I couldn't rerun it myself, so I rebased and force-pushed.

jeffra (approved):
Thank you @aphedges for all your hard work on this! It's very much appreciated :)

aphedges:
@jeffra, thanks for merging it!

I needed to suppress some errors for undeclared variables I could not fix myself. All of the lines would cause a crash if they are run. Should I make issues for them so they aren't just lost in this PR? They are all bugs waiting to happen.

PR #2053, opened 2022-06-25
Fix conflict between Tutel and top-2 gate in MoE layer

yetiansh:
Previously, using both Tutel optimization and top-2 gating in MoE model training would fail. If we enable both Tutel and top-2, `MoELayer` would try to unpack top-2 gate's output at [here](https://github.com/microsoft/DeepSpeed/blob/7bae53d154f44877d17c9c5efe40b144440272bf/deepspeed/moe/sharded_moe.py#L502), which would fail because top-2 gate does not produce these number of outputs.

Fix by checking the gate's type when constructing `MoELayer`.


awan-10:
Thank you for the PR @yetiansh :) It looks good to me. Alex had added tutel support so let me tag him and ask for a quick review.

@alexandremuzio - can you please review this real quick?

awan-10:
@yetiansh - can you please follow the guide here and update your PR? I see its failing in format checks.

https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md

alexandremuzio:
Looks good to me. Thanks!

yetiansh:
Thanks @alexandremuzio @awan-10. I've run the `pre-commit` and it looks like running format checking workflow needs your approval.

yetiansh:
Hi, is this PR still active? @awan-10 @alexandremuzio

awan-10:
Sorry for the delay in getting back @yetiansh. I approved this PR so tests can run. Will merge it as soon as tests pass. Thank you!

awan-10 (approved):

deepspeed/moe/sharded_moe.py
```diff
@@ -474,13 +474,17 @@ def __init__(self,
...
-        self.use_tutel = use_tutel and TUTEL_INSTALLED
+        self.use_tutel = use_tutel and TUTEL_INSTALLED and gate.k == 1
 
         if self.use_tutel:
             logger.info('Using Tutel optimizations.')
         elif use_tutel and not TUTEL_INSTALLED:
             logger.warning("Tutel optimization requested but not installed. "
                            "Proceeding without Tutel.")
+        elif use_tutel and TUTEL_INSTALLED and gate.k != 1:
+            logger.warning(
```
jeffra, line 485: Can we wrap this in a `if torch.distributed.get_rank() ==0:`?

jeffra (approved):

deepspeed/moe/sharded_moe.py
yetiansh: Yeah it is possible. But I wonder should we also wrap other warnings and infos? For example, L480 and L482-483?

PR #2058, opened 2022-06-27
Wrong partition_id while copying fp32_params -> fp16 params in Z2 for MoE

siddharth9820:
Post the optimizer step, Z2 uses incorrect partition ids for copying updated fp-32 params to their fp-16 counterparts. This leads to only a fraction of the non-expert parameters being trained. For reference, I have the loss curves before and after the bug fix.

Base Model - 1.3B
Number of Experts - 8
Batch Size - 256
Machine - Azure A100 40GB
Number of GPUs - 8
Dataset - BookCorpus

image

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1762,7 +1762,8 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
                 for group in self.single_partition_of_fp32_groups:
                     group.grad = None  # in step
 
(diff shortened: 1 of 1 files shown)


siddharth9820:
Here's the comparison with ZeRO disabled.
image

jeffra (approved):
Excellent catch @siddharth9820, we discussed offline the existing issue w. the offload code path as well. We discussed this as a follow up PR though.

siddharth9820:
A separate PR coming soon for the offload fix. Stay tuned!

PR #2074, opened 2022-07-06
Tensor parallelism for Mixture of Experts

siddharth9820:
Note: This PR is in conjunction with this [PR](https://github.com/microsoft/Megatron-DeepSpeed/pull/63) on the Megatron-Deepspeed repo.

This PR adds tensor parallelism for non-experts. This combined with ZeRO-2 allows us to scale to roughly 2x larger base models than ZeRO-2. When tensor parallelism is enabled only for non-experts, there are duplicate tokens at each gate. It is important to drop the duplicates before they reach the experts, otherwise we run into convergence issues. In the current implementation, we drop tokens right before the AlltoAll and gather them right after the AlltoAll. These calls are done in `sharded_moe.py`

Update: This PR now supports tensor parallelism for experts as well.


siddharth9820:
Comparing loss curves with no tensor parallelism

image

deepspeed/moe/layer.py
```diff
@@ -54,6 +55,7 @@ def __init__(self,
         self.ep_size = min(
             ep_size,
             num_experts)  # the ep size should be less than the number of experts
+        self.MoE_mp_size = MoE_mp_size
```
awan-10, line 58: MoE_mp_size --> moe_mp_size

deepspeed/moe/layer.py
```diff
@@ -94,7 +96,7 @@ def _create_process_groups(self):
             print(
                 f"No existing process group found, creating a new group named: {self.expert_group_name}"
             )
-            if groups.mpu is None:
+            if self.MoE_mp_size == 1:
```
awan-10, line 99: if groups.mpu is None or self.moe_mp_size == 1:

siddharth9820:
I have also added a switch to enable tensor parallelism for experts as well. 
image

siddharth9820:
A larger run
image

deepspeed/moe/layer.py
```diff
@@ -94,9 +97,11 @@ def _create_process_groups(self):
             print(
                 f"No existing process group found, creating a new group named: {self.expert_group_name}"
             )
-            if groups.mpu is None:
+            if (groups.mpu is None) or (not self.enable_expert_tensor_parallelism):
+                # no mpu implies no tensor parallelism
```
siddharth9820, line 101: update this comment

deepspeed/utils/groups.py
```diff
@@ -237,11 +243,10 @@ def _create_expert_data_and_model_parallel(expert_parallel_size_, mpu):
     _DATA_PARALLEL_GROUP = mpu.get_data_parallel_group()
     _MODEL_PARALLEL_GROUP = mpu.get_model_parallel_group()
 
+    group_name = f"ep_size_{expert_parallel_size_}"
```
siddharth9820, line 246: double check with pyramid moe

deepspeed/utils/mappings.py
```diff
@@ -0,0 +1,72 @@
+import torch
+import deepspeed
+
+
+def _gather_tokens(input_, dim=0):
+    """Gather tensors and concatinate them along a dimension"""
```
conglongli, line 6: concatenate? Better to do a spelling check.

siddharth9820: yep, the comments can be all over the place. I'll fix them.

deepspeed/utils/mappings.py
```diff
@@ -0,0 +1,72 @@
...
+    total_chunks = mpu.get_tensor_model_parallel_world_size()
+    this_chunk = mpu.get_tensor_model_parallel_rank()
+    assert input_.shape[dim] % total_chunks == 0, f"input dimension {dim} ({input_.shape[dim]}) is not divisible by tensor parallel world size ({total_chunks})"
+    chunk_size = input_.shape[dim] // total_chunks
+
+    return torch.narrow(input_, dim, this_chunk * chunk_size, chunk_size)
+
+
+class _GatherTokens(torch.autograd.Function):
+    """Reduce scatter output of self attention for MoE"""
```
siddharth9820, line 43: fix this comment

deepspeed/moe/sharded_moe.py
```diff
@@ -519,6 +520,10 @@ def forward(self, *input: Tensor, **kwargs: Any) -> Tensor:
         if self.wall_clock_breakdown:
             self.timers('falltoall').start()
 
+        if groups._get_expert_model_parallel_world_size() == 1:
+            # no slicing for experts means we need to drop tokens
```
conglongli, line 524: Can we have a bit more detailed explanation here? To me "no expert slicing" has no direct relationship with "token dropping". And also good to explain how it's different from the drop_tokens in the gating functions in the same file.

deepspeed/utils/mappings.py
```diff
@@ -0,0 +1,72 @@
...
+        ctx.dim = dim
+        return _gather_tokens(input_, dim)
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        return _drop_tokens(grad_output, ctx.dim), None
+
+
+class _DropTokens(torch.autograd.Function):
+    "Drop tokens (this is a hacky approach until we can do reduce scatter)"
```
siddharth9820, line 59: fix this comment

siddharth9820:
after fixing local number of experts for full tensor parallelism, the convergence issue seems to be gone.
image

conglongli (approved):
Tested that both non-expert MP and expert MP work fine. Still need a full code review from @awan-10

awan-10 (approved):

deepspeed/utils/mappings.py
```diff
@@ -0,0 +1,95 @@
+'''
```
jeffra, line 1: If this is only used for MoE can we move it to `deepspeed/moe`?

jeffra:
Let's add some functional unit tests to ensure the new code paths are triggered in our tests. This will ensure things are at least functionally working in the future. Would be great to have basic correctness unit tests as well but we can discuss that offline.

deepspeed/utils/mappings.py
siddharth9820: Yes, it's only used for MoEs. I'll move this file.

siddharth9820:
@jeffra I have added some tests in `tests/unit/test_moe_tp.py`. Can you please check if they are good enough for now?

tests/unit/test_moe_tp.py
```diff
@@ -0,0 +1,160 @@
...
+from .simple_model import args_from_dict
+from .util import required_torch_version
+from deepspeed.moe.layer import MoE
+
+try:
+    from apex import amp  # noqa: F401
+    _amp_available = True
+except ImportError:
+    _amp_available = False
+amp_available = pytest.mark.skip(_amp_available, reason="apex/amp is not installed")
```
jeffra, line 14: Let's remove this, it doesn't appear your test uses `amp`.

jeffra (approved):

jerryli1981:
> Comparing loss curves with no tensor parallelism

Hi, I tested the tensor parallelism for MoE, the loss curves still higher and then NAN after 600 millions of tokens

PR #2085, opened 2022-07-11
Trajepl/nebula ckpt engine

trajepl:


trajepl:
> It feels like to me that the `NebulaCheckpointEngine` should be part of the Nebula library itself. Is it possible to make Nebula APIs align more with torch save/load APIs? Then on the DeepSpeed or any other Nebular user will require less code change to support this feature.

Currently, it is hard to make nebula save/load api totally same with torch save/load, as nebula relys on the tag to build up the concept of checkpoint to manager the saved files.

deepspeed/runtime/engine.py
```diff
@@ -2289,7 +2299,8 @@ def load_moe_state_dict(checkpoint_path,
                             old_moe_load,
                             model=None,
                             mpu=None,
-                            num_experts=1):
+                            num_experts=1,
+                            checkpoint_engine=CheckpointEngine()):
```
tjruwase, line 2303: Why not pass self.checkpoint_engine instead of CheckpointEngine()?

deepspeed/runtime/engine.py
```diff
@@ -2444,7 +2455,9 @@ def load_checkpoint(self,
                         load_optimizer_states=True,
                         load_lr_scheduler_states=True,
                         load_module_only=False,
-                        custom_load_fn=None):
+                        custom_load_fn=None,
+                        enable_nebula_load=True,
```
tjruwase, line 2459: Will this not break backwards-compatibility by forcing existing load_checkpoint() calls in clients to use nebula? Can you clarify why nebula_* args need to be added to this function?

deepspeed/runtime/engine.py
```diff
@@ -2467,6 +2480,12 @@ def load_checkpoint(self,
         ``load_checkpoint()`` wants a pristine model. If insisting to do so, please reinitialize engine
         before ``load_checkpoint()``.
         """
+        checkpoint_engine_tmp = self.checkpoint_engine
+        if not self.config is None and \
+                self._config.nebula_config.enabled and \
+                enable_nebula_load == False:
+            self.checkpoint_engine = CheckpointEngine()
```
tjruwase, line 2487: Can you please explain why `self.checkpoint_engine` is reinitialized here in addition to the initialization in DeepSpeedEngine constructor?

deepspeed/runtime/engine.py
```diff
@@ -2467,6 +2480,12 @@ def load_checkpoint(self,
         ``load_checkpoint()`` wants a pristine model. If insisting to do so, please reinitialize engine
         before ``load_checkpoint()``.
         """
+        checkpoint_engine_tmp = self.checkpoint_engine
```
tjruwase, line 2483: I don't think this temporary swap of checkpoint_engine is needed here since NebulaCheckpointEngine will be initialized by the client through deepspeed configuration. Is that correct?

deepspeed/runtime/checkpoint_engine/nebula_checkpoint_engine.py
```diff
@@ -0,0 +1,64 @@
...
+                return None
+        else:
+            checkpoint = torch_nebula.get_checkpoint(tag=tag,
+                                                     persist_path=self.nebula_load_path)
+        partition = checkpoint.load(partititon_name, map_location=map_location)
+        logger.info(
+            f"[Nebula] Loaded {path} under tag{tag} from {self.nebula_load_path}.")
+        return partition
+
+    def commit(self, tag):
```
tjruwase, line 61: `commit` seems to be nebula-specific and not needed by other checkpoint engines, such as `torch.save`. Is this correct?

deepspeed/runtime/state_dict_factory.py
```diff
@@ -25,19 +28,24 @@ def get_sd_loader_json(json_file):
             return SDLoaderFactory.get_sd_loader(ckpt_list, sd_type, version)
 
     @staticmethod
-    def get_sd_loader(ckpt_list, sd_type='Megatron', version=None):
+    def get_sd_loader(ckpt_list,
+                      sd_type='Megatron',
+                      version=None,
+                      checkpoint_engine=None):
```
tjruwase, line 34: Make `checkpoint_engine` required.

deepspeed/runtime/state_dict_factory.py
```diff
@@ -25,19 +28,24 @@ def get_sd_loader_json(json_file):
...
         if sd_type == 'Megatron':
-            return MegatronSDLoader(ckpt_list, version)
+            return MegatronSDLoader(ckpt_list, version, checkpoint_engine)
         else:
             assert False, '{} checkpoint type is not supported'.format(sd_type)
 
 
 class SDLoaderBase(ABC):
-    def __init__(self, ckpt_list, version):
+    def __init__(self, ckpt_list, version, checkpoint_engine=None):
```
tjruwase, line 42: Make `checkpoint_engine` required. This will help to avoid unexpected creation of CheckpointEngine objects at runtime.

deepspeed/runtime/state_dict_factory.py
```diff
@@ -25,19 +28,24 @@ def get_sd_loader_json(json_file):
...
         self.module_key = None
         self.ckpt_list = ckpt_list
-        self.check_ckpt_list()
         self.version = version
+        self.checkpoint_engine = CheckpointEngine(
```
tjruwase, line 46: Simplify by making `checkpoint_engine` a required argument.

deepspeed/runtime/state_dict_factory.py
```diff
@@ -193,8 +204,8 @@ def sanity_check(self, ckpt_file_name):
 
 
 class MegatronSDLoader(SDLoaderBase):
-    def __init__(self, ckpt_list, version):
-        super().__init__(ckpt_list, version)
+    def __init__(self, ckpt_list, version, checkpoint_engine=None):
```
tjruwase, line 207: Make `checkpoint_engine` required argument.

tjruwase:
@trajepl, thanks for this PR. Could you please consider linking a tutorial?

deepspeed/runtime/checkpoint_engine/checkpoint_engine.py
```diff
@@ -0,0 +1,23 @@
...
+        logger.info(f"Saved {path}.")
+        return None
+
+    def load(self, path: str, map_location=None):
+        logger.info(f"Loading checkpoint from {path}...")
+        partition = torch.load(path, map_location=map_location)
+        logger.info(f"Loaded checkpoint from {path}.")
+        return partition
+
+    def commit(self, tag):
```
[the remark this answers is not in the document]
trajepl: It plays the role to tell main process current checkpoint is saved completely. For raw torch, it will be same as following code.
```python
        if save_latest and self.global_rank == 0:
            with open(os.path.join(save_dir, 'latest'), 'w') as fd:
                fd.write(tag)
```

Nebula need this signal to start the persistence processing for checkpoint.

deepspeed/runtime/engine.py
trajepl: It is class method where we can not get the variables from instance.
```python
    @staticmethod
    def load_moe_state_dict(checkpoint_path,
                            tag,
                            state_dict,
                            old_moe_load,
                            model=None,
                            mpu=None,
                            num_experts=1,
                            checkpoint_engine=CheckpointEngine()):
```

deepspeed/runtime/engine.py
trajepl: Please refer to the above case #1.
customer want to reuse the existing checkpoints saved by deepspeed, then use nebula for saving.

deepspeed/runtime/engine.py
tjruwase: Fair point. But right now, this will default to torch checkpoint engine, correct?

deepspeed/runtime/checkpoint_engine/nebula_checkpoint_engine.py
tjruwase: Yes, it is fine to add a `commit` API. I just wanted to understand what it means.

trajepl:
> @trajepl, thanks for this PR. Could you please consider linking a tutorial?
Here is the nebula wiki, please feel free to leave your comments:
https://gentle-moss-0d9631610.1.azurestaticapps.net/tutorial.html

deepspeed/runtime/checkpoint_engine/checkpoint_engine.py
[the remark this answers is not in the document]
trajepl: image
Actually for nebula.save, we have the following steps to persist a checkpoint into disk:
1. **tier1**: we pin the shared memory and write the checkpoints files into shared memory(let say under /dev/shm/). It is a synced way which will block the training.
2. **tier3**: when nebule service received the `commit` signal from SDK, nebula service will start a persistent process which move the checkpoints from tier1 to tier3(equal to `nebula_load_path_tier3` where you commented). After tier3 done, the nebula meta info will be wrote into tier3 path.

So for your questions:
> For nebula, do we need a single commit for each checkpoint file or just one for the save_checkpoint() call?
    Nebula needs every training process to call `commit` to tell the services current checkpoint can be moved to tier3. But origin torch only write the meta file for checkpoint management.

>  what does the return value mean?
Emmm, it is dummy return for torch I think. If needed, maybe we can move the torch meta file operation into this method only for rank0? What do you think? :)

deepspeed/runtime/engine.py
trajepl: Yep~

deepspeed/runtime/checkpoint_engine/checkpoint_engine.py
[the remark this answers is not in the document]
trajepl: One thing I want to call out is that: the files structures are totally different for original torch path and nebula tier3 path.

deepspeed/runtime/engine.py
trajepl: Nope, it will not work. For case 2/3, it is doable in the side of 'initialize()'. But for case1, we must to have torch and nebula checkpoint engine ready to call `torch.load` and `nebula.save`.

It is hard for nebula to directly load the files saved by torch. Nebula maintained the customized checkpoint management services. 

Also, in real usage, if case 1/2/3 happen at one training process, it will be hard to directly call `deepspeed.initialize()` to make it ready.

deepspeed/runtime/checkpoint_engine/checkpoint_engine.py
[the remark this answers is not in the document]
tjruwase: It is fine for the checkpoint file structure to be differerent between torch and nebula tier3, it simply means that `torch.load()` cannot be used for nebula checkpoints. This is not a concern for deepspeed. 

I think the `fd.write(tag)` should remain inside deepspeed and outside nebula for a number of reasons: (1) it is not currently created by torch since it is not checkpoint state, but metadata, (2) It is used by deepspeed and existing clients for checkpoint management, and (3) it is very small data and fast to write. Typically, it is used to identify the latest checkpoint file in a folder of many checkpoint files, but clients can also use tags in other ways. So, can you explain why this part has to change for nebula?

deepspeed/runtime/engine.py
tjruwase: In that case, can the nebula checkpoint engine not be initialized for different load modes: (i) load with torch and (ii) load with nebula. This mode can be configured in the config. This will hide the complexity in the nebula checkpoint engine. How about this approach?

tjruwase: Can you describe a situation, where cases 1, 2, and 3 will happen in a single training run?

tjruwase:
> > @trajepl, thanks for this PR. Could you please consider linking a tutorial?
> > Here is the nebula wiki, please feel free to leave your comments:
> > https://gentle-moss-0d9631610.1.azurestaticapps.net/tutorial.html

Thanks for sharing the tutorial. I did not see the use of `commit()` in the quick start examples, perhaps I just missed it?

deepspeed/runtime/checkpoint_engine/checkpoint_engine.py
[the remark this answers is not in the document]
trajepl: It makes sense to keep `fd.write(tag)` out of nebula. Actually we did that in code as well.
```python
# Save latest checkpoint tag
dist.barrier()
self.checkpoint_engine.commit(tag)
if save_latest and self.global_rank == 0:
    with open(os.path.join(save_dir, 'latest'), 'w') as fd:
        fd.write(tag)
```
Then for:
1. `torch`: the `commit` was used to align the API. No extra efforts for condition(`if nebula else...`).
2. `nebula`: the `commit` will send the signal to nebula services to start tier3.

The `fd.write(tag)` is workable for torch, but cannot replace `commit()` to tell nebula service current checkpoint is saved.

So for you question:  
> why this part has to change for nebula?
We need the `commit()` for nebula when all files under a tag are saved. But it will just return dummy true for `torch`.

deepspeed/runtime/engine.py
trajepl: > Can you describe a situation, where cases 1, 2, and 3 will happen in a single training run?

I were thinking that the case where:
1. User want to load back the checkpoint saved by deepspeed when the job starts. (_case1_)
2. Let us say, after several steps(global_step1000), user's workflow will call `load_checkpoint` to load nebula checkpoint back for testing/validation. or something else, just to verify the status of saved checkpoints(_case2/3_).

In this case, the config would not work as it is one-off setting for training.
**But user must to change the code when they change the case from 1 to 2 or other cases.** :( . And the config can avoid this inconvenience. 

> In that case, can the nebula checkpoint engine not be initialized for different load modes: (i) load with torch and (ii) load with nebula. This mode can be configured in the config. This will hide the complexity in the nebula checkpoint engine. How about this approach?

Continue on the last one comment. Maybe we can assume user only `load_checkpoint` in a consistent case for a training. If so, as your suggestions, it would be better to put the arguments into configs. Any other concerns? If not, I can help move them into config. @tjruwase

trajepl:
> > > @trajepl, thanks for this PR. Could you please consider linking a tutorial?
> > > Here is the nebula wiki, please feel free to leave your comments:
> > > https://gentle-moss-0d9631610.1.azurestaticapps.net/tutorial.html
> 
> Thanks for sharing the tutorial. I did not see the use of `commit()` in the quick start examples, perhaps I just missed it?

Oh, sorry. My bad. The tutorial shows the case where user know the concrete numbers for files user want to save(`tn.Checkpoint('global_step10', num_of_files=100)`: 100 files will be saved).

But in most cases, users don't know the file numbers, they can init the checkpoint class by: `tn.Checkpoint('global_tag100', num_of_files=-1)`.

It will be like for commit mode:
```python
        checkpoint = tn.Checkpoint(tag=global_tag, num_of_files=-1)
        print("tn.Checkpoint finish")
        checkpoint.save(partition_name, sd)

        print("checkpoint.save finish")

        checkpoint.commit()
        print("checkpoint.commit finish")
```

deepspeed/runtime/pipe/module.py
```diff
@@ -563,7 +564,7 @@ def ckpt_layer_path_list(self, ckpt_dir, local_layer_idx):
         ckpt_files.sort()
         return ckpt_files
 
-    def save_state_dict(self, save_dir):
+    def save_state_dict(self, save_dir, checkpoint_engine=None):
```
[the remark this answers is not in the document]
trajepl: Fixed.

deepspeed/runtime/pipe/module.py
```diff
@@ -584,9 +585,12 @@ def save_state_dict(self, save_dir):
...
                  for k,
                  v in orig_state_dict.items()})
-            torch.save(final_state_dict, model_ckpt_path)
+            checkpoint_engine.save(final_state_dict, model_ckpt_path)
 
-    def load_state_dir(self, load_dir, strict=True):
+    def load_state_dir(self,
+                       load_dir,
+                       strict=True,
+                       checkpoint_engine=CheckpointEngine()):
```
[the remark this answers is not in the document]
trajepl: Fixed.

deepspeed/runtime/state_dict_factory.py
trajepl: Fixed.

deepspeed/runtime/state_dict_factory.py
trajepl: Fixed.

deepspeed/runtime/state_dict_factory.py
trajepl: Fixed.

deepspeed/runtime/engine.py
tjruwase: Thanks for explaining all the considerations. I think for now moving to config is sufficient. We can always address other user scenarios in the future when they come up. But for now, I think the simplicity of hiding the arguments in the config is priority.

deepspeed/runtime/engine.py
tjruwase: I want to confirm that after moving the args to config as discussed that this change will go away.

deepspeed/runtime/engine.py
tjruwase: This change should also go away once the args are moved to config to allow loading a non-nebula checkpoint, right?

deepspeed/runtime/checkpoint_engine/checkpoint_engine.py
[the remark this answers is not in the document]
tjruwase: > It makes sense to keep `fd.write(tag)` out of nebula. Actually we did that in code as well.

Sorry, I misread the PR. You are correct that you did not change this code.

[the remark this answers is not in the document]
tjruwase: > Then for:
> 
> 1. `torch`: the `commit` was used to align the API. No extra efforts for condition(`if nebula else...`).
> 2. `nebula`: the `commit` will send the signal to nebula services to start tier3.
> 
> The `fd.write(tag)` is workable for torch, but cannot replace `commit()` to tell nebula service current checkpoint is saved.
> 
> So for you question:
> 
> > why this part has to change for nebula?
> > We need the `commit()` for nebula when all files under a tag are saved. But it will just return dummy true for `torch`.

So, this brings up the importance of clarifying the semantics of nebula and other checkpoint engines. I don't know if there is standard terminology for these concepts. But it seems to me that one key question is whether persistence is implicit or explicit. Persistence with `torch.save()` is implicit because checkpoint files are persisted on disk on return, while nebula is explicit because an additional call is required to make it persist the files. I think this is an important consideration for designing the general CheckpointEngine abstraction. We can discuss this further offline.

deepspeed/runtime/engine.py
trajepl: Yep, you are right. Removed in the latest commit.

deepspeed/runtime/engine.py
trajepl: Yep. Thanks for confirming. Removed now.

tjruwase (approved):

jeffra (approved):

PR #2089, opened 2022-07-12
Release swap buffers for persisted params

tjruwase:
Avoid hogging swap buffers of persisted params since they will never be swapped out. 
Option to limit number of parameters pinned in GPU memory. 
Minor logging tweaks.


deepspeed/runtime/engine.py
```diff
@@ -1406,10 +1410,8 @@ def _configure_zero_optimizer(self, optimizer):
 
         elif zero_stage == ZeroStageEnum.weights:
             assert not self.has_moe_layers, "MoE not supported with Stage 3"
-            logger.info("Initializing ZeRO Stage 3") if dist.get_rank() == 0 else None
-            from deepspeed.runtime.zero.stage3 import DeepSpeedZeroOptimizer_Stage3
-
             if isinstance(optimizer, DummyOptim):
+                logger.info("Creating ZeRO Offload") if dist.get_rank() == 0 else None
```
mrwyattii, line 1414: Can we replace this with `log_dist` (and maybe the logger call on [line 302](https://github.com/microsoft/DeepSpeed/blob/49c1f78ddcbb794acb2f1fbb4c3d1866ff0d431c/deepspeed/runtime/engine.py#L302) as well)

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -669,9 +669,13 @@ def get_model():
 
         # Remote device is the device where parameter partitions are stored
         # It can be same as local_device or it could be CPU or NVMe.
-        self.remote_device = self.local_device if remote_device is None else remote_device
-        self.pin_memory = pin_memory if (self.remote_device
-                                         == OffloadDeviceEnum.cpu) else False
+        self.remote_device = self.local_device if remote_device in [
+            None,
```
mrwyattii, line 673: Could we make the default of `remote_device` be `OffloadDeviceEnum.none`? Or do we need to support the option to pass `None`?

mrwyattii (changes requested):

deepspeed/runtime/zero/partition_parameters.py
tjruwase: Unfortunately, the caller could be outside deepspeed, e.g., Megatron, HF and other integrations, and `None` is a convenient default value for those cases.

deepspeed/runtime/engine.py
tjruwase: Done

mrwyattii (approved):

PR #2091, opened 2022-07-13
fixed "None type has no len()"

xiazeyu:
When using autotunning to calculate throughput with `transformer < 4.16.2`, It breaks with `None type has no len()`.

It seems `elapsed_records` is still `None` when being called at the first run.

This modification helps to save elapsed to `self.elapsed_records`, so that it won't throw an exception.


cli99:
@xiazeyu, thanks for catching this bug!

Issue #2080, opened 2022-07-08
[BUG] can't install with ROCm


stalkerg:
It happened just during `pip install` for deepspeed-0.6.5:

```
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [8 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/tmp/pip-install-ruwdavjo/deepspeed_caadfe808785436583e97206db673e2c/setup.py", line 39, in <module>
          rocm_version = OpBuilder.installed_rocm_version()
        File "/tmp/pip-install-ruwdavjo/deepspeed_caadfe808785436583e97206db673e2c/op_builder/builder.py", line 192, in installed_rocm_version
          with open('/opt/rocm/.info/version-dev', 'r') as file:
      FileNotFoundError: [Errno 2] No such file or directory: '/opt/rocm/.info/version-dev'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed
```

I have ROCm but it is installed into the system, not into `/opt/rocm` and I have no `version-dev` file. ROCm PyTorch is working fine. 
It's Gentoo Linux.

mrwyattii:
@stalkerg thanks for bringing this to our attention. It seems that we expect ROCm to be installed at `/opt/rocm/` [here](https://github.com/microsoft/DeepSpeed/blob/73b5d9833a113c47758fbe9f8538614ce34ea95e/op_builder/builder.py#L192)

I'll look into how we can avoid assuming the install location and add a fix. In the meantime, if you modify the path on the file:line I linked to and point it to your rocm install location, it should fix the problem.

mrwyattii:
@stalkerg could you please try again with #2093 ?
```
git clone https://github.com/microsoft/DeepSpeed.git && cd DeepSpeed
git pull origin pull/2093/head
pip install .
```

stalkerg:
Thanks! But looks like it will not work because I am installing https://pytorch.org/get-started/locally/ 
```
pip3 install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/rocm5.1.1
```
and it contains ROCm libraries but without `.info/version-dev` file or any `dev` or `debug` information. I also have no ROCM_HOME. 
[...]

PR #2093, opened 2022-07-13
Remove hardcoded ROCm install path

mrwyattii:
Addresses #2080


jeffra (approved):

aphedges:
I've seen test failures like this before when working on my PRs. There are some flaky tests that assume a certain execution time, and the threshold seems to be low enough to cause these intermittent failures.

mrwyattii:
> I've seen test failures like this before when working on my PRs. There are some flaky tests that assume a certain execution time, and the threshold seems to be low enough to cause these intermittent failures.

@aphedges yes, we've been noticing them more frequently too. We may increase the allowed threshold, but usually rerunning the tests will allow all inference tests to pass.

Issue #2046, opened 2022-06-23
[BUG] AttributeError: 'NewGELUActivation' object has no attribute '__flops__'


xiazeyu:
**Describe the bug**
When running https://github.com/microsoft/DeepSpeedExamples/tree/master/autotuning/hf/gpt2, it throws AttributeError: 'NewGELUActivation' object has no attribute '__flops__'

**To Reproduce**
./test_tune.sh tune

**Expected behavior**
Runs successfully

**ds_report output**
[2022-06-23 22:20:46,880] [WARNING] [partition_parameters.py:60:<module>] unable to find torch.distributed._all_gather_base. will fall back to torch.distributed.all_gather which will result in suboptimal performance. please consider upgrading your pytorch installation.
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
cpu_adam ............... [YES] ...... [OKAY]
cpu_adagrad ............ [YES] ...... [OKAY]
fused_adam ............. [YES] ...... [OKAY]
fused_lamb ............. [YES] ...... [OKAY]
sparse_attn ............ [YES] ...... [OKAY]
transformer ............ [YES] ...... [OKAY]
stochastic_transformer . [YES] ...... [OKAY]
async_io ............... [YES] ...... [OKAY]
utils .................. [YES] ...... [OKAY]
quantizer .............. [YES] ...... [OKAY]
transformer_inference .. [YES] ...... [OKAY]
--------------------------------------------------
DeepSpeed general environment info:
torch install path ............... ['/home/xiaze/miniconda3/envs/ds_18/lib/python3.9/site-packages/torch']
torch version .................... 1.8.1+cu111
torch cuda version ............... 11.1
torch hip version ................ None
nvcc version ..................... 11.1
deepspeed install path ........... ['/home/xiaze/miniconda3/envs/ds_18/lib/python3.9/site-packages/deepspeed']
deepspeed info ................... 0.6.6+6719b46b, 6719b46b, master
deepspeed wheel compiled w. ...... torch 1.8, cuda 11.1

**System info (please complete the following information):**
 - OS: Ubuntu 20.04
[...]

PR #2106, opened 2022-07-18
Dont overwrite hook handles in flop profiler

Sanger2000:
This is a solution to #2046.

As explained there:
> I ran into the same issue. The problem here is that huggingface instantiates activation function modules like `NewGELUActivation` at the python global scope. So, when deepspeed recursively registers hooks to model, several forward hooks are added for the same `NewGELUActivation` object. But, they all overwrite the same removable handle attributes saved to the object. Then, when deepspeed tries to remove these hooks, it can only remove 1, leaving several stale hooks on the module.


cli99 (approved):
Thanks for the PR.

PR #2111, opened 2022-07-19
reduce ds-inference log verbosity

jeffra:


mrwyattii (approved):

stas00:
this now prints the same log - one line per layer - but on rank 0 only - if that's the intention then this PR works. Thank you, Jeff!

I just wasn't sure if the user cares for 72 printouts of (identical?) config for BLOOM - perhaps some are different - it's hard to tell.

RezaYazdaniAminabadi (approved):

PR #2127, opened 2022-07-23
[zero-3] print warning once and support torch parameter

awan-10:


deepspeed/runtime/zero/parameter_offload.py
```diff
@@ -44,10 +47,13 @@ def _apply_to_tensors_only(module, functional, backward_function, outputs):
         return functional.apply(module, backward_function, outputs)
     else:
         if not is_builtin_type(outputs):
-            logger.warning(
-                f"A module has unknown inputs or outputs type ({type(outputs)}) and the tensors embedded in it cannot be detected. "
-                "The ZeRO-3 hooks designed to trigger before or after backward pass of the module relies on knowing the input and "
-                "output tensors and therefore may not get triggered properly.")
+            global warned
+            if not warned:
```
jeffra, line 51: Can we also only do this warning on rank 0?

jeffra (approved):

PR #2148, opened 2022-07-27
Enable fused_lamb_cuda_kernel on ROCm

rraminen:
This PR is to enable fused_lamb_cuda_kernel on ROCm

The below HIP cooperative groups are available on ROCm
> coalesced_group 
> coalesced_threads


rraminen:
cc: @jithunnair-amd

jeffra (approved):

PR #2152, opened 2022-07-28
fix: fix BF16_Optimizer compatibility issue

shjwudp:
Fix BF16_Optimizer compatibility issue with optimizer state 0-dim tensor, `tensor.narrow` does not support 0-dim tensor.

### DIFF
--- a/deepspeed/runtime/bf16_optimizer.py
+++ b/deepspeed/runtime/bf16_optimizer.py
@@ -417,7 +417,7 @@ class BF16_Optimizer(ZeROOptimizer):
                                   hp_frag_address.numel)
                 for key,
                 value in self.optimizer.state[flat_hp_partition].items()
-                if torch.is_tensor(value)
+                if torch.is_tensor(value) and value.dim() > 0
             }
 
             lp_frag_address = fragment_address(start=fragment_start - lp_start,


tjruwase:
@shjwudp, thanks for this PR. Can you share a bit more context on when this fails?

shjwudp:
> @shjwudp, thanks for this PR. Can you share a bit more context on when this fails?

OK, this is really an edge case, in fairseq's Adafactor implementation, the RMS is calculated by the formula `tensor.norm(2) / (tensor.numel() ** 0.5) `, which is a 0-dim tensor.
https://github.com/facebookresearch/fairseq/blob/main/fairseq/optim/adafactor.py#L223

tjruwase (approved):

tjruwase:
Interesting, thanks for sharing. So, to clarify, are you applying bf16_optimizer to adafactor for large model training? If so, we would be very interested in your experience. I suspect there might be issues with checkpoint load/stores, perhaps :).

toriving:
@shjwudp 
hello.
I am using bf16 adafactor same as you.
In my case, I use the Huggingface's adafactor, and when I use it with deepspeed, the loss does not drop and it is not learning.
Have you ever experienced any of these problems?
Or, what happens if you learn without modifying this?

PR #2160, opened 2022-08-01
[deepspeed/autotuner] Bug fix for batch size binary search

rahilbathwal5:
The current implementation terminates when `low = high - 1` and in the process skips checking `high` as a possible value for the `max_micro_batch_size`.


rahilbathwal5:
Closing in favor of https://github.com/microsoft/DeepSpeed/pull/2162.

PR #2165, opened 2022-08-02
[deepspeed/autotuner] Fix for timer for gas > 1

rahilbathwal5:
The current implementation stops the timer even for non gradient accumulation boundary steps which artificially inflates the throughput when gas > 1.


tjruwase:
@rahilbathwal5, this also needs updating to master. Thanks!

rahilbathwal5:
@tjruwase Would you mind taking a look at this PR please? Thanks!

PR #2170, opened 2022-08-02
Pydantify DeepSpeed configs

mrwyattii:


mrwyattii:
splitting this work into multiple PRs, work on sub-configs is now in #2210

Issue #2169, opened 2022-08-02
[BUG][0.6.7]DeepSpeed is trying to load the whole model to 1 GPU instead of partitioning


lanking520:
**Describe the bug**

OOM on GPU when running GPU = 4. My GPU have 40GB memory each (A100) and my CPU memory is 488GB. However, I am trying to load the OPT-30B model with 80GB size. In this case, after creating 4 processes using DeepSpeed launcher, the system goes GPU OOM after 10 min. 

In theory, if the model is loaded on CPU, DeepSpeed should slice each layer to one GPU so the load can be balanced. In that case, we would expect each GPU take 80 / 4 = 20GB. I also tested GPU=8, it falls into the same issue. It seemed the DeepSpeed.init() seemed to copy the whole CPU-memory model to 1 single GPU.

**To Reproduce**
I am running with 8 GPU instance with A100. 

```python
import os
import deepspeed
import torch
from transformers import pipeline, AutoModelForCausalLM, AutoTokenizer

local_rank = int(os.getenv('LOCAL_RANK', '0'))
world_size = int(os.getenv('WORLD_SIZE', '1'))

model_name = 'facebook/opt-30b'
model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
tokenizer = AutoTokenizer.from_pretrained("facebook/opt-30b", use_fast=False)

generator.model = deepspeed.init_inference(generator.model,
                                           mp_size=world_size,
                                           dtype=torch.float,
                                           replace_method='auto',
					   replace_with_kernel_inject=True)

generator = pipeline(task='text-generation', model=model, tokenizer=tokenizer, device=local_rank)

string = generator("DeepSpeed is", do_sample=True, min_length=50)
if not torch.distributed.is_initialized() or torch.distributed.get_rank() == 0:
    print(string)
```

```
deepspeed --num_gpus 8 test.py
```

**Expected behavior**

Should be just normal?

**ds_report output**

```
# ds_report
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
[...]

PR #2179, opened 2022-08-03
Fixing model partitioning without injection

RezaYazdaniAminabadi:
This addresses https://github.com/microsoft/DeepSpeed/issues/2169

cc: @arashb


jeffra (approved):

jeffra:
Can we add a unit test that uses LinearLayer in this way? In a new PR is fine

PR #2181, opened 2022-08-04
Simplifying inference API

RezaYazdaniAminabadi:


arashb (approved):

awan-10:
Our API has been changed significantly since then so this PR is no longer active/relevant.

PR #2202, opened 2022-08-09
Refactor dist tests: Checkpointing

tjruwase:
Refactor checkpointing distributed unit tests.


tests/unit/checkpoint/test_lr_scheduler.py
```diff
@@ -0,0 +1,165 @@
...
+                              (1,
+                               False),
+                              (2,
+                               False),
+                              (2,
+                               True),
+                              (3,
+                               False),
+                              (3,
+                               True)])
```
mrwyattii, line 106: this parametrize can be removed if you move the one above this (for `test_checkpoint_lr_schedule`) to the class level (just above the `class TestLR...`)

tests/unit/checkpoint/test_latest_checkpoint.py
```diff
@@ -0,0 +1,62 @@
...
+            "steps_per_print": 1,
+            "optimizer": {
+                "type": "Adam",
+                "params": {
+                    "lr": 0.00015
+                }
+            }
+        }
+        hidden_dim = 10
+        args = args_from_dict(tmpdir, config_dict)
```
mrwyattii, line 24: not super important, but I've been trying to remove unnecessary code. This args_from_dict is not necessary anymore and you can replace all `deepspeed.initialize(args=args, ...` with `deepspeed.initialize(config=config_dict, ...`

tests/unit/checkpoint/test_moe.py
```diff
@@ -0,0 +1,119 @@
...
+            checkpoint_correctness_verification(args,
+                                                models=models,
+                                                hidden_dim=hidden_dim,
+                                                tmpdir=tmpdir,
+                                                load_optimizer_states=True,
+                                                load_lr_scheduler_states=False,
+                                                fp16=config_dict["fp16"]["enabled"],
+                                                empty_tag=True,
+                                                base_optimizers=optimizers,
+                                                seq_dataloader=True)
```
mrwyattii, line 48: What's the purpose of having the local function `_helper`? couldn't this just be un-indented once and run the same? I think this applies to multiple tests here

tests/unit/checkpoint/test_other_optimizer.py
```diff
@@ -0,0 +1,151 @@
...
+
+        _test_checkpoint_unfused_optimizer(args=args,
+                                           models=models,
+                                           hidden_dim=hidden_dim,
+                                           load_optimizer_states=True)
+
+        _test_checkpoint_unfused_optimizer(args=args,
+                                           models=models,
+                                           hidden_dim=hidden_dim,
+                                           load_optimizer_states=False)
```
mrwyattii, line 73:
```suggestion
        checkpoint_correctness_verification(
            args,
            models=models,
            hidden_dim=hidden_dim,
            tmpdir=tmpdir,
            load_optimizer_states=True)
            
        checkpoint_correctness_verification(
            args,
            models=models,
            hidden_dim=hidden_dim,
            tmpdir=tmpdir,
            load_optimizer_states=False)
```

This might be easier to read (and is less code) than using a local function

mrwyattii: We could even make this a for loop:
`for load_opt_state in [True, False]:`

tests/unit/checkpoint/test_pipeline.py
```diff
@@ -0,0 +1,119 @@
...
+from tests.unit.simple_model import *
+
+from tests.unit.checkpoint.common import checkpoint_correctness_verification
+
+import pytest
+
+
+@pytest.fixture(params=[2])
+def stages(request):
+    return request.param
```
mrwyattii, line 12: What's the purpose of having this be a fixture? Is there a reason we may want to change the value of `stages` later?

tests/unit/checkpoint/test_pipeline.py
```diff
@@ -0,0 +1,119 @@
...
+            # PipeTopo(num_pp=4,
+            #          num_dp=1)),
+            #(PipeTopo(num_pp=2,
+            #          num_dp=2),
+            # PipeTopo(num_pp=2,
+            #          num_dp=2)),
+            #(PipeTopo(num_pp=4,
+            #          num_dp=1),
+            # PipeTopo(num_pp=2,
+            #          num_dp=2)),
```
mrwyattii, line 82: If these are all supposed to be commented out, this parametrize is not needed

mrwyattii (changes requested):

tests/unit/checkpoint/test_moe.py
tjruwase: Fixed.

tests/unit/checkpoint/test_other_optimizer.py
tjruwase: Done

tests/unit/checkpoint/test_pipeline.py
tjruwase: Yes, it is currently all [commented out](https://github.com/microsoft/DeepSpeed/blob/f82846d7df6a129668bea447a0f953f6cf345efb/tests/unit/test_checkpointing.py#L712).

I am not sure the reason, but I seem to recall this being a flaky unit test.

tests/unit/checkpoint/test_pipeline.py
tjruwase: I thought it was the right way to reproduce a default argument value as done in the original [test](https://github.com/microsoft/DeepSpeed/blob/f82846d7df6a129668bea447a0f953f6cf345efb/tests/unit/test_checkpointing.py#L661). Will remove and simplify

tests/unit/checkpoint/test_latest_checkpoint.py
tjruwase: Done

tests/unit/checkpoint/test_pipeline.py
tjruwase: Done

tests/unit/checkpoint/test_lr_scheduler.py
tjruwase: Done

tests/unit/checkpoint/test_pipeline.py
mrwyattii: That makes sense, I think it's probably fine either way!

mrwyattii (approved):
lgtm

PR #2205, opened 2022-08-09
Add support of OPT models

### LINKED ISSUE #1978: DeepSpeed Inference support for OPT
Right now OPT (https://huggingface.co/docs/transformers/model_doc/opt) can only be supported via custom kernel injection policy. It would be great if there's official support. Thanks!

arashb:
This PR:
- adds OPT layer policy
- adds support of choosing activation function for MLP 
- adds support of ReLU activation function which is used in OPT MLP
- fixes an issue related to positional embedding in OPT models
- refactor the code to be able to set whether pre- or post-attention layer norm is needed at the policy level
- add support of opt-350m model which the layer norm is applied after attention layer 

fixes #1978


jeffra (approved):

deepspeed/ops/transformer/inference/transformer_inference.py
```diff
@@ -589,7 +592,8 @@ def forward(ctx,
                 mlp_gemm_func,
                 fused_gemm_gelu,
                 vector_matmul_func,
-                bias_residual_func):
+                bias_residual_func,
+                activation_func_type=ActivationFuncType.GELU):
```
RezaYazdaniAminabadi, line 596: I think this argument (`activation_func_type`) is not used anywhere, right? We can just remove it

RezaYazdaniAminabadi (approved):
Thanks @arashb, all looks great :)

deepspeed/ops/transformer/inference/transformer_inference.py
arashb: `activaiton_func_type` is passed to `mlp_gemm_func` so we can decide which activation kernel to launch in MLP.

PR #2212, opened 2022-08-12
Extend scratch buffer for long prompts

cmikeh2:
Segfault fix for long sequence length prompt.


csrc/transformer/inference/csrc/pt_binding.cpp
```diff
@@ -305,7 +312,15 @@ void attention_unfused(T* prev_key_cont,
...
     float gemm_beta = 0.0;
-    T* workspace = (T*)output + bsz * seq_len * heads * k;
+    T* workspace;
+    if (seq_len == 1) {
+        workspace = (T*)output + bsz * seq_len * heads * k;
+    } else {
+        // If we are doing the prompt, switch to the tail workspace
+        T* scratch = (T*)Context::Instance().GetWorkSpace();
+        workspace = scratch + (Context::Instance().get_workspace_size() / sizeof(T)) -
+                    heads * seq_len * seq_len;
```
RezaYazdaniAminabadi, line 322: @cmikeh2, do we want to add bsz here too, since we might get a batched input?

csrc/transformer/inference/csrc/pt_binding.cpp
```diff
@@ -591,14 +606,19 @@ std::vector<at::Tensor> ds_qkv_gemm(at::Tensor& input,
                                     at::Tensor& beta,
                                     const float epsilon,
                                     bool add_bias,
-                                    unsigned num_layers)
+                                    unsigned num_layers,
+                                    bool bloom_seq_len,
```
RezaYazdaniAminabadi, line 610: maybe change this to `is_bloom`?

csrc/transformer/inference/csrc/pt_binding.cpp
cmikeh2: Yeah, this is primed to cause a segfault on any batched input.

jeffra (approved):

PR #2214, opened 2022-08-12
Type checking in zero-3 to avoid silent bugs

awan-10:
This is a follow-up PR to the existing one: https://github.com/microsoft/DeepSpeed/pull/2127

The goal is to keep this PR open and investigate the issue in more detail while the PR above removes the annoying repeated printing of the warning.


HeyangQin:
Hi @awan-10, it seems this has already been merged by another of your PR https://github.com/microsoft/DeepSpeed/pull/2127. Do we still need this PR?

PR #2232, opened 2022-08-17
Add tensor parallel inference unit tests

mrwyattii:


tests/unit/inference/test_inference.py
```diff
@@ -269,12 +269,62 @@ def test(
...
+@pytest.mark.parametrize("model_w_task",
+                         [("gpt2",
+                           "text-generation"),
+                          ("EleutherAI/gpt-neox-20b",
+                           "text-generation"),
+                          ("bigscience/bloom-3b",
+                           "text-generation")],
+                         ids=["gpt2",
+                              "gpt-neox",
+                              "bloom"])
```
jeffra, line 291: Can we add OPT into the list but just comment it out? will make it easier to turn back on in a future PR that fixes those issues.

jeffra (approved):

tests/unit/inference/test_inference.py
mrwyattii: I don't think OPT models are supported with multi-gpu at the moment

.github/workflows/nv-inference.yml
```diff
@@ -40,7 +41,7 @@ jobs:
           git clone https://github.com/huggingface/transformers
           cd transformers
           # if needed switch to the last known good SHA until transformers@master is fixed
-          # git checkout 1cc453d33
+          git checkout v4.21.2
```
jeffra, line 44: I think we can change to this be latest master from HF, after we merge: https://github.com/microsoft/DeepSpeed/pull/2291. But I want to re-test #2291 after this PR is merged.

jeffra (approved):

PR #2250, opened 2022-08-23
Enable contiguous gradients with Z1+MoE

siddharth9820:
MoE training with zero stage 1 only works with `contiguous_gradients=True`.


siddharth9820:
@tjruwase

tjruwase (approved):

jeffra (approved):
This does have the side effect of making `contiguous_gradients` default True for ZeRO-1. But this is probably okay

PR #2253, opened 2022-08-23
Refactor universal checkpointing and tensor fragments

tjruwase:
First step in enabling universal checkpoints for all zero stages. This moves the building blocks of universal checkpoint format outside of the bf16_optimizer. 

Basic testing verifies that bf16_optimizer can use refactored logic to resume training from a tp=2,pp=2,dp=2 checkpoint into different tp/pp/dp shapes. 

@stas00, FYI


stas00:
super! thank you for working on that, Tunji!

and down the road perhaps we can switch to this functionality for `zero_to_fp32.py` - so it'll convert to the universal checkpoint and then easily pluck the fp32 weights.

jeffra (approved):

PR #2276, opened 2022-08-31
Memory Access Utility

cmikeh2:
### Simplifying Loads
This PR introduces a header file for simplifying coarse loads when writing CUDA kernels. As an example, to induce a 16 byte load on half precision data, we need to perform a cast to an unrelated `float4` datatype:
```
float4* half_data = reinterpret_cast<float4*>(half_ptr);
float4 local_data = half_data[idx];
// Reinterpret back to half precision data
```
This decreases the readability and increases the verbosity writing kernels. The `memory_access_utils.h` header introduces a set of APIs to perform the loads natively.
```
__half local_data[8];
mem_access::load_global<16>(local_data, half_ptr);
// Iterate over the local_data array
```
This both removes the reference to the unrelated type while also clearly specifying what address space a memory transaction occurs between. The library also supports specifying cache hints on supported platforms:
```
mem_access::load_global<8, mem_access::LoadPolicy::CacheStreaming>(register_ptr, global_ptr);
```
For specific cache operators, see the header file.

Integration for this has begun in `gelu.cu`. Correspondingly, unit tests have been introduced to validate the accuracy of the reworked kernel.

### Asynchronous Copies
The new header file also introduces support for asynchronous copies between global and shared memory on supported architectures (currently Ampere). The asynchronous copies may be used as follows:
```
// Issue two 16 byte copies, their ordering is not defined
mem_access::memcpy_async<16>(shr_ptr_1, gbl_ptr_1);
mem_access::memcpy_async<16>(shr_ptr_2, gbl_ptr_2);
// Group them into a commit group
mem_access::memcpy_async_fence();
// Do other things
// Once we need the data, we wait until 0 commit groups are still in flight
mem_access::memcpy_async_wait<0>();
```
### Other Changes
The header files for the inference transformer are changed in this PR. This allows us to add `memory_access_utils.h` to the primary includes directory (`csrc/includes`), and still include it from both training and inference transformer kernel code.


awan-10 (approved):

jeffra (approved):
🚀

PR #2278, opened 2022-08-31
add a new unit test for cuda ops

awan-10:
This PR introduces a test for our CUDA kernels. The main PR authored by Connor is here: https://github.com/microsoft/DeepSpeed/pull/2276

The unit test in this PR is factored out from there so we can merge it independently.

### DIFF
--- a/tests/unit/ops/transformer/inference/test_bias_gelu.py
+++ b/tests/unit/ops/transformer/inference/test_bias_gelu.py
@@ -10,7 +10,6 @@ if not deepspeed.ops.__compatible_ops__[InferenceBuilder.NAME]:
 inference_module = None
 torch_minor_version = None
 
-
 def allclose(x, y):
     assert x.dtype == y.dtype
     rtol, atol = {torch.float32: (5e-4, 5e-5), torch.float16: (3e-2, 2e-3)}[x.dtype]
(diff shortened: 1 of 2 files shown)


tests/unit/ops/transformer/inference/test_bias_gelu.py
```diff
@@ -0,0 +1,46 @@
+import pytest
+import torch
+import deepspeed
+from deepspeed.ops.op_builder import InferenceBuilder
+
+if not deepspeed.ops.__compatible_ops__[InferenceBuilder.NAME]:
+    pytest.skip("Inference ops are not available on this system",
+                allow_module_level=True)
+
+inference_module = InferenceBuilder().load()
```
jeffra, line 10: This will invoke an op build which will initialize cuda, i think this is the core of the test failures. Let's set this to `None` here and then build it only if `inference_module is None` inside each of the `def test_*` methods?

awan-10: @cmikeh2 -- please see the comment above. I think this is the problem.

awan-10: @jeffra -- thank you! Pushed a possible fix. Please check.

jeffra (approved):

PR #2285, opened 2022-09-01
Fp32 accuracy bug fix

RezaYazdaniAminabadi:
This PR fixes the stride of the attention mask mem. layout used in `attention_unfused` kernel. 

Additionally, two utility functions `infer_transformer_type()` and `get_attn_mask_stride()` are introduced to encapsulate the logic in separate functions.


jeffra (approved):
@arashb, @cmikeh2, @awan-10: we discussed some ideas on how to clean this up. Approving for now but perhaps we fix this up in an upcoming refresh/refactor.

PR #2291, opened 2022-09-02
only override forward if using cuda-graph

jeffra:
https://github.com/huggingface/transformers/pull/18261 introduces model arg validation, which is not compatible with how ds-inference was originally setup. We no longer need to do all of the things we previously did in an engine forward. This PR simplifies the inference engine a bit and also protects against a case where mp>1 and cuda graph are both enabled which is currently not compatible.


deepspeed/inference/engine.py
```diff
@@ -162,10 +162,10 @@ def __init__(self,
             torch.cuda.set_rng_state(_rng_state.cpu())
 
         if self.mp_world_size > 1:
+            assert not self.enable_cuda_graph, "Cuda graph is not supported for model parallelism"
+        elif self.enable_cuda_graph:
             self.model_orig_fwd = self.module.forward
             self.module.forward = self.forward
-        else:
-            self.module.register_forward_pre_hook(self._pre_forward_hook)
```
RezaYazdaniAminabadi, line 168: The forward hook was initially added to make sure that all tensors are on GPU! But I think it is just safe to remove it as it is the responsibility of user to put them on device!

RezaYazdaniAminabadi (approved):

PR #2296, opened 2022-09-06
finish test_bias_residual test

GuanhuaWang:
@cmikeh2 @awan-10  For concept understanding, I finished the test code for `fused_bias_residual` kernel, and it passes pytest. Feel free to review.

### DIFF
--- a/tests/unit/ops/transformer/inference/test_bias_residual.py
+++ b/tests/unit/ops/transformer/inference/test_bias_residual.py
@@ -1,0 +1,48 @@
+import pytest
+import torch
+import deepspeed
+from deepspeed.ops.op_builder import InferenceBuilder
+
+if not deepspeed.ops.__compatible_ops__[InferenceBuilder.NAME]:
+    pytest.skip("Inference ops are not available on this system",
+                allow_module_level=True)
+
+inference_module = None
+torch_minor_version = None
+
+def allclose(x,y):
+    assert x.dtype == y.dtype
+    assert x.shape == y.shape
+    rtol, atol = {torch.float32: (5e-4, 5e-5), torch.float16: (3e-2, 2e-3)}[x.dtype]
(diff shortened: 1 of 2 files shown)


awan-10:
@GuanhuaWang - thanks for making the test. Can you please fix the formatting so the automatic tests run? 

Please see the https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md guide to fix the formatting etc.

tests/unit/ops/transformer/inference/test_bias_residual.py
```diff
@@ -0,0 +1,48 @@
...
+torch_minor_version = None
+
+def allclose(x,y):
+    assert x.dtype == y.dtype
+    assert x.shape == y.shape
+    rtol, atol = {torch.float32: (5e-4, 5e-5), torch.float16: (3e-2, 2e-3)}[x.dtype]
+    return torch.allclose(x,y, rtol=rtol, atol=atol)
+
+def run_bias_residual_reference(inputs, residuals, bias):
+    return inputs
```
awan-10, line 20: Is the reference implementation just to return inputs? or do you plan to work on this further?

GuanhuaWang: In our pybinding, right now this one is disabled and only return input. 
https://github.com/microsoft/DeepSpeed/blob/master/csrc/transformer/inference/csrc/pt_binding.cpp#L556

I think it is ok for now. We can update this part when the bias_residual binding is updated.

GuanhuaWang:
> @GuanhuaWang - thanks for making the test. Can you please fix the formatting so the automatic tests run?
> 
> Please see the https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md guide to fix the formatting etc.


Done formatting @awan-10

PR #2307, opened 2022-09-09
Add unit tests for residual_add kernel

arashb:
This PR adds unittests for both `launch_gptj_residual_add` and `launch_bias_residual` kernels (which kernel gets tested depends on the value of `mlp_after_attn` argument). 

The C++ APIs of these kernels are subject to refactoring, but this will be pushed in another PR. 
The parameter `preln` is not tested in this PR as it most probably changes after refactoring described above. 

TODOs:
- [x] fix the combination of `add_bias=False` and `mlp_after_attn=True`
- [x] add support of `mp_size > 1`

CC: @awan-10 @cmikeh2

### DIFF
--- a/tests/unit/ops/transformer/inference/test_residual_add.py
+++ b/tests/unit/ops/transformer/inference/test_residual_add.py
@@ -1,0 +1,106 @@
+"""
+Copyright 2022 The Microsoft DeepSpeed Team
+"""
+
+import pytest
+import torch
+import deepspeed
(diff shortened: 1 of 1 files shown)


arashb:
the tests in this PR are going to pass after the #2310 fix is merged.

awan-10 (approved):
Thanks, Arash!! This PR looks great.

cmikeh2 (approved):

PR #2315, opened 2022-09-09
add quant unit test

GuanhuaWang:
This PR adds unit test for `ds_quantize` kernel, In this example, we test over `torch.float16` data format. 

We tested in following two cases:

1. INT8 quant/dequant over tensor in 1 group. 

> Here we form 1 single quantization group. We quantize `input_tensor` with data format of `torch.float16` to `torch.int8`, and then convert back to `torch.float16` tensor.

2. INT4 quant/dequant over tensor partitioned into 16 groups. 

> Here we form 16 quantization groups. We quantize `input_tensor` with data format of `torch.float16` to `int4`, then convert back to `torch.float16` tensor.

TODOs:
- [x] fix unit-tests compiling error of `undefined symbol: curandSetPseudoRandomGeneratorSeed` by adding `-lcurand` flag inside `op_builder/quantizer.py`
- [x] create new folder for quantizer tests as `tests/unit/ops/quantizer`

The C++/CUDA APIs of the kernels are subject to refactoring, but it will lead to another PR.

CC: @awan-10 @cmikeh2


tests/unit/ops/transformer/inference/test_quant.py
```diff
@@ -0,0 +1,38 @@
+import torch
+import pytest
+from deepspeed.ops import op_builder
+
+quantizer_cuda_module = op_builder.QuantizerBuilder().load()
```
cmikeh2, line 5: I believe this is causing the unrelated tests to fail. Let's refactor this to use the loading style used either [here](https://github.com/microsoft/DeepSpeed/blob/a691ec605b2383ecac21ecf33a9ba1c0cba4dd01/tests/unit/ops/transformer/inference/test_bias_gelu.py#L42).

tests/unit/ops/transformer/inference/test_quant.py
```diff
@@ -0,0 +1,38 @@
...
+
+    scale = q_range / (2 * torch.max(input_min.abs(), input_max.abs()))
+    input_flat = (input_flat * scale).round().clamp(-q_range // 2, q_range // 2 - 1)
+    # dequantize
+    return input_flat.reshape(inputs.shape).to(torch.int8) / scale.view(-1).to(
+        torch.float16)
+
+
+@pytest.mark.inference
+@pytest.mark.parametrize("input_tensor", torch.rand(8, 8, dtype=torch.float16).cuda())
```
cmikeh2, line 30: I think we could benefit from some more test cases here. Looking at the kernels, we should test that it partitions in the correct number of groups correctly and things like that as well as larger tensor/higher dimensioned Tensors.

cmikeh2 (commented):
I think we should move the test file to `tests/unit/ops/quantizer` to match the structure of the C++ source code and to make it clear this won't apply to the existing dequantization kernels we have there and any future quantization kernels that are added for inference.

tests/unit/ops/transformer/inference/test_quant.py
awan-10: @GuanhuaWang -- yes, please see the code above. We set the module to none at global level and only load it inside the function that needs it.

GuanhuaWang: @cmikeh2 @awan-10  move the test file to `tests/unit/ops/quantizer`, also used the loading style used either [here](https://github.com/microsoft/DeepSpeed/blob/a691ec605b2383ecac21ecf33a9ba1c0cba4dd01/tests/unit/ops/transformer/inference/test_bias_gelu.py#L42).

tests/unit/ops/transformer/inference/test_quant.py
GuanhuaWang: @cmikeh2 added more tests and also explicitly mentioned the group size boundary in comment.

tests/unit/ops/quantizer/test_quant.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def run_quant_dequant(inputs, groups, bits):
+    global quantizer_cuda_module
+    if quantizer_cuda_module is None:
+        quantizer_cuda_module = op_builder.QuantizerBuilder().load()
+    return quantizer_cuda_module.ds_quantize_fp16(inputs, groups, bits)
+
+
+@pytest.mark.inference
```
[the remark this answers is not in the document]
GuanhuaWang: @cmikeh2  done.

tests/unit/ops/quantizer/test_quant.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.mark.parametrize("tensor_shape", [(8, 8), (128, 256)])
+def test_quant_dequant(tensor_shape):
+    input_tensor = torch.rand((tensor_shape), dtype=torch.float16).cuda()
+
+    # test 8bit quant/dequant on tensor partitioned in 1 group.
+    ref_input_8bit_1group = input_tensor.clone().detach()
+    ref_out_8bit_1group = quantize_dequantize_ref(ref_input_8bit_1group, 8)
+    # run_quant_dequant will do quantize then dequantize and return the dequantized value.
+    ds_out_8bit_1group = run_quant_dequant(input_tensor, 1, 8)
```
cmikeh2, line 45: Is the quantization process in-place? Does that have repercussions for doing both the 8bit-1group and 4bit-16group on the same Tensor? If so maybe add another detach or factor it out as another parameterized aspect?

GuanhuaWang: detached the tensor for both ds and ref now. Thx for find this bug. @cmikeh2

awan-10 (approved):

PR #2327, opened 2022-09-15
Kernel Data Conversion Utility

cmikeh2:
This PR introduces a change to where/how we define macros and constants for device kernels and introduces a new utility header file for more flexible and clear datatype conversions.

#### Macros and Constants
Previously, we defined macros and constants across multiple header files for use in specific modules. Since these macros tended to define common functionality about the compilation targets rather than module specific code, I have unified them into a new header file `ds_kernel_utils.h` so that we have consistency on what these macros mean across different modules. 

Furthermore, this introduces a new `constexpr` variable `hw_warp_size` which should match the externally (and `const`) variable `warpSize` for each platform. This makes the hardware warp size available at compile time (useful for templating) while also provides a starting point for ensuring that kernel code is appropriately dimensioned for the target hardware and doesn't presume the size of a wavefront.

#### Conversion Utils
`conversion_utils.h` is a new utility header file that simplifies type conversions in device code. For kernels that switch between common data types like `__half` and `float`, it is necessary to use verbose intrinsics that make templating even simple functions more difficult. Additional, implicit type-casts in C++ can harm code readability.

This header file introduces a universal interface for conversion between the primary types from within kernels. An example of this is conversion from `float2` to `__half2`:
```
// previous version
__half2 h_data = __float22half2(f_data);
// new api
__half2 h_data = conversion::to<__half2>(f_data);
```
While slightly more verbose, it means a consistent code pattern for all type conversions. It also allows for better code writing efficiency by better enabling templating. For activation functions, we frequently up-cast half precision values to a floating point in order to maintain precision for non-linear functions. This would require a call to an intrinsic, preventing us from using the same implementation for both `float` and `__half`, even though the code scheduling is identical. With the conversion util however:
```
float up_cast = conversion::to<float>(data);
up_cast = activation(data);
T data = conversion::to<T>(up_cast);
```
If `T` is `float` the conversion evaluates to no-op. We can now use the same memory scheduling implementation with no code changes though. An example of this is demonstrated with `fused_bias_gelu` in the PR.

A final benefit is that this removes the dependency on remembering verbose intrinsic names.

#### Areas for Feedback
1. Since we are refactoring some of these macros, are there different names that we'd like to use for them?
2. For the conversion utility, would a shorter namespace name (`convert`) be preferred. Are there ways the API could be cleaner?
3. The conversion utility focuses on conversions for single elements rather than arrays of elements. Supporting this could further simplify user code at the cost of potentially worse code patterns. Right now I'm handling this by forcing the user to implement the array version themselves, but I can understand the argument for including some default implementations.

@jeffra @RezaYazdaniAminabadi @awan-10


RezaYazdaniAminabadi (approved):
Looks great! :-) Thanks @cmikeh2 for perparing such useful utility, I think this would help us develop faster, reduce the code and improve the readability.

PR #2331, opened 2022-09-18
mem access for quantize kernel

GuanhuaWang:
add mem-access code refactoring in quantization kernel code, which supports both fp16 and fp32 input.

@cmikeh2 @awan-10 

Do want to mention, thx @cmikeh2 for helping me fix bugs in quant kernels.


csrc/quantization/quantizer.cu
```diff
@@ -1,42 +1,45 @@
...
+    cg::thread_block_tile<32> g =
+        cg::tiled_partition<32>(b);  // warp, 32 not optimal for AMD which should be 64.
 
     int gid = threadIdx.x >> 5;
     int lane = threadIdx.x & 0x1f;
     int warp_num = blockDim.x >> 5;
     int id = threadIdx.x;
 
-    float2* vals_cast = reinterpret_cast<float2*>(vals);
+    constexpr int granularity = 8;
```
cmikeh2, line 20: Let's increase granularity to 16 here to match the float implementation. This should also simplify the launch code.

csrc/quantization/quantizer.cu
```diff
@@ -1,42 +1,45 @@
...
+    constexpr int vals_per_access = granularity / sizeof(__half);
 
-    float2 data[MAX_REG];
+    __half data[MAX_REG * vals_per_access];
```
cmikeh2, line 23: This array is much larger than the register file anyways; rather than store here, let's just store into a temp array on each loop iteration and read from global memory on both loops.

tests/unit/ops/quantizer/test_quant.py
```diff
@@ -7,7 +7,7 @@
 
 def allclose(x, y):
     assert x.dtype == y.dtype
-    rtol, atol = {torch.float32: (2e-1, 5e-2), torch.float16: (2e-1, 5e-2)}[x.dtype]
+    rtol, atol = {torch.float32: (2e-1, 5e-1), torch.float16: (2e-1, 5e-1)}[x.dtype]
```
cmikeh2, line 10: Is there a reason we reduced our error tolerance here? Quant-dequant (especially with floating point values) should   be very precise.

csrc/quantization/quantizer.cu
```diff
@@ -1,42 +1,45 @@
...
 
     int group_id = blockIdx.x;
 
     {
-        int group_index = id;
+        int group_index = id * vals_per_access;
```
cmikeh2, line 28: Let's refactor some of these variable names (particularly the ones that just pass through other values).

csrc/quantization/quantizer.cu
```diff
@@ -103,28 +96,31 @@ __global__ void quantize_kernel(float* vals, int group_size, int num_bits)
     int warp_num = blockDim.x >> 5;
     int id = threadIdx.x;
 
-    float4* vals_cast = reinterpret_cast<float4*>(vals);
+    constexpr int granularity = 16;
+    constexpr int vals_per_access = granularity / sizeof(float);
 
-    float4 data[MAX_REG];
+    float data[MAX_REG * vals_per_access];
```
cmikeh2, line 102: Where applicable, apply the comments for the `__half` implementation to the `float` implementation.

csrc/quantization/quantizer.cu
GuanhuaWang: done

csrc/quantization/quantizer.cu
GuanhuaWang: done all above. already pushed new commit @cmikeh2 feel free to check the latest version.

tests/unit/ops/quantizer/test_quant.py
GuanhuaWang: > Is there a reason we reduced our error tolerance here? Quant-dequant (especially with floating point values) should be very precise.

I manually saw few values has a little bigger variance (but still looks similar), that is the reason I changed the bound.

cmikeh2 (approved):
This looks great to me! Thanks for putting in the work to refactor the test as well to catch the edge cases we found!

PR #2332, opened 2022-09-18
NCCL Backend

Quentin-Anthony:
@awan-10 @jeffra 

This PR introduces the dedicated NCCL backend, enabling communication calls directly through NCCL and without going through PyTorch's distributed module.


op_builder/comm.py
```diff
@@ -0,0 +1,66 @@
...
+
+MV2 = f'/opt/mvapich2/gdr/2.3.6/no-mpittool/no-openacc/cuda11.3/mofed5.4/mpirun/gnu9.3.0'
+OMPI = f'/usr/local/mpi'
+
+MPI_INCL = OMPI + '/include'
+MPI_LIB = '-L' + OMPI + '/lib -lmpi'
+
+CUDA = '/usr/local/cuda'
+CUDA_INCL = CUDA + '/include'
+CUDA_LIB = '-L' + CUDA + '/lib64 -lm -lcuda -lcudart'
```
jeffra, line 12: Where is the NCCL lib/include path set? Would it be possible to switch between different versions of NCCL? For example, when nvidia releases a new version of NCCL tomorrow can we more quickly add it in here than way for pytorch to be re-built with it 3 months later.

jomayeri:
Closing due to age and divergence from develop branch. Please redo changes from there if still needed.

PR #2333, opened 2022-09-20
Refactor residual add kernels

arashb:
This PR:
- [x] refactors the `residual_add_bias` kernel launch function to a templated function
- [x] exposes fp16/fp32 residual add kernels at python level
- [ ] ~~refactors the corresponding cuda kernels into their own separate `.cu` files~~
- [ ] ~~adds negative tests for type/shape matching of input tensors~~

NOTE: the tests for type/shape matching will be added in a separate PR at the python level.

### DIFF
--- a/deepspeed/ops/transformer/inference/transformer_inference.py
+++ b/deepspeed/ops/transformer/inference/transformer_inference.py
@@ -601,6 +601,7 @@ class DeepSpeedMLPFunction(Function):
                 fused_gemm_gelu,
                 vector_matmul_func,
                 bias_residual_func,
+                residual_add_func,
                 activation_func_type=ActivationFuncType.GELU):
 
         if attn_nw is None:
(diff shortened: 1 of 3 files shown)


arashb:
> > * [ ]   refactors the corresponding cuda kernels into their own separate `.cu` files
> 
> We can probably skip this for now? I think in all likelihood we'll want to refactor the kernel into a templated implementation anyways, at which point they'd end up in the same file.

That's fine with me

deepspeed/ops/transformer/inference/transformer_inference.py
```diff
@@ -710,6 +711,9 @@ def __init__(self,
         self.bias_residual_func = inference_cuda_module.bias_residual_fp16 if config.fp16 or config.q_int8 else \
                                     inference_cuda_module.bias_residual_fp32
 
+        self.residual_add_func = inference_cuda_module.residual_add_bias_fp16 if config.fp16 or config.q_int8 else \
```
arashb, line 714: @RezaYazdaniAminabadi when running with `int8` quantization, the `residual_add_bias_fp16` needs to be called? Where we check the correct dtype of tensors?

RezaYazdaniAminabadi: they should be fp16

arashb:
Note `nv-torch18-v100` tests failure is due to another issue which @mrwyattii will fix soon.

cmikeh2 (approved):

awan-10 (approved):
Awesome! Thank you Arash for this PR.

jeffra (approved):

Issue #2300, opened 2022-09-07
[BUG]GPT Models fail for long inputs and or outputs during inference


mallorbc:
**Describe the bug**

When using GPTJ or GPT Neo 2.7B with DeepSpeed inference if you give it the short simple "DeepSpeed is" like the tutorial shows, and generate only 50 tokens or so, then everything works.

However, when you give the model a long input, such as 1000 tokens or so, and or when you give a small input and want to generate many tokens, the system breaks.  

Through my many tries of trying to fix the issue, I have gotten errors similar to that of #2062 where illegal memory is accessed.  I have gotten errors with regards to nan/inf.  Sometimes the model does not error out but rather gives garbage output once a certain length is reached, similar to that of #2233 

**To Reproduce**
1. Install torch, transformers,etc
2. Install DeepSpeed, either from source, the latest tag, or from one of the non merged PRs I reference
3. Have a long text input in "input_data.txt"
4. Run code below
5. Notice bad results in some form described above

Note that when not specifying the min length, what sometimes happens is the model generates a few tokens but then stops.  Specify a long min lenghth gurantees issues.

```
import os
import deepspeed
import torch
from transformers import pipeline
from transformers import AutoTokenizer, AutoModelForCausalLM
import argparse
import deepspeed
# os.environ["CUDA_LAUNCH_BLOCKING"] = '1'
if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument('-m',"--model_name", type=str, default='EleutherAI/gpt-j-6B')
    parser.add_argument('--local_rank', type=int, default=-1,
                    help='local rank passed from distributed launcher')
    parser = deepspeed.add_config_arguments(parser)
    args = parser.parse_args()
    model_name = args.model_name

    with open('input_data_long.txt', 'r') as f:
        input_text = f.read()

    local_rank = int(os.getenv('LOCAL_RANK', '0'))
    world_size = int(os.getenv('WORLD_SIZE', '1'))

    model = AutoModelForCausalLM.from_pretrained(model_name,torch_dtype=torch.float16)
    tokenizer = AutoTokenizer.from_pretrained(model_name)
    generator = pipeline('text-generation', model=model, tokenizer=tokenizer, device=local_rank,torch_dtype=torch.float16)



    generator.model = deepspeed.init_inference(generator.model,
                                            mp_size=world_size,
                                            dtype=torch.half,
                                            replace_method='auto',
                        replace_with_kernel_inject=True)
[...]

PR #2344, opened 2022-09-22
Fix issue with corrupted output on long generation for GPT

andrewchernyh:
Fix issue described in https://github.com/microsoft/DeepSpeed/issues/2300

Only updating MAX_OUT_TOKES doesn't helps and main issue is in temp_buf that was allocated just after `output` tensor

Then in `attention_unfused` 
`T* workspace = (T*)output + bsz * seq_len * heads * k;` where output is `temp_buf` from `ds_softmax_context`
In the result `temp_buf` overlaps `query_cont` allocated in  `ds_softmax_context`

I've put `temp_buf` just after `kv_cache` and it helps

Minimal steps to reproduce:
```bash
python3 benchmarks/inference/gpt-bench.py -m EleutherAI/gpt-neo-125M --kernel-inject --deepspeed --dtype=fp32 --max-tokens=1020 --trials 1
```

### DIFF
--- a/csrc/transformer/inference/csrc/pt_binding.cpp
+++ b/csrc/transformer/inference/csrc/pt_binding.cpp
@@ -107,7 +107,7 @@ void allocate_workspace(size_t hidden_dim,
                         unsigned num_layers,
                         size_t head_size = 128)
 {
-    size_t _workSpaceSize = 16 * (hidden_dim * batch_size * max_seq_len) +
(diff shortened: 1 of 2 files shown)


andrewchernyh:
Not actual more, see https://github.com/microsoft/DeepSpeed/issues/2300#issuecomment-1255466874

PR #2354, opened 2022-09-24
Extend residual_add kernel tests to cover pre_attn_norm

arashb:
This PR extends the `residual_add_bias` kernel tests to also cover the cases where the layer norm is applied before/after attention layer.


awan-10 (approved):
Thanks! Looks good.

jeffra (approved):

PR #2356, opened 2022-09-25
 Refactor `fused_bias_residual` kernels for better readability

arashb:
This PR does not change the algorithm or the computation of the kernels (the tests are passing as before), and only: 
- renames the variables to more readable alternatives
- adds C++ safeguards

### DIFF
--- a/csrc/transformer/inference/csrc/gelu.cu
+++ b/csrc/transformer/inference/csrc/gelu.cu
@@ -154,120 +154,127 @@ void launch_bias_add(T* input, const T* bias, int hidden_size, int batch_size, c
 template void launch_bias_add<float>(float*, const float*, int, int, cudaStream_t);
 template void launch_bias_add<__half>(__half*, const __half*, int, int, cudaStream_t);
 
-__global__ void fused_bias_residual(float* input,
-                                    float* output,
-                                    float* attn,
-                                    float* bias,
-                                    float* attnbias,
-                                    int total_count,
-                                    int intermediate_size,
-                                    float mp_scale,
-                                    bool preln)
+__global__ void fused_bias_residual(float* residual,
+                                    const float* hidden_state,
+                                    const float* attn,
+                                    const float* bias,
+                                    const float* attn_bias,
+                                    const int total_count,
+                                    const int intermediate_size,
+                                    const float mp_scale,
+                                    const bool preln)
 {
-    float4* input_cast = reinterpret_cast<float4*>(input);
-    float4* output_cast = reinterpret_cast<float4*>(output);
-    float4* attn_cast = reinterpret_cast<float4*>(attn);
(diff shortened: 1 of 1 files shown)


csrc/transformer/inference/csrc/gelu.cu
```diff
@@ -154,120 +154,127 @@ void launch_bias_add(T* input, const T* bias, int hidden_size, int batch_size, c
...
-__global__ void fused_bias_residual(float* input,
-                                    float* output,
-                                    float* attn,
-                                    float* bias,
-                                    float* attnbias,
-                                    int total_count,
-                                    int intermediate_size,
-                                    float mp_scale,
-                                    bool preln)
+__global__ void fused_bias_residual(float* residual,
```
arashb, line 157: @RezaYazdaniAminabadi in these kernels the computation output is stored in `residual` as opposed to final `hidden_state`. Then as output of the MLP [here](https://github.com/microsoft/DeepSpeed/blob/993264388db97b63a77547c3092fc82fc9425ac7/deepspeed/ops/transformer/inference/transformer_inference.py#L646) we return the residual value. Wouldn't make sense to store the output in hidden_state  and return the hidden state?

cmikeh2: The underlying storage for `residual` is actually the original input tensor to the Transformer layer (which we don't otherwise modify), whereas `hidden_state` is backed by an offset into our scratchpad memory. That scratchpad section would be overwritten by another part of the Transformer (the allocation for hidden states is [here](https://github.com/microsoft/DeepSpeed/blob/993264388db97b63a77547c3092fc82fc9425ac7/csrc/transformer/inference/csrc/pt_binding.cpp#L1106) and conflicts with the QKV GEMM allocation [here](https://github.com/microsoft/DeepSpeed/blob/993264388db97b63a77547c3092fc82fc9425ac7/csrc/transformer/inference/csrc/pt_binding.cpp#L748)).

Long term I don't know how safe it necessarily is to assume that we can freely modify the input Tensor though, so modifying our scratchpad structure might be the most correct option? I don't know if anyone else thinks there's risk in using the input tensor this way though.

RezaYazdaniAminabadi: Thanks for bringing up this @arashb. Also, thanks @cmikeh2 for the main reason of not putting data in the hidden_state here. I think the safe way is probably to create another tensor for the output if this is needed in other places of the pipeline, however, this imposes more memory-usage of the inference system which is not desired. So, I think let's keep the default as writing in `residual` and if required by the users we can add a flag to create tensors for each layer's individual outputs. How does it sound?

RezaYazdaniAminabadi (approved):

jeffra (approved):

PR #2359, opened 2022-09-26
Fix issue with corrupted output on long generation for GPT

andrewchernyh:
Reopened and updated PR  https://github.com/microsoft/DeepSpeed/pull/2344
Fix issue described in https://github.com/microsoft/DeepSpeed/issues/2300

Only updating MAX_OUT_TOKES doesn't helps and main issue is in temp_buf that was allocated just after output tensor

Then in attention_unfused
T* workspace = (T*)output + bsz * seq_len * heads * k; where output is temp_buf from ds_softmax_context
In the result temp_buf overlaps query_cont allocated in ds_softmax_context

I've put temp_buf just after kv_cache and it helps

Minimal steps to reproduce:

`python3 benchmarks/inference/gpt-bench.py -m EleutherAI/gpt-neo-125M --kernel-inject --deepspeed --dtype=fp32 --max-tokens=1020 --trials 1`


andrewchernyh:
@microsoft-github-policy-service agree [company="Inworld.AI"]

andrewchernyh:
@microsoft-github-policy-service agree company="Inworld.AI"

csrc/transformer/inference/csrc/pt_binding.cpp
```diff
@@ -461,7 +461,7 @@ std::vector<at::Tensor> ds_softmax_context(at::Tensor& query_key_value,
     auto kv_cache = workspace + offset + (hidden_dim / heads) * (is_prompt ? 0 : soft_len - 1);
     size_t value_offset = bsz * Context::Instance().GetMaxTokenLenght() * hidden_dim;
 
-    T* temp_buf = (T*)output.data_ptr() + at::numel(output);
+    T* temp_buf = (T*)kv_cache+value_offset+value_offset;
```
RezaYazdaniAminabadi, line 464: by using this part of the memory, we are corrupting the cache of the consecutive layer. May I suggest keeping the temp unchanged and change the workspace used for the attention-score [here ](https://github.com/andrewchernyh/DeepSpeed/blob/fix-gpt-long-generation/csrc/transformer/inference/csrc/pt_binding.cpp#L357)to always point at the last portion of the memory buffer? Note, that the main issue is that we are currently writing and reading at the same place when doing the last GeMM in this function and by moving the workspace to the end of the buffer, we can be sure of not corrupting the memory.

csrc/transformer/inference/includes/inference_context.h
```diff
@@ -97,12 +97,13 @@ class Context {
         if (!_free_memory_size) { cudaMemGetInfo(&_free_memory_size, &total_size); }
 
         size_t activation_size = 16 * hidden_dim * batch_size;
+        size_t temp_size = 2 * hidden_dim * batch_size;
```
RezaYazdaniAminabadi, line 100: Can we change this to max_prompt_length * max_prompt_length  * batch_size * heads, which is the memory required for doing the attention part? We can either define a macro for max_prompt_length  and set it to a large-enough input-length (like 2K) or we can get it from the input? I would prefer the former.

RezaYazdaniAminabadi (changes requested):

csrc/transformer/inference/csrc/pt_binding.cpp
RezaYazdaniAminabadi: you can do this, by just keeping the else part.

csrc/transformer/inference/includes/inference_context.h
andrewchernyh: In such case GenWorkSpace requires num_heads, that is not provided by caller. It's easy to pass current prompt_length, but num_heads requires changing signature (args list) for `ds_qkv_gemm`, and `ds_linear_layer`, maybe it's better to make additional call for pt for internal memory allocation? With configuration, like prompt_length, max_sequence_length for model (2048 for GPT-neo for example)?

csrc/transformer/inference/csrc/pt_binding.cpp
andrewchernyh: Sure, will do it

csrc/transformer/inference/includes/inference_context.h
RezaYazdaniAminabadi: Got it, that makes sense. By the way, I think the max_sequence_length is not always limited to a number for all models. For example, in BLOOM architecture you can potentially produce as many tokens as needed. So, let's not get it from a config and we can just assume a large enough number like 8K (if the available memory allows).

andrewchernyh: I'll try to implement workspace allocation. And I think, it's better to write in docs, that DeepSpeed inference doesn't support parallel (in multiple threads) or interleaved generation (few tokens for one prompt, that few tokens for another, and few from first one), because internal state will be corrupted

RezaYazdaniAminabadi: I agree, but let's first finish this part as it is an urgent issue, and we can get to the other issue as you suggested. We can discuss it in another thread/PR.
Thanks,
Reza

RezaYazdaniAminabadi: Please let me know if you need help to finish this PR asap.

andrewchernyh: Memory allocation updated.

csrc/transformer/inference/csrc/pt_binding.cpp
andrewchernyh: Fixed

csrc/transformer/inference/csrc/pt_binding.cpp
```diff
@@ -352,15 +354,10 @@ void attention_unfused(T* prev_key_cont,
...
-    } else {
-        // If we are doing the prompt, switch to the tail workspace
-        T* scratch = (T*)Context::Instance().GetWorkSpace();
-        workspace = scratch + ((Context::Instance().get_workspace_size() / sizeof(T)) -
-                               bsz * heads * seq_len * soft_len);
-    }
+    // Always use the tail workspace
+    T* scratch = (T*)Context::Instance().GetWorkSpace();
+    T *workspace = scratch + ((Context::Instance().get_workspace_size() / sizeof(T)) -
+                            bsz * heads * seq_len * soft_len);
```
RezaYazdaniAminabadi, line 360: we should divide the number of heads by mp_size so as not to request for more memory.

csrc/transformer/inference/includes/inference_context.h
```diff
@@ -97,14 +99,21 @@ class Context {
         if (!_free_memory_size) { cudaMemGetInfo(&_free_memory_size, &total_size); }
 
         size_t activation_size = 16 * hidden_dim * batch_size;
+        size_t temp_size = batch_size * num_heads * prompt_len * prompt_len * elem_size;
```
RezaYazdaniAminabadi, line 102: divide the heads by mp_size here, too.

RezaYazdaniAminabadi (approved):

csrc/transformer/inference/includes/inference_context.h
andrewchernyh: mp_size is not available here, so I added GetAttentionUnfusedWorkspace to the Context

andrewchernyh: @RezaYazdaniAminabadi  Also I'm first-time contributor in DeepSpeed, and it's required to approve workflow running

andrewchernyh: fixed, all checks are passed

PR #2369, opened 2022-09-27
Add Unittest Builder, change bias_gelu unittest

lokoppakmsft:
This PR:
1) Creates a pt_binding file for functions that are only used for testing.
2) Creates new Inference UnittestBuilder to use with new binding file
3) changes test_bias_gelu test for reference implementation
4) Added pytest mark inference_unittest


op_builder/unittest_builder.py
```diff
@@ -0,0 +1,15 @@
+from .transformer_inference import InferenceBuilder
+
+
+class UnitTestBuilder(InferenceBuilder):
+    NAME = "deepspeed_unittest"
+
+    def __init__(self, name=None):
```
cmikeh2, line 7: The `__init__` implementation looks like matches the InferenceBuilder exactly, do we need to override it at all here?

csrc/transformer/inference/csrc/pt_binding_unittest.cpp
```diff
@@ -0,0 +1,33 @@
+/*
+Copyright 2022 The Microsoft DeepSpeed Team
+*/
+
+#include <c10/cuda/CUDAStream.h>
+#include <torch/extension.h>
+#include <stdexcept>
```
cmikeh2, line 7: Some of these includes aren't necessary, let's pull all that aren't necessary to compile.

cmikeh2 (commented):
I'm good with the basic structure of things. I think we might want to potentially add "Inference" to the name of the builder and pytest flags though to differentiate it from other tests modules we may want to add in the future to cover other parts of DeepSpeed, but others can probably make that determination.

op_builder/unittest_builder.py
lokoppakmsft: removed

csrc/transformer/inference/csrc/pt_binding_unittest.cpp
lokoppakmsft: removed unnecessary includes

lokoppakmsft:
> I'm good with the basic structure of things. I think we might want to potentially add "Inference" to the name of the builder and pytest flags though to differentiate it from other tests modules we may want to add in the future to cover other parts of DeepSpeed, but others can probably make that determination.

added Inference to the name of the builder

cmikeh2:
Moving in a different direction for backend inference refactoring.

PR #2371, opened 2022-09-27
Use CUDA events for inference model profiling

mrwyattii:


awan-10 (approved):
Looks great! @mrwyattii -- please make sure you manually compare the performance and also see if both gpt and bert style models are working.

PR #2380, opened 2022-09-30
Fix the MLP output tensor's shape

arashb:
This PR fixes the shape of the MLP layer's output tensor which was set to the size of intermediate weights instead of the size of output weights.


RezaYazdaniAminabadi (approved):

arashb:
The failed unittest is the usual gpt-j-6B OOM issue. CC: @mrwyattii

PR #2403, opened 2022-10-07
Allowing base64 encoded ds config passed as command line argument and using it for autotuner

cli99:
This PR allows user to pass in base64 encoded json dict as command line argument for "--deepspeed" or "--deespeed_config", and thus enabling the autotuner to launch jobs on remote workers without copying over the config file.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -924,12 +924,6 @@ class DeepSpeedEngine(Module):
                     args, "deepspeed_config") and args.deepspeed_config is not None
             ), "DeepSpeed requires --deepspeed_config to specify configuration file"
 
-            assert os.path.isfile(
-                args.deepspeed_config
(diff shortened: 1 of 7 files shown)


deepspeed/runtime/engine.py
```diff
@@ -924,12 +924,6 @@ def _do_args_sanity_check(self, args):
                     args, "deepspeed_config") and args.deepspeed_config is not None
             ), "DeepSpeed requires --deepspeed_config to specify configuration file"
 
-            assert os.path.isfile(
-                args.deepspeed_config
```
tjruwase, line 928: Rather than removing this assert, could we instead do correctness check for base64 encoded config? It is better for failures to occur early here during sanity check than later in DeepSpeedConfig constructor.

cli99: the autotuner no longer needs this encoding feature, closing the PR. thanks for the comment

PR #2409, opened 2022-10-09
deepspeed/launcher/launch.py: add option enable_each_rank_log

guoyejun:
It is not easy to check the console output because all the ranks output logs (with logs from underlying software libraries) into console. With this patch, the output of each rank are dumped into different log files.

### DIFF
--- a/deepspeed/launcher/launch.py
+++ b/deepspeed/launcher/launch.py
@@ -92,6 +92,12 @@ def parse_args():
                         default=0,
                         help="main launching process pid, for internal pid tracking")
 
+    parser.add_argument(
+        "--enable_each_rank_log",
+        default="None",
+        type=str,
+        help="redirect the stdout and stderr from each rank into different log files")
+
     # positional
     parser.add_argument("training_script",
                         type=str,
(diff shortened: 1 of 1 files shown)


deepspeed/launcher/launch.py
```diff
@@ -219,7 +223,13 @@ def main():
                 cmd.append(f"--local_rank={local_rank}")
             cmd += args.training_script_args
 
-            process = subprocess.Popen(cmd, env=current_env)
+            if args.enable_each_rank_log:
+                date_str = time.strftime("%Y%m%d%H%M%S", time.localtime())
+                log_file = os.path.join(PID_FILE_BASEPATH, f"{date_str}_rank{local_rank}.log")
```
jeffra, line 228: Really love this PR, thank you @guoyejun! Could I request two small changes though? Let's allow the log file path to be user configurable. Perhaps that path can be an argument to `--enable_each_rank_log`?

Also, can you use `dist_rank` in the filename instead of `local_rank`? This would allow someone to write out the log files to a shared filesystem and still be able to distinguish between global ranks in multi-node deployments.

jeffra:
One other request, in order to merge we need to address the formatting errors. Can you run our formatter? For details see: https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites

deepspeed/launcher/launch.py
guoyejun: sure, nice idea, will update the PR.

guoyejun:
looks that we are still waiting for the formatting and unit-tests, anything I need to do to trigger them? thanks.

guoyejun:
I take a look at the CI failure, and they are also there at both https://github.com/microsoft/DeepSpeed/pull/2429 and https://github.com/microsoft/DeepSpeed/pull/2428.  I think it is a CI system issue, not caused by this PR.

FAILED unit/inference/test_inference.py::TestModelTask::test[EleutherAI/gpt-j-6B-text-generation-fp16-noCG]

guoyejun:
any other comment? thanks.

jeffra (approved):
Thank you @guoyejun!

PR #2411, opened 2022-10-10
fixes #2389

### LINKED ISSUE #2389: [BUG] Bloated MoE expert checkpoint files with ZeRO 2
**Describe the bug**
It appears that with ZeRO 2 MoE layer expert parameters are replicated across all ranks by sharing a single tensor storage (with different offsets).
Since all MoE layer experts are stored in individual files for all ranks the default `torch.save()` behavior stores the underlying tensor's storage (even though a given expert is only covering a fraction of the storage).

Thus we end up with the following file structure per layer N:
```
layer_N_expert_E_mp_rank_00_model_states.pt
...
layer_N_expert_2_mp_rank_00_model_states.pt
layer_N_expert_1_mp_rank_00_model_states.pt
layer_N_expert_0_mp_rank_00_model_states.pt
```

Where _each_ file has the storage for all E experts across layer N, but the tensors in the files are only using a fraction of it.
When `torch.load()`ed each end up being tensors with independent storage.

It seems like using `.clone().detach()` when saving MoE expert tensors to individual files should address the issue.

But I was wondering about the reason behind making each expert and individual file for all ranks. E.g. if we stored a single file for expert layer for all ranks then the storage would have been reused for the loaded tensors and this single file would not be bloated.

I'll provide a test shortly, willing to contribute the fix.

**To Reproduce**

**Expected behavior**
Checkpoint files should be compact, no redundancy.

**ds_report output**

**Screenshots**

 - OS: Amazon Linux 4.14.252-195.483.amzn2.x86_64
 - GPU count and types: 8x A100 on a single machine
 - Interconnects (if applicable): single machine
 - Python version: 3.8
 - Any other relevant info about your setup: DeepSpeed 0.6.0, 0.7.4

**Launcher context**
DeepSpeed runner/launcher

**Docker context**
Custom docker image

**Additional context**
Using MoE layers, ZeRO 2

clumsy:
truncating expert param storage for checkpointing

### DIFF
--- a/tests/unit/simple_model.py
+++ b/tests/unit/simple_model.py
@@ -43,9 +43,16 @@ class SimpleMoEModel(torch.nn.Module):
     def __init__(self, hidden_dim, num_experts=4, ep_size=1, use_residual=False):
         super(SimpleMoEModel, self).__init__()
         self.linear = torch.nn.Linear(hidden_dim, hidden_dim)
-        linear2 = torch.nn.Linear(hidden_dim, hidden_dim)
+        expert = torch.nn.Linear(hidden_dim, hidden_dim)
+        # using two MoE layers to check implications of sharing a single storage
         self.linear2 = MoE(hidden_size=hidden_dim,
-                           expert=linear2,
+                           expert=expert,
+                           ep_size=ep_size,
+                           use_residual=use_residual,
+                           num_experts=num_experts,
+                           k=1)
+        self.linear3 = MoE(hidden_size=hidden_dim,
+                           expert=expert,
                            ep_size=ep_size,
                            use_residual=use_residual,
                            num_experts=num_experts,
@@ -53,9 +60,9 @@ class SimpleMoEModel(torch.nn.Module):
         self.cross_entropy_loss = torch.nn.CrossEntropyLoss()
 
     def forward(self, x, y):
-        hidden_dim = x
-        hidden_dim = self.linear(hidden_dim)
+        hidden_dim = self.linear(x)
         output, _, _ = self.linear2(hidden_dim)
+        output, _, _ = self.linear3(output)
         hidden_dim = hidden_dim + output
         sentence_embed = hidden_dim.mean(1)
         return self.cross_entropy_loss(sentence_embed, y)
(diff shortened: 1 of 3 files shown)


tests/unit/simple_model.py
```diff
@@ -43,19 +43,26 @@ class SimpleMoEModel(torch.nn.Module):
...
                            ep_size=ep_size,
                            use_residual=use_residual,
                            num_experts=num_experts,
                            k=1)
         self.cross_entropy_loss = torch.nn.CrossEntropyLoss()
 
     def forward(self, x, y):
-        hidden_dim = x
-        hidden_dim = self.linear(hidden_dim)
+        hidden_dim = self.linear(x)
```
tjruwase, line 63: What is the rationale for this change?

clumsy:
> 



> @clumsy please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> 

@microsoft-github-policy-service agree [company="Amazon"]

tests/unit/simple_model.py
tjruwase: Yes, you are correct. Thanks!

mrwyattii:
> @tjruwase Could you please help me identify why some of these unit tests fail? It's mostly related to sparse attention which is seemingly unrelated to my change. And there are also 2 tests (but not all) failing with `CUDA out of memory` for MoE checkpointing which I didn't get for my device.

@clumsy We were seeing problems with our self-hosted CI runners yesterday. I believe they are resolved now. I have re-run the failing jobs, and they should pass this time.

mrwyattii:
> Oh, that makes sense. Thanks, @mrwyattii! It looks like all but one check passed now. And the one that failed - looks like it failed the same way in a seemingly unrelated to my change place. Do we need to rerun it again?

Let me take a closer look at this today - not sure why it's failing

mrwyattii:
Looks like tests are passing, with exception of one inference test that is getting OOM error. It's unrelated to any changes in this PR, so I'll merge it anyway

PR #2426, opened 2022-10-15
Add TestInjectionPolicy inference unittest class for testing custom injection policies

lekurile:
This PR adds a `TestInjectionPolicy` inference unittest class for testing custom injection policies.

This test differs from the existing tests in that the `injection_policy` dictionary is explicitly specified when calling the DeepSpeed `init_inference` API.

The `google/t5-v1_1-small` text2text-generation model and the `roberta-large` fill-mask model are added as tests with the injection policy explicitly specified.

This is done to expand our unittest coverage to test the path where the `replace_wo_policy` function is invoked (see GH-2387).


tests/conftest.py
```diff
@@ -2,10 +2,13 @@
 
 import sys
 import pytest
+import os
 from os.path import abspath, dirname, join
 import torch
 import warnings
 
+os.environ['PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION'] = 'python'
```
mrwyattii, line 10: Can we add a comment above this line indicating for which test this is needed?

tests/unit/inference/test_inference.py
```diff
@@ -252,14 +266,28 @@ def test(
...
+                pipe.model,
+                mp_size=1,
+                dtype=dtype,
+                injection_policy={
+                    T5Block: ('SelfAttention.o',
+                              'EncDecAttention.o',
+                              'DenseReluDense.wo')
+                },
+                enable_cuda_graph=enable_cuda_graph,
+            )
```
mrwyattii, line 280: Adding additional if statements to the main test function could become messy. I think this is ok for now, but we should think about how we can have a single `init_inference` call.

mrwyattii (approved):
Add a comment to the one line and then good to merge!

tests/conftest.py
lekurile: Added a comment describing that this is needed for the T5 unittest.

tests/unit/inference/test_inference.py
lekurile: Thanks for the suggestion. After our discussion, we decided to refactor the testing to include a `TestInjectionPolicy` class, specifically designed to test explicit injection policies.

awan-10 (approved):
Great!

tests/unit/inference/test_inference.py
```diff
@@ -323,6 +334,66 @@ def test(
...
+        (("roberta-large",
+          "fill-mask"),
+         {
+             RobertaLayer: ('output.dense')
+         }),
+    ],
+    ids=["t5",
+         "roberta"],
+)
+@pytest.mark.parametrize("world_size", [1, 2])
```
mrwyattii, line 357: I'm surprised this works for our `DistributedTest` class, as it was certainly not designed to work this way. The `world_size` parameter should be a property of the class. See https://github.com/microsoft/DeepSpeed/blob/799120e7e44cefde7db8005295d40c2fa1eaf28d/tests/unit/common.py#L82

tests/unit/inference/test_inference.py
```diff
@@ -323,6 +334,66 @@ def test(
...
+@pytest.mark.parametrize("dtype", [torch.float], ids=["fp32"])
+@pytest.mark.parametrize("enable_cuda_graph", [False], ids=["noCG"])
```
mrwyattii, line 359: These are not required as the `dtype` and `enable_cuda_graph` in the test method definition have default values.

mrwyattii (changes requested):

tests/unit/inference/test_inference.py
lekurile: Updated to no longer parameterize the world_size.

tests/unit/inference/test_inference.py
lekurile: Removed the defaults from the test method and kept parametrize to override the fixtures.

mrwyattii (approved):

PR #2431, opened 2022-10-18
[memory estimators] new config args sync

stas00:
this PR updates the doc with demos' output of memory estimators to the current config params.

@jeffra, @tjruwase


tjruwase (approved):

stas00:
this is a doc-only change, you don't need to get the unit tests to run (since they don't run).

jeffra (approved):

PR #2432, opened 2022-10-18
only add deps if extra is explicitly called

jeffra:
Now that we have two features that depend on different triton versions we can no longer assume what triton version to install. Sparse attn was the only op that used this feature but previous to this PR we added triton to the dependency list iff it was detected as compatible during install time. This was causing version conflicts if we wanted to install the `sd` extra that requires triton 2.0.0. Also the `sd` extra was missing from setup.py, this PR adds it in.


mrwyattii (approved):
LGTM

PR #2436, opened 2022-10-20
Reduction Kernel Utility

cmikeh2:
This PR introduces a kernel utility for block-scope reductions. The purpose of this to provide a high performance, easy to use interface for the common reduction patterns we see in kernels.

### Usage

To give a usage example, let's look at how it might be used in asymmetric quantization. In asymmetric quantization, for a group of elements it is necessary to establish the range of the values by performing both a minimum and maximum reduction on the input. This accomplished by through a serial reduction within a thread and a parallel reduction across threads. This utility targets the parallel reduction:

```cpp
using rop = reduce::ROpType;
using cg = cooperative_groups;

__global__ void asym_quant(/* args */)
{
    cg::thread_block tb = cg::this_thread_block();
    cg::thread_block_tile<hw_warp_size> warp = cg::tiled_partition<hw_warp_size>(tb);

    float min_val, max_val;
    /* load data and perform serial reduction */
    /* `min_val` and `max_val` contain thread min/max */

    reduce::block<rop::Min, rop::Max>(tb, warp, min_val, max_val);
	
    /* `min_val` and `max_val` now contain the block-scoped min/max */
    /* quantize the data */
}
```

With this utility, block-scope reductions, even for multiple values in parallel, should be a one-line  function call.

### Limitations

1. We currently only support up to 4 parallel reductions together. This can be extended with some additional boilerplate code, but I am unaware of any use of this sort of reduction at this time.
2. Currently, the library implements the core reduction algorithm. There is an additional use case for something like mean that could be expected to fall under this reduction utility. However, this requires knowledge of how many values may have been serially reduced, and so blurs the lines between pure reduction and use-case specific. A mean reduction should probably be implemented on top of this to handle this common use case.
3. We don’t support partitioned reductions wherein multiple warps may cooperate on a reduction, but not all warps of the threadblock. This is a code pattern not common in something like layer norm or ZeroQuant-style quantization, but is more common in softmax where the span of reductions is typically smaller.


RezaYazdaniAminabadi (approved):

RezaYazdaniAminabadi:
LGTM. Thanks for all the great work here. Just wondering about the last limitation, can it not be configured so that we can use it in more places? I think we have fixed this for the softmax and we can get an idea from there, right?

cmikeh2:
> LGTM. Thanks for all the great work here. Just wondering about the last limitation, can it not be configured so that we can use it in more places? I think we have fixed this for the softmax and we can get an idea from there, right?

Correct, there's no technical reason to not support this, just need to add an interface and some minor tweaks to get there.

Issue #2319, opened 2022-09-13
[BUG] Can't load checkpoint without having shared filesystem in multi-node training when multi-node setup config remains same


pacman100:
**Describe the bug**
1. Background: Use deepspeed (use ZeRO-1) for multi-node training, save optimizers to resume training.
[...]

PR #2438, opened 2022-10-20
Fix Bug #2319

jomayeri:
Fixing bug [#2319 ](https://github.com/microsoft/DeepSpeed/issues/2319).

Adding a flag called `use_node_local_storage` to the ds_confg to allow for node local storage of zero and model states checkpoints instead of requiring a shared file system.


deepspeed/runtime/engine.py
```diff
@@ -2885,7 +2879,12 @@ def _checkpoint_tag_validation(self, tag):
             elif not valid:
                 logger.warning(msg)
 
-    def save_checkpoint(self, save_dir, tag=None, client_state={}, save_latest=True):
+    def save_checkpoint(self,
+                        save_dir,
+                        tag=None,
+                        client_state={},
+                        save_latest=True,
+                        use_node_local_storage=False):
```
tjruwase, line 2887: Thanks for working on this. However, this node_local_storage option is better configured through ds_config since it is constant for a given run. Also, users can activate with ds_config change rather than code change.  
This option should specifically go [here](https://github.com/microsoft/DeepSpeed/blob/99fde3b7a573f2a261dc5ef9380870ce30a1db9d/deepspeed/runtime/constants.py#L368).

jeffra:
Let's also add this new config to the config-json docs: https://github.com/microsoft/DeepSpeed/blob/master/docs/_pages/config-json.md

jeffra (approved):
Thanks @jomayeri, this is great.

tjruwase (approved):

PR #2445, opened 2022-10-25
Support iterators with incompletely defined __len__ functions

codedecde:
If data_iter has an incomplete definition of __len__, this would throw an error with the check. But as long as the __iter__ function works correctly, the semantics of the data_iterator are unaffected. Making the check more explicit for handling this.


deepspeed/runtime/pipe/engine.py
```diff
@@ -4,17 +4,14 @@
...
 from deepspeed.utils.timer import ThroughputTimer
 
-from ..engine import DeepSpeedEngine, MEMORY_OPT_ALLREDUCE_SIZE
-from ..utils import PartitionedTensor
 from ..dataloader import RepeatingLoader
-
-from .module import PipelineModule, PipelineError
-from . import p2p
-from . import schedule
+from ..engine import MEMORY_OPT_ALLREDUCE_SIZE, DeepSpeedEngine
```
tjruwase, line 11: Are these imports needed?

codedecde: These were already present. My linter just sorted them.

loadams: Merged changes, ran our formatter

loadams:
@codedecde - is this still a bug that you'd like to see fixed?

tjruwase (approved):

PR #2448, opened 2022-10-27
Implement apply_rotary_pos for rotate_half and rotary_dim > 32 in transformer inference

twaka:
Current implementation of `apply_rotary_pos` for `rotate_half` is optimized but only supports `rotary_dim` up to 32, because `__shfl_sync` cannot exchange value beyond warp size 32.

https://github.com/microsoft/DeepSpeed/blob/3b3ba3c2566a66d651393028af6c923b32eeffaa/csrc/transformer/inference/csrc/apply_rotary_pos_emb.cu#L195

This issue limits the applicability of deepspeed inference for certain models and also causes unexpected behavior.
To solve the issue, this PR implements `apply_rotary_pos` kernel for `rotate_half` and `rotary_dim > 32`.

### DIFF
--- a/csrc/transformer/inference/csrc/apply_rotary_pos_emb.cu
+++ b/csrc/transformer/inference/csrc/apply_rotary_pos_emb.cu
@@ -208,6 +208,53 @@ __global__ void apply_rotary_pos_emb1(__half* mixed_query,
 #endif
 }
 
+// update inplace mixed_query and key_layer, which are [B A S N]
+template <typename T>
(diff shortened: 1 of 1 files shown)


cmikeh2:
I'd like to apologize for not addressing this sooner! The current version of DS-Inference now includes an optimized implementation for rotate half supporting head sizes above 32 (and covering all common head sizes). If you believe you have a scenario that is not met by the current kernel, please re-open this PR.

PR #2450, opened 2022-10-27
Deepspeed quantization library v0.1

lokoppakmsft:
This PR introduced deepspeed quantization library. 
**This includes the following components:**

1. Support for Int4 and Int8
2. Support for both symmetric and asymmetric quantization

**Abstractions introduced:**
`quantize::GroupStats` : abstract container for characterizing the quantization group (abs-max for symmetric, min-max for asymmetric)
`quantize::Params` : abstract container for holding quantization parameters (scale, scale+offset)
`quantize::_chunk`: Quantize 16 bytes of data
`quantize::local_array` : Quantize a register-file persisted array and store it to global memory


cli99:
thanks @lokoppakmsft. why do we have two header files?

lokoppakmsft:
> thanks @lokoppakmsft. why do we have two header files?

the purpose of quantization.h is that only stuff in this header file like enums can be exposed to CPP files for example using enums in binding. If we merge these header files, cpp compiler might complain about CUDA code in it.

cmikeh2 (approved):

RezaYazdaniAminabadi (approved):

RezaYazdaniAminabadi:
Looks great, thanks :)

PR #2454, opened 2022-10-30
don't gather partitioned activations for mp size 1

guoyejun:


guoyejun:
the ci failure is caused by the ci system itself, I created an 'empty' PR at https://github.com/microsoft/DeepSpeed/pull/2455 and also see the same failures.

mrwyattii:
> the ci failure is caused by the ci system itself, I created an 'empty' PR at #2455 and also see the same failures.

I'm working on a fix for this. It seems a recent update to the default version of torch installed by pip has caused the error.

tjruwase:
@guoyejun, thanks for PR. Can you please describe the issue that this is fixing? And is it possible to include a unit test for this case? Thanks!

guoyejun:
hi @tjruwase it is for performance: 
if the model parallel size is 1 (no model parallel), we don't need to do all_gather for the activations.  

The reason that I did not add "or mp_size > 1" in below 'if' is that, we can remove some copy with my change.  Actually, the below 'if' can be moved ahead, it can be in another PR.
        if mp_group is not None:
            dist.all_gather(partitions, partitions[mp_rank], group=mp_group)

guoyejun:
> @guoyejun, thanks for PR. Can you please describe the issue that this is fixing? And is it possible to include a unit test for this case? Thanks!

for the unit test, can you share the current unit test for gather_partitioned_activations, actually, i think the that unit test might already test with mp size =1, I'll add it if not.

tjruwase (approved):

tjruwase:
@guoyejun, thanks!

PR #2451, opened 2022-10-28
Inference support for encoder decoder

RezaYazdaniAminabadi:

PR #2464, opened 2022-11-01
Create a new folder structure to isolate model-specific code in DS

awan-10:
This is a small refactor proposal so that we can continue our model-specific implementation and development under a single folder. I have moved the ones that I could think of. I see that Reza has a new PR #2451 that will add even more model-specific code and/or model implementations.

CC @jeffra @RezaYazdaniAminabadi @arashb @cmikeh2 

Hacks needed to get this right: 
- num_layers to be passed in DS Attention so we had to get rid of the static DeepSpeedTransformerInference.layer_id variable. 

EDIT: Reza and I reviewed the code. The num_layers hack is not needed as we can just switch from DSTransformer.layer_id to DSAttn.layer_id.


deepspeed/model_implementations/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .diffusers.unet import DSUNet
```
jeffra, line 1: We can't import this unless `diffusers` is available, since it's not a hard dependency of deepspeed. Similar for `transformers` we cannot assume it's installed

awan-10: Thanks! We don't need to import it in the __init__ so I removed it. The other two are not importing any external dependencies so they are okay to keep.

RezaYazdaniAminabadi (approved):

cmikeh2 (approved):

Issue #2352, opened 2022-09-23
[REQUEST] BF16 mixed precision => grad accum in fp32


stas00:
**Is your feature request related to a problem? Please describe.**

We have proven with the BLOOM training that BF16 is by far more superior for mixed precision training than FP16, using Megatron-Deepspeed

But the latter is very complex, it'd be much easier for folks to use standalone ZeRO for training in bf16 mixed precision.

But for this to work we need ZeRO to support grad accumulation in fp32, similar to the recently added `BF16Optimizer`.

So this is the feature request to **backport `BF16Optimizer`'s fp32 grad accumulation to ZeRO-1,2,3.**

Once this is done I can safely tell those who are interested in an easier to solution to use ZeRO.

@tjruwase, @jeffra

stas00:
@tjruwase, would it be possible to implement this? We are ready to start using ZeRO-3/bf16 for the multi-modal training.

Thank you very much!

tjruwase:
@stas00, is it better to close this or #2768? They are the same thing, right?

stas00:
Hi Tunji - you're the owner so it's up to you to decide. The new one is a duplicate of this one, so typically the earliest one stays.

And I don't agree with the other request that it should be hardcoded to fp32, it should be a user choice.

tjruwase:
Makes sense, will close the newer one and reference this appropriately. 

Yes, the accumulation type will be configurable. Hopefully, we should have a WIP pushed later this week. It would be great to get your usual feedback as we iterate on a solution.

stas00:
Fantastic news, Tunji. Thank you.

And, yes, we would be happy to experiment with your WIP PR.

michaelroyzen:
Amazing, thanks!

bestbzw:
hi I also met the same problem.

bestbzw:
hi I also met the same problem. @tjruwase have you found a solution?

danyang-rainbow:
Great, looking forward to see this new release!

michaelroyzen:
Any update on this @tjruwase?

stas00:
Please see: https://github.com/microsoft/DeepSpeed/pull/2847

tjruwase:
Closing as completed by #2847.

Issue #1835, opened 2022-03-15
Add explicit gradient_accumulation_dtype config


jeffra:
DeepSpeed has support for several dtypes now (i.e., fp32, fp16, bf16). However, it's becoming less clear what parts of training are using what dtypes and what time. For example, in #1801 we added support for BF16 training + FP32 gradient accumulation and optimizer stage sharding (zero stage 1) when pipeline parallelism is enabled. This is only triggered if your config is in the following scenario:

Note: PP is used on the client side and not in the ds_config, but it's use or not also decides what code paths are supported or not.

```json
# pipeline-parallelism: enabled
"bf16": {
   "enabled": true
},
"zero_optimization": {
    "stage": 0
}
```
--> BF16 training + FP32 gradient accumulation + ZeRO stage 1 optimizer sharding via `deepspeed/runtime/bf16_optimizer.py`

```json
# pipeline-parallelism: enabled
"bf16": {
   "enabled": true
},
"zero_optimization": {
    "stage": 1
}
```
--> BF16 training + BF16 gradient accumulation + ZeRO stage 1 optimizer sharding via `deepspeed/runtime/zero/stage_1_and_2.py`

The proposal is to introduce a config like the following:

```json
"bf16": {
   "enabled": true
},
"gradient_accumulation_dtype": "fp32",
"zero_optimization": {
    "stage": 1
}
```
--> 

The proposal is to add a new option in the ds_config: `gradient_accumulation_dtype`. In this case we would dispatch to the right version of ZeRO depending on what mode is selected by the user to make it more explicit what is happening.

I've started a table to try and express all of these possible cases and which ones would be supported and which would not. It feels a bit overly complicated in some ways however. This also doesn't consider cases where zero is disabled `"stage": 0`.

| bf16 | fp16 | grad-accu-dtype | PP | ZeRO (1,2,3)  | Result       | ZeRO implementation |
| ---- | ---- | --------------- | -- | ------ | ------------ | ------------------- |
| T    | T    | \*              | \* | \*     | Error |                     |
|      |      |                 |    |        |              |                     |
| T    | F    | fp16            | \* | \*     | NotSupported |                     |
| T    | F    | bf16            | \* | \*     | OKAY         | stage\_1\_and\_2.py |
| T    | F    | fp32            | T  | 1      | OKAY         | bf16\_optimizer.py  |
| T    | F    | fp32            | F  | 1      | NotSupported |                     |
[...]

PR #2470, opened 2022-11-03
Make bf16_optimizer work for non pipeline parallelism

tjruwase:
Initial step for #2352. Now, bf16_optimizer is enabled for the following ds_config
```json
   "zero_optimization": {
         "stage": 0,
         ...
   },
   "bf16": {
     "enabled": true
  },
 "fp16": {
   "enabled": false
}
...
```


jomayeri (approved):

jeffra:
I realized tonight that if we're opening non-PP setups to use bf16_optimizer it should make supporting a gradient accumulation dtype a bit easier (see table in #1835). Please take a look, but I think we could support `gradient_accu_dtype` which would help us map to the right backend zero implementation via the following below.

```python
assert not (self.fp16_enabled() and self.bfloat16_enabled()), "cannot enable both fp16 and bf16"
model_dtype = torch.float32
if fp16_enabled():
    model_dtype = torch.float16
elif bfloat16_enabled():
    model_dtype = torch.bfloat16

if model_dtype == self.gradient_accu_dtype():
    # -> stage_1_and_2.py
elif self.bf16_enabled() and self.gradient_accu_dtype() == torch.float32:
    # -> bf16_optimizer.py
else:
    raise NotImplementedError("unsupported mix of model dtype and gradient accu dtype")
```

This way someone's ds-config could be:

```json
"zero_optimization": {
    "stage": 1,
    ...
},
"bf16": {
    "enabled": true
},
"fp16": {
    "enabled": false
},
"gradient_accumulation_dtype": "bf16"
...
```

Issue #2473, opened 2022-11-03
[BUG] null default values in autotuning json causing failures


jomayeri:
**Describe the bug**
In the autotuning json config the default values for `results_dir` and `exps_dir` are listed for both as `null`. Using this in a json config results in the following error:
![image](https://user-images.githubusercontent.com/114769929/199814885-69541ee7-f0b4-4d25-8b95-a7c97c170684.png)

Removing  `results_dir` and `exps_dir`  and letting them use the internal default values produces the correct behavior.

**To Reproduce**
Repro file:
![image](https://user-images.githubusercontent.com/114769929/199815144-a7d42689-f2c4-465f-935b-9abb366987d4.png)

Repro config:
![image](https://user-images.githubusercontent.com/114769929/199815217-28683b11-9ae4-4d7d-bcee-37f4cd2fc8c3.png)

Repro Command 
`deepspeed --num_gpus=1 repro.py --deepspeed --deepspeed_config=ds_config.json`

**Expected behavior**
It seems that listing the actual default values instead of `null` as the default value would fix the confusion.

jomayeri:
Suggested fix: https://github.com/microsoft/DeepSpeed/compare/master...jomaayeri/debug-2473

PR #2476, opened 2022-11-04
Updating autotune json default in docs.

jomayeri:
Changing the defaults listed in the docs for the autotune json from `null` to the values found in constants.py. Addresses the issues found in bug [2373](https://github.com/microsoft/DeepSpeed/issues/2473).


cli99 (approved):
thanks!

PR #2481, opened 2022-11-07
fix accelerate link in README

kyoto7250:
Thank you for the great library :)

I noticed that the link to `accelerate` is broken.

This PR is to correct the link.

ref:
https://huggingface.co/docs/accelerate/usage_guides/deepspeed


jeffra (approved):
Thank you @kyoto7250!

PR #2483, opened 2022-11-07
Fix CI issues related to cupy install

mrwyattii:
We were seeing many errors related to installed version of `cupy` or multiple versions of `cupy` installed in our CI. The problem stems from cupy version needing to match with the CUDA runtime version rather than CUDA driver version. For example, if we have CUDA 11.3 runtime (from `nvcc --version`), CUDA 11.6 driver (from `nvidia-smi`), and pytorch installed with version 11.6, then in our `setup.py` we were incorrectly installing the `cupy` version to match the driver version (i.e., 11.6) since we detected the version from `torch.version.cuda`. In this PR I correct that and use the version from `nvcc --version`.

Additionally, a few other changes to update our CI:
- update torch latest from 1.12 -> 1.13
- CI runners for torch latest were updated from CUDA 11.3 -> 11.6
- move the python environment printout to after DeepSpeed install. This was causing trouble when debugging the problem initially, as installing DeepSpeed changes the python environment
- change nv-nightly test to use latest torch version (we are running inference tests here, so this should match nv-inference torch version)

### DIFF
--- a/.github/workflows/amd.yml
+++ b/.github/workflows/amd.yml
@@ -37,6 +37,12 @@ jobs:
           pip install --upgrade pip
           pip uninstall --yes torch torchvision triton
           pip install torch torchvision --extra-index-url https://download.pytorch.org/whl/rocm5.1.1
+          # Since the cupy install that happens in DS's setup.py gets the cuda
+          # version from pytorch/driver api (and not nvcc/runtime api), it's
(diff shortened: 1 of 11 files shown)


jeffra (approved):
Thank you @mrwyattii, this is great!

PR #2489, opened 2022-11-08
Make data contiguous before the inplace reshape-copy_ function.

lokoppakmsft:
Reshape returns a copy of original tensor when the data is not contiguous and not a view of the input tensor. This is causing the any inplace copy_ operation to not copy to the original tensor. This issue caused hugging face low_cpu_mem_usage feature to fail with deepspeed since a few parameters with `low_cpu_mem_usage` are not contigious and skipping the transposed copy,


lokoppakmsft:
@RezaYazdaniAminabadi , @jeffra, @awan-10

mrwyattii (approved):

Issue #2463, opened 2022-11-01
[BUG]Call torch.cuda.synchronize each time to reduce gradients in ZeRO Stage2


li-yi-dong:
**Describe the bug**
When training with Stage2, I noticed that `DeepSpeedZeroOptimizer` would synchronize the whole GPU before each time to launch a new `reduce`:
```Python
def average_tensor(self, tensor):
    if self.overlap_comm:
        torch.cuda.synchronize()
        stream = self.reduction_stream
    else:
        stream = torch.cuda.current_stream()

    with torch.cuda.stream(stream):
        if not self.reduce_scatter:
            self.gradient_reduction_w_predivide(tensor)
            return
        ...
```
Timeline:
![image](https://user-images.githubusercontent.com/73142299/199200577-f180dc45-011d-4e51-95a0-789c721609d9.png)

This reduces the opportunity for overlapping of the communication and hurts the performance a lot.
Why this synchronization is needed? Is it possible to remove it?

samyam:
Hi @li-yi-dong, the synchronize ensures that the `tensor` that we are about to reduce in the `average_tensor` already has the gradients in it. Without the synchronization, the reduction of the `tensor` can initiate on the reduction stream without the results of the gradient calculation being placed in the `tensor` by the current stream. 

Essentially what is happening is as follows:

```
with torch.cuda.stream(torch.cuda.current_stream()):
    tensor.copy_(compute_gradients())

#Ensure the compute gradients and copy complete before the reduction on a different stream
torch.cuda.synchronize()

with torch.cuda.stream(self.reduction_stream):
    dist.reduce(tensor)
```
    
The actual overlapping happens between the reduction and the next gradient computation happening later in the backward pass. 

Please let me know if you have any further questions.

li-yi-dong:
> Hi @li-yi-dong, the synchronize ensures that the `tensor` that we are about to reduce in the `average_tensor` already has the gradients in it. Without the synchronization, the reduction of the `tensor` can initiate on the reduction stream without the results of the gradient calculation being placed in the `tensor` by the current stream.
> 
> Essentially what is happening is as follows:
> 
> ```
> with torch.cuda.stream(torch.cuda.current_stream()):
>     tensor.copy_(compute_gradients())
> 
> #Ensure the compute gradients and copy complete before the reduction on a different stream
> torch.cuda.synchronize()
> 
> with torch.cuda.stream(self.reduction_stream):
>     dist.reduce(tensor)
[...]

PR #2500, opened 2022-11-12
Remove unnecessary device synchronization for stage 2

li-yi-dong:
#2463
I tested this with a simple example:
```Python
import torch
dist.init_process_group(backend='nccl')

stream = torch.cuda.Stream()


def trace_handler(p):
    p.export_chrome_trace("trace.json")


with torch.profiler.profile(
        with_stack=True,
        record_shapes=True,
        with_flops=True,
        activities=[
            torch.profiler.ProfilerActivity.CPU,
            torch.profiler.ProfilerActivity.CUDA,
        ],
        schedule=torch.profiler.schedule(
            wait=0,
            warmup=0,
            active=1,
            repeat=1
        ),
        on_trace_ready=trace_handler
) as prof:
    a = torch.rand((10000, 10000), device='cuda', dtype=torch.float)
    # warm up
    dist.all_reduce(a)

    for i in range(50):
        a.acos_()
        
    stream.wait_stream(torch.cuda.default_stream())
    
    with torch.cuda.stream(stream):
        a.asin_()
        dist.reduce(a, dst=0)
    
    torch.cuda.synchronize()
    prof.step()
```

Here's the result:
image


tjruwase:
@li-yi-dong, thanks for this PR. Could you please do some testing of impact of training convergence?

li-yi-dong:
> @li-yi-dong, thanks for this PR. Could you please do some testing of impact of training convergence?

@tjruwase 
I'm using it to train Stable Diffusion. Nothing strange happened. And I see 6% improvement in term of training speed.

li-yi-dong:
Is there any unit test or integrated test in DeepSpeed? It's my first time contributing. I'm not familiar with how to do the tests.

li-yi-dong:
Sorry, I found that the parameters may not be updated in some cases. I'll give it a double check🙃

tjruwase:
> Is there any unit test or integrated test in DeepSpeed? It's my first time contributing. I'm not familiar with how to do the tests.

Unfortunately, we don't have tests for this case, as it is quite subtle. However, I am encouraged that your model testing showed no issues.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -939,8 +939,8 @@ def gradient_reduction_w_predivide(self, tensor):
 
     def average_tensor(self, tensor):
         if self.overlap_comm:
-            torch.cuda.synchronize()
             stream = self.reduction_stream
+            stream.wait_stream(torch.cuda.default_stream())
```
tjruwase, line 943: Can you please explain why it waits for `default_stream()`, rather than `current_stream()` before line 942? What do you expect `default_stream()` to be?

li-yi-dong: I think in this context, `current_stream()` is `default_stream()`. I do agree `current_stream()` is a better choice.

tjruwase: Sounds good. Sorry for the delayed response.

tjruwase (approved):

li-yi-dong:
@tjruwase

li-yi-dong:
@tjruwase 
Hi, when will this PR be merged?

tjruwase:
@li-yi-dong, apologies for the delay. We had a CI issue over the holidays that caused a backlog of PRs to be merged. Your approved PR will be merged as soon as the CI issue is resolved. Thanks!

PR #2504, opened 2022-11-13
Abstract accelerator (step 1)

delock:
This PR is step 1 of fully merge #2221.

Because #2221 touches a lot of files and could often cause conflict when merge with master, so we break this PR into multiple step 1 for a smooth merge process.

Step 1: Add interface for abstract accelerator,  an CUDA implementation of abstract accelerator, and op_builder preparation for op_builder abstraction.
Step 2: Port deepspeed code in small batches, each batch control in the quantity of 20 files.
Step 3: Port deepspeed tests and benchmarks in small batches
Step 4: fix any hard coded CUDA code in newly introduced code, add a format checker for unported code in newly introduced code and add the checker in workflow.

For whole code change and design of abstract accelerator support in DeepSpeed, refer to #2221


tjruwase (approved):

jomayeri (approved):

mrwyattii (approved):

deepspeed/accelerator/abstract_accelerator.py
```diff
@@ -0,0 +1,198 @@
...
+    def __init__(self):
+        self._name = None
+        self._communication_backend_name = None
+        self.BFloat16Tensor = None
+        self.ByteTensor = None
+        self.DoubleTensor = None
+        self.FloatTensor = None
+        self.HalfTensor = None
+        self.IntTensor = None
+        self.LongTensor = None
```
mrwyattii, line 15: Are these values required to be defined for all inheriting classes? It could be useful to use `@abstractproperty` for them, so we ensure all accelerators define them.

delock: `_name` and `_communication_backend_name` already has get function defined so they should not be defined as `@abstractproperty`.   The rest tensor type class should be defined for all inheriting class and can be defined as `@abstractproperty`.

Note: the following names are not actually used by DeepSpeed, they are defined for compleness:
`BFloat16Tensor`, `DoubleTensor`, `HalfTensor`

delock: Updated according to comments

jeffra:
Thank you @delock! This is great. I do have one comment/question though.

I am thinking about the work required to add new ops into DeepSpeed. From what I understand in this PR a developer will need to modify the following files with their new op builder, right?

op_builder/all_ops.py
op_builder/builder_names.py
op_builder/__init__.py
(for a cuda op): deepspeed/accelerator/cuda_accelerator.py 

I see a note that maybe we can remove the op references in `op_builder/__init__.py`?

For the builder names is it possible to take advantage of the [builder name](https://github.com/microsoft/DeepSpeed/blob/78d4ca1f4bb2f0e8a7f2cfd052b0e7733ab41898/op_builder/transformer_inference.py#L6) itself instead redefining it in `op_builder/builder_names.py`?

My primary concern is when we add new ops it will be hard to remember all the places that need changes, previously this was isolated to only `op_builder/__init__.py` and we had a TODO to even make that automatically picked up via the op_builder folder.

delock:
Thanks for the comments @jeffra

The builder files involved in add a new builder has two parts, one part is the abstraction part and another part is accelerator part.   With this in mind, we can divide the files like the following:

- abstraction part
`op_builder/all_ops.py`
`op_builder/builder_names.py`

- accelerator specific part
`deepspeed/accelertor/cuda_accelerator.py`
`op_builder/__init__.py`   (The part below TODO will be eventurally removed when merge is fully done)
builder files that is under op_builder/*.py

for xpu extension, we also have `xpu_accelerator.py` and `op_builder` directory that holds xpu specific part.
`intel_extension_for_deepspeed/xpu_accelerator.py`
`intel_extension_for_deepspeed/op_builder/__init__.py`
builder files that is under `op_builder/*.py`

I think the desire in automation is mainly in the following parts:
1. Whether we can define ALL_OPS without list builders manually?
2. in `xpu_accelerator.py` or `cuda_accelerator.py`, can we maintain the list of builders automatically, when new builders are defined in the extension?
3. Whether the relationship between symbol and builder name can be associated automatically?

For 1, I think its doable with certain reflection mechanism and naming convention.  We will need to be able to infer op_builder directory, this can be done by let each accelerator extension report their op_builder directory.  After that we can iterate through modules under op_builder directory and their members to find out op_builder_name with certain naming pattern, then we can append get_accelerator().create_op_builder(op_builder_name) to ALL_OPS.   The logic looks like below:
```
op_builder_dir = get_accelerator().op_builder_dir()
for file in *.py files that is not 'builder.py' under op_builder_dir
    for op_builder_name with *Builder pattern in file
        additionally check whether op_builder_name represents a subclass of OpBuilder if possible
        builder = get_accelerator().create_op_builder(op_builder_name)
        ALL_OPS.append(builder)
```
This mechanism should solve concern 1.

For 2, I think its better to let extension owner to decide what to do, there is only two rules to follow.  a) put all builder files in a single place so reflection mechanism in 1 can enumerate all the builders in the extension.  b) in create_op_builder return builder object according to name.  It should be quite natural to follow.   The extention owner could also optionally use a similar mechanism as 1 to associate name to builder class, this is up to extension owner.

For 3, we can use similar mechanism in 1, and use self reflection to add symbol-name relationships
for builder_names
```
import sys
this_module = sys.modules[__name__]
for each op_builder_name reflected
    this_modules.__dict__[op_builder_name] = op_builder_name
```

So if we follow up 1 and 3, developers don't need to maintain `all_ops.py` and `builder_names.py`

> Thank you @delock! This is great. I do have one comment/question though.
> 
> I am thinking about the work required to add new ops into DeepSpeed. From what I understand in this PR a developer will need to modify the following files with their new op builder, right?
> 
> op_builder/all_ops.py op_builder/builder_names.py op_builder/**init**.py (for a cuda op): deepspeed/accelerator/cuda_accelerator.py
> 
> I see a note that maybe we can remove the op references in `op_builder/__init__.py`?
> 
> For the builder names is it possible to take advantage of the [builder name](https://github.com/microsoft/DeepSpeed/blob/78d4ca1f4bb2f0e8a7f2cfd052b0e7733ab41898/op_builder/transformer_inference.py#L6) itself instead redefining it in `op_builder/builder_names.py`?
> 
> My primary concern is when we add new ops it will be hard to remember all the places that need changes, previously this was isolated to only `op_builder/__init__.py` and we had a TODO to even make that automatically picked up via the op_builder folder.

delock:
The latest commit follow up this comment.
> Thanks for the comments @jeffra
> 
> The builder files involved in add a new builder has two parts, one part is the abstraction part and another part is accelerator part. With this in mind, we can divide the files like the following:
> 
> * abstraction part
>   `op_builder/all_ops.py`
>   `op_builder/builder_names.py`
> * accelerator specific part
>   `deepspeed/accelertor/cuda_accelerator.py`
>   `op_builder/__init__.py`   (The part below TODO will be eventurally removed when merge is fully done)

delock:
Hi @jeffra does the latest change resolve your comments?

jeffra (approved):
This is great, thank you @delock!

delock:
Hi @tjruwase if there is no other comments can we have this PR merged?   Then I will update #2221 and start to create PR for accelerator interface integration code.

tjruwase:
> Hi @tjruwase if there is no other comments can we have this PR merged? Then I will update #2221 and start to create PR for accelerator interface integration code.

Will merge once CI completes. Thanks!

deepspeed/accelerator/abstract_accelerator.py
```diff
@@ -0,0 +1,223 @@
...
+    @abc.abstractmethod
+    def lazy_call(self, callback):
+        ...
+
+    @abc.abstractmethod
+    def communication_backend_name(self):
+        ...
+
+    # Tensor operations
+    @abc.abstractproperty
```
mrwyattii, line 177: It looks like `abstractproperty` is actually deprecated now and it is suggested we use
```python
@property
@abstractmethod
```

https://docs.python.org/3/library/abc.html#abc.abstractproperty

delock: changed as suggested.   Because these property are read only so we don't need additional setter for them.

PR #2509, opened 2022-11-15
Update docs to autogenerate pydantic config model docs

mrwyattii:
We are moving our configs to use pydantic. As part of this work we also want to have autogenerated docs. This PR adds that to the code docs and demonstrates these features on the `DeepSpeedZeroConfig` and `DeepSpeedInferenceConfig`.

The changes can be seen here:
https://deepspeed.readthedocs.io/en/rtd-staging/inference-init.html#deepspeed.inference.config.DeepSpeedInferenceConfig
https://deepspeed.readthedocs.io/en/rtd-staging/zero3.html#zero-configurations


deepspeed/runtime/config_utils.py
```diff
@@ -47,6 +47,9 @@ class MyExampleConfig(DeepSpeedConfigModel):
                                       new_param_fn=(lambda x: int(x)))
     """
     def __init__(self, strict=False, **data):
+        """
+        test one two
```
tjruwase, line 51: What is this documenting?

mrwyattii: haha thanks for catching this :)

tjruwase (approved):

PR #2511, opened 2022-11-15
Add unittest to chek huggingface low_cpu_mem_usage_flag

lokoppakmsft:


tests/unit/inference/test_inference.py
```diff
@@ -337,6 +337,55 @@ def test(
         assert assert_fn(bs_output, ds_output)
 
 
+@pytest.mark.seq_inference
+@pytest.mark.parametrize("model_w_task",
+                         [("EleutherAI/gpt-j-6B",
```
mrwyattii, line 342: I think we should use a different model here. I recently moved our `gpt-j-6B` test to multi-GPU because we were often seeing OOM errors when running on a single GPU (see #2457 and #2493)

mrwyattii (changes requested):
Looks like we need to add `accelerate` to our installs on the inference CI workflows based on the error message.

Also, what is the motivation for this test? Was inference breaking with `low_cpu_mem_usage` enabled?

lokoppakmsft:
> Looks like we need to add `accelerate` to our installs on the inference CI workflows based on the error message.
> 
> Also, what is the motivation for this test? Was inference breaking with `low_cpu_mem_usage` enabled?

It was breaking before [this fix](https://github.com/microsoft/DeepSpeed/commit/f2710bbe1dc19717b3c99e910cb453ea20ae8b08), and we saw a lot of interest in this feature, so adding this unittest to not regress in the support

loadams:
Closing in favor of #4184

PR #2512, opened 2022-11-16
Adding Gradient Accumulation Data Type Config

jomayeri:
Adding a gradient accumulation data type config parameter. The goal is to make it more explicit via the DeepSpeed config which parts of training are using which data types.

Partial fix for #1835


deepspeed/runtime/engine.py
```diff
@@ -1135,44 +1151,52 @@ def _configure_optimizer(self, client_optimizer, model_parameters):
...
+                model, self.optimizer = amp.initialize(
+                    self.module, basic_optimizer, **amp_params
+                )
+                self._set_client_model(model)
+                self._broadcast_model()
+                # TODO: maybe need to broadcast experts differently?
+            elif self.fp16_enabled():
+                self.optimizer = self._configure_fp16_optimizer(basic_optimizer)
+            elif self.bfloat16_enabled():
+                self.optimizer = self._configure_bf16_optimizer(basic_optimizer)
```
jeffra, line 1191: need to remove this case

jeffra:
Need to add unit tests to test basic functionality for different modes and if NotImplementedErrors are raised for the cases we don't support

deepspeed/runtime/config.py
```diff
@@ -905,6 +909,10 @@ def _initialize_params(self, param_dict):
             USE_NODE_LOCAL_STORAGE_CHECKPOINT,
             USE_NODE_LOCAL_STORAGE_CHECKPOINT_DEFAULT)
 
+        data_types_params = get_data_types_params(param_dict)
```
jeffra, line 912: @mrwyattii to confirm if we have pydantic support for the main runtime config yet?

mrwyattii: We do not for the main runtime config. That will be a larger refactor, but I would like to prioritize it and get all the configs using pydantic.

deepspeed/runtime/engine.py
```diff
@@ -78,6 +78,8 @@
 from deepspeed.profiling.flops_profiler.profiler import FlopsProfiler
 from deepspeed.utils.logging import print_json_dist
 
+from deepspeed.inference.config import DtypeEnum
```
mrwyattii, line 81: We should probably move the `DTypeEnum` to `deepspeed/runtime/config_utils.py` or `deepspeed/runtime/config.py` if it's being used in multiple configs. We can do that in this PR or a future PR when we refactor the main runtime config to use pydantic.

deepspeed/runtime/config.py
jeffra: No problem, just wanted to make sure we didn't add a new config using the old approach if it was already converted. I imagine the main runtime config will be a bigger undertaking than the others.

deepspeed/runtime/engine.py
```diff
@@ -1106,6 +1125,62 @@ def ids_list(group):
...
+            if model_dtype == torch.float16:
+                return 'fp16'
+            # else optimizer_implementation = 'basic'
+        elif model_dtype == torch.bfloat16 and grad_accum_dtype == torch.float32:
+            return 'bf16'
+        else:
+            raise NotImplementedError(
+                "unsupported mix of model dtype and gradient accummulation type")
+
+        return 'basic'
```
tjruwase, line 1182: Does `basic` have a special meaning? Could we not just return `None`

deepspeed/runtime/engine.py
```diff
@@ -1106,6 +1125,62 @@ def ids_list(group):
...
+            if not is_zero_supported_optimizer(basic_optimizer):
+                assert (
+                    self.zero_allow_untested_optimizer()
+                ), 'You are using an untested ZeRO Optimizer. Please add <"zero_allow_untested_optimizer": true> in the configuration file to use it.'
+
+                if self.global_rank == 0:
+                    logger.warning(
+                        "**** You are using ZeRO with an untested optimizer, proceed with caution *****"
+                    )
+            return 'zero'
```
tjruwase, line 1152: Return `ZERO_OPTIMIZATION` from  https://github.com/microsoft/DeepSpeed/blob/999aa66550cf59b189608980364e3a305c6e315a/deepspeed/runtime/zero/config.py#L42

jomayeri: Ah thanks I was looking for that one.

tjruwase (approved):

PR #2522, opened 2022-11-18
Fixes for torch 1.14 due to new torch.numel return type

jeffra:
For some reason torch 1.14 changes the return type of `torch.numel(..)` to be a Tensor instead of an `int`. This change caused several issues in our flops profiler and onebit tests. See previously failing tests here: https://github.com/microsoft/DeepSpeed/actions/runs/3484228630/jobs/5828554740

This PR should fix these issues by either reverting to `x.numel()` or converting the Tensor returned by `torch.numel(x)` to an int.

image

### DIFF
--- a/deepspeed/profiling/flops_profiler/profiler.py
+++ b/deepspeed/profiling/flops_profiler/profiler.py
@@ -478,47 +478,40 @@ def _prod(dims):
     return p
 
 
-def torch_numel(tensor):
-    n = torch.numel(tensor)
-    if isinstance(n, torch.Tensor):
-        n = n.item()
-    return n
-
-
 def _linear_flops_compute(input, weight, bias=None):
     out_features = weight.shape[0]
-    macs = torch_numel(input) * out_features
+    macs = input.numel() * out_features
     return 2 * macs, macs
 
 
 def _relu_flops_compute(input, inplace=False):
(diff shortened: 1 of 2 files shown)


deepspeed/profiling/flops_profiler/profiler.py
```diff
@@ -478,40 +478,47 @@ def _prod(dims):
...
+    n = torch.numel(tensor)
+    if isinstance(n, torch.Tensor):
+        n = n.item()
+    return n
+
+
 def _linear_flops_compute(input, weight, bias=None):
     out_features = weight.shape[0]
-    macs = torch.numel(input) * out_features
+    macs = torch_numel(input) * out_features
```
tjruwase, line 490: Would `input.numel()` not work?

jeffra: It should? I wasn't sure the reasoning behind using `torch.numel()` in this case so opted to preserve the original style. @cli99 might know more here?

cli99: input.numel() works as well

jeffra: Cool! I’ll switch the PR to that style then

PR #2523, opened 2022-11-18
Set zero stage to 0 in autotuning when prescale gradients is used

cli99:
In autotuning, the profiling run uses zero stage 3 which currently does not support prescale gradients. Thus this PR does this: when prescale_gradients is true, the autotuning profiling run uses zero stage 0 (ZeRO disabled).


mrwyattii (approved):

loadams:
Closing as stale for now.

PR #2217, opened 2022-08-13
Ds-inference Int8 support through ZeroQuant technology

RezaYazdaniAminabadi:
This PR adds the Int8 support using the ZeroQuant technology introduced [here ](https://arxiv.org/abs/2206.01861). Note that the kernels added in this PR is just doing a dequantization of the int8-weight matrices and the GeMM operation is still done in the FP16 format using FP16 tensor-core ops. However, the kernels mentioned in this paper can directly operate on the int8 data, by using the int8 tensor-cores if available (such as on A100), which can improve the GeMM throughput by 2x. There will be another PR which uses these kernels to give a much faster inference through MII system. 

Here is some performance evaluation:

FP16 on 8 A100-80G:
bsz | latency | tput | tput-per-GPU | Tflops/GPU
-- | -- | -- | -- | --
1 | 60.8 | 16.5 | 2.1 | 1.4
8 | 58.7 | 136.2 | 17.0 | 12.0
32 | 64.6 | 495.5 | 61.9 | 43.6
64 | 68.4 | 936.2 | 117.0 | 82.4
100 | 78.4 | 1274.7 | 159.3 | 112.2

INT8 on 4 A100-80G:
bsz | latency | tput | tput-per-GPU | Tflops/GPU
-- | -- | -- | -- | --
1 | 160.5 | 6.2 | 1.6 | 0.5
8 | 165.1 | 48.5 | 12.1 | 4.3
32 | 178.7 | 179.1 | 44.8 | 15.8
64 | 190.7 | 335.6 | 83.9 | 29.5
100 | 213.0 | 469.4 | 117.4 | 41.3

INT8 on 8 A100-80G:
bsz | latency | tput | tput-per-GPU | Tflops/GPU
-- | -- | -- | -- | --
1 | 92.9 | 10.8 | 1.3 | 0.5
8 | 95.5 | 83.7 | 10.5 | 3.7
32 | 107.5 | 297.6 | 37.2 | 13.1
64 | 116.5 | 549.6 | 68.7 | 24.2
128 | 133.4 | 959.3 | 119.9 | 42.2
196 | 166.7 | 1175.5 | 146.9 | 51.7
210 | 170.3 | 1232.8 | 154.1 | 54.2
240 | 177.9 | 1349.1 | 168.6 | 59.4
256 | 181.8 | 1408.2 | 176.0 | 62.0

cc: @stas00 @jeffra

PR #2526, opened 2022-11-18
Add 4-bit quantized inference to run BLOOM-176B on 2 A100 GPUs

RezaYazdaniAminabadi:
This PR adds the support for 4-bit quantization at DeepSpeed-Inference to be able to run large-scale model such as BLOOM-176B using 2x/4x lower number of GPUs compared to INT8/FP16 inference pipeline.

As the first evaluation of the accuracy on **2 A100 GPUs**, we see some good quality text generated using the below prompt:

``` 
in=DeepSpeed is a machine learning framework 
out=DeepSpeed is a machine learning framework in C. The aim of the project is to create an accurate and efficient algorithm, 
which is easy to expand and improve. It is very useful in the classification of an input. The key to the success of the project is the
use of the features of the C programming language. The project has a lot of code, but the most important part of their is in the 
C language,
```
### How to run inference
You can find the running scripts [here](https://github.com/RezaYazdaniAminabadi/transformers-bloom-inference/tree/int4-bloom/bloom-inference-scripts). For running the model using 4-bits, you can start with a [FP16](https://huggingface.co/microsoft/bloom-deepspeed-inference-fp16) or [INT8](https://huggingface.co/microsoft/bloom-deepspeed-inference-int8) checkpoint, and the DeepSpeed-Inference pipeline can generate the 4-bit checkpoint on-the-fly and run the inference on as low as 2 A100-80G GPUs. Here is the command used to generate the above text:
```
deepspeed --num_gpus 2 bloom-ds-inference.py --name microsoft/bloom-deepspeed-inference-int8 --dtype int4 --batch_size 1 --benchmark
```
Here is the performance stats for running on 2 A100 GPUs:
```
*** Performance stats:
Throughput per token including tokenize: 283.44 msecs
Start to ready to generate: 120.887 secs
Tokenize and generate 438 (bs=1) tokens: 27.788 secs
Start to finish: 148.675 sec
```
Compared to 4 GPU [int8-performance](https://github.com/microsoft/DeepSpeed/pull/2217), the latency increases from 160.5 ms to 283.4 ms for batch-1 while reducing the number of GPUs by 2x, which reduces the inference cost by about 13%.

We will add more performance results to check the throughput improvement.

### More to come

1. Even though this is only supported for BLOOM, we are working on generalizing this feature for the rest of models in order to reduce the inference cost and democratize the inference of such huge models.
2. We will add some more tests to measure the accuracy metrics of the Int4 inference pipeline.
3. We will add faster inference solution using the int-4 quantized inference through DeepSpeed-MII.

cc: @jeffra @yaozhewei @cmikeh2


deepspeed/model_implementations/transformers/ds_transformer.py
```diff
@@ -128,7 +128,7 @@ def forward(
             input = input[0]
         input_type = input.dtype
 
-        if (self.config.fp16 or self.config.q_int8) \
+        if (self.config.fp16 or self.config.qunatize) \
```
mrwyattii, line 131:
```suggestion
        if (self.config.fp16 or self.config.quantize) \
```

groenenboomj:
@RezaYazdaniAminabadi Is there an updated version of this PR? I'm having some issues running out of the box on CUDA using the following related branch in transformers: 
https://github.com/RezaYazdaniAminabadi/transformers-bloom-inference/tree/int4-bloom

PR #2534, opened 2022-11-22
Ensure MOE is initialized for SD

cmikeh2:
Since the pipeline for SD is not a `torch.nn.Module`, we need to ensure that the local `moe` variable is set in the else condition. For now, since we don't support MOE with SD I'm setting it to `False`.

@RezaYazdaniAminabadi @awan-10 @jeffra


jeffra (approved):
Good catch @cmikeh2!

PR #2536, opened 2022-11-22
Change Where DS/Triton is Used in Stable Diffusion

cmikeh2:
This PR makes some tweaks to when we use Triton/DS kernels for diffusers attention. The performance impact of this change should be < 1% with a large reduction in visual difference between HF and DS outputs.


jeffra:
If this is configurable can we have it default off for CC < 8.0? I believe currently on older gpus triton fails

RezaYazdaniAminabadi (approved):

Issue #2532, opened 2022-11-22
Make DS inference config readable from a json file


awan-10:
@mrwyattii -- Please take a look at this issue for context: https://github.com/huggingface/transformers-bloom-inference/pull/37/files

@stas00 wants us to make the new DS inference config accept ```dict_or_path``` just like zero does?

PR #2537, opened 2022-11-22
Make DS-Inference config readable from JSON

mrwyattii:
Addresses #2532

- `config` can now be a path to a JSON file
- Includes a refactor of how we handle cases where both `config` and `kwargs` are passed to `init_inference` to fix the case where `kwargs` can overwrite `config` values
- Add unit tests for `DeepSpeedInferenceConfig`


awan-10 (approved):
Looks great!

PR #2538, opened 2022-11-22
Remove PP Grad Tail Check

Quentin-Anthony:
After the patch in https://github.com/microsoft/DeepSpeed/pull/1400 for BigScience, the final element of the `inputs` tuple is conditional on whether its grad is null (https://github.com/microsoft/DeepSpeed/blob/v0.7.5/deepspeed/runtime/pipe/engine.py#L995).

This will always fail if `elt.grad is None` in (https://github.com/microsoft/DeepSpeed/blob/v0.7.5/deepspeed/runtime/pipe/engine.py#L1026) with a `IndexError: tuple index out of range` because the `index_grad_tail` in (https://github.com/microsoft/DeepSpeed/blob/v0.7.5/deepspeed/runtime/pipe/engine.py#L1006) doesn't exist.

We're facing this issue intermittently whenever the grad tail is full of `nan`. I propose we just always include the grad tail.

@dashstander @jeffra @ShadenSmith


deepspeed/runtime/pipe/engine.py
```diff
@@ -990,9 +990,7 @@ def _exec_send_grads(self, buffer_id):
             if isinstance(inputs, tuple):
                 first_input = inputs[0]
                 assert all([torch.is_tensor(elt) for elt in inputs[1:]])
-                inputs_grad_tail = [
-                    elt.grad for elt in inputs[1:] if elt.grad is not None
-                ]
+                inputs_grad_tail = [elt.grad for elt in inputs[1:]]
```
ShadenSmith, line 993: It's been ages since I worked on this code unfortunately...could the extra `None`s result in mismatched number of communications in the backward pass from stage? I think that was the original intention. Our first version of pipelined GPT in Megatron used floating-point attention masks, but once switching to non-differentiable types things got trickier.

StellaAthena: We are seeing errors happen because of the lack of `None`s currently, as shown in https://github.com/EleutherAI/gpt-neox/issues/884

loadams:
Hi @Quentin-Anthony - is this an active PR you'd still like to see merged?  Just going through some older PRs and trying to see if we should review/prioritize or close.

Quentin-Anthony:
> Hi @Quentin-Anthony - is this an active PR you'd still like to see merged? Just going through some older PRs and trying to see if we should review/prioritize or close.

Yes I'm still interested! I'll take another look sometime over the next few days and decide how best to close this out.

loadams:
> > Hi @Quentin-Anthony - is this an active PR you'd still like to see merged? Just going through some older PRs and trying to see if we should review/prioritize or close.
> 
> Yes I'm still interested! I'll take another look sometime over the next few days and decide how best to close this out.

Thanks!  Other than a current known failure, it is passing all tests at least, but we will just want to figure out who would be best to review.

RUAN-ZX:
Hi, I have met the same problem as well. I wonder if we can move forwards and merge this solution ASAP?

tjruwase (approved):

PR #2539, opened 2022-11-22
Pass down the new DS inference config to replace_transformer_layer.

awan-10:
This PR is the next in the series of PRs we will need to pass down the new DS Inference config from init -> inference engine -> replace_transformer_layer etc. 

@RezaYazdaniAminabadi @mrwyattii @cmikeh2 -- please review this :)

### DIFF
--- a/deepspeed/inference/engine.py
+++ b/deepspeed/inference/engine.py
@@ -344,34 +344,7 @@ class InferenceEngine(Module):
                           enable_cuda_graph=config.enable_cuda_graph)
 
         if isinstance(self.module, torch.nn.Module):
-            replace_transformer_layer(
-                client_module,
-                self.module,
-                triangular_masking=config.triangular_masking,
-                policy=config.injection_policy_tuple,
-                mp_size=config.tensor_parallel.tp_size,
-                mp_group=self.mp_group,
-                ep_group=self.ep_group,
-                expert_mp_group=self.expert_mp_group,
-                config=self.config,
-                fp16=(config.dtype == torch.half) or (config.dtype == torch.int8),
-                training=False,
-                return_tuple=config.return_tuple,
-                quantize=(config.dtype == torch.int8),
-                quantize_settings=(self.quantization_scales,
-                                   self.quantize_merge_count,
-                                   self.mlp_extra_grouping,
-                                   self.quantize_groups),
-                replace_with_kernel_inject=config.replace_with_kernel_inject,
-                moe=config.moe,
-                moe_experts=config.moe.moe_experts,
-                moe_type=config.moe.moe_type,
-                training_mp_size=config.training_mp_size,
-                checkpoint_dict=checkpoint,
-                save_mp_checkpoint_path=config.save_mp_checkpoint_path,
-                base_dir=config.base_dir,
-                enable_cuda_graph=config.enable_cuda_graph,
-                max_out_tokens=config.max_out_tokens)
+            replace_transformer_layer(client_module, self.module, checkpoint, config)
 
     def _get_all_ckpt_names(self, checkpoints_path, tag):
         ckpt_file_pattern = self._get_ckpt_name(checkpoints_path,
--- a/deepspeed/module_inject/replace_module.py
+++ b/deepspeed/module_inject/replace_module.py
@@ -286,74 +286,33 @@ def generic_injection(module, fp16=False, enable_cuda_graph=True):
                 setattr(module, name, new_module)
 
 
-def replace_transformer_layer(orig_layer_impl,
-                              model,
-                              policy=None,
-                              micro_batch_size=-1,
-                              config=None,
-                              seed=-1,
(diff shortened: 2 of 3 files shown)


deepspeed/module_inject/replace_module.py
```diff
@@ -468,23 +429,24 @@ def replace_with_policy(child,
...
                     mlp_act_func_type=policy.mlp_act_func_type,
-                    training_mp_size=training_mp_size,
+                    training_mp_size=config.training_mp_size,
                     bigscience_bloom=bigscience_bloom,
-                    max_out_tokens=max_out_tokens,
+                    max_out_tokens=config.max_out_tokens,
                     scale_attn_by_inverse_layer_idx=scale_attn_by_inverse_layer_idx)
 
-            if quantize and quantize_settings is not None:
+            if quantize and config.quantize_settings is not None:
```
mrwyattii, line 437: Is `config.quantize_settings` a parameter? I think you want `config.quant`. Also, the default value for `config.quant` is `{}` and this is not equivalent to `None`

deepspeed/module_inject/replace_module.py
```diff
@@ -468,23 +429,24 @@ def replace_with_policy(child,
...
                     new_module = transformer_inference.DeepSpeedMoEInference(
                         transformer_config,
-                        mp_group=mp_group,
-                        ep_group=None if ep_group is None else ep_group[num_experts],
-                        expert_mp_group=None
-                        if expert_mp_group is None else expert_mp_group[num_experts],
+                        mp_group=config.tensor_parallel.tp_group,
+                        ep_group=None if config.moe.ep_group is None else
+                        config.moe.ep_group[num_experts],
+                        expert_mp_group=None if config.moe.expert_mp_group is None else
```
mrwyattii, line 448: `config.moe.expert_mp_group` should be `config.moe.ep_mp_group`

deepspeed/module_inject/replace_module.py
```diff
@@ -468,23 +429,24 @@ def replace_with_policy(child,
...
+                    max_out_tokens=config.max_out_tokens,
                     scale_attn_by_inverse_layer_idx=scale_attn_by_inverse_layer_idx)
 
-            if quantize and quantize_settings is not None:
+            if quantize and config.quantize_settings is not None:
                 (quantization_scales,
                  merge_count,
                  mlp_extra_grouping,
-                 quantize_groups) = quantize_settings
+                 quantize_groups) = config.quantize_settings
```
cmikeh2, line 441: As Mike mentioned above, I think this should refer to `config.quant` (or do we want to add an alias to `quantize_settings`?)

deepspeed/module_inject/replace_module.py
awan-10: oh, good catch. quantize_settings is a local thing and that has to be cleaned up in a separate PR after Reza's int4 PR gets merged. 

Let me fix this part differently for now.

deepspeed/module_inject/replace_module.py
awan-10: Actually, thinking a bit more about this new name. I think it should be called ```config.moe.expert_tp_group``` (as we use it for expert's tensor slicing or tensor parallelism degree). @RezaYazdaniAminabadi -- what do you think?

deepspeed/module_inject/replace_module.py
awan-10: Removing quantize_settings completely as we don't use that code path anymore based on Reza's feedback.

deepspeed/module_inject/replace_module.py
awan-10: We can decide on this later. For now, I just followed Mike's suggestion.

RezaYazdaniAminabadi (approved):

deepspeed/inference/engine.py
```diff
@@ -346,34 +345,12 @@ def _apply_injection_policy(self, config, client_module=None):
...
-                save_mp_checkpoint_path=config.save_mp_checkpoint_path,
-                base_dir=config.base_dir,
-                enable_cuda_graph=config.enable_cuda_graph,
-                max_out_tokens=config.max_out_tokens)
+            # config is our DeepSpeedInferenceConfig and self.config is the HF model config
+            replace_transformer_layer(client_module,
+                                      self.module,
+                                      checkpoint,
+                                      config,
+                                      self.config)
```
mrwyattii, line 353: `config` and `self.config` is a little bit confusing, maybe we can address this in a future PR?

mrwyattii: or perhaps use `inf_config` or `self._config`

mrwyattii (approved):

deepspeed/inference/engine.py
awan-10: Sounds good. We do need a future PR to clean this part. The self.config is coming from huggingface.

PR #2540, opened 2022-11-22
Remove mutable default parameter in `init_inference()`

aphedges:
A mutable default value is dangerous because editing it will change the value for all future calls to the function. The value is itself edited later in the function, so this problem will likely be encountered sooner or later.

---

I noticed this when was looking over the recent changes. This problem was just introduced in 8b4318b950d09d975c62cd8f4285821fa670ce87.


mrwyattii (approved):
Thank you!

aphedges:
You're welcome!

PR #2545, opened 2022-11-23
Automatic tensor parallelism for HF models

molly-smith:
Adds module_parser.py for parsing modules and generating injection policy dictionaries.

Adds parser_policies.py which stores new policy dictionaries for over 100 new huggingface models.

Modifies deepspeed inference engine. Adds new replace method option "dict" which enables automatic selection of corresponding injection policy based on model.

@awan-10 
@RezaYazdaniAminabadi


deepspeed/inference/engine.py
```diff
@@ -122,7 +124,31 @@ def __init__(self, model, config):
...
+                        new_key = SML(child)
+                    if new_key is not None:
+                        return new_key
+
+            new_key = SML(model)
+            if new_key is not None:
+                key = str(new_key).split('models.')[1].split('.')[0]
+            assert key in ParserPolicies.parser_policy_map, "dict replace method not supported for this model"
+            for client_module, injection_policy in ParserPolicies.parser_policy_map[key].items():
+                client_module = import_string('transformers.models.' + key +
```
jeffra, line 144: I’m not familiar with the ‘import_string’ package, but would could importlib be used here? I am always wary of adding new dependencies :)

https://docs.python.org/3.9/library/importlib.html#importlib.import_module

molly-smith: I don't think import_module will work here because it can only import modules and I need a class object returned here which is not included in any module.

jeffra: Can we do what they’re recommending here: 

https://github.com/rochacbruno-archive/import_string

I see the package is deprecated as well, so if we can avoid using it that would probably be ideal

deepspeed/module_inject/test_module_parser.sh
```diff
@@ -0,0 +1,9 @@
+#Runs through all huggingface models and runs module_parser
+
+set -x
+
+model_dir="./transformers/src/transformers/models"
+
+for d in $model_dir/*/modeling_*.py; do
+	python module_parser.py -f $d
```
jeffra, line 8: Would this style test be a good candidate to add into our unit tests? Might be a good way to detect when `transformers` makes changes that break our assumptions?

molly-smith: In a next phase, I'm going to work on changing the parser such that we parse the model structure from the pipeline instead of the transformers repo. That way we don't need to store injection policies and instead generate them at runtime. And we shouldn't need to check for changes to the transformer repo.

deepspeed/inference/engine.py
jeffra: Awesome, saw the updates, thanks @molly-smith!

jeffra:
Let's add a new unit test that triggers this code path and uses this model: https://huggingface.co/Salesforce/codegen-2B-mono

molly-smith:
Closing PR since we are skipping to v2: https://github.com/microsoft/DeepSpeed/pull/2670

PR #2550, opened 2022-11-24
Fix invalid check of recorded parameter orders in zero stage3.

inkcherry:
In zerostage3, class ```PartitionedParameterCoordinator ``` . the variable ```__param_order``` is assigned by ```construct_parameter_trace_from_module_trace()```. These two checks are earlier than the assignment. Although these checks could pass. Ranks will always broadcast or compare the empty tensors, the trace check is invalid.


tjruwase (approved):

inkcherry:
@tjruwase  Thanks for your review. It seems that ut failure is not caused by the code of the pr, and similar problems have occurred in other prs. would you please check it?

PR #2551, opened 2022-11-25
Fix bugs in l_aux and exp_counts calculation in moe top2gating

yyyyyt123:
Hi DeepSpeed Team,

There are some errors in the calculation of `l_aux` and `exp_counts` in the `top2gating` function in `deepspeed/moe/sharded_moe.py`. Since it only uses the result of `mask1` to generate the `exp_counts` and the `l_aux` (`mask1` is the 1st's expert choice per token).
This patch fixes this by adding `mask2` in the calculation (the 2nd expert choice per token).

Please verify that the rest of DeepSpeed is still working after applying this patch.

### DIFF
--- a/deepspeed/accelerator/__init__.py
+++ b/deepspeed/accelerator/__init__.py
@@ -1,0 +1,2 @@
+from .abstract_accelerator import DeepSpeedAccelerator
+from .real_accelerator import get_accelerator, set_accelerator
--- a/deepspeed/accelerator/abstract_accelerator.py
+++ b/deepspeed/accelerator/abstract_accelerator.py
@@ -1,0 +1,230 @@
+import abc
+from abc import ABC
+
+
+class DeepSpeedAccelerator(ABC):
+    def __init__(self):
+        self._name = None
+        self._communication_backend_name = None
+
+    # Device APIs
+    @abc.abstractmethod
+    def device_name(self, device_index):
+        ...
+
+    @abc.abstractmethod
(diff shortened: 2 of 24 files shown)


awan-10:
@yyyyyt123 - thanks for the PR. Do you mind sharing what kind of example/test you are using to verify the correctness of your proposed solution? 

@ykim362 -- kindly review this change and approve if it looks good to you. It looks okay to me but I don't fully understand/remember the background of this one. The original fairscale code does not use this method for l_aux so I need your input. Thanks!

ykim362:
@awan-10 
I think it's a design choice, but the original Gshard paper also used only mask1 for the aux loss.

@yyyyyt123 
By any chance, do you have experimental results that using both mask1 and mask2 together give a better performance?

https://arxiv.org/pdf/2006.16668.pdf

image

ykim362:
When I thought about it again, this change would affect the training dynamics quite a bit. (back prop on the first routing choice only vs.  back prop on the first & second choice compounded.)

awan-10 (approved):

awan-10 (changes requested):
@yyyyyt123 -- please add a boolean flag and protect this new code with that so you can use it but default it to False. Something like def top2gating(..,use_both_masks=False,..).

yyyyyt123:
@awan-10 Thanks for the reply! I will add a boolean flag to the top2gating like your mentioned.

PR #2559, opened 2022-12-01
Fix bugs in exp_count and add a boolean flag on whether to use both masks in calculating l_aux

yyyyyt123:
1. Correct the calculation of `exp_count` in top2gating.

2. Add a boolean flag on whether to use both masks in calculating `l_aux`

@awan-10 I'm so sorry that I made some misoperations in recorrecting the pr [#2551](https://github.com/microsoft/DeepSpeed/pull/2551). So I open a new pull request of same content as [#2551](https://github.com/microsoft/DeepSpeed/pull/2551).

### DIFF
--- a/deepspeed/moe/sharded_moe.py
+++ b/deepspeed/moe/sharded_moe.py
@@ -277,7 +277,8 @@ def top1gating(logits: Tensor,
 
 def top2gating(logits: Tensor,
                capacity_factor: float,
-               min_capacity: int) -> Tuple[Tensor,
+               min_capacity: int,
+               use_both_masks=False) -> Tuple[Tensor,
                                            Tensor,
                                            Tensor,
                                            Tensor]:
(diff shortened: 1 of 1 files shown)


deepspeed/moe/sharded_moe.py
```diff
@@ -309,12 +310,18 @@ def top2gating(logits: Tensor,
     locations2 += torch.sum(mask1, dim=0, keepdim=True)
 
     # gating decisions
-    exp_counts = torch.sum(mask1, dim=0).detach().to('cpu')
+    exp_counts = torch.sum(mask1 + mask2, dim=0).detach().to('cpu')
```
ykim362, line 313: I think this needs to follow the config?

yyyyyt123: @ykim362  Thanks for your reply! But I think only the calculation of `l_aux` should follow the config. I thought the `exp_count` shows how many tokens are dispatched to the experts in different routing choices. I profiled the tokens dispatch pattern of the top1&top2 gate by printing the return value `self.deepspeed_moe.exp_counts` in `moe/layer.py/MoE/forward`. I found that in the `top2gating` function, the `exp_count` only counts the first routing choice (ignore the second routing choice). So I think `exp_count` might not have to follow the config?

yyyyyt123: I think in the `top2gating` function, the `exp_count` should count both the first&second expert choice of each token so that we can get the correct dispatch pattern. Feel free to correct me if I say something wrong.

jomayeri:
@yyyyyt123 Please respond if you would like to keep this PR open. Otherwise I will close it. Thanks.

PR #2560, opened 2022-12-01
Abstract accelerator (step 2)

delock:
This PR is step 2 of fully merge https://github.com/microsoft/DeepSpeed/pull/2221.

This PR integrated deepspeed accelerator interface in the following files:
1. setup related files
2. files that use DeepSpeed op builders.
3. Changes in accelerator interface that fix the issue that cannot install deepspeed when pytorch is not installed


delock:
The test error in nv-mii / unit-tests seems to be an error in the test itself, the error says BloomModel is not compatible with generate

https://github.com/microsoft/DeepSpeed/actions/runs/3598044999/jobs/6060428284#step:6:213

E               	details = "Exception calling application: The current model class (BloomModel) is not compatible with `.generate()`, as it doesn't have a language model head. Please use one of the following classes instead: {'BloomForCausalLM'}"
E               	debug_error_string = "UNKNOWN:Error received from peer ipv4:127.0.0.1:50050 {created_time:"2022-12-01T17:37:37.209027217-08:00", grpc_status:2, grpc_message:"Exception calling application: The current model class (BloomModel) is not compatible with `.generate()`, as it doesn\'t have a language model head. Please use one of the following classes instead: {\'BloomForCausalLM\'}"}"

A later message says BloomModel is not supported fro text-generation.

The model 'BloomModel' is not supported for text-generation. Supported models are ['BartForCausalLM', 'BertLMHeadModel', 'BertGenerationDecoder', 'BigBirdForCausalLM', 'BigBirdPegasusForCausalLM', 'BlenderbotForCausalLM', 'BlenderbotSmallForCausalLM', 'BloomForCausalLM', 'CamembertForCausalLM', 'CodeGenForCausalLM', 'CTRLLMHeadModel', 'Data2VecTextForCausalLM', 'ElectraForCausalLM', 'ErnieForCausalLM', 'GPT2LMHeadModel', 'GPTNeoForCausalLM', 'GPTNeoXForCausalLM', 'GPTNeoXJapaneseForCausalLM', 'GPTJForCausalLM', 'MarianForCausalLM', 'MBartForCausalLM', 'MegatronBertForCausalLM', 'MvpForCausalLM', 'OpenAIGPTLMHeadModel', 'OPTForCausalLM', 'PegasusForCausalLM', 'PLBartForCausalLM', 'ProphetNetForCausalLM', 'QDQBertLMHeadModel', 'ReformerModelWithLMHead', 'RemBertForCausalLM', 'RobertaForCausalLM', 'RoCBertForCausalLM', 'RoFormerForCausalLM', 'Speech2Text2ForCausalLM', 'TransfoXLLMHeadModel', 'TrOCRForCausalLM', 'XGLMForCausalLM', 'XLMWithLMHeadModel', 'XLMProphetNetForCausalLM', 'XLMRobertaForCausalLM', 'XLMRobertaXLForCausalLM', 'XLNetLMHeadModel'].

deepspeed/env_report.py
```diff
@@ -79,31 +80,35 @@ def nvcc_version():
...
-        ("deepspeed wheel compiled w.",
-         f"torch {torch_info['version']}, " +
-         (f"hip {torch_info['hip_version']}"
-          if hip_version else f"cuda {torch_info['cuda_version']}")),
+         )
     ]
+    if get_accelerator().device_name() == 'cuda':
+        hip_version = None
+        if hasattr(torch.version, 'hip'):
+            hip_version = torch.version.hip
```
mrwyattii, line 97:
```suggestion
        hip_version = getattr(torch.version, "hip", None)
```

mrwyattii (changes requested):

deepspeed/env_report.py
delock: will make this change.

delock:
@mrwyattii code had updated according to both comments.  Thanks!

deepspeed/__init__.py
```diff
@@ -6,308 +6,308 @@
...
+        ds_inference_config = DeepSpeedInferenceConfig(**config_dict)
 
-    ds_inference_config = DeepSpeedInferenceConfig(**config_dict)
+        engine = InferenceEngine(model, config=ds_inference_config)
 
-    engine = InferenceEngine(model, config=ds_inference_config)
+        return engine
 
-    return engine
+except ImportError:
```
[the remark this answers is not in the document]
tjruwase: @delock, torch is a hard requirement for using deepspeed. Can you please explain the scenario you are trying to support here? Thanks!

delock:
Hi @mrwyattii are there any other comments?   Previous comments had been resolved with changes.

tjruwase:
@delock, thanks for the updates. I think you might have missed my question about wrapping torch import in `try...except`. Can you please explain the scenario that requires this, since torch is a hard requirement for DeepSpeed?

delock:
> @delock, thanks for the updates. I think you might have missed my question about wrapping torch import in `try...except`. Can you please explain the scenario that requires this, since torch is a hard requirement for DeepSpeed?

Hi @tjruwase, thanks for reminding.  During deepspeed installation step, pytorch is not needed prior to deepspeed installation.   At this step, DeepSpeed needs to get full list of op builders with the following line:
https://github.com/microsoft/DeepSpeed/pull/2560/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R34
`from op_builder.all_ops import ALL_OPS`

Because op builders are provided by accelerator interface extension, which could be provided by an external module, so in `op_builder/all_ops.py`, `deepspeed.accelerator.get_accelerator()` needs to be used to create op builders specific to accelerator from external module.   So this import line triggers execution of `deepspeed/__init__.py`
https://github.com/microsoft/DeepSpeed/blob/master/op_builder/all_ops.py#L7, which contains `import torch`, which triggers build error.

The change in this PR wrap part of `deepspeed/__init__.py` code that use torch, which allows installation get full list of op builders without pytorch installed.  If this change is removed, will get the following error during installation if torch is not installed.

```(base) gma@gma-mobl2:~/DeepSpeedSYCLSupport$ python setup.py install
[WARNING] Unable to import torch, pre-compiling ops will be disabled. Please visit https://pytorch.org/ to see how to properly install torch on your system.
 [WARNING]  unable to import torch, please install it if you want to pre-compile any deepspeed ops.
  File "/home/gma/DeepSpeedSYCLSupport/setup.py", line 34, in <module>
    from op_builder.all_ops import ALL_OPS
  File "/home/gma/DeepSpeedSYCLSupport/op_builder/all_ops.py", line 7, in <module>
  File "/home/gma/DeepSpeedSYCLSupport/deepspeed/__init__.py", line 9, in <module>
    import torch
ModuleNotFoundError: No module named 'torch'
```

If we have these lines, installation can success.

Here's some more information we get from deepspeed installation process:
Before deepspeed installation, pytorch does not need to be installed, the pre-CI tests of deepspeed does not have pytorch pre-installed as well.   During deepspeed installation, pytorch will be installed from requirements automatically.  This line in deepspeed also indicates that torch is not always needed before installation.  https://github.com/microsoft/DeepSpeed/blob/master/setup.py#L26

delock:
Hi @tjruwase did this answer your question?  Any other question/comments?

> > @delock, thanks for the updates. I think you might have missed my question about wrapping torch import in `try...except`. Can you please explain the scenario that requires this, since torch is a hard requirement for DeepSpeed?
> 
> Hi @tjruwase, thanks for reminding. During deepspeed installation step, pytorch is not needed prior to deepspeed installation. At this step, DeepSpeed needs to get full list of op builders with the following line: https://github.com/microsoft/DeepSpeed/pull/2560/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R34 `from op_builder.all_ops import ALL_OPS`
> 
> Because op builders are provided by accelerator interface extension, which could be provided by an external module, so in `op_builder/all_ops.py`, `deepspeed.accelerator.get_accelerator()` needs to be used to create op builders specific to accelerator from external module. So this import line triggers execution of `deepspeed/__init__.py` https://github.com/microsoft/DeepSpeed/blob/master/op_builder/all_ops.py#L7, which contains `import torch`, which triggers build error.
> 
> The change in this PR wrap part of `deepspeed/__init__.py` code that use torch, which allows installation get full list of op builders without pytorch installed. If this change is removed, will get the following error during installation if torch is not installed.
> 
> ```
> [WARNING] Unable to import torch, pre-compiling ops will be disabled. Please visit https://pytorch.org/ to see how to properly install torch on your system.
>  [WARNING]  unable to import torch, please install it if you want to pre-compile any deepspeed ops.
>   File "/home/gma/DeepSpeedSYCLSupport/setup.py", line 34, in <module>
>     from op_builder.all_ops import ALL_OPS
>   File "/home/gma/DeepSpeedSYCLSupport/op_builder/all_ops.py", line 7, in <module>
>     from deepspeed.accelerator import get_accelerator
>   File "/home/gma/DeepSpeedSYCLSupport/deepspeed/__init__.py", line 9, in <module>
>     import torch
> ModuleNotFoundError: No module named 'torch'
> ```
> 
> If we have these lines, installation can success.
> 
> Here's some more information we get from deepspeed installation process: Before deepspeed installation, pytorch does not need to be installed, the pre-CI tests of deepspeed does not have pytorch pre-installed as well. During deepspeed installation, pytorch will be installed from requirements automatically. This line in deepspeed also indicates that torch is not always needed before installation. https://github.com/microsoft/DeepSpeed/blob/master/setup.py#L26

tjruwase:
@delock, thanks for the explanation. I think I understand the problem now. It seems to me that most of `deepspeed/__init__.py` is not needed for op_builder, and it was written for engine initialization which occurs after client calls `deepspeed.initialize()`. So, I think we need to split `deepspeed/__init__.py` into two parts: (i) engine initialization, and (ii) non-engine initialization, such as op building, symbol lookup, etc.  Please share your thoughts.  We will also discuss internally.

delock:
Hi @tjruwase.   Yes the root cause for such construction is that op_builder is used at two different stages:
1. during deepspeed installation stage, at this stage deepspeed engine does not expected to be functional, we can see that for op_builder to be accessible, op_builder are accessed through op_builder, rather than deepspeed.ops.op_builder.   I believe this is to decouple op_builder from deepspeed engine during installation time.
2. during deepspeed execution stage, at this stage, op_builder are accessed through deepspeed.ops.op_builder, which is installed as part of deepspeed package.   From the code we see there is symlink of op_builder and csrc under deepspeed/ops, these symlinks exist because they are needed at both build stage and execution stage.

So I think to resolve deepspeed/__init__.py issue, we can build a symlink of accelerator following the way op_builder and csrc does.   accelerator itself should not depend on deepspeed engine, so it could exist independent of deepspeed engine at build time, then we don't need construction in deepspeed/__init__.py.
> @delock, thanks for the explanation. I think I understand the problem now. It seems to me that most of `deepspeed/__init__.py` is not needed for op_builder, and it was written for engine initialization which occurs after client calls `deepspeed.initialize()`. So, I think we need to split `deepspeed/__init__.py` into two parts: (i) engine initialization, and (ii) non-engine initialization, such as op building, symbol lookup, etc. Please share your thoughts. We will also discuss internally.

delock:
@tjruwase  @mrwyattii is there any remaining issue needs to be resolved for this PR?   Thanks!

tjruwase (approved):

tjruwase:
@delock, thanks! This looks good to me.

delock:
Hi @mrwyattii, there is still 'change required' status of this PR, but the change requst had already been responded.  Is this state needs to be manually resolved before this PR can be merged?   Thanks!

tjruwase:
@delock, apologies for the delay. We are eager to merge this PR and actually holding back other dependent PRs to avoid further delay. @mrwyattii should be back from vacation tomorrow.

delock:
@tjruwase thanks!   Glad to know its on the queue.  Looking forward to fully integrate accelerator abstraction into DeepSpeed.
> @delock, apologies for the delay. We are eager to merge this PR and actually holding back other dependent PRs to avoid further delay. @mrwyattii should be back from vacation tomorrow.

mrwyattii (approved):

tohtana:
Hi @delock,  

I installed deepspeed using pip with -e option, and found that the changes in rev 9548d48 cause the following error: 

```
File "/DEEPSPEED_REPO_ROOT/op_builder/builder.py", line 453, in load
from ...git_version_info import installed_ops, torch_info 
ImportError: attempted relative import beyond top-level package 
```

I encountered this error after I ran `run_ds.sh` in [CIFAR in DeepSpeedExamples](https://github.com/microsoft/DeepSpeedExamples/tree/master/cifar).
But I think you can reproduce it with many other programs using deepspeed.

When we have deepspeed installed with -e, we have builder.py in both `deepspeed/ops/op_builder/` and `op_builder/` (the former is a symbolic link). 
I guess Python interpreter recognizes the path of builder.py as `op_builder/` at runtime and reference to `...git_version_info` fails. 

Do you have any idea how to fix this?

jeffra:
@delock would this change work in your environment? It seems to fix the issue on my tests:

```diff
diff --git a/op_builder/builder.py b/op_builder/builder.py
index 39544f8c..a1730abc 100644
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -450,7 +450,7 @@ class OpBuilder(ABC):
             extra_link_args=self.strip_empty_entries(self.extra_ldflags()))

     def load(self, verbose=True):
-        from ...git_version_info import installed_ops, torch_info
+        from deepspeed.git_version_info import installed_ops, torch_info
         if installed_ops[self.name]:
             # Ensure the op we're about to load was compiled with the same
             # torch/cuda versions we are currently using at runtime.
```

tohtana:
I confirmed that the fix proposed by @jeffra works and submitted it as PR #2687.

PR #2564, opened 2022-12-02
Update the locator for Megatron-LM

rapsealk:
Hello, DeepSpeed Team!

I found that the official document is guiding us to the old location of Megatron-LM tutorial.


jeffra (approved):
Thank you!

PR #2566, opened 2022-12-02
Support fp32 gradaccum for bf16 model

delock:
Recently deepspeed added sanity check for optimizer as the following link
https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L1133

However, FP32 gradient accumulation data type is the default gradient accumulation type for bfloat16 model.   DeepSpeed code itself also set gradient accumulation data type to FP32 for BF16 model.
https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L811

This PR allows this combination in sanity check, so bfloat16 model can pass sanity check with default configuration.

The following artical mentioned using FP32 accumulation type for BFloat16 datatype.
https://cloud.google.com/blog/products/ai-machine-learning/bfloat16-the-secret-to-high-performance-on-cloud-tpus

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1130,9 +1130,14 @@ class DeepSpeedEngine(Module):
         ), "Amp and ZeRO are not currently compatible, please use (legacy) fp16 mode which performs similar to amp opt_mode=O2"
         if zero_enabled:
             if model_dtype != grad_accum_dtype:
-                raise NotImplementedError(
-                    "Model data type and gradient accumulation data type must be equal to use ZeRO"
-                )
+                if model_dtype == torch.bfloat16 and grad_accum_dtype == torch.float32:
+                    # when model_dtype is bfloat16, grad_accum_dtype of float32 should be valid, because
+                    # bfloat16 needs fp32 accumulation to maintain accuracy
+                    pass
+                else:
+                    raise NotImplementedError(
+                        "Model data type and gradient accumulation data type must be equal to use ZeRO"
+                    )
             if not is_zero_supported_optimizer(basic_optimizer):
                 assert (
                     self.zero_allow_untested_optimizer()
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1150,9 +1150,14 @@ class DeepSpeedEngine(Module):
             return ZERO_OPTIMIZATION
         elif amp_enabled:
             if model_dtype != grad_accum_dtype:
-                raise NotImplementedError(
-                    "Model data type and gradient accumulation data type must be equal to use Amp"
-                )
+                if model_dtype == torch.bfloat16 and grad_accum_dtype == torch.float32:
+                    # when model_dtype is bfloat16, grad_accum_dtype of float32 should be valid, because
+                    # bfloat16 needs fp32 accumulation to maintain accuracy
+                    pass
+                else:
+                    raise NotImplementedError(
+                        "Model data type and gradient accumulation data type must be equal to use Amp"
+                    )
             if model_dtype == torch.bfloat16 or model_dtype == torch.float16:
                 raise NotImplementedError(
                     "Cannot enable both amp with (legacy) fp16 or bfloat16 mode")


tjruwase:
@delock, can you please share more details on the issue you are running into? 

Currently, bf16_optimizer is activated with `zero_stage=0` even though it provides optimizer state partitioning. Sorry that this is quite confusing :(.

delock:
Hi @tjruwase .  When we train a 3.6b BF16 model with ZeRO stage 2, we get the following error message.  The error is new after we merge with upstream code recently, so we think it is due to new _do_optimizer_sanity_check behavior in the code.

We use a ported Megatron-LM from https://github.com/microsoft/DeepSpeedExamples/tree/master/megatron/Megatron-LM which allows us to run on XPU device with BF16 data type.  The model is still internally developing so we cannot share it, but functionality is still the same with public one.   If use this PR, this error will disappear and we can train the model as before merge.

If there are any more details needed let us know.

```
  File "/home/gma/frameworks.ai.benchmarking.other.deepspeed/deepspeed/runtime/engine.py", line 1137, in _do_optimizer_sanity_check
  File "/home/gma/frameworks.ai.benchmarking.other.deepspeed/deepspeed/runtime/engine.py", line 1137, in _do_optimizer_sanity_check
        optimizer_wrapper = self._do_optimizer_sanity_check(basic_optimizer)optimizer_wrapper = self._do_optimizer_sanity_check(basic_optimizer)

      File "/home/gma/frameworks.ai.benchmarking.other.deepspeed/deepspeed/runtime/engine.py", line 1137, in _do_optimizer_sanity_check
  File "/home/gma/frameworks.ai.benchmarking.other.deepspeed/deepspeed/runtime/engine.py", line 1137, in _do_optimizer_sanity_check
optimizer_wrapper = self._do_optimizer_sanity_check(basic_optimizer)
  File "/home/gma/frameworks.ai.benchmarking.other.deepspeed/deepspeed/runtime/engine.py", line 1137, in _do_optimizer_sanity_check
    optimizer_wrapper = self._do_optimizer_sanity_check(basic_optimizer)
  File "/home/gma/frameworks.ai.benchmarking.other.deepspeed/deepspeed/runtime/engine.py", line 1137, in _do_optimizer_sanity_check
    optimizer_wrapper = self._do_optimizer_sanity_check(basic_optimizer)
  File "/home/gma/frameworks.ai.benchmarking.other.deepspeed/deepspeed/runtime/engine.py", line 1137, in _do_optimizer_sanity_check
    optimizer_wrapper = self._do_optimizer_sanity_check(basic_optimizer)
  File "/home/gma/frameworks.ai.benchmarking.other.deepspeed/deepspeed/runtime/engine.py", line 1137, in _do_optimizer_sanity_check
    raise NotImplementedError(
NotImplementedError: Model data type and gradient accumulation data type must be equal to use ZeRO
    raise NotImplementedError(
NotImplementedError: Model data type and gradient accumulation data type must be equal to use ZeRO
    raise NotImplementedError(
NotImplementedError: Model data type and gradient accumulation data type must be equal to use ZeRO
    raise NotImplementedError(
NotImplementedError: Model data type and gradient accumulation data type must be equal to use ZeRO
        raise NotImplementedError(raise NotImplementedError(

NotImplementedErrorNotImplementedError: : Model data type and gradient accumulation data type must be equal to use ZeROModel data type and gradient accumulation data type must be equal to use ZeRO

    raise NotImplementedError(
NotImplementedError: Model data type and gradient accumulation data type must be equal to use ZeRO
        raise NotImplementedError(raise NotImplementedError(

    raise NotImplementedError(NotImplementedError
: NotImplementedErrorModel data type and gradient accumulation data type must be equal to use ZeRO:
    NotImplementedErrorModel data type and gradient accumulation data type must be equal to use ZeROraise NotImplementedError(:

Model data type and gradient accumulation data type must be equal to use ZeRO
raise NotImplementedError(NotImplementedError
: Model data type and gradient accumulation data type must be equal to use ZeRO
NotImplementedError: Model data type and gradient accumulation data type must be equal to use ZeRO

```

The model configuration json is as follows:

```
{
  "train_batch_size": 96,
  "steps_per_print": 1,
  "optimizer": {
    "type": "Adam",
    "params": {
      "lr": 0.00015,
      "weight_decay": 1e-2
    }
  },
  "zero_optimization": {
    "stage": 2
  },
  "communication_data_type": "bfp16",
  "gradient_clipping": 1.0,
  "fp16": {
    "enabled": false,
    "loss_scale": 0,
    "loss_scale_window": 1000,
    "hysteresis": 2,
    "min_loss_scale": 1
  },
  "bfloat16": {
    "enabled": true,
    "loss_scale": 1.0
  },
  "activation_checkpointing": {
    "partition_activations": true,
    "contiguous_memory_optimization": false
  },
  "flops_profiler": {
    "enabled": true,
    "profile_step": -1,
    "module_depth": -1,
    "top_modules": 1,
    "detailed": true,
    "output_file": null
  }
}
```

The model launch command is as follow, MP_SIZE is configured as 1.

```
gpt_options=" \
       --model-parallel-size ${MP_SIZE} \
       --num-layers 30 \
       --hidden-size 3072 \
       --num-attention-heads 32 \
       --batch-size 8 \
       --seq-length 2048 \
       --max-position-embeddings 2048 \
       --train-iters 10 \
       --resume-dataloader \
       --train-data webtext \
       --lazy-loader \
       --tokenizer-type GPT2BPETokenizer \
       --split 949,50,1 \
       --distributed-backend ccl \
       --lr 0.00015 \
       --no-load-optim \
       --lr-decay-style cosine \
       --weight-decay 1e-2 \
       --clip-grad 1.0 \
       --warmup .01 \
       --checkpoint-activations \
       --deepspeed-activation-checkpointing \
       --bf16 \
"
```

delock:
> @delock, can you please share more details on the issue you are running into?
> 
> Currently, bf16_optimizer is activated with `zero_stage=0` even though it provides optimizer state partitioning. Sorry that this is quite confusing :(.

Hi @tjruwase , I want to understand more about `zero_stage=0` and bf16_optimizer, I see from this link that zero_stage=0 means there is no zero optimization, is that right? https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/zero/config.py#L71

I read more about this sanity check, it seems `get_data_types()` is called from `_do_optimizer_sanity_check()` only.   And when model datatype is `bfloat16` and `self._config.grad_accum_dtype` is `None`, `get_data_types()` will return `(torch.bfloat16, torch.float32)` as `model_type` and `grad_accum_type`.   And this combination will fail sanity if zero optimization is enabled in config.

tjruwase:
@delock, apologies for breaking your code. Also, thanks to your analysis, I think I understand what what is wrong. But, first let me provide two clarifications
1. For bf16_optimizer.py, `training dtype` != `grad_accum_dtype`. In this case, specifically `bf16` != `fp32`. 
2. For all other optimizers, `zero_stage=[0,1,2,3]`, we expect `training dtype` == `grad_accum_dtype`.

These are hardcoded behaviors that we are trying to fix with the recent PR. In particular, we want to enable arbitrary combinations of `training dtype` and `grad_accum_dtype`. 

For your failing `zero_stage=2` case, I expect that it was running with `training dtype` == `grad_accum_dtype` == `bf16`, and not doing `grad_accum_dtype=fp32`? Please confirm. 

Regarding the breaking PR, the intention is that if `grad_accum_dtype` is not specified then it should default to `training dtype`. So, for your case, the expected behavior of `get_data_types()` is to return `(torch.bfloat16, torch.bfloat16)` since `self._config.grad_accum_dtype` is `None`. 

If you can confirm the above, I think the correct fix is to modify the following snippet:
https://github.com/microsoft/DeepSpeed/blob/d4cab2cec2c54b0bf9121dcd094f56754c581f6e/deepspeed/runtime/engine.py#L809-L811

Replace with the following: 
```python
        if self._config.grad_accum_dtype == None:
            if model_dtype == torch.bfloat16 and not self.zero_optimization():
                grad_accum_dtype = torch.float32  
```

Can you please try this fix on your side? Thanks!

danyang-rainbow:
> 

I tried this code, it works, thanks!

delock:
> ```python
> ```python
> not self.zero_optimization():
>                 grad_accum_dtype = torch.float32  
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> ```

Hi @tjruwase , yes the alternative fix you mention works on gpt 3.6b BF16 model with zero stage 2.  I have updated the PR with suggested fix.  Yes its true that for bf16 model with ZeRO optimizations, bf16_optimizer is not used.  So accumulation datatype should be the same as model datatype.  Thanks for explaination!

jomayeri:
Why modify the code? Why got just specify `grad_accum_dtype: "bf16"` in the config?

tjruwase:
> Why modify the code? Why not just specify `grad_accum_dtype: "bf16"` in the config?

Because we don't want to break backwards-compatibility.  This why the default value for `grad_accum_dtype` should be different from `bf16_optimizer` versus the zero optimizer implementations.

tjruwase (approved):

danyang-rainbow:
I have a question that when can bfloat16 model training with fp32 gradient accumulation and zero-1,2,3 be supported?

PR #2575, opened 2022-12-05
Support N-dimension input in quantization kernel

lokoppakmsft:


cmikeh2 (approved):

csrc/quantization/pt_binding.cpp
```diff
@@ -136,10 +136,12 @@ at::Tensor dequantize(at::Tensor& quantized_data,
                               .layout(at::kStrided)
                               .device(at::kCUDA)
                               .requires_grad(false);
-    const int final_dim_size = num_decompressed_elems(quantized_data, num_bits);
```
cmikeh2, line 139: If we aren't using this `num_decompressed_elems` anymore can we also remove the implementation for it?

cmikeh2 (changes requested):
I'm seeing unit test failures that are tied to the use of `quantized_data` rather than `output` for determining the number of elements for the dequantize implementation.

csrc/quantization/pt_binding.cpp
lokoppakmsft: Fixed it, all tests passed locally now ;

lokoppakmsft:
> I'm seeing unit test failures that are tied to the use of `quantized_data` rather than `output` for determining the number of elements for the dequantize implementation.

FIxed

cmikeh2 (approved):

Issue #2584, opened 2022-12-08
[BUG]AttributeError: module 'deepspeed' has no attribute 'ops'


SiTu1O:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**ds_report output**
Please run `ds_report` to give us details about your setup.

**Screenshots**

**System info (please complete the following information):**
 - OS: [e.g. Ubuntu 18.04]
 - GPU count and types [e.g. two machines with x8 A100s each]
 - Interconnects (if applicable) [e.g., two machines connected with 100 Gbps IB]
 - Python version
 - Any other relevant info about your setup

**Launcher context**
Are you launching your experiment with the `deepspeed` launcher, MPI, or something else?

**Docker context**
Are you using a specific docker image that you can share?

**Additional context**

yaozhewei:
Could you give more details? We cannot clearly follow the "reproduce steps"

jeffra:
I was able to reproduce what I think the issue is but it only seems to happen on python 3.6. It appears this regression was introduced since v0.7.5 release. I am bisecting further now and will push a fix soon.

<img width="895" alt="image" src="https://user-images.githubusercontent.com/645595/206585702-761cedb2-1082-422b-9fc0-ad6409b10b7e.png">

jeffra:
@SiTu1O, this should be resolved now. Please re-open if you still see the issue.

PR #2589, opened 2022-12-09
Fix issues w. python 3.6 + add py-version checks to CI

jeffra:
* Fixes https://github.com/microsoft/DeepSpeed/issues/2584
* Adds python version checks to our CI for basic installation + ds_report, this will catch issues like the one addressed in this PR in the future. It will not however catch runtime issues that are unique to python versions.
* Adds fix for newer `diffusers` versions that require unet.config access

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -9,15 +9,15 @@ on:
     branches:
       '**'
 
-concurrency:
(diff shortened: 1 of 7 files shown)


mrwyattii (approved):
LGTM

cmikeh2 (approved):

PR #2590, opened 2022-12-09
Move layer norm to new schedule

lokoppakmsft:
Update schedule for Layernorm kernel to  
1) improve performance
2) match upcoming quantization kernel schedule


lokoppakmsft:
@cmikeh2 Please review

csrc/transformer/inference/csrc/layer_norm.cu
```diff
@@ -200,7 +200,11 @@ Template arg:
     StoreResidual: controls whether the residual calculation is stored
         or not. When set to false, the input `res_output` is unused.
 */
-template <typename T, int UNROLL, bool StoreResidual>
+template <typename T, int UNROLL,
```
cmikeh2, line 203: Can we fold `internal_unroll` and `UNROLL` into a single parameter? They're always used as a pair in the implementation as far as I can tell.

csrc/transformer/inference/csrc/layer_norm.cu
```diff
@@ -200,7 +200,11 @@ Template arg:
...
+          int internal_unroll,
+          int threads_per_group,
+          int max_threads,
+          bool StoreResidual>
```
cmikeh2, line 207: Let's refactor `StoreResidual` to `preLN` style argument to make it more clear why the store is happening (I believe you already made this change internally for quant).

csrc/transformer/inference/csrc/layer_norm.cu
```diff
@@ -200,7 +200,11 @@ Template arg:
     StoreResidual: controls whether the residual calculation is stored
         or not. When set to false, the input `res_output` is unused.
 */
-template <typename T, int UNROLL, bool StoreResidual>
```
cmikeh2, line 203: For non `typename T` template parameters, let's use `camelCase` moving forward to consistently differentiate the variables from template params.

cmikeh2 (commented):
Minor style comments.

csrc/transformer/inference/csrc/layer_norm.cu
lokoppakmsft: done

csrc/transformer/inference/csrc/layer_norm.cu
lokoppakmsft: changed a few names; quant style may not be possible here since we have only one output when it is post residual layernorm. please let me know if this change looks good

cmikeh2: This looks good, I just meant more broadly that referring to pre-LN makes it a little more clear why we want to store the residual so what you did works great.

deepspeed/ops/transformer/transformer.py
```diff
@@ -170,7 +170,7 @@ def forward(ctx,
                 norm_w,
                 norm_b,
                 config):
-
+        print(config.__dict__)
```
cmikeh2, line 173: Need to remove extra prints

lokoppakmsft: removed thanks

cmikeh2 (approved):

PR #2591, opened 2022-12-09
Remove GatheredParameters context from replace_with_policy

lekurile:
This PR removes the zero-infernece `GatheredParameters` context from `replace_with_policy` due to no longer needing zero-inference after the introduction of meta tensor support for BLOOM.


lekurile:
@jeffra, @RezaYazdaniAminabadi, @mrwyattii

We make use of the `GatheredParameters` context in the `replace_wo_policy` function as well:
https://github.com/microsoft/DeepSpeed/blob/35eabb0a336e7a8e9950a550475ceaebda42066c/deepspeed/module_inject/replace_module.py#L759

I'm assuming we should still keep it there, but just wanted to make sure.

jeffra (approved):

PR #2599, opened 2022-12-13
Migrate ops tests to new inference_ops marker

cmikeh2:
This will run the op tests sequentially since the overhead of running them in parallel is greater than the parallelization benefits. In practice, I saw this reduce the latency of the `nv-inference` workflow from ~60 minutes to ~30 minutes.


jeffra (approved):
Thank you @cmikeh2, this is great

PR #2603, opened 2022-12-13
fixes #2498

### LINKED ISSUE #2498: [BUG] Incorrect batch_size used for ThroughputTimer
**Describe the bug**
`ThroughputTimer` is started/stopped for model steps, but in `runtime/engine` it uses the micro_batch_size only:
[here](https://github.com/microsoft/DeepSpeed/blob/ee39187d8f07f7efc615f64affa8cc60ccf41eb5/deepspeed/runtime/engine.py#L297)

Compare the above line to the following that uses the right value in `runtime/pipe/engine`: [here](https://github.com/microsoft/DeepSpeed/blob/b8fb9c3f1a8a8e3e574a7d53e83ce2b72d471aa3/deepspeed/runtime/pipe/engine.py#L107)

**To Reproduce**
1. Enable `wall_clock_breakdown`

**Expected behavior**
This timer should use  `self.train_micro_batch_size_per_gpu() * self.gradient_accumulation_steps()`.

**ds_report output**
N/A

**Screenshots**
N/A

 - OS: `Amazon Linux 2`
 - GPU count and types: `8 x A100`
 - Interconnects (if applicable): `4 x 100 Gbps`
 - Python version: `3.8`
 - Any other relevant info about your setup: 

**Launcher context**
Launching the experiment with the `deepspeed` launcher.

**Docker context**
Can't share

**Additional context**
DeepSpeed v0.7.3

clumsy:
taking gradient accumulation steps into account for throughput calculation

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -308,8 +308,7 @@ class DeepSpeedEngine(Module):
         self.timers = SynchronizedWallClockTimer()
         # Throughput timer
         self.tput_timer = ThroughputTimer(
-            batch_size=self.train_micro_batch_size_per_gpu(),
-            num_workers=self.dp_world_size,
+            batch_size=self.train_batch_size(),
             steps_per_output=self.steps_per_print(),
             monitor_memory=False,
         )
@@ -2168,7 +2167,8 @@ class DeepSpeedEngine(Module):
 
             report_progress = self.global_rank == 0 if self.global_rank else True
 
-        self.tput_timer.stop(report_progress)
+        self.tput_timer.stop(global_step=self.is_gradient_accumulation_boundary(),
+                             report_speed=report_progress)
 
         self._stop_timers(self.engine_timers.step_timers)
 
--- a/deepspeed/utils/timer.py
+++ b/deepspeed/utils/timer.py
@@ -136,7 +136,6 @@ class ThroughputTimer:
     def __init__(
         self,
         batch_size,
-        num_workers,
         start_step=2,
         steps_per_output=50,
         monitor_memory=False,
@@ -146,15 +145,13 @@ class ThroughputTimer:
         self.start_time = 0
         self.end_time = 0
         self.started = False
-        self.batch_size = batch_size
-        if batch_size is None:
-            self.batch_size = 1
-        self.num_workers = num_workers
+        self.batch_size = 1 if batch_size is None else batch_size
         self.start_step = start_step
         self.epoch_count = 0
-        self.local_step_count = 0
-        self.total_step_count = 0
+        self.micro_step_count = 0
+        self.global_step_count = 0
         self.total_elapsed_time = 0
+        self.step_elapsed_time = 0
         self.steps_per_output = steps_per_output
         self.monitor_memory = monitor_memory
         self.logging = logging_fn
@@ -167,7 +164,7 @@ class ThroughputTimer:
 
     def update_epoch_count(self):
         self.epoch_count += 1
-        self.local_step_count = 0
+        self.micro_step_count = 0
 
     def _init_timer(self):
         self.initialized = True
@@ -175,54 +172,60 @@ class ThroughputTimer:
     def start(self):
         self._init_timer()
         self.started = True
-        if self.total_step_count >= self.start_step:
+        if self.global_step_count >= self.start_step:
             torch.cuda.synchronize()
             self.start_time = time.time()
 
-    def stop(self, report_speed=True):
+    def stop(self, global_step=False, report_speed=True):
         if not self.started:
             return
         self.started = False
-        self.total_step_count += 1
-        self.local_step_count += 1
-        if self.total_step_count > self.start_step:
+        self.micro_step_count += 1
+        if global_step:
+            self.global_step_count += 1
+
+        if self.start_time > 0:
             torch.cuda.synchronize()
             self.end_time = time.time()
             duration = self.end_time - self.start_time
             self.total_elapsed_time += duration
+            self.step_elapsed_time += duration
 
-            curr_samples_sec = (self.batch_size * self.num_workers) / duration
-
-            if report_speed:
-                self.logging(
-                    "{}/{}, RunningAvgSamplesPerSec={}, CurrSamplesPerSec={}, MemAllocated={}GB, MaxMemAllocated={}GB"
-                    .format(
-                        self.epoch_count,
-                        self.local_step_count,
-                        self.avg_samples_per_sec(),
-                        curr_samples_sec,
-                        round(torch.cuda.memory_allocated() / 1024**3,
-                              2),
-                        round(torch.cuda.max_memory_allocated() / 1024**3,
-                              2),
-                    ))
-                if self.monitor_memory:
-                    virt_mem = psutil.virtual_memory()
-                    swap = psutil.swap_memory()
-                    self.logging("{}/{}, vm percent: {}, swap percent: {}".format(
-                        self.epoch_count,
-                        self.local_step_count,
-                        virt_mem.percent,
-                        swap.percent,
-                    ))
+            if global_step:
+                if report_speed and self.global_step_count % self.steps_per_output == 0:
+                    self.logging(
+                        "epoch={}/micro_step={}/global_step={}, RunningAvgSamplesPerSec={}, CurrSamplesPerSec={}, "
+                        "MemAllocated={}GB, MaxMemAllocated={}GB".format(
+                            self.epoch_count,
+                            self.micro_step_count,
+                            self.global_step_count,
+                            self.avg_samples_per_sec(),
+                            self.batch_size / self.step_elapsed_time,
+                            round(torch.cuda.memory_allocated() / 1024**3,
+                                  2),
+                            round(torch.cuda.max_memory_allocated() / 1024**3,
+                                  2),
+                        ))
+                    if self.monitor_memory:
+                        virt_mem = psutil.virtual_memory()
+                        swap = psutil.swap_memory()
+                        self.logging(
+                            "epoch={}/micro_step={}/global_step={}, vm %: {}, swap %: {}"
+                            .format(
+                                self.epoch_count,
+                                self.micro_step_count,
+                                self.global_step_count,
+                                virt_mem.percent,
+                                swap.percent,
+                            ))
+                self.step_elapsed_time = 0
 
     def avg_samples_per_sec(self):
-        if self.total_step_count > 0:
-            samples_per_step = self.batch_size * self.num_workers
-            total_step_offset = self.total_step_count - self.start_step
+        if self.global_step_count > 0:
+            total_step_offset = self.global_step_count - self.start_step
             avg_time_per_step = self.total_elapsed_time / total_step_offset
             # training samples per second
-            return samples_per_step / avg_time_per_step
+            return self.batch_size / avg_time_per_step
         return float("-inf")
 
 
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -104,9 +104,7 @@ class PipelineEngine(DeepSpeedEngine):
 
         self._force_grad_boundary = False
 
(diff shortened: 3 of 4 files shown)


deepspeed/runtime/engine.py
```diff
@@ -297,7 +297,8 @@ def __init__(
         self.timers = SynchronizedWallClockTimer()
         # Throughput timer
         self.tput_timer = ThroughputTimer(
-            batch_size=self.train_micro_batch_size_per_gpu(),
+            train_micro_batch_size_per_gpu=self.train_micro_batch_size_per_gpu(),
+            gradient_accumulation_steps=self.gradient_accumulation_steps(),
```
tjruwase, line 301: We don't need timer to be aware of `gradient accumulation steps` so that client for example could report progress on micro steps if they wish.

deepspeed/utils/timer.py
```diff
@@ -146,15 +147,18 @@ def __init__(
...
+        self.train_micro_batch_size_per_gpu = (1
+                                               if train_micro_batch_size_per_gpu is None
+                                               else train_micro_batch_size_per_gpu)
+        self.gradient_accumulation_steps = (1 if gradient_accumulation_steps is None else
+                                            gradient_accumulation_steps)
         self.num_workers = num_workers
         self.start_step = start_step
         self.epoch_count = 0
-        self.local_step_count = 0
+        self.local_microstep_count = 0
```
tjruwase, line 158: Can simply be `self.microstep_count`.

deepspeed/utils/timer.py
```diff
@@ -167,62 +171,71 @@ def __init__(
...
 
     def _init_timer(self):
         self.initialized = True
 
     def start(self):
         self._init_timer()
         self.started = True
-        if self.total_step_count >= self.start_step:
+        if self.local_microstep_count >= self.start_step * self.gradient_accumulation_steps:
```
tjruwase, line 182: Can be `if self.global_step_count >= self.start_step:`

deepspeed/utils/timer.py
```diff
@@ -167,62 +171,71 @@ def __init__(
...
             torch.cuda.synchronize()
             self.start_time = time.time()
 
     def stop(self, report_speed=True):
```
tjruwase, line 186: As suggested,
`def stop(self, global_step=False, report_speed=True)`

deepspeed/utils/timer.py
```diff
@@ -167,62 +171,71 @@ def __init__(
...
         if not self.started:
             return
         self.started = False
-        self.total_step_count += 1
-        self.local_step_count += 1
-        if self.total_step_count > self.start_step:
+        self.local_microstep_count += 1
+        gradient_accumulation_boundary = (self.local_microstep_count %
+                                          self.gradient_accumulation_steps == 0)
+        if gradient_accumulation_boundary:
```
tjruwase, line 193:
```python
   if global_step:
     self.global_step_count += 1
```

deepspeed/utils/timer.py
```diff
@@ -167,62 +171,71 @@ def __init__(
...
+            self.total_step_count += 1
+
+        if self.local_microstep_count > self.start_step * self.gradient_accumulation_steps:
```
tjruwase, line 196: `if self.global_step_count >= self.start_step`

deepspeed/utils/timer.py
```diff
@@ -167,62 +171,71 @@ def __init__(
...
-                    swap = psutil.swap_memory()
-                    self.logging("{}/{}, vm percent: {}, swap percent: {}".format(
-                        self.epoch_count,
-                        self.local_step_count,
-                        virt_mem.percent,
-                        swap.percent,
-                    ))
+            self.step_elapsed_time += duration
+
+            if gradient_accumulation_boundary:
```
tjruwase, line 203:
```python
   if global_step:
     self.global_step_count += 1
     if report_speed:
         self.logging.... 
```

deepspeed/utils/timer.py
clumsy: This, I believe, won't work as we need to accumulate the duration of the first micro steps before first full step after `self.start_step`

deepspeed/utils/timer.py
clumsy: I can change this, sure.

deepspeed/runtime/engine.py
```diff
@@ -2059,8 +2058,7 @@ def step(self, lr_kwargs=None):
             else:
                 self._take_model_step(lr_kwargs)
 
-            report_progress = self.global_rank == 0 if self.global_rank else True
-
+        report_progress = self.global_rank == 0 if self.global_rank else True
```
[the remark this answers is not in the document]
clumsy: Ack!

deepspeed/runtime/engine.py
clumsy: I this case, do we also want to use `batch_size=self.train_batch_size()` and don't even pass `num_workers`? It makes little sense to have it for the same reason, otherwise we need to accumulate samples I believe...

deepspeed/utils/timer.py
tjruwase: My idea is to have the client explicitly control when micro_steps and global_steps are updated so that the timer does not have to reason about or track the relationship of these two. Specifically, 
1. micro_steps is updated when client calls timer.stop()
2. global_steps is updated when client calls timer.stop(...,global_step=True,...) 

Does that make sense?

deepspeed/runtime/engine.py
tjruwase: I think we should keep `micro_batch_size` and `num_workers` since the timer.start()/stop() are invoked on micro batch size granularity. As I suggested in elsewhere, the client will inform timer of global batch size boundaries.

deepspeed/utils/timer.py
clumsy: I can change it to `if self.start_time > 0:` if there are no objections.

deepspeed/utils/timer.py
clumsy: Sounds good, will change.

deepspeed/utils/timer.py
tjruwase: `self.start_step` is defined in terms of global steps, so ideally should be compared with that.

deepspeed/utils/timer.py
clumsy: On the second though, don't we have a functional dependency between `global_step` and `report_speed`? The only case where the two have different values is at gradient accumulate step for non-0 rank, where we don't report value anyway. Or is it an instance of the debugging use-case you were referring previously? E.g. if we want to examine the contents of `tput_timer` for non-0 rank that is.

deepspeed/utils/timer.py
tjruwase: Good point. This logic is one of the oldest in DeepSpeed, and I feel our understanding of useful logging has changed a lot since then, so okay to break this. So how about we log 3 values
1. `self.micro_step_count`: number of micro steps of current epoch, will be reset every epoch
2. `self.global_step_count`: number of global steps of current epoch. Reset every epoch, I think?
3. `self.epoch_count`: number of epochs

What do you think?

deepspeed/utils/timer.py
tjruwase: Yes, reporting statistics on rank0 only is just a convention. But timer statistics is similarly collected by all ranks. So yes, this could be useful for debugging.

deepspeed/runtime/engine.py
clumsy: In this case, should I also change pipe/engine [here](https://github.com/microsoft/DeepSpeed/blob/ef869377e91660a649d16d8f416841ed9b5dbed3/deepspeed/runtime/pipe/engine.py#L107) to use `self.train_micro_batch_size_per_gpu()` without gradient accumulation steps as well?

tjruwase: But to your point, we could simplify as 
 ```python
      train_micro_batch_size=self.train_micro_batch_size_per_gpu() * self.dp_world_size
````

That will simplify timer code.

File: deepspeed/utils/timer.py (hunk 7 above)
Comment (clumsy) [reply deepspeed/utils/timer.py:203]: Sounds good to me, do you have any particular format in mind? `{epoch}/{micro_step}/{global_step}` or `epoch={}/micro_step={}/global_step={}`

File: deepspeed/runtime/engine.py (hunk 1 above)
Comment (clumsy) [reply deepspeed/runtime/engine.py:312]: Right, since you mentioned you want the caller to decide the reported batch size, rather than let the timer calculate it (in the spirit of `explicit is better than implicit`). I will make the change.

File: deepspeed/utils/timer.py [hunk 9]
```diff
@@ -167,62 +164,72 @@ def __init__(
...
     def avg_samples_per_sec(self):
-        if self.total_step_count > 0:
-            samples_per_step = self.batch_size * self.num_workers
-            total_step_offset = self.total_step_count - self.start_step
+        if self.global_step_count > 0:
+            total_step_offset = self.global_step_count - self.start_step
             avg_time_per_step = self.total_elapsed_time / total_step_offset
             # training samples per second
-            return samples_per_step / avg_time_per_step
+            return self.batch_size / avg_time_per_step
```
tjruwase, line 232: Should be `return self.micro_batch_size * self.global_step_count / avg_time_per_step`?

deepspeed/utils/timer.py
```diff
@@ -167,62 +164,72 @@ def __init__(
...
+                            .format(
+                                self.epoch_count,
+                                self.micro_step_count,
+                                self.global_step_count,
+                                virt_mem.percent,
+                                swap.percent,
+                            ))
+                self.step_elapsed_time = 0
+
+    def _samples_per_sec(self, duration):
```
tjruwase, line 223: This function now seems broken as it references deleted class field.

deepspeed/utils/timer.py
clumsy: `avg_time_per_step` is already average step duration across `self.global_step_count`, so the average throughput (per 1 step) is `self.batch_size / avg_time_per_step`. Or, equivalently, `self.batch_size * total_step_offset / total_elapsed_time`

deepspeed/runtime/engine.py
```diff
@@ -308,8 +308,7 @@ def __init__(
         self.timers = SynchronizedWallClockTimer()
         # Throughput timer
         self.tput_timer = ThroughputTimer(
-            batch_size=self.train_micro_batch_size_per_gpu(),
-            num_workers=self.dp_world_size,
+            batch_size=self.train_batch_size(),
```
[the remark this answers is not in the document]
clumsy: This is the correct value and the gist of https://github.com/microsoft/DeepSpeed/issues/2498 bug report.
We aggregate micro step durations `gradient_accumulation_steps` times. Meaning we processed `train_micro_batch_size_per_gpu() * gradient_accumulation_steps` in one global step per rank, so the overall (global step) throughput across all `self.dp_world_size` ranks is `(train_micro_batch_size_per_gpu() * gradient_accumulation_steps) * self.dp_world_size`.
This is also verified in the unit test in this PR.

deepspeed/utils/timer.py
clumsy: Great catch, it's dead code now, removing.

deepspeed/runtime/engine.py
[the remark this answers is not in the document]
tjruwase: Okay, now I understand what you are doing, `self.batch_size` is actually global batch size. That works, thanks for explaining. 

One more request, could you please rename `self.batch_size` to `self.global_batch_size` to improve clarity?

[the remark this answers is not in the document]
clumsy: We actually have a term for this value, it's `train_batch_size` (it's not just DeepSpeed-specific even), why not use it and avoid ambiguity?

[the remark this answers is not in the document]
tjruwase: Agree

tjruwase (approved):

PR #2623, opened 2022-12-17
[inference] ds-attention refactor w.r.t. ops

jeffra:
First of a few phase effort to unify the python front-end w.r.t. to DS kernels.


deepspeed/ops/transformer/inference/op_binding/base.py
```diff
@@ -0,0 +1,9 @@
+import torch
+import deepspeed
+
+
+class BaseOp(torch.nn.Module):
+    def __init__(self, builder="InferenceBuilder"):
```
cmikeh2, line 6: It looks like every child op is taking in the config right now, so I think it makes sense to maybe make this explicit with the base class taking in the config as well.

deepspeed/ops/transformer/inference/op_binding/vector_matmul.py
```diff
@@ -0,0 +1,22 @@
...
+        self.config = config
+        if self.config.fp16:
+            self.vector_matmul_func = self.inference_cuda_module.vector_matmul_fp16
+        else:
+            self.vector_matmul_func = self.inference_cuda_module.vector_matmul_fp32
+
+    def forward(self,
+                input: torch.Tensor,
+                weight: torch.Tensor,
+                q_scale: torch.Tensor,
```
cmikeh2, line 18: Since `q_scale` is always an attribute of the `weight`, I think we can collapse these two arguments into one.

deepspeed/ops/transformer/inference/ds_attention.py
```diff
@@ -435,13 +69,54 @@ def __init__(self,
...
-        if inference_cuda_module is None:
-            builder = op_builder.InferenceBuilder()
-            inference_cuda_module = builder.load()
+        self.qkv_func = QKVGemmOp(config)
+        self.score_context_func = SoftmaxContextOp(config)
+        self.linear_func = LinearOp(config)
+        self.vector_matmul_func = VectorMatMulOp(config)
+        self.softmax_func = SoftmaxOp(config)
+
+    def _transpose_for_scores(self, x, key=False, reshape=False):
```
cmikeh2, line 78: This function looks to be unused on all branches, can probably prune.

cmikeh2 (approved):

PR #2625, opened 2022-12-19
Optimize megatron unit test time

mrwyattii:
Change how we install Nvidia Apex (avoid compiling lots of things we don't actually need to for this test). As a result, reducing dependency install time from ~12m to ~20s:

Before:
image

Now:
image


mrwyattii:
Turns out we cannot use the python-only Apex install and so we are stuck with a super long Apex install for these tests

PR #2634, opened 2022-12-20
[zero] `_apply_to_tensors_only` handle `torch.Tensor` subclasses

stas00:
In a complex code base I run into this warning:

> [WARNING] [parameter_offload.py:55:_apply_to_tensors_only] A module has unknown inputs or outputs type (<class 'torch.nn.parameter.Parameter'>) and the tensors embedded in it cannot be detected. The ZeRO-3 hooks designed to trigger before or after backward pass of the module relies on knowing the input and output tensors and therefore may not get triggered properly.

But `torch.nn.parameter.Parameter` is a subclass of `torch.Tensor`.

This PR fixes this by replacing the exact check for `torch.Tensor` with a more flexible check that includes its subclasses as well.

Thank you!

@tjruwase

### DIFF
--- a/deepspeed/runtime/zero/parameter_offload.py
+++ b/deepspeed/runtime/zero/parameter_offload.py
@@ -46,7 +46,8 @@ def _apply_to_tensors_only(module, functional, backward_function, outputs):
                                                   outputs[key])
         return outputs
 
-    elif type(outputs) is torch.Tensor:
+    elif isinstance(outputs, torch.Tensor):
+        # this also applies to torch.Tensor's subclasses like torch.nn.parameter.Parameter
         return functional.apply(module, backward_function, outputs)
     else:
         if not is_builtin_type(outputs):


mrwyattii (approved):

tjruwase (approved):

stas00:
the megatron CI failure seems to be related to a wrong numpy version - unrelated to this PR

jeffra:
Funny enough, I added similar code in an old PR w. @awan-10: https://github.com/microsoft/DeepSpeed/pull/2127/commits/0ec57a41f90b0c62d43b25845c2b1a2016cc20ec. The CI logs won't load anymore but I recall it was removed it due to a test failing for a strange reason we didn't understand at the time (https://github.com/microsoft/DeepSpeed/pull/2127/commits/7ef1ca8583c3d5a36431985777347db754355b6f). If all the tests pass now then that's wonderful :)

stas00:
@jeffra, in:
```
elif isinstance(outputs, (torch.Tensor, torch.nn.parameter.Parameter)):
```
if i'm not mistaken `torch.nn.parameter.Parameter` will never be reached because it's a subclass of `torch.Tensor` so the first check will always succeed for an output type of either of the 2.

jeffra:
> @jeffra, in [0ec57a4](https://github.com/microsoft/DeepSpeed/commit/0ec57a41f90b0c62d43b25845c2b1a2016cc20ec):
> 
> ```
> elif isinstance(outputs, (torch.Tensor, torch.nn.parameter.Parameter)):
> ```
> 
> if i'm not mistaken `torch.nn.parameter.Parameter` will never be reached because it's a subclass of `torch.Tensor` so the first check will always succeed for an output type of either of the 2.

I agree, I believe my old commit and yours will do the same thing. But the current code in master will only match on Tensor and not Parameter. I recall something strange happened previously when including Parameters, but hopefully it’s resolved now since it’s been almost 6mo

stas00:
and if there is a problem, should it not assert and not warn and do nothing? it's so so difficult to notice these warnings - there are so many logs.

jeffra:
> and if there is a problem with the model, should it not assert and not warn and do nothing? it's so so difficult to notice these warnings - there are so many logs.
> 
> 
> 
> My personal opinion is that warnings should be completely eliminated in modern software - at least in the DL software with tens of thousands of logs. If there is a problem assert, if there is no problem move along.
> 
> 
> 
> Also this particular warning doesn't tell user much about what to do to solve it. The wording makes sense to the designer of the deepspeed code but is not very actionable for the user reading it. IMHO. As you can see I tried to fix the warning rather than try to understand why it was triggered as I have no idea what the problem is.

It appears the test failed. I agree with you I think warnings are way too easy to get ignored and in this case also agree it's not even clear what the user should do. I think seeing as the test fails there's a bug in z3 here but not sure exactly why this triggers it.

stas00:
it's `FAILED unit/runtime/zero/test_zero_context.py::test_ext_param_return`

```
 def test_ext_param_return():
        setup_serial_env()
    
        net = DanglingExt()
    
        args = SimpleNamespace(local_rank=0)
        engine, optim, _, _ = deepspeed.initialize(args=args,
                                                   model=net,
                                                   model_parameters=net.parameters(),
                                                   config=config)
    
        for _ in range(5):
            input = torch.rand(net.dim).to(engine.device).half()
>           loss = engine(input)

unit/runtime/zero/test_zero_context.py:315: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../unit-test-venv/lib/python3.8/site-packages/torch/nn/modules/module.py:1194: in _call_impl
    return forward_call(*input, **kwargs)
../unit-test-venv/lib/python3.8/site-packages/deepspeed/utils/nvtx.py:11: in wrapped_fn
    return func(*args, **kwargs)
../unit-test-venv/lib/python3.8/site-packages/deepspeed/runtime/engine.py:1833: in forward
    loss = self.module(*inputs, **kwargs)
../unit-test-venv/lib/python3.8/site-packages/torch/nn/modules/module.py:1212: in _call_impl
    result = forward_call(*input, **kwargs)
unit/runtime/zero/test_zero_context.py:293: in forward
    out = self.container(input)
../unit-test-venv/lib/python3.8/site-packages/torch/nn/modules/module.py:1212: in _call_impl
    result = forward_call(*input, **kwargs)
unit/runtime/zero/test_zero_context.py:281: in forward
    act2, bias = self.dangler(act1)
../unit-test-venv/lib/python3.8/site-packages/torch/nn/modules/module.py:1212: in _call_impl
    result = forward_call(*input, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = DanglingAttention(
  in_features=16, out_features=16, bias=True
  (d_linear): DanglingBias(in_features=16, out_features=16, bias=True)
)
input = tensor([ 0.2150, -0.1658,  0.0734, -0.3599, -0.5244, -0.2744,  0.1663, -0.2009,
        -0.1085,  0.1218, -0.0302, -0.... 0.2286, -0.1608,  0.0660],
       device='cuda:0', dtype=torch.float16,
       grad_fn=<PostBackwardFunctionBackward>)

    def forward(self, input):
        out = super().forward(input)
        if self.return_obj:
            out_obj = self.d_linear(out)
            assert out_obj.bias.ds_status == ZeroParamStatus.AVAILABLE
            # forward the external param
            return out_obj.out, out_obj.bias
        else:
            out, bias = self.d_linear(out)
>           assert bias.ds_status == ZeroParamStatus.AVAILABLE
E           AttributeError: 'Tensor' object has no attribute 'ds_status'
```

stas00:
Looking into `nn.Linear` https://pytorch.org/docs/stable/_modules/torch/nn/modules/linear.html#Linear
`bias` is  `nn.Parameter`
```
class Linear(Module):
[...]

    def __init__(self, in_features: int, out_features: int, bias: bool = True,
                 device=None, dtype=None) -> None:
        factory_kwargs = {'device': device, 'dtype': dtype}
        super(Linear, self).__init__()
        self.in_features = in_features
        self.weight = Parameter(torch.empty((out_features, in_features), **factory_kwargs))
        if bias:
            self.bias = Parameter(torch.empty(out_features, **factory_kwargs))
        else:
            self.register_parameter('bias', None)
```

samyam:
From a cursory look, it seems that the issue is not that the bias did not get converted into a ZeRO parameter, instead the issue seems to be that the bias returned by self.d_linear(out) is not the same object as self.d_linear.bias, and therefore a parameter of the module. 

Notice the output of the print statements:
```            
class DanglingAttention(torch.nn.Linear):
    def __init__(self, dim=16, return_obj=False):
      ....

    def forward(self, input):
        out = super().forward(input)
        if self.return_obj:
            out_obj = self.d_linear(out)
            assert out_obj.bias.ds_status == ZeroParamStatus.AVAILABLE
            # forward the external param
            return out_obj.out, out_obj.bias
        else:

          out, bias = self.d_linear(out)
          print(f"Parameter Ids: {[id(p) for p in self.d_linear.parameters()]}")
          print(f"Original vs returned bias id: {(id(self.d_linear.bias), id(bias))}")
          assert self.d_linear.bias.ds_status == ZeroParamStatus.AVAILABLE
          return out, bias

Output:
Parameter Ids: [139751043104336, 139751043104416]
Original vs returned bias id: (139751043104416, 139749617570176)
```
The assertion fails because `bias` is not the same as `self.d_linear.bias`. The latter is part of the module as you can see the id matches one of the parameter IDs of the module and has been converted in a ZeRO3 parameter, but the former has a different id.

Looks like python function might be making a copy of the tensor before returning it.

stas00:
Thank you for the investigation, @samyam. always awesome to hear your insights!

Once we remove the recursive list/dict handling `_apply_to_tensors_only` can be reduced to just (pseudo-code):
```
if outputs is torch.Tensor:
else:
 return outputs
```
and this PR changes it to do that for any subclass of `torch.Tensor`.

and so you're saying that `backward_function` mutates the tensor and strips all the deepspeed paritioned tensor metadata from it.

Should it then check if it's a non-partitioned DS tensor instead?

jeffra:
@stas00, @samyam and I deep dived this issue and think we figured out what is going on. Samyam created a new PR from your branch and pushed it here: https://github.com/microsoft/DeepSpeed/pull/2642

It appears that the bias parameter was being registered which is good but that the returned bias was modified (thus new object id) which meant it was lost. We modified the way zero handles this case to stash the original parameter so it can be updated in the external parameters dictionary correctly.

jeffra:
Closing as https://github.com/microsoft/DeepSpeed/pull/2642 is covering this now

PR #2640, opened 2022-12-21
Refactor/Pydantify monitoring config

mrwyattii:
Splitting work from https://github.com/microsoft/DeepSpeed/pull/2210 to make it easier to review and merge.

Monitor already used Pydantic, but updating it to use the `DeepSpeedConfigModel` like all other configs and added auto-generated config docs.


jeffra:
@Quentin-Anthony for awareness

Quentin-Anthony (commented):
> @Quentin-Anthony for awareness/review

I don't see any issues as of now. 

@mrwyattii -- I can review again if you make further changes. Feel free to let me know when this PR is finalized if you'd like me to take another look.

deepspeed/monitor/monitor.py
```diff
@@ -27,14 +27,14 @@ def __init__(self, monitor_config):
         self.tb_monitor = None
         self.wandb_monitor = None
         self.csv_monitor = None
-        self.enabled = monitor_config.tensorboard_enabled or monitor_config.csv_monitor_enabled or monitor_config.wandb_enabled
+        self.enabled = monitor_config.tensorboard.enabled or monitor_config.csv_monitor.enabled or monitor_config.wandb.enabled
```
tjruwase, line 30: Could this be simplified by adding an `enabled()` function to `DeepSpeedMonitorConfig` to the details?

deepspeed/monitor/wandb.py
```diff
@@ -10,10 +10,10 @@ def __init__(self, monitor_config):
         check_wandb_availability()
         import wandb
 
-        self.enabled = monitor_config.wandb_config.enabled
-        self.group = monitor_config.wandb_config.group
-        self.team = monitor_config.wandb_config.team
-        self.project = monitor_config.wandb_config.project
+        self.enabled = monitor_config.wandb.enabled
```
tjruwase, line 13: `WandbConfig` instead of `DeepSpeedMonitorConfig`?

tjruwase (approved):

mrwyattii:
@tjruwase thanks for the feedback! I incorporated the suggestions. I think I tried this initially but it was failing the unit tests. After closer investigation, it was easy to modify the unit tests so we can pass just the relevant `*_monitor` configs to `tensorboard`, `wandb`, and `csv_monitor`.

PR #2641, opened 2022-12-21
Refactor/Pydantify comms config 

mrwyattii:
Splitting work from https://github.com/microsoft/DeepSpeed/pull/2210 to make it easier to review and merge.

Comms already used Pydantic, but updating it to use the DeepSpeedConfigModel like all other configs

TODO:
- [ ] Refactor how the config is used
- [ ] Verify that this code is tested somehow in unit tests
- [ ] Add auto-gen docs

### DIFF
--- a/deepspeed/comm/comm.py
+++ b/deepspeed/comm/comm.py
@@ -75,7 +75,7 @@ from deepspeed.comm.utils import *
 
 
 def _configure_using_config_file(config):
-    if config.comms_logger_enabled:
+    if config.enabled:
         comms_logger.configure(config)
 
 
--- a/deepspeed/comm/config.py
+++ b/deepspeed/comm/config.py
@@ -3,29 +3,16 @@ Copyright (c) Microsoft Corporation
 Licensed under the MIT license.
 """
 
-from pydantic import BaseModel
-from .constants import *
+from deepspeed.runtime.config_utils import DeepSpeedConfigModel
 
 
-class CommsConfig(BaseModel):
-    class Config:
-        validate_all = True
-        validate_assignment = True
-        use_enum_values = True
-        extra = 'forbid'
+def get_comms_config(param_dict):
+    return DeepSpeedCommsConfig(**param_dict.get("comms_logger", {}))
 
 
-class CommsLoggerConfig(CommsConfig):
-    enabled: bool = COMMS_LOGGER_ENABLED_DEFAULT
-    prof_all: bool = COMMS_LOGGER_PROF_ALL_DEFAULT
-    prof_ops: list = COMMS_LOGGER_PROF_OPS_DEFAULT
-    verbose: bool = COMMS_LOGGER_VERBOSE_DEFAULT
-    debug: bool = COMMS_LOGGER_DEBUG_DEFAULT
-
-
-class DeepSpeedCommsConfig:
-    def __init__(self, ds_config):
-        self.comms_logger_enabled = 'comms_logger' in ds_config
-
-        if self.comms_logger_enabled:
-            self.comms_logger = CommsLoggerConfig(**ds_config['comms_logger'])
+class DeepSpeedCommsConfig(DeepSpeedConfigModel):
+    enabled: bool = False
+    prof_all: bool = True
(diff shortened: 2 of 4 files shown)


mrwyattii:
@Quentin-Anthony could you take a look at this PR? Also, do we have any unit tests that cover the comm logger? I couldn't find any and I'm wondering if you might be able to add some or help me add some?

Quentin-Anthony:
Yeah I can take a look. 

One immediate feedback is that https://github.com/microsoft/DeepSpeed/pull/2641/commits/3bff0be54b86c9a31ab7eb3ee26edd497ff22a22 will still fail because of the other checks in https://github.com/microsoft/DeepSpeed/blob/841ee593b943de4fd14db7d352d5bcb6b25f2ff9/deepspeed/comm/comm.py#L82 will fail here: https://github.com/microsoft/DeepSpeed/blob/841ee593b943de4fd14db7d352d5bcb6b25f2ff9/deepspeed/comm/comm.py#L97-L107

Lemme push a quick commit that replaces these assignments with `setattr()` calls in case the attr doesn't exist. That way we can do away with the `__init__()` func in `class CommsLogger` as you intended.

Quentin-Anthony:
There are no unit tests for the comms logger. I'll create some and commit them to this PR.

mrwyattii:
> Yeah I can take a look.
> 
> One immediate feedback is that [3bff0be](https://github.com/microsoft/DeepSpeed/commit/3bff0be54b86c9a31ab7eb3ee26edd497ff22a22) will still fail because of the other checks in
> 
> https://github.com/microsoft/DeepSpeed/blob/841ee593b943de4fd14db7d352d5bcb6b25f2ff9/deepspeed/comm/comm.py#L82
> 
> will fail here:
> https://github.com/microsoft/DeepSpeed/blob/841ee593b943de4fd14db7d352d5bcb6b25f2ff9/deepspeed/comm/comm.py#L97-L107
> 
> Lemme push a quick commit that replaces these assignments with `setattr()` calls in case the attr doesn't exist. That way we can do away with the `__init__()` func in `class CommsLogger` as you intended.

I'm not sure I follow. The code block you linked to should create those attributes if they don't already exist. I think defining `self.enabled` in the `CommsLogger.__init__()` is the only one necessary.

Quentin-Anthony:
> > Yeah I can take a look.
> > One immediate feedback is that [3bff0be](https://github.com/microsoft/DeepSpeed/commit/3bff0be54b86c9a31ab7eb3ee26edd497ff22a22) will still fail because of the other checks in
> > https://github.com/microsoft/DeepSpeed/blob/841ee593b943de4fd14db7d352d5bcb6b25f2ff9/deepspeed/comm/comm.py#L82
> > 
> > will fail here:
> > https://github.com/microsoft/DeepSpeed/blob/841ee593b943de4fd14db7d352d5bcb6b25f2ff9/deepspeed/comm/comm.py#L97-L107
> > 
> > Lemme push a quick commit that replaces these assignments with `setattr()` calls in case the attr doesn't exist. That way we can do away with the `__init__()` func in `class CommsLogger` as you intended.
> 
> I'm not sure I follow. The code block you linked to should create those attributes if they don't already exist. I think defining `self.enabled` in the `CommsLogger.__init__()` is the only one necessary.

You're right. I misread the error log and got very confused, looks like. Ignore my comment :)

mrwyattii:
> There are no unit tests for the comms logger. I'll create some and commit them to this PR.

@Quentin-Anthony Are you still able to do this? If not, I can put something together and get your input.

Quentin-Anthony:
> > There are no unit tests for the comms logger. I'll create some and commit them to this PR.
> 
> @Quentin-Anthony Are you still able to do this? If not, I can put something together and get your input.

My bad! This slipped my mind. Working on it now.

Quentin-Anthony:
I have tests in https://github.com/mrwyattii/DeepSpeed/pull/1. Let me know if you have any comments :)

mrwyattii:
> @mrwyattii -- Anything else needed here?

Thanks @Quentin-Anthony! These look great

PR #2642, opened 2022-12-21
[zero-3] Handle forward parameter return correctly in nested cases

samyam:
Extension of https://github.com/microsoft/DeepSpeed/pull/2634

cc/ @stas00, @tjruwase, @mrwyattii

### DIFF
--- a/deepspeed/runtime/zero/parameter_offload.py
+++ b/deepspeed/runtime/zero/parameter_offload.py
@@ -140,11 +140,13 @@ class PreBackwardFunction(torch.autograd.Function):
         if not hasattr(module, "applied_pre_backward_ref_cnt"):
             module.applied_pre_backward_ref_cnt = 0
         module.applied_pre_backward_ref_cnt += 1
+        #print(f"After Forward: {ctx.module.__class__.__name__}")
         outputs = outputs.detach()
         return outputs
 
     @staticmethod
     def backward(ctx, *args):
+        #print(f"Before Backward: {ctx.module.__class__.__name__}")
         ctx.pre_backward_function(ctx.module)
         return (None, None) + args
 
--- a/deepspeed/runtime/zero/parameter_offload.py
+++ b/deepspeed/runtime/zero/parameter_offload.py
@@ -46,7 +46,8 @@ def _apply_to_tensors_only(module, functional, backward_function, outputs):
                                                   outputs[key])
         return outputs
 
-    elif type(outputs) is torch.Tensor:
+    elif isinstance(outputs, torch.Tensor):
+        # this also applies to torch.Tensor's subclasses like torch.nn.parameter.Parameter
         return functional.apply(module, backward_function, outputs)
     else:
         if not is_builtin_type(outputs):
--- a/deepspeed/runtime/zero/parameter_offload.py
+++ b/deepspeed/runtime/zero/parameter_offload.py
@@ -48,14 +48,11 @@ def _apply_to_tensors_only(module, functional, backward_function, outputs):
 
     elif isinstance(outputs, torch.Tensor):
         # this also applies to torch.Tensor's subclasses like torch.nn.parameter.Parameter
-        param = None
-        if is_zero_param(outputs):
-            param = outputs
         touched_outputs = functional.apply(module, backward_function, outputs)
 
-        #using the temp to pass though the id of the original zero param
-        if param is not None:
-            touched_outputs.ds_param_alias = param
+        # restore zero param attributes if those get stripped by `backward_function`
+        if not is_zero_param(touched_outputs) and is_zero_param(outputs):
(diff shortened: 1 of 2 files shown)


tests/unit/runtime/zero/test_zero_context.py
```diff
@@ -296,6 +295,8 @@ def forward(self, input):
         assert len(self._external_params) == 0
         assert len(self.container._external_params) == 1
         assert len(self.container.dangler._external_params) == 0
+        assert id(self.container.dangler.d_linear.bias
```
stas00, line 298: may be adding a comment here from your explanation earlier - as it'd be difficult to remember why the test checks for `id` equivalency.

deepspeed/runtime/zero/parameter_offload.py
```diff
@@ -134,13 +143,11 @@ def forward(ctx, module, pre_backward_function, outputs):
         if not hasattr(module, "applied_pre_backward_ref_cnt"):
             module.applied_pre_backward_ref_cnt = 0
         module.applied_pre_backward_ref_cnt += 1
-        #print(f"After Forward: {ctx.module.__class__.__name__}")
```
stas00, line 137: let's keep these debug prints commented out - these help a lot when tracing bugs

deepspeed/runtime/zero/parameter_offload.py
```diff
@@ -134,13 +143,11 @@ def forward(ctx, module, pre_backward_function, outputs):
...
         outputs = outputs.detach()
         return outputs
 
     @staticmethod
     def backward(ctx, *args):
-        #print(f"Before Backward: {ctx.module.__class__.__name__}")
```
stas00, line 143: same as above

stas00 (commented):
I'm not familiar with the external param logic, so I will let others review.

but the rest looks good.

Thank you, @samyam

deepspeed/runtime/zero/parameter_offload.py
```diff
@@ -46,8 +46,17 @@ def _apply_to_tensors_only(module, functional, backward_function, outputs):
...
+        # this also applies to torch.Tensor's subclasses like torch.nn.parameter.Parameter
+        param = None
+        if is_zero_param(outputs):
+            param = outputs
+        touched_outputs = functional.apply(module, backward_function, outputs)
+
+        #using the temp to pass though the id of the original zero param
+        if param is not None:
+            touched_outputs.ds_param_alias = param
+        return touched_outputs
```
stas00, line 59: Do we need to worry that this will now introduce an overhead for all tensors while we are dealing with an edge case?

I'm not sure if this is any faster since the copy has been made already

```suggestion
        param = None
        if is_zero_param(outputs):
            param = outputs
        touched_outputs = functional.apply(module, backward_function, outputs)

        #using the temp to pass though the id of the original zero param
        if param is not None and not is_zero_param(touched_outputs):
            touched_outputs.ds_param_alias = param
        return touched_outputs
```

tests/unit/runtime/zero/test_zero_context.py
```diff
@@ -263,8 +263,8 @@ def forward(self, input):
             # forward the external param
             return out_obj.out, out_obj.bias
         else:
+
             out, bias = self.d_linear(out)
-            assert bias.ds_status == ZeroParamStatus.AVAILABLE
```
tjruwase, line 267: Is something like this not needed?

```python
   assert hasattr(bias, 'ds_status') or hasattr(bias, 'ds_param_alias')
   z3_bias = bias if hasattr(bias, 'ds_status') else bias.ds_param_alias
   assert z3_bias.ds_status == ZeroParamStatus.AVAILABLE
```

jeffra: Oh great catch @tjruwase, we were so focused on the first part of the problem that forget to add back the status check :)

jeffra:
Thank you @stas00 and @tjruwase for the good comments/suggestions. I accepted your 2nd suggestion @stas00 for the code change, it looked good and also passes. I noticed all the tests passed earlier, so we should be able to get this merged soon.

tjruwase (approved):

stas00:
I was thinking more about this - and I still think this situation is broken wrt the warning.

If a backward function is skipped, this a reason for an assert and not let's-swipe-it-under-the-carpet-may-be-it-will-be-noticed type of warning. This is big deal, no? Deepspeed is supposed to not interfere with the normal functioning of the model and here it does when it skips `backward`.

So we did deal with `nn.Parameter` type of output and all is good in this scenario since it no longer is skipped. But any other `outputs` will be a problem. The optimizer be missing some of the grads.

And, of course, what is the user to do when this happens? I think the most important part of using `assert` is that it will point out to the user which module this issue came from via traceback and then they can actually do something about it.

cc: @jeffra

Issue #2379, opened 2022-09-29
[Question] How to preshard a model for tensor parallism


lanking520:
Currently we are trying to run inference with pretrained BLOOM model. However, the loading takes very long due to DeepSpeed sharding in runtime. Since there is a pre-sharded version of BLOOM:

```
microsoft/bloom-deepspeed-inference-fp16
```

Is that possible to share the script that done the above job or any guidance on how to preshard to speed up loading experience?

mrwyattii:
Hi @lanking520 for loading this presharded version:
```python
import os
import torch
import deepspeed

from huggingface_hub import snapshot_download

model = "microsoft/bloom-deepspeed-inference-fp16"

local_rank = int(os.getenv("LOCAL_RANK", "0"))
world_size = int(os.getenv("WORLD_SIZE", "1"))

repo_root = snapshot_download(model)
checkpoints_json = os.path.join(repo_root, "ds_inference_config.json")
model = deepspeed.init_inference(
    model,
    mp_size=world_size,
    base_dir=repo_root,
    dtype=torch.half,
    checkpoint=checkpoints_json,
    replace_method="auto",
    replace_with_kernel_inject=True,
)
```

@RezaYazdaniAminabadi would be able to give more details about creating presharded versions for other models.

lanking520:
@mrwyattii thanks for the reply, I know how to load the model. But wondering more on how we can pre-shard the model. 

This also applies to other model like OPT models. Loading them in CPU and sharding take crazy long time. Maybe it could be done in one-off, save the sharded one to disk and next time skip loading them on CPU again. Would appreciate If there is any instruction that can share to save the sharded model with DeepSpeed

archieCanada:
@lanking520 

Here is how I did it (if I correctly understand the issue):

say 5-GB shards
```
python -c 'from transformers import AutoModelForSeq2SeqLM; \
model = AutoModelForSeq2SeqLM.from_pretrained("model_path/model_name"); \
model.save_pretrained("/path/to/model_name", max_shard_size="5GB")
```
this step will require 2x model-size cpu memory and then a bit more.

and then use the resulting model like so:
```
python -c 'from transformers import AutoModelForSeq2SeqLM; \
model = AutoModelForSeq2SeqLM.from_pretrained("/path/to/model_name");
```

lanking520:
> @lanking520 
> 
> 
> 
> Here is how I did it (if I correctly understand the issue):
> 
> 
> 
[...]

PR #2643, opened 2022-12-22
Add correctness check for sharded checkpoint test

mrwyattii:
Discussion on #2379 has indicated that there are correctness issues when loading certain models from sharded checkpoints.

@RezaYazdaniAminabadi

### DIFF
--- a/tests/unit/inference/test_checkpoint_sharding.py
+++ b/tests/unit/inference/test_checkpoint_sharding.py
@@ -3,7 +3,7 @@ import pytest
 import torch
 import deepspeed
 from deepspeed.model_implementations import DeepSpeedTransformerInference
-from transformers import AutoConfig, AutoModelForCausalLM
+from transformers import AutoConfig, AutoTokenizer, AutoModelForCausalLM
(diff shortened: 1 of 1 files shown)


tests/unit/inference/test_checkpoint_sharding.py
```diff
@@ -69,7 +69,9 @@ class TestCheckpointShard(DistributedTest):
     world_size = 2
 
     def test(self, model_name, dtype, class_tmpdir, save_shard):
+        local_rank = int(os.getenv("LOCAL_RANK", "0"))
         world_size = int(os.getenv("WORLD_SIZE", "1"))
+
```
tjruwase, line 74: Should this entire test be skipped if `dtype != torch.float16`, and thus eliminate the `if` condition in line 101?

tjruwase (approved):

tests/unit/inference/test_checkpoint_sharding.py
mrwyattii: No, we still want to ensure we can properly load the checkpoints in `int8` (we verify this with `check_dtype`)

PR #2547, opened 2022-11-24
Fix quantized-inference & Add generic support of checkpoint loading

RezaYazdaniAminabadi:
This PR generalizes the checkpoint loading/saving at DeepSpeed-Inference, to make the model-initialization faster for different tensor-parallelism degrees. This also allows to store the model in different tp-sharded configurations as requested by several users (https://github.com/microsoft/DeepSpeed/issues/2442 and https://github.com/microsoft/DeepSpeed/issues/2379). This has been already tested with several model architectures, GPT-J, GPT-NeoX, OPT, and BLOOM. 
Moreover, it enables users to quantize the model using the saved checkpoint. 

To use the inference on these models using the new model-checkpoint loading, you can use this [PR](https://github.com/microsoft/DeepSpeedExamples/pull/223), and more specifically this [test suite](https://github.com/microsoft/DeepSpeedExamples/blob/meta-inference/inference/huggingface/text-generation/inference-test.py). Here, I show some examples of running inference with GPTJ-6B and OPT-30B. 

### GPTJ-6B

Let's assume you download the checkpoint somewhere under '/tmp' directory and you want to use DeepSpeed with meta-tensor to load the checkpoint. To do so, you can use the following command:

```
deepspeed --num_gpus 1 inference-test.py --name EleutherAI/gpt-j-6B --ds_inference --use_kernel --use_meta_tensor --checkpoint_path '/tmp/
```

To save the new checkpoint, with any TP size, you can pass the directory in which you want to store it through the `save_mp_checkpoint_path` argument. Then, in order to run the model in INT8-quantized mode, you can pass `--dtype int8` to convert the model to INT8 format, and by passing a new path to save the model, you can get the quantized model. Here is the output of this model in INT8-quantized mode:
```
[2022-11-29 00:11:48,186] [INFO] [logging.py:68:log_dist] [Rank -1] DeepSpeed-Inference config: {'layer_id': 0, 'hidden_size': 4096, 'intermediate_size': 16384, 'heads': 16, 'num_hidden_laye
rs': -1, 'fp16': True, 'pre_layer_norm': True, 'local_rank': -1, 'stochastic_mode': False, 'epsilon': 1e-05, 'mp_size': 1, 'q_int8': True, 'scale_attention': True, 'triangular_masking': True
[...]

Issue #1687, opened 2022-01-10
[BUG] RuntimeError: Ninja is required to load C++ extensions


ShivamSharma2705:
Hi, 

I am getting the following error when running pretrain_gpt.sh 

--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
cpu_adam ............... [NO] ....... [OKAY]
cpu_adagrad ............ [NO] ....... [OKAY]
fused_adam ............. [NO] ....... [OKAY]
fused_lamb ............. [NO] ....... [OKAY]
sparse_attn ............ [NO] ....... [OKAY]
transformer ............ [NO] ....... [OKAY]
stochastic_transformer . [NO] ....... [OKAY]
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
 [WARNING]  async_io: please install the libaio-devel package with yum
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
async_io ............... [NO] ....... [NO]
transformer_inference .. [NO] ....... [OKAY]
utils .................. [NO] ....... [OKAY]
quantizer .............. [NO] ....... [OKAY]
--------------------------------------------------
DeepSpeed general environment info:
torch install path ............... ['/people/shar703/anaconda3/envs/deepspeed/lib/python3.8/site-packages/torch']
torch version .................... 1.8.2+cu111
torch cuda version ............... 11.1
nvcc version ..................... 11.1
deepspeed install path ........... ['/qfs/people/shar703/scripts/mega_ai/deepspeed_megatron/DeepSpeed/deepspeed']
deepspeed info ................... 0.5.9+1d295ff, 1d295ff, master
deepspeed wheel compiled w. ...... torch 1.8, cuda 11.1
**** Git info for Megatron: git_hash=1ac4a44 git_branch=main ****
using world size: 1, data-parallel-size: 1, tensor-model-parallel size: 1, pipeline-model-parallel size: 1 
using torch.float16 for parameters ...
------------------------ arguments ------------------------
[...]

PR #2645, opened 2022-12-23
Fix issue w. bloom int8 when changing tp size

jeffra:
When trying to load https://huggingface.co/microsoft/bloom-deepspeed-inference-int8 on 8 x A100-40GB this triggers a crash during checkpoint loading:

image

After bisecting it appears to be introduced in https://github.com/microsoft/DeepSpeed/pull/2547, which refactored the way `num_groups` is used during quantization. After peer-debugging w. @cmikeh2 we determined that `num_groups` in this case should be set to 32. We've tested both 176B and 7.1B versions of Bloom where we change from TP 4->8 and TP 2->4 respectively.

Also reported by @JingfengYang and confirmed fixes their issue in https://github.com/microsoft/DeepSpeed/pull/2217#issuecomment-1362681481

### DIFF
--- a/deepspeed/module_inject/replace_module.py
+++ b/deepspeed/module_inject/replace_module.py
@@ -142,6 +142,8 @@ class GroupQuantizer:
         self.group_size = group_size
         self.num_bits = num_bits
         self.q_int8 = q_int8
+        # TODO(jeff): need to check w. Reza on why this is needed when changing tp size w. bloom
+        self.num_groups = 32
 
     def quantize(self, inputs, qkv=True, count=1, parallel_dim=0):
         if not self.q_int8 or not qkv:


JingfengYang:
Thanks for your quick fix. It works as well on my side.

cmikeh2 (approved):

JingfengYang:
@jeffra When I'm trying to run BLOOM 175B (microsoft/bloom-deepspeed-inference-fp16) with 2 server (each with 8 x 40GB A100), I first ran into this problem https://github.com/microsoft/DeepSpeed/issues/1687 and solve it according to suggestions in that thread. But there is still another problem: 
`(deepspeed/module_inject/load_checkpoint.py", line 78, in load_parameters 
scale = scale.view( 
AttributeError: 'NoneType' object has no attribute 'view')`
image
Could you please help fix this? Thanks!

JingfengYang:
The command of running to this issue is: 
deepspeed --hostfile=hostfile bloom-inference-scripts/bloom-ds-inference.py --name microsoft/bloom-deepspeed-inference-fp16
by using https://github.com/huggingface/transformers-bloom-inference/tree/main/bloom-inference-scripts

I think the reason is that in deepspeed/module_inject/load_checkpoint.py, you set scale = None in line 48, but you directly call scale.view in line 77.
image

JingfengYang:
any updates on this issue? thanks!

JingfengYang:
Another kind reminder on this issue. Thanks!

JingfengYang:
@jeffra , any updates on this issue? thanks!

PR #2653, opened 2022-12-27
ZeRO3 handling frozen weights

### LINKED ISSUE #2615: [REQUEST] efficiently deal with frozen weights during training

This is a need for HF `accelerate` and `m4` projects.

We have more and more situations where a large part of the model that's being trained is frozen. 

At the moment the only way to emulate frozen weights in deepspeed is to create a param group with `lr=0`, which works but a huge amount of memory is wasted to allocate states that aren't being trained (grads too?) and the compute is wasted too.

Ideally, if `requires_grad == False` those weights shouldn't be in the optimizer and not allocating any optim state/grad memory.

Thank you!

@tjruwase 

@pacman100 - would you like to add/expand anything to this request?

tjruwase:
Frozen weights don't get optimizer state. 

Fix #2615


stas00:
You rock, Tunji!

I will test it and get back to you - slurm is currently very bad at yielding so this could take time

jomayeri (approved):

jeffra (approved):

PR #2657, opened 2022-12-29
fix flops profiler multiple wraps issue

NouamaneTazi:
Hello, first of all, thank you @cli99 for the very useful profiler.
I added support for `baddbmm` and most importantly. I added a mechanism to prevent infinitely wrapping functions, which lead to ever increasing throughput when we restarted the profiling.

cc @jeffra


NouamaneTazi:
@microsoft-github-policy-service agree [company="huggingface"]

NouamaneTazi:
@microsoft-github-policy-service agree company="huggingface"

cli99:
@NouamaneTazi , sorry for overlooking the PR. Can you check if the multiple wraps issue is resolved in the current master? Also we are using func.__str__ rather than func.__name__ now to avoid duplicating names

PR #2663, opened 2023-01-04
[fp16] lower `initial_scale_power` to `16`

stas00:
I'm proposing to change the default `initial_scale_power` to `16` from the current `32`. Here is why:

From [wikipedia](https://en.wikipedia.org/wiki/Half-precision_floating-point_format)

> The minimum strictly positive (subnormal) value is 2**−24 ≈ 5.96 × 10−8. The minimum positive normal value is 2**−14 ≈ 6.10 × 10−5. The maximum representable value is (2−2**−10) × 215 = 65504.

So I guess if the loss were to be `2**−24` then the maximum possible loss scale could be `2**48` `(24+16)` before it overflows, so `2**32` mathematically passes as a legit loss scale, except it's fantastically improbable.

But practically have you ever seen `loss<1` ? And `loss>1` then takes us to `initial_scale_power=16` as the practical starting point, which is likely to lead to just a few skipped optim states.

(while this might sounds as a pointless change - who cares about a few skipped steps which are likely to be totally insignificant when training for thousands of steps, these things affect situations like writing tests, or debugging a failing from the beginning training, etc.) 

--------------------

I hope this is not a backward compatibility breaking change, as someone not specifying `initial_scale_power` explicitly and relying on the default `32` will now have a slightly different outcome as they will start training sooner (less skipping). If it is, then we should leave the default `32` but change the doc to use `16` and add a note to why.

So please kindly discuss among your colleagues whether the proposed change is a good idea as is, or whether it'd be safer to not change the default, but the docs only. Thank you.

@tjruwase


tjruwase:
@stas00, thanks for this PR and for raising the backwards-compatibility issue. We will discuss internally.

jeffra (approved):

PR #2665, opened 2023-01-04
[GatheredParameters] fix memory leak

stas00:
Currently on exit from `GatheredParameters` with `modified_rank=None` memory is leaked as the gathered param remains gathered (the leak remains until the param is gathered again, most likely the first `forward`).

This PR fixes this problem by re-partitioning the param on `exit` from the `GatheredParameters` context.

A new test is supplied that reproduces this scenario and which fails prior to this PR.

Another test that was relying on the side-effect of this leak has been rewritten to do it properly.

 @tjruwase


tjruwase (approved):

stas00:
is it possible `nv-torch18-p40` is running an old code? I have just fixed:

```
tests/unit/runtime/zero/test_zero.py::TestZeroToFP32::test_2_param_groups[3] 
tests/unit/runtime/zero/test_zero.py::TestZeroToFP32::test_1_param_group[3]
```

so not sure why it's failing still

PR #2670, opened 2023-01-04
Automatic tensor parallelism v2

molly-smith:
Automatic tensor parallelism added for when no injection policy is provided. Injection policy is determined at runtime by parsing model from pipeline.

Unittest added for marian model. Testing capability added for translation and summarization tasks. Checks for fill-mask models that use angle bracket `<mask>` vs. square bracket `[MASK]`.


RezaYazdaniAminabadi (approved):

tests/unit/inference/test_inference.py
```diff
@@ -415,6 +447,57 @@ def test(
         assert assert_fn(bs_output, ds_output)
 
 
+@pytest.mark.seq_inference_
```
jeffra, line 450: I think there's a typo here so that these tests are run in the inference workflow:
`@pytest.mark.seq_inference_` -> `@pytest.mark.seq_inference`

tests/unit/inference/test_inference.py
```diff
@@ -139,8 +141,25 @@ def invalid_model_task_config(model_w_task, dtype, enable_cuda_graph):
...
+
+    def check_angle_bracket_mask(model):
+        for name in angle_bracket_mask_models:
+            if name in model:
+                return True
+        return False
+
     if task == "fill-mask":
-        if "roberta" in model:
+        if check_angle_bracket_mask(model):
```
mrwyattii, line 162:
```suggestion
        if any(map(lambda x: x in model, angle_bracket_mask_models)):
```
We could avoid having `check_angle_bracket_mask` with this one-liner

tests/unit/inference/test_inference.py
```diff
@@ -209,6 +239,8 @@ def assert_fn(model_w_task):
         "token-classification": token_classification_assert,
         "text-generation": text_generation_assert,
         "text2text-generation": text2text_generation_assert,
+        "translation": translation_assert,
+        "summarization": summarization_assert
```
mrwyattii, line 243: The summarization task is never used - should we add a model for this task?

molly-smith: I could add a second model to my unit test for summarization, but it wouldn't necessarily be testing tensor parallelism any more than before. Do you still think I should add it?

mrwyattii: no, that's probably not necessary. Thanks!

PR #2677, opened 2023-01-08
Abstract accelerator (step 3)

delock:
This PR is step 3 of fully merge https://github.com/microsoft/DeepSpeed/pull/2221.

This PR integrated deepspeed accelerator interface in deepspeed/.    After this PR, deepspeed could work with an accelerator extension following DeepSpeed accelerator abstraction interface.

Remaining files including UT, docs and benchmark will be integrated in an followup CI.


tjruwase:
@delock, seems there is a conflict.

deepspeed/runtime/zero/linear.py
```diff
@@ -29,8 +30,13 @@ def print_rank_0(message, debug=False, force=False):
 
 
 try:
-    autocast_custom_fwd = torch.cuda.amp.custom_fwd
-    autocast_custom_bwd = torch.cuda.amp.custom_bwd
+    device = get_accelerator().device_name()
```
tjruwase, line 33: I think it might be better to push the `try ... except` only into the `cuda` code path.

```python

if device == 'cuda':
   try:
     ...
   except:
     ...
else:
     autocast_custom_fwd = noop_decorator
     autocast_custom_bwd = noop_decorator
```

tjruwase (approved):

deepspeed/runtime/zero/linear.py
delock: Good suggestion, code updated.

deepspeed/runtime/swap_tensor/utils.py
```diff
@@ -179,9 +180,10 @@ def __init__(self, num_elems, count, dtype):
         self.count = count
         self.dtype = dtype
         self.all_buffers = [
-            torch.zeros(num_elems,
-                        device='cpu',
-                        dtype=dtype).pin_memory() for _ in range(count)
+            get_accelerator().pin_memory(
```
mrwyattii, line 183: Why does this need to be changed to use the `accelerator.pin_memory()` method?

deepspeed/runtime/zero/stage3.py
```diff
@@ -269,7 +269,7 @@ def __init__(self,
         self.__params_in_ipg_bucket: List[Parameter] = []
         self.is_gradient_accumulation_boundary: bool = True
 
-        self.__param_reduce_events: Deque[Event] = collections.deque()
+        self.__param_reduce_events: Deque[get_accelerator().Event] = collections.deque()
```
mrwyattii, line 272: This would break if the accelerator doesn't have `Event` defined. Perhaps we need an `event_type()` method for `*_Accelerator` classes?

deepspeed/runtime/swap_tensor/utils.py
delock: For a CPU tensor t, `t.pin_memory()` would pin this tensor to GPU memory.  In new pytorch version (1.13.1), a device string can be passed to pin_memory to specify current device.  This is how `get_accelerator().pin_memory(tensor)` is implemented for XPU:
https://github.com/intel/intel-extension-for-deepspeed/blob/main/intel_extension_for_deepspeed/xpu_accelerator.py#L178

However, in early version of pytorch (1.8.1), pin_memory does not accept device string as named argument, so in CUDA implementation we use the no parameter form to keep pytorch compatibility.
https://github.com/microsoft/DeepSpeed/blob/master/accelerator/cuda_accelerator.py#L215

So we introduced an interface `pin_memory()` in Accelerator for compatibility.  For CUDA device, we will call `pin_memory()` without device parameter so both pytorch 1.8.1 and pytorch 1.13.1 will work.  For XPU device, we will specificy xpu device explicitly to make pin_memory goes to XPU path.

If we run the following with torch 1.13.1 on CUDA device, pin_name with device argument will pass:
```
>>> import torch
>>> a = torch.zeros(3,3)
>>> b = a.pin_memory(device="cuda:0")
>>> a.is_pinned()
False
>>> b.is_pinned()
True
```
With torch 1.8.1 the third line will have an error, indicating pin_memory does not accept device as paramter.

deepspeed/runtime/zero/stage3.py
delock: This should be get_accelerator().Event(), and the method call should return the event type.   I'll fix this.

https://github.com/microsoft/DeepSpeed/blob/master/accelerator/cuda_accelerator.py#L106

delock: Event and Stream is now define and impelmented as @abstractmethod + @property, so reference their names in data types would return torch.cuda.Event and torch.cuda.Stream.  We can also use the same interface to create instances of Event and Stream by add parenthsis after it:
https://github.com/microsoft/DeepSpeed/pull/2677/files#diff-063566e7665f62d56f954817035fb24b21a1395225ca5d1ae792e60220c98e42R88

`get_accelerator().Event`: would return torch.cuda.Event as class type
`get_accelerator().Event()`: would create new object from torch.cuda.Event

deepspeed/runtime/swap_tensor/utils.py
mrwyattii: thanks for the explanation! I learned something new today :)

deepspeed/runtime/zero/stage3.py
mrwyattii: Looks good, thank you!

delock:
The latest commit is to fix a test issue with develop install (pip install -e .).

I wonder whether deepspeed can add a workflow for _develop_ install?  The difference of op_builder in build time and runtime make it sensitive to develop install (under develop install, both op_builder and deepspeed.ops.op_builder are visible).  @tjruwase

PR #2687, opened 2023-01-11
Fix import error of op_builder

tohtana:
This PR fixes an import error of in `builder.py`. (See also #2560)

The changes in 9548d48 causes the following error when you install deepspeed with `-e` option.

```
File "/DEEPSPEED_REPO_ROOT/op_builder/builder.py", line 453, in load
from ...git_version_info import installed_ops, torch_info 
ImportError: attempted relative import beyond top-level package 
```

When we have deepspeed installed with -e, we have builder.py in both `deepspeed/ops/op_builder/` and `op_builder/` (the former is a symbolic link). Python interpreter recognizes the path of builder.py as `op_builder/` at runtime and reference to `...git_version_info` fails. 

This PR fixes the issue by importing the module with the absolute path (proposed by @jeffra at #2560).

### DIFF
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -450,7 +450,7 @@ class OpBuilder(ABC):
             extra_link_args=self.strip_empty_entries(self.extra_ldflags()))
 
     def load(self, verbose=True):
-        from ...git_version_info import installed_ops, torch_info
+        from deepspeed.git_version_info import installed_ops, torch_info
         if installed_ops[self.name]:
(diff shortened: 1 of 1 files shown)


conglongli (approved):

jeffra:
Let's wait to double check w. @delock that this works in their environment.

jeffra (approved):
Looks like @delock agrees this fixes the problem

delock:
Yes, just confirmed this patch works well in our local environment.  Thanks!

PR #2688, opened 2023-01-11
Fix runtime bug in developer install mode

delock:
This PR fix a bug that prevents deepspeed to work properly under developer install (`python setup.py develop`).   The root cause is in CUDA_Accelerator, when `op_builder_dir()` is called, CUDA_Accelerator suppose to return **"op_builder"** during build time, and **"deepspeed.ops.op_builder"** during runtime.   In previous implementation, CUDA_Accelerator would try to import `op_builder`.  If this import is successful, CUDA_Accelerator would assume this is build time and return "op_builder".  Otherwise it would assume this is runtime and return "deepspeed.ops.op_builder".

However, if deepspeed is installed through developer install, `op_builder` module would still be accessible, so CUDA_Accelerator would get confused and return "op_builder" in `op_builder_dir()`.   This would eventurally causing op builders be created from op_builder path and throw an error from [`builder.py#L453`](https://github.com/microsoft/DeepSpeed/blob/master/op_builder/builder.py#L453), which is supposed to be called from `deepspeed.ops.op_builder` only.

To resolve this issue, we introduced an optional argument in `accelerator.get_accelerator`, and call `get_accelerator(for_setup=True)` in the very beginning of setup.py.  get_accelerator would subsequently pass this argument to CUDA_Accelerator initializer.   When `op_builder_dir` is called, CUDA_Accelerator would use the value in this argument to decide whether to return "op_builder" or "deepspeed.ops.op_builder", so CUDA_Accelerator won't be confused whether its in  build time and runtime, and work correctly.

Note that XPU_Accelerator does not need to have this argument in initializer.  Because XPU_Accelerator is defined in sperate package, so its op builder path is the same during build time and runtime, so no distinguish is needed.

### DIFF
--- a/accelerator/cuda_accelerator.py
+++ b/accelerator/cuda_accelerator.py
@@ -12,9 +12,10 @@ except ImportError:
 
 
 class CUDA_Accelerator(DeepSpeedAccelerator):
-    def __init__(self):
+    def __init__(self, from_setup=False):
         self._name = 'cuda'
         self._communication_backend_name = 'nccl'
+        self._from_setup = from_setup
 
         # begin initialize for create_op_builder()
         # put all valid class name <--> class type mapping into class_dict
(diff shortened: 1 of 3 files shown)


delock:
Closing this since #2687 would also fix this bug.

PR #2703, opened 2023-01-13
[ZeRO] non-MoE stage 1 requires CG disabled

jeffra:
MoE w. stage 1 requires contiguous gradients (CG) is enabled, which was fixed in https://github.com/microsoft/DeepSpeed/pull/2250. However, this introduces a performance regression when not using MoE. This PR reverts the non-MoE case to ensure CG is disabled.

/cc @siddharth9820, @tjruwase


siddharth9820:
looks good to me

tjruwase (approved):

PR #2705, opened 2023-01-14
Preserve prior op builder backend api

jeffra:
#2560 and #2504 changed the way deepspeed creates op builders, this PR aims to preserve the previous API for deepspeed op developers.

I have not changed all the places to use the new API so many tests will still fail. I want to get early feedback from @delock and @tjruwase first before proceeding with reverting the other changes to ensure everything still works as intended on the Intel side.

Current abstract accelerator API:

```python
from deepspeed.ops.op_builder.builder_names import CPUAdamBuilder
self.ds_opt_adam = get_accelerator().create_op_builder(CPUAdamBuilder).load()
```

Prior API and proposed in this PR:

```python
from deepspeed.ops.op_builder import CPUAdamBuilder
self.ds_opt_adam = CPUAdamBuilder().load()
```


delock:
Hi @jeffra , looks like this is a better way of using builder, without need to call create_op_builder() at every place op builder is needed.

In op_builder/__init__.py, all op builders implemented in deepspeed will be enumerated and their name will be used to build op builders from accelerator extension, I think its a good way to keep the old interface.
https://github.com/microsoft/DeepSpeed/pull/2705/files#diff-3e903a37c259ed263b3073df3b02281c9f4dd22b4f98dd47c4d76a164d376c20R42

In our extension, currently we didn't implement all the builders.  Those builders essential to our customer's workload will be implemented first.   So when DeepSpeed try to get op builder from names that is not implemented in our accelerator extension, **None** will be returned.   Currently DeepSpeed is build around this awareness.  I'll try this PR in our local environment and see if that still works.

delock:
Hi @jeffra , we tested this change in our local environment.  We applied your change and modified all the places that use op builder.  This change works fine on our devices.

When will this change be made, before or after #2677 ?  Another option maybe apply this change on #2677 and merge together.

tjruwase (approved):

jeffra:
> Hi @jeffra , we tested this change in our local environment. We applied your change and modified all the places that use op builder. This change works fine on our devices.
> 
> When will this change be made, before or after #2677 ? Another option maybe apply this change on #2677 and merge together.

Thanks so much for trying this out. If possible could we add it to #2677?

delock:
> > Hi @jeffra , we tested this change in our local environment. We applied your change and modified all the places that use op builder. This change works fine on our devices.
> > When will this change be made, before or after #2677 ? Another option maybe apply this change on #2677 and merge together.
> 
> Thanks so much for trying this out. If possible could we add it to #2677?

Sure, I'll add this change to #2677, thanks!

delock:
@jeffra The builder closure needs to return the Builder class during runtime, like implemented as follow.
https://github.com/microsoft/DeepSpeed/pull/2677/commits/0e63cf227fdfdfcdd83f64a0cc247fcf6f0eecbc

Without this change, the following line will fail, since "AsyncIOBuilder" is actually a closure function than a real Builder class.
https://github.com/huggingface/transformers/blob/main/tests/deepspeed/test_deepspeed.py#L103

With the change above, module outside deepspeed which use deepspeed OpBuilder should experience no change of behavior as before.

Issue #2636, opened 2022-12-21
RuntimeError: Special sequence length found in softmax backward [BUG]


FarzanT:
**Describe the bug**
DeepSpeedTransformerLayer seems to be not compatible with sequence lengths larger than 2048, because of `launch_attn_softmax_backward_v2` here:
https://github.com/microsoft/DeepSpeed/blob/7bcb4fabeb6c2a84998ab41af62bd0cc9eb25f2e/csrc/transformer/softmax_kernels.cu#L551-L582

I have created a transformer-based model that has a fixed input sequence length of 12896. The model works ok with a custom transformer module written in PyTorch, and I have also integrated it with DeepSpeed Distributed Training in PyTorch Lightning, with FP16 precision.
I copied the exact model, but replaced the transformer layers with the transformer kernel described in the **very brief** example provided in the tutorials: https://www.deepspeed.ai/tutorials/transformer_kernel/
However, the model throws different errors based on whether I use a mask or not.
Please see below for the full stack trace.

**To Reproduce**

Here's a relatively small example to reproduce the problem. (please let me know if I'm using the transformer kernel incorrectly!):
```python
import copy
import torch
from deepspeed import DeepSpeedTransformerConfig, DeepSpeedTransformerLayer
from torch import nn


class DeepSpeedTransformerEncoder(nn.Module):
    def __init__(self, **ds_config):
        super().__init__()
        config = DeepSpeedTransformerConfig(batch_size=ds_config['batch_size'],
                                            hidden_size=ds_config['hidden_size'],
                                            heads=ds_config['heads'],
                                            attn_dropout_ratio=0.0,
                                            hidden_dropout_ratio=ds_config['hidden_dropout_ratio'],
                                            num_hidden_layers=ds_config['num_hidden_layers'],
                                            initializer_range=0.02,
                                            seed=ds_config['seed'],
                                            fp16=ds_config['fp16'],
                                            pre_layer_norm=ds_config['pre_layer_norm'],
                                            attn_dropout_checkpoint=ds_config['attn_dropout_checkpoint'],
                                            normalize_invertible=ds_config['normalize_invertible'],
                                            gelu_checkpoint=ds_config['gelu_checkpoint'])

        self.layers = nn.ModuleList([
            copy.deepcopy(DeepSpeedTransformerLayer(config))
            for _ in range(config.num_hidden_layers)
        ])

    def forward(self, x, **kwargs):
        for layer in self.layers:
[...]

PR #2709, opened 2023-01-16
Fix softmax backward

RezaYazdaniAminabadi:
This PR addresses https://github.com/microsoft/DeepSpeed/issues/2636


csrc/transformer/softmax_kernels.cu
```diff
@@ -575,10 +671,15 @@ void launch_attn_softmax_backward_v2(T* out_grad,
...
+            <<<grid_dim, block_dim, 0, stream>>>(out_grad, soft_inp, seq_length);
+    else if (seq_length <= 8192)
+        softmax_backward_kernel_v2<T, 256>
+            <<<grid_dim, block_dim, 0, stream>>>(out_grad, soft_inp, seq_length);
     else
-        throw std::runtime_error(
-            std::string("Special sequence length found in softmax backward, seq_length: ") +
-            std::to_string(seq_length));
+        softmax_backward_kernel_arbitrary_length<<<grid_dim, block_dim, 0, stream>>>(
+            out_grad, soft_inp, seq_length / (4 << ((sizeof(T) & 2) >> 1)));
```
cmikeh2, line 682: For `seq_length > 8192` and `seq_length % 8 != 0` (for `__half`) we'll still have slight correctness issues if I'm reading this correctly. For training this shouldn't be an issue typically, but should we throw a runtime error here just to be safe?

RezaYazdaniAminabadi: That's a good point, will do that.

RezaYazdaniAminabadi: also, the sequence will be padded here:
https://github.com/microsoft/DeepSpeed/blob/fix-softmax-backward/deepspeed/ops/transformer/transformer.py#L178
So, as you said there is no concern on the training side.

cmikeh2: If we're padding I don't see a need to add a runtime error.

cmikeh2 (approved):

PR #2727, opened 2023-01-20
Enable tensor fragments for zero 2 & 3

### LINKED ISSUE #2290: [REQUEST] individual tensor `.grad` getter/setter for flattened tensors
As we know DS flattens individual tensors by optim groups so that each tensor and its attributes can't be accessed by the user once the ds engine takes over.

As @tjruwase is working on the API used by the universal checkpoint to get/set fragments of data/grads/optim states across multiple gpus, I think the next stage is to override some accessors in those flattened tensors to transparently get/set data on those.

For example `weight.grad` could behind the scenes get/set the sharded fragments - allowing users normal access to important data. e.g. researchers often need to access grads. optim states is another really important data that needs to be accessible. e.g. we wanted this to diagnose spikes during the training, but alas that data wasn't accessible.

And of course should look at what other accessors could be overriden and ideally the flattened tensors would appear to the user as normal tensors.

Thank you!

@tjruwase

tjruwase:
Enable tensor fragmentation in zero stage 2 & 3
Fix #2290
TODO: 
test w/ offload (at least cpu to start with)
test w/o offload


deepspeed/utils/tensor_fragment.py
```diff
@@ -58,9 +61,56 @@ def get_full_hp_param(self, optim_state_key=None):
...
+                                       lp_frag_address.start,
+                                       lp_frag_address.numel)
+
+        hp_frag_address = self._hp_mapping.hp_fragment_address
+        hp_mapping = self._hp_mapping
+        # lp_grad_fragment = torch.narrow(hp_mapping.gradient_dict[hp_mapping.param_group_index],
+        #                                 0,
+        #                                hp_frag_address.start,
+        #                                hp_frag_address.numel)
+
```
stas00, line 91:
```suggestion
        from deepspeed.utils import logger
        if hp_mapping.param_group_index not in hp_mapping.gradient_dict or hp_mapping.gradient_dict[hp_mapping.param_group_index] is None:
            logger.warning("Gradients are only available immediately after backward and before egine step")
            return None
```

stas00: not sure if we want to warn or assert here though. You know my stance on warnings...

If the user calling this function they must want the gradients, so it's a usage error if they call it at the wrong timing and thus perhaps it should be an assert?

tjruwase: I agree that assert is appropriate here.

deepspeed/utils/tensor_fragment.py
```diff
@@ -58,9 +61,59 @@ def get_full_hp_param(self, optim_state_key=None):
...
+        hp_frag_address = self._hp_mapping.hp_fragment_address
+        hp_mapping = self._hp_mapping
+        # lp_grad_fragment = torch.narrow(hp_mapping.gradient_dict[hp_mapping.param_group_index],
+        #                                 0,
+        #                                hp_frag_address.start,
+        #                                hp_frag_address.numel)
+        from deepspeed.utils import logger
+        if hp_mapping.param_group_index not in hp_mapping.gradient_dict or hp_mapping.gradient_dict[hp_mapping.param_group_index] is None:
+            logger.warning("Gradients are only available immediately after backward and before egine step")
+            return None
```
stas00, line 94:
```suggestion
        if hp_mapping.param_group_index not in hp_mapping.gradient_dict or hp_mapping.gradient_dict[hp_mapping.param_group_index] is None:
           raise ValueError("Gradients are only available immediately after backward and before egine step")
```

stas00: changed to assert

stas00:
Tunji, please kindly add to the OP a TODO list:
- test w/ offload and w/o it (at least cpu to start with) (as this PR currently breaks under cpu offload)
Thanks.

stas00:
current breakages.

1. not working under cpu_offload since elf.averaged_gradients doesn't get populated in stage_1_and_2.py
```
        if self.cpu_offload is False:
            for i, _ in enumerate(self.bit16_groups):

                if not i in self.averaged_gradients or self.averaged_gradients[i] is None:
                    self.averaged_gradients[i] = self.get_flat_partition(
```

2. getting shapes wrong:

```
  File "m4/training/main.py", line 155, in <module>
    train_logs = trainer.train()
  File "/gpfswork/rech/cnw/uue59kq/repos/m4_3/m4/training/trainer.py", line 997, in train
    ) = self._do_batch(
  File "/gpfswork/rech/cnw/uue59kq/repos/m4_3/m4/training/trainer.py", line 425, in _do_batch
    self._log_deepspeed_training_stats()
  File "/gpfswork/rech/cnw/uue59kq/repos/m4_3/m4/training/trainer.py", line 474, in _log_deepspeed_training_stats
    hp_grad = safe_get_full_grad(lp)
  File "/gpfswork/rech/cnw/commun/conda/lucile-m4_3/lib/python3.8/site-packages/deepspeed/utils/tensor_fragment.py", line 106, in safe_get_full_grad
    return param.get_full_hp_grad()
  File "/gpfswork/rech/cnw/commun/conda/lucile-m4_3/lib/python3.8/site-packages/deepspeed/utils/tensor_fragment.py", line 96, in get_full_hp_grad
    reduce_fragment.data.copy_(hp_grad_fragment.data)
RuntimeError: The size of tensor a (1534) must match the size of tensor b (1028) at non-singleton dimension 0 
```

stas00:
Also a bonus feature would be to be able to return unscaled and clipped grads (before step)

deepspeed/runtime/zero/stage3.py
```diff
@@ -2109,6 +2148,26 @@ def get_fp32_grad_partitions(self) -> Dict[int, Dict[int, Tensor]]:
...
+    def get_fp32_grad_for_param(self, param) -> Tensor:
+        self.__reduce_and_partition_stream.synchronize()
+        reduce_buffer = torch.zeros(param.ds_shape, dtype=torch.float32, device=param.device).flatten()
+
+        if self.offload_optimizer:
+            group_idx, dest_offset, num_elements = self.grad_position[self.get_param_id(param)]
+            fp32_grad = self.fp32_partitioned_groups_flat[group_idx].grad.narrow(
+                0,
+                dest_offset,
+                num_elements)
```
stas00, line 2160:
```suggestion
            fp32_grad = self.fp32_partitioned_groups_flat[group_idx].grad.narrow(
                0,
                dest_offset,
                num_elements) .to(device=param.device)
```

This is needed for the offload to work

stas00:
OK, so z3/grads works fine with 1 gpu w/ and w/o offload - great.

but with 2 gpus, sometimes it works, sometimes it doesn't. for example it fails for a param `nn.Parameter(torch.zeros(1))`

If I add debug:
```
        print(f"{fp32_grad.shape} => {reduce_buffer.shape}")
```

it gives: `torch.Size([1]) => torch.Size([1])`

and then fails with:

```
    work = group._allgather_base(output_tensor, input_tensor)
RuntimeError: output tensor size must be equal to world_size times input tensor size
```

so I think there are 2 problems here, 
1. it seems it expects the chunks we gather to be of equal size, 
2. but also for some reason we have 2 chunks of size 1 when the full tensor is also of size 1. `1+1 =! 1`

stas00:
ok, so the following version will allow any lengths of shards:

```
    def get_fp32_grad_for_param(self, param) -> Tensor:
        self.__reduce_and_partition_stream.synchronize()

        if self.offload_optimizer:
            group_idx, dest_offset, num_elements = self.grad_position[self.get_param_id(param)]
            fp32_grad = self.fp32_partitioned_groups_flat[group_idx].grad.narrow(
                0,
                dest_offset,
                num_elements).to(device=param.device)
        else:
            fp32_grad = self.__param_id_to_grad_partition[param.ds_id].float()

        output = [None] * self.partition_count
        torch.distributed.all_gather_object(output,
                        fp32_grad,
                        group=self.dp_process_group)

        print(f"need total {param.ds_shape} got shards {[x.shape for x in output]}")

        return torch.cat([x.to(device=param.device) for x in output]).view(param.ds_shape)
```

but it still has an issue in the case of param of len `1` - I think we aren't dealing with padding here and the 2nd shard must be just padding.

indeed, if I dump the `output` in this case, it's indeed padding (0) on the 2nd gpu:

```
result [tensor([62.5625], device='cuda:0'), tensor([0.], device='cuda:1')]
```

stas00:
When we finish z3 here is an important next feature to support for `grads` getter: support flags `clipped=True/False` and `scaled=True/False` - currently both get done in `step`, but it'd be helpful for the developer to see these different versions as well.

deepspeed/runtime/zero/stage3.py
```diff
@@ -2115,6 +2153,26 @@ def get_fp32_grad_partitions(self) -> Dict[int, Dict[int, Tensor]]:
 
         return grad_dict
 
+    def get_fp32_grad_for_param(self, param) -> Tensor:
+        self.__reduce_and_partition_stream.synchronize()
```
stas00, line 2157:
```suggestion
    def get_fp32_grad_for_param(self, param) -> Tensor:

        if not param.requires_grad:
            return None

        self.__reduce_and_partition_stream.synchronize()
```
this takes care of frozen params

stas00:
one more but important small todo item - doc - should we add it to some file under `doc` in the repo?

Something along the lines:

```
        backward(loss)
        [...]
        from deepspeed.utils import safe_get_full_fp32_param, safe_get_full_grad, safe_get_full_optimizer_state
        # must be called after `backward` and before `step`
        for n, lp in self.vl_model.named_parameters():
            hp = safe_get_full_fp32_param(lp)
            exp_avg = safe_get_full_optimizer_state(lp, "exp_avg")
            exp_avg_sq = safe_get_full_optimizer_state(lp, "exp_avg_sq")
            hp_grad = safe_get_full_grad(lp)
        [...]
       optimizer.step()
```

tjruwase:
@stas00, would this be an appropriate location for the doc? 
https://deepspeed.readthedocs.io/en/latest/zero3.html#

stas00:
Yes, that would be perfect!

tjruwase:
@stas00, please review the docs draft here: https://deepspeed.readthedocs.io/en/rtd-staging/zero3.html#debugging

docs/code-docs/source/zero3.rst
```diff
@@ -293,3 +293,37 @@ parallelism to fit them in limited GPU memory.
...
 .. autoclass:: deepspeed.zero.TiledLinear
     :members:
+
+
+Debugging
+---------
+
+Debugging ZeRO training is complicated by the partitioning of parameters, gradients, and optimizer states.
+To simplify debugging, DeepSpeed provides the following routines for accessing individual model states
+in their full form (i.e., un-partitioned).
```
stas00, line 303:
```suggestion
Debugging ZeRO training is complicated by the partitioning of parameters, gradients, and optimizer states.
None of these 3 groups of parameters can be normally accessed because of that. To overcome that DeepSpeed provides the following utilities for accessing individual model states in their unpartitioned form.

Please note that these utilities must be called by all processes, even if you decide to do something with the result only in the main process. If all processes don't participate these utilities will hang waiting for all processes to send their contribution.
```

docs/code-docs/source/zero3.rst
```diff
@@ -293,3 +293,37 @@ parallelism to fit them in limited GPU memory.
...
+
+.. autofunction:: deepspeed.utils.safe_get_full_fp32_param
+
+.. autofunction:: deepspeed.utils.safe_get_full_grad
+
+.. autofunction:: deepspeed.utils.safe_get_full_optimizer_state
+
+
+These routines can be used in a training loop as shown in the following snippet.
```
stas00, line 313: ah, you prefer "routines" instead of "utilities" - that works, let's then rename these back int he paras above, to keep things consistent.

docs/code-docs/source/zero3.rst
```diff
@@ -293,3 +293,37 @@ parallelism to fit them in limited GPU memory.
...
+
+.. code-block:: python
+
+    backward(loss)
+    [...]
+    from deepspeed.utils import safe_get_full_fp32_param, safe_get_full_grad, safe_get_full_optimizer_state
+    for n, lp in model.named_parameters():
+        # 1. grad lookup must be called after `backward` and before `step` for z2
```
stas00, line 321:
```suggestion
        # 1. grad lookup must be called after `backward` and before `step` for zero2
```

stas00: and for zero3?

docs/code-docs/source/zero3.rst
stas00: please review/extend my additions on when the data is valid.

docs/code-docs/source/zero3.rst
```diff
@@ -293,3 +293,38 @@ parallelism to fit them in limited GPU memory.
...
     :members:
+
+
+Debugging
+---------
+
+Debugging ZeRO training is complicated by the partitioning of parameters, gradients, and optimizer states.
+None of these 3 groups of parameters can be normally accessed because of that. To overcome that DeepSpeed provides the following utilities for accessing individual model states in their unpartitioned form.
+
+Please note that routines utilities must be called by all processes, even if you decide to do something with the result only in the main process. If all processes don't participate these utilities will hang waiting for all processes to send their contribution.
```
stas00, line 304:
```suggestion
Debugging ZeRO training is complicated by the partitioning of parameters, gradients, and optimizer states.
None of these 3 groups of tensors (model states) can be normally accessed because of that. To overcome that DeepSpeed provides the following routines for accessing individual model states in their unpartitioned form.

Important: Please note that these utilities must be called by all processes participating in the training, even if you decide to do something with the result only in the main process. If all processes don't participate these utilities will hang waiting for all processes to send their contribution.

Additionally, you must be aware that these routines return correct data only in specific phases of the training. So for examples the gradients are valid after ``backward`` and before ``step``. The optimizer states are updated after ``step``. Same goes for fp32 master weights.
```

stas00: sorry, I was still editing the suggestion - I reloaded it - now it's good to go.

deepspeed/utils/tensor_fragment.py
```diff
@@ -58,9 +62,95 @@ def get_full_hp_param(self, optim_state_key=None):
...
+    if hasattr(param, '_hp_mapping'):
+        return param.get_full_hp_param()
+    return None
+
+
+def safe_get_full_optimizer_state(param, optim_state_key):
+    """Assemble and return the fp32 optimizer state of a low-precision (e.g., fp16) parameter.
+
+        Args:
+            param (``torch.nn.Parameter``): The low-precision parameter
```
stas00, line 115: might it be confusing to the user who might have no idea what low-precision means in this context? perhaps just one of the parameters from the model or something like that?

deepspeed/utils/tensor_fragment.py
```diff
@@ -58,9 +62,95 @@ def get_full_hp_param(self, optim_state_key=None):
...
+        return param.grad
+
+    # ZeRO stage 3 param
+    if hasattr(param, 'ds_id'):
+        return param._z3_optimizer.get_fp32_grad_for_param(param)
+
+    # ZeRO stage 1, 2, and bf16_optimizer params
+    if hasattr(param, '_hp_mapping'):
+        return param.get_full_hp_grad()
+    return None
```
stas00, line 144:
```suggestion
    # ZeRO stage 3 param
    if hasattr(param, 'ds_id'):
        return param._z3_optimizer.get_fp32_grad_for_param(param)

    # ZeRO stage 1, 2, and bf16_optimizer params
    if hasattr(param, '_hp_mapping'):
        return param.get_full_hp_grad()

    if param.grad is not None:
        return param.grad

        return None
```

this solves the hanging. 

no idea why for `size[1]` params, on many gpus, some of them have `param.grad is not None`

jeffra (approved):

deepspeed/runtime/zero/stage3.py
```diff
@@ -408,8 +410,13 @@ def _setup_for_real_optimizer(self):
                     0,
                     offset,
                     param.partition_numel())
+            # self.param_offset[param.ds_id] = offset
```
stas00, line 413:
```suggestion
```

deepspeed/utils/tensor_fragment.py
```diff
@@ -58,9 +62,107 @@ def get_full_hp_param(self, optim_state_key=None):
...
+            raise ValueError(
+                "Gradients are only available immediately after backward and before engine step"
+            )
+
+        lp_grad_fragment = gradient_dict[hp_mapping.param_group_index][
+            self._index_in_param_group]
+        hp_grad_fragment = lp_grad_fragment.to(torch.float32).flatten()
+        # print(
+        #     f'{dist.get_rank()=} {self.shape=}  {hp_grad_fragment.shape=}  {hp_grad_fragment=}'
+        # )
```
stas00, line 86:
```suggestion
```

deepspeed/utils/tensor_fragment.py
```diff
@@ -58,9 +62,107 @@ def get_full_hp_param(self, optim_state_key=None):
...
+
+        lp_frag_address = self._hp_mapping.lp_fragment_address
+        reduce_fragment = torch.narrow(reduce_buffer,
+                                       0,
+                                       lp_frag_address.start,
+                                       lp_frag_address.numel)
+        # print(
+        #     f'{dist.get_rank()=} {lp_frag_address=} {reduce_fragment.shape=} {reduce_fragment=}'
+        # )
```
stas00, line 95:
```suggestion
```

deepspeed/utils/tensor_fragment.py
```diff
@@ -70,9 +172,9 @@ def get_hp_fragment_mapping(lp_param,
 
     fragment_start = max(lp_start, hp_start)
     fragment_end = min(lp_end, hp_end)
-    #            print(
-    #                f'{self.dp_rank=} {lp_start=} {lp_end-lp_start=} {hp_start=} {hp_end-hp_start=} {fragment_start=} {fragment_end-fragment_start=}'
-    #            )
+    # print(
+    #     f'{dist.get_rank()=} {lp_start=} {lp_end-lp_start=} {hp_start=} {hp_end-hp_start=} {fragment_start=} {fragment_end-fragment_start=}'
+    # )
```
stas00, line 177:
```suggestion
```

stas00 (commented):
Not sure if we want to keep any commented out code. Up to you, Tunji!

but otherwise we retested it here - good to merge asap! as my team is juggling multiple DS PRs at the moment so merging this would help a lot.

Thank you!

SaulLu:
`e5d3b54ba897125abb6462dd7111bd75c5c22a49` 

Thank you very much again for this work! This is a really useful API to be able to improve our models and understand what is going on.

I would like to leave a message, because as Stas had passed on the information, on an early commit in this PR we had observed a hang issue on a parameter of size 1. I then tested the PR at the commits `3d68746ebe0e7de0ea4f4cbcaa0696cc1b267670` and `e5d3b54ba897125abb6462dd7111bd75c5c22a49` which solved this hang issue but the last version (commit `b8d7b878318fbabef91a609dddf83d550b3ac795`) generates a hang again.

stas00:
have you by chance validated with `py-spy` that it's the same issue or possibly a new one?

Alternatively, if you apply this patch:

```
diff --git a/deepspeed/utils/tensor_fragment.py b/deepspeed/utils/tensor_fragment.py
index 82047357..a8a6db4a 100644
--- a/deepspeed/utils/tensor_fragment.py
+++ b/deepspeed/utils/tensor_fragment.py
@@ -136,9 +136,6 @@ def safe_get_full_grad(param):
         Args:
             param (``torch.nn.Parameter``): A model parameter
     """
-    if param.grad is not None:
-        return param.grad
-
     # ZeRO stage 3 param
     if hasattr(param, 'ds_id'):
         return param._z3_optimizer.get_fp32_grad_for_param(param)
@@ -147,6 +144,9 @@ def safe_get_full_grad(param):
     if hasattr(param, '_hp_mapping'):
         return param.get_full_hp_grad()

+    if param.grad is not None:
+        return param.grad
+
     return None
```

does the problem go away?

stas00:
That's exciting, Lucile - thank you for the update - we can then focus on solving other issues.

stas00:
Let's merge this, Tunji?

deepspeed/runtime/zero/stage3.py
```diff
@@ -1260,9 +1266,10 @@ def __partition_grads(self,
         offload_fp32_gradients = {}
         offload_fp32_offsets = {}
         for param, grad_partition in zip(params_to_release, grad_partitions):
-            if param.partition_numel() * dist.get_rank(
-                    self.dp_process_group) > param.ds_numel:
+            if param.partition_numel() * (
+                    1 + dist.get_rank(self.dp_process_group)) > param.ds_numel:
```
tjruwase, line 1270: @stas00, turns out this fix is buggy because it will skip when the rank contains a tensor fragment and padding.

stas00: but we added it specifically to solve the problem of another one off.

remember we had 2+6, instead of 1+7, because rank0 always returns 0, and so rank1 gets assigned the role of rank 0

where did you discover the bug - was it the existing test?

tjruwase: This is the failing [test](https://github.com/microsoft/DeepSpeed/blob/81b4d5db06b70ed7bc188568f90d89af32f876b1/tests/unit/runtime/zero/test_zero.py#L532). I think we created a new [unit test](https://github.com/microsoft/DeepSpeed/blob/1b6003bd45f8fdea04e7f7729cdfc878887e70c4/tests/unit/runtime/zero/test_zero_tensor_fragment.py#L83) for the original problem, makes pass with update in my local box.

stas00: I see. 

I still think that logic is problematic - since it tries to count ranks, but multiplies by 0 for the first rank which will always lead to a `False`  as `0 < !0` which is wrong, since the first rank's numel isn't 0.

I think it just happens to work. And we found this totally accidentally when I was debugging the hanging problem.

stas00: So when you undid the fix we applied you actually created a one-off problem, but it's very possible the rest of the code was written to rely on the side effect of this situation and thus it was compensated on elsewhere.

but I'm not insisting that it has to be fixed. If the tests pass it's fine with me. It'll probably re-appear in the future.

tjruwase: Actually, I would like to revisit the hanging problem and the overall correctness of this logic. My new sense is that the condition for skipping is when the rank only contains paddings. In that case, I think the check is whether the first byte of the rank is less than true payload. I would like to sync to chat more on this.

stas00: the new fix we figured out is perfect, can resolve this one.

stas00:
can we try again please Tunji?

tjruwase:
> can we try again please Tunji?

Trying to merge this today but fighting a strange transformers CI issue :(.

stas00:
I just merged something on transformers a few hours ago - perhaps we broke something there - checking your CI

stas00:
It looks like someone broke the `zero_to_fp32.py` file it now starts with:

```

'''Copyright The Microsoft DeepSpeed Team'''

#!/usr/bin/env python
```

stas00:
This is the breakage https://github.com/microsoft/DeepSpeed/pull/2889 - @jeffra

tjruwase:
@stas00, thanks for the quick fix. I am pulling your PR into mine.

Issue #2737, opened 2023-01-23
[BUG] Loading MoE Checkpint with load_optimizer_states=False should tolerate missing optim_state files


clumsy:
**Describe the bug**
Even though not loaded into optimizer state dict for MoE model with `load_optimizer_states=False` the current code attempts to load the files at https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L2843, but they could be (intentionally or not) missing.
The only workaround right now is to use `load_module_only` which is not ideal if we want to load LR scheduler states for example.

**To Reproduce**
1. Delete `optim_state` files from the checkpoint
2. Try loading DeepSpeed checkpoint with `load_optimizer_states=False`
3. Observe `FileNotFoundError`:
```
Traceback (most recent call last):
--
...
_, checkpoint_state_dict = model.load_checkpoint(
File "/usr/local/lib/python3.8/dist-packages/deepspeed/runtime/engine.py", line 2415, in load_checkpoint
load_path, client_states = self._load_checkpoint(load_dir,
File "/usr/local/lib/python3.8/dist-packages/deepspeed/runtime/engine.py", line 2489, in _load_checkpoint
optim_checkpoint = torch.load(optim_load_path,
File "/usr/local/lib/python3.8/dist-packages/torch/serialization.py", line 594, in load
with _open_file_like(f, 'rb') as opened_file:
File "/usr/local/lib/python3.8/dist-packages/torch/serialization.py", line 230, in _open_file_like
return _open_file(name_or_buffer, mode)
File "/usr/local/lib/python3.8/dist-packages/torch/serialization.py", line 211, in __init__
super(_open_file, self).__init__(open(name, mode))
FileNotFoundError: [Errno 2] No such file or directory: '<REDACTED>/expp_rank_0_mp_rank_00_optim_states.pt'

```

**Expected behavior**
Loading checkpoints loads all specified state dicts, but for optimizer - it's initialized from model weights. `client_state` should reflect this.

**ds_report output**
```
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
[...]

PR #2738, opened 2023-01-23
fix load_optimizer_states for MoE (#2737)

clumsy:
When `load_optimizer_states=False` is used for MoE `load_checkpoint` - do not attempt to load the optimizer state files.

This currently fails as DeepSpeed still attempts to load those, even though they are not used afterwards.

Adding parameterized unit tests for various cases.

Verified via: `pytest tests/unit/checkpoint/test_moe_checkpoint.py -k 'test_checkpoint_moe_and_zero'`
```
= 6 passed, 1 deselected, 102 warnings in 156.67s (0:02:36) =
```

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -916,7 +916,7 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
             rank_and_offsets = []
             real_dp_process_group = []
             curr_size = 0
-            prev_id = -1
+            prev_id, prev_process_group = -1, None
 
             process_group = self.dp_process_group
             # count = 0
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/engine.py
```diff
@@ -2908,30 +2916,19 @@ def get_sparse_tensor_module_names(original_set,
...
-            deepspeed_states.append('optimizer')
+            ])
 
         client_state = {
             key: value
             for key,
             value in checkpoint.items() if not key in deepspeed_states
         }
 
-        if not load_optimizer_states and not load_module_only:
```
clumsy, line 2934: I believe this is a bug, we should indicate the `optimizer` state was used when `load_optimizer_states=True` and `load_module_only=False`

clumsy:
Hi @tjruwase, I almost got this to work, but for some reason when I suppress loading optimizer states for Stage3 tensor correctness fails for model parameters in unit test.
Do you have an idea why?
Is there's something besides optimizer states in Stage3 optimizer states dictionary or is there some side-effect of loading Stage3 optimizer for model parameters?

PR #2741, opened 2023-01-23
Change zero_grad() argument to match pytorch

loadams:
Change zero_grad() argument from `set_grads_to_None=True` to `set_to_none=False` to match PyTorch  (docs [here](https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.zero_grad.html)).  This will be a breaking change but is not expected to cause new issues downstream for users.

The only existing code I could find that used the `set_grads_to_None` argument is in the DeepSpeedExamples repo under the now deprecated megatron folder there.

### DIFF
--- a/deepspeed/runtime/fp16/fused_optimizer.py
+++ b/deepspeed/runtime/fp16/fused_optimizer.py
@@ -130,14 +130,14 @@ class FP16_Optimizer(DeepSpeedOptimizer):
 
         return
 
-    def zero_grad(self, set_grads_to_None=True):
+    def zero_grad(self, set_to_none=False):
         """
         Zero FP16 parameter grads.
         """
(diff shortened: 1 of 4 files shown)


tjruwase (approved):

jeffra:
@loadams, I think we also need to change our internal usage of `zero_grad()` -> `zero_grad(set_to_none=True)` since now the default no-arg behavior has changed.

For example I think this call should be changed: https://github.com/microsoft/DeepSpeed/blob/fe728e3ed880f27de2c21234f12b7aa6f672e825/deepspeed/runtime/zero/stage_1_and_2.py#L1769

There may be a few others as well based on a quick search.

jeffra (approved):
LGTM, thanks @loadams :)

PR #2742, opened 2023-01-24
Fixing Optimizer Sanity Check

jomayeri:
Fixing the optimizer sanity check to make BF16 + FP32 gradients + ZeRO1 a special case. All other combinations follow the pattern of matching model dtype to gradient accumulation dtype for ZeRO.


tjruwase (approved):

michaelroyzen:
Why does ZeRO not support fp32 accumulation? Otherwise it's incredibly imprecise. @stas00

stas00:
I agree but I don't work at MSFT :)

ping @tjruwase

PR #2743, opened 2023-01-24
Fix hardcoded instances to fp16 in optimizer creation log messages to the correct dtype.

loadams:
Fix places where in log messages we have hardcoded fp16 as the dtype but bf16 is also supported and should be correctly printed to the user.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1433,7 +1433,6 @@ class DeepSpeedEngine(Module):
         initial_dynamic_scale = self.initial_dynamic_scale()
         dynamic_loss_args = self.dynamic_loss_scale_args()
         clip_grad = self.gradient_clipping()
-        model_dtype, grad_accum_dtype = self.get_data_types()
         if APEX_INSTALLED:
             fused_opts = (apex.optimizers.FusedAdam, FusedAdam)
         else:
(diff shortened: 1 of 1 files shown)


stas00 (commented):
Looks great, thank you, @loadams!

deepspeed/runtime/engine.py
```diff
@@ -1433,14 +1433,16 @@ def _configure_fp16_optimizer(self, optimizer):
...
+        model_dtype, grad_accum_dtype = self.get_data_types()
         if APEX_INSTALLED:
             fused_opts = (apex.optimizers.FusedAdam, FusedAdam)
         else:
             fused_opts = FusedAdam
         if isinstance(optimizer, fused_opts) \
                 or self.optimizer_name() in [ONEBIT_ADAM_OPTIMIZER, ZERO_ONE_ADAM_OPTIMIZER]:
             if self.dynamic_loss_scale():
-                log_dist("Creating fp16 optimizer with dynamic loss scale", ranks=[0])
+                log_dist(f'Creating {model_dtype} optimizer with dynamic loss scale',
```
SaulLu, line 1444: For this message, we initialize only the `FP16_Optimizer` class afterwards, it is perhaps safer not to change the message in this case, no?

deepspeed/runtime/engine.py
```diff
@@ -1470,7 +1470,7 @@ def _configure_fp16_optimizer(self, optimizer):
                     has_moe_layers=self.has_moe_layers,
                 )
         else:
-            log_dist("Creating fp16 unfused optimizer with dynamic loss scale",
+            log_dist(f'Creating {model_dtype} unfused optimizer with dynamic loss scale',
```
SaulLu, line 1473: I would have the same question for this message since the only optimizer that is initialized after is `optimizer = FP16_UnfusedOptimizer`

SaulLu (commented):
Thanks for working on clarifying the messages, I have left two comments in places where I wonder if the change is not conducive to letting an unwanted setup through

deepspeed/runtime/engine.py
SaulLu: Thank you! Everything looks good for me

PR #2745, opened 2023-01-24
Pin transformers unit test version

mrwyattii:
We are seeing an error pop up in the transformers unit test after a recent update. Pinning the version of transformers used until we can do a deeper dive in the cause of this error.


mrwyattii:
This may have been resolved by https://github.com/huggingface/transformers/pull/21283, Do not merge

jeffra:
I think we can close this, it appears fixed now:
https://github.com/microsoft/DeepSpeed/actions/runs/3999217629/jobs/6862843895

PR #2761, opened 2023-01-30
modify engine.py for formatting

xiexbing:


deepspeed/runtime/engine.py
```diff
@@ -114,12 +112,11 @@
 
 
 def split_half_float_double_sparse(tensors):
-    device_type = get_accelerator().device_name()
     supported_types = [
-        "torch.{}.HalfTensor".format(device_type),
-        "torch.{}.FloatTensor".format(device_type),
-        "torch.{}.DoubleTensor".format(device_type),
-        "torch.{}.BFloat16Tensor".format(device_type),
+        "torch.cuda.HalfTensor",
```
tjruwase, line 116: This is undoing the accelerator abstraction PR. Maybe this is a bad conflict resolution?

deepspeed/runtime/engine.py
```diff
@@ -225,7 +229,7 @@ def __init__(
         self.eigenvalue = None
         self.block_eigenvalue = None
         self.gas_boundary_ctr = 0
-        self.dist_backend = get_accelerator().communication_backend_name()
+        self.dist_backend = "nccl"
```
tjruwase, line 232: Ditto

xiexbing:
OK, I guess I merged from an older version of engine.py which doesn't reflect the new changes on engine.py for other purposes. I'll simply modify my part from an updated engine.py.

PR #2766, opened 2023-01-30
Create tensor parallelism blog/tutorial

molly-smith:
Creates blog for automatric tensor parallelism feature.


docs/_tutorials/automatic-tensor-parallelism.md
```diff
@@ -0,0 +1,129 @@
...
+# Initialize the DeepSpeed-Inference engine
+pipe.model = deepspeed.init_inference(
+    pipe.model,
+    mp_size=world_size,
+    dtype=torch.float
+)
+output = pipe('Input String')
+```
+
+Previously, to run inference with only tensor parallelism for the models that we don't support kernels, you can pass an injection policy that shows the two specific linear layers on a Transformer Encoder/Decoder layer: 1) the attention output GeMM and 2) layer output GeMM. We need these parts of the layer to add the required all-reduce communication between GPUs to merge the partial results across model-parallel ranks. Below, we bring an example that shows how you can use deepspeed-inference with a T5 model:
```
awan-10, line 22: Should be past tense for this entire paragraph.

docs/_tutorials/automatic-tensor-parallelism.md
```diff
@@ -0,0 +1,129 @@
...
+
+To enable tensor parallelism, you need to set the `ds_inference` to True for the compatible models:
+
+```bash
+deepspeed --num_gpus <num_gpus> DeepSpeedExamples/inference/huggingface/text-generation/inference-test.py --name <model> --batch_size <batch_size> --ds_inference
+```
+
+## OPT 13B Inference Performance Comparison
+
+The following results were collected using V100 SXM2 32GB GPUs.
```
jeffra, line 65: (from discussion) let's add the flag(s) used to reproduce the numbers below. Should we also add mem bandwidth below to the table?

docs/_tutorials/automatic-tensor-parallelism.md
```diff
@@ -0,0 +1,129 @@
+---
+title: "Automatic Tensor Parallelism for HuggingFace Models"
+tags: inference
+---
+
+This tutorial demonstrates the new automatic tensor parallelism feature for inference. Previously, the user needed to provide an injection policy to DeepSpeed to enable tensor parallelism. DeepSpeed now supports automatic tensor parallelism for HuggingFace models by default as long as kernel injection is not true and an injection policy is not provided. This is convenient for when the injection policy of a model is not known and for improving performance of models without kernel injection support.
```
mrwyattii, line 6:
```suggestion
This tutorial demonstrates the new automatic tensor parallelism feature for inference. Previously, the user needed to provide an injection policy to DeepSpeed to enable tensor parallelism. DeepSpeed now supports automatic tensor parallelism for HuggingFace models by default as long as kernel injection is not enabled and an injection policy is not provided. This allows our users to improve performance of models that are not currently supported via kernel injection, without providing the injection policy.
```

Small suggestion with wording of the last sentence.

docs/_tutorials/automatic-tensor-parallelism.md
```diff
@@ -0,0 +1,129 @@
...
+import deepspeed
+pipe = pipeline("text2text-generation", model="google/t5-v1_1-small", device=local_rank)
+# Initialize the DeepSpeed-Inference engine
+pipe.model = deepspeed.init_inference(
+    pipe.model,
+    mp_size=world_size,
+    dtype=torch.float
+)
+output = pipe('Input String')
+```
```
mrwyattii, line 20:
```suggestion
import os
import torch
import transformers
import deepspeed

local_rank = int(os.getenv("LOCAL_RANK", "0"))
world_size = int(os.getenv("WORLD_SIZE", "1"))

# create the model pipeline
pipe = transformers.pipeline(task="text2text-generation", model="google/t5-v1_1-small", device=local_rank)
# Initialize the DeepSpeed-Inference engine
pipe.model = deepspeed.init_inference(
    pipe.model,
    mp_size=world_size,
    dtype=torch.float
)
output = pipe('Input String')
```

We should probably make the example code runnable if a user were to copy, paste, and run it.

docs/_tutorials/automatic-tensor-parallelism.md
```diff
@@ -0,0 +1,129 @@
...
+pipe = pipeline("text2text-generation", model="google/t5-v1_1-small", device=local_rank)
+# Initialize the DeepSpeed-Inference engine
+pipe.model = deepspeed.init_inference(
+    pipe.model,
+    mp_size=world_size,
+    dtype=torch.float,
+    injection_policy={T5Block: ('SelfAttention.o', 'EncDecAttention.o', 'DenseReluDense.wo')}
+)
+output = pipe('Input String')
+```
```
mrwyattii, line 38: Same comment as above code snippet, let's make it runnable if a user were to copy and paste.

Additionally, would it make more sense to have this example first? We show the old way of doing things where the user must specify the `injection_policy` value and then show that we can achieve the same without the user providing that thanks to the new automatic tensor parallelism.

docs/_tutorials/automatic-tensor-parallelism.md
molly-smith: Hi @jeffra, I included the test_performance flag. I didn't include the bandwidth numbers because I felt Tflops captured enough information for the comparison. I also didn't save the bandwidth numbers but I could calculate them from the Tflops results.

docs/_tutorials/automatic-tensor-parallelism.md
molly-smith: That's a good point. I referenced the same example from the inference tutorial page. I had assumed it was runnable myself.

docs/_tutorials/automatic-tensor-parallelism.md
molly-smith: I originally had this example first but @awan-10 suggested I reverse it.

lekurile (approved):
Looks good to me!

docs/_tutorials/automatic-tensor-parallelism.md
```diff
@@ -0,0 +1,129 @@
+---
+title: "Automatic Tensor Parallelism for HuggingFace Models"
+tags: inference
+---
+
+This tutorial demonstrates the new automatic tensor parallelism feature for inference. Previously, the user needed to provide an injection policy to DeepSpeed to enable tensor parallelism. DeepSpeed now supports automatic tensor parallelism for HuggingFace models by default as long as kernel injection is not enabled and an injection policy is not provided. This allows our users to improve performance of models that are not currently supported via kernel injection, without providing the injection policy.
+
+```python
+# create the model
```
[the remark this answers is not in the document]
molly-smith: Thanks for the suggestion. I think headers are a good idea so I added them.

jeffra (approved):

PR #2783, opened 2023-02-02
Add hjson support for user configs

mrwyattii:
Also added tests to verify DeepSpeed configs are properly loaded from dict, json, and hjson formats.

@stas00 @jeffra


stas00:
whoohoo! Finally this is happening! 

Thank you so much, Michael and the team!

lekurile (approved):

Issue #2785, opened 2023-02-03
[BUG] wrong `overlapping_partition_gradients_reduce_epilogue` comment?


stas00:
**Describe the bug**

It says it runs only for stage 2,
https://github.com/microsoft/DeepSpeed/blob/c5b983e92ec71d0bbd1deb110537db4eafbe1fb5/deepspeed/runtime/engine.py#L1907-L1909
but I see it running for stage 3 as well when stepping with debugger.

tjruwase:
Yes, the comment was true when only stage 1 and 2 existed, and should have been deleted since stage 3 includes stage 2.

stas00:
thank you, Tunji: suggested PR here: https://github.com/microsoft/DeepSpeed/pull/2786

PR #2786, opened 2023-02-03
remove outdated comment

stas00:
Fixes: https://github.com/microsoft/DeepSpeed/issues/2785#issuecomment-1414875492

Tunji:
> Yes, the comment was true when only stage 1 and 2 existed, and should have been deleted since stage 3 includes stage 2.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1904,7 +1904,6 @@ class DeepSpeedEngine(Module):
         self.optimizer.is_gradient_accumulation_boundary = self.is_gradient_accumulation_boundary(
         )
 
-        # ZeRO stage 2 communicates during non gradient accumulation boundaries as well
         if self.zero_optimization_partition_gradients():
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/engine.py
```diff
@@ -1904,7 +1904,6 @@ def allreduce_gradients(self, bucket_size=MEMORY_OPT_ALLREDUCE_SIZE):
         self.optimizer.is_gradient_accumulation_boundary = self.is_gradient_accumulation_boundary(
         )
```
tjruwase, line 1906:
```suggestion
# ZeRO stage >= 2 communicates during non gradient accumulation boundaries as well
```

tjruwase (approved):

deepspeed/runtime/engine.py
stas00: Thank you for the correct fix, Tunji!

jeffra (approved):

Issue #2602, opened 2022-12-13
[BUG]injection_policy_tuple   value is not a valid tuple (type=type_error.tuple)


chaoyujiang:
**Describe the bug**

Traceback (most recent call last):
  File "./test_ds.py", line 49, in <module>
    replace_with_kernel_inject=True)
  File "/home/yjchao/venv_37/lib/python3.7/site-packages/deepspeed/__init__.py", line 311, in init_inference
    engine = InferenceEngine(model, config=ds_inference_config)
  File "/home/yjchao/venv_37/lib/python3.7/site-packages/deepspeed/inference/engine.py", line 121, in __init__
    config.injection_policy_tuple = injection_policy
  File "pydantic/main.py", line 385, in pydantic.main.BaseModel.__setattr__
pydantic.error_wrappers.ValidationError: 1 validation error for DeepSpeedInferenceConfig
injection_policy_tuple
  value is not a valid tuple (type=type_error.tuple)
[2022-12-13 17:01:34,118] [INFO] [launch.py:318:sigkill_handler] Killing subprocess 88447

code:

model_name_or_path = '/root/model/opt/opt-13b/'
model = OPTForCausalLM.from_pretrained(model_name_or_path, torch_dtype=torch.float16)

batch_size = 8
seq_length = 320
iter_num = 1
max_length = 1
in_ids = np.random.randint(low=1, high=3000, size=(batch_size, seq_length)).astype(
    np.int64
)
in_ids = torch.Tensor(in_ids).long().to(local_rank)

injection_policy={opt_transformer: module_inject.replace_policy.HFOPTLayerPolicy}
model = deepspeed.init_inference(model,
                                 mp_size=1,
                                 dtype=data_type,
                                 injection_policy=injection_policy,
                                 replace_method='auto',
                                 replace_with_kernel_inject=True)

**To Reproduce**
1. Simple inference script to reproduce
2. What packages are required and their versions
3. How to run the script
4. ...

**Expected behavior**

**ds_report output**

(venv_37) [root@iZ2ze85vftf3u8xeax2cjkZ yjchao]# ds_report
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
[...]

PR #2791, opened 2023-02-03
Allow abcmeta types for injection policy for container policies

molly-smith:
Fix "not valid tuple"  when using container policies: https://github.com/microsoft/DeepSpeed/issues/2602

Users have been working around the issue by "tuple-ing" the policy in inference engine then "un-tuple-ing" in replace_module.py. Instead we can allow tuple and ABCMeta types for policies.


molly-smith:
closing since issue has been resolved with https://github.com/microsoft/DeepSpeed/pull/2831

PR #2792, opened 2023-02-03
Add container load checkpoint error reporting + refactor

lekurile:
This PR refactors the organization of meta tensor checkpoint loading as follows:
* Move `get_param_names()` abstract method definition from `TransformerPolicy` into `MetaTensorContainer`
* Model-specific `get_param_names()` definitions moved from policy into model-specific container
* `selected_policy_g`, `megatron_v2_g`, and `transformer_config_g` globals replaced with a single `container_g` global, since the container will contain all of the information those globals previously captured
* `ckpt_load_enabled` flag added to containers that's set to `False` by default in the `base.py` container and gets set to `True` when the `MetaTensorContainer` feature is inherited
* Assertion added to `replace_transformer_layer` before performing checkpoint loading to check if `ckpt_load_enabled == True`, otherwise an error message will be printed saying that the container does not support meta tensor checkpoint loading.

The aim of these changes is to more closely couple meta tensor checkpoint loading code to the `MetaTensorContainer` and to allow for better error reporting of load checkpoint use on model types that don't support this feature.

### DIFF
--- a/deepspeed/module_inject/containers/base.py
+++ b/deepspeed/module_inject/containers/base.py
@@ -21,6 +21,7 @@ class BaseTransformerContainer(ABC):
 
         self.megatron_v2 = self.policy.is_megatron_v2
         self.scale_attention = self.policy.scale_attention
+        self.ckpt_load_enabled = False
 
         # configuration for models. todo: can this be moved to a pydantic model config?
         self.hidden_size = None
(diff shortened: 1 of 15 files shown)


awan-10 (approved):
Excellent work on refactoring this. Thank you, Lev. The PR looks good to me.

PR #2796, opened 2023-02-06
Update ROCm version

rraminen:
This PR is to updated the ROCm version from 5.1.1 to 5.3 in amd.yml

cc: @jithunnair-amd

### DIFF
--- a/.github/workflows/amd.yml
+++ b/.github/workflows/amd.yml
@@ -29,7 +29,7 @@ jobs:
 
       - name: Install pytorch
         run: |
-          pip install torch torchvision --extra-index-url https://download.pytorch.org/whl/rocm5.1.1
+          pip install torch torchvision --extra-index-url https://download.pytorch.org/whl/rocm5.3
           python -c "import torch; print('torch:', torch.__version__, torch)"
           python -c "import torch; print('CUDA available:', torch.cuda.is_available())"
 


tjruwase (approved):

.github/workflows/amd.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
       - name: Install pytorch
         run: |
-          pip install torch torchvision --extra-index-url https://download.pytorch.org/whl/rocm5.1.1
+          pip install torch torchvision --extra-index-url https://download.pytorch.org/whl/rocm5.3
```
mrwyattii, line 32: This line causes the following version of torch to be installed on our CI nodes:
`1.13.1+cu117`

It doesn't look like torch has released wheels compiled with 5.3 - is this a mistake?

mrwyattii (changes requested):

.github/workflows/amd.yml
rraminen: Hi @mrwyattii, you are correct, rocm 5.3 is installing torch version `1.13.1+cu117`. Where as 5.1.1 installs torch version `1.13.1+rocm5.1.1`. 

@jithunnair-amd, should we not proceed with this PR? Please advice.

jithunnair-amd: Hey folks, if you're looking for the latest wheels, those would be at: https://download.pytorch.org/whl/rocm5.4.2/. Is that what you want?

jithunnair-amd: lol, just saw this is an old PR from our side. But yes, if we want to update ROCm version used, it would be 5.4.2. We can go ahead and update the PR. cc @rraminen

Issue #2733, opened 2023-01-20
Tests should fail indicating actual number of GPUs is below desired world_size


clumsy:
I faced this cryptic error while running tests on a device with a single GPU.

DeepSpeed: `master`
PyTroch: `1.12.1`
NCCL: `2.10.3`

### Current Behavior

Steps to reproduce:
1. `pytest tests/unit/checkpoint/test_moe_checkpoint.py -k 'test_checkpoint_moe_and_zero'`
2. Observe:
```
Traceback (most recent call last):
  File "/opt/conda/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/opt/conda/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/opt/conda/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/conda/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/azzhipa/workspace/m5/DeepSpeed/tests/unit/common.py", line 165, in _dist_init
    dist.barrier()
  File "/home/azzhipa/workspace/m5/DeepSpeed/tests/unit/common.py", line 165, in _dist_init
    dist.barrier()
  File "/home/azzhipa/workspace/m5/DeepSpeed/deepspeed/comm/comm.py", line 127, in log_wrapper
    return func(*args, **kwargs)
  File "/home/azzhipa/workspace/m5/DeepSpeed/deepspeed/comm/comm.py", line 127, in log_wrapper
    return func(*args, **kwargs)
  File "/home/azzhipa/workspace/m5/DeepSpeed/deepspeed/comm/comm.py", line 469, in barrier
    return cdb.barrier(group=group, async_op=async_op, device_ids=device_ids)
  File "/home/azzhipa/workspace/m5/DeepSpeed/deepspeed/comm/comm.py", line 469, in barrier
    return cdb.barrier(group=group, async_op=async_op, device_ids=device_ids)
  File "/home/azzhipa/workspace/m5/DeepSpeed/deepspeed/comm/torch.py", line 160, in barrier
    return torch.distributed.barrier(group=group,
  File "/home/azzhipa/workspace/m5/DeepSpeed/deepspeed/comm/torch.py", line 160, in barrier
    return torch.distributed.barrier(group=group,
[...]

Issue #2482, opened 2022-11-07
[BUG] NCCL rank 1 and rank 0 on same GPU during pytest for test_lr_scheduler.py - single CUDA GPU computer


Thomas-MMJ:
**Describe the bug**
when running

`pytest ./tests/unit/checkpoint/test_lr_scheduler.py`

It fails with
```
_____________________________________________ TestLRSchedulerCheckpoint.test_checkpoint_lr_scheduler[0-False] ______________________________________________
Worker 0 exited with code 1
------------------------------------------------------------------- Captured stdout call -------------------------------------------------------------------
[2022-11-07 09:30:39,330] [INFO] [comm.py:633:init_distributed] Initializing TorchBackend in DeepSpeed with backend nccl
------------------------------------------------------------------- Captured stderr call -------------------------------------------------------------------
Process Process-1:
Process Process-2:
Traceback (most recent call last):
Traceback (most recent call last):
  File "/home/username/anaconda3/envs/diffusers/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/username/anaconda3/envs/diffusers/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/username/anaconda3/envs/diffusers/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/username/anaconda3/envs/diffusers/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/mnt/c/Users/username/DeepSpeed/tests/unit/common.py", line 166, in _dist_init
    dist.barrier()
  File "/mnt/c/Users/username/DeepSpeed/tests/unit/common.py", line 166, in _dist_init
    dist.barrier()
  File "/home/username/anaconda3/envs/diffusers/lib/python3.9/site-packages/deepspeed/comm/comm.py", line 126, in log_wrapper
    return func(*args, **kwargs)
  File "/home/username/anaconda3/envs/diffusers/lib/python3.9/site-packages/deepspeed/comm/comm.py", line 126, in log_wrapper
    return func(*args, **kwargs)
  File "/home/username/anaconda3/envs/diffusers/lib/python3.9/site-packages/deepspeed/comm/comm.py", line 458, in barrier
    return cdb.barrier()
[...]

PR #2799, opened 2023-02-07
Check device count before running dist tests

HeyangQin:


HeyangQin:
@tjruwase I reworked the previous PR. This PR would check GPU count against world size for all dist tests so it avoids issues like https://github.com/microsoft/DeepSpeed/issues/2733 and https://github.com/microsoft/DeepSpeed/issues/2482 for all the dist unit tests

tjruwase (approved):

tests/unit/common.py
```diff
@@ -317,6 +317,15 @@ def __call__(self, request):
...
             world_size = self.world_size
 
+        if isinstance(world_size, int):
+            max_world_size = world_size
+        else:
+            max_world_size = max(world_size)
+        if torch.cuda.device_count() < max_world_size:
+            pytest.skip(
+                f"Skipping test because not enough GPUs are available: {max_world_size} required, {torch.cuda.device_count()} available"
+            )
```
mrwyattii, line 327: I think this check needs to happen in the `DistributedExec` class or we may still run into problems with `DistributedFixture`. Also, we still want to run a test that has multiple world sizes if possible. With the current changes, a test that has world sizes of [2,4] would get skipped altogether on a machine with 2 GPUs.

I suggest adding the following lines:
```python
if torch.cuda.device_count() < num_procs:
    pytest.skip(f"Skipping test because not enough GPUs are available: {num_procs} required, {torch.cuda.device_count()} available")
```
To this method:
https://github.com/microsoft/DeepSpeed/blob/b6030a913b8cb14aa5f9529d8d7854cd15acefc0/tests/unit/common.py#L107

mrwyattii (changes requested):

HeyangQin:
@mrwyattii Hi Michael, could you take a look at the updated PR? Thanks!

mrwyattii (approved):

PR #2803, opened 2023-02-08
deepspeed/runtime/utils.py: reset_peak_memory_stats when empty cache

guoyejun:


guoyejun:
to make the logic clear.   The peek stats should be cleared after cache is empty.

tjruwase (approved):

PR #2805, opened 2023-02-08
ZeRO 1 and 2 Gradient Accumulation Dtype.

jomayeri:
Adding configurable gradient accumulation data type to ZeRO stages 1 and 2.


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -730,36 +733,38 @@ def independent_gradient_partition_epilogue(self):
...
-        # All gradients required by the step
-        # are in self.averaged_gradients
-        self.zero_grad(set_to_none=True)
-        see_memory_usage(f"End ipg_epilogue")
+        if after_reduction:
+            self._release_ipg_buffers()
+            # No need to keep the gradients anymore.
+            # All gradients required by the step
+            # are in self.averaged_gradients
+            self.zero_grad(set_to_none=True)
```
jomayeri, line 767: @tjruwase In Stage 1 should we wait to zero param.grad until after the reduction?

tjruwase: Yes

PR #2807, opened 2023-02-08
Make z3 respect comm dtype

### LINKED ISSUE #2789: [BUG] Communication dtype is broken in zero 3
https://github.com/microsoft/DeepSpeed/blob/0a73e6e6137a91e1b776f725b637f8b37a75f8e7/deepspeed/runtime/zero/stage3.py#L1168

Only works for float32.

tjruwase:
Fix #2789 
@stas00


deepspeed/runtime/zero/stage3.py
```diff
@@ -213,7 +213,7 @@ def __init__(self,
         self.reduce_bucket_size = int(reduce_bucket_size)
 
         if self.reduce_scatter:
-            assert self.communication_data_type in (torch.float16, torch.bfloat16), f"ZeRO-3 supports only float16 or bfloat16 communication_data_type with reduce scatter enabled. Got: '{self.communication_data_type}'"
+            assert self.communication_data_type in (torch.float16, torch.bfloat16, torch.float32), f"ZeRO-3 supports only float16 or bfloat16 communication_data_type with reduce scatter enabled. Got: '{self.communication_data_type}'"
```
stas00, line 216: Tunji, the message is mismatching now - missing fp32

and if all 3 are ok, then why do we need this assert in the first place?

tjruwase: 1. Good catch will fix the assert message. 
2. Maybe 8bit or 64bit might not work? I am not sure, either.

stas00: The change was added here:

https://github.com/microsoft/DeepSpeed/pull/1487/files#diff-e6635a81d2c2bf0938b5f83b1c4945e0f344e0106191a6960df8ee4aa64cc55fL1313-L1316

stas00: The original check was:

```
assert (
            not self.allreduce_always_fp32()
        ), "ZeRO does not support 'fp32_allreduce': true"
```

any idea why and whether this is still true? if it is then we potentially broke something with your commit this morning.

tjruwase: I don't think that is still true. That was when ZeRO did not work for fp32 training.

stas00: then remove that assert completely please.

tjruwase: Will do

tjruwase: Assert in engine.py removed.

jomayeri (approved):

Issue #2760, opened 2023-01-29
[BUG] Zeros key and value cache when get_present = True


mzusman:
**Describe the bug**
Running a forward pass with `get_present = True` returns wrong values for key and value tensor. They contain a lot of zeros.

**To Reproduce**
1. Simple inference script to reproduce

```python
import os
import time
import deepspeed
import torch
import torch.nn as nn
from deepspeed.module_inject.replace_policy import TransformerPolicy, DSPolicy
from torch.nn import Parameter,LayerNorm

class TransformerInferenceCustomPolicy(TransformerPolicy):
    _orig_layer_class = None
    def __init__(self, client_module,inference):
        super().__init__(inference)
        self.client_module = client_module
        if TransformerInferenceCustomPolicy._orig_layer_class is None:
            TransformerInferenceCustomPolicy._orig_layer_class = DeepSpeedTransformerInferenceCustomLayer
        pass

    def get_hidden_heads(self):
        return self.client_module.attention.query_key_value.weight.shape[1], \
                self.client_module.attention.heads

    def attention(self):
        qkvw = self.client_module.attention.query_key_value.weight
        qkvb = self.client_module.attention.query_key_value.bias
        return self.linear_layer, \
            qkvw, \
            qkvb, \
            self.client_module.attention.dense.weight, \
            self.client_module.attention.dense.bias, \
            self.scale_attention, \
            self.is_megatron_v2

    def mlp(self):
        return self.linear_layer, \
            self.client_module.mlp.dense_to_4h.weight,\
            self.client_module.mlp.dense_to_4h.bias, \
            self.client_module.mlp.dense_to_h.weight, \
            self.client_module.mlp.dense_to_h.bias


    def layerNorm(self):
        attention_layernorm = self.client_module.post_attention_layernorm
        transformer_layernorm = self.client_module.input_layernorm
        return attention_layernorm.weight, \
            attention_layernorm.bias, \
            transformer_layernorm.weight, \
            transformer_layernorm.bias

class WeightBiasContainer(torch.nn.Module):
    # Wrapper for a param containing weight and bias
    def __init__(self, weight, bias):
        super().__init__()
        self.weight = weight
        self.bias = bias


class DeepSpeedMLPCustom(nn.Module):
    def __init__(self, hidden_size, intermediate_multiplier=4, layernorm_epsilon=1e-5):
        super().__init__()
        intermediate_size = hidden_size * intermediate_multiplier

        inter_w = nn.Parameter(
[...]

PR #2828, opened 2023-02-14
fix return prev key and value , added strides to from_blob

mzusman:
fixing issue https://github.com/microsoft/DeepSpeed/issues/2760 
basically just fixing from_blob strides , before the fix the functions returned key and value cache that was populated by zeros.


jeffra (approved):
Excellent, thank you @mzusman!

Also reviewed by @cmikeh2

mzusman:
@microsoft-github-policy-service agree company="AI21 Labs"

PR #2834, opened 2023-02-15
fix inference MLP all_reduce dislocation bug

baodii:


baodii:
@CaoZhongZ

CaoZhongZ:
We believe that when distribute a GEMM by cutting K dimension, the residual addition should come after the allreduce. Or residual will be added multiple times.

PR #2847, opened 2023-02-17
ZeRO Gradient Accumulation Dtype.

jomayeri:
Adding configurable gradient accumulation data type to all ZeRO stages.
Added two gradient attributes to the parameter object:
1. `param.grad_reduc`: Used only as a pointer to reference which gradient tensor should be used in reduction. The tensor object pointed to by this attribute is used in all reduction operations.
2. `param.grad_accum`: Used as a pointer or standalone tensor representing the accumulated gradients. The tensor object pointed to by this attribute is used in all accumulation operations.

For stages 1 & 2.
Two modes:
1. `model_dtype == gradient_accumulation_dtype`
2. `model_dtype != gradient_accumulation_dtype`

**Stage 1, Mode 1**:
- `param.grad_reduc = param.grad`
- `param.grad_accum = param.grad`

**Stage 1, Mode 2**:
- `param.grad_reduc = param.accum_grad`
- `param.grad_accum = standalone tensor`

**Stage 2, Mode 1**:
- `param.grad_reduc = param.grad`
- `param.grad_accum = param.grad`

**Stage 2, Mode 2**:
- `param.grad_reduc = param.grad`
- `param.grad_accum = standalone tensor`


tjruwase:
@stas00, FYI

stas00:
Thank you very much for working on this important feature, Joe!

stas00 (commented):
While reading the proposed changes I struggled to have clarify about the various similar yet different names/attributes. Probably if I were to read it some 10 times I would get used to it. 

So here is some food for thought, but feel free to ignore if it doesn't resonate.


1. name starting with `.grad` are actual gradients, so `s/accum_grad/grad_accum/`

- `param.grad_accum` (this is a variation on `.grad`)

and when there is an assignment:

`param.grad_accum = param.grad`

this is loud and clear

also this aligns well with: `self.grad_accum_dtype`

Same in context:
```
param.grad_accum = param.grad
if self.grad_accum_dtype: ...
```
this seems to flow well for me.

2. then there is the bool flag

`self.reduce_accum_grad` - it'd be confusing to revert the order of words if suggestion (1) resonates, and also I'm not 100% it's clear from that attribute it's a bool flag (due to abbreviations we don't know what's meant)

So perhaps `self.should_reduce_grad_accum`? or at least: `self.reduce_grad_accum`

----------------

Same logic applies to `param.reduc_grad` and here we also get into a clash with `self.reduce_accum_grad` - 2 different abbreviations starting with `reduc`.

Perhaps let's spell each word fully out first and then decide what needs to be abbreviated in what way?

----------------

And here is another combo that reads weird:

```
gradient_accumulation_dtype=grad_accum_dtype,
[...]
self.grad_accum_dtype = gradient_accumulation_dtype
```
sometimes full spelling, other times abbreviated...

stas00:
This is much easier to follow, @jomayeri - thank you for the renames!

deepspeed/runtime/engine.py
```diff
@@ -1208,13 +1208,9 @@ def _do_optimizer_sanity_check(self, basic_optimizer):
                         "**** You are using ZeRO with an untested optimizer, proceed with caution *****"
                     )
 
-            if model_dtype == torch.bfloat16 and grad_accum_dtype == torch.float32 and self.zero_optimization_stage(
-            ) == 1:
-                return BFLOAT16
-
-            if model_dtype != grad_accum_dtype:
+            if model_dtype != grad_accum_dtype and self.zero_optimization_stage() == 3:
```
tjruwase, line 1211: This change makes bf16_optimizer unusable, right?

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -833,6 +850,64 @@ def increment_value(dictionary, key):
...
+            param.grad_reduc = param.grad
+
+    # Called after a memory defrag moves tensor object via copy
+    def point_grad_to_reduc(self, param):
+        if self.use_grad_accum_for_reduction:
+            param.grad_accum = param.grad_reduc
+        else:
+            param.grad = param.grad_reduc
+
+    def create_set_grad_accum_hooks(self):
```
tjruwase, line 893: Hooks can incur runtime overheads that could be non-trivial for zero stage 1, so I would like to avoid them. Couldn't we call `set_grad_accum` for all params in the optimizer constructor?

stas00:
@jomayeri, if I may share a useful gh trick with you - you might want to add:

```
Fixes: https://github.com/microsoft/DeepSpeed/issues/2352
```

to the OP and it'll automatically link to close the issue on the merge of this one. And the others will also know which PR to try out when they land on the Issue page.

deepspeed/runtime/engine.py
jomayeri: Correct, I believe it replaces all use cases of the current BF16 Optimizer.

deepspeed/runtime/zero/stage_1_and_2.py
jomayeri: It could be replaced by nested for loops in the backward pass.

deepspeed/runtime/engine.py
tjruwase: @stas00, what do you think? We probably need more testing to validate correctness parity before disabling bf16_optimizer. 

Also, bf16_optimizer has superior memory efficiency, and probably throughput. We need to validate that as well.

deepspeed/runtime/zero/stage_1_and_2.py
tjruwase: Let's chat about this.

deepspeed/runtime/engine.py
stas00: I'm not sure what you are asking, Tunji. 

We know for sure that some people out there use Megatron-Deepspeeed (as we get Issues), so it's almost certain someone will want the BF16Optimizer in a working condition. If that's the question.

tjruwase: Yes, that is the question. How to ensure BF16Optimizer users retain access?

tjruwase: Sorry for not being clear. Let me back up a bit. With this PR, both zero_stage_1 now supports `model_dtype=bf16` and `accum_dtype=fp32` similar to bf16_optimizer.  And so, with that dtypes combination and `stage=1`, zero_stage_1 optimizer will now be instantiated instead of bf16_optimizer. I don't think this is desirable, at least not until we validated the correctness and performance equivalence. So, @jomayeri, can you please restore the old lines 1211--1213 and place it above the new lines 1211--1213. In other words, we need both logics. We can chat more for clarifications. Thanks!

stas00: I thought this was the case, thank you for elaborating, Tunji. 

As I was suggesting a year ago we needed some specific flag in the config that activates `BF16Optimizer`. It keeps on biting us the way it is now.

Schedule a call for Monday? Please ping me on slack when it's good for you

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -682,7 +696,8 @@ def reduce_gradients(self, pipeline_parallel=False):
         if not self.overlap_comm:
             for i, group in enumerate(self.bit16_groups):
                 for param in group:
-                    if param.grad is not None:
+                    self.point_grad_reduc(param)
+                    if param.grad_reduc is not None:
```
tjruwase, line 700: I think this could cause an attribute error for non-trainable parameters, e.g., during finetuning. Is this change necessary?

jomayeri: The change exists because in Stage 1, Mode 1 `param.reduc_grad` points to `param.grad` and in Mode 2 it points to `param.accum_grad`. In the case of Mode 2 `param.grad` is always None. The `reduc_grad` is used as an alias to be apathetic to the Mode.  I'm not sure about the finetuning process but the array being looped over (`self.bit16_groups`) only contains trainable parameters as far as I'm aware based on the constructor. 
```Python
 298         for i, param_group in enumerate(self.optimizer.param_groups):
 299             partition_id = dist.get_rank(group=self.real_dp_process_group[i])
 300
 301             # push this group to list before modify
 302             # TODO: Explore simplification that avoids the extra book-keeping by pushing the reordered group
 303             trainable_parameters = []
 304             for param in param_group['params']:
 305                 if param.requires_grad:
 306                     param.grad_accum = None
 307                     param.grad_reduc = None
 308                     trainable_parameters.append(param)
 309             self.bit16_groups.append(trainable_parameters)
```

tjruwase: Good point, only trainable parameters will be accessed in this loop. Since this loop is only used for stage 1, it means `param.grad` or `param.grad_reduc` is should never be None. Perhaps this should be an assert?

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1561,14 +1627,23 @@ def zero_grad(self, set_to_none=False):
...
                 if set_to_none:
                     p.grad = None  # epilogue and in step
+                    p.grad_accum = None
+                    p.grad_reduc = None
                 else:
                     if p.grad is not None:
                         p.grad.detach_()
                         p.grad.zero_()
+                    if p.grad_accum is not None:
+                        p.grad_accum.detach_()
```
tjruwase, line 1642: I don't think this necessary since these are our buffers which should never be exposed externally (e.g., to autograd).

danyang-rainbow:
bros, can we just merge in master, and test by ourselves.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -833,6 +850,50 @@ def increment_value(dictionary, key):
...
+    # Clear the tensor the reduction gradient attribute is pointing to
+    def clear_grad_reduc_pointer(self, param):
+        param.grad_reduc = None
+        if self.use_grad_accum_for_reduction:
+            param.grad_accum = None
+        else:
+            param.grad = None
+
+    # Point to the gradient tensor to use in reduction
+    def set_grad_reduc_pointer(self, param):
```
tjruwase, line 882: I think we can avoid this setter by using a getter to access the gradient to reduce. Something like 

```python
  def get_gradient_for_reduction(self, param):
    return param.grad_accum if self.use_grad_accum_for_reduction else param.grad
```

Then all uses of `param.grad_reduc` can be replaced.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -276,6 +277,18 @@ def __init__(self,
...
         self.dtype = self.optimizer.param_groups[0]['params'][0].dtype
+        self.gradient_accumulation_dtype = gradient_accumulation_dtype
+
+        if self.dtype != self.gradient_accumulation_dtype:
+            self.use_separate_grad_accum = True
+        else:
+            self.use_separate_grad_accum = False
+        if self.use_separate_grad_accum and not self.partition_gradients:
+            self.use_grad_accum_for_reduction = True
+            self.reduction_dtype = self.gradient_accumulation_dtype
```
tjruwase, line 288: I think `self.reduction_dtype` might be redundant/incorrect.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -289,9 +302,12 @@ def __init__(self,
...
             # push this group to list before modify
             # TODO: Explore simplification that avoids the extra book-keeping by pushing the reordered group
-            trainable_parameters = [
-                param for param in param_group['params'] if param.requires_grad
-            ]
+            trainable_parameters = []
+            for param in param_group['params']:
+                if param.requires_grad:
+                    param.grad_accum = None
+                    param.grad_reduc = None
```
tjruwase, line 309: `param.grad_reduc` might be redundant.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -833,6 +850,50 @@ def increment_value(dictionary, key):
...
+
+    # Point to the gradient tensor to use in reduction
+    def set_grad_reduc_pointer(self, param):
+        if self.use_grad_accum_for_reduction:
+            param.grad_reduc = param.grad_accum
+        else:
+            param.grad_reduc = param.grad
+
+    # Called after a memory defrag moves tensor object via copy
+    def reset_pointers_to_reduc_after_copy(self, param):
```
tjruwase, line 889: I think this could be replaced by 

```python
   def update_gradient_for_accumulation(self, param, new_grad):
        if ....: 
           param.grad_accum.copy_(new_grad)`
        else:
           param.grad = new_grad ``

danyang-rainbow:
so, who can help to move this pr on.QAQ

KeremTurgutlu:
Thanks for this change, looks like tests are passing, so is it ready for use? Also, would this work for stage 0 (DDP) as well?

tjruwase (approved):

zaptrem:
@tjruwase It is my understanding that when training with BF16 enabled this PR means gradients are now accumulated in FP32, but master weights are still stored as BF16. Is this true? Has this been shown to not hurt model performance? If not, is there a way to store the master weights as FP32 as is done in DeepSpeed's FP16 mixed precision training?

tjruwase:
> @tjruwase It is my understanding that when training with BF16 enabled this PR means gradients are now accumulated in FP32, but master weights are still stored as BF16. Is this true? Has this been shown to not hurt model performance? If not, is there a way to store the master weights as FP32 as is done in DeepSpeed's FP16 mixed precision training?

This PR does not affect the master weight precision, fp32, in the optimizer. It seems provides an option to do gradient accumulation in fp32, even though fwd/bwd is in bf16 which creates bf16 gradients.

zaptrem:
> > @tjruwase It is my understanding that when training with BF16 enabled this PR means gradients are now accumulated in FP32, but master weights are still stored as BF16. Is this true? Has this been shown to not hurt model performance? If not, is there a way to store the master weights as FP32 as is done in DeepSpeed's FP16 mixed precision training?
> 
> This PR does not affect the master weight precision, fp32, in the optimizer. It seems provides an option to do gradient accumulation in fp32, even though fwd/bwd is in bf16 which creates bf16 gradients.

Thanks. To be clear, are you saying master weights are stored in BF16 or FP32 when BF16 is enabled? I tried to switch my FP16 run (which saves weights as FP32) to BF16 and got an error regarding missing BF16 weight files. When I manually initialized in BF16 by loading the params converted to FP32, the loss jumped a little and quickly started coming back down again.

Issue #2800, opened 2023-02-07
[BUG] `TrainSchedule` seems to use one more buffer than what's needed


jaywonchung:
I believe that `TrainSchedule`, which implements the synchronous 1F1B pipeline training schedule, uses one more buffer than what is actually needed in every pipeline stage except for the last.

https://github.com/microsoft/DeepSpeed/blob/d323abd80f62bebb9924bb85feb72b57c25af50d/deepspeed/runtime/pipe/schedule.py#L243-L247

1. Imagining a 4-stage pipeline with 8 microbatches, the first stage will do four forward computations and then one backward computation. So thinking in the high level, it should need only four buffers to remember the activations for the first four forward passes. However, `num_pipe_buffers` will return `5` for such a case (`self.stages == 4`, `self.stage_id == 0`, and `self.microbatches == 8`).
2. In any case, the largest value of `self.stage_id` is `self.stages - 1`. Thus, `buffers` is never smaller than 2. However, this function currently returns `max(2, buffers)`, which hints a one-off mistake. `max`ing with 2 itself makes sense because the last stage, which only has at most one in-flight microbatch and theoretically only needs one buffer, actually needs two buffers in order for `send` and `recv` to not deadlock or overwrite to each other.

Changing to the following, training should run well in theory.

```python
    def num_pipe_buffers(self):
        buffers = min(self.stages - self.stage_id, self.micro_batches)
        return max(2, buffers)
```

However, it dies with an assertion error triggered here

https://github.com/microsoft/DeepSpeed/blob/d323abd80f62bebb9924bb85feb72b57c25af50d/deepspeed/runtime/pipe/engine.py#L1032

which is part of `_exec_send_grads` and ensures that intermediate activations from the previous stage in the `'input'` buffer has the `.grad` attribute, which are the Jacovian-vector products produced by `_exec_backward_pass`.

Why does this happen? Inspecting the `PipeInstruction`s generated by the modified `TrainSchedule` class for the penultimate stage (`self.stage_id == 2` where `self.stages == 4` and `self.microbatches == 8`):

```python
>>> pprint(list(FixBufferTrainSchedule(8, 4, 2)), width=120)
[[-1],
 [-2],
 [0, RecvActivation(buffer_id=0), ForwardPass(buffer_id=0)],
 [-1, SendActivation(buffer_id=0)],
 [1, RecvActivation(buffer_id=1), ForwardPass(buffer_id=1)],
[...]

PR #2862, opened 2023-02-20
Fix buffer size for pipeline parallel and communication schedule

tohtana:
As pointed out by @jaywonchung in #2800, the scheduler for pipeline parallelism allocates more buffer than necessary.
This PR fixes the buffer size and also changes the schedule for communication to prevent the buffer from collapsing.

See #2800 for the discussion and validation.


deepspeed/runtime/pipe/schedule.py
```diff
@@ -243,7 +243,7 @@ def steps(self):
     def num_pipe_buffers(self):
         """As many buffers as the distance from this stage to the last stage.
         """
```
jaywonchung, line 245:
```suggestion
        """Return the number of pipeline buffers required for this stage.
        
        This is equivalent to the maximum number of in-flight forward passes,
        since we need to remember the activations of forward passes in order
        to run backpropagation. For synchronous 1F1B, this is equivalent to
        the index difference between this stage and the last stage.
        """
```
Just thought this was a good chance to improve explanation. Also now that we removed the `+1`, the word "distance" no longer makes sense, at least in the sense the original code writer intended.

ShadenSmith: Totally agreed, this is a great suggestion.

ShadenSmith (approved):
Thank you @tohtana and @jaywonchung !!!!!

PR #2863, opened 2023-02-20
Remove deprecated `torch._six` imports

### LINKED ISSUE #2845: [BUG] No module named 'torch._six'
Torch does not seem to support torch._six anymore and it has been removed.
Refer - https://github.com/pytorch/pytorch/pull/94709

DeepSpeed still has dependency on it. Example in `runtime/utils.py`

```
from torch._six import inf
```

Torch version to be precise - 2.0.0a0+git5a54537

yasyf:
Closes #2845.


yasyf:

mrwyattii (approved):

jeffra:
@yasyf [we currently recommend users use torch 1.8+](https://github.com/microsoft/deepspeed#requirements), it appears this older version of torch does not allow `from torch import inf`. Perhaps we can change these imports to be dependent on torch version? Not sure what version of torch this was supported, I just tried on 1.13 and it works there.

yasyf:
@jeffra godo catch! it's only a problem if you're using pytorch `2.0.0+` (nightly). I'll make it conditional

yasyf:
@jeffra test failure seems unrelated?

PR #2866, opened 2023-02-21
Data efficiency library update

conglongli:
1. Update blog/tutorial based on new results on BERT pretrain/GPT-2 finetune (corresponding update to examples: https://github.com/microsoft/DeepSpeedExamples/pull/238 and https://github.com/microsoft/Megatron-DeepSpeed/pull/113).
2. Fix the case when user provides a collate function.
3. Fix the case when user does not provide a mpu.
4. Fix checkpoint saving issues when applying the library for MoE model.
5. Use "/tmp" for unit test.


xiaoxiawu-microsoft (approved):
great jobs :)

PR #2888, opened 2023-02-23
fix iteration timing used in autotuning when gradient_accumulation_steps > 1

cli99:
the number of flops (flops profiler) is captured at global step boundary, thus when gradient_accumulation_steps > 1, the iteration time (from fwd, bwd and step timers capturing micro steps) used in the autotuner needs to be multiplied by the gradient_accumulation_steps.


deepspeed/runtime/engine.py
```diff
@@ -2252,11 +2253,13 @@ def _autotuning_exit(self):
...
             ],
                                        reset=False)
-            titer = msg[FORWARD_GLOBAL_TIMER] + msg[BACKWARD_GLOBAL_TIMER] + msg[
-                STEP_GLOBAL_TIMER]
+            titer = (msg[FORWARD_GLOBAL_TIMER] + msg[BACKWARD_GLOBAL_TIMER] +
+                     msg[STEP_GLOBAL_TIMER]) * self.gradient_accumulation_steps()
             msg["latency"] = titer
             msg["FLOPS_per_gpu"] = self.flops * 1_000_000 * self.gradient_accumulation_steps(
             ) / titer
+            msg["num_flops"] = self.flops
```
tjruwase, line 2261: Can "num_flops" and "fwd_duration" be symbolic constants?

21Ovi (commented):
will work

tjruwase (approved):

Issue #2793, opened 2023-02-04
[BUG] For pytorch transformers, do the linear layer hidden sizes need to be divisible by the world size?


jihan-yin:
**Describe the bug**
I'm reading through the deepspeed code, and for wrapping a model without kernel inject (in my case, T5), I see many instances in the code where the model's parameters are split into contiguous chunks according to the world size. However, in each case, it seems like if the split dimension is not divisible by the world size, we simply ignore the remainder parameters? [ex1](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/module_inject/replace_module.py#L413), [ex2](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/module_inject/replace_module.py#L466). 

If I'm working with T5 (qkv hidden dim = 4096), using deepspeed with 3 devices, it seems like there's a single weight for each layer that is not copied over (4096 mod 3 = 1). What happens to this weight? It seems to cause errors down the line when running inference, so is it a hard requirement that everything (linear dim, num heads, embedding size, hidden state dimension, etc) needs to be divisible by the world size (or `mp_size`)?

**To Reproduce**
1. Run the T5 code from https://www.deepspeed.ai/tutorials/inference-tutorial/  with `CUDA_VISIBLE_DEVICES=0,1,2` on a >=3 gpu machine

**Expected behavior**
Either for inference to work as intended, or for some assertion error. However, we get
```
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/ubuntu/anaconda3/envs/pytorch_p38/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/ubuntu/src/models/spellbook_celery/spellbook_celery/deepspeed/testing/run_sanity_test.py", line 82, in run_test_suite
    run_default_test()
  File "/home/ubuntu/src/models/spellbook_celery/spellbook_celery/deepspeed/testing/run_sanity_test.py", line 55, in run_default_test
    resp = get_response()
  File "/home/ubuntu/src/models/spellbook_celery/spellbook_celery/deepspeed/router.py", line 79, in get_response
    raise Exception(f"Deepspeed server encountered error: {data[ERROR_FIELD]}")
Exception: Deepspeed server encountered error: Traceback (most recent call last):
[...]

PR #2891, opened 2023-02-24
Assert mp_size is factor of model dimensions

molly-smith:
The number of GPUs or mp_size needs to be a factor of a model's hidden dimension, embedded dimension, number of attention heads, etc. Otherwise we encounter various tensor size errors as described in https://github.com/microsoft/DeepSpeed/issues/2793


deepspeed/module_inject/replace_module.py
```diff
@@ -464,20 +464,28 @@ def _slice_embedding(child, name, conv_linear_layer):
 
         def update_mp_params(child):
             if hasattr(child, 'n_heads'):
+                assert child.n_heads%mp_size == 0, "Model params must be divisible by mp_size"
```
tjruwase, line 467: The following changes would make the assert message more meaningful and actionable for users:
1. Include the actual dimension (e.g., hidden size) that violates the divisibility requirement. 
2. Include the invalid dimension value.
3. Report all the failing dimensions in one message (optional).

jeffra (approved):

PR #2896, opened 2023-02-24
Collect CUDA Graphs for Unique Input Signatures

DannyBruno:
Collect unique cuda graph based on input signatures. Solves bug where w/ cuda graph enabled increasing batch size will break. Added unit test.


deepspeed/inference/engine.py
```diff
@@ -480,25 +494,34 @@ def _create_cuda_graph(self, *inputs, **kwargs):
...
             if torch.is_tensor(inputs[i]):
-                self.static_inputs[i].copy_(inputs[i])
+                static_inputs[i].copy_(inputs[i])
         for k in kwargs:
             if torch.is_tensor(kwargs[k]):
-                self.static_kwargs[k].copy_(kwargs[k])
-        self._cuda_graphs.replay()
-        return self.static_output
+                static_kwargs[k].copy_(kwargs[k])
+        self._cuda_graphs[shape_signature].replay()
```
RezaYazdaniAminabadi, line 523: this way of accessing the Graphs can be costly for the performance. instead of accessing the graphs directly from a dictionary, let's use a list of graphs. We either need to have indirect access to the list by having the mapping to the right index of the list for each shape-signature, or get the list-index directly from the shape_signature.

RezaYazdaniAminabadi: the same issue applys for when accessing the static_inputs, static_kwargs, and static_output of that CUDA-Graph (which are lines 513, 514, and 515).

RezaYazdaniAminabadi (changes requested):

tests/unit/inference/test_inference.py
```diff
@@ -557,3 +557,53 @@ def test(self, model_family, model_name, task):
                        ds_output["results"][task]["ppl"])
         #assert ds_time <= bs_time
         assert ppl_diff < 0.01
+
+
+@pytest.mark.inference
+@pytest.mark.parametrize("model_w_task",
+                         [("bigscience/bloom-560m",
```
cmikeh2, line 564: We don't currently have CG support for generative models, can we switch this to a BERT variant?

tests/unit/inference/test_inference.py
```diff
@@ -557,3 +557,53 @@ def test(self, model_family, model_name, task):
...
+                                              replace_with_kernel_inject=True,
+                                              replace_method='auto',
+                                              enable_cuda_graph=True)
+
+        # Run pipe test with increasing and then decreasing batch size
+        success = True
+        try:
+            pipe(query, batch_size=2)
+            pipe([query] * 4, batch_size=4)
+            pipe(query, batch_size=1)
```
cmikeh2, line 603: Let's use a 1-4-1 batch pattern here to make sure the b=1 graph is still valid after we've run/created the b=4 graph.

tests/unit/inference/test_inference.py
mrwyattii: yup, I noticed this in the inference test errors. I'll make that change.

deepspeed/inference/engine.py
```diff
@@ -583,6 +605,14 @@ def _local_cuda_graph_used(self, module):
 
             return sub_module_cuda_graph
 
+    def _check_for_workspace_invalidation(self, input_signature):
+        batch_size = input_signature[1][0][0]
```
DannyBruno, line 609: Not confident that this will always be the correct way to retrieve batch size. Is there a better way?

mrwyattii:
Closing this in a favor of a future different approach @DannyBruno and I have discussed.

PR #2905, opened 2023-02-26
deepspeed.init_distributed() support for TCP protocols

noabauma:
This little fix in the code will let users (e.g. on CSCS Piz Daint) use the Slurm workload manager combined with the init_method for TCP protocols. For init_method with TCPs, the rank and world_size have to be communicated.

### DIFF
--- a/deepspeed/comm/comm.py
+++ b/deepspeed/comm/comm.py
@@ -604,8 +604,8 @@ def init_distributed(dist_backend=None,
                      init_method=None,
                      dist_init_required=None,
                      config=None,
-                     rank=None,
-                     world_size=None):
+                     rank=-1,
+                     world_size=-1):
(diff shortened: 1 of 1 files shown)


jeffra:
Thank you @noabauma for your contribution! This change looks good to me but is missing one important piece. You'll need to pass `rank` and `world_size` to the torch init_process call itself: https://github.com/microsoft/DeepSpeed/blob/87eaf8f99abdebac1207565e3c3ee49cf151f29b/deepspeed/comm/torch.py#L30-L36

noabauma:

Dear @jeffra,

I totally forgot to modify TorchBackend as well, thanks for noting that. I hope this little code documentation is also fine. If not, don't hesitate to contact me again. Thank you for letting me do this little contribution to DeepSpeed.

jeffra (approved):
LGTM, thanks @noabauma!

PR #2907, opened 2023-02-27
better eval sampler for val or test dataset

mayank31398:
* uses DistributedSampler instead of SequentialSampler for validation sets to avoid duplicating the datasets on the GPUs


conglongli (approved):
LGTM

mayank31398:
@conglongli looks like the nv-transformers-v100 tests are failing. Not sure if this is related to the PR.
Can you comment on that?

tjruwase:
> @conglongli looks like the nv-transformers-v100 tests are failing. Not sure if this is related to the PR. Can you comment on that?

Seems to be related to #2909. Should be fixed soon.

PR #2916, opened 2023-03-01
Codegen Inference Support

RezaYazdaniAminabadi:
This PR adds the inference-kernel support for the CodeGen models, such as Salesforce/codegen-xx-multi


deepspeed/module_inject/containers/gptj.py
```diff
@@ -77,21 +77,33 @@ def __init__(self, client_module, inference=True):
...
+            HFGPTJLayerPolicy._orig_layer_class = [
+                transformers.models.gptj.modeling_gptj.GPTJBlock,
+                transformers.models.codegen.modeling_codegen.CodeGenBlock
+            ]
         except:
             HFGPTJLayerPolicy._orig_layer_class = None
 
     def get_hidden_heads(self):
-        return self.client_module.attn.q_proj.weight.shape[1], \
+        return self.client_module.attn.out_proj.weight.shape[1], \
```
molly-smith, line 88: Does this not break gpt-j? Gptj and codegen have different naming schemes for qkv. When I was briefly looking into adding codegen support I created a separate container file like this https://github.com/microsoft/DeepSpeed/compare/master...mosm/codegen

RezaYazdaniAminabadi: but I am using out_proj here, and gptj has the same naming convention: https://github.com/huggingface/transformers/blob/main/src/transformers/models/gptj/modeling_gptj.py#L115

ganler:
@RezaYazdaniAminabadi Thanks for the great work! I am trying to use your patch to run codegen-16B-mono but unfortunately failed with "FileNotFoundError: [Errno 2] No such file or directory: '/home/jiawei/.conda/envs/g/lib/python3.8/site-packages/deepspeed/ops/csrc/transformer/inference/csrc/pt_binding.cpp'". The command I was using is `deepspeed --num_gpus 1 --num_nodes 1 inference-test.py --name Salesforce/codegen-16B-mono --ds_inference --use_kernel --test_performance`. Could you kindly advise me how to make it work? Thanks!

PR #2939, opened 2023-03-03
AutoTP Assert Kernel Injection Support

molly-smith:
If AutoTP fails, check if model is supported by kernel injection and provide suggestion to use kernel injection instead. Clarify to users that some models that are not supported by autoTP can still be used with kernel injection.


lekurile (approved):

awan-10 (approved):
@molly-smith -- this looks good to me. Please add some context/description to the PR so we know why we are doing it.

PR #2940, opened 2023-03-03
Update Inference Engine checkpoint loading + meta tensor assertions

lekurile:
This PR adds a `model_device_meta` attribute to the `InferenceEngine` that's used to:
* assert that `replace_with_kernel_inject == True` if meta tensors are used, since we only support meta tensors when kernel injection is enabled.
* Allow the `InferenceEngine` to load checkpoints via `_load_checkpoint()` when a checkpoint is passed to the `init_inference` API only when meta tensors are not used.

This PR also adds an assertion in the `initialize_tensors` function of the base container to check that if the model is using meta tensors, that the corresponding model container uses the meta tensor feature.


deepspeed/inference/engine.py
```diff
@@ -151,6 +151,13 @@ def __init__(self, model, config):
             assert pkg_version.parse(torch.__version__) >= pkg_version.parse("1.10"), \
                 "If you want to use cuda graph, please upgrade torch to at least v1.10"
 
+        # Check if model passed to engine is loaded w/ meta tensors, in which case
+        # kernel injection must be enabled.
```
awan-10, line 155: Please update the comment to add why this only works for HF models.

awan-10 (approved):

deepspeed/inference/engine.py
lekurile: Added a note saying that the device type is sourced assuming a Hugging Face hierarchy.

Issue #2717, opened 2023-01-18
[BUG] Make InferenceModule enable_cuda_graph more flexible.


tchaton:
**Describe the bug**

The following with `enable_cuda_graph=True` breaks as the entire model can't be traced under a cuda graph.

```python
hf_auth_key = os.getenv("HF_AUTH_KEY")
if not hf_auth_key:
    raise ValueError("HF_AUTH_KEY is not set")

pipe = diffusers.StableDiffusionPipeline.from_pretrained(
    "runwayml/stable-diffusion-v1-5",
    use_auth_token=hf_auth_key,
    torch_dtype=torch.float16,
    revision="fp16")

print(pipe)
pipe = deepspeed.init_inference(pipe.to("cuda"), dtype=torch.float16, enable_cuda_graph=True)
```

However, by introducing a global cuda graph and local ones through the policies like DSUnet, etc.. works. This enables to gain an extra 100-150 ms.

```python
class InferenceEngine(InferenceEngine):

    def __init__(self, *args, enable_cuda_graph_global: bool = False, **kwargs):
        super().__init__(*args, **kwargs)
        self.enable_cuda_graph_global = enable_cuda_graph_global

    def forward(self, *inputs, **kwargs):
        """Execute forward propagation
        Arguments:
            *inputs: Variable length input list
            **kwargs: variable length keyword arguments
        """
        start = None
        if self.model_profile_enabled and self.enable_cuda_graph_global:
            torch.cuda.synchronize()
            start = time.time()

        if self.enable_cuda_graph_global:
            if self.cuda_graph_created:
                outputs = self._graph_replay(*inputs, **kwargs)
            else:
                self._create_cuda_graph(*inputs, **kwargs)
                outputs = self._graph_replay(*inputs, **kwargs)
        else:
            outputs = self.module(*inputs, **kwargs)

        if self.model_profile_enabled and self.enable_cuda_graph_global:
            torch.cuda.synchronize()
            duration = time.time() - start
            self._model_times.append(duration)

        return outputs
```

You can found the code there: https://github.com/Lightning-AI/stablediffusion/blob/lit/ldm/deepspeed_replace.py#L34

**To Reproduce**
1. Simple inference script to reproduce
2. What packages are required and their versions
3. How to run the script
4. ...

**Expected behavior**

**ds_report output**
Please run `ds_report` to give us details about your setup.

**Screenshots**

**System info (please complete the following information):**
[...]

PR #2941, opened 2023-03-04
Check for local CUDA graphs when enable_cuda_graph=True

lekurile:
This PR addresses issue #2717.

Currently, if `enable_cuda_graph=True`, the `InferenceEngine` will attempt to create a CUDA graph on the entire `forward()` function. However, if DeepSpeed implementations contain _local_ CUDA graphs, capturing the entire `forward()` function is not possible and instead the local CUDA graphs should take precedence.

This PR updates the way the `InferenceEngine` builds its CUDA graphs by checking to see if any of the submodules contain CUDA graph implementations during the engine `__init__` phase and storing the result in `self.local_cuda_graph`. If local CUDA graph implementations are detected and `enable_cuda_graph=True`, the `InferenceEngine` forward function will not attempt to capture a CUDA graph and instead will simply call `self.module()`, allowing local CUDA graph implementations be created and used.

A new ABC `CUDAGraph` "feature" class is added to `deepspeed/model_implementations/features` that's inherited by DSVAE, DSUNet, and DSClipEncoder implemenations. This ensures that implementations using the `CUDAGraph` class will have the same naming for `self.enable_cuda_graph`, which is searched for during `InferenceEngine` initialization to detect if local CUDA graphs are used.

After these changes, the `"runwayml/stable-diffusion-v1-5"` model was tested with `enable_cuda_graph=True` across 10 prompts and showed a ~4% performance improvement.

Thanks to @cmikeh2 for the support on this.

### DIFF
--- a/deepspeed/inference/engine.py
+++ b/deepspeed/inference/engine.py
@@ -532,8 +532,6 @@ class InferenceEngine(Module):
             for name in module.__dict__.keys():
                 sub_module = getattr(module, name)
 
-                print(f"DEBUG: sub_module_cuda_graph = {sub_module_cuda_graph}")
-                print(f"DEBUG: sub_module = {sub_module}")
                 if self._module_match(sub_module) and hasattr(sub_module,
                                                               "enable_cuda_graph"):
                     sub_module_cuda_graph = True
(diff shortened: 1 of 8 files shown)


jeffra (approved):

cmikeh2 (approved):
LGTM!

Issue #2904, opened 2023-02-26
[BUG]


Rishabh-Choudhry:
**Describe the bug**

When using deepspeed accelerator with pytorch lightning, I get this error. The issue seems to be the wrong path calling in the deepspeed library for **UtilsBuilder** class.
> File "deepspeed/runtime/engine.py", line 392, in __init__
>     util_ops = UtilsBuilder().load()
> TypeError: 'NoneType' object is not callable

The issue gets resolved after updating the path from `from deepspeed.ops.op_builder import UtilsBuilder` to `from deepspeed.ops.op_builder.utils import UtilsBuilder`.

**To Reproduce**
Use `strategy=DeepSpeedStrategy(stage=2)` with torch lightning trainer.

**Expected behavior**
The deepspeed accelerator should run as normal with pytorch lightning.

**System info (please complete the following information):**
 - OS: Ubuntu
 - GPU count and types : 1 x NVIDIA T4
 - Python version : 3.8.12

tjruwase:
@Rishabh-Choudhry, thanks for reporting this issue. Do you mean submitting a PR?

jeffra:
@Rishabh-Choudhry and @yasyf, I can't seem to reproduce this issue. Can you share what version of deepspeed you are using?  I tried latest `master` and also v0.8.1.

I am trying to reproduce via:

```python
from deepspeed.ops.op_builder import UtilsBuilder
util_ops = UtilsBuilder().load()
```

I am glad to hear the suggested fix from @Rishabh-Choudhry works, but i worry other ops might also have the same problem but not being triggered in your setting. Maybe I need to try w. pytorch lightning directly though.

yasyf:
@jeffra I am using `master` (with pytorch nightly)

jeffra:
@yasyf, does the above snippet reproduce the stack trace or is it only reproducible from lightning?

jeffra:
Hi @yasyf, I've pushed a PR (https://github.com/microsoft/DeepSpeed/pull/2963) can you give the related branch a try on your side to confirm if it fixes your issue or not?

yasyf:
@jeffra fixed! totally didn't realize I still had colossal installed :)

jeffra:
Excellent! Glad to hear :) @Rishabh-Choudhry plesse reopen if this is not fixed for you

SamieeS:
Hi @jeffra ,

[...]

Issue #2690, opened 2023-01-11
[BUG] the `benchmarks` folder is included upon installation


hadim:
I noticed that while inspecting the conda package during my attempt to create a conda forge build.

![image](https://user-images.githubusercontent.com/528003/211929466-436decd1-a733-4db6-8a67-c948736575a5.png)

The fix is likely as simple as adding `benchmarks` to `packages=find_packages(exclude=[....])` in the `setup.py` file.

jeffra:
Thank you @hadim, this is a good catch. I'll create a PR to fix this shortly.

hadim:
`benchmarks` is still included in 0.8.0 but only the subfolder `benchmarks/communications/` and its content.

Maybe in `packages=find_packages(exclude=` you need to specify `benchmarks.communications` or `benchmarks.*`?

You can double-check this by building a wheel locally and opening it to check which folders have been included.

![image](https://user-images.githubusercontent.com/528003/213032333-bf190571-5b21-4d45-bef1-4c3b10370bf2.png)

---

It's not a critical issue but it's best to remove it to prevent any conflicts with other libraries.

jeffra:
Thank you @hadim for bringing this up. I think now that i am looking at all of this closer we might just want to change this to be something like `packages=find_packages(include=['deepspeed', 'deepspeed.*']),`. But I am doing some local testing first to make sure this works as expected.

jeffra:
Can you confirm this fixes the issue from your perspective?

https://github.com/microsoft/DeepSpeed/pull/2714

hadim:
It seems ok to me but I haven't setup deepspeed locally to confirm.

If you build a wheel locally with `pip wheel` you can then open the wheel and check what has been included. Also any CI installing deepspeed in non editable mode will confirm all the required deepspeed modules have been included correctly.

jeffra:
I installed locally with the wheel we upload to pypi. We do build via sdist, which is a bit different: https://github.com/microsoft/DeepSpeed/blob/3f210c971590d5d23acc57004330f9ae38001748/release/release.sh#L34

Our CI will install and run all the tests on my PR to check that everything is all still good. My local tests before pushing the PR all looked good though. I installed in a fresh virtualenv and inspected the lib/python3.8/site-packages for that virtualenv to see what was installed. I repro'd your issue with the current master where you see that benchmarks/communication still exists.

hadim:
[...]

PR #2963, opened 2023-03-07
Fix issue between our abstract accelerator and colossalai's version of op_builder

jeffra:
ColossalAI has an [adapted version of DeepSpeed's op_builder](https://github.com/hpcaitech/ColossalAI/tree/main/op_builder), however it appears they are installing it as a top-level package called `op_builder` as it's [not excluded in their setup.py](https://github.com/hpcaitech/ColossalAI/blob/ea0b52c12ee7598fe126ed0d8b0557f7e8a0e999/setup.py#L154-L164). We have also had similar issues accidentally installing extra packages like this in the past (e.g., https://github.com/microsoft/DeepSpeed/issues/2690). They might be able to just add this to their package exclude list but not sure.

The issue (https://github.com/microsoft/DeepSpeed/issues/2904) occurs when a user has both ColossalAI and DeepSpeed >= 0.8.0 installed. DeepSpeed v0.8.0 introduced the concept of an abstract accelerator. In order to support pre-compilation of ops we attempt to import `op_builder` and we assumed this would import from the local path, however if another package installs `op_bulder` as a package python will pick that up instead. This was made a bit more confusing to debug since ColossalAI's op_builder reuses the same builder names as ours (e.g., `CPUAdamBuilder`).

The fix in this PR forces the `op_builder` import to be relative to the local path in order to detect if we are in a pre-compilation mode or JIT compilation mode.

Tagging core contributors on DeepSpeed's abstract accelerator for awareness/review: @delock, @tjruwase

### DIFF
--- a/accelerator/cuda_accelerator.py
+++ b/accelerator/cuda_accelerator.py
@@ -224,8 +224,9 @@ class CUDA_Accelerator(DeepSpeedAccelerator):
 
     def op_builder_dir(self):
         try:
-            # during installation time op_builder is visible as a local path, otherwise return deepspeed.ops.op_builder
-            from .op_builder import builder  # noqa: F401
+            # is op_builder from deepspeed or a 3p version? this should only succeed if it's deepspeed
+            # if successful this also means we're doing a local install and not JIT compile path
+            from op_builder import __deepspeed__  # noqa: F401
             return "op_builder"
         except ImportError:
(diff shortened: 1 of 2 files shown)


mrwyattii (approved):

delock:
Thanks!  The __deepspeed__ technique is very enlightening.

PR #2972, opened 2023-03-08
SCR checkpoint engine

adammoody:
The existing checkpoint engine model is a close fit for plugging in the Scalable Checkpoint / Restart library, which like Nebula, caches checkpoint files in node-local storage.  There are a few extensions that could be added that would help.  I wanted to open this PR as a means for discussion.

Could we move directory creation in the checkpoint engine, e.g., maybe add a new ``checkpoint_engine.makedirs()`` that could replace ``os.makedirs()`` calls like this one:

https://github.com/microsoft/DeepSpeed/blob/58a4a4d4c19bda86d489ac171fa10f3ddb27c9d6/deepspeed/runtime/pipe/module.py#L590

In some situations, it's useful to delay directory creation until the files are copied.  In particular, some checkpoint libraries do not copy every checkpoint, in which case, it's good to avoid creating directories at all.

Just as the checkpoint sequence has "start" and "end" operations in the form of ``create()`` and ``commit()``, it would be useful to have start and end operations for the restart sequence.  Maybe new functions like ``checkpoint_engine.open()`` and ``checkpoint_engine.close()``?  One would then call ``checkpoint_engine.load()`` in between ``open()`` and ``close()``.  This makes it easier for the library to allow the application to read from temporary paths.

For the existing engines, these calls would just be NOP's.  In the case of SCR, one must read checkpoint files between calls to ``scr.start_restart()`` and ``scr.complete_restart()``.


loadams:
Hi @adammoody - apologies for the very slow reply on this.  Is this a PR that you would still like to see merged?  If so, would you be able to - when you have time - merge the latest master and then we can review it?

Let us know, if not we can close this, but we will get the right people to review it after it is updated.  Thanks!

adammoody:
Thanks, @loadams .  Actually, before getting into details on this PR, there is a related PR to add "open/close" operations to the checkpoint engine:

https://github.com/microsoft/DeepSpeed/pull/3009

Could we start with that one?

It also needs to be refreshed, but it would be help if folks could consider the ideas there and let me know whether the approach sounds reasonable.

PR #2981, opened 2023-03-09
pre-commit check for torch.cuda in code

delock:
This PR add checks for CUDA specific code in DeepSpeed.  The purpose is to avoid accidental usage of CUDA code in new code.

Two kinds of code are checked against:
1. `torch.cuda` code which calls PyTorch cuda runtime.
2. `tensor.cuda()` which converts tensor to cuda device.

Also port remaining cuda specific code which can be ported.


deepspeed/runtime/activation_checkpointing/checkpointing.py
```diff
@@ -499,7 +499,7 @@ def get_cpu_activations_for_backward(args, inputs):
 class CheckpointFunction(torch.autograd.Function):
     """This function is adapted from torch.utils.checkpoint with
        two main changes:
-           1) torch.cuda.set_rng_state is replaced with `_set_cuda_rng_state`
+           1) torch .cuda.set_rng_state is replaced with `_set_cuda_rng_state`
```
delock, line 502: This change is to avoid trigger torchcuda check.  According to the comment this cannot be replaced by get_accelerator().  So add a space between `torch` and `.cuda` to get around.

jeffra: I wonder if we could easily add a `#ignore-cuda` comment to the line (or similar) that we could filter during the check?

delock: Sounds good, I'll make the change.

jeffra (approved):
this is great, thank you @delock

PR #3022, opened 2023-03-15
BF Optimizer Attribute Checks

jomayeri:
Adding attribute checks for when the BF Optimizer and ZeRO are both enabled.


tjruwase (approved):

tjruwase:
@jomayeri, is there an issue to link this to? Thanks!

Issue #1896, opened 2022-04-19
[BUG] DeepSpeed zero_to_fp32.py script ignores some layers while creating FP32 checkpoints from DS ZeRO checkpoints.


rohitdwivedula:
**Problem**: Trying to convert DeepSpeed zero checkpoints to `PyTorch state_dict`s leads to one layer not being present in the generated state dict. I am using the `zero_to_fp32.py` script. I'm trying to train a GPT2 like model, and it looks like the `lm_head` (linear layer) of the model is not being correctly included in checkpoints for some reason.
 
# Description 
I'm trying to train a GPT2-like model using DeepSpeed (and some code from [huggingface/transformers](https://github.com/huggingface/transformers)). The model looks something like this:

```
CustomGPT2LMHeadModel(
  (transformer): GPT2Model(
    (wte): Embedding(50257, 1024)
    (wpe): Embedding(1024, 1024)
    (drop): Dropout(p=0.1, inplace=False)
    (h): ModuleList(
      (0): GPT2Block(
        (ln_1): LayerNorm((1024,), eps=1e-05, elementwise_affine=True)
        (attn): GPT2Attention(
          (c_attn): Conv1D()
          (c_proj): Conv1D()
          (attn_dropout): Dropout(p=0.1, inplace=False)
          (resid_dropout): Dropout(p=0.1, inplace=False)
        )
        (ln_2): LayerNorm((1024,), eps=1e-05, elementwise_affine=True)
        (mlp): GPT2MLP(
          (c_fc): Conv1D()
          (c_proj): Conv1D()
          (dropout): Dropout(p=0.1, inplace=False)
        )
      )
      (1): GPT2Block(
        (ln_1): LayerNorm((1024,), eps=1e-05, elementwise_affine=True)
        (attn): GPT2Attention(
          (c_attn): Conv1D()
          (c_proj): Conv1D()
          (attn_dropout): Dropout(p=0.1, inplace=False)
          (resid_dropout): Dropout(p=0.1, inplace=False)
        )
        (ln_2): LayerNorm((1024,), eps=1e-05, elementwise_affine=True)
        (mlp): GPT2MLP(
          (c_fc): Conv1D()
          (c_proj): Conv1D()
          (dropout): Dropout(p=0.1, inplace=False)
        )
      )
    )
    (ln_f): LayerNorm((1024,), eps=1e-05, elementwise_affine=True)
  )
[...]

PR #3033, opened 2023-03-16
Recover shared parameters

ShijieZZZZ:
To address this issue:

[[BUG] DeepSpeed zero_to_fp32.py script ignores some layers while creating FP32 checkpoints from DS ZeRO checkpoints. ](https://github.com/microsoft/DeepSpeed/issues/1896)

Shared parameters that hold reference are missing when extracts fp32 weights from checkpoint. This PR recovers such shared parameters by linking them with their partners.


tjruwase (approved):

mayank31398:
Thanks a lot.
Ill try to give this a shot sometime today :)

tjruwase:
@stas00 FYI

mayank31398:
@ShijieZZZZ https://github.com/microsoft/DeepSpeed/pull/3295

stas00:
This code unfortunately doesn't work, since `data_ptr()` always points to `0` in this code, as all compared tensors are of 0 size. And so it re-ties the shared param to whatever first param it encounters. So the recovered checkpoint has invalid shared params.

I think it only happens to work if the first param is the that's shared.

stas00:
in `_zero3_consolidated_16bit_state_dict` we use `ds_id` to identify shared params:

https://github.com/microsoft/DeepSpeed/blob/a094c9763de8d42107cbffd0bb9abb8056aa3c60/deepspeed/runtime/engine.py#L3225-L3228

PR #3036, opened 2023-03-16
fix check params

satpalsr:
Check params was removed for [Bloom](https://github.com/microsoft/DeepSpeed/pull/2083).
Not evident why.


jeffra (approved):
Good catch, this was accidentally added. Thank you @satpalsr

jeffra:
Removing this `return` seems to cause failures on some of our unit tests, need to investigate further before merging.

lekurile:
@satpalsr Looks like this check was removed in GH-2083 and was likely done due to certain models having mixed dtypes for various parameters. I think this likely warrants a deeper discussion about what purpose `__check_params()` serves in the runtime engine and if we should look to remove it altogether.

In the meantime, I think we can close this PR since simply removing the `return` in the function will result in errors in the unit tests. E.g. `TestSaveTensorClone.test_save_tensor_clone[False-2]`:
```
ValueError: torch.float32 is enabled but the following parameters have dtype that is not torch.float32: [('linears.0.weight', torch.float16), ('linears.0.bias', torch.float16), ('linears.1.weight', torch.float16), ('linears.1.bias', torch.float16), ('linears.2.weight', torch.float16), ('linears.2.bias', torch.float16), ('linears.3.weight', torch.float16), ('linears.3.bias', torch.float16)]
```

PR #3041, opened 2023-03-17
[CPU] Support Intel CPU inference

delock:
### Summary:
This PR provides adds Intel CPU support to DeepSpeed by extending [DeepSpeedAccelerator](https://github.com/microsoft/DeepSpeed/blob/master/accelerator/abstract_accelerator.py) Interface.   It allows user to run LLM inference with Intel CPU with Auto Tensor Parallelism or kernel injection (currently kernel injection is experimental and only support BLOOM).   This PR comes with the following features:
1. CPUAccelerator
2. Low latency communication backend CCLBackend that call oneCCL collective communication library directly instead of calling through torch.distributed.
3. Kernel injection fallback path to aid development of DeepSpeed accelerator.   Allowing kernel injection path to function before kernels fully developed.
4. **(Merged)** ~~Reduce peak memory usage during AutoTP model loading stage. (Also in seperte PR https://github.com/microsoft/DeepSpeed/pull/3102)~~
5. **(Merged)** ~~BF16 inference datatype support.~~
6. **(Merged)** ~~LLAMA auto tensor parallel.(Also in PR https://github.com/microsoft/DeepSpeed/pull/3170)~~
7. **(Merged)** ~~Efficient distributed inference on host with multiple CPU NUMA cluster, including multiple CPU socket or sub NUMA cluster (SNC).  (Also in seperate PR https://github.com/microsoft/DeepSpeed/pull/2881)~~
8. **(Merged)** ~~BLOOM AutoTP and/or tensor parallelism through user specified policy. (Also in PR https://github.com/microsoft/DeepSpeed/pull/3035)~~

Note: this PR keeps all the pieces for testing.  The other PRs mentioned here are supposed to be reviewed seperatedly to allow small steps of work.

### Major components:
1. core binding support -- changes under deepspeed/launcher
2. CCLBackend -- changes under deepspeed/com
3. CPUAccelerator -- files from the following places:
- accelerator/cpu_accelerator.py
- op_builder/cpu/
- csrc/cpu/
4. Accelerator selection logic -- accelerator/real_accelerator.py
5. Kernel injection fallback path -- changes from deepspeed/ops/transformer/inference/op_binding/
6. bfloat16 inference support -- changes scattered around

### CPU Inference example
This section is work in progress


deepspeed/module_inject/replace_module.py
```diff
@@ -257,8 +258,8 @@ def replace_attn_block(child, policy):
     if isinstance(module, torch.nn.Module):
         pass
     else:
-        if fp16 is False:
-            raise ValueError("Generic injection only supported with FP16")
+        if fp16 is False and bf16 is False:
+            raise ValueError("Generic injection only supported with FP16 of BF16")
```
rogerxfeng8, line 262: typo: or BF16

delock:
@tjruwase we plan to add a new workflow to validate installation for CPU and run some inference related test.   If there is anything we should know before add new workflow just let us know.  Thanks!

delock:
> @tjruwase we plan to add a new workflow to validate installation for CPU and run some inference related test. If there is anything we should know before add new workflow just let us know. Thanks!

The CPU workflow requires machine with AVX2 or above.  Is there any self-hosted keywords I can use to get machine with AVX2 or above?

tjruwase:
> > @tjruwase we plan to add a new workflow to validate installation for CPU and run some inference related test. If there is anything we should know before add new workflow just let us know. Thanks!
> 
> The CPU workflow requires machine with AVX2 or above. Is there any self-hosted keywords I can use to get machine with AVX2 or above?

We are discussing this internally, and will respond asap. In the meantime, you can try https://github.com/microsoft/DeepSpeed/blob/master/.github/workflows/nv-torch-latest-cpu.yml

jeffra:
> @tjruwase we plan to add a new workflow to validate installation for CPU and run some inference related test. If there is anything we should know before add new workflow just let us know. Thanks!

Thanks @delock, this sounds great. Will github provided runner(s) work for this? If so, please extend the existing `nv-torch-latest-cpu.yml` workflow that @tjruwase referenced above.

delock:
> > > @tjruwase we plan to add a new workflow to validate installation for CPU and run some inference related test. If there is anything we should know before add new workflow just let us know. Thanks!
> > 
> > 
> > The CPU workflow requires machine with AVX2 or above. Is there any self-hosted keywords I can use to get machine with AVX2 or above?
> 
> We are discussing this internally, and will respond asap. In the meantime, you can try https://github.com/microsoft/DeepSpeed/blob/master/.github/workflows/nv-torch-latest-cpu.yml

Thanks @tjruwase  @jeffra !  We did some investigation and found at least one instance used to run workflow comes with AVX2 instruction set but Intel Extension for Pytorch complains no AVX2 support.
https://github.com/microsoft/DeepSpeed/actions/runs/4498976192/jobs/7916249285?pr=3041#step:6:1115

I talked with engineer who worked on PyTorch, he has some idea on that behavior and would investigate early next week.   Once we have conclusion will give you an update.

delock:
> > > @tjruwase we plan to add a new workflow to validate installation for CPU and run some inference related test. If there is anything we should know before add new workflow just let us know. Thanks!
> > 
> > 
> > The CPU workflow requires machine with AVX2 or above. Is there any self-hosted keywords I can use to get machine with AVX2 or above?
> 
> We are discussing this internally, and will respond asap. In the meantime, you can try https://github.com/microsoft/DeepSpeed/blob/master/.github/workflows/nv-torch-latest-cpu.yml

Hi @tjruwase , we have investigated this issue and found that the root cause is some old hypervisor does not emulate xcr register used by Intel Extension for Pytorch，which detects  XSAVE status for AVX2.  We plan to remove this detection in next release.  Before that, we will use a temporary Intel Extension for PyTorch wheel to build cpu workflow.

So to summary, the issue is not because machine does not support AVX2.   I think we can build the workflow using the default ubuntu-20.04 instance for github and see if we can always get instance with AVX2 or above.

csrc/cpu/includes/comm.h
```diff
@@ -0,0 +1,15 @@
+// Copyright The Microsoft DeepSpeed Team
```
delock, line 1: It looks like we can remove this file

.github/workflows/cpu-inference.yml
```diff
@@ -0,0 +1,68 @@
+name: cpu-inference
```
delock, line 1: CPU workflow, still WIP

accelerator/cuda_accelerator.py
```diff
@@ -26,7 +26,7 @@ def __init__(self):
 
         for _, module_name, _ in pkgutil.iter_modules([os.path.dirname(op_builder_module.__file__)]):
             # avoid self references
-            if module_name != 'all_ops' and module_name != 'builder':
+            if module_name != 'all_ops' and module_name != 'builder' and module_name != 'cpu':
```
delock, line 29: Currently CPU op_builders is under directory `op_builder/cpu`.  iter_modules for CUDA op_builder will see `cpu/` directory.  Filter it out to avoid enumerate CPU op builders.

deepspeed/comm/ccl.py
```diff
@@ -0,0 +1,48 @@
...
+def build_ccl_op():
+    builder = get_accelerator().create_op_builder("CCLCommBuilder")
+    ccl_cpp_module = builder.load()
+    print(f'DeepSpeed {builder.absolute_name()} built successfully')
+    return ccl_cpp_module
+
+
+class CCLBackend(TorchBackend):
+
+    def __init__(self, name='ccl', rank=-1, world_size=-1, mpu=None, timeout=None, init_method=None):
```
delock, line 20: Currently we only support broadcast, barrier and allreduce.   Will support other collectives when CPU training is being worked on.

csrc/cpu/comm/ccl.cpp
```diff
@@ -0,0 +1,345 @@
+#include <torch/extension.h>
```
delock, line 1: Need to do a round of clean up to keep only active codes and remove the rest.

deepspeed/inference/engine.py
```diff
@@ -34,6 +34,58 @@
 INFERENCE_MODEL_TIMER = "model-forward-inference"
```
delock, line 36: Changes in this file related to AutoTP for BLOOM, also appear in PR#3035

deepspeed/launcher/constants.py
```diff
@@ -8,6 +8,7 @@
 
 OPENMPI_LAUNCHER = 'openmpi'
```
delock, line 9: On CPU, we support MPICH and Intel MPI (IMPI) launcher.

deepspeed/module_inject/auto_tp.py
```diff
@@ -104,6 +104,10 @@ def tp_parser(model):
                         gem_list = gem_list + [layer_list[i - 1]]
                 elif 'out_proj' in layer:
                     gem_list = gem_list + [layer]
+                elif 'o_proj' in layer:
```
delock, line 107: support LLAMA AutoTP, will also have seperate PR.

deepspeed/module_inject/replace_module.py
```diff
@@ -24,6 +24,8 @@
 
 from .utils import policy_to_ds_container
 
+import gc
```
delock, line 27: Changes in this file are mainly tracked by #3102 which is to reduce memory consumption in AutoTP

deepspeed/ops/transformer/inference/op_binding/mlp_gemm.py
```diff
@@ -12,16 +13,33 @@ class MLPGemmOp(BaseOp):
...
-                                                  self.config.q_int8, self.config.mlp_act_func_type)
+        if self.mlp_gemm_func != None:
+            output, residual_add = self.mlp_gemm_func(input, residual, input_bias, weight_interm, weight_out, bias,
+                                                      gamma, beta, self.config.epsilon, self.config.pre_layer_norm,
+                                                      self.config.mlp_after_attn, weight_interm.scale,
+                                                      weight_out.scale, self.config.q_int8,
+                                                      self.config.mlp_act_func_type)
+        else:
+            # fallback
+            if self.config.mlp_after_attn:
```
delock, line 37: Fallback path is introduced to faciliate writing kernels for new backend.  Allowing backend to integrate kernel injection in incremental manner.  Currently we have only built fallback path for BLOOM model yet.

deepspeed/ops/transformer/inference/op_binding/mlp_gemm.py
```diff
@@ -12,18 +13,34 @@ class MLPGemmOp(BaseOp):
...
-            self.config.mlp_act_func_type, self.config.transposed_mode)
+        if self.mlp_gemm_func != None:
+            output, residual_add = self.mlp_gemm_func(
+                input, residual, input_bias, weight_interm, weight_out, bias, gamma, beta, self.config.epsilon,
+                self.config.pre_layer_norm, self.config.mlp_after_attn,
+                weight_interm.scale if hasattr(weight_interm, 'scale') else torch.empty(1),
+                weight_out.scale if hasattr(weight_out, 'scale') else torch.empty(1), self.config.q_int8,
+                self.config.mlp_act_func_type, self.config.transposed_mode)
+        else:
+            # fallback
```
delock, line 37: Need to have a mechanism to enforce going through fallback path to test fallback path on cuda device.  Consider use an environment variable 'DS_KI_ENFORCE_FALLBACK'

csrc/cpu/comm/ccl.cpp
delock: Done.

.github/workflows/cpu-inference.yml
tjruwase: @delock, any update on the workflow? Is this PR ready for review?

delock: @tjruwase Yes, this PR is ready for review.   We have a minimum set for CPU workflow that can pass and will need your comments on what should be included in CPU workflow.

accelerator/real_accelerator.py
```diff
@@ -44,17 +44,41 @@ def get_accelerator():
...
+        # Currently we detect whether intel_etension_for_pytorch is installed
+        # in the environment and use CPU_Accelerator if the answewr is True.
+        # An alternative might be detect whether CUDA device is installed on
+        # the system but this comes with two pitfalls:
+        # 1. the system may not have torch pre-installed, so
+        #    get_accelerator().is_avaiable() may not work.
+        # 2. Some scenario like install on login node (without CUDA device)
+        #    and run on compute node (with CUDA device) may cause mismatch
+        #    between installation time and runtime.
+        try:
```
delock, line 60: Also need to add mechanism to override auto detection.  The logic should be controlled by environment variable:
DS_USE_CUDA_ACCELERATOR ==> CUDA path
DS_USE_CPU_ACCELERATOR ==> CPU path
DS_USE_XPU_ACCELERATOR ==> XPU path
This could allow user to have a more explicit way to use certain accelerator

delock: Done.

deepspeed/ops/transformer/inference/op_binding/mlp_gemm.py
delock: Use an environment variable DS_FORCE_KI_FALLBACK better be in a seperate PR.

accelerator/cpu_accelerator.py
```diff
@@ -0,0 +1,262 @@
...
+    def set_device(self, device_index):
+        return
+
+    def current_device(self):
+        return 0
+
+    def current_device_name(self):
+        return 'cpu'
+
+    def device_count(self):
```
delock, line 55: TODO: Number of devices on CPU could be flexible.  A more reasonable number is same number as number of sockets or sub numa clusters.   We should use numa util to help decide device_count for CPU Accelerator.

delock: fixed.

accelerator/cpu_accelerator.py
```diff
@@ -0,0 +1,271 @@
...
+        self._communication_backend_name = 'ccl'
+        self.max_mem = psutil.Process().memory_info().rss
+
+        def _check_and_mapping_mpich_env():
+            # Workaround when using Intel MPI launcher
+            # Model can check "RANK" and "WORLD_SIZE" for rank and world size
+            import os
+            if "RANK" not in os.environ and "PMI_RANK" in os.environ:
+                os.environ['RANK'] = os.environ.get('PMI_RANK')
+                print("mapping environment variable PMI_RANK to RANK")
```
tjruwase, line 26: To reduce log size, can you please change the 3 prints as follows:
1. Restrict to only rank 0
2. Use [log_dist](https://github.com/microsoft/DeepSpeed/blob/085981bf1caf5d7d0b26d05f7c7e9487e1b35190/deepspeed/utils/logging.py#L75).
3. Combine into single print statement

accelerator/cpu_accelerator.py
```diff
@@ -0,0 +1,271 @@
...
+        from deepspeed.utils.numa import get_numa_cores
+        # Count NUMA node for number of cpu accelerators. On machine with HBM
+        # In flat mode, HBM is in separate NUMA node with no cores on this node.
+        # Ignore these NUMA nodes with no cores.
+        numa_core_lists = get_numa_cores()
+        core_count = 0
+        for core_list in numa_core_lists:
+            if len(core_list) > 0:
+                core_count += 1
+        return core_count
```
tjruwase, line 65: It seems inconsistent that device  count can be > 1, but other device api functions like `current_device()` behave as if there is only one CPU device. 
I am concerned about client logic such as below that relies on some form of unique device identifier.  Can you please explain why this is fine? 

```python
num_devices = get_accelerator().device_count() 
 all_devices = [num_devices]
  for i, dev in enumerate(all_devices)
        if i == get_accelerator().current_device():
           # Do something about my device
        else:
          # Do something about other devices
```

accelerator/cpu_accelerator.py
```diff
@@ -0,0 +1,271 @@
...
+    @property
+    def IntTensor(self):
+        return torch.IntTensor
+
+    @property
+    def LongTensor(self):
+        return torch.LongTensor
+
+    def pin_memory(self, tensor):
+        return tensor.pin_memory(device=self.current_device_name())
```
tjruwase, line 228: I am curious how this works. I thought `pin_memory()` did not take [arguments](https://pytorch.org/docs/stable/generated/torch.Tensor.pin_memory.html?highlight=pin_memory).

accelerator/cuda_accelerator.py
```diff
@@ -167,7 +154,11 @@ def amp(self):
         return None
 
     def is_available(self):
-        return torch.cuda.is_available()
+        try:
+            import torch
+            return torch.cuda.is_available()
+        except ImportError:
+            return True
```
tjruwase, line 161: Two questions here:
1. Why not `import torch.cuda`?
2. Why return `True` if import fails?

deepspeed/inference/engine.py
```diff
@@ -343,16 +336,36 @@ def load_model_with_checkpoint(self, r_module):
         def load(module, state_dict, prefix):
             args = (state_dict, prefix, {}, True, [], [], error_msgs)
             if hasattr(module, 'weight'):
+                if module.weight.data.is_meta:
+                    # meta tensor cannot be casted or copied to, so we need to replace it with a normal tensor here
+                    module.weight = torch.nn.parameter.Parameter(data=torch.empty_like(module.weight.data,
```
tjruwase, line 341: Should this meta tensor replacement happen every time or only when using CPUAccelerator? In other words, does this PR change behavior for other accelerators?

deepspeed/inference/engine.py
```diff
@@ -432,16 +455,18 @@ def _load_checkpoint(self, load_dir, load_module_strict=True, tag=None):
         else:
             sd_loader = SDLoaderFactory.get_sd_loader_json(load_dir, self.checkpoint_engine)
 
-        if type(sd_loader) is list:
-            self.sd = torch.load(sd_loader[0], map_location='cpu')
+        checkpoint = sd_loader['checkpoints']
```
tjruwase, line 458: Is this backwards-compatible? It seems that `sd_loader` has changed?

setup.py
```diff
@@ -285,6 +285,7 @@ def create_dir_symlink(src, dest):
           'Documentation': 'https://deepspeed.readthedocs.io',
           'Source': 'https://github.com/microsoft/DeepSpeed',
       },
+      dependency_links=['https://developer.intel.com/ipex-whl-stable-cpu'],
```
jeffra, line 288: Is this different from https://pypi.org/project/intel-extension-for-pytorch/?

deepspeed/launcher/constants.py
tjruwase: Is there any issue to resolve here? Or was this just for documentation?

deepspeed/inference/engine.py
```diff
@@ -343,16 +336,36 @@ def load_model_with_checkpoint(self, r_module):
...
+                    # meta tensor cannot be casted or copied to, so we need to replace it with a normal tensor here
+                    module.weight = torch.nn.parameter.Parameter(data=torch.empty_like(module.weight.data,
+                                                                                       device="cpu"),
+                                                                 requires_grad=module.weight.data.requires_grad)
                 if 'query_key_value' in prefix:
                     module.weight = self.mp_replace.qkv_copy(module.weight.data, state_dict[prefix + 'weight'])
                 else:
                     module.weight = self.mp_replace.copy(module.weight.data, state_dict[prefix + 'weight'])
             else:
+                if module.norm.weight.data.is_meta:
```
tjruwase, line 349: Same here.

deepspeed/inference/engine.py
```diff
@@ -343,16 +336,36 @@ def load_model_with_checkpoint(self, r_module):
...
+                    # meta tensor cannot be casted or copied to, so we need to replace it with a normal tensor here
+                    module.norm.weight = torch.nn.parameter.Parameter(
+                        data=torch.empty_like(module.norm.weight.data, device="cpu"),
+                        requires_grad=module.norm.weight.data.requires_grad)
                 module.norm.weight = self.mp_replace.copy(module.norm.weight.data, state_dict[prefix + 'weight'])
             if prefix + 'bias' in self.key_list:
                 if hasattr(module, 'norm'):
+                    if module.norm.bias.data.is_meta:
```
tjruwase, line 357: Same here

deepspeed/inference/engine.py
```diff
@@ -343,16 +336,36 @@ def load_model_with_checkpoint(self, r_module):
...
+                        # meta tensor cannot be casted or copied to, so we need to replace it with a normal tensor here
+                        module.norm.bias = torch.nn.parameter.Parameter(
+                            data=torch.empty_like(module.norm.bias.data, device="cpu"),
+                            requires_grad=module.norm.bias.data.requires_grad)
                     module.norm.bias = self.mp_replace.copy(module.norm.bias, state_dict[prefix + 'bias'])
                 else:
+                    if module.bias.data.is_meta:
+                        # meta tensor cannot be casted or copied to, so we need to replace it with a normal tensor here
```
tjruwase, line 365: Same here

deepspeed/inference/engine.py
```diff
@@ -381,6 +394,15 @@ def load_module_recursive(module, prefix='', level=0):
 
         load_module_recursive(r_module)
 
+        embedding_weight = None
+
+        for n, p in r_module.named_parameters():
+            if "word_embeddings." in n or "embed_tokens." in n or "wte." in n:
+                embedding_weight = p
+        if embedding_weight is not None and hasattr(r_module, "lm_head") and hasattr(
+                r_module.lm_head, "weight") and r_module.lm_head.weight.is_meta:
```
tjruwase, line 403: I don't understand this logic. 
1. Are we setting `r_module.lm_head.weight` to the last embedding weight in the module parameters? 
2. Is it expected that `r_module.lm_head.weight` will be overwritten multiple times?

deepspeed/utils/timer.py
```diff
@@ -42,25 +42,40 @@ def __init__(self, name):
             self.event_timers = []
             self.start_event = None
             self.elapsed_records = None
+            self.start_time = 0.0
```
tjruwase, line 45: I think it would be cleaner to implement separate timers for synchronized devices (i.e., CPU) and asynchronous devices (i.e., GPU). This will help to avoid the repeated `is_synchronized_device()` calls and reduce the timer overhead.

op_builder/builder.py
```diff
@@ -433,7 +433,7 @@ def builder(self):
 
     def load(self, verbose=True):
         from deepspeed.git_version_info import installed_ops, torch_info
-        if installed_ops[self.name]:
+        if self.name in installed_ops.keys() and installed_ops[self.name]:
```
tjruwase, line 436: Could simplify as `if installed_ops.get(self.name, False)`

deepspeed/inference/engine.py
sywangyi: this replacement happens just once if the module weight is meta, after the replacement, it's not meta any more. and yes. it works for other accelerator because forward could not work if the weight is in meta device(no real data.) in every accelerators. I not only check in cpu but also in CUDA RTX8000

deepspeed/inference/engine.py
sywangyi: previously, the autoTP checkpoint path could not work with KI checkpoint format. KI checkout point is like "{"type": "ds_model", "checkpoints": ["/skyrex01/wangyi/gpt-j6/hub/models--EleutherAI--gpt-j-6B/snapshots/f98c709453c9402b1309b032f40df1c10ad481a2/pytorch_model.bin"], "version": 1.0}
"

deepspeed/inference/engine.py
sywangyi: same above

deepspeed/inference/engine.py
sywangyi: same above

deepspeed/inference/engine.py
sywangyi: same above

deepspeed/inference/engine.py
sywangyi: yes. KI also has such logic in https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/module_inject/load_checkpoint.py#L245
it will not overwritten multiple time only happen once when r_module.lm_head.weight.is_meta = true, will be false after the replacement.

deepspeed/inference/engine.py
delock: This benefits all acclerators.  On CPU we need this change in order to run model with large weight on server with less host memory.   It could also benefit server with multiple CUDA device, reduce the host memory needed.   This work is also tracked seperately in the following PR.  The plan is merge #3102 first so this PR alone can be more CPU specific.
https://github.com/microsoft/DeepSpeed/pull/3102

tjruwase: @sywangyi and @delock, thanks for the explanation. In that case, this sounds like a bug fix rather than a CPUAccelerator feature. Was this code working previously?

deepspeed/inference/engine.py
tjruwase: @sywangyi, thanks for the explanation. 

I am not familiar with KI code, so I will leave it to the experts to comment further. But it seems to me that if the intention is to modify exactly one parameter then it would be clearer to extract that one parameter rather than looping over `model.parameters()`.

deepspeed/inference/engine.py
tjruwase: I see. I will also leave this for KI experts to review.

deepspeed/inference/engine.py
sywangyi: yes, it's a bug fix. the previous checkpoint loading code in AutoTP path could not work with KI checkpoint json, and it's not sharded checkpoint loading just like KI.

tjruwase: Awesome! Thanks for the confirmation and bug fix.

tjruwase:
@delock, @mrwyattii, 

I notice the cpu-inference CI is failing for the following reason: 
image

Is it possible to have a filter to exclude this CI based on hardware requirements?

delock:
> @delock, @mrwyattii,
> 
> I notice the cpu-inference CI is failing for the following reason: image
> 
> Is it possible to have a filter to exclude this CI based on hardware requirements?
The detection part in this test shows this machine actually has AVX2 instruction set.
https://github.com/microsoft/DeepSpeed/actions/runs/4804701965/jobs/8550432875?pr=3041#step:4:48
image
However the control register are not correctly emulated by hypervisor which cause this problem, note the XCR value is not correct here.
https://github.com/microsoft/DeepSpeed/actions/runs/4804701965/jobs/8550432875?pr=3041#step:4:74

Currently we are using a hotfix for Intel Extension for Pytorch to get around this issue by not relying on XCR.  So Intel Extension for Pytorch could be more robust.  However it is not effective for this specific run.  Will need some investigation to understand why.

op_builder/builder.py
delock: Thanks for the suggestion, will change.

deepspeed/env_report.py
```diff
@@ -50,7 +50,7 @@ def op_report(verbose=True):
     for op_name, builder in ALL_OPS.items():
         dots = "." * (max_dots - len(op_name))
         is_compatible = OKAY if builder.is_compatible(verbose) else no
-        is_installed = installed if installed_ops[op_name] else no
+        is_installed = installed if (op_name in installed_ops.keys() and installed_ops[op_name]) else no
```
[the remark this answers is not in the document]
delock: will change.

csrc/cpu/foo.cpp
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0
+
+// DeepSpeed Team
+
+#include <torch/extension.h>
```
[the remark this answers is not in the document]
delock: This file it not needed, will remove.

deepspeed/launcher/constants.py
delock: just documentation, will mark as resolved.

accelerator/cpu_accelerator.py
delock: Yes, on CPU pin_memory should be a _nop_.  According to some discussion from https://stackoverflow.com/questions/55563376/pytorch-how-does-pin-memory-work-in-dataloader, pin_memory is used to accelerate copy tensor from CPU to GPU device.   Since this is a CPU Accelerator, the request to pin memory should not be necessary.   So CPU_Accelerator should simply return the tensor.

delock:
The Intel Extension for Pytorch we used here is a hotfix for the AVX2 instruction set detection.  In this particulary run seems the public wheel is downloaded from pypi.  We will have a new public release containing the AVX2 instruction set detection fix, then default installation of Intel Extension for Pytorch won't show this error.
> Currently we are using a hotfix for Intel Extension for Pytorch to get around this issue by not relying on XCR. So Intel Extension for Pytorch could be more robust. However it is not effective for this specific run. Will need some investigation to understand why.

accelerator/cpu_accelerator.py
delock: Thanks for the suggestion.  We plan to follow this PR https://github.com/microsoft/DeepSpeed/pull/3372 to move environment setting into multinode runner, rather than check and map in accelerator init function.  After that, we will remove the mapping and logging code in this PR.

accelerator/cpu_accelerator.py
delock: Yes, this will be a concern.   CPU as accelerator is different from discrete device that CPU is more 'divisible'.  For example, two sockets 2x56 core system can be used as a single accelerator with 112 cores, or 2 accelerators with 56 cores, or 4 accelerators with 28 cores.

A better logic is as follow:
1. Multinode runner: let device_count() return env LOCAL_SIZE and current_device() return env LOCAL_RANK.
2. Singlenode runner:  also use LOCAL_SIZE and LOCAL_RANK

If LOCAL_SIZE and LOCAL_RANK are not set, this is the case in runner when deepspeed is launched. (https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/launcher/runner.py#L405),  in this case return number of NUMA nodes on host.  In this situation we assume current_device() will not be called because local rank had not been decided yet.

delock: Will change to this new logic.   After the change a simple test result will be following:
1. Single node launching, without specify num_gpus, the number of rank is decided by numa nodes:
```
$ deepspeed --bind_cores_to_rank torch_dist.py
...
rank = 0, size = 2 device_count = 2 current_device = 0
rank = 1, size = 2 device_count = 2 current_device = 1
...

```

2. Two nodes launching, number of devices is specified in hostfile, each host has 2 slots
```
$ deepspeed --bind_cores_to_rank --hostfile=hostfile.txt --launcher impi --master_addr 192.168.10.44 torch_dist.py
...
rank = 0, size = 4 device_count = 2 current_device = 0
rank = 1, size = 4 device_count = 2 current_device = 1
rank = 3, size = 4 device_count = 2 current_device = 1
rank = 2, size = 4 device_count = 2 current_device = 0
...
```

setup.py
delock: Will remove this line.

deepspeed/utils/timer.py
delock: Seperate timer will require caller to treat synchronized device and asynchronous device differently.  By looking at a typical caller like this link(https://github.com/microsoft/DeepSpeed/blob/master/tests/onebit/test_mpi_perf.py#L56),  keep caller not distinguish between two kinds of device is still possible.

An alternative could be call `is_synchronized_device()` once in `init` function and store the result, then use that result to avoid repeatedly function call.

delock: Changed to call is_synchronized_device() only once during timer init.

delock:
@tjruwase need your input about UT data format.  Currently on pytorch CPU backend we only support BF16 training and inference.  This will break UTs using FP16 data format.   We are thinking about ways to allow UT skip certain data format basing on backend capability.   There are two options:
1. Add new interface in abstract accelerator to claim accelerator capability (combination of FP16/BF16 training/inference, and maybe int8/FP8 as well).   So UT would be able to make skip decision basing on return value of this interface.  Model code can also use this capability to decide which datatype to use for training or inference.
2. Check and skip base on accelerator name, this will not require accelerator interface change, but the implementation in UT will not be general enough and may require more change if there are other accelerator involved.

Which one do you prefer?  We plan to add UT skipping in a seperate PR, after CPU support is merged.  After that we are able to run more unit tests for CPU backend.

deepspeed/ops/transformer/inference/op_binding/mlp_gemm.py
```diff
@@ -40,35 +43,46 @@ def forward(self,
...
-                weight_out,
-                gamma,
-                self.config.epsilon,
-                weight_interm.scale if hasattr(weight_interm, 'scale') else torch.empty(1),  # type: ignore
-                weight_out.scale if hasattr(weight_out, 'scale') else torch.empty(1),  # type: ignore
-                self.config.dtype == torch.int8,
-                self.config.mlp_act_func_type,
-                self.config.transposed_mode)
+            # fallback
+            if self.config.mlp_after_attn and not self.config.transposed_mode:
```
delock, line 80: We plan to guard fallback path with a environment variable and not triggered by default.  Fallback path does help for debugging when we enable kernels in new backend.   But it also potentially cause confusion, we observe people think inference kernels had been 'implemented' when fallback path is triggered, so we will guard these fallback path with environment variable.

tjruwase:
> Hi @tjruwase need your input about UT data format. Currently on pytorch CPU backend we only support BF16 training and inference. This will break UTs using FP16 data format. We are thinking about ways to allow UT skip certain data format basing on backend capability. There are two options:
> 
> 1. Add new interface in abstract accelerator to claim accelerator capability (combination of FP16/BF16 training/inference, and maybe int8/FP8 as well).   So UT would be able to make skip decision basing on return value of this interface.  Model code can also use this capability to decide which datatype to use for training or inference.
> 2. Check and skip base on accelerator name, this will not require accelerator interface change, but the implementation in UT will not be general enough and may require more change if there are other accelerator involved.
> 
> Which one do you prefer? We plan to add UT skipping in a seperate PR, after CPU support is merged. After that we are able to run more unit tests for CPU backend.

@delock, thanks for raising this question. I prefer option 1 of adding new interface to abstract accelerator. 

For datatypes, I think we can add an interface that returns a list of supported torch data types. 
`def supported_dtypes(self) -> List[torch.dtype]` 
This interface will be future proof to allow new data types without interface change. Client code can query the return list of specific types. 

For training and inference, I don't think we need an accelerator interface for these capabilities. To me, the difference between training and inference is whether or not reduction is supported by the communication library.  Please let me know if I am missing anything. 

@jeffra, @mrwyattii FYI

deepspeed/comm/comm.py
```diff
@@ -189,26 +186,34 @@ def is_available() -> bool:
...
+    backend_name = get_accelerator().communication_backend_name()
+
+    if backend_name == NCCL_BACKEND:
+        if nccl_backend is not None and nccl_backend.is_initialized():
+            cdb = nccl_backend
+    elif backend_name == MPI_BACKEND:
+        if mpi_backend is not None and mpi_backend.is_initialized():
+            cdb = mpi_backend
+    elif backend_name == CCL_BACKEND:
+        if ccl_backend is None:
```
tjruwase, line 207: Why ccl not be handled like other backends and code simplified to 

```python
try: 
  ...
  elif backend_name == CCL_BACKEND:
    if ccl_backend is not None and ccl_backend .is_initialized():
      cdb = ccl_backend 
except ...
```

delock:
> 

The difference between training and inference is mainly from software priority aspect.  Sometimes a backend has forward OPs for a data type implemented first, then implement backward of the same OP in next release, so it would make sense to let a backend 'claim' support of certain datatype for inference only or training+inference.

deepspeed/comm/comm.py
delock: > Why ccl not be handled like other backends and code simplified to
> 
> ```python
> try: 
>   ...
>   elif backend_name == CCL_BACKEND:
>     if ccl_backend is not None and ccl_backend .is_initialized():
>       cdb = ccl_backend 
> except ...
> ```

Yes, this part should be simplified and backend creation should be put into init_deepspeed_backend.  The latest comm code reflect this change.

deepspeed/comm/comm.py
```diff
@@ -189,26 +191,23 @@ def is_available() -> bool:
...
+
+    if backend_name == NCCL_BACKEND:
+        if nccl_backend is not None and nccl_backend.is_initialized():
+            cdb = nccl_backend
+    elif backend_name == MPI_BACKEND:
+        if mpi_backend is not None and mpi_backend.is_initialized():
+            cdb = mpi_backend
+    elif backend_name == CCL_BACKEND:
+        if ccl_backend is not None and ccl_backend.is_initialized():
+            cdb = ccl_backend
```
tjruwase, line 210: I think we need to throw an exception if there is no backend match. Is that correct?

deepspeed/comm/comm.py
tjruwase: Thanks. With that change, I think `--prefer_deepspeed_comm`  is no longer needed and can be removed.

tjruwase:
> @tjruwase The difference between training and inference is mainly from software priority aspect. Sometimes a backend has forward OPs for a data type implemented first, then implement backward of the same OP in next release, so it would make sense to let a backend 'claim' support of certain datatype for inference only or training+inference.

I was not aware that backward ops are more challenging for an accelerator than forward ops. Is this referring to availability of high-performance kernel implementations? Can you please explain a bit more?

I think gradient reduction is another requirement for training, and that is provided by communication library like CCL. Is that not the case?

deepspeed/comm/comm.py
delock: Will remove prefer_deepspeed_comm and use CCLBackend as default CPU backend.  Note that CCLBackend only implemented allreduce/broadcast/barrier so it works for inference only.   We could extend it to other collectives if training support is needed.

deepspeed/comm/comm.py
delock: If no backend match, DeepSpeed will fallback to `TorchBackend` from the following line.
https://github.com/microsoft/DeepSpeed/pull/3041/files/e4e92e0a9838591cc32c5a5bb1e1f101123044a2#diff-2bf8bf2cd0ea4637cdd3397e6e27d38ce59e0ebc8daafcc583dbfcd8c092b0beR602

The logic is `TorchBackend` is a safe fallback when an accelerator had not implemented its own comm module.  i.e. if `NCCLBackend` had not been implemented, `TorchBackend` will be used.

delock:
From my understanding, the challenge is more in engineering effort.  The effort include implement backward OPs in primitive library, then integrate these primitive into PyTorch, then do extensive testing to ensure OOB training experience for that datatype.   It involves engineering effort in both development, integration and testing, so implement inference and training step by step will help focus efforts.

Different datatype do need gradient reduction for different datatype, and it will need support from communication library.   If the communication library is developed in house, coordinate datatype support with backward ops seems to be a reasonable choice.  May vary by case.

> > @tjruwase The difference between training and inference is mainly from software priority aspect. Sometimes a backend has forward OPs for a data type implemented first, then implement backward of the same OP in next release, so it would make sense to let a backend 'claim' support of certain datatype for inference only or training+inference.
> 
> I was not aware that backward ops are more challenging for an accelerator than forward ops. Is this referring to availability of high-performance kernel implementations? Can you please explain a bit more?
> 
> I think gradient reduction is another requirement for training, and that is provided by communication library like CCL. Is that not the case?

deepspeed/ops/transformer/inference/op_binding/mlp_gemm.py
delock: Fallback path are now guarded and will not be turned on by default.  Fallback path are used for debugging purpose, i.e. integrate one kernel before other kernels is available.

deepspeed/comm/comm.py
tjruwase: Yes, I am aligned with CCLBackend as default CPU backend

tjruwase:
@delock, thanks for the discussion. Perhaps we should handle datatype support and OP support differently in the UTs. 
1. For datatype support, let us use the proposed abstract accelerator interface: `def supported_dtypes(self) -> List[torch.dtype]`
2. For training/inference, let us use accelerator name to detect support. 

In other words, going back to your original option 1 and 2 proposals. We shall use option 1 for datatypes and option 2 for training/inference. What do you think?

deepspeed/comm/comm.py
tjruwase: Okay, I see. I think that is fine. Thanks!

delock:
> @delock, thanks for the discussion. Perhaps we should handle datatype support and OP support differently in the UTs.
> 
> 1. For datatype support, let us use the proposed abstract accelerator interface: `def supported_dtypes(self) -> List[torch.dtype]`
> 2. For training/inference, let us use accelerator name to detect support.
> 
> In other words, going back to your original option 1 and 2 proposals. We shall use option 1 for datatypes and option 2 for training/inference. What do you think?

Sounds good.   We will add the new interface and enable more UT in a seperate pull request.

tjruwase:
> The Intel Extension for Pytorch we used here is a hotfix for the AVX2 instruction set detection. In this particulary run seems the public wheel is downloaded from pypi. We will have a new public release containing the AVX2 instruction set detection fix, then default installation of Intel Extension for Pytorch won't show this error.
> 
> > Currently we are using a hotfix for Intel Extension for Pytorch to get around this issue by not relying on XCR. So Intel Extension for Pytorch could be more robust. However it is not effective for this specific run. Will need some investigation to understand why.

This seems to still be failing in CI. Are you running the new release of AVX2 instruction set detection fix yet? 
https://github.com/microsoft/DeepSpeed/actions/runs/4936158586/jobs/8823357797?pr=3041

I believe this is remaining issue to merge this PR.

delock:
We will have a new release for Intel Extension for PyTorch for AVX2 detection very soon, will update install link to fix the CI workflow for CPU.
> > The Intel Extension for Pytorch we used here is a hotfix for the AVX2 instruction set detection. In this particulary run seems the public wheel is downloaded from pypi. We will have a new public release containing the AVX2 instruction set detection fix, then default installation of Intel Extension for Pytorch won't show this error.
> > > Currently we are using a hotfix for Intel Extension for Pytorch to get around this issue by not relying on XCR. So Intel Extension for Pytorch could be more robust. However it is not effective for this specific run. Will need some investigation to understand why.
> 
> This seems to still be failing in CI. Are you running the new release of AVX2 instruction set detection fix yet? https://github.com/microsoft/DeepSpeed/actions/runs/4936158586/jobs/8823357797?pr=3041
> 
> I believe this is remaining issue to merge this PR.

delock:
Intel Extension for PyTorch had been updated to lastest version and AVX2 detection issue had been resolved.

> We will have a new release for Intel Extension for PyTorch for AVX2 detection very soon, will update install link to fix the CI workflow for CPU. A reference to the issue: [intel/intel-extension-for-pytorch#326](https://github.com/intel/intel-extension-for-pytorch/issues/326)
> 
> > > The Intel Extension for Pytorch we used here is a hotfix for the AVX2 instruction set detection. In this particulary run seems the public wheel is downloaded from pypi. We will have a new public release containing the AVX2 instruction set detection fix, then default installation of Intel Extension for Pytorch won't show this error.
> > > > Currently we are using a hotfix for Intel Extension for Pytorch to get around this issue by not relying on XCR. So Intel Extension for Pytorch could be more robust. However it is not effective for this specific run. Will need some investigation to understand why.
> > 
> > 
> > This seems to still be failing in CI. Are you running the new release of AVX2 instruction set detection fix yet? https://github.com/microsoft/DeepSpeed/actions/runs/4936158586/jobs/8823357797?pr=3041
> > I believe this is remaining issue to merge this PR.

Issue #2985, opened 2023-03-09
[BUG]  DeepSpeed Inference reports Signal code: Integer divide-by-zero when Seq length is 4096 for GPT2


zhen-jia:
**Describe the bug**
I am using DeepSpeed inference to run a GPT-2 model that has seq length 4096. It works fine for small input tokens. but is fails when I give it a input with shape like [1, 4000]. with the error of divide-by-zero. I traced the error a little bit, and the error comes from [softmax_context_func](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/ops/transformer/inference/op_binding/softmax_context.py#L34) .

**To Reproduce**
1. Simple inference script to reproduce

``` Python
# file name: test_hg.py
import torch
import transformers
import deepspeed
import gc
from transformers import AutoConfig, AutoModelForCausalLM

deepspeed.init_distributed("nccl")
world_size = 1
config = AutoConfig.from_pretrained('./config.json')
model = AutoModelForCausalLM.from_config(config, torch_dtype=torch.bfloat16)
model = model.eval()
torch.cuda.empty_cache()


model = deepspeed.init_inference(model,
                                 mp_size=world_size,
                                 dtype=torch.float16,
                                 replace_method='auto',
                                 replace_with_kernel_inject=True)

torch.cuda.empty_cache()
gc.collect()

model = model.module
batch = 1
seq_length = 4000

input_ids = torch.randint(1, 1000, (batch, seq_length), dtype=torch.int64)
attention_mask = torch.ones((batch, seq_length), dtype=torch.int64)

input_ids = input_ids.cuda()
attention_mask = attention_mask.cuda()


model(input_ids=input_ids, attention_mask=attention_mask)

```
`config.json` is
```
{
  "_num_labels": 1,
  "activation_function": "gelu_new",
  "architectures": [
    "GPT2LMHeadModel"
  ],
  "attn_pdrop": 0.1,
  "bos_token_id": 1,
  "embd_pdrop": 0.1,
  "eos_token_id": 1,
  "id2label": {
    "0": "LABEL_0"
  },
  "initializer_range": 0.02,
  "label2id": {
    "LABEL_0": 0
  },
  "layer_norm_epsilon": 1e-05,
  "model_type": "gpt2",
  "n_ctx": 4096,
  "n_embd": 1920,
  "n_head": 16,
  "n_inner": null,
  "n_layer": 24,
  "n_positions": 4096,
  "resid_pdrop": 0.1,
[...]

PR #3046, opened 2023-03-17
Softmax Scheduling Cleanup

cmikeh2:
This PR cleans up softmax scheduling:

1. Removes edge case whereby for long sequence lengths there would be a divide by 0 on the host (See https://github.com/microsoft/DeepSpeed/issues/2985)
2. Removes over allocation of threads for medium-sized sequences
3. Adds clarifying comments for what the behavior should be.


jeffra (approved):
As long as tests pass I am good with this change.

PR #3049, opened 2023-03-17
cleanup tp_parser

satpalsr:
Remove redundant code.

### DIFF
--- a/deepspeed/module_inject/auto_tp.py
+++ b/deepspeed/module_inject/auto_tp.py
@@ -118,5 +118,6 @@ class AutoTP():
                 gem_list = list(set(gem_list))
                 policy_list = AutoTP.update_policy_list(policy_list, module, gem_list)
                 gem_list = []
-        assert len(policy_list), "AutoTP not supported for model. Please use kernel injection since container policy for model exists."
+        assert len(policy_list), "AutoTP not supported for model. Please use kernel injection since container policy for model exists." \
+        if AutoTP.kernel_supported(module_list) else "Not able to determine model policy automatically. Please provide policy."
         return policy_list
--- a/deepspeed/module_inject/auto_tp.py
+++ b/deepspeed/module_inject/auto_tp.py
@@ -91,7 +91,6 @@ class AutoTP():
 
     def tp_parser(model):
         policy_list = []
-        module_list = []
         layer_list = []
         gem_list = []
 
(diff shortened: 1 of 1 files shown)


deepspeed/module_inject/auto_tp.py
```diff
@@ -119,6 +118,5 @@ def tp_parser(model):
                 gem_list = list(set(gem_list))
                 policy_list = AutoTP.update_policy_list(policy_list, module, gem_list)
                 gem_list = []
-        assert len(policy_list), "AutoTP not supported for model. Please use kernel injection since container policy for model exists." \
-        if AutoTP.kernel_supported(module_list) else "Not able to determine model policy automatically. Please provide policy."
+        assert len(policy_list), "AutoTP not supported for model. Please use kernel injection since container policy for model exists."
```
molly-smith, line 121: Not all models that fail here have kernel support.

molly-smith (changes requested):

deepspeed/module_inject/auto_tp.py
satpalsr: kernel support was checked twice. [Here](https://github.com/microsoft/DeepSpeed/blob/49fb30810127764bcad40540f0d0edf905cfc9fe/deepspeed/module_inject/auto_tp.py#L99) & [here](https://github.com/microsoft/DeepSpeed/blob/f1e4fb0b878e48d6fe33dbb2e0969d791d698e86/deepspeed/module_inject/auto_tp.py#L123).

satpalsr: Hm, correct. Thanks.

satpalsr:
Hey @molly-smith , Look #3130 
I think we should assert len(policy_list) == 2
and also change this
https://github.com/microsoft/DeepSpeed/blob/7faae44d507ea609696094b2fac2300ac3b24562/deepspeed/module_inject/auto_tp.py#L114

to allow more layers?
```
elif any(cond in layer for cond in ['outer', 'attention.dense']):
```

What to you think?

molly-smith:
> Hey @molly-smith , Look #3130 I think we should `assert len(policy_list) == 2` and also change this
> 
> https://github.com/microsoft/DeepSpeed/blob/7faae44d507ea609696094b2fac2300ac3b24562/deepspeed/module_inject/auto_tp.py#L114
> 
> to allow more layers?
> 
> ```
> elif any(cond in layer for cond in ['outer', 'attention.dense']):
> ```
> 
> What do you think?

Looks like this issue was resolved. Closing this PR.

PR #3051, opened 2023-03-18
Fix batch checking

HuangLK:
`batch` may be a `list`, depending on the behavior of the dataset, and it has no effect on subsequent execution


satpalsr:
@HuangLK Duplicate of PR #3042

PR #3060, opened 2023-03-20
Empty ZeRO3 partition cache

### LINKED ISSUE #3025: [BUG] zero3 memory leak on return from training loop
**Describe the bug**

Hmm, I thought I fixed the leak here https://github.com/microsoft/DeepSpeed/pull/2665 but there is one more of a similar nature, where the model gets gathered, but not ungathered when the weights are no longer needed.

When entering the training loop the weights are sharded. But on exit they are no longer sharded, so the memory isn't getting freed. Say, you want to do something else with the gpu - e.g. additional eval - it's going to OOM most likely.

**To Reproduce**

```
$ cat test.py
import deepspeed
from deepspeed.ops.op_builder import CPUAdamBuilder

from unit.common import DistributedTest, DistributedFixture
from unit.simple_model import *
from unit.util import required_minimum_torch_version

from unit.checkpoint.common import *

import pytest

class TestZeROCheckpointNew(DistributedTest):
    world_size = 1

    def test_1_load_module_only_new(self):
        tmpdir = 'zero_stage'
        zero_stage = 3
        config_dict = {
            "train_batch_size": 2,
            "optimizer": {
                "type": 'Adam'
            },
            "fp16": {
                "enabled": True,
                "initial_scale_power": 8
            },
            "zero_optimization": {
                "stage": zero_stage,
            }
        }
        hidden_dim = 10

        with deepspeed.zero.Init(config_dict_or_path=config_dict):
            models = [SimpleModel(hidden_dim, empty_grad=False) for _ in range(2)]

        print("CREATE", models[0].linears[0].weight[:5])

        dtype = torch.half
        ds_model = create_deepspeed_model(config_dict=config_dict,
                                          model=models[0],
                                          base_optimizer=None)

        data_loader = random_dataloader(model=ds_model,
                                        total_samples=5,
                                        hidden_dim=hidden_dim,
                                        device=ds_model.device,
                                        dtype=dtype)

        print("ORIGIN 1", models[0].linears[0].weight)
        for _, batch in enumerate(data_loader):
            loss = ds_model(batch[0], batch[1])
            ds_model.backward(loss)
            ds_model.step()

        print("ORIGIN 2", models[0].linears[0].weight[0,:5])
```

```
PYTHONPATH=. pytest --torch_ver 1.13.1 --cuda_ver 11.7 test.py -sv
```
gives:
```
ORIGIN 1 Parameter containing:
tensor([], device='cuda:0', dtype=torch.float16, requires_grad=True)
ORIGIN 2 tensor([ 0.0699, -0.0974, -0.1665, -0.1251, -0.1195], device='cuda:0',
       dtype=torch.float16, grad_fn=<SliceBackward0>)
```
you can see that the model wasn't partitioned and memory leaked for the remainder of the program.

It should be `tensor([], device='cuda:0', ...)` after `step` maybe?

As far as what the solution should be, I think `step` should call `partition` internally, similar to the fix in https://github.com/microsoft/DeepSpeed/pull/2665 - it scatters the updated weights anyway.

@tjruwase

tjruwase:
API to free GPU memory consumed by zero3 partition cache. 
Fix #3025


stas00:
oh, sorry, could this new method be added to the API docs please? Thank you, Tunji!

Perhaps with context:

> By default at the end of the training some parameters will remain unpartitioned and use up gpu memory. This happens as an optimization should you resume training again. If you'd like to clear out the cached parameters that use up gpu memory, you can call:

`deepspeed_engine.empty_partition_cache()`

deepspeed/runtime/zero/parameter_offload.py
```diff
@@ -259,6 +259,9 @@ def get_param_coordinator(self, training):
 
         return self.param_coordinators[training]
 
+    def empty_partition_cache(self):
+        self.partition_all_parameters()
```
jeffra, line 263: Can we add a docstring for this similar to what's in `docs/code-docs/source/zero3.rst`, then I think we can add it directly on our read-the-docs site as well.

stas00:
Thank you for adding the doc - looks great, Tunji!

PR #3064, opened 2023-03-21
fix pop off grad mistakenly

HuangLK:
In my case, `inputs[0]` and `inputs[1]` are partitioned grads, and there is no `mask` in the `inputs`, which pops up grads and causes an error


jeffra:
I believe this is related and might be fixed by [#2538](https://github.com/microsoft/DeepSpeed/pull/2538). Will get @duli2012 or @ShadenSmith to take a closer look.

Issue #2858, opened 2023-02-19
Compilation error for 0.8.1 with CUDA 11.2


hadim:
The conda-forge bot is building the new deepspeed packages for 0.8.1. See https://github.com/conda-forge/deepspeed-feedstock/pull/6#issuecomment-1436120690 for context.

All the builds for CUDA 11.2 are failing because of the below error:

```
[...]

PR #3085, opened 2023-03-23
remove `torch.cuda.is_available()` check when compiling ops

jinzhen-lin:
`torch.cuda.is_available()` is not necessary here. And I would cause https://github.com/microsoft/DeepSpeed/issues/2858 when compiling deepspeed >= 0.8.1 on a machine without gpu (e.g. docker image build).


tjruwase:
@jinzhen-lin, thanks for your contribution. But can you please provide some more details on the issue fixed by this PR? In my experience, the commented code works fine on machines without gpu including this [CI](https://github.com/microsoft/DeepSpeed/blob/master/.github/workflows/nv-torch-latest-cpu.yml). Thanks!

jinzhen-lin:
@tjruwase Sorry for absence of cpu builds checking before PR.

I notice that the cpu-only target environments was introduced recently (after v0.8.0) and deepspeed is mainly for gpu now. So we should always assume user want a cuda build, and we should do a cpu build when:

- we cannot get the cuda in the build environment or cuda version is incompatible with torch cuda version
- user specify a environment variable (e.g. `DS_BUILD_OPS_CPU`)

tjruwase:
@jinzhen-lin, thanks for updating the PR. This is an improvement but not quite cross-compilation. Nevertheless, this will suffice for now.

jeffra (approved):

PR #3102, opened 2023-03-27
add sharded checkpoint loading for AutoTP path to reduce the peak mem…

sywangyi:
…ory in initialization stage


sywangyi:
@delock @yao-matrix

sywangyi:
Hi，@molly-smith,  this PR is meant to reduce the host memory per Rank,  support shard loading in AutoTP path, same with shard loading in kernel injection path.

tjruwase (approved):

PR #3112, opened 2023-03-29
Optimize Softmax Kernel

molly-smith:
This PR improves the softmax kernel CUDA time. 

Achieves up to 65% softmax speedup with bloom-1.1b FP16.
Achieves up to 63% softmax speedup with opt-1.3b FP32.

image
image


cmikeh2 (approved):
Super excited to see this go in!

PR #3116, opened 2023-03-30
Disable ZeRO loading when load_module_only=True

ploshkin:
Hi,

I encountered the issue when loading weights of the pretrained model for further fine-tuning. The model was trained using bfloat16 in multi-GPU multi-host regime. I wanted to launch training on a different number of GPUs, but didn't find out how to tell DeepSpeed not to load ZeRO-states.

Based on the description of the `load_module_only` argument of the `load_checkpoint()` method I assume that the additional check will solve this issue.


tohtana:
Hi @ploshkin, 
Sorry that we haven't responded for a long time. Do you still have this issue with the latest version?
I will handle this if the latest code still has this issue.

tjruwase:
Closing for lack of activity

PR #3166, opened 2023-04-09
stage_1_and_2.py: do gradient scale only for fp16

guoyejun:


guoyejun:
for bf16, the gradient scale is not needed.

tjruwase (approved):

PR #3191, opened 2023-04-12
Add HE support for the rest of model containers

RezaYazdaniAminabadi:
Tested with GPT-J, GPT-Neo and GPT-Neox.


deepspeed/module_inject/containers/base.py
```diff
@@ -347,6 +347,10 @@ def copy_data_to_new_module(self):
         self.module.norm_w.data.copy_(self.input_nw.to(get_accelerator().current_device_name()))
         self.module.norm_b.data.copy_(self.input_nb.to(get_accelerator().current_device_name()))
 
+    def update_merged_qkv(self):
+        if hasattr(self, '_update_merged_qkv'):
+            self.update_merged_qkv()
```
lekurile, line 352: `self.update_merged_qkv()` should be `self._update_merged_qkv()` so that the Megatron feature function is called.

lekurile, line 352: `self.update_merged_qkv()` should be `self._update_merged_qkv()` so that the Megatron feature function is called.

deepspeed/module_inject/containers/features/megatron.py
```diff
@@ -24,6 +24,22 @@ def transpose_qkv_alignment(self, x):
         else:
             return torch.cat((q.reshape(-1), k.reshape(-1), v.reshape(-1)), dim=-1).reshape(x.shape)
 
+    def _align_qkv(self, x):
```
lekurile, line 27: Nit: Perhaps we can rename the two functions defined in this feature class, `_align_qkv` and `transpose_qkv_alignment`, to more explicit names since both functions have somewhat similar names w/ "qkv" and "align" in both.

lekurile (approved):

deepspeed/module_inject/containers/base.py
RezaYazdaniAminabadi: fixed it. Thanks @lekurile :)

deepspeed/module_inject/containers/features/megatron.py
```diff
@@ -24,6 +24,22 @@ def transpose_qkv_alignment(self, x):
...
+        (q, k, v) = torch.split(x_1, (x_1.shape[div_dim] // 3), dim=div_dim)
+        if len(q.shape) > 2:
+            return torch.cat((q.reshape(-1, q.shape[-1]), k.reshape(-1, q.shape[-1]), v.reshape(-1, q.shape[-1])),
+                             dim=0).reshape(x.shape)
+        else:
+            return torch.cat((q.reshape(-1), k.reshape(-1), v.reshape(-1)), dim=-1).reshape(x.shape)
+
+    def _update_merged_qkv(self):
+        self.qkvw = torch.nn.parameter.Parameter(self._align_qkv(self.qkvw).contiguous())
+        self.qkvb = torch.nn.parameter.Parameter(self._align_qkv(self.qkvb).contiguous())
```
mrwyattii, line 41: @RezaYazdaniAminabadi is self.qkvb only defined for containers inheriting from `features/megatron.py`? I wonder if we could isolate the `if self.qkvb is not None:` branches to the megatron container, rather than add to the base container?

cmikeh2 (approved):

jeffra (approved):

BoFeng2477:
So what does HE stands for?

PR #3201, opened 2023-04-12
fix whitespace

eisene:
Fixed by clang-format


eisene:

Issue #3202, opened 2023-04-12
[BUG] "with deepspeed.zero.Init()" is not idempotent


eisene:
**Describe the bug**
Intuitively, the Init() context seems like it should be idempotent. It should activate model partitioning, and calling it again shouldn't have any unexpected consequences.

However, currently, nesting the Init() context causes a hard to debug infinite recursion. This causes bugs in libraries that use DeepSpeed, notably Huggingface Transformers. For example, loading a pre-trained checkpoint of a custom encoder-decoder stack that uses the EncoderDecoderModel class results in a model that crashes when you train it.

It looks like this:

```shell
deepspeed train_bert_ds.py --checkpoint_dir . --num_layers 2 --h_dim 32                                      (deepspeed) 
[2023-04-12 16:22:59,798] [WARNING] [runner.py:181:fetch_hostfile] Unable to find hostfile, will proceed with training with local resources only.
[2023-04-12 16:22:59,806] [INFO] [runner.py:527:main] cmd = /home/eeisenst/miniconda3/envs/deepspeed/bin/python3.9 -u -m deepspeed.launcher.launch --world_info=eyJsb2NhbGhvc3QiOiBbMF19 --master_addr=127.0.0.1 --master_port=29500 --enable_each_rank_log=None train_bert_ds.py --checkpoint_dir . --num_layers 2 --h_dim 32
[2023-04-12 16:23:00,536] [INFO] [launch.py:133:main] WORLD INFO DICT: {'localhost': [0]}
[2023-04-12 16:23:00,536] [INFO] [launch.py:139:main] nnodes=1, num_local_procs=1, node_rank=0
[2023-04-12 16:23:00,536] [INFO] [launch.py:150:main] global_rank_mapping=defaultdict(<class 'list'>, {'localhost': [0]})
[2023-04-12 16:23:00,536] [INFO] [launch.py:151:main] dist_world_size=1
[2023-04-12 16:23:00,536] [INFO] [launch.py:153:main] Setting CUDA_VISIBLE_DEVICES=0
2023-04-12 16:23:01.543 | INFO     | __main__:log_dist:53 - [Rank 0] Creating Experiment Directory
2023-04-12 16:23:01.562 | INFO     | __main__:log_dist:53 - [Rank 0] Experiment Directory created at bert_pretrain.2023.4.12.13.23.1.addjtvxg
[...]

PR #2989, opened 2023-03-10
Nested zero.Init() and dynamically defined model class

tohtana:
This PR aims to fix the following issues described in #2811 and #2812.

- Nested `zero.Init()` causes infinite recursion
-  `zero.Init()` does not properly partition parameters of a model class that is dynamically defined inside the context

According to the discussion on #2812, this change enables `zero.Init()` to detect dynamically defined classes and properly sets wrappers to the constructors of the classes. This also manages the hierarchical contexts of `zero.Init()`.

PR #3203, opened 2023-04-12
Make deepspeed.zero.Init() idempotent

eisene:
This is a proposed fix for #3202, it makes the deepspeed.zero.Init() context ignore any nested enters, ie any additional enters that were not preceded by an exit.

It also fixes some whitespace and fixes a forgotten cls._old_init in init_subclass in the Init() context entry.


tjruwase:
@eisene, thanks for this PR. Can you please check for similarities with #2989?

eisene:
Unless I'm misunderstanding something, this is what I see:

First of all, in #2989 the test TestNewClassDeclaredInsideInit defined inside tests/unit/runtime/zero/test_zero_dynamic_class.py appears to have a bug. It should test whether model.fc.weight has the ds_id attribute without initializing the deepspeed_engine. The way it's implemented right now it always passes even if I comment out the two lines

```python
            cls.__init__ = partition_after(cls.__init__)
```

in partition_parameters.py. However, when checking for the attribute on model.fc.weight without initializing the deepspeed_engine, I get the expected failure when commenting out the above two lines.

In my pull request, the test TestNewClassDeclaredInsideInitFailure.test_new_class_declared_inside_init_failure fails because my way of doing it doesn't crash if you define a new class inside the Init() context. In my way of doing things I believe this is the correct behavior and there shouldn't be any problems with partitioning the parameters of such subclasses. This is because I add the line

```python
        def _init_subclass(cls, **kwargs):
            cls._old_init = cls.__init__    # <-- THIS IS THE ADDED LINE
            cls.__init__ = partition_after(cls.__init__)
```

This activates partitioning for all subclasses that are defined inside an Init() context. This seems to have been the desired behavior to begin with, except someone forgot to set the _old_init?

In the other pull request #2989 the all_wrapped_classes set boils down to enforcing the behavior that _init_subclass is only ever called together with _enable_class. This is why it never crashes because _old_init is missing, because it's always set inside _enable_class. But wouldn't it just be better to set _old_init inside _init_subclass? That seems to guarantee that all subclasses, whether they're dynamic or whatever, wherever they're defined, are always partitioned.

So in my way of doing things there should _not_ be a RuntimeError in TestNewClassDeclaredInsideInitFailure.test_new_class_declared_inside_init_failure and instead everything should behave correctly in this case. I can add these tests to my pull request and change the test_new_class_declared_inside_init_failure to be a test_new_class_declared_inside_init without a failure, to check whether the MyModel class was processed by DeepSpeed by checking for the ds_id attribute on model.fc.weight?

eisene:
This is the change I'm proposing to the test:

```python
# Copyright (c) Microsoft Corporation.

# DeepSpeed Team

import torch

import deepspeed


class TestNewClassDeclaredInsideNestedInit(DistributedTest):
    world_size = 1

    def test_new_class_declared_inside_nested_init(self):
        ds_config = dict(train_batch_size=1, zero_optimization=dict(stage=3))

        with deepspeed.zero.Init(config_dict_or_path=ds_config):

            class MyModel(torch.nn.Module):

                def __init__(self):
                    super().__init__()
                    self.fc = torch.nn.Linear(4, 4)

            with deepspeed.zero.Init(config_dict_or_path=ds_config):
                model = MyModel()

        # deepspeed_engine, *_ = deepspeed.initialize(model=model, config_params=ds_config)
        # ensure that zero3 processed the parameter
        # assert hasattr(deepspeed_engine.fc.weight, "ds_id")
        assert hasattr(model.fc.weight, "ds_id")


class TestNewClassDeclaredInsideInit(DistributedTest):
    world_size = 1

    def test_new_class_declared_inside_init(self):
        ds_config = dict(train_batch_size=1, zero_optimization=dict(stage=3))

        with deepspeed.zero.Init(config_dict_or_path=ds_config):

            class MyModel(torch.nn.Module):

                def __init__(self):
                    super().__init__()
                    self.fc = torch.nn.Linear(1, 1)

            model = MyModel()
        # ensure that zero3 processed the parameter
        assert hasattr(model.fc.weight, "ds_id")
```

tjruwase:
@stas00, FYI

tjruwase:
@samyam, FYI

eisene:
The HF accelerate tests fail because deepspeed.initialize calls shutdown_init_context directly. Then it's called again during __exit__, causing the counter to go negative. So I made shutdown_init_context also idempotent.

This clearly seems to be the intended behavior. Whoever wrote deepspeed.initialize seems to have known that there would be a second call to shutdown_init_context and expected it to do nothing.

eisene:
Agreed. I can make the engine object an optional argument to the context and have it store the state on the engine object if provided. If not provided it can raise a deprecation warning of some kind and store the state in the globals. Thoughts?

stas00:
The single-model concept has been used until now and the problem is that `zero.Init` is integrated into `from_pretrained` so that it can automatically partition the model transparently for the user. At that point there is no engine, this happens long before `deepspeed.initialize` so I'm not sure what engine object could be provided here. 

I think this new situation of multiple models calls for a major redesign of the `zero.Init` + `deepspeed.initialize` combo and how they interplay - but clearly `globals` won't work here, unless they are somehow magically not shared between models. I will let @tjruwase comment on that thanks to his work on https://github.com/microsoft/DeepSpeedExamples/tree/master/applications/DeepSpeed-Chat

tjruwase:
> @tjruwase, one other thing to be concerned about in the context of the discussion of [huggingface/diffusers#3076](https://github.com/huggingface/diffusers/pull/3076) is all those `global` objects - these will be a problem when using multiple models which don't all necessarily use the same setup. One model may use zero.Init whereas another won't. I mention it here since there are a few globals used in the zero.Init code. These probably should be tied to the engine object instead.

Regarding this, I had started process of eliminating globals in zero.Init. This was partially done [here](https://github.com/microsoft/DeepSpeed/blob/cc67f22f60f7ec20c61a53b4c3da4e8799d7de5a/deepspeed/runtime/zero/partition_parameters.py#L603-L607) for this PR #2953. There remains more global state to eliminate, and I will prioritize that. 

The goal is to optionally associate independent zero.Init context objects with each models.  The deepspeed engine object of a model is then naturally associated with the zero.Init context object of the model by a subsequent `deepspeed.intialize()`.  I would like to discuss more on this to clarify the API and design.

eisene:
At a high level this looks reasonable to me. I think what would be most helpful here is if I outlined how I'm using DS and what sorts of features I would potentially find helpful that may be relevant to the design principles.

My use case for DS is to train relatively large models, ~3B-7B, on consumer hardware. I'm working on a side project to make a machine translation system for a low resource language. The idea is to translate long texts, so I'm experimenting with different ways to modify existing model architectures to make them better at very long text and then fine-tuning them with DS. I'm leaning heavily on modular model design and using available pre-trained models in each slot. In this way what I'm doing is similar to the other discussions. However, in my case all the models are trained together on a single task, so my situation is simpler, at least for now. I do have ideas that will involve outright separate models, as well.

For example, I have a pre-trained encoder for my source language that I pre-trained myself from scratch, I have a large decoder into the target language that I take from the HF Hub, and I have a helper model that encodes some long context for the current sentence being translated and injects this into the decoder in some way. I'm experimenting with different architectures for this high level idea. This can create fairly complex model creation and data loading code.

ZeRO stage 3, especially NVMe offloading, has been a total game changer for me here. Unlike people working with large clusters and a large budget, I'm trying to squeeze every drop of performance from whatever hardware I happen to have lying around. I think being able to use DS to squeeze large models onto small hardware is very important, and not just inference but fine-tuning as well.

It may be useful to be able to manually decide that some parameters/gradients/optimizer state will go to some specific device, or at least be kept together. For example, I could have a smaller helper model that I want to place on a cheaper machine, ensure that all parameters/gradients/optimizer state for the helper model live on the same machine, and only have it send the output/final gradients of the helper model across the network. Depending on the size of the helper model, it seems possible to me that this could matter, even with all the optimizaton DS is already doing.

I also need NVMe checkpointing, but that's a separate conversation.

tjruwase:
> 
> I also need NVMe checkpointing, but that's a separate conversation.

@eisene, thanks for sharing this. It is very helpful to us to know important features to our users. I think our increased emphasis on modularity should help with flexible combination of models and features. You may be aware of our recent [DeepSpeed Chat](https://github.com/microsoft/DeepSpeedExamples/tree/master/applications/DeepSpeed-Chat) release where we support up to 5 models on a single rank to implement [RLHF finetuning](https://github.com/microsoft/DeepSpeedExamples/tree/master/applications/DeepSpeed-Chat/training/step3_rlhf_finetuning).

tjruwase:
@eisene, regarding this PR, can you confirm that your proposed test fails with master branch? Thanks!

eisene:
The test that's supposed to fail is `test_new_class_declared_inside_init` declared inside `tests/unit/runtime/zero/test_zero_dynamic_class.py`. I can confirm that it fails on the master branch for me.

tests/unit/runtime/zero/test_zero_nesting_init.py
```diff
@@ -20,6 +20,16 @@ def test_nesting_init(self):
...
-        assert hasattr(deepspeed_engine.weight, "ds_id")
+        assert hasattr(model.weight, "ds_id")
+
+    def test_initialize_inside_init(self):
+        ds_config = dict(train_batch_size=1, zero_optimization=dict(stage=3))
+
+        with deepspeed.zero.Init(config_dict_or_path=ds_config):
+            assert (deepspeed.zero.partition_parameters.zero_init_enabled == 1)
+            model = torch.nn.Linear(4, 4)
+            _, *_ = deepspeed.initialize(model=model, config_params=ds_config)
```
tjruwase, line 32: I see two issues here: 
1.  `deepspeed.initialize()` should not modify init_context in an externally visible way. For example, `deepspeed.zero.partition_parameters.zero_init_enabled` should be the same before and after. 
2. The following test fails with infinite recursion error.
```python
import torch 
import deepspeed 

ds_config = dict(train_batch_size=1, zero_optimization=dict(stage=3))
with deepspeed.zero.Init(config_dict_or_path=ds_config):
    with deepspeed.zero.Init(config_dict_or_path=ds_config):
        model = torch.nn.Linear(4, 4)
        _, *_ = deepspeed.initialize(model=model, config_params=ds_config)
```

Ideally, `deepspeed.initialize` should temporarily disable any existing init_context and re-enable afterwards.

eisene: I think I understand. I see at least two ways to deal with this:

1. The way it works now - hard shut down the `Init()` context in `initialize()`. This way the partitioning hook is installed in `DeepSpeedEngine` and then removed in `initialize()`. This seems to me to be brittle - if ever `DeepSpeedEngine` needs to be created outside `initialize()` it always needs to be accompanied by `shutdown_init_context()` for reasons that are not immediately obvious. Perhaps this is fine, I defer to your expertise here.
2. I tried explicitly excepting DeepSpeedEngine from the partitioning hook in partition_parameters.py:

```python
    def _enable_class(cls):
        if cls == deepspeed.runtime.engine.DeepSpeedEngine:
            return
        cls._old_init = cls.__init__
        cls.__init__ = partition_after(cls.__init__)

    def _disable_class(cls):
        if cls == deepspeed.runtime.engine.DeepSpeedEngine:
            return
        cls.__init__ = cls._old_init
```

This looks like it works. Maybe it would be better to make a class decorator that excludes classes from partitioning and decorate `DeepSpeedEngine` with it? Then you never have to call `shutdown_init_context()` explicitly.

tohtana: Hi @eisene, thank you for your great improvements. This PR significantly simplifies the management of `zero.Init()` contexts while it allows us to flexibly declare a new model class inside `zero.Init()`

Since the current code in this PR throws in some cases as pointed by @tjruwase, I submitted a separate PR #3592.
This can run the code shown by @tjruwase without an error, and also restores the context of  `zero.Init()` after `deepspeed.initialize()`.

I would be happy if we could discuss with you all including @tjruwase and @stas00.

eisene:
The issue has been resolved by #3592. The decorator is introduced separately in #3865.

PR #3205, opened 2023-04-12
zero3 checkpoint frozen params

### LINKED ISSUE #3090: [BUG] save/load checkpoint in zero3 fails to preserve frozen weights
**Describe the bug**

I was scratching my head on why I wasn't getting on resume the same weights as what was going in during model save.

The problem stems from having some weights that are frozen. The checkpoint saves only optimizer states which don't have frozen weights in it! so those weights are lost.

So when we resume - all the frozen weights aren't there and remain random. This would be extremely subtle to catch if only a few weights are frozen.

I validated that if I unfreeze everything, the resume works just fine - what goes in - it comes out correctly
so basically we need to figure out what to do with frozen weights, so that they survive resume. 

Using `stage3_gather_16bit_weights_on_model_save` isn't an option for LLM. And even if there was enough cpu memory it'd be extremely inefficient if needing to save checkpoints frequently during training.

@tjruwase

tjruwase:
- Enable checkpoint load/save of frozen params in zero stage 3. 
- Fix #3090 

- Pending task: Update zero_to_fp32.py to recover frozen weights.


tjruwase:
@stas00, FYI

stas00:
When the checkpoint is saved, I get almost all frozen weights saved with `size[0]`

```
python tools/convert_checkpoint/inspect_checkpoint.py /hf/m4-master-3/save_dir/opt_step-10/accelerator_state/pytorch_model/zero_pp_rank_0_mp_rank_00_model_states.pt
loading checkpoint file: /hf/m4-master-3/save_dir/opt_step-10/accelerator_state/pytorch_model/zero_pp_rank_0_mp_rank_00_model_states.pt
[tensor] module.lm_head.weight = torch.Size([0])
[tensor] module.lm_head.additional_fc.weight = torch.Size([0])
[tensor] module.model.decoder.embed_tokens.weight = torch.Size([0])
[...]
```

I think they need to be gathered before saving.

stas00:
I'm also thinking would this even work if there is a huge model with a lot of frozen params? There might not be enough memory to gather them all. Perhaps should save their fp16 shards instead? that would be much faster.

shaankhosla:
Hi @stas00 and @tjruwase, thanks for your work on this. I'm just checking to see if this would fix an error I'm getting using DeepSpeed and LoRA. Let me know if this isn't the place to ask.

I'm able to train "t5" using DeepSpeed Stage 3 and LoRA, however when I run the `load_state_dict_from_zero_checkpoint` command I get an error `KeyError: '_forward_module.model.base_model.model.encoder.embed_tokens.weight'`

Thanks again for all your help!

shaankhosla:
> 

Here it is: #3291 :)

jeffra (approved):

jomayeri (approved):

stas00:
Thank you for the quick solving and merge, Tunji and the team!

PR #3256, opened 2023-04-15
Fix missing scale attributes for GPTJ

cmikeh2:
This PR fixes two regressions introduced in the DeepSpeed chat release for GPT-J:

1. Checks for the `scale` attribute on all parameters before accessing.
2. Changes workspace offsets to avoid scenario where we are double using a buffer and over-writing data.


jeffra (approved):

heroes999:
@cmikeh2  I'm now using TP=8 with huggingface to inference llama, and I found memory overlap of input and output. I think it is due to the workspace offset? I see it is now 4 * buf_size，but may not be enough for TP=8?

PR #3289, opened 2023-04-18
Update pydantic version and corresponding changes

loadams:
PyDantic is preparing to update to v2 and we need to add support for their breaking changes from their [migration plan](https://github.com/pydantic/pydantic/blob/1896500fb6bc371270b94f343390e2121dc0e193/docs/migration.md).

- Overall migration plan [here](https://github.com/pydantic/pydantic/blob/1896500fb6bc371270b94f343390e2121dc0e193/docs/migration.md#migration-notes).
- Changes to config [here](https://github.com/pydantic/pydantic/blob/1896500fb6bc371270b94f343390e2121dc0e193/docs/migration.md#changes-to-config).
-

### DIFF
--- a/requirements/requirements-readthedocs.txt
+++ b/requirements/requirements-readthedocs.txt
@@ -4,6 +4,6 @@ hjson
 packaging
 psutil
 py-cpuinfo
-pydantic==2.0a4
+pydantic
 torch
 tqdm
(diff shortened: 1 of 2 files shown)


mrwyattii:
Looks like v2 will not be compatible with v1 -  we will need to add `pydantic>=2.0.0` to requirements once we make the switch. It might be worth adding `pydantic<2.0.0` to the requirements before that time.

PR #3290, opened 2023-04-18
Fix pydantic and autodoc_pydantic version to <2.0.0 until support is added.

loadams:
- pydantic is working on updating to a 2.0.0 that is not backwards compatible, we will need to add this, but since this will break our CI when it is released, this is a preventative step.
- autodoc_pydantic hasn't addressed how they will handle this, though it will be tracked in [this issue](https://github.com/mansenfranzen/autodoc_pydantic/issues/146).  I assume they will update to 2.0.0 once they add support as well.
- Once pydantic 2.0.0 is released, fixes will be added in #3289.

### DIFF
--- a/csrc/includes/conversion_utils.h
+++ b/csrc/includes/conversion_utils.h
@@ -266,7 +266,12 @@ DS_D_INLINE float2 to(__nv_bfloat162 val)
 template <>
 DS_D_INLINE __half to(double val)
 {
+#ifdef __HIP_PLATFORM_HCC__
+    float val_f = __double2float_rn(val);
+    return __float2half(val_f);
+#else
     return __double2half(val);
+#endif
 }
 template <>
 DS_D_INLINE __half to(float val)


mrwyattii (approved):

mansenfranzen:
@loadams You can now test the new **autodoc_pydantic** version supporting pydantic v2 via `pip install autodoc_pydantic==2.0.0`. The dev docs already showcase v2 support (see [here](https://autodoc-pydantic.readthedocs.io/en/main-2.x/index.html)). Migration guide is available [here](https://autodoc-pydantic.readthedocs.io/en/main-2.x/users/faq.html#migration-guide-from-v1-to-v2).

Any feedback is more than appreciated. The current **autodoc_pydantic** version `2.0.0` on pypi remains yanked for one week until final release.

mrwyattii:
@mansenfranzen Thanks for the update! We have another PR that we are working on for full v2 support: #3290 

We were waiting on support from the `autodoc_pydantic` (thanks for the work there!) and we will finish this work soon.

PR #3306, opened 2023-04-19
Fix memory leak in zero2 contiguous gradients

hablb:
No usage of extra_large_param_to_reduce if contiguous_gradients is False. It keeps reference of the param for the lifetime of the application.


tjruwase (approved):

hablb:
@tjruwase not sure why amd-mi200 test failed

PR #3311, opened 2023-04-20
hybrid_engine: check tuple size when fusing lora params

adammoody:
As written, I believe the ``if`` conditions might always evaluate to ``True``, since the body of the ``for`` loop would not execute if the ``lora_params`` list is 0-length.  I think the intent here was to verify that the tuple has three elements.


adammoody:
@tjruwase , here is a PR for the ``lora_param`` question.

Issue #3228, opened 2023-04-14
[BUG] batch_size check failed with zero 2 (deepspeed v0.9.0)


chenmingjiong:
**Describe the bug**
`AssertionError: Check batch related parameters. train_batch_size is not equal to micro_batch_per_gpu * gradient_acc_step * world_size 16 != 2 * 1 * 1`

This error only occurs when using deepspeed v0.9.0 and zero stage 2.
My code trains normally with (deepspeed v0.9.0 + zero stage 3), or (deepspeed v0.8.3 +zero stage 2).

**ds_report output**
```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /data/alpaca-lora/finetune_accelerate.py:520 in <module>                            │
│                                                                                                  │
│   517                                                                                            │
│   518                                                                                            │
│   519 if __name__ == "__main__":                                                                 │
│ ❱ 520 │   main()                                                                                 │
│   521                                                                                            │
│                                                                                                  │
│ /data/alpaca-lora/finetune_accelerate.py:347 in main                                │
│                                                                                                  │
│   344 │   │   eval_dataloader,                                                                   │
│   345 │   │   optimizer,                                                                         │
│   346 │   │   lr_scheduler,                                                                      │
│ ❱ 347 │   ) = accelerator.prepare(                                                               │
│   348 │   │   model, train_dataloader, eval_dataloader, optimizer, lr_scheduler                  │
│   349 │   )                                                                                      │
│   350 │   accelerator.print(model)                                                               │
│                                                                                                  │
│ /home/chenmingrui/.local/lib/python3.10/site-packages/accelerate/accelerator.py:1118 in prepare  │
│                                                                                                  │
│   1115 │   │   │   old_named_params = self._get_named_parameters(*args)                          │
│   1116 │   │                                                                                     │
│   1117 │   │   if self.distributed_type == DistributedType.DEEPSPEED:                            │
│ ❱ 1118 │   │   │   result = self._prepare_deepspeed(*args)                                       │
│   1119 │   │   elif self.distributed_type == DistributedType.MEGATRON_LM:                        │
│   1120 │   │   │   result = self._prepare_megatron_lm(*args)                                     │
│   1121 │   │   else:                                                                             │
│                                                                                                  │
│ /home/chenmingrui/.local/lib/python3.10/site-packages/accelerate/accelerator.py:1415 in          │
│ _prepare_deepspeed                                                                               │
│                                                                                                  │
[...]

PR #3324, opened 2023-04-20
Fix for dist not being initialized when constructing main config

mrwyattii:
Fix for #3228 

Also removing compatibility for torch < 1.8 in `deepspeed.comm` because we no longer officially support older torch versions.

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -125,17 +125,8 @@ def initialize(args=None,
 
     global dist
     from deepspeed import comm as dist
-    if dist_init_required is None:
-        dist_init_required = not dist.is_initialized()
-
-    if dist_init_required is False:
-        assert (
-            dist.is_initialized() is True
-        ), "Torch distributed not initialized. Please set dist_init_required to True or initialize before calling deepspeed.initialize()"
-    else:
-        if not dist.is_initialized():
-            dist_backend = get_accelerator().communication_backend_name()
-            dist.init_process_group(backend=dist_backend)
+    dist_backend = get_accelerator().communication_backend_name()
+    dist.init_distributed(dist_backend=dist_backend, dist_init_required=dist_init_required)
 
     # Set config using config_params for backwards compat
     if config is None and config_params is not None:
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -125,22 +125,17 @@ def initialize(args=None,
 
     global dist
     from deepspeed import comm as dist
-    from deepspeed.comm import supported_torch_version
-    # This supported_torch_version check is for torch1.2 compatibility only
-    dist_backend = get_accelerator().communication_backend_name()
-    if supported_torch_version:
-        dist.init_distributed(dist_backend=dist_backend, dist_init_required=dist_init_required)
-    else:
-        if dist_init_required is None:
-            dist_init_required = not dist.is_initialized()
+    if dist_init_required is None:
+        dist_init_required = not dist.is_initialized()
 
-        if dist_init_required is False:
-            assert (
-                dist.is_initialized() is True
-            ), "Torch distributed not initialized. Please set dist_init_required to True or initialize before calling deepspeed.initialize()"
-        else:
-            if not dist.is_initialized():
-                dist.init_process_group(backend=dist_backend)
+    if dist_init_required is False:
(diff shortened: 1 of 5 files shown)


deepspeed/__init__.py
```diff
@@ -119,6 +123,25 @@ def initialize(args=None,
 
     assert model is not None, "deepspeed.initialize requires a model"
 
+    global dist
+    from deepspeed import comm as dist
+    from deepspeed.comm import supported_torch_version
+    # This supported_torch_version check is for torch1.2 compatibility only
```
stas00, line 129: btw, you no longer support pt-1.2, so just as well remove that old code?

stas00:
I confirm that this fixes the problem. Could we please merge it asap, since other PRs are afflicted as well.

Thank you!

deepspeed/__init__.py
mrwyattii: @stas00 we were discussing this internally and agree we should remove it since we don't officially support older torch versions now. Latest commit removes it.

stas00: Spring clean up, yay! 

Thank you, Michael!

mrwyattii:
> p.s. The problem manifests in distributed Accelerate. Not sure how easy it'd be to write a test since Accelerate isn't easy to set up.

Do you have a minimal example we could add to our tests to ensure we don't break this in the future?

stas00:
As I'm not on Accelerate team, I'm just a user, let's ping @pacman100 who might be able to help with contributing such test. 

But since it's could take days it's far more urgent to merge your fix first. 

Thank you.

deepspeed/runtime/engine.py
```diff
@@ -234,6 +233,7 @@ def __init__(
 
         global dist
         from deepspeed import comm as dist
```
stas00, line 235: may be just moving this import to the top of the file and then `global dist` is no longer needed? I could be wrong

I did a quick test with the following and it seems to work:
```
diff --git a/deepspeed/runtime/engine.py b/deepspeed/runtime/engine.py
index f3480f76..e9541092 100644
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -95,6 +95,7 @@ from deepspeed.accelerator import get_accelerator
 from deepspeed.ops.op_builder import UtilsBuilder

 from deepspeed.runtime.config import DtypeEnum
+from deepspeed import comm as dist

 MEMORY_OPT_ALLREDUCE_SIZE = 500000000

@@ -141,8 +142,6 @@ BACKWARD_REDUCE_GLOBAL_TIMER = 'backward_allreduce'
 STEP_MICRO_TIMER = 'step_microstep'
 STEP_GLOBAL_TIMER = 'step'

-dist = None
-

 class EngineTimers(object):
     r"""Wallclock timers for DeepSpeedEngine"""
@@ -231,9 +230,6 @@ class DeepSpeedEngine(Module):

         self.checkpoint_engine = None

-        global dist
-        from deepspeed import comm as dist
-
         self._is_gradient_accumulation_boundary = None
         self.scale_wrt_gas = None

```

tjruwase (approved):

stas00:
Thank you for the quick fix and merge, Michael!

The only remaining step is a new minor release, since many users will not know to install from master.

jeffra:
> Thank you for the quick fix and merge, Michael!
> 
> The only remaining step is a new minor release, since many users will not know to install from master.

@stas00 v0.9.1 is now live:
https://pypi.org/project/deepspeed/0.9.1/
https://github.com/microsoft/DeepSpeed/releases/tag/v0.9.1

stas00:
Amazing! Thank you, Jeff!

pacman100:
> As I'm not on Accelerate team, I'm just its user, let's ping @pacman100 who might be able to help with contributing such test.

Hello, as this issue only comes up when using multi-gpu/multi-node setup, all the current slow tests in accelerate are failing (see the screenshots below). I don't know if DeepSpeed runs these slow tests on their front.

Commands to run the slow tests from accelerate main folder on a node with 2 gpus
```
export RUN_SLOW=true
python -m pytest -s -v ./tests/deepspeed
```

Screenshot 2023-04-21 at 12 02 39 PM
Screenshot 2023-04-21 at 12 02 51 PM

Issue #2911, opened 2023-02-28
[BUG] overflow warning needs to be different for fp16 and non-fp16


stas00:
**Describe the bug**

This code has an issue when it is run under non-fp16 regime.

https://github.com/microsoft/DeepSpeed/blob/da84e60d98d2e90f6f2094a219c98c8b41582eb9/deepspeed/runtime/zero/stage3.py#L1837-L1842

There are no scalers under bf16/fp32. So this warning is alarming to see - we rushed to see if somehow the config was broken, but it wasn't.

It should only say the `Attempted loss scale:...` part under fp16.

Most likely the same applies to its counterpart in stage 1/2.

Also do you think it'd be helpful to tell the user specifically if it's Inf vs. NaN? Since NaN isn't really an overflow or does it? Perhaps one of you with a more rigorous math background knows better. I think overflow is one of many types of NaN, thus NaN isn't always on Overflow. Please correct me if I'm wrong. 

The reason I'm asking this question is to help the user to know what to look for, NaNs, Infinity, else.

@tjruwase

tjruwase:
@stas00, I am curious what happened to the training in this case, since there is no loss scaling for non-fp16. Did the subsequent iterations continue to report overflows?

I am equally fuzzy about the relationships of overflows, NaNs, and Infinity, so will ask in the team.

stas00:
> @stas00, I am curious what happened to the training in this case, since there is no loss scaling for non-fp16. Did the subsequent iterations continue to report overflows?

We don't know yet as the training also crashes. When we looked at the grads (Yay!!!) it was actually a mix of Infinity and NaNs.

So the main part of this request is to separate the scaler overflow report (fp16) from all other overflow reports.

Plus perhaps the warning should say that the NaN was in the grads and not loss. A NaN loss would be a disaster.

> I am equally fuzzy about the relationships of overflows, NaNs, and Infinity, so will ask in the team.

Some good reading I found: https://stackoverflow.com/questions/59335027/what-is-nan-not-a-number-in-the-words-of-a-beginner

Sanster:
In my testing, training with stage2 bf16 also kept showing this message. Does this have any relevance to your question? Thanks.

<img width="1728" alt="截屏2023-03-03 19 53 30" src="https://user-images.githubusercontent.com/3998421/222713934-f295e459-8888-41d6-bfe3-d63b954daf6a.png">

tjruwase:
[...]

PR #3370, opened 2023-04-24
fixing default communication_data_type for bfloat16_enabled and docs

### LINKED ISSUE #2071: [BUG] FP16 used for all reduce even if BFLOAT16 is enabled
Even if BFLOAT16 is enabled, unless the communication datatype is also explicitly set to bfp16, fp16 is still used: https://github.com/microsoft/DeepSpeed/blob/9b70ce56e7af89d5226f9b06ebe1137407f371dc/deepspeed/runtime/engine.py#L702-L711

Moreover in zero, even if the communication datatype is set to bfp16 at some place the default value is fp16
https://github.com/microsoft/DeepSpeed/blob/9b70ce56e7af89d5226f9b06ebe1137407f371dc/deepspeed/runtime/zero/stage_1_and_2.py#L1334-L1338

clumsy:
fixes #2071

#2145 fixed the case when `communication_data_type` is set explicitly, unfortunately the default value for when `bfloat16_enabled=True` is not `bfp16`.
Adding unit tests for default `communication_data_type`.
Correcting the docs that mentioned `communication_data_type` as being of type `boolean`.

Before the fix:
```
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[fp16-fp16] PASSED                                                                                                                                    [ 11%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[fp16-bfp16] PASSED                                                                                                                                   [ 22%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[fp16-fp32] PASSED                                                                                                                                    [ 33%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[bfp16-fp16] PASSED                                                                                                                                   [ 44%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[bfp16-bfp16] PASSED                                                                                                                                  [ 55%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[bfp16-fp32] PASSED                                                                                                                                   [ 66%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[default-fp16] PASSED                                                                                                                                 [ 77%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[default-bfp16] FAILED                                                                                                                                [ 88%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[default-fp32] PASSED                                                                                                                                 [100%]

=============================================================================================================== FAILURES ===============================================================================================================
______________________________________________________________________________________________ TestZeroDtypeCocktail.test[default-bfp16] 
```

After the fix:
```
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[fp16-fp16] PASSED                                                                                                                                    [ 11%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[fp16-bfp16] PASSED                                                                                                                                   [ 22%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[fp16-fp32] PASSED                                                                                                                                    [ 33%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[bfp16-fp16] PASSED                                                                                                                                   [ 44%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[bfp16-bfp16] PASSED                                                                                                                                  [ 55%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[bfp16-fp32] PASSED                                                                                                                                   [ 66%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[default-fp16] PASSED                                                                                                                                 [ 77%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[default-bfp16] PASSED                                                                                                                                [ 88%]
tests/unit/runtime/half_precision/test_bf16.py::TestZeroDtypeCocktail::test[default-fp32] PASSED                                                                                                                                 [100%]
```

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -778,6 +778,9 @@ class DeepSpeedEngine(Module):
         if self.fp16_enabled():
             return torch.float16
 
+        if self.bfloat16_enabled():
+            return torch.bfloat16
+
         return torch.float32
 
     def postscale_gradients(self):
--- a/docs/_pages/config-json.md
+++ b/docs/_pages/config-json.md
@@ -181,7 +181,7 @@ Example of <i>**scheduler**</i>
 
 ### Communication options
 
-<i>**communication_data_type**</i>: [boolean]
+<i>**communication_data_type**</i>: [string]
 
 | Description                                                                                                                   | Default |
 | ----------------------------------------------------------------------------------------------------------------------------- | ------- |
--- a/tests/unit/runtime/half_precision/test_bf16.py
+++ b/tests/unit/runtime/half_precision/test_bf16.py
@@ -287,7 +287,7 @@ class TestZeroEmptyGrad(DistributedTest):
 
 
 @pytest.mark.parametrize("comp_type", [torch.float16, torch.bfloat16, torch.float], ids=["fp16", "bfp16", "fp32"])
-@pytest.mark.parametrize("comm_type", [torch.float16, torch.bfloat16], ids=["fp16", "bfp16"])
+@pytest.mark.parametrize("comm_type", [torch.float16, torch.bfloat16, None], ids=["fp16", "bfp16", "default"])
 class TestZeroDtypeCocktail(DistributedTest):
     world_size = 2
 
@@ -312,8 +312,11 @@ class TestZeroDtypeCocktail(DistributedTest):
             "zero_optimization": {
                 "stage": 2
             },
-            "communication_data_type": type_str[comm_type]
         }
+        if comm_type is not None:
+            config_dict["communication_data_type"] = type_str[comm_type]
+        else:
+            comm_type = comp_type
         hidden_dim = 10
 
         model = SimpleModel(hidden_dim)


tjruwase (approved):

Sanster:
@clumsy According to the discussion here  https://github.com/microsoft/DeepSpeed/issues/2911#issuecomment-1457181013, it seems intentional that bf16's default communication_data_type is set to float32. I'm curious, did you encounter any issues when using bf16 + fp32 communication_data_type?

clumsy:
Yes @Sanster, let me mention the regression issue we faced.

Perhaps more importantly this `communication_data_type` conversion didn't even happen in v0.6.0 for example. So in our `bfloat16=true` experiments (and w/o setting `communication_data_type` explicitly) the gradients were reduced as `bfloat16` (note no fp32 conversion).
When we switched to v0.7.3 we were unpleasantly pressurized that the model trained using newer DeepSpeed under-performed vs. older v0.6.0. It took us significant effort to bisect DeepSpeed versions and changes to identify the offending change.
So in our case not only we're willingly sacrificing precision in mantissa using bfloat16 to fit a larger model, but we were also inadvertently suffering from precision loss in exponent due to fp16 (that was default and hard-coded in v0.7.3) + during conversion between the two.

Now back to your question of whether to default to fp32. In the spirit of `Explicit is better than implicit` I'd like to avoid this, especially since the doc clearly states [here](https://github.com/microsoft/DeepSpeed/blob/4d269c6e4db390c2b427decc863e0a29df2fa99f/deepspeed/runtime/constants.py#L85): `By default, this feature is not enabled ('none' value)` - so as a user I don't expect any conversion happening behind the scenes as much as possible.
Another reason not to use fp32 by default I think is the fact that it [was not even possible to use it with `reduce_scatter` in ZeRO2 before](https://github.com/microsoft/DeepSpeed/blob/v0.7.3/deepspeed/runtime/zero/stage_1_and_2.py#L225).

tjruwase:
> When we switched to v0.7.3 we were unpleasantly pressurized that the model trained using newer DeepSpeed under-performed vs. older v0.6.0. It took us significant effort to bisect DeepSpeed versions and changes to identify the offending change.

@clumsy, thanks for sharing this explanation. Apologies for breaking BC with the above change. I am aligned with `Explicit is better than implicit` and will ensure we adhere better to that in practice.

PR #3371, opened 2023-04-25
add parameter "strict"

GongCQ:
add parameter "strict=strict" when layer.load_state_dict() in PipelineModule.load_state_dir()


tjruwase:
Fixed by different PR

Issue #3156, opened 2023-04-06
[BUG] RuntimeError: still have inflight params [<bound method Init._convert_to_deepspeed_param.<locals>.ds_summary of Parameter containing:


marsggbo:
**Describe the bug**

what's the possible reason for error below

```
  File "/home/xihe/xinhe/distNAS/DeepspeedNAS/train.py", line 200, in train_zero
    engine.backward(loss)
  File "/home/xihe/xinhe/deepspeed/DeepSpeed/deepspeed/utils/nvtx.py", line 11, in wrapped_fn
    ret_val = func(*args, **kwargs)
  File "/home/xihe/xinhe/deepspeed/DeepSpeed/deepspeed/runtime/engine.py", line 1980, in backward
    self.optimizer.backward(loss, retain_graph=retain_graph)
  File "/home/xihe/xinhe/deepspeed/DeepSpeed/deepspeed/utils/nvtx.py", line 11, in wrapped_fn
    ret_val = func(*args, **kwargs)
  File "/home/xihe/xinhe/deepspeed/DeepSpeed/deepspeed/runtime/zero/stage3.py", line 2088, in backward
    self._get_param_coordinator(training=True).reset_step()
  File "/home/xihe/xinhe/deepspeed/DeepSpeed/deepspeed/runtime/zero/partitioned_param_coordinator.py", line 185, in reset_step
    raise RuntimeError(
RuntimeError: still have inflight params [<bound method Init._convert_to_deepspeed_param.<locals>.ds_summary of Parameter containing:
```

**ds_report output**

```
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
async_io ............... [YES] ...... [OKAY]
cpu_adagrad ............ [NO] ....... [OKAY]
cpu_adam ............... [YES] ...... [OKAY]
fused_adam ............. [YES] ...... [OKAY]
fused_lamb ............. [YES] ...... [OKAY]
quantizer .............. [NO] ....... [OKAY]
random_ltd ............. [NO] ....... [OKAY]
sparse_attn ............ [NO] ....... [OKAY]
[...]

Issue #3068, opened 2023-03-21
[BUG] INFLIGHT parameters after evaluation


xiamengzhou:
**Describe the bug**
I adapted my training process from the hugging face trainer.py, so most of my trainer is similar to theirs. My model includes a language model and an external module for learning other parameters using information from the main model. I put the main and external parameters in separate groups in the optimizer. While testing the code on larger models with deepspeed, I encountered an assertion error after 50 training steps and 1 evaluation round. The error is related to the embedding matrix, which remains in "INFLIGHT" status after resuming training, while all other parameters are "AVAILABLE".

> AssertionError: {'id': 786, 'status': 'INFLIGHT', 'numel': 38603520, 'ds_numel': 38603520, 'shape':
> (50265, 768), 'ds_shape': (50265, 768), 'requires_grad': False, 'grad_shape': None, 'persist': False,
> 'active_sub_modules': {3}}

The program runs fine if I simply train a language model without an external module. 

**To Reproduce**
I can provide the code if necessary, but hope that I can get some help on how to further debug this issue :) 

**Expected behavior**
I expected the training runs fine after evaluation. 

**Screenshots**

**System info (please complete the following information):**
 - OS: Linux  4.18.0-425.13.1.el8_7.x86_64
 - GPU count and types: 1 machine x2 A100s 
 - Interconnects (if applicable)
 - Python version: python 3.8

**Launcher context**
I am using deepspeed to launch my program.

vlievin:
I am experiencing the same issue on a single node using `lightning` and a `transformers.BertModel` at the second epoch of lightning's "sanity checking":

> assert param.ds_status == ZeroParamStatus.AVAILABLE, param.ds_summary()
AssertionError: {'id': 0, 'status': 'INFLIGHT', 'numel': 0, 'ds_numel': 23440896, 'shape': (0,), 'ds_shape': (30522, 768), 'requires_grad': True, 'grad_shape': None, 'persist': False, 'active_sub_modules': {702}}

**system info**
- OS: Ubuntu 20.04.5 LTS, kernel: Linux 5.15.0-1030-gcp, arch: x86-64
- GPU count and types: 1 x A100-SXM4-40GB
- Python version: 3.9.16

**Relevant dependencies**
- `lightning==2.0.0`
- `transformers==4.27.1`
- `deepspeed==0.8.3`
- `torch==2.0.0`

HeyangQin:
[...]

PR #3380, opened 2023-04-25
make parameters status shared by all PartitionedParameterCoordinator instances

HeyangQin:
There could be multiple PartitionedParameterCoordinator instances, yet they currently manage the parameters in a standalone manner. Let's say we have PartitionedParameterCoordinator A and B. When A puts some parameters inflight, B is not aware of that and when B tries to use these parameters it will just error out. This PR addresses this issue by making the __InflightParamRegistry shared among all PartitionedParameterCoordinator instances.

This PR fixes https://github.com/microsoft/DeepSpeed/issues/3068 and hopefully would fix https://github.com/microsoft/DeepSpeed/issues/3156 as well 😉

### DIFF
--- a/deepspeed/runtime/zero/partitioned_param_coordinator.py
+++ b/deepspeed/runtime/zero/partitioned_param_coordinator.py
@@ -42,6 +42,7 @@ class ZeRoTraceMode(Enum):
 
 class PartitionedParameterCoordinator:
     """Handles partitioning and gathering of parameters."""
+    __inflight_param_registry = None
 
     class __InflightParamRegistry(UserDict):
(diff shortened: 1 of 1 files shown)


justHungryMan:
Hi, thanks for contributing to DeepSpeed. I've been experimenting with your commit but the [same problem](https://github.com/microsoft/DeepSpeed/issues/3156) persists.

There is a problem when using the generate function of a huggingface with zero3 in Pytorch Lightning.

HeyangQin:
Close as replaced by https://github.com/microsoft/DeepSpeed/pull/3462

Issue #2994, opened 2023-03-10
[BUG] Does bf16 support Zero stage 1 with pipeline?


lyj201002:
**Describe the bug**
I'm using Deepspeed-Megatron although,  using pipeline parallelism and setting 
```
"bf16": {
    "enabled": "auto"
  }
```
will step into the `NotImplementedError` in 
```python
#/opt/conda/lib/python3.8/site-packages/deepspeed/runtime/pipe/engine.py
def _exec_reduce_grads(self):
      self._force_grad_boundary = True
      if self.pipeline_enable_backward_allreduce:
          if self.bfloat16_enabled():
              if self.zero_optimization_stage() == 0:
                  self._bf16_reduce_grads()
              else:
                  assert self.zero_optimization_stage() == 1, "only bf16 + z1 are supported"
                  raise NotImplementedError()
          else:
              self.allreduce_gradients(bucket_size=MEMORY_OPT_ALLREDUCE_SIZE)
      self._force_grad_boundary = False
```
But when using transformer integrated deepspeed with Zero Stage 1/2/3. It work fine
The only diff I found is that, in megatron-deepspeed the model was a subclass of  PipelineModule whereas in transformer is not

I wonder whether deepspeed now support pipeline bf16 with zero stage 1 or just my code mistakes

tjruwase:
@lyj201002, thanks for reporting this bug. We are working on a fix.

PR #3399, opened 2023-04-27
Add ZeRO 1 support to PP for BF16.

jomayeri:
Addressing issue #2994

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -241,11 +241,10 @@ class PipelineEngine(DeepSpeedEngine):
         self._force_grad_boundary = True
         if self.pipeline_enable_backward_allreduce:
             if self.bfloat16_enabled():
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/pipe/engine.py
```diff
@@ -241,11 +241,10 @@ def _exec_reduce_grads(self):
         self._force_grad_boundary = True
         if self.pipeline_enable_backward_allreduce:
             if self.bfloat16_enabled():
-                if self.zero_optimization_stage() == 0:
+                if self.zero_optimization_stage() < 2:
```
tjruwase, line 244: Replace constant 2 with `ZeroStageEnum.gradients` as in  https://github.com/microsoft/DeepSpeed/blob/f7d71ec15376bc8188e5e8eca9c79177536c6e5d/deepspeed/runtime/zero/config.py#L67

tjruwase (approved):

PR #3407, opened 2023-04-28
[zero_to_fp32] fix shared param recovery

stas00:
Fixes: the


stas00:
@tjruwase, thinking aloud here - also what's the point of saving tensor placeholders, if they are stripped of their ds metadata - this is again a source of confusion and here was the cause of the bug. If something isn't usable let's not put it into the checkpoint.

mayank31398:
Thanks @stas00 
Ill test this out sometime today :)

tjruwase (approved):

Issue #3002, opened 2023-03-13
[BUG] Peft Training with Zero.Init() and Zero3 will increase GPU memory every forward step


dumpmemory:
**Describe the bug**
when i using Peft LoRA to train a gpt2 model, the gpu memory increase with every forward step with Zero3 adn zero.init function. when i disable zero.init, it worked as normal. 

**To Reproduce**
1. the all details can be found in https://github.com/huggingface/peft/issues/161

**Expected behavior**
run with no gpu memory increasing
**ds_report output**
Please run `ds_report` to give us details about your setup.
```bash

--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
 [WARNING]  async_io: please install the libaio-devel package with yum
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
async_io ............... [NO] ....... [NO]
cpu_adagrad ............ [NO] ....... [OKAY]
cpu_adam ............... [NO] ....... [OKAY]
fused_adam ............. [NO] ....... [OKAY]
fused_lamb ............. [NO] ....... [OKAY]
quantizer .............. [NO] ....... [OKAY]
random_ltd ............. [NO] ....... [OKAY]
 [WARNING]  please install triton==1.0.0 if you want to use sparse attention
sparse_attn ............ [NO] ....... [NO]
spatial_inference ...... [NO] ....... [OKAY]
transformer ............ [NO] ....... [OKAY]
stochastic_transformer . [NO] ....... [OKAY]
transformer_inference .. [NO] ....... [OKAY]
utils .................. [NO] ....... [OKAY]
--------------------------------------------------
DeepSpeed general environment info:
torch install path ............... ['/opt/miniconda/lib/python3.8/site-packages/torch']
torch version .................... 1.12.1
deepspeed install path ........... ['/opt/miniconda/lib/python3.8/site-packages/deepspeed']
deepspeed info ................... 0.8.2, unknown, unknown
torch cuda version ............... 11.3
[...]

PR #3413, opened 2023-04-30
Save tensors in context of memory_efficient_linear

tohtana:
By default, `torch.nn.functional.linear` is replaced with `LinearFunctionForZeroStage3`. However, `LinearFunctionForZeroStage3` causes memory leak in some usecases.

In PEFT's LoRA mentioned in #3002, the weight is passed after 'transpose'. 

```python
result = F.linear(x, transpose(self.weight, self.fan_in_fan_out), bias=self.bias)
```

LinearFunctionForZeroStage3 saves the weight in a map and the key is the object's ID. But a new transposed weight is created and the ID changes for each iteration. So the saved weights will increase through iterations.

This PR simply saves weight and bias in the context instead of the IDs.
I don't understand the intention of using IDs to store the weight. If saving IDs instead of tensors is a crucial part of this module, we need another approach to fix this.

### DIFF
--- a/deepspeed/runtime/zero/linear.py
+++ b/deepspeed/runtime/zero/linear.py
@@ -26,8 +26,6 @@ from deepspeed.runtime.utils import noop_decorator
 from deepspeed import comm as dist
 from deepspeed.accelerator import get_accelerator
 
-tensor_map = {}
-
 
 def print_rank_0(message, debug=False, force=False):
     if dist.get_rank() == 0 and (debug or force):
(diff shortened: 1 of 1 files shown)


tjruwase (approved):

tohtana:
@tjruwase Thank you for merging this PR!

As I mentioned, I didn't understand the intention of saving tensors in a global map.
I would be happy to fix again if we find any problem with this PR.

tjruwase:
@tohtana, I think your solution is the correct one.

tohtana:
Thank you for your reviewing, @tjruwase!

PR #3426, opened 2023-05-02
Avoid non_blocking copy from GPU to CPU

mmhab:
In Zero3 flow, when we do non_blocking copy from GPU to CPU and then copies the tensor to GPU (using the to() function) there is a race condition between these functions and there may be cases in which the to() function is executed before the copy_() completes. See the following discussion: https://discuss.pytorch.org/t/should-we-set-non-blocking-to-true/38234/17 
Therefore I think it is safer to do clone() instead of the non_blocking to().


tohtana:
Hello @mmhab, thank you for your PR.

As far as I understand, the preceding `copy_()` (L1209) is D2H copy and the non-blocking `to()` (L1212) is H2D copy.
They both run on the same CUDA stream. So it is ensured that the `to()` runs after the `copy_()` completes.

What are your thoughts on this?

mmhab:
Hi @tohtana,
I agree with you.
Thanks!

PR #3440, opened 2023-05-04
[MiCS] [Fix] saving and loading model checkpoint logic for MiCS sharding

zarzen:
- Only the first partition group will save the model checkpoints 
- Need to avoid call `dist.barrier` on the `WORLD` group.
- Including the support for loading the partitioned model checkpoint as ZeRO-3


zarzen:
Hi there, one of the unit test is failed, but I didn't see the corresponding error. 
Can someone help to retry with the `nv-torch19-v100 / unit-tests`

zarzen:
Hi @samadejacobs @tjruwase 
would you mind taking a look at this PR? It is an easy fix.
The only thing that can affect the current logic of ZeRO-3 is the in `deepspeed/runtime/engine.py: _create_zero_checkpoint_files`, in which I let the `dist.barrier` function get called on ranks within the `self.optimizer.dp_process_group`.

tests/unit/checkpoint/test_mics_optimizer.py
```diff
@@ -0,0 +1,61 @@
...
+                }
+            },
+            "fp16": {
+                "enabled": True,
+                "initial_scale_power": 8
+            },
+            "wall_clock_breakdown": True,
+            "zero_optimization": {
+                "stage": 3,
+                "mics_shard_size": 2
```
tjruwase, line 41: Is it possible to test different `mics_shard_size` values, e.g., 1, 2, and 4?

deepspeed/comm/torch.py
```diff
@@ -20,13 +20,24 @@ def is_torch_two():
         return False
 
 
+def torch_ver_ge_1_13():
```
tjruwase, line 23: This change seems to be independent of checkpointing. Is that correct?

zarzen: Right this is independent to checkpointing part, but found the issue along the for fixing checkpointing

tests/unit/checkpoint/test_mics_optimizer.py
zarzen: Yes, it is possible for 2 and 4. Not sure the effect on 1, because if 1 is used, it means zero-2.

tjruwase: I guess you mean 1 would be vanilla zero-3. The expectation is that it would be functional but not performant, right? I think that is okay for a unit test as it will help us catch unexpected breakages.

deepspeed/comm/torch.py
tjruwase: Sounds good. Thanks for the clarification!

tests/unit/checkpoint/test_mics_optimizer.py
zarzen: my mistake. For `"mics_shard_size": 1`, I think it is equivalent to DDP, in which each device holds a complete copy of the model states (grads, params, optimizer states)

tjruwase: Thanks for the clarification. I think it would still be a good test case.

zarzen: I see, thanks for the suggestions, I added more test as you suggested

tjruwase: Sorry, I don't see the modification to make shard size configurable in the unit test. Perhaps, I am missing something. 

What I was expecting was something like
```python
@pytest.mark.parametrize('shard_size', [1, 2, 4])
class TestMiCSCheckpoint(DistributedTest):
   def _toy_model_config(self, shard_size):
    ...
     "mics_shard_size": shard_size
    ...
    def test_load_optimizer_state(self, tmpdir, shard_size):   
      ....
      config_dict, hidden_dim, models = self._toy_model_config(shard_size)

```

zarzen: sry, I was looking into a wrong file. 
The tests are updated with the shard_size [1,2,4] covered.

tjruwase: Awesome! Thanks!

tjruwase:
@zarzen, I am unable to rebase this PR. Did you set some restrictions on your branch?

zarzen:
I think it is due to the permission policy on `dmlc` account. 
I just rebased with the master branch. Will take a look at the permission setup.

PR #3501, opened 2023-05-10
Align InferenceEngine to store ms in _model_times

HolyFalafel:
When using cuda_events, the measured model time is stored in ms.
   When not using cuda_events, the measured model time was stored in seconds.
   This commit fixes the units and aligns them to store ms, the same as elapsed() function.
   This was observed when running the following pytest:
   unit/inference/test_model_profiling.py::TestModelProfiling::test[False-True-roberta-base-fill-mask]

   Returned values were:
     count=0 e2e_t=895.174312 model_t=0.8529715538024902
     count=1 e2e_t=7.500252 model_t=0.0041310787200927734
     count=2 e2e_t=3.887346 model_t=0.0018568038940429688
     count=3 e2e_t=3.577845 model_t=0.0016334056854248047
     count=4 e2e_t=3.43976 model_t=0.0016703605651855469
     count=5 e2e_t=3.310903 model_t=0.0016107559204101562
     count=6 e2e_t=3.299556 model_t=0.001603841781616211
     count=7 e2e_t=3.605722 model_t=0.0015969276428222656
     count=8 e2e_t=3.273741 model_t=0.0015516281127929688
     count=9 e2e_t=3.46306 model_t=0.0016617774963378906

   The units difference is observed here, when model_t is in ther order of 10e-3 comparing to e2e_t


mrwyattii (approved):
Thanks @HolyFalafel

Issue #713, opened 2021-01-30
Single node launcher issue


jeffra:
When running something like: `deepspeed --include worker-1 test.py` from worker-0 we currently run test.py only on worker-0. This is due to this line of code in our runner:

https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/launcher/runner.py#L307

We currently prevent `pdsh` from launching the job if the number of workers is 1. However, we do not currently check to make sure that the 1 worker is the local worker we are invoking the `deepspeed` launcher from.

PR #3519, opened 2023-05-11
Extend DeepSpeed inference initialization API with a 'quantize_groups' argument

sakogan:
This PR allows setting the number of quantization groups to be used during inference for the given model weights. Once merged, one should be able to control this number by passing the `quantize_groups` argument to a `deepseed.init_inference` call, e.g., like is done in this inference test script: https://github.com/microsoft/DeepSpeedExamples/compare/master...sakogan:DeepSpeedExamples:inference-test-enhance
(Note that the latter example will be submitted to the DeepSpeedExamples repo as a separate PR if this PR is merged.)

Setting 'quantize_groups' is allowed only when the quantization is enabled, i.e., when `dtype` is set to `int8`.

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -21,7 +21,7 @@ from .runtime.engine import ADAM_OPTIMIZER, LAMB_OPTIMIZER
 from .runtime.hybrid_engine import DeepSpeedHybridEngine
 from .runtime.pipe.engine import PipelineEngine
 from .inference.engine import InferenceEngine
-from .inference.config import DeepSpeedInferenceConfig
+from .inference.config import DeepSpeedInferenceConfig, QuantizationConfig
 from .runtime.lr_schedules import add_tuning_arguments
 from .runtime.config import DeepSpeedConfig, DeepSpeedConfigError
 from .runtime.activation_checkpointing import checkpointing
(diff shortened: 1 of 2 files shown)


RezaYazdaniAminabadi:
Hi @sakogan,

The changes in this PR looks good to me. May I ask that you link the PR on the example side so that I know how to use this feature?
Thanks,
Reza

sakogan:
thanks for the review @RezaYazdaniAminabadi. I will submit the corresponding PR into the examples repo shortly.

sakogan:
@RezaYazdaniAminabadi I opened a PR [#713](https://github.com/microsoft/DeepSpeedExamples/pull/713) on the examples repo that adds the corresponding argument to the inference test script

sakogan:
Yes. Here is an example command:
`deepspeed /path/to/DeepSpeedExamples/inference/huggingface/text-generation/inference-test.py --batch_size 1 --use_kernel --use_meta_tensor --dtype int8 --model bigscience/bloom-7b1 --quantize_groups 32`

deepspeed/__init__.py
```diff
@@ -337,6 +337,14 @@ def init_inference(model, config=None, **kwargs):
...
     config_dict.update(kwargs)
 
+    # Set the number of weight quantization groups if an optional 'quantize_groups' argument is given
+    if "quantize_groups" in config_dict:
+        if not ("dtype", torch.int8) in config_dict.items():
+            raise ValueError("'dtype' argument expected int8 when 'quantize_groups' argument is provided")
+        quant = QuantizationConfig()
+        quant.weight.q_groups = config_dict.pop("quantize_groups")
+        config_dict["quant"] = quant
+
```
mrwyattii, line 347: I believe you are adding `quantize_groups` as a shortcut for the admittedly convoluted current quantize config settings? For example, with this you could just pass `quantize_groups=2` rather than `quant={"weight":{"q_groups":2}}`.

But perhaps we should look into how we can simplify the quantize settings or at the very least add this logic to the `DeepSpeedInferenceConfig` class as a pydantic validator (so that the config logic is consolidated there).

kiucho:
Hi, dose deepspeed support int8 inference? With next command I get following error
command: `deepspeed --num_gpus 1 inference-test.py --use_kernel --dtype float16 --model EleutherAI/gpt-j-6b`
`!!!! kernel execution error. (m: 4096, n: 3, k: 12288, error: 15)`

Issue #1300, opened 2021-08-11
why use for all-reduce when `contiguous_gradients` is False?


zarzen:
Hi there, 

On single node with 8A100 GPUs connected with nvlinks, I can still observe the performance benefits with `contiguous_gradients` set to False. e.g., 10B model backward time: *~1329.30ms* vs ~ *1593.24ms* (contiguous_gradients=False vs. contiguous_gradients=True). 

I am wondering where is the benefits coming from. The all-reduce suppose to incur more communication overheads than reduce-scatter operation. 

Could please explain a littile bit about why use all-reduce for each gradient tensor? (I not completely sure that zero3 will do all-reduce for all gradient tensors. But base on the code path at here, I guess so)
https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/zero/stage3.py#L2314 

And what could be the overheads introduced with setting `contiguous_gradients=True`? 

Thanks!

Issue #622, opened 2020-12-28
Zero stage 2 : contiguous_gradients and "Reduce"


anijmlt:
I am running the 8B model as described in Table 1 of the [Zero paper.](https://www.computer.org/csdl/pds/api/csdl/proceedings/download-article/1oeOS4nBiik/pdf) on 8 GPUs.
I notice that the "contiguous_gradients" setting in the config seems to control whether "Reduce" or "Allreduce" is used for gradient reduction in backward pass. That is, I see the following in the NCCL Debug log only when "contiguous_gradients" is "true":
![image](https://user-images.githubusercontent.com/37815929/103239520-94ca9a00-4902-11eb-9df5-329f1a1a7f91.png)

I see this being referenced in #264 but it wasn't clear why "contiguous_gradients" should control the communication pattern. As the answer in #264 mentioned it should only "defragment the memory during backward propagation".

Issue #264, opened 2020-06-15
Deepspeed communication pattern clarification


nishantagrawalgit:
Queries regarding communcation calls in Deepspeed for Stage1(Zero1) and Stage2(Zero2):

1.  As per paper, Stage1(Zero1) has Optimizer split implementation and we should see below calls on 2 GPU's(1 layer GPT2/345m model) with NCCL:
    1 call - Reduce_Scatter for Gradients 
    1 call - All_Gather for optimizer
    
	As per details mentioned in one of prior github issues(Deepspeed) that reduce_scatter for stage 1 is not implemented. Thats why it calls all_reduce for stage1 Zero optimizer. Is it still not implemented? 
	
    Based on the logs we are seeing All_Reduce calls in the backpass + Reduce_Scatter calls post backpass before optimizer step. So can you explain why reduce_scatter needed here as we can just do all_reduce for the gradient and all_gather for optimizer? 
	
	We are seeing an overflow occur at the end of Optimizer step(we expected to see All_Gather comms at the end of the optimizer which we dont observe). Can you pls explain why overflow occurs as mentioned in the below logs:  
	rank 0 detected overflow nan in tensor 0:1 shape torch.Size([50304, 1024])
	[deepspeed] OVERFLOW! Skipping step. Attempted loss scale: 4294967296, reducing to 4294967296

    
2.  For Stage2(Zero2), Zero with Optimizer+gradient split- In the backpass the gradient should get scattered to gpu that owns that gradient chunk for reduce, and then perform all gather for optimizer(as per paper). In the run logs we are seeing:
	No calls in the backpass which means no comms and compute overlap in backpass because of enabling fusion buffer. As a result we have see only 2 All_Reduce calls post backpass for gradients. Pls confirm if our understanding is correct?

samyam:
@nishantagrawalgit  Thank you for your interest in DeepSpeed. 

1. 
a) Reduce Scatter for Stage 1 is now already implemented. It is the default option for Stage 1, and we have removed the allreduce version from Stage 1
  
We use reduce_scatter instead of all_reduce to reduce the communication volume by half.

[...]

PR #3545, opened 2023-05-15
Resolve hard dependency on MOE for contiguous_gradients on stage 1

abhilash1910:
Motivation for This PR:
In [engine.py](https://github.com/microsoft/DeepSpeed/blob/5c6da1f001f936234a31a238e71ca386e34eb51a/deepspeed/runtime/engine.py#L1408) there is a dependency on contiguous_gradients on MoE, for Stage1 which would imply that even with "contiguous_geadients" enabled, Stage 1 would still default to ["buffered_allreduce" ](https://github.com/microsoft/DeepSpeed/blob/5c6da1f001f936234a31a238e71ca386e34eb51a/deepspeed/runtime/zero/stage_1_and_2.py#L1210) during reduce_ipg_grads.   If MoE is disabled (no experts) , then even with contiguous_gradients set, we would have to see allreduce pathway for Stage 1; which is a hard dependency on MoE. So this PR is to retain the condition that irrespective of MoE , contiguous_gradients if set should remain set on deepspeed. In this case, without MoE and inducing contiguous_gradients=True allows stage 1 to show reduce collective in place of allreduce (as it should). 
With MoE experts set it also goes through the reduce path (does not break functionality). 
Also this is for issues : #622  #264  #1300 

@tjruwase   requesting review .

Ps: CLA is signed .


abhilash1910:
@tjruwase  could you please review ? cont_gradients should not be hard binded with Moe in engine , as this would mean enabling Moe layers everytime to see cont_grads.  This also affects performance, hence seeking your review on this.

awan-10:
@abhilash1910  -- Thanks for the PR! Have you tested performance and found issues? It will be helpful to add some more details if you can. 

@RezaYazdaniAminabadi, @jeffra, and @tjruwase - I think we should accept this PR. Do you guys remember we did something like this but never pushed a public PR? Thoughts?

abhilash1910:
Thanks @awan-10 , I found that decoupling the moe provides reduce collective ; I am checking for perf if there is any impact . I will add the log file for the perf characterisations .  But moe should be detachable from contiguous gradients.

PR #3551, opened 2023-05-16
Fixing bf16 test 

jomayeri:
The test TestAdamBF16ZeroOneCycleCompatibility was failing do to missing `train_batch_size` in its config.

### DIFF
--- a/tests/unit/runtime/half_precision/test_bf16.py
+++ b/tests/unit/runtime/half_precision/test_bf16.py
@@ -27,6 +27,7 @@ class TestAdamBF16ZeroOneCycleCompatibility(DistributedTest):
             pytest.skip("cpu-adam is not compatible")
 
         config_dict = {
+            "train_batch_size": 4,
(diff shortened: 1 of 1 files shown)


tests/unit/runtime/half_precision/test_bf16.py
```diff
@@ -27,6 +27,7 @@ def test(self, zero_stage=2, use_cpu_offload=False):
             pytest.skip("cpu-adam is not compatible")
 
         config_dict = {
+            "train_batch_size": 4,
```
tjruwase, line 30: It is more robust to use `"train_micro_batch_size_per_gpu": 1`

jomayeri: Should I do this for all tests in that file?

tjruwase: Yes, that would be great. Thanks!

tjruwase (approved):

PR #3553, opened 2023-05-16
Expose Consecutive Hysteresis to Users

Quentin-Anthony:
There's already a nice `consecutive_hysteresis` feature in the DynamicLossScaler that replenishes the `hysteresis` whenever a non-overflowing iteration is encountered. This is useful for training runs that periodically face instabilities or bad samples.

This PR exposes this existing feature to users via a new config option `consecutive_hysteresis`, which is added to the fp16 dict like:

```
"fp16": {
    "enabled": true,
    "loss_scale": 0,
    "loss_scale_window": 1000,
    "hysteresis": 2,
    "consecutive_hysteresis": true,
    "min_loss_scale": 1
}
```

So if `"consecutive_hysteresis": true`, `"hysteresis: 2"`, and every 2k iterations (outside the loss window) there's a single loss overflow, we'd restore the `hysteresis` to 2 after each overflow and won't update the loss scale. Alternatively, if `"consecutive_hysteresis": false`, we'd update the loss scale after every `hysteresis=2` loss overflows.

### DIFF
--- a/deepspeed/runtime/config.py
+++ b/deepspeed/runtime/config.py
@@ -18,6 +18,7 @@ from .fp16.loss_scaler import (
     INITIAL_LOSS_SCALE,
     SCALE_WINDOW,
     DELAYED_SHIFT,
+    CONSECUTIVE_HYSTERESIS,
     MIN_LOSS_SCALE,
 )
 from .config_utils import (
@@ -217,6 +218,7 @@ def get_dynamic_loss_scale_args(param_dict):
                 INITIAL_LOSS_SCALE: 2**init_scale,
                 SCALE_WINDOW: scale_window,
                 DELAYED_SHIFT: delayed_shift,
+                CONSECUTIVE_HYSTERESIS: consecutive_hysteresis,
                 MIN_LOSS_SCALE: min_loss_scale,
             }
 
--- a/deepspeed/runtime/config.py
+++ b/deepspeed/runtime/config.py
@@ -210,7 +210,7 @@ def get_dynamic_loss_scale_args(param_dict):
             init_scale = get_scalar_param(fp16_dict, FP16_INITIAL_SCALE_POWER, FP16_INITIAL_SCALE_POWER_DEFAULT)
             scale_window = get_scalar_param(fp16_dict, FP16_LOSS_SCALE_WINDOW, FP16_LOSS_SCALE_WINDOW_DEFAULT)
             delayed_shift = get_scalar_param(fp16_dict, FP16_HYSTERESIS, FP16_HYSTERESIS_DEFAULT)
-            is_consecutive = get_scalar_param(fp16_dict, FP16_CONSECUTIVE_HYSTERESIS, FP16_CONSECUTIVE_HYSTERESIS_DEFAULT)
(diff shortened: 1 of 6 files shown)


deepspeed/runtime/fp16/loss_scaler.py
```diff
@@ -190,6 +191,8 @@ def update_scale(self, overflow):
             self.last_overflow_iter = self.cur_iter
         else:
             if self.consecutive_hysteresis:
+                hysteresis_msg = f"Consecutive hysteresis is enabled. Restoring hysteresis to {self.delayed_shift}"
+                logger.info(hysteresis_msg)
```
jeffra, line 195: Can you wrap the logger.info around an `if dist.get_rank() == 0:` so we only see one log instead of `world_size`?

deepspeed/runtime/config.py
```diff
@@ -204,11 +204,14 @@ def get_dynamic_loss_scale_args(param_dict):
...
         ]
         if any(arg in list(fp16_dict.keys()) for arg in dynamic_loss_args):
             init_scale = get_scalar_param(fp16_dict, FP16_INITIAL_SCALE_POWER, FP16_INITIAL_SCALE_POWER_DEFAULT)
             scale_window = get_scalar_param(fp16_dict, FP16_LOSS_SCALE_WINDOW, FP16_LOSS_SCALE_WINDOW_DEFAULT)
             delayed_shift = get_scalar_param(fp16_dict, FP16_HYSTERESIS, FP16_HYSTERESIS_DEFAULT)
+            consecutive_hysteresis = get_scalar_param(fp16_dict, FP16_CONSECUTIVE_HYSTERESIS,
+                                                      FP16_CONSECUTIVE_HYSTERESIS_DEFAULT)
             min_loss_scale = get_scalar_param(fp16_dict, FP16_MIN_LOSS_SCALE, FP16_MIN_LOSS_SCALE_DEFAULT)
             loss_scale_args = {
                 INITIAL_LOSS_SCALE: 2**init_scale,
```
jeffra, line 217: I don't think `consecutive_hysteresis` is being used currently in the PR. Looking at how the code is used w. ZeRO stage 1 and 2 at least, I think we need to pass `consecutive_hysteresis` to the `loss_scale_args`. I could be wrong though, it's been a bit since I've looked at this part of the code.

deepspeed/runtime/fp16/loss_scaler.py
Quentin-Anthony: Resolved in https://github.com/microsoft/DeepSpeed/pull/3553/commits/d3a48460422ba916b1ae21f3c392274ff3d6991d

deepspeed/runtime/config.py
Quentin-Anthony: Ah that was a big oversight on my part. It's kind of hard to test this feature so this slipped through the cracks.

I think it's resolved in https://github.com/microsoft/DeepSpeed/pull/3553/commits/156ef6c120528347ac2b096f816c2dc6e54cd734. Please double check me.

jeffra (approved):

PR #3579, opened 2023-05-19
[profiling]add show_straggler argument to log_summary()

delock:
In tensor parallel inference straggler effect is one of the factor that impacts scaling efficiency.  Between any two allreduce of tensor parallel, one worker may run slower than other workers, this worker become the straggler for this duration.  Different duration (between two allreduces) might have different straggler.

When do profiling, Straggler effect usually show as part of communication latency.  This is because faster worker needs to wait for straggler during communication, this usually cause long communicatio latency.  Part of communication latency is waiting for straggler to start, the rest is the time real communication takes.

It would helpful to quantify straggler effect in a workload, this PR extends communciation logging by add an optional argument to `log_summary`:
```
log_summary(show_straggler=True)
```
When `show_straggler` is set to True, communication logging would show a new table showing communciation latency and straggler effect.  Communication latency is defined as the duration between the time when all ranks started communication and the communication finished.   Straggler effect is define as the time between current rank started communication and all ranks started communication.

Below is an example:
```
Comm. Op            Message Size        Count               Total Latency(ms)   Avg Latency(ms)     tput_avg (Gbps)     busbw_avg (Gbps)
broadcast
                    4.94 KB             1                   0.22                0.22                0.18                0.18
all_reduce
                    32.0 KB             3472                1281.64             0.34                1.71                0.86
                    1.0 MB              112                 178.83              1.51                11.17               5.58
log_summary_barrier
                    0B                  1                   1.33                1.33                0.00                0.00
_______________________________
Breakdown with straggler effect
-------------------------------
Comm. Op            Message Size        Count               Total comm lat(ms)  Total straggler(ms) Avg comm lat(ms)    Avg straggler(ms)
broadcast
                    4.94 KB             1                   0.22                0.00                0.22                0.00
all_reduce
                    4.0 B               1                   0.10                0.00                0.10                0.00
                    32.0 KB             3472                966.38              315.26              0.24                0.08
                    1.0 MB              112                 178.52              0.32                1.51                0.00
log_summary_barrier
                    0B                  1                   0.12                1.21                0.12                1.21
```
The column 'Total straggler(ms)' show the straggler time.  In this case, for 32KB allreduce, real communciation latency is 966.38ms and straggler effect is 315.26ms.  Which shows significant straggler effect in the workload.

In this PR, straggler effect is defined as:
```
sum(latency)-sum(allreduce(latency, OP=MIN))
```


deepspeed/utils/comms_logging.py
```diff
@@ -139,6 +144,34 @@ def log_all(self):
...
+            for record_name in self.comms_dict.keys():
+                if print_log:
+                    print(record_name)
+                for msg_size, vals in sorted(self.comms_dict[record_name].items()):
+                    # vals[0] is the count for each msg size
+                    count = vals[0]
+                    # vals[1] is a list of latency records for each msg size
+                    lats = torch.tensor(vals[1])
+                    min_lats = torch.tensor(vals[1])
+                    dist.all_reduce(min_lats, op=ReduceOp.MIN)
```
tjruwase, line 169: I am not convinced that this logic will compute straggler as defined. Since we are logging from rank 0 perspective, it seems this will report the difference between rank 0 and the fastest rank (i.e., with minimum latency). This will be fine if rank 0 latency is assumed to be greater or equal to other ranks. Is it okay to make this assumption?

docs/_tutorials/comms-logging.md
```diff
@@ -114,3 +114,14 @@ broadcast | [Caller Func: _broadcast_model]
 reduce_scatter_tensor | [Caller Func: reduce_scatter_fn]
                     678.86 MB           80                  1527.17             13.94               1211.75             1136.01
 ```
+
+Straggler effect can be shown by supply optional argument `show_straggler=True` to `deepspeed.comm.log_summary()` call.   Straggler effect is defined as the time a rank waits for the slowest rank to start communication.  For each collective, `log_summary` would get the minimum collective time among all ranks, compute straggler effect as follows:
```
tjruwase, line 118: supply -> supplying

delock: Will fix, thanks!

deepspeed/utils/comms_logging.py
delock: Straggler effect is the additional overhead introduced from syncing between different ranks.  Take a simple example, suppose a work load has two stages, each stage take 1s on average and has 0.3s average variance.

Assume there are two ranks:
Rank 0:
    stage 1: 0.8s
    stage 2: 1.4s
Total time = 2.2s
Rank 1:
    stage 1: 1.2s
    stage 2: 0.6s
Total time = 1.8s

If there is no syncing between two ranks, rank 0 will take 2.2s and rank 1 will take 1.8s.  If we run the workload for several times, we will observe rank 0 average runtime is 2.0s and rank 1 average runtime is 2.0s.  This is the base line we use to measure straggler effect.

If there is syncing between two ranks, assume allreduce take 0.1s, then at the end of each stage the fast rank will wait for slow rank.  So we will observe the following stage time of each rank:
Rank 0:
    stage 1: 1.3s --> allreduce time = 0.5s  = 0.4s (wait for another rank) + 0.1s (allreduce time itself)
    stage 2: 1.5s  --> allreduce time = 0.1s (allreduce time itself)
Total time = 2.8s, 0.6s longer than without syncing = 0.4s spending on waiting other rank + 0.2s allreduce time
Rank 1:
    stage 1: 1.3s --> allreduce time = 0.1s (allreduce time itself)
    stage 2: 1.5s --> allreduce time = 0.9s = 0.8s (wait for another rank) + 0.1s (allreduce time itself)
Total time = 2.8s, 1.0s longer than without synching = 0.8s spending on waiting other rank + 0.2s allreduce time

We can see that for each rank, the additional time spent when compared with no synching is the time waiting for other rank plus allreduce time itself.  The time waiting for other rank is the straggler effect of that rank.  If we run this setting for many times and observe a specific rank, average straggler effect will be 0.6s.

So its okay if rank 0 has smaller latency than other rank, in this specific allreduce, the straggler effect of rank 0 is zero, but in other allreduce, rank 0 may still has longer latency than some other ranks.  Straggler effect of rank 0 will reflect how much duration rank 0 spending on wait for other ranks, and this will statistically represent the straggler effect of other ranks.

Note straggler effect assume all accelerator has same performance but with variance/jitting introduced by random event such as cache miss, TLB miss, OS scheduling etc.   So statistically rank0 straggler effect can be used to study straggler effect of other ranks.

tjruwase: Thanks for the detailed explanation. I think it is okay to measure straggler effect from perspective of rank 0. I was just thinking that it is possible to get largest straggler effect across the ranks by doing an extra allreduce. Something like 
```python
  local_straggeler_effect = sum(latency)-sum(allreduce(latency, OP=MIN))
  global_straggeler_effect = allreduce(local_straggler_effect, OP=MAX)
```
But I agree with your point that `local_stragger_effect` is sufficient over many iterations. I am fine with this now. Thanks!

tjruwase (approved):

PR #3596, opened 2023-05-23
Fix op_builder against PyTorch nightly

malfet:
After https://github.com/pytorch/pytorch/pull/100557 PyTorch is free to use C++17 feature in it's headers, such as `std::optional`, `if constexpr` etc.

This update language standard for DS extensions to use C++17 standard by default

### DIFF
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -673,8 +673,8 @@ class CUDAOpBuilder(OpBuilder):
             cuda_major, _ = installed_cuda_version()
             args += [
                 '-allow-unsupported-compiler' if sys.platform == "win32" else '', '--use_fast_math',
-                '-std=c++17' if cuda_major > 10 else '-std=c++14',
-                '-U__CUDA_NO_HALF_OPERATORS__', '-U__CUDA_NO_HALF_CONVERSIONS__', '-U__CUDA_NO_HALF2_OPERATORS__'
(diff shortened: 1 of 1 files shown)


stas00:
For context please see the breakage reported when trying to build deepspeed with pt-nightly as of last week:
https://github.com/pytorch/pytorch/issues/102082

stas00:
@malfet, will your changes work with older pytorch versions? if I'm not mistaken deepspeed needs to work with pt-1.9 or higher.

malfet:
> @malfet, would your proposed changes work with older pytorch versions? if I'm not mistaken deepspeed needs to work with pt-1.9 or higher.

@stas00 I have not checked with pt-1.9, but I don't see why not (it should clearly work with 1.13/2.0), as C++17 is backward compatible with C++14. It might print a few warnings (for example because older PyTorch builds use [`std::iterator`](https://en.cppreference.com/w/cpp/iterator/iterator)

stas00:
@jeffra

malfet:
Looking into formatting failure, but will not push a change until other signals have finished.

cmikeh2 (approved):

jeffra (approved):
Thank you @malfet! 🚀

fecet:
Still cannot build, could you please provide the command and environment while building?

dainey94:
yes sir !! Thanks Sir

Tải xuống Outlook dành cho Android<https://aka.ms/AAb9ysg>
________________________________

PR #3642, opened 2023-05-31
fix ccl_backend and residual_add problems

dc3671:
This PR fix two problems:
- Currently, when ccl_backend is set but not supported, it will fallback to TorchBackend. But in the `__init__` function of CCLBackend, it will first initialize TorchBackend and then check whether it should fallback.
  - This will wrongly initialize torch distributed environment before fallback.
  - And the `RANK` and `WORLD_SIZE` env may not exist, which will raise error.
- About residual_add op, it contains two kernel. If only one kernel, like _vector_add, raised exception, the other will also be set to None.
  - So I split it into two try-block.

@delock please take a look~

### DIFF
--- a/deepspeed/comm/ccl.py
+++ b/deepspeed/comm/ccl.py
@@ -24,6 +24,11 @@ def build_ccl_op():
 class CCLBackend(TorchBackend):
 
     def __init__(self, name='ccl', rank=-1, world_size=-1, mpu=None, timeout=None, init_method=None):
+        self.ccl_comm_op = build_ccl_op()
+        if self.ccl_comm_op == None:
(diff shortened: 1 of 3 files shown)


delock:
@dc3671 What is the case that one kernel may raise exception but the other may not?

dc3671:
> @dc3671 What is the case that one kernel may raise exception but the other may not?

When use idex and only one is implemented.

dc3671:
@jeffra @cmikeh2 Could you please help review this PR? Thanks

delock:
Hi @jeffra can this PR be reviewed?
This PR should also fix https://github.com/microsoft/DeepSpeed/issues/3660

dc3671:
@loadams Hi, CI passed. Can you approve this, plz?

PR #3683, opened 2023-06-05
[MoE] remove deepcopy to avoid incompatible torch.jit.script modules.

awan-10:
> **Warning**
> This is a breaking change for the MoE API, instead of passing an nn.Module representing an `expert` we now instead require passing a function that creates the expert. This gets around the deepcopy issues discussed below.

Some user modules can use torch.jit.script for their MLP/experts which will lead to compatibility problems. See this for more details; https://github.com/pytorch/pytorch/issues/18106

This PR tries to remove deepcopy with a slight change to the DS MoE API.

### DIFF
--- a/deepspeed/moe/experts.py
+++ b/deepspeed/moe/experts.py
@@ -4,7 +4,6 @@
 # DeepSpeed Team
 
 import torch
-import copy
 
 
 class Experts(torch.nn.Module):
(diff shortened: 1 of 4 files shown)


jeffra (approved):

jeffra:
Not necessarily in the end, closing

PR #3709, opened 2023-06-08
DeepSpeed overview in Japanese 

conglongli:


tohtana (approved):
This looks good, thank you!

PR #3728, opened 2023-06-09
remove UtilsBuilder load, use torch (un)flatten ops

inkcherry:
Torch (un)flatten currently has no performance issues, we have tested it on the CUDA device. Skip compiling unnecessary util.

It will not depend on the deepspeed compile process for many training processes. This may also be friendly to other device integration.

### DIFF
--- a/deepspeed/runtime/bf16_optimizer.py
+++ b/deepspeed/runtime/bf16_optimizer.py
@@ -7,9 +7,9 @@ from collections import OrderedDict
 import torch
 import sys
 import os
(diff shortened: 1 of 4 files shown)


jeffra:
Thank you @inkcherry! This is a great reminder, I see that torch has updated their (un)flatten implementations https://github.com/pytorch/pytorch/pull/58006. I also did a benchmark just now and see no difference between versions.

In this PR can you also remove: `op_builder/utils.py`

jeffra (approved):

Issue #3645, opened 2023-05-31
[BUG] killing (ctrl-C) trainer on single-node multi-gpu has sigkill_hanlder error


sei-amellinger:
**Describe the bug**
During training on a single node multi-gpu configuration the sigkill_handler generates this error:

```
[2023-05-30 20:29:14,854] [INFO] [launch.py:428:sigkill_handler] Killing subprocess 19380
    main()
  File "/opt/conda/lib/python3.8/site-packages/deepspeed/launcher/runner.py", line 556, in main
    result.wait()
  File "/opt/conda/lib/python3.8/subprocess.py", line 1083, in wait
    return self._wait(timeout=timeout)
  File "/opt/conda/lib/python3.8/subprocess.py", line 1806, in _wait
    (pid, sts) = self._try_wait(0)
  File "/opt/conda/lib/python3.8/subprocess.py", line 1764, in _try_wait
    (pid, sts) = os.waitpid(self.pid, wait_flags)
  File "/opt/conda/lib/python3.8/site-packages/deepspeed/launcher/runner.py", line 548, in sigkill_handler
    result_kill = subprocess.Popen(kill_cmd, env=env)
NameError: free variable 'kill_cmd' referenced before assignment in enclosing scope
```

It is looks like that on this line  (https://github.com/microsoft/DeepSpeed/blob/49a73549b91cc0a589d3a44e59dd4687e774d296/deepspeed/launcher/runner.py#L551) where kill_cmd is assigned will not be called when multi_node_exec is false but PDSH_LAUNCHER is true causing the signal handler to be installed but the kill_cmd to not be initialized.

**To Reproduce**
1. start deepspeed training something using "deepspeed <script>" on a single-node multi-gpu configuration.
2. Ctrl-c while actively training
3. See sigkill trace

**Expected behavior**
That the sigkill handler should handle the signal properly and bring things down.

**ds_report output**
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
[...]

PR #3730, opened 2023-06-09
Single Node is using unreferenced pdsh kill cmd while terminating

abhilash1910:
Motivation for this PR:
Issue: #3645 (https://github.com/microsoft/DeepSpeed/issues/3645)
While terminating, the process flows to PDSH  (since it is default launcher in args) and since PDSH is not initialized there is no kill cmd . This causes the incorrect reference to kill_cmd mentioned in the issue. The resolution is to fallback to result.wait() and then call SIGTERM (from parent to child as by default) by adding a condition that  kill_cmd only be called when pdsh launcher is actually used.  This is getting reproduced in scaleup runs while terminating . 
@loadams @mrwyattii @tjruwase requesting review.
(PS:CLA is signed)


jeffra (approved):
thank you @abhilash1910!

PR #3731, opened 2023-06-09
FP8 unittest for H100

jomayeri:


jeffra (approved):
Thanks @jomayeri, this is great!

PR #3736, opened 2023-06-12
fix hybrid engine mlp module

tensor-tang:


tensor-tang:

PR #3739, opened 2023-06-12
Zero3 Fix allreduce optimization for extra large tensor

hablb:
Grad tensors that don't fit in the bucket flat buffer are not added to it, but still added to params_in_ipg_bucket if such tensors exists use reduce_scatter of params_in_ipg_bucket instead of allreduce. since allreduce assumes all grads are in ipg_bucket_flat_buffer.

Add test for reduce scatter=false
Fix padding to zeros instead of undefined values


tjruwase:
@hablb, thanks for this contribution. To help the review process, can you please split into separate PRs for Profiler and Zero changes?

hablb:
Profiler changes were not mine. some kind of merge issue.
I created a new PR
https://github.com/microsoft/DeepSpeed/pull/3832

PR #3748, opened 2023-06-14
DeepSpeed-Triton for Inference

stephen-youn:
This PR introduces the Triton to DeepSpeed ([README](https://github.com/microsoft/DeepSpeed/blob/1eadebdf5bbdb0f59cedc69976be62987f9232ae/blogs/deepspeed-triton/README.md))
- We have integrated [Triton](https://github.com/openai/triton), an open source compiler for GPU programming, into DeepSpeed, which further boosts the inference speed of BERT-like models in float16 precision.
- It supports BERT-type model inference and providing 1.14~1.68x speedup for different models and GPUs.


blogs/deepspeed-triton/README.md
```diff
@@ -0,0 +1,96 @@
...
+                                        max_out_tokens=pipe.tokenizer.model_max_length)
+```
+
+
+## Running BERT inference with Triton kernels
+
+We use an example of Bert-base here.
+
+```python
+pip install deepspeed
```
jeffra, line 67: I think users will have to now do: `pip install deepspeed[triton]` so that the triton requirements file will be used. Similar change below on the other pip install command.

blogs/deepspeed-triton/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+# NOTE
+<!-- **_NOTE:_** -->
+* For more information on how to use DeepSpeed, please visit our [GitHub Page](https://github.com/microsoft/DeepSpeedExamples) and our [website](https://www.deepspeed.ai/), where you can find blog posts, tutorials, and documentation.
+
+* This feature is currently only supported for BERT, Roberta and other BERT-like models, and not for text-generation models yet.
+
+* To achieve the best performance with Triton optimization, you need to activate CUDA graph and ‘triton_autotune’ in the DeepSpeed config. CUDA graph prevents the overhead of JIT compilation and a deep call stack in Triton. ‘triton_autotune’ executes an initial step to find the most suitable parameters for Triton kernels, which may take some time.
+
+* We used [the latest Triton release](https://pypi.org/project/triton/2.0.0.post1/) in our experiments.
```
jeffra, line 94: "latest" will change over time, can we say that we're using the 2.0.0.post1 release explicitly here?

jeffra (approved):

cmikeh2 (approved):

deepspeed/inference/config.py
```diff
@@ -152,6 +153,18 @@ class DeepSpeedInferenceConfig(DeepSpeedConfigModel):
     can run faster using the graph replay method.
     """
 
+    use_triton: bool = False
+    """
+    Use this flag to use triton kernels for inference ops.
+    """
+
+    triton_autotune: bool = False
```
RezaYazdaniAminabadi, line 161: any reason why this should be false by default?

blogs/deepspeed-triton/README.md
```diff
@@ -0,0 +1,96 @@
...
+# 2. How to use Triton in Deepspeed
+
+You can enable Triton compilers to optimize these kernels by setting a flag in the DeepSpeed config file.
+
+```
+pipe = pipeline('fill-mask', model='bert-base-cased', framework='pt', device=0)
+pipe.model = deepspeed.init_inference(pipe.model,
+                                        dtype=torch.float16,
+                                        replace_with_kernel_inject=True,
+                                        replace_method='auto',
```
awan-10, line 54: remove the "replace_method" argument. It is not needed anymore

awan-10 (approved):

deepspeed/inference/config.py
stephen-youn: if turned on, users have to pay for the large cost for the initial triton autotune.

PR #3600, opened 2023-05-24
fix opt-350m shard loading issue in AutoTP

sywangyi:

PR #3749, opened 2023-06-14
fix autoTP checkpoint load for OPT models

dc3671:
For OPT-30B or other OPT series model, state_dict keys is: `(['decoder.embed_tokens.wei ght', 'decoder.embed_positions.weight', 'decoder.project_out.weight', 'decoder.project_in.weight', 'decoder.layers.0.self_attn.k_proj.weight', ' decoder.layers.0.self_attn.k_proj.bias'`

While the model start with a extra level0 `model`:
image

So we need to skip the level0 when load checkpoint for OPT series models like BLOOM series models.

Please kindly review @tjruwase @jeffra  cc @sywangyi @delock

### DIFF
--- a/deepspeed/module_inject/replace_module.py
+++ b/deepspeed/module_inject/replace_module.py
@@ -801,7 +801,7 @@ def skip_level_0_prefix(model, name):
         key = re.search(r": (.*?)Stack", model)
     if key is None:
         key = re.match(r"(.*?)Model", model)
-    if key is not None and key.group(1).lower() in "bloom":
+    if key is not None and key.group(1).lower() in ["bloom", "opt"]:
         # if keys start with 'model.', don't skip level 0 prefix
         if not re.match("^model[.]", name):
             return True


sywangyi:
Hi, I have already upload PR for it https://github.com/microsoft/DeepSpeed/pull/3600

sywangyi:
no merged yet. you could try the fix.

dc3671:
> no merged yet. you could try the fix.

Ok, I'll try it. My PR can be closed if yours is merged.

sywangyi:
> > no merged yet. you could try the fix.
> 
> Ok, I'll try it. My PR can be closed if yours is merged.

facebook/opt-350m and facebook/opt-125m key name is different. that's why we did not catch the failure in ci, you change is only for opt-350m, but will lead to regression to opt-125m

dc3671:
duplicated with https://github.com/microsoft/DeepSpeed/pull/3600 .

PR #3756, opened 2023-06-15
fix the bug of save model in the bf16+zero1+pp mode

L-hongbin:
Fix the bug " AttributeError: 'BF16_Optimizer' object has no attribute 'bit16_groups' " when using bf16 + zero1 + pp to train model and saving the bf16 optimizer state


jomayeri (approved):

nomadlx:
Why was this closed? What's wrong with this fix?

PR #3759, opened 2023-06-16
fix the bug of save bf16 optimizer state in the bf16+zero1+pp mode

L-hongbin:
Fix the bug " AttributeError: 'BF16_Optimizer' object has no attribute 'bit16_groups' " when using bf16 + zero1 + pp to train model and saving the bf16 optimizer state

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -3165,7 +3165,8 @@ class DeepSpeedEngine(Module):
         # if we don't use it, we get parameters ordered incorrectly
         if hasattr(self.optimizer, "round_robin_bit16_groups"):
             bit16_groups = self.optimizer.round_robin_bit16_groups
-        elif self.bfloat16_enabled() and not self.zero_optimization():
+        elif self.bfloat16_enabled() and (not self.zero_optimization() or self.zero_optimization_stage() == 1 
+            and hasattr(self.optimizer, 'bf16_groups')):
             bit16_groups = self.optimizer.bf16_groups
         else:
             bit16_groups = self.optimizer.bit16_groups if self.zero_optimization_stage(
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -3165,8 +3165,9 @@ class DeepSpeedEngine(Module):
         # if we don't use it, we get parameters ordered incorrectly
         if hasattr(self.optimizer, "round_robin_bit16_groups"):
             bit16_groups = self.optimizer.round_robin_bit16_groups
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/engine.py
```diff
@@ -3165,7 +3165,9 @@ def _get_zero_param_shapes(self):
         # if we don't use it, we get parameters ordered incorrectly
         if hasattr(self.optimizer, "round_robin_bit16_groups"):
             bit16_groups = self.optimizer.round_robin_bit16_groups
-        elif self.bfloat16_enabled() and not self.zero_optimization():
+        elif self.bfloat16_enabled() and (not self.zero_optimization() or
```
tjruwase, line 3168:
```suggestion
        elif self.bfloat16_enabled() and hasattr(self.optimizer, "bf16_groups")
```

tjruwase: Can you try this simplication?

inkcherry:
I have also encountered this problem， "In the two configurations below, both will opt for the 'BF16optimizer' in the selection logic of the optimizer. Despite the differing 'config' settings, they follow the same execution path with bf16 + zero1 + pp. However, when employing the former configuration, self.zero_optimization() is set to 1, whereas in the latter configuration, it's set to 0." If  attempt to save optimizer states using the first configuration,  will encounter this issue.

personally I think configuration 2 may be somewhat misleading.
1. 
----------------------
 "zero_optimization": {
    "stage": 1
  },
  "bf16": {"enabled": true},
  "fp16": {"enabled": false},
  "data_types": {
    "grad_accum_dtype": "fp32"
  }
----------------------
2. 
----------------------
  "zero_optimization": {
    "stage": 0
  },
  "bf16": {"enabled": true},
  "fp16": {"enabled": false},
  "wall_clock_breakdown": false
-----------------------

tjruwase:
@inkcherry, your analysis is correct. And yes, configuration 2 is misleading. But we use `bf16_optimizer` in this case for historical reasons from the bloom176b training and because we have not written a non-zero optimizer wrapper that works for bf16 training.

deepspeed/runtime/engine.py
tjruwase: @L-hongbin, does this suggestion not work for you?

zte-tcb:
I have the same error. I want to know that if I use this code to modify DeepSpeed now, Can it work correctly at BF16?

tjruwase:
@L-hongbin, are you still working on this PR?

tjruwase:
@inkcherry, are you still interested in this PR. It seems @L-hongbin is no longer interested, so I want to close. Thanks!

PR #3761, opened 2023-06-16
Update vae.py

mzamini92:
Since the `DSVAE` class is already inheriting from `torch.nn.Module`, there is no need to inherit from `CUDAGraph` as well. You can remove the `CUDAGraph` inheritance. Instead of using `self.vae.requires_grad_(requires_grad=False)`, you can use `torch.no_grad()` context manager during initialization to disable gradient computation for the `self.vae` module.  The `_graph_replay_decoder`, `_graph_replay_encoder`, and `_graph_replay` methods can benefit from the `@torch.no_grad()` decorator.


mzamini92:
> ```
> @microsoft-github-policy-service agree
> ```

loadams:
@mzamini92 - since this PR made it quite a while before being lost, do you think it is still worth merging, if so could you resolve the merge conflicts?  Or should we close this PR?

PR #3790, opened 2023-06-22
[zero] revert PR #3166, it disabled grad clip for bf16

jeffra:
#3166 inadvertently disabled gradient clipping when using bf16. The original intention was to remove grad scaling when using bf16, however when using bf16 the loss scale is forced to 1.0 which will result in no grad scaling occurring.


guoyejun:
thanks, good catch, yes, this is the grad clip to avoid exploding, no matter which format is used. It is independent with loss scale of fp16.

tjruwase (approved):

PR #3806, opened 2023-06-25
zero/mics.py: use on_accelerator instead of cuda only

guoyejun:


delock:
@guoyejun can you also update https://github.com/microsoft/DeepSpeed/blob/master/scripts/check-torchcuda.py to catch is_cuda usage in the code?

guoyejun:
@delock  sure, it can be in another separate PR

PR #3595, opened 2023-05-23
Add Ascend NPU accelerator support

CurryRice233:
Add Ascend NPU accelerator support

PR #3831, opened 2023-06-28
Make Ascend NPU available

hipudding:
NPU accelerator support is introduced in (#3595).
This commit provides two enhancements:
  1. Add a new accelerator_name 'npu' for choosing, it can be specified by environment variable or auto detected.
  2. Optimize auto detect code in get_accelerator to avoid too many layers of exception throwing.


hipudding:

accelerator/real_accelerator.py
```diff
@@ -65,34 +65,53 @@ def get_accelerator():
...
+            except ImportError as e:
+                raise ValueError(
+                    f'NPU_Accelerator requires torch_npu, which is not installed on this system.')
+            pass
         elif accelerator_name == 'cuda':
             pass
         else:
             raise ValueError(
-                f'DS_ACCELERATOR must be one of "cuda", "cpu", or "xpu".  Value "{accelerator_name}" is not supported')
+                f'DS_ACCELERATOR must be one of "cuda", "cpu", "xpu", or "npu".  Value "{accelerator_name}" is not supported')
```
tjruwase, line 79: Could we introduce a global list of accelerator names and use that here?
I mean something like the following

```python
DS_ACCELERATOR_LIST = ['cuda', 'cpu', ...]

...
raiseValueError(f'DS_ACCELERATOR environment variable must be one of {DS_ACCELERATOR_LIST} ...'
```

hipudding: Sure. Will change to use global list of accelerators next push.

hipudding:
@tjruwase Good day. Could you please review this PR again and trigger workflows? Thanks.

tjruwase:
@hipudding, thanks for this PR. We are reviewing currently. Apologies for the delay.

hipudding:
:-) Thanks.

hipudding:
1. Rebase my commits to master HEAD to avoid too many merge back from master.
2. Change commit title to remove PR number because it will added automatically when PR merged.
3. No code changed.

hipudding:
It  seems the failing test  case is not related to these commits. Please re-trigger this failed workflow, Thanks.

accelerator/cpu_accelerator.py
```diff
@@ -234,6 +234,9 @@ def on_accelerator(self, tensor):
 
     # create an instance of op builder and return, name specified by class_name
     def create_op_builder(self, op_name):
+        # avoid abstract classes
```
tjruwase, line 237: @hipudding, can you please explain these changes? 

@delock, FYI

accelerator/cuda_accelerator.py
```diff
@@ -234,8 +234,9 @@ def _lazy_init_class_dict(self):
             op_builder_dir = self.op_builder_dir()
             op_builder_module = importlib.import_module(op_builder_dir)
             for _, module_name, _ in pkgutil.iter_modules([os.path.dirname(op_builder_module.__file__)]):
-                # avoid self references
-                if module_name != 'all_ops' and module_name != 'builder' and module_name != 'cpu':
+                # avoid self references and skip other backends
+                if module_name != 'all_ops' and module_name != 'builder' and module_name != 'no_impl' \
+                    and module_name != 'cpu' and module_name != 'npu':
```
tjruwase, line 239:
```suggestion
                if module_name != 'all_ops' and module_name != 'builder' and module_name not in [no_impl', 'cpu', 'npu']:
```

accelerator/cpu_accelerator.py
hipudding: Sure, of course.

I moved no_impl class out of cpu folder as a public class for every backend. If any backend didn't have any ops, ds_report will not print any op (including NotImplementedBuilder, because the op_builder/{new backend} folder is empty, ds_report can't find any modules). For cpu accelerator, if any ops is not implement, it will fallback to NotImplementedBuilder, but it didn't skip the abstract classes(not regular ops), which means if ds_report scanning module named OpBuilder or CPUOpBuilder, it will still return the fallback op(NotImplementedBuilder), and it will print NotImplementedBuilder in ds_report result.

1. Considering cuda_accelerator skip abstract class, I make them behave the same. See cuda_acclerator._lazy_init_class_dict
2. NotImplementedBuilder is not considered as a normal op any more, it should not show in ds_report.
3. Make every backend behave the same no matter it has any op_builders or not.

accelerator/cuda_accelerator.py
hipudding: How about make all module_names we want to exclude into a list?

accelerator/cpu_accelerator.py
delock: @tjruwase @hipudding  I'm more favor to decouple accelerator backends than reuse between accelerator backends.   Since it's hard for a developer to test on all the backends, decouple would avoid cross interfere between accelerator changes.  I I think its better each accelerator backend implement its own NotImplementedBuilder, even the logic is almost the same.

accelerator/cuda_accelerator.py
delock: I remember `cpu` is included because `cpu/` is a subdirectory of op_builder and get reflected.   I think rather than make a list to include `cpu` and `npu`, there are two options:
1. Exclude module names that is _directory_, so subdirectory created for other accelerators would be excluded.
2. Move CUDA op_builder under `cuda/` subdirectory and reflect op builder from that subdirectory, this would also make directory structure more clean.

accelerator/cpu_accelerator.py
hipudding: I agree woth @delock, I will revert the last commit and leave a NotImplementBuilder for NPU.

accelerator/cuda_accelerator.py
hipudding: I used to think about Option 1. But exclude all directory may exclude directories that contains ops. For example, cuda may add a group of similar ops and put them in a directory. 

And Move CUDA op_builder under cuda/ maybe a better choice, I didn't do this to avoid too may code change. 
I will use Option2 for next commit.

hipudding: After thinking about these two options, I want to choose Option1.

accelerator/cpu_accelerator.py
hipudding: Fixed according to @delock 's suggestion.

op_builder/npu/no_impl.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+from .builder import NPUOpBuilder
+
+
+class NPUNotImplementedBuilder(NPUOpBuilder):
```
delock, line 9: I think both `NPUNotImplementedBuilder` and `CPUNotIMplemetnedBuilder` can have the same name `NotImplementedBuilder`.  The accelerator interface design allows each accelerator backend to implement op builder with same functionality with same name but different implementation.   So `NotImplementedBuilder` have same functionality and could have different implementation in CPU and NPU.   This also opens the door for DeepSpeed to check whether certain OpBuilder is implemented in the future.

hipudding: Sure. Will change to same name next commit. Thanks.

tjruwase (approved):

hipudding:
Good day. This PR is approved and ready for merge. But one workflow is failed due to OOM(not related to these changes). Could you retrigger this workflow and merge this PR? Thanks.

Issue #3837, opened 2023-06-28
[BUG] Incorrect type check in engine.py for CPU training 


learning-chip:
**Describe the bug**

The type check inside function `split_half_float_double_sparse` of [runtime/engine.py](https://github.com/microsoft/DeepSpeed/blob/v0.9.5/deepspeed/runtime/engine.py#L115-L124) does not recognize CPU tensors. Simply because it is checking for string `"torch.cpu.FloatTensor"` while the CPU tensor type is `"torch.FloatTensor"` (no `"cpu"`)

https://github.com/microsoft/DeepSpeed/blob/fc9e1ee00e673b2ce2a433c4e34c5440e76c9f3e/deepspeed/runtime/engine.py#L115-L124

**To Reproduce**

Enable Intel CPU backend of deepspeed:

```bash
pip install torch==1.13.1+cpu
pip install deepspeed==0.9.5
pip install intel_extension_for_pytorch==1.13+cpu -f https://developer.intel.com/ipex-whl-stable-cpu
pip install oneccl_bind_pt==1.13+cpu -f https://developer.intel.com/ipex-whl-stable-cpu
# also need to build oneCCL itself
```

Modify [small_model_debugging/test_model.py](https://github.com/microsoft/DeepSpeed/blob/v0.9.5/tests/small_model_debugging/test_model.py) to work on CPU:
- set `"torch_adam": True` in `config_dict` to skip op builder, and set `"enabled": False` for `"fp16"`
- set `dtype=torch.float` for `train_data` to avoid half precision

Running the script leads to error `AssertionError: attempting to reduce an unsupported grad type: torch.FloatTensor`. Remove the assertion (line 123~124 in DS 0.9.5) then it trains fine.

**ds_report output**
Please run `ds_report` to give us details about your setup.

PR #3633, opened 2023-05-30
[CPU] Skip CPU support unimplemented error

Yejing-Lai:
This PR aims to add CPU inference UT support. We skip CPU support unimplemented errors and update cpu inference workflow. 
Skip logic:
1. Builder name does not exist or is not implemented, then skip.
2. Add an API to get the supported data types by accelerator, which can help us skip GPU fp16 UT.

PR #3842, opened 2023-06-29
Support cpu tensors without direct device invocation

abhilash1910:
Motivation:
Fix for reproducible issue #3837 on cpu. On cpus direct invocation of torch.cpu.tensor leads to dtype mismatch.
Another way would be to have something like :
["torch.DoubleTensor" if device_type == 'cpu else '"torch.{}.DoubleTensor".format(device_type)] for all elements in the supported list , but that would eliminate "torch.cpu.DoubleTensor" ,etc from the scope. 
@jeffra requesting review.
  
CLA is signed


tjruwase:
@abhilash1910, thanks for this PR. I think this PR needs some work that leverages PR #3633 for the following reasons. 
1. As you observed, strings like `torch.cpu.DoubleTensor` are invalid, and the underlying issue is that existing code is now broken with the support for accelerators other than cuda.  So, it will be better to remove code that generates this invalid strings. 
2. #3633 is adding API for querying the dtypes supported by the accelerator, and it seems to me that this entire logic could be rewritten to use that API. I think this API is less brittle since it does not rely on string formatting, and is easier to maintain since the respective accelerator owners can update their list of supported dtypes. 

Please share your thoughts. Thanks!

abhilash1910:
@tjruwase yes I think that would be a proper fix , instead of having separate dtypes, we can directly leverage abstract accelerator interface. Let me go through the changes for this . Thanks (Making this draft for now).

tjruwase:
@abhilash1910, thanks for your alignment. I will push to get #3633 merged asap. I left some comments on there.

abhilash1910:
Yes sure, I will work with my colleague Yejing to make this work.

delock:
> @abhilash1910, thanks for this PR. I think this PR needs some work that leverages PR #3633 for the following reasons.
> 
> 1. As you observed, strings like `torch.cpu.DoubleTensor` are invalid, and the underlying issue is that existing code is now broken with the support for accelerators other than cuda.  So, it will be better to remove code that generates this invalid strings.
> 2. [[CPU] Skip CPU support unimplemented error #3633](https://github.com/microsoft/DeepSpeed/pull/3633) is adding API for querying the dtypes supported by the accelerator, and it seems to me that this entire logic could be rewritten to use that API. I think this API is less brittle since it does not rely on string formatting, and is easier to maintain since the respective accelerator owners can update their list of supported dtypes.
> 
> Please share your thoughts. Thanks!

@tjruwase I have a question, for the following check, is the check focusing on the data type, or data type+device type is needed?   If only data type is important, then proper way is strip device type from t.type(), then compare with data type list.

tjruwase:
> @tjruwase I have a question. For the following check, is the check focusing on the data type, or data type+device type is needed? If only data type is important, then proper way is strip device type from t.type(), then compare with data type list.

Yes, the focus is on checking the data type supported by the device. My feedback is based on avoiding any assumptions of string format combinations of device and data type. For example as shown in this [list](https://pytorch.org/docs/1.12/tensors.html#data-types), the same dtype is formatted differently for cpu and cuda tensors.  So, I 
suggested relying on `torch dtype` which is canonical, and which has an accelerator API to retrieve. 

Please let me know your thoughts.

delock:
> 

Yes, dtype is better.  Some additional changed in _reduce_non_expert_gradients and _reduce_expert_gradients will be needed accordingly.

tjruwase:
> Yes, dtype is better. Some additional changed in _reduce_non_expert_gradients and _reduce_expert_gradients will be needed accordingly.

@delock, thanks for the pointer. @abhilash1910, could you please help handle those changes in your PR?

abhilash1910:
@tjruwase @delock _reduce_non_expert_gradients and _reduce_expert_gradients uses SparseTensor class, would it make sense to replace type with dtype there ?

tjruwase:
@abhilash1910, you raise an important issue with my proposal that I had overlooked. This is that `SparseTensor.type()` is a string while `torch.dtype` is not. My proposal would result in populating `supported_types` with objects of different types, which creates all kinds of complications. 
image

One idea is to populate the list with `[f'{t}' for t in accelerator.supported_dtypes()]` to harmonize the object types. But the problem is that while `SparseTensor.type()` will work for lookups, neither `torch.Tensor.type()` nor `torch.Tensor.dtype` will work. In other words, we will always need different lookup logic for torch tensors on one hand, and deepspeed sparse_tensors on the other hand. 

Another potential issue with existing code is that we don't check whether the dtype of the underlying [tensor](https://github.com/microsoft/DeepSpeed/blob/0a0819b78554a5130b7c1b31d18779d39421dade/deepspeed/runtime/engine.py#L2321) of a SparseTensor is supported by the accelerator. I need to double check this concern with my teammates. 

Therefore, I am now wondering it would better to take a new approach that makes this difference explicit, and uses `isinstance(t, torch.Tensor)` and `isinstance(t, SparseTensor)` appropriately to guard the lookups. I think this will make the code easier to maintain since the logic is more explicit. Also, I think is more intuitive for `split_half_float_double_sparse()` [function](https://github.com/microsoft/DeepSpeed/blob/0a0819b78554a5130b7c1b31d18779d39421dade/deepspeed/runtime/engine.py#L115) to return separate buckets for sparse and dense tensors for the callers to handle appropriately. 

These are just some thoughts. Please let me know what you think. Thanks!

deepspeed/runtime/engine.py
```diff
@@ -114,14 +114,11 @@
...
 def split_half_float_double_sparse(tensors):
     device_type = get_accelerator().device_name()
-    supported_types = [
-        "torch.{}.HalfTensor".format(device_type), "torch.{}.FloatTensor".format(device_type),
-        "torch.{}.DoubleTensor".format(device_type), "torch.{}.BFloat16Tensor".format(device_type),
-        SparseTensor.type()
-    ]
-
+    supported_types = get_accelerator().supported_dtypes()
+    supported_types.append(SparseTensor.type())
```
jeffra, line 118: @tjruwase and I just discussed offline.

Let's remove this line (supported_types.append(SparseTensor.type())), then also add an attribute in SparseTensor to stash the dense tensor dtype like this:

`self.dtype = self.orig_dense_tensor`

Inside here: 
https://github.com/microsoft/DeepSpeed/blob/0a0819b78554a5130b7c1b31d18779d39421dade/deepspeed/runtime/sparse_tensor.py#L16

abhilash1910:
> @abhilash1910, you raise an important issue with my proposal that I had overlooked. This is that `SparseTensor.type()` is a string while `torch.dtype` is not. My proposal would result in populating `supported_types` with objects of different types, which creates all kinds of complications. image
> 
> One idea is to populate the list with `[f'{t}' for t in accelerator.supported_dtypes()]` to harmonize the object types. But the problem is that while `SparseTensor.type()` will work for lookups, neither `torch.Tensor.type()` nor `torch.Tensor.dtype` will work. In other words, we will always need different lookup logic for torch tensors on one hand, and deepspeed sparse_tensors on the other hand.
> 
> Another potential issue with existing code is that we don't check whether the dtype of the underlying [tensor](https://github.com/microsoft/DeepSpeed/blob/0a0819b78554a5130b7c1b31d18779d39421dade/deepspeed/runtime/engine.py#L2321) of a SparseTensor is supported by the accelerator. I need to double check this concern with my teammates.
> 
> Therefore, I am now wondering it would better to take a new approach that makes this difference explicit, and uses `isinstance(t, torch.Tensor)` and `isinstance(t, SparseTensor)` appropriately to guard the lookups. I think this will make the code easier to maintain since the logic is more explicit. Also, I think is more intuitive for `split_half_float_double_sparse()` [function](https://github.com/microsoft/DeepSpeed/blob/0a0819b78554a5130b7c1b31d18779d39421dade/deepspeed/runtime/engine.py#L115) to return separate buckets for sparse and dense tensors for the callers to handle appropriately.
> 
> These are just some thoughts. Please let me know what you think. Thanks!

Yes my thoughts exactly.  I was thinking of adding a dtype getter inside the sparsetensor to make it consistent, @tjruwase @jeffra . Let me make the additions.

deepspeed/runtime/engine.py
```diff
@@ -114,14 +114,10 @@
...
-        "torch.{}.HalfTensor".format(device_type), "torch.{}.FloatTensor".format(device_type),
-        "torch.{}.DoubleTensor".format(device_type), "torch.{}.BFloat16Tensor".format(device_type),
-        SparseTensor.type()
-    ]
-
+    supported_types = get_accelerator().supported_dtypes()
+    
     for t in tensors:
-        assert t.type() in supported_types, f"attempting to reduce an unsupported grad type: {t.type()}"
+        assert t.dtype in supported_types, f"attempting to reduce an unsupported grad type: {t.type()}"
```
tjruwase, line 120: `t.type()` -> `t.dtype`

abhilash1910:
@tjruwase @jeffra   does _reduce_non_expert_gradients and _reduce_expert_gradients also need dtype changes , since we are adding the attribute . Also not sure why the CLA is appearing.

deepspeed/runtime/sparse_tensor.py
```diff
@@ -15,6 +15,7 @@ class SparseTensor(object):
 
     def __init__(self, dense_tensor=None):
         self.orig_dense_tensor = dense_tensor
+        self.dtype = self.orig_dense_tensor
```
tjruwase, line 18: `self.dtype = self.orig_dense_tensor.dtype`

deepspeed/runtime/engine.py
```diff
@@ -114,14 +114,10 @@
...
-        "torch.{}.HalfTensor".format(device_type), "torch.{}.FloatTensor".format(device_type),
-        "torch.{}.DoubleTensor".format(device_type), "torch.{}.BFloat16Tensor".format(device_type),
-        SparseTensor.type()
-    ]
-
+    supported_types = get_accelerator().supported_dtypes()
+    
     for t in tensors:
-        assert t.type() in supported_types, f"attempting to reduce an unsupported grad type: {t.type()}"
+        assert t.dtype in supported_types, f"attempting to reduce an unsupported grad type: {t.dtype}"
```
tjruwase, line 120: I notice that `t.type()` is still used in line 124. This wouldn't work.

abhilash1910: Could you elaborate on this? t.type() was replaced with t.dtype
thanks

tjruwase: Sure, I mean the below line 124
image

abhilash1910:
@tjruwase  could you please re-review? Thanks

deepspeed/runtime/engine.py
```diff
@@ -2324,30 +2323,32 @@ def _get_gradients_for_reduction(self):
         return non_expert_grads, expert_grads
 
     def _reduce_non_expert_gradients(self, grads, elements_per_buffer):
-        split_buckets = split_half_float_double_sparse(grads)
-        for _, bucket_tuple in enumerate(split_buckets):
-            bucket_type, bucket = bucket_tuple
-
+        split_sparse_tensor_buckets, split_dense_tensor_buckets = split_half_float_double_sparse(grads)
```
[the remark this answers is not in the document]
tjruwase: Please consider the above simplification. In particular, dense and sparse buckets can be handled seperately without need for a `zip`. Thanks!

abhilash1910:
@tjruwase  could you re-trigger CI and re-review ? Thanks

tjruwase (approved):

tjruwase:
@abhilash1910, thanks for persevering to deliver this awesome contribution.

tjruwase:
@abhilash1910, can you please use this [guide](https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites) to address the formatting issues? Thanks!

deepspeed/runtime/engine.py
```diff
@@ -118,21 +118,20 @@
...
+        assert t.dtype in supported_types, f"attempting to reduce an unsupported grad type: {t.dtype}"
 
-    buckets = []
+    sparse_tensor_buckets, dense_tensor_buckets = [], []
     for i, dtype in enumerate(supported_types):
-        bucket = [t for t in tensors if t.type() == dtype]
-        if bucket:
-            buckets.append((dtype, bucket))
-    return buckets
+        sparse_bucket = [t for t in tensors if t.dtype == SparseTensor.dtype]
```
tjruwase, line 128: CI is failing because of `SparseTensor.dtype`, since `dtype` is not a class field

tjruwase, line 128:
```suggestion
        sparse_bucket = [t for t in tensors if isinstance(t, SparseTensor)]
```

deepspeed/runtime/engine.py
```diff
@@ -118,21 +118,20 @@
...
+        dense_bucket = [t for t in tensors if t.dtype != SparseTensor.dtype]
```
tjruwase, line 129:
```suggestion
        dense_bucket = [t for t in tensors if not isinstance(t, SparseTensor)]
```

abhilash1910: Thanks I was thinking of the same,

abhilash1910:
@tjruwase  could you help re-run CI ? Thanks

tjruwase:
@abhilash1910, did you notice this CI failure?
image

deepspeed/runtime/engine.py
```diff
@@ -122,12 +122,12 @@ def split_half_float_double_sparse(tensors):
     supported_types = get_accelerator().supported_dtypes()
 
     for t in tensors:
-        assert t.dtype in supported_types, f"attempting to reduce an unsupported grad type: {t.dtype}"
+        assert t.dtype in supported_types if not isinstance(t, str), f"attempting to reduce an unsupported grad type: {t.dtype}"
```
tjruwase, line 125: @abhilash1910, do you know why a t of type str showed up here?

abhilash1910: yes quite strange, where this is coming from as the supported_dtypes (cuda/cpu) are not in string

abhilash1910:
@tjruwase  could you review once? I think this might resolve this issue ; I was not able to pinpoint the exact cause through locally, so any pointers will be helpful.

tjruwase:
> @tjruwase could you review once? I think this might resolve this issue ; I was not able to pinpoint the exact cause through locally, so any pointers will be helpful.

Were you able to repro the issue in locally?

abhilash1910:
@tjruwase  could you help re-trigger CI? Thanks

delock:
Hi @abhilash1910 can you check and fix the following error?
https://github.com/microsoft/DeepSpeed/actions/runs/6952457019/job/18941598524?pr=3842#step:8:4568
```
  File "/tmp/actions-runner/_work/DeepSpeed/DeepSpeed/unit-test-venv/lib/python3.8/site-packages/deepspeed/runtime/engine.py", line 126, in split_half_float_double_sparse
    assert t.dtype in supported_types, f"attempting to reduce an unsupported grad type: {t.dtype}"
AttributeError: 'str' object has no attribute 'dtype'
```

abhilash1910:
@tjruwase  could you retrigger CI (issue seems to be fix now)? Thanks.

abhilash1910:
@tjruwase  could you help re-trigger the CI and re-review ? Much appreciated.

delock:
Hi @abhilash1910 can you clarify whether current failures in CI is related to your PR or just a test issue?   Thanks!

abhilash1910:
@delock  I think that it might be a test issue as I am able to run the CI for the sparse test locally. I changed the pathway of code and still I see the same allclose issue.  @tjruwase  could you suggest any modifications on this? This is strange as I tested in an isolated env and did not get the issue.

delock:
Hi @abhilash1910 some suggestions:
1. provide more details (hw, sw, log ...) of your local run so there might be hint of difference.
2. try to modify the test as a standalone workload (not using pytest) so debugging could be possible
3. seperate assert for dense and sparse grad so you will know which type of tensor have difference.  This might narrow down the location of possible bug.

abhilash1910:
Thanks @inkcherry for highlighting the boundary issue ; seems it will pass the CI now. 
@tjruwase  could you help retrigger the CI ? Thanks

inkcherry:
> Thanks @inkcherry for highlighting the boundary issue ; seems it will pass the CI now. @tjruwase could you help retrigger the CI ? Thanks
 
I could reproduce the CI issue in my local env and It could passed currently. 
FYI: @delock

delock:
@tjruwase the previous error in CI workflow `unit/runtime/sparse_tensor/test_averaging_sparse_gradients.py` should be fixed by commit from @inkcherry.  Can you help restart the workflow?  Thanks!

delock:
Hi @abhilash1910 are the following two errors related to your change?
```
FAILED unit/inference/quantization/test_intX_quantization.py::TestQuantizedInt::test_zero3_int4_post_init_quant_nvme_offload
FAILED unit/inference/quantization/test_intX_quantization.py::TestQuantizedInt::test_zero3_int4_quantized_initialization_nvme_offload
```

abhilash1910:
@delock I think this failure is related to this PR , but it seems to be arising after the previous fix . I will take a look at it .

PR #3844, opened 2023-06-29
autoTP for fused qkv weight

inkcherry:
To make autoTP work with some fusedqkv model.
The qkv weight of one rank format after TP should be consistent with no TP, and the heads should be correctly divided and arranged, so that the model can correctly parse the qkv matrix when do inference.

However, some models have different qkv fuse types, and we have found three types. Now types are temporarily named based on model name.

Assuming there is num_ head=n, qx represents the weight of the n-th q head, shape=(hidden dim, hidden dim)

Bloom (huggingface):
q1, k1, v1, q2, k2, v2,..., qn, kn, vn

GLM or WebGLM (huggingface)

q1, q2, q3,.. qn, k1, k2, k3,...,kn, v1, v2, v3,..., vn

Codegen (huggingface):

q1, q2,..., q(n/4), k1, k2,..., k(n/4), v1, v2,..., v(n/4), q(n/4+1),...,q(n/2), k(n/4+1),...,k(n/2), v(n/4+1),...,v(n/2),...

At first, we attempted to infer autoTP GLM but received incorrect results, but there were no errors or warnings. We think that we can add warnings to unrecognized fused qkv for user to better troubleshoot the problem.

For other model use fused_qkv, they can also work by matching or adding fused types.


inkcherry:
update: we conducted some tests on intel GPU as follows:

- glm 350M     mp=1/2   work
- codegen 350M  mp=1/2  work
- codegen 6B      mp=1/2/4 work   
- glm     10B     mp=1/2/4/8 work
- codegen16B     mp=1/2/3/6  work

inkcherry:
@RezaYazdaniAminabadi  @molly-smith 
Could you please help review this pr?  Will appreciate your feedback.

deepspeed/module_inject/replace_module.py
```diff
@@ -398,16 +399,104 @@ def _replace(child, name, conv_linear_layer):
...
+
+            def _glm_type_transpose(input, mp_size):
+                #input : [3*hidden_dim, hidden_dim](weight) or [3*hidden_dim](bias)
+
+                shape = input.shape
+                dst_shape = shape[0] // mp_size
+                src_split = torch.split(input, shape[0] // 3, dim=0)
+
+                qkv_split = [torch.split(src_s, shape[0] // 3 // mp_size, dim=0) for src_s in src_split]
+                split_fusedqkv = [
```
abhilash1910, line 468: split_fusedqkv (cat operations) can be made as a separate method since there may be other variations using the same logic flow. @inkcherry ? 
Also changes look good, @molly-smith @RezaYazdaniAminabadi  could you please review and provide suggestions?

deepspeed/module_inject/replace_module.py
```diff
@@ -398,16 +399,104 @@ def _replace(child, name, conv_linear_layer):
...
+                            get_accelerator().current_device_name())
+                else:
+                    data = mp_replace.copy(new_weight, child.weight.data)
+                    bias_data = None if child.bias is None else mp_replace.copy(new_bias, child.bias.data).to(
+                        get_accelerator().current_device_name())
+
                 setattr(child, "replaced", True)
                 return LinearLayer(weight=data.to(get_accelerator().current_device_name()), bias=bias_data)
 
+        def require_tp_fused_qkvw(name):
```
RezaYazdaniAminabadi, line 422: @inkcherry, can we please move all these logic into a separate file, so that we can then refactor the tp-sharding logic based on this? thanks

deepspeed/module_inject/replace_module.py
RezaYazdaniAminabadi: Yes, I agree, and since we have the same logic for when using the TP for when the inference kernels are enabled then we can reuse this.

inkcherry:
Thanks for your suggestion: )  The file has been modified.
@RezaYazdaniAminabadi @tjruwase 

cc  @guoyejun @abhilash1910

RezaYazdaniAminabadi (approved):

molly-smith:
Thanks @inkcherry for this contribution. It would be nice to update the supported/unsupported model lists in docs/_tutorials/automatic-tensor-parallelism.md and maybe add one of these fused qkv weight models to a unit test. T

molly-smith (approved):

inkcherry:
> Thanks @inkcherry for this contribution. It would be nice to update the supported/unsupported model lists in docs/_tutorials/automatic-tensor-parallelism.md and maybe add one of these fused qkv weight models to a unit test.

sure, I placed it in this PR
https://github.com/microsoft/DeepSpeed/pull/4057

andre-bauer:
Bloom won't work with this since it's not taking mp_size into account

PR #3850, opened 2023-06-29
Reduce Unit Test Times (Part 3)

mrwyattii:
Continuing work from #3829 and #3838

Changes in this PR:
- Set common environment variables in shared `setup-env` workflow to reduce copied code and have more consistent test environments
- Download directory for `torchvision.datasets.CIFAR10` can now be set by an environment variable, defaults to `/tmp/`
- Added a new `reuse_dist_env` option for tests inheriting from `DistributedTest`
  - `DistributedTest` now uses `mp.pool` to persist the processes with the distributed environment between tests
  - Off by default
  - Allows us to avoid the overhead of creating and destroying lots of distributed environments for tests that run many param combinations
- `nv-torch-latest-v100`
  - Added a coverage report
  - Used coverage report (and feedback from @tjruwase and @ShadenSmith) to reduce the total number of tested parameters and make the parameters we are testing to be a more efficient use of CI resources (e.g. reducing model sizes)
- `nv-inference`
  - Added coverage report
- Time reductions:
  - `nv-torch-latest-cpu` ~8m --> ~5m
  - `nv-torch-latest-v100` ~85m --> ~45m

Changes for future PRs:
- `nv-inference`
  - Utilize `coverage` to reduce the total number of tests to a bare minimum that touches all inference code
  - Reduce number of parameterize options
- Add email notifications to the team when nightly tests fail
- Add a `coverage` report for other tests

### DIFF
--- a/.github/workflows/cpu-inference.yml
+++ b/.github/workflows/cpu-inference.yml
@@ -76,5 +76,5 @@ jobs:
           source oneCCL/build/_install/env/setvars.sh
           unset TORCH_CUDA_ARCH_LIST # only jit compile for current arch
           cd tests
-          TRANSFORMERS_CACHE=~/tmp/transformers_cache/ pytest -m 'inference' unit/inference/test_inference_config.py
-          TRANSFORMERS_CACHE=~/tmp/transformers_cache/ pytest -k TestDistAllReduce unit/comm/test_dist.py
+          TRANSFORMERS_CACHE=~/tmp/transformers_cache/ pytest $PYTEST_OPTS -m 'inference' unit/inference/test_inference_config.py
+          TRANSFORMERS_CACHE=~/tmp/transformers_cache/ pytest $PYTEST_OPTS -k TestDistAllReduce unit/comm/test_dist.py
--- a/.github/workflows/cpu-inference.yml
+++ b/.github/workflows/cpu-inference.yml
@@ -73,7 +73,6 @@ jobs:
         run: |
           source oneCCL/build/_install/env/setvars.sh
           unset TORCH_CUDA_ARCH_LIST # only jit compile for current arch
-          if [[ -d ./torch-extensions ]]; then rm -rf ./torch-extensions; fi
           cd tests
-          TRANSFORMERS_CACHE=~/tmp/transformers_cache/ TORCH_EXTENSIONS_DIR=./torch-extensions pytest -m 'inference' unit/inference/test_inference_config.py
-          TRANSFORMERS_CACHE=~/tmp/transformers_cache/ TORCH_EXTENSIONS_DIR=./torch-extensions pytest -k TestDistAllReduce unit/comm/test_dist.py
+          TRANSFORMERS_CACHE=~/tmp/transformers_cache/ pytest -m 'inference' unit/inference/test_inference_config.py
+          TRANSFORMERS_CACHE=~/tmp/transformers_cache/ pytest -k TestDistAllReduce unit/comm/test_dist.py
--- a/.github/workflows/setup-venv/action.yml
+++ b/.github/workflows/setup-venv/action.yml
@@ -25,6 +25,7 @@ runs:
         echo TORCH_EXTENSIONS_DIR=./torch-extensions/ >> $GITHUB_ENV
         echo HF_DATASETS_CACHE=/blob/datasets_cache/ >> $GITHUB_ENV
         echo MEGATRON_CKPT_DIR=/blob/megatron_ckpt/ >> $GITHUB_ENV
+      shell: bash
     - id: print-env
       run: |
         which python
--- a/.github/workflows/setup-venv/action.yml
+++ b/.github/workflows/setup-venv/action.yml
@@ -23,8 +23,10 @@ runs:
         echo TEST_DATA_DIR=/blob/ >> $GITHUB_ENV
         echo TRANSFORMERS_CACHE=/blob/transformers_cache/ >> $GITHUB_ENV
         echo TORCH_EXTENSIONS_DIR=./torch-extensions/ >> $GITHUB_ENV
+        echo TORCH_CACHE=/blob/torch_cache/ >> $GITHUB_ENV
         echo HF_DATASETS_CACHE=/blob/datasets_cache/ >> $GITHUB_ENV
         echo MEGATRON_CKPT_DIR=/blob/megatron_ckpt/ >> $GITHUB_ENV
+        echo PYTEST_OPTS="--color=yes --durations=0 --verbose -rF" >> $GITHUB_ENV
       shell: bash
     - id: print-env
       run: |
--- a/.github/workflows/setup-venv/action.yml
+++ b/.github/workflows/setup-venv/action.yml
@@ -18,6 +18,13 @@ runs:
         pip install wheel # required after pip>=23.1
         echo PATH=$PATH >> $GITHUB_ENV # Make it so venv is inherited for other steps
       shell: bash
+    - id: set-env-vars
+      run: |
+        echo TEST_DATA_DIR=/blob/ >> $GITHUB_ENV
+        echo TRANSFORMERS_CACHE=/blob/transformers_cache/ >> $GITHUB_ENV
+        echo TORCH_EXTENSIONS_DIR=./torch-extensions/ >> $GITHUB_ENV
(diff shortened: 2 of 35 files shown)


tests/unit/ops/accelerators/test_accelerator_backward.py
```diff
@@ -244,30 +244,23 @@ def run_backward(ds_config, seq_len, atol=1e-2, verbose=False):
...
 
 
-#test_backward[3-1024-120-16-24-True-True-0.05]
-#test_backward[3-1024-52-16-24-False-True-0.2]
-# 3-128-54-2-24-False-True-0.2
-@pytest.mark.parametrize('batch_size, hidden_size, seq_len, heads, num_layers, is_preln, use_fp16, atol',
-                         [
-                             (64,160,128,2,24,False,True, 0.2),
-                             (64,1600,128,2,4,False,True, 0.2),
-                             (8,1600,128,25,3,True,True, 0.05),
```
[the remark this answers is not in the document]
mrwyattii: This is great to know, I'll add these back and also leave a comment about why we have them!

tests/unit/common.py
```diff
@@ -40,11 +42,10 @@ def get_xdist_worker_id():
 
 
 def get_master_port():
-    master_port = os.environ.get('DS_TEST_PORT', '29503')
-    xdist_worker_id = get_xdist_worker_id()
-    if xdist_worker_id is not None:
-        master_port = str(int(master_port) + xdist_worker_id)
-    return master_port
+    # Select a random open port
```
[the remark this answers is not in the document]
mrwyattii: Right, with the cached distributed processes, we need to make sure we're grabbing an unused port and this seemed like an easy solution. I wasn't aware that `DS_TEST_PORT` was used anywhere, but I'll double check

.github/workflows/cpu-inference.yml
```diff
@@ -75,7 +75,6 @@ jobs:
         run: |
           source oneCCL/build/_install/env/setvars.sh
           unset TORCH_CUDA_ARCH_LIST # only jit compile for current arch
-          if [[ -d ./torch-extensions ]]; then rm -rf ./torch-extensions; fi
           cd tests
-          TRANSFORMERS_CACHE=~/tmp/transformers_cache/ TORCH_EXTENSIONS_DIR=./torch-extensions pytest -m 'inference' unit/inference/test_inference_config.py
```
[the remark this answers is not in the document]
mrwyattii: `TORCH_EXTENSIONS_DIR` is now set for all workflows in `setup-venv/action.yml`

tests/unit/runtime/half_precision/onebit/test_onebit.py
```diff
@@ -1159,20 +1124,12 @@ def test(self, topo_config):
         }
 
         topo = PipeTopo(**topo_config)
-        steps = 500  # Must be >=100
-
-        # Allocate model for consistent initial weights.
-        init_net = AlexNetPipe()
+        steps = 10
```
tjruwase, line 1127: Should this be 100?

tests/unit/runtime/half_precision/onebit/test_onebit.py
```diff
@@ -1109,27 +1080,21 @@ def test_overflow(self, tmpdir):
...
-            "num_dp": 1
-        },
     ],
 )
 class TestOneBitLambFP16Pipeline(DistributedTest):
     world_size = 4
 
     def test(self, topo_config):
+        global_batch_size = 4
+        assert global_batch_size % self.world_size == 0, "Global batch size must be divisible by world size"
```
tjruwase, line 1094: I don't think this is needed since batch-related assertions already done in deepspeed: https://github.com/microsoft/DeepSpeed/blob/ce535945e6dacb6ffbf17a7cbf7543fbd86385dd/deepspeed/runtime/config.py#L878

tests/unit/runtime/pipe/test_pipe.py
```diff
@@ -41,9 +41,11 @@ class TestPipeCifar10(DistributedTest):
     def test(self, topo_config):
         skip_on_arch(min_arch=7)
 
+        global_batch_size = 4
+        assert global_batch_size % self.world_size == 0
         config_dict = {
-            "train_batch_size": 16,
-            "train_micro_batch_size_per_gpu": 4,
+            "train_batch_size": global_batch_size,
+            "train_micro_batch_size_per_gpu": int(global_batch_size / self.world_size),
```
tjruwase, line 48: It is much better to set `gradient_accumulation_steps` here to 1 rather than setting `micro_batch_size`

tests/unit/runtime/half_precision/onebit/test_onebit.py
mrwyattii: I have it at 10 because we're not doing any correctness checks here (e.g., comparing loss with a baseline), just verifying that the Pipeline runs

tests/unit/runtime/zero/test_zero.py
```diff
@@ -946,17 +1023,29 @@ def __init_weights(self, module):
 @pytest.mark.parametrize("zero_grad", [True, False])
 @pytest.mark.parametrize("prefetching", [True, False])
 @pytest.mark.parametrize("reduce_scatter", [True, False])
-@pytest.mark.parametrize("model_class", [
-    EltwiseMultiplicationTestNetwork_Dict, EltwiseMultiplicationTestNetwork_NamedTuple,
-    EltwiseMultiplicationTestNetwork_namedtuple, EltwiseMultiplicationTestNetwork_Tuple,
-    EltwiseMultiplicationTestNetwork_List
-])
+@pytest.mark.parametrize(
+    "model_class",
```
tjruwase, line 1027: This formatting looks odd because of the white lines.

tests/unit/runtime/half_precision/onebit/test_onebit.py
jeffra: Can we ensure we are escaping fp16 overflow within 10 steps? If we are not then the optimizer step won't be triggered at all

tests/unit/runtime/zero/test_zero.py
mrwyattii: I'm not sure why but `yapf` wants to format it like this now

tests/unit/runtime/half_precision/onebit/test_onebit.py
mrwyattii: I reverted this back to 100 steps - I just checked and it doesn't save _that_ much time with having it at only 10 steps

jeffra (approved):

tjruwase (approved):

tests/unit/common.py
```diff
@@ -40,11 +42,10 @@ def get_xdist_worker_id():
...
+    with socket.socket() as s:
+        s.bind(('', 0))
+        return s.getsockname()[1]
```
mrwyattii, line 48:
```suggestion
        return str(s.getsockname()[1])
```

.github/workflows/setup-venv/action.yml
```diff
@@ -18,6 +18,16 @@ runs:
...
       shell: bash
+    - id: set-env-vars
+      run: |
+        echo TEST_DATA_DIR=/blob/ >> $GITHUB_ENV
+        echo TRANSFORMERS_CACHE=/blob/transformers_cache/ >> $GITHUB_ENV
+        echo TORCH_EXTENSIONS_DIR=./torch-extensions/ >> $GITHUB_ENV
+        echo TORCH_CACHE=/blob/torch_cache/ >> $GITHUB_ENV
+        echo HF_DATASETS_CACHE=/blob/datasets_cache/ >> $GITHUB_ENV
+        echo MEGATRON_CKPT_DIR=/blob/megatron_ckpt/ >> $GITHUB_ENV
+        echo PYTEST_OPTS="--color=yes --durations=0 --verbose -rF" >> $GITHUB_ENV
```
sfc-gh-sbekman, line 29: @mrwyattii, for `-rF` the `pytest` doc doesn't mentioned `F` only `f` :

```
  -r chars              Show extra test summary info as specified by chars: (f)ailed, (E)rror, (s)kipped, (x)failed, (X)passed, (p)assed, (P)assed with
                        output, (a)ll except passed (p/P), or (A)ll. (w)arnings are enabled by default (see --disable-warnings), 'N' can be used to
                        reset the list. (default: 'fE').
                        
```

should it be `-rf`?

PR #3865, opened 2023-07-03
Decorator to avoid partitioning DeepSpeedEngine

eisene:
Following up on #3202 and the discussion in #3592, this introduces a class decorator called ```NoPartitioningDecorator``` that excludes a class from the partitioning hooks installed by the ```Init``` context in ```_enable_class``` and ```_init_subclass```. With this there is no need to manually manage the shutdown of the ```Init``` context inside ```deepspeed.initialize```.

### DIFF
--- a/deepspeed/runtime/zero/partition_parameters.py
+++ b/deepspeed/runtime/zero/partition_parameters.py
@@ -38,6 +38,14 @@ param_count = 0
 partitioned_param_data_shape = [0]
 zero_init_context = 0
 top_level_context = None
+_cls_excluded_from_partitioning = set()
+
+
+class NoPartitioningDecorator:
+    def __call__(self, cls):
+        global _cls_excluded_from_partitioning
+        _cls_excluded_from_partitioning.add(cls)
+        return cls
 
 
 class NoGatherHandle:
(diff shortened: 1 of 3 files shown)


tjruwase:
@eisene, thanks for this PR.
 
@tohtana, and @stas00 FYI

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -38,6 +38,14 @@
 partitioned_param_data_shape = [0]
 zero_init_context = 0
 top_level_context = None
+_cls_excluded_from_partitioning = set()
```
stas00, line 41: Now that deepspeed supports multiple models instantiation won't introducing a global state like this potentially lead to problems?

May be  `_cls_excluded_from_partitioning` should be cleared at the end of Init

It's hard to say if the user will want the same behavior for a sub-module with the same name in both instances or not.

eisene: This is intended for internal use inside DeepSpeed, to avoid having to explicitly shut down the Init context inside deepspeed.initialize. If we want to let users exclude their own classes from partitioning I think we need to rethink this. Do you have a usecase in mind?

stas00: I was just mentioning this since DS started to support multiple concurrent models, so it's hard to tell what users will try to do. That's why I'm voicing a concern about a global variable that could impact a complex setup.

I don't have a concrete case in mind right now, other than someone trying to train 2 models where one submodule is shared.

eisene: Agreed. @tjruwase I believe you have experience with this kind of use case?

tjruwase: @eisene, yes, I have being trying to eliminate global variables from this file to support partitioning of multiple models, e.g., RLHF training.

eisene:
This is too old now I think, no longer relevant.

Issue #3735, opened 2023-06-12
[BUG] RuntimeError: inflight params error when using DeepSpeed for Reinforcement Learning


shyustc:
**Description**
I am encountering a persistent error when using DeepSpeed for reinforcement learning, specifically with the GLM model as the actor and critic. The error persists even after upgrading to the latest versions of DeepSpeed and PyTorch Lightning.
The error occurs when I set ZeRO-3 in DeepSpeed and attempt to generate experiences using the actor model. The exact error message is: "RuntimeError: still have inflight params". Here is the relevant portion of the stack trace:

**log out**
```
   File "/home/jovyan/rlhf0/miniconda3/envs/py38cu113/lib/python3.8/site-packages/deepspeed/runtime/engine.py", line 1736, in forward
    loss = self.module(*inputs, **kwargs)
  File "/home/jovyan/rlhf0/miniconda3/envs/py38cu113/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1151, in _call_impl
    hook_result = hook(self, input, result)
  File "/home/jovyan/rlhf0/miniconda3/envs/py38cu113/lib/python3.8/site-packages/deepspeed/utils/nvtx.py", line 15, in wrapped_fn
    ret_val = func(*args, **kwargs)
  File "/home/jovyan/rlhf0/miniconda3/envs/py38cu113/lib/python3.8/site-packages/deepspeed/runtime/zero/parameter_offload.py", line 329, in _end_of_forward_hook
    self.get_param_coordinator(training=False).reset_step()
  File "/home/jovyan/rlhf0/miniconda3/envs/py38cu113/lib/python3.8/site-packages/deepspeed/runtime/zero/partitioned_param_coordinator.py", line 185, in reset_step
    raise RuntimeError(f"still have inflight params "
RuntimeError: still have inflight params [<bound method Init._convert_to_deepspeed_param.<locals>.ds_summary of Parameter containing:
```

**ds_report**
```
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
[...]

PR #3884, opened 2023-07-05
Separate ZeRO3 InflightParamRegistry for train and eval

HeyangQin:
After PR 3462, the InflightParamRegistry is bounded with the model so the InflightParamRegistry is the same for both training and eval, which is why there will be leftover parameters in flight that causes inflight parameter issues. With this fix and https://github.com/microsoft/DeepSpeed/pull/3883, the step3 runs well and it doesn't cause regression for the Lightning side (which is what PR 3462 tries to fix). Potentially related issues: https://github.com/microsoft/DeepSpeed/issues/3068 https://github.com/microsoft/DeepSpeed/issues/3156
https://github.com/microsoft/DeepSpeed/issues/3735
https://github.com/microsoft/DeepSpeedExamples/issues/337
https://github.com/microsoft/DeepSpeedExamples/issues/591.

### DIFF
--- a/.github/workflows/nv-lightning-v100.yml
+++ b/.github/workflows/nv-lightning-v100.yml
@@ -24,7 +24,7 @@ jobs:
 
       - name: Install pytorch
         run: |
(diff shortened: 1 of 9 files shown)


awan-10 (approved):

shyustc:
After updating to the latest version, I found that the previous issue has been fixed. However, I still encountered the 'inflight' error in the subsequent code at different locations.
#3735

PR #3886, opened 2023-07-05
Multiple zero stage 3 related fixes

tjruwase:
- zero.Init composition
- Add missing tensor allocators: torch.randn, torch.tensor
- Remove assert in overridden module.apply
- module.apply override disabling option
@stas00 

TODO
- [x] module.apply override documentation
- [ ]


deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -877,12 +896,21 @@ def _update_persist_config(self, ds_config):
         Init.param_persistence_threshold = ds_config.zero_config.param_persistence_threshold
         Init.model_persistence_threshold = ds_config.zero_config.model_persistence_threshold // self.num_partitions
 
+    def _zero_init_param(self, param):
+        self._convert_to_deepspeed_param(param)
```
tohtana, line 900: Do we need this conditional branch? 
When `if dist.get_world_group() == self.get_dp_process_group()`, `dist.get_global_rank(self.get_dp_process_group(), 0)` returns 0.
The "else" branch seems to cover another branch.

tohtana (commented):
Overall this looks good to me. I left a brief comment.

deepspeed/runtime/zero/partition_parameters.py
tjruwase: The `if` branch covers when we are using only data parallelism. The else branch covers the case of tensor parallelism (or pipeline parallelism) where the rank 0 of each tensor slice may not be the global rank 0.

tohtana (approved):

PR #3919, opened 2023-07-10
[CPU] Use allreduce_low_latency for AutoTP and implement low latency allreduce for CPU backend (single node)

delock:
In optimize DeepSpeed AutoTP inference workload, allreduce latency is critical to inference scaling.  This PR introduce a new deepspeed.comm interface `allreduce_low_latency` to allow communication backend to implement a low latency version of allreduce, which allows communication backend or library to optimize allreduce performance with different strategy.

In this PR we implemented a low latency allreduce on CCLBackend for CPU backend.  Experiment shows for very small messages, low latency allreduce bring more than 2x performance boost, compared to calling oneCCL library which is optimized for large message size with concurrency.

The opportunity of low latency allreduce comes from the fact that allreduce in training is different from allreduce in inference tensor parallel.  In training, allreduce usually has:
1. Large message size
2. Concurrency
Optimize of large message size and concurrency improves communcation in training but also complicate the software infrastructure.  In inference the message size is small and there is no need for concurrency.  Thus these software design won't help allreduce latency but might hurt it.  Seperate the path for latency and throughput gives an opportunity to optimize for latency in inference scenario.


tjruwase:
@delock, thanks for the PR. I am concerned that `allreduce_low_latency` could be a misleading name since the low latency is only achieved for small message sizes and low concurrency. A user could mistakenly think that this function could help improve allreduce performance for training as well. Can we explore names that more descriptive of the target uses, such as: 
1. ` small_message_allreduce`
2. `inference_allreduce`

Thanks!

csrc/cpu/comm/ccl.cpp
```diff
@@ -5,12 +5,245 @@
...
+    return _mm512_cvtusepi32_epi16(t_value);
+}
+
+void reduce_2_bf16_buffers(int num_elements, void* in_out, void* in)
+    __attribute__((target("avx512bw")));
+
+void reduce_bf16_buffers(int num_elements, int num_buffers, struct allreduce_workspace* workspace)
+    __attribute__((target("avx512bw")));
+
+void reduce_2_f32_buffers(int num_elements, void* in_out, void* in)
```
tjruwase, line 112: Is this a typo, `f32` ---> `fp32`?

csrc/cpu/comm/ccl.cpp
```diff
@@ -5,12 +5,245 @@
...
+    __attribute__((target("avx512bw")));
+
+void reduce_all_buffers(struct allreduce_workspace* workspace,
+                        int num_elements,
+                        c10::ScalarType scalar_type,
+                        int num_buffers)
+{
+    switch (scalar_type) {
+        case c10::ScalarType::BFloat16:
+            if (num_buffers >= 3 && num_buffers <= 8) {
```
tjruwase, line 125: What is the significance of the upper bound of `8`? Can we replace with symbolic constant with descriptive name?

tjruwase: Also, I think the lower bound is more intuitve to read as `num_buffers > 2` instead of `num_buffers >= 3`.

csrc/cpu/comm/ccl.cpp
```diff
@@ -193,6 +450,70 @@ void all_reduce_caching(torch::Tensor& data,
...
+    int data_size = 0;
+    bool data_type_fallback = false;
+
+    switch (data.scalar_type()) {
+        case c10::ScalarType::BFloat16: data_size = numel * 2; break;
+        case c10::ScalarType::Float: data_size = numel * 4; break;
+        default: data_type_fallback = true;
+    }
+
+    if (data_size > MAX_BUF_SIZE || (numel % 16) != 0 || data_type_fallback || !all_ranks_local_p) {
```
tjruwase, line 471: Would be good to replace `16` with a descriptive symbolic constant. 
I assume this is some alignment constraint. If so, I am surprised the same value is used for both bf16 and fp32.

csrc/cpu/comm/ccl.cpp
```diff
@@ -206,6 +527,7 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
     m.def("get_world_size", &get_world_size, "get world size");
     m.def("broadcast", &broadcast, "ccl broadcast");
     m.def("all_reduce", &all_reduce, "ccl all_reduce");
+    m.def("inference_allreduce", &inference_allreduce, "low latency all_reduce implementation");
```
tjruwase, line 530: Please add some unit tests for this.

csrc/cpu/comm/ccl.cpp
delock: Thanks for point out.  Will change to fp32.

csrc/cpu/comm/ccl.cpp
delock: 8 is a number limited by the implementation of REPEAT(N, x) macro and switch case that use REPEAT macro.  It can be extended to a bigger number if needed.  I'll change it to a symolic constant and instructions how to make this number bigger.

csrc/cpu/comm/ccl.cpp
delock: Thanks!  Added a new constant `VECTOR_LENGTH_IN_BYTES` and define it to 32.  This would reflect the length of vector used in reduce kernels (256bits=>32bytes).   Will check with `data_size % VECTOR_LENGTH_IN_BYTES` here, and also use the same constant to control loop steps in reduce kernels.

delock:
> > @delock, thanks for the PR. I am concerned that `allreduce_low_latency` could be a misleading name since the low latency is only achieved for small message sizes and low concurrency. A user could mistakenly think that this function could help improve allreduce performance for training as well. Can we explore names that more descriptive of the target uses, such as:
> > 
> > 1. ` small_message_allreduce`
> > 2. `inference_allreduce`
> > 
> > Thanks!
> 
> Thanks for the comments @tjruwase , I think `inference_allreduce` would be better since in training scenario concurrency could be assumed. Would make the change.

Will use `inference_all_reduce` to keep naming consistency with `all_reduce`.

csrc/cpu/comm/ccl.cpp
tjruwase: @delock, thanks. 

@mrwyattii, can you please take a look at the unit test changes?

tests/unit/comm/test_dist.py
```diff
@@ -123,6 +123,23 @@ def test(self):
...
 
 
+class TestDistInferenceAllReduce(DistributedTest):
+    device_count = get_accelerator().device_count()
+    if device_count >= 4:
+        world_size = [1, 2, 4]
+    elif device_count >= 2:
+        world_size = [1, 2]
+    else:
+        world_size = [1]
```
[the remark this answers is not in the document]
delock: Thanks! Will fix world_size to 4

tjruwase (approved):

PR #3931, opened 2023-07-12
add flash_attn as deepspeed ops

YizhouZ:
Hi, we would like to add flash attention mentioned by paper https://arxiv.org/abs/2205.14135 as a deepspeed op.

This PR is to create an interface for developers to implement fmha on their own devices.


stephen-youn:
from where do u get "FlashAttentionBuilder"? where is the cuda kernel for flash attn?

YizhouZ:
Closing this PR since Megatron-DS already has this layer https://github.com/microsoft/Megatron-DeepSpeed/blob/main/megatron/model/transformer.py#L340.

PR #3784, opened 2023-06-21
ZeRO++

HeyangQin:
We are releasing [ZeRO++](https://arxiv.org/abs/2306.10209), a system of communication optimization strategies built on top of ZeRO to offer unmatched efficiency for large model training, regardless of batch size limitations or cross-device bandwidth constraints. ZeRO++ leverages quantization, in combination with data, and communication remapping, to reduce total communication volume by 4x compared with ZeRO, without impacting model quality.

ZeRO++ has three sets of communication optimizations:

- Quantized weight communication for ZeRO (qwZ)
- Hierarchical weight partition for ZeRO (hpZ) 
- Quantized gradient communication for ZeRO (qgZ)

Please refer to our ZeRO++ [paper](https://arxiv.org/abs/2306.10209) and [blog](https://www.microsoft.com/en-us/research/blog/deepspeed-zero-a-leap-in-speed-for-llm-and-chat-model-training-with-4x-less-communication/) for details. 

ZeRO++ is particularly useful for:

a) Small batch-size per GPU: Whether pre-training large models on thousands of GPUs or fine-tuning them on hundreds or even dozens of GPUs, when batch-size per GPU is small, ZeRO++ offers up to 2.2x higher throughput compared to ZeRO, directly reducing training time and cost.

b) Low-bandwidth clusters: ZeRO++ enables low-bandwidth clusters to achieve similar throughput as those with 4x higher bandwidth. Therefore, ZeRO++ makes efficient large model training accessible across a wider variety of clusters.

Contributors:
This code release was made possible by the contributions of the following people from the DeepSpeed Team:
Guanhua Wang, Heyang Qin, Sam Ade Jacobs, Connor Holmes, [Samyam Rajbhandari](https://www.microsoft.com/en-us/research/people/samyamr/), [Olatunji Ruwase](https://www.microsoft.com/en-us/research/people/olruwase/), Ammar Ahmad Awan, Jeff Rasley, Michael Wyatt,  [Yuxiong He](https://www.microsoft.com/en-us/research/people/yuxhe/) (team lead)

PR #3933, opened 2023-07-12
Use device_name instead of device index to support other device

hipudding:
Create tensor with device=Integer will always choose cuda as its deivce for current pytorch version(2.1), other device should use device={device}:{index}.

Change get_accelerator().current_device() to
get_accelerator().current_device_name() to support other devices.


tjruwase:
@hipudding, can you please share more details on the failure this is supposed to fix? Each accelerator can implement `current_device()` as appropriate, so I don't see how this API results in `cuda` devices. Thanks!

hipudding:
> @hipudding, can you please share more details on the failure this is supposed to fix? Each accelerator can implement `current_device()` as appropriate, so I don't see how this API results in `cuda` devices. Thanks!

Yes. For example, I'm using NPU as the backend and want to create a empty tensor.

```
a_tensor = tensor.empty(10, device=get_accelerator().current_device()
```

For every accelerator, current_device() will return the current index of the backend. Suppose we are using npu:1, current_device() will return Integer 1. Then, the code above equals to:

```
a_tensor = tensor.empty(10, device=1)
```

But pytorch will use Cuda as it's backend if device is a Integer.
image

If these code want to work with every backend, it should specify the device name, so I changed current_device() to current_device_name(), which will return a device name and its index.

tjruwase:
@hipudding, I see your point. I agree that this quite an incovenience of torch, but I was suggesting that rather than changing deepspeed code, you could follow xpu_accelerator [implementation](https://github.com/intel/intel-extension-for-deepspeed/blob/223a4edbbeca89503933231191df45f6537aaf54/intel_extension_for_deepspeed/xpu_accelerator.py#L27-L28). That is working without needing this PR.

tjruwase:
On second thought, perhaps I should confirm what xpu_accelerator is actually doing. @delock, how do you avoid the problem solved by this PR? Thanks.

delock:
`current_device_name()` should be the right API to use.  `current_deivce()` will return an index, which should only be used when a number is needed, i.e. iterate all device indexes.

I see these codes are introduced from zero++ (https://github.com/microsoft/DeepSpeed/pull/3784), it should be misuse of `current_device()`.  We have not started to evaluate zero++ yet, so probably why `xpu` didn't encounter this issue.

tjruwase:
@delock, thanks for the explanation. That makes sense.

tjruwase (approved):

PR #3936, opened 2023-07-12
Update zero_to_fp32.py - to support deepspeed_stage_1

PicoCreator:
Since deepspeed 1 checkpoint structure is identical to deepspeed 2 (AFAIK), we should just change the stage check and add support accordingly

However I am not 100% sure if this is intentional by design, or some coincidence in my use case - so might need someone with more knowledge on this topic to weigh in 🤔

### DIFF
--- a/deepspeed/utils/zero_to_fp32.py
+++ b/deepspeed/utils/zero_to_fp32.py
@@ -5,7 +5,7 @@
 
 # DeepSpeed Team
 
-# This script extracts fp32 consolidated weights from a zero 2 and 3 DeepSpeed checkpoints. It gets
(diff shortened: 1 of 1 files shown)


tjruwase:
@stas00, do you remember why stage 1 was excluded?

tjruwase:
@PicoCreator, thanks for this PR.

PicoCreator:
Glad to see my wild guess, work and be of use (found a few issues on dependent projects that encounted this)

I currently use deepspeed 1 to train small toy models (<=3B), as fast as possible, and to test param/model architecture changes =) on consumer hardware (where the gpu-to-gpu communication of deepspeed 2 is noticeable)

PR #3939, opened 2023-07-12
Re-enable skipped unit tests

mrwyattii:
Replacing `required_torch_version` utility function with `required_minimum_torch_version`. The old function was causing several tests to be skipped on runners that use torch 2.x


tjruwase (approved):

jeffra (approved):
LGTM, might be good to switch to use something like [packaging](https://packaging.pypa.io/en/stable/index.html) to help prevent issues like this in the future though.

```python
def required_minimum_torch_version(min_version="1.8"):
    from packaging import version
    return version.parse(min_version) < version.parse(torch.__version__)
```

We're currently using this in a few spots across DeepSpeed, e.g., https://github.com/microsoft/DeepSpeed/blob/aa54dba02532f3f487cc635dfdace8fa95593db0/deepspeed/__init__.py#L50
https://github.com/microsoft/DeepSpeed/blob/aa54dba02532f3f487cc635dfdace8fa95593db0/deepspeed/runtime/zero/stage_1_and_2.py#L2169

mrwyattii:
> LGTM, might be good to switch to use something like [packaging](https://packaging.pypa.io/en/stable/index.html) to help prevent issues like this in the future though.

Added this in the latest commit and refactored the `required_*_torch_version` functions to just `required_torch_version`

Issue #3358, opened 2023-04-23
[BUG]nvcc fatal  when DS_BUILD_TRANSFORMER_INFERENCE=1 


kuangdao:
i user pytorch 1.12 with cuda 11.6, and with ds config of 

DS_BUILD_FUSED_ADAM=1
DS_BUILD_CPU_ADAM=1
DS_BUILD_UTILS=1
DS_BUILD_TRANSFORMER=0
DS_BUILD_STOCHASTIC_TRANSFORMER=0
DS_BUILD_TRANSFORMER_INFERENCE=1
DS_BUILD_OPS=0

use pip3 install deepspeed --global-option="build_ext" --global-option="-j8" 

and the finnal error is  error: command '/usr/local/cuda/bin/nvcc' failed with exit code 1 

nvcc fatal   : Unknown option '-Wno-reorder'

mrwyattii:
What version of pip do you have installed? `pip --version`

kuangdao:
0.9.1 error bug 0.9.0 works

mrwyattii:
@kuangdao you get this bug with deepspeed `0.9.1` but not with `0.9.0`?

gfzum:
I solved this by simply removed additional options, i.e. `DS_BUILD_OPS=1 pip install deepspeed`

jockeyyan:
> 

You doesn't solve the issue as you never build extensions any more.

hipudding:
I get the same error.
```
torch                         1.13.1+cu117
torchaudio                    0.13.1+cu117
torchvision                   0.14.1+cu117
```

deepspeed==0.9.2
cuda==11.7

nvcc fatal   : Unknown option '-Wno-reorder'
nvcc fatal   : Unknown option '-Wall'

crclark:
Same error with [the latest Nvidia pytorch Docker image](https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/rel-23-05.html#rel-23-05). It happens to me with both `0.9.0` and `0.9.4`.

Nvidia driver version: 525.116.03. 
Using `nvidia-container-runtime`.
RTX 6000 Ada GPU.

Minimal repro:

```dockerfile
FROM nvcr.io/nvidia/pytorch:23.05-py3

# Set the working directory in the container to /app
WORKDIR /app

RUN pip install --upgrade pip
RUN apt-get update
RUN apt-get install -y libaio-dev
[...]

Issue #3067, opened 2023-03-21
[BUG] Cannot build with AIO


fecet:
I create a conda environment to build DeepSpeed as follows:
```
conda create -n dp \
    -c "nvidia/label/cuda-11.7.0"  \
    -c pytorch \
    -c conda-forge \
	cuda compilers \
    pytorch torchvision torchaudio pytorch-cuda=11.7 
```
Then I build deepspeed from source:
```
DS_BUILD_OPS=1 DS_BUILD_SPARSE_ATTN=0 pip install . | tee /dev/tty
```
It says that `libaio` cannot be founded. 
I'm on Arch Linux and I can confirm I have installed it, then I use conda to install another one:
```
conda install libaio
```
then the previous command can continue and start to build, but throw an error after a while
```
[...]

Issue #2885, opened 2023-02-23
parallel build no longer works


stas00:
I used to prebuild deepspeed with:
```
DS_BUILD_CPU_ADAM=1 DS_BUILD_AIO=1 DS_BUILD_TRANSFORMER_INFERENCE=1 DS_BUILD_UTILS=1 pip install -e . --global-option="build_ext" --global-option="-j8" --no-cache -v --disable-pip-version-check 2>&1 | tee build.log
```

it no longer works. I have to use `--global-option="-j1"` for it to succeed.

Log file attached. [build-log.txt](https://github.com/microsoft/DeepSpeed/files/10818135/build-log.txt)

Thank you!

stas00:
> Hi @stas00, I'm not able to reproduce the error you are seeing locally (I copy and pasted the command you shared). Can you share a bit more information?
> 
>     * Is the error message the same when you repeat the command?

well, my build script removes the `build` dir before building.

I have tested that if I don't remove the `build` dir, and run the build 2 times, the process succeeds.

my script is:
```
#!/bin/bash
rm -rf build
DS_BUILD_CPU_ADAM=1 DS_BUILD_AIO=1 DS_BUILD_TRANSFORMER_INFERENCE=1 DS_BUILD_UTILS=1 pip install -e . --global-option="build_ext" --global-option="-j8" --no-cache -v --disable-pip-version-check 2>&1 | tee build.log
```
 
>     * What version of nvcc and cuda do you have?

`Cuda compilation tools, release 11.7, V11.7.64`

>     * What version of pytorch do you have?

`pt=1.13.1, cuda=11.7, nccl=(2, 14, 3)`

>     * Is this with the latest master branch of DeepSpeed or another version?

master, just retested with today's master - same

mrwyattii:
Thanks @stas00 - I've matched your environment as close as I can, but I'm still not seeing this error. I'll try on a few different system and let you know. Just curious, do you see this on other systems as well?

stas00:
I tried on another system, but `DS_BUILD_TRANSFORMER_INFERENCE=1` build was failing at all, it build just fine with -j8 otherwise.

And actually once I removed `DS_BUILD_TRANSFORMER_INFERENCE=1` from the system I made the OP report with the -j8 problem disappears as well. So it has something to do with the inference component.

stas00:
[...]

PR #3943, opened 2023-07-13
add reproducible compilation environment

fecet:
Currently, deepspeed only verifies the compilation process on Docker, which may not work on many clusters or not have the permission to install deepspeed using Docker. This makes precompiled deepspeed ops very challenging, especially considering that the compilation chain tools can vary significantly between different systems. There have been many issues complaining about their inability to compile ops in their own environment, say https://github.com/microsoft/DeepSpeed/issues/3890 https://github.com/pytorch/pytorch/pull/100557 https://github.com/microsoft/DeepSpeed/issues/3358 https://github.com/microsoft/DeepSpeed/issues/3067

Conda-forge provides a cross-platform compilation toolchain that, if we maintain a robust Conda environment, can make precompiled ops available to everyone and solve the issue of reproducibility.

I verify the environment on Arch LInux and Ubuntu 20.04 for pytorch and pytorch-nightly. For pytorch nightly, `DS_BUILD_AIO` should be used as it seems that op doesn't support `c++17` yet. And parallel parallel build option should be disabled as https://github.com/microsoft/DeepSpeed/issues/2885

### DIFF
--- a/docs/_tutorials/advanced-install.md
+++ b/docs/_tutorials/advanced-install.md
@@ -123,6 +123,16 @@ fail. Therefore, if you need to you can override the default location with the h
  TORCH_EXTENSIONS_DIR=./torch-extensions deepspeed ...
 ```
 
+### conda environment for building from source
+
+If you encounter difficulties during compilation using the default system environment, you can try the conda environment we provide, which includes the necessary compilation toolchain and PyTorch. 
+
+```bash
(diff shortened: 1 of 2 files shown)


fecet:

tjruwase:
@fecet, thanks for this amazing PR that also includes documentation. We will review right away, apologies for the delay.

PR #3962, opened 2023-07-14
add lm_head and embed_out tensor parallel

Yejing-Lai:
This PR aims to add lm_head and embed_out layer tensor parallel. This applies to models whose last layer is named lm_head/embed_out, such as llama, gpt-j, bloom, opt, and so on.

### DIFF
--- a/deepspeed/module_inject/layers.py
+++ b/deepspeed/module_inject/layers.py
@@ -29,6 +29,36 @@ class LinearAllreduce(nn.Module):
         return output
 
 
+class LmHeadLinearAllreduce(nn.Module):
+
+    def __init__(
+        self,
+        weight,
+        rank,
+        world_size,
+        bias=None,
+        mp_group=None,
+    ):
+        super(LmHeadLinearAllreduce, self).__init__()
+        self.weight = weight
+        self.bias = bias
+        self.mp_group = mp_group
+        self.rank = rank
+        self.world_size = world_size
+
+    def forward(self, input):
+        assert input.shape[
+            -1] % self.world_size == 0, 'Please ensure that self.world_size is divisible by input.shape[-1]'
+        input_shard = input.shape[-1] // self.world_size
+        output = torch.matmul(input[:, :, self.rank * input_shard:(self.rank + 1) * input_shard],
+                              self.weight.transpose(-1, -2))
+        if self.mp_group is not None:
+            dist.all_reduce(output, group=self.mp_group)
+        if self.bias is not None:
+            output += self.bias
+        return output
+
+
 class LinearLayer(nn.Module):
 
     def __init__(self, weight_shape=None, dtype=torch.half, weight=None, bias=None):
--- a/deepspeed/module_inject/layers.py
+++ b/deepspeed/module_inject/layers.py
@@ -53,7 +53,7 @@ class LmHeadLinearAllreduce(nn.Module):
         output = torch.matmul(input[:, :, self.rank * input_shard:(self.rank + 1) * input_shard],
                               self.weight.transpose(-1, -2))
         if self.mp_group is not None:
-            dist.all_reduce(output, group=self.mp_group)
+            dist.inference_all_reduce(output, group=self.mp_group)
         if self.bias is not None:
             output += self.bias
         return output
--- a/deepspeed/module_inject/replace_module.py
+++ b/deepspeed/module_inject/replace_module.py
@@ -275,6 +275,8 @@ def replace_transformer_layer(orig_layer_impl, model, checkpoint_dict, config, m
         _autotp.update_linear_policies()
 
         # 4. Replace modules
+        if "lm_head" in str(module) or 'embed_out' in str(module):
+            return _autotp._replace_last_linear_module(module)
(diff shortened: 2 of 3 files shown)


RezaYazdaniAminabadi:
> This PR aims to add lm_head and embed_out layer tensor parallel. This applies to models whose last layer is named lm_head/embed_out, such as llama, gpt-j, bloom, opt, and so on.

Yejing-Lai:
> > This PR aims to add lm_head and embed_out layer tensor parallel. This applies to models whose last layer is named lm_head/embed_out, such as llama, gpt-j, bloom, opt, and so on.
> > Hi @Yejing-Lai
> 
> Thanks for the nice PR to add the parallelism in the last embedding linear layer. I was just wondering how much of the performance impact this has, considering that you have the communication overhead? Also, you can split the weight in different ways, either across the hidden_dim requiring an all-reduce (which is what you did), or across the embedding dimension which then needs an all-gather. Either way, has their own tradeoffs, in the case of embedding-dim we can reduce the amount of computation if there is a very large vocabulary, however, with sharding across the model dimension, we reduce the amount of communication in this case. I think having both of these options makes sense and we can later decide on which needs to be enabled.

Hi @RezaYazdaniAminabadi. We observed an increase in E2E performance on SPR after adding lm_head/embed_out TP. I think the amount of calculations reduced by splitting weights on hidden_dim or embedding_dim is the same, and they all distribute the weights equally to each rank. Thanks~

deepspeed/module_inject/layers.py
```diff
@@ -29,6 +29,36 @@ def forward(self, input):
...
+        self.world_size = world_size
+
+    def forward(self, input):
+        assert input.shape[
+            -1] % self.world_size == 0, 'Please ensure that self.world_size is divisible by input.shape[-1]'
+        input_shard = input.shape[-1] // self.world_size
+        output = torch.matmul(input[:, :, self.rank * input_shard:(self.rank + 1) * input_shard],
+                              self.weight.transpose(-1, -2))
+        if self.mp_group is not None:
+            dist.all_reduce(output, group=self.mp_group)
```
delock, line 56: Should use `inference_all_reduce` since https://github.com/microsoft/DeepSpeed/pull/3919 is merged.

deepspeed/module_inject/layers.py
```diff
@@ -29,6 +29,36 @@ def forward(self, input):
...
+        self.weight = weight
+        self.bias = bias
+        self.mp_group = mp_group
+        self.rank = rank
+        self.world_size = world_size
+
+    def forward(self, input):
+        assert input.shape[
+            -1] % self.world_size == 0, 'Please ensure that self.world_size is divisible by input.shape[-1]'
+        input_shard = input.shape[-1] // self.world_size
```
delock, line 52: We can actually make this work with input.shape[-1] not divisible by world_size.  Need to consider relationship between this pR and the following PR https://github.com/microsoft/DeepSpeed/pull/4011

delock:
Divide among hidden dim will be a simpler solution.   It is consistent with other layers and we can reuce `inferece_all_reduce` to reduce communication latency further.   Low communcation size also means it more friendly to scaleout scenario.

delock:
Hi @RezaYazdaniAminabadi, is this PR still under review, or embedding parallel method still needs more consideration?

deepspeed/module_inject/replace_module.py
```diff
@@ -562,6 +564,14 @@ def replace_module(model, orig_class, replace_fn, _replace_policy, checkpoint=No
...
                 replaced_module.lm_head, "weight") and replaced_module.lm_head.weight.is_meta:
             replaced_module.lm_head.weight = embedding_weight
+
+    # enable tensor parallel for the last linear
+    if hasattr(replaced_module, "lm_head") and hasattr(replaced_module.lm_head,
+                                                       "weight") and not replaced_module.lm_head.weight.is_meta:
+        replaced_module = replace_fn(replaced_module, ("lm_head", ), 0, "lm_head")
+    elif hasattr(replaced_module, "embed_out") and hasattr(replaced_module.embed_out,
+                                                           "weight") and not replaced_module.embed_out.weight.is_meta:
+        replaced_module = replace_fn(replaced_module, ("embed_out", ), 0, "embed_out")
```
dc3671, line 574: Hi @RezaYazdaniAminabadi @delock, I refactored some code here, so that it will not integrate so deeply with former replace logic. And it will replace the `lm_head` or `embed_out` Linear **after** all other replace/load logics are finished. It will be more decoupling and cleaner now.

dc3671:
About the performance, in our simple test, it will get an around 13% end2end improvement for bloom-176B on per token latency.

dc3671:
Also, if need to make this function optional, I think we need to add an item in DeepSpeed's Config. Maybe this needs further discussion?

RezaYazdaniAminabadi (approved):
All LGTM, thanks everyone for making this part of inference faster :)

dc3671:
Hi @tjruwase , I have fixed the merge conflict.

dc3671:
Hi @tjruwase , the CI problems are fixed.

PR #3972, opened 2023-07-17
Fix get_param_coordinator() in _end_of_forward_hook()

mmhab:
In DeepSpeed Chat we call the _end_of_forward_hook both in the training and the eval mode so the training should not be always False.


HeyangQin:
@mmhab Thank you for contributing to DeepSpeed! For this call to `get_param_coordinator()`, it is protected by `torch._C.is_grad_enabled`, so it should only be called when the model is in eval mode. Is there any exception that we missed? If so, please provide an example.

cc @tjruwase

PR #3973, opened 2023-07-17
fix bloom bugs about inference and TP

dawson-chen:
The code on the master branch(20230717) are incompatible with bloom due to two issues:

1. Bloom Inference bug in softmax
    An error occurs when doing Bloom inference in the softmax function, and the reason is unclear. However, replacing the ds-softmax function with alternative operator in Torch could solve this problem.
    
3. Bloom TP error
    The parameters copied from Bloom model when useing TP is in worry shape. This issue can be resolved by defining customized mp-functions in the BloomContainer.


AllenChennn:

delock:
Hi @AllenChennn are you running bloom on CPU or CUDA device?   Are you running with AutoTP or kernel injection?

AllenChennn:
> Hi @AllenChennn are you running bloom on CPU or CUDA device? Are you running with AutoTP or kernel injection?

No, I'am not using AutoTP, as this is not a feature supported by HybridEngine.

Issue #3979, opened 2023-07-18
[BUG] BFloat16 training accuracy issue with zero 0/1 (the more rank, the less accuracy) in CIFAR10


delock:
**Describe the bug**
When using BF16 to train CIFAR10 following https://github.com/microsoft/DeepSpeedExamples/pull/651.   I encounter accuracy loss with the following conditions:
1. BF16 training
2. zero stage 0 and 1
3. world_size >= 2

**To Reproduce**
1. Use PR https://github.com/microsoft/DeepSpeedExamples/pull/651
2. Goto DeepSpeedExamples/training/cifar/
3. Change ds_config.json to change 'fp16' to 'bf16'
4. run the `run_ds.sh` with multiple accelerators on the system
5. Observe accuracy loss with multiple ranks.  For 2 ranks, accuracy is 52%, much less than 57% of 1 rank.   For 8 ranks, accuracy drops to around 43%
6. This can be observed from two CUDA cards.  The 8 rank result is observed from a WIP CPU training branch.  Should also be observed on 8 CUDA card system
7. When change data type to fp32 (on CPU system), there is no issue with 2 ranks or 8 ranks.  When change to zero stage 2, there is no accuracy issue.

**Expected behavior**
BF16 train accuracy the same as single rank.

**ds_report output**
```
[2023-07-18 12:16:47,983] [INFO] [real_accelerator.py:133:get_accelerator] Setting ds_accelerator to cpu (auto detect)
[2023-07-18 12:16:48,129] [INFO] [real_accelerator.py:133:get_accelerator] Setting ds_accelerator to cpu (auto detect)
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
deepspeed_not_implemented  [NO] ....... [OKAY]
deepspeed_ccl_comm ..... [NO] ....... [OKAY]
fused_adam ............. [NO] ....... [OKAY]
--------------------------------------------------
DeepSpeed general environment info:
torch install path ............... ['/home/gma/anaconda3/envs/dscpu/lib/python3.11/site-packages/torch']
torch version .................... 2.0.1+cu117
[...]

PR #2507, opened 2022-11-15
CUDA optional deepspeed ops

tjruwase:
Reboot of #1280

PR #3991, opened 2023-07-19
[CPU] FusedAdam and CPU training support

delock:
This PR implements FusedAdam and also enables CPU training with ZeRO stage 0-3.  Note FP16 data type are not supported for CPU training.  BF16 is the preferred datatype.

For ZeRO stage 0, this PR depends on #3842.

Whats in this PR:
1. FusedAdam kernel, the implementation is from CPU Adam kernel and had been modified to suite the case.
2. Fix bugs in CPU Accelerator interface implementation which had been exposed by training workload.
3. Handling of synchronized accelerator in DeepSpeed runtime.

This PR had been verified on CIFAR10 in DeepSpeedExamples, A PR had been submitted to support BF16 data type for CIFAR10 https://github.com/microsoft/DeepSpeedExamples/pull/651 .  Note for Zero0/1, there is an accuracy bug in DeepSpeed supporting BF16 training (https://github.com/microsoft/DeepSpeed/issues/3979), for Zero2/3, accuracy is correct.


accelerator/cpu_accelerator.py
```diff
@@ -139,7 +139,7 @@ def reset_max_memory_cached(self, device_index=None):
         return
 
     def memory_stats(self, device_index=None):
-        return self.get_rss()
+        return None
```
tjruwase, line 142: Why return `None`? Is there some other way to observe memory statistics of cpu_accelerator?

csrc/cpu/adam/cpu_adam.h
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0
+
+// DeepSpeed Team
+
+#pragma once
+
+#define NOMINMAX  // Windows idiosyncrasy
+                  // https://stackoverflow.com/questions/4913922/possible-problems-with-nominmax-on-visual-c
+
```
tjruwase, line 10: This file has some duplication with `csrc/include/cpu_adam.h`. Do you plan to merge them?

csrc/cpu/adam/simd.h
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0
+
+// DeepSpeed Team
+
+#pragma once
```
tjruwase, line 6: This is also similar to `csrc/includes/simd.h`. It will get good to merge to avoid duplicate files in the repo.

accelerator/cpu_accelerator.py
delock: From the usage of memory_stats caller is expecting a dict or None.   The attribute DeepSpeed is using is `num_alloc_retries` which is GPU specific.
https://github.com/microsoft/DeepSpeed/blob/fb9aebbf254d6d4535e42df75cb08b7c1dbd03f2/deepspeed/runtime/zero/stage3.py#L1895

Any suggestions?  Besides `None`, alternative options is return a dict with attribute `num_alloc_retries` set to 0 and emulate other attributes in the future when needed.

csrc/cpu/adam/cpu_adam.h
delock: the CUDA version of cpu_adam.h provide Adam in the form of optimizer object.  We changed it to stateless function call to fit the scenario of FusedAdam.  Besides the CUDA version contains CUDA specific code, so we think its better to make a copy of cpu_adam.h to decouple.

csrc/cpu/adam/simd.h
delock: This file is identical to `csrc/includes/simd.h`, let me see if the duplication could be removed.

accelerator/cpu_accelerator.py
tjruwase: Thanks for pointing this out. Here are my thoughts. 
1. We should return `dict` of interesting memory statistics. I think it is okay if keys are different across accelerators.
2. We should change DeepSpeed code to first check `num_alloc_retries` in the returned dict before executing any logic.

What do you think?

delock: Updated memory_stats to report `allocated_bytes.all.current` and `allocated_bytes.all.peak`, also check whether key `num_alloc_retries` exist first in use point.

csrc/cpu/adam/cpu_adam.h
tjruwase: I understand. In the long run, it would be nice to remove duplicate code by having the CUDA cpu_adam inherit from the basic cpu_adam. But that will be a future PR.

delock:
@tjruwase note we also have a PR in DeepSpeedExamples to demostrate CIFAR10 training with CPU.  The PR also fix test broken for pytorch 2.0 because of API change.
https://github.com/microsoft/DeepSpeedExamples/pull/651

csrc/cpu/adam/cpu_adam.h
tjruwase: @delock, apologies for revisiting this issue. But I was reminded that we had made cpu_adam runnable in CPU only environment with PR #2507. Moreover, we created this [unit test](https://github.com/microsoft/DeepSpeed/blob/23a11a39510e2aefb48236e3d2672a7dcbfc42a3/tests/unit/ops/adam/test_cpu_adam.py#L90) for this scenario since `torch.optim.AdamW` can also work in cpu only environment. 

So I am unsure that the changes in `csrc/cpu` are actually needed. Did you try to use existing for to support CPU training? Thanks!

delock:
Also added CPUAdamBuilder for CPU accelerator, so HelloDeepSpeed which use cpu offload can run.

tjruwase (approved):

csrc/cpu/adam/cpu_adam.h
tjruwase: Awesome! Thanks!

PR #4033, opened 2023-07-25
remove duplicate check for pp and zero stage

inkcherry:
duplicate  with https://github.com/microsoft/DeepSpeed/blob/f3943cf9109226ed3ecf2d5dbb639a11cd925555/deepspeed/runtime/pipe/engine.py#L58

no need to check again.


tjruwase:
@inkcherry, thanks for this PR. Do you mind replacing the `2` in the assert with `ZeroStageEnum.gradients`? Thanks!

tjruwase (approved):

PR #4044, opened 2023-07-26
load linear layer weight with given dtype

polisettyvarma:
bf16 inference fails due to data type mismatch as half is default value


polisettyvarma:
@mrwyattii please review this as well

nelyahu:
Hi @jeffra :) can you please help reviewing this change?

polisettyvarma:
Hi @RezaYazdaniAminabadi @mrwyattii @jeffra @awan-10 @cmikeh2 @arashb 
please review this change.

deepspeed/module_inject/load_checkpoint.py
```diff
@@ -248,7 +248,10 @@ def load_module_recursive(module, prefix='', level=0):
                     if child.__class__ is nn.LayerNorm:
                         child = Normalize(dim=ds_shape[-1], dtype=child.weight.dtype, eps=child.eps)
                         setattr(module, name, child)
-                    elif child.__class__ in [nn.Linear, ColumnParallelLinear, RowParallelLinear]:
+                    elif child.__class__ is nn.Linear:
+                        child = LinearLayer(weight_shape=child.weight.shape, dtype=child.weight.dtype, bias=child.bias)
+                        setattr(module, name, child)
+                    elif child.__class__ in [ColumnParallelLinear, RowParallelLinear]:
```
tjruwase, line 254: Why is this fix not applied to `ColumnParallelLinear` or `RowParallelLinear` as well? Can they not experience the same dtype mismatch problem?

polisettyvarma: https://github.com/facebookresearch/fairscale/blob/main/fairscale/nn/model_parallel/layers.py#L239-L247
both ColumnParallelLinear and RowParallelLinear doesn't accept dtype.

RezaYazdaniAminabadi: Hi @polisettyvarma,

Thanks for fixing this issue.
I think @tjruwase is right here, we should do this change for all different type of the linear layers. Note that we don't want to send dtype to ColumnParallelLinear or RowParallelLinear, but replacing them with LinearLayer.

RezaYazdaniAminabadi:
> Hi @RezaYazdaniAminabadi @mrwyattii @jeffra @awan-10 @cmikeh2 @arashb please review this change.

The changes look good to me 👍 I think we can merge it after fixing the comment from @tjruwase

deepspeed/module_inject/load_checkpoint.py
polisettyvarma: @RezaYazdaniAminabadi should i open another PR on fairscale to add dtype ?

RezaYazdaniAminabadi: I don't think it is needed. The dtype for those linears will be determined either by loading checkpoint or in the higher level of stack when they convert the model to the dtype they need. For DeepSpeed, since this injection is happening after model is created and converted, we need to specify the right dtype which is coming from the base layer's weight parameter dtype.

tjruwase (approved):

PR #4048, opened 2023-07-26
Fix deadlock when SHM based allreduce spin too fast

delock:
This PR fix a potential deadlock when SHM based allreduce spin too fast.  This is caused if the computation between two allreduce finishes too fast, so rank 0 would spin the state to next allreduce, while other rank still exiting previous allreduce.  This situation will cause a deadlock.

This PR fix by let all other rank check rank 0's state in a more robust way before exiting previous allreduce.


delock:
@dc3671 @baodii Please also help review this change.

csrc/cpu/comm/ccl.cpp
```diff
@@ -526,7 +534,10 @@ void inference_all_reduce(torch::Tensor& data, py::object op, py::object group,
         workspace[world_rank].state = 0;
     }
     if (world_rank != 0) {
-        wait_buffer_state_until(0, 0);
+        // if rank 0 spin too fast it could be in state 1 of next allreduce
+        // in this case wait_buffer_state_until(0, 0) may cause deadlock
+        // what we are certain is when rank 0 finishes the state won't be 2
+        wait_buffer_state_until_not(0, 2);
```
tjruwase, line 540: Can we please represent the states with symbolic constants with descriptive names to enhance readability and maintenance? Thanks!

delock: state changed to enum type.  Thanks for the suggestion!

PR #4053, opened 2023-07-27
Spread layers more uniformly when using partition_uniform

marcobellagente93:
The floor operation in the old implementation accumulates residuals which are all dumped in the last stage, potentially causing highly unbalanced partitions when both the residual and the number of pipeline stage is large.
The proposed change spreads out the layers more uniformly resulting in a more balanced partition.

As an example:

- old: partition_unform(num_items=38, num_parts=8) returns [0, 4, 8, 12, 16, 20, 24, 28, 38]
- new: partition_unform(num_items=38, num_parts=8) returns [0, 5, 10, 15, 20, 25, 30, 34, 38]


marcobellagente93:

mrwyattii:
@marcobellagente93 could you please run the precommit and commit any changes made?
```bash
pre-commit run --all-files
git add deepspeed/runtime/utils.py
git commit -m "formatting"
git push origin more-uniform-pipeline
```

tjruwase (approved):

PR #4077, opened 2023-08-01
Fix module parameter replacement for HybridEngine

XuehaiPan:
This PR fixes a bug for the module parameter replacement for HybridEngine. The original implementation runs a number of non-op assignments in a for-loop, which will not affect the original PyTorch module.

For example:

```python
for dst, src in [(mod.a, A), (mod.b, B), ...]:
    dst = src

for dst, src in [(mod.a, A), (mod.b, B), ...]:
    del dst
```

is not equal to:

```python
mod.a = A
mod.b = B
...

del mod.a
del mod.b
```

the line `dst = src` will only drop the reference to the original tensor. This statement only makes the variable `dst` point to a new object in the for-loop body.

This PR fixes this issue by calling `setattr` and `delattr` on the PyTorch module:

```python
for m, dst_name, src in [(mod, 'a', A), (mod, 'b', B), ...]:
    setattr(m, dst_name, src)

for m, dst_name, src in [(mod, 'a', A), (mod, 'b', B), ...]:
    delattr(m, dst_name)
```

NOTE: `nn.Module.__setattr__` and `nn.Module.__delattr__` magic methods have been modified in `nn.Module`.

### DIFF
--- a/deepspeed/module_inject/containers/features/gated_mlp.py
+++ b/deepspeed/module_inject/containers/features/gated_mlp.py
@@ -37,16 +37,19 @@ class HybridGatedMLPContainer(HybridEngineContainer):
         # Only need to alter behavior if we can't do the normal destructive copy
         if self.module.mlp.inter_w is None:
             params = [
-                (self.module.mlp.inter_up_w, self.inter_up_w),
-                (self.module.mlp.inter_up_b, self.inter_up_b),
-                (self.module.mlp.inter_gate_w, self.inter_gate_w),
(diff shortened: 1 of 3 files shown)


XuehaiPan:

RezaYazdaniAminabadi (approved):
Thanks for the PR, @XuehaiPan. The changes looks good to me 👍

awan-10:
@XuehaiPan - thanks for this PR! Can you please a client code or test that we can use to see the behavior? What exactly did you use to test/develop this?

PR #4088, opened 2023-08-04
Fix for lm head weights for models such as llama

puneeshkhanna:
Llama has tie word embeddings as False.
When loading from checkpoint, lm head weight is by default set to embedding weight. Add a check of tie word embeddings when setting lm head weight to embedding weight. In case of llama models, this updation of lm head weight with embedding weight won't happen and lm head weights will be correctly  set during load module recursive. Without this fix, llama models may give junk output also.


puneeshkhanna:
Can some one please review ?

polisettyvarma (commented):
LGTM

tjruwase (approved):

PR #4101, opened 2023-08-07
use correct ckpt path when base_dir not available

polisettyvarma:
results in incorrect ckpt path when base_dir not available


deepspeed/module_inject/replace_module.py
```diff
@@ -328,6 +328,8 @@ def replace_fn(child, _policy, layer_id=0, prefix="", state_dict=None):
         ckpt_mp_size = checkpoint_dict.get('tp_size', len(ckpt_list))
         ckpt_mp_size = checkpoint_dict.get('mp_size', ckpt_mp_size)
         base_dir1 = checkpoint_dict.get('base_dir', config.base_dir)
+        if base_dir1 is None:
+            base_dir1 = ""
```
mrwyattii, line 332: Perhaps we should make the default value be `""` in the inference config?
https://github.com/microsoft/DeepSpeed/blob/85dc854b11a66ec18e0f54e23eb72eeeddf9c81a/deepspeed/inference/config.py#L205

mrwyattii (approved):

PR #4102, opened 2023-08-07
add type checker ignore

EeyoreLee:
add type checker ignore to resolve that pylance can't resolved noqa annotation. Cause many people use vscode and Pylance. And Pylance isn't a linter, it's a static type checker. It can't support `# noqa`.  So I add some `# type: ignore` to resolve the type checker problems.


EeyoreLee:

mrwyattii (approved):

PR #4115, opened 2023-08-09
[CPU][Bugfix] Make uid and addr_port part of SHM name in CCL backend

delock:
This PR fix a potential bug in CCLBackend.   Previous the SHM name used for CCLBackend is a fixed name, this would cause conflict between different users or the same user launch multiple DeepSpeed instance.   This PR make uid and addr_port part of SHM name, thus fix the potential conflict.

### DIFF
--- a/csrc/cpu/comm/ccl.cpp
+++ b/csrc/cpu/comm/ccl.cpp
@@ -323,31 +323,30 @@ void initialize(int size, int rank, torch::Tensor& kvs_data)
     _ccl_comms.emplace_back(ccl::create_communicator(size, rank, kvs));
 
     auto addr_string = std::getenv("MASTER_ADDR");
-    if (addr_string == NULL) {
-        addr_string = "";
-    }
+    if (addr_string == NULL) { addr_string = ""; }
     auto port_string = std::getenv("MASTER_PORT");
-    if (port_string == NULL) {
-        port_string = "";
-    }
+    if (port_string == NULL) { port_string = ""; }
     char shm_name[NAME_BUF_SIZE];
-    snprintf(shm_name, NAME_BUF_SIZE, "%s_%d_%s_%s", SHM_BUFFER_NAME, getuid(), addr_string, port_string);
+    snprintf(shm_name,
+             NAME_BUF_SIZE,
+             "%s_%d_%s_%s",
+             SHM_BUFFER_NAME,
+             getuid(),
+             addr_string,
+             port_string);
     // create shared workspace for SHM based allreduce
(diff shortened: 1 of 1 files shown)


delock:
@tjruwase ideally deepspeed/comm should be called back when the workload is exiting, this gives communication library a chance to release all the resource gracefully, including named SHM which would only be released when system rebooted.  Is there such mechanism in DeepSpeed?

tjruwase:
@delock, did you have a question about graceful shutdown of deepspeed?

delock:
> @delock, did you have a question about graceful shutdown of deepspeed?

It would be helpful if deepspeed/comm/ccl.py can get a chance to exit gracefully.  I experiemented with CCLBackend.__del__() method but it seems the process termination mechanism in launch.py didn't allow CCLBackend.__del__() being called before process terminated.

As I experiment in UT.  __del__() of class will be called if process killed by SIGINT, but will not be called if process killed by SIGTERM, I wonder whether process exit of DeepSpeed can be further improved.

Currently, the SHM object will remain in /dev/shm after DeepSpeed exit, we don't expect it to be a big issue because there will be only one copy for each UID+ADDR+PORT combination.

tjruwase:
> As I experiment in UT. **del**() of class will be called if process killed by SIGINT, but will not be called if process killed by SIGTERM, I wonder whether process exit of DeepSpeed can be further improved.

Thanks for this question. You are correct that graceful shutdown is an area that needs improvement for DeepSpeed. We recently added this [method](https://github.com/microsoft/DeepSpeed/blob/0c75f4a3f937febc8c15610fcab7b81466b216c7/deepspeed/runtime/engine.py#L360-L362) to help, but it requires client call. We have experimented other automatic ideas without success. We are very open to any help here. Thanks!

delock:
> > As I experiment in UT. **del**() of class will be called if process killed by SIGINT, but will not be called if process killed by SIGTERM, I wonder whether process exit of DeepSpeed can be further improved.
> 
> Thanks for this question. You are correct that graceful shutdown is an area that needs improvement for DeepSpeed. We recently added this [method](https://github.com/microsoft/DeepSpeed/blob/0c75f4a3f937febc8c15610fcab7b81466b216c7/deepspeed/runtime/engine.py#L360-L362) to help, but it requires client call. We have experimented other automatic ideas without success. We are very open to any help here. Thanks!

Thanks for the link.  Currently this fix should be stable enough for multi-user and multi-instance environment.   This PR should be ready to merge from our side.

tjruwase (approved):

Issue #4116, opened 2023-08-09
[REQUEST] add ssh port argument for multi-node training/inference


Hiromasa-H:
**Is your feature request related to a problem? Please describe.**
Currently, the code assumes that the SSH port for connections is set to the default port 22 . However, in some cases, the SSH port might be configured to a value other than 22 for security reasons. This limitation prevents the utilization of the multi-node feature, as the code lacks the functionality to specify a custom port number for SSH connections.

**Describe the solution you'd like**
I would like to be able to specify the SSH port when using the multi-node feature, like so:
```bash
deepspeed --hostfile=hostfile.txt \
          --master_addr=$ip_address \
          --port=2222  \
          test.py 
```

**Describe alternatives you've considered**
An alternative is to reconfigure the SSH ports to be 22, but as mentioned earlier, this may not be possible due to security reasons.

**Additional context**
I have altered [deepspeed/launcher/runner.py](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/launcher/runner.py) to enable specification of SSH ports in the following PR: https://github.com/microsoft/DeepSpeed/pull/4117#issue-1843517576

Hiromasa-H:
https://github.com/microsoft/DeepSpeed/pull/4117#issue-1843517576

PR #4117, opened 2023-08-09
added port argument for ssh

Hiromasa-H:
This is the PR for issue https://github.com/microsoft/DeepSpeed/issues/4116.


deepspeed/launcher/runner.py
```diff
@@ -193,6 +193,8 @@ def parse_args(args=None):
                         "numbers and range. i.e. 1,3-5,7 => [1,3,4,5,7].  When not "
                         "specified, all cores on system would be used rank binding")
 
+    parser.add_argument("--port", type=int, default=22, help="SSH port to use for remote connections")
```
jeffra, line 196: Can we change this flag to be `--ssh_port` instead? I just want to make it clear which port we are defining here, since we also have `--master_port`.

deepspeed/launcher/runner.py
```diff
@@ -564,6 +566,7 @@ def main(args=None):
         else:
             cmd = runner.get_cmd(env, active_resources)
 
+    env["PDSH_SSH_ARGS_APPEND"] += f" -p {args.port} "
```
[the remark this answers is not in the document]
jeffra: I agree w. this suggestion from @loadams. The deepspeed launcher supports different multi-node backends, `pdsh` is the default and most common but it would be good to move the `PDSH_SSH_ARGS_APPEND` variable into the `PDSHRunner` class inside multinode_runner.py.

deepspeed/launcher/runner.py
```diff
@@ -432,7 +434,7 @@ def main(args=None):
     if multi_node_exec and not args.no_ssh_check:
         first_host = list(active_resources.keys())[0]
         try:
-            subprocess.check_call(f'ssh -o PasswordAuthentication=no {first_host} hostname',
+            subprocess.check_call(f'ssh -o PasswordAuthentication=no -p {args.port} {first_host} hostname',
```
jeffra, line 437: I think we only want to pass `-p {args.port}` iff --ssh_port is set. In many cases we have users setting up a [ssh config](https://phoenixnap.com/kb/ssh-config) that sets specific port/connection parameters. If we force passing `-p 22` as the default then this will break a ssh config host that may have a port that is not 22.

Hiromasa-H:

deepspeed/launcher/runner.py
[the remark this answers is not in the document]
Hiromasa-H: Thank you for your feedback. I've moved this part of the code into the PDSHRunner class as specified, and made the necessary changes.

deepspeed/launcher/runner.py
Hiromasa-H: Thank you for your feedback. I've added an if statement to make sure the `-p` is only added on when `args.ssh_port` is specified.

deepspeed/launcher/runner.py
Hiromasa-H: Apologies for the confusing name. Fixed as requested.

deepspeed/launcher/runner.py
```diff
@@ -193,6 +193,8 @@ def parse_args(args=None):
                         "numbers and range. i.e. 1,3-5,7 => [1,3,4,5,7].  When not "
                         "specified, all cores on system would be used rank binding")
 
+    parser.add_argument("--ssh_port", type=int, default=22, help="SSH port to use for remote connections")
```
jeffra, line 196: I think the later checks for `if args.ssh_port is not None` will never trigger since the default here is `22`. Can we set the default to `None`?

loadams:
Hi @Hiromasa-H - apologies for the slow review, wanted to test one thing on our side first, but will review this today.  Thanks for your patience.

deepspeed/launcher/runner.py
```diff
@@ -432,10 +434,16 @@ def main(args=None):
...
+            if args.ssh_port is not None:  # only specify ssh port if it is specified
+                subprocess.check_call(f'ssh -o PasswordAuthentication=no -p {args.ssh_port} {first_host} hostname',
+                                      stderr=subprocess.DEVNULL,
+                                      stdout=subprocess.DEVNULL,
+                                      shell=True)
+            else:
+                subprocess.check_call(f'ssh -o PasswordAuthentication=no {first_host} hostname',
+                                      stderr=subprocess.DEVNULL,
+                                      stdout=subprocess.DEVNULL,
+                                      shell=True)
```
mrwyattii, line 446:
```suggestion
            ssh_check_cmd = "ssh -o PasswordAuthentication=no "
            if args.ssh_port is not None:
                ssh_check_cmd += f"-p {args.ssh_port} "
            ssh_check_cmd +=f"{first_host} hostname"
            subprocess.check_call(ssh_check_cmd,
                                  stderr=subprocess.DEVNULL,
                                  stdout=subprocess.DEVNULL,
                                  shell=True)
```

mrwyattii (approved):
Let's avoid copying code with `subprocess.check_call` (see my suggestion).

Otherwise, LGTM

loadams (approved):

Hiromasa-H:
@loadams Sorry, I did not intend to rush you. Thank you for your taking your time with a through review.

PR #4120, opened 2023-08-09
tolerating missing optimizer states for MoE [2nd attempt]

### LINKED ISSUE #2737: [BUG] Loading MoE Checkpint with load_optimizer_states=False should tolerate missing optim_state files

clumsy:
Fixes #2737

When `load_optimizer_states=False` is used for MoE `load_checkpoint` - do not attempt to load the optimizer state files.

This currently fails as DeepSpeed still attempts to load those, even though they are not used afterwards.

Previous attempt: https://github.com/microsoft/DeepSpeed/pull/2738


clumsy:
For your consideration, @tjruwase.

clumsy:
@tjruwase is there on-going issue with HuggingFace integration tests? This looks unrelated to my changes.

tjruwase:
> @tjruwase is there on-going issue with HuggingFace integration tests? This looks unrelated to my changes.

I am not sure if there is an on-going issue with HF. Are you able to run the tests on your side?

deepspeed/runtime/engine.py
```diff
@@ -2710,26 +2710,29 @@ def _load_checkpoint(self,
...
-                largest_group_name = groups._get_max_expert_size_name()
-                expp_rank = groups._get_expert_parallel_rank(largest_group_name)
-                optim_load_path = self._get_optimizer_ckpt_name(load_dir, tag, expp_rank)
-                optim_checkpoint = self.checkpoint_engine.load(optim_load_path, map_location=torch.device('cpu'))
-            else:
-                optim_checkpoint = checkpoint
-
             has_zero_optimizer_state = self.zero_optimization() or self.bfloat16_enabled()
             if load_optimizer_states and self.optimizer is not None and not has_zero_optimizer_state:
+                if self.has_moe_layers:
```
tjruwase, line 2721: Since MoE is composable with ZeRO, I think this MoE logic should be independent of `has_zero_optimizer_state`.

clumsy:
> > @tjruwase is there on-going issue with HuggingFace integration tests? This looks unrelated to my changes.
> 
> 
> 
> I am not sure if there is an on-going issue with HF. Are you able to run the tests on your side?

I'll double-check.

deepspeed/runtime/engine.py
clumsy: I can change the code if you insist, @tjruwase. But then it won't be equivalent in terms of effects to what it was previously, i.e. we only `load_state_dict` for `self.optimizer` if it's `not has_zero_optimizer_state`.

clumsy: Or do you mean the fact that it's returned in `client_state` regardless?

clumsy:
@tjruwase HuggingFace tests work fine for me:

```
====== 1 failed, 42 passed, 94 skipped, 30 warnings in 296.84s (0:04:56) =======
```

The only failure is some hardware incompatibility:

```
self = <op_builder.async_io.AsyncIOBuilder object at 0x7f7371078880>
verbose = False

    def jit_load(self, verbose=True):
        if not self.is_compatible(verbose):
>           raise RuntimeError(
                f"Unable to JIT load the {self.name} op due to it not being compatible due to hardware/software issue. {self.error_log}"
            )
E           RuntimeError: Unable to JIT load the async_io op due to it not being compatible due to hardware/software issue. None
```

Triggering another build.

deepspeed/runtime/engine.py
tjruwase: @clumsy, apologies my initial comment was not accurate. Also, this code has evolved in a very weird way as you have observed. Please give me a bit more time to remind myself of the original intent of this code.

tjruwase: By the way, what happens if you pass `load_module_only=True`?

clumsy: Only model states will be loaded, but not LR schedule, data sampler and other auxiliary states. Optimizer state will be recreated from the model parameters, @tjruwase. See [here](https://github.com/microsoft/DeepSpeed/blob/19e9a7c0284cc886d84afd7ec58721af5e045811/deepspeed/runtime/engine.py#L2713)

deepspeed/runtime/engine.py
```diff
@@ -2710,26 +2710,29 @@ def _load_checkpoint(self,
...
+                    largest_group_name = groups._get_max_expert_size_name()
+                    expp_rank = groups._get_expert_parallel_rank(largest_group_name)
+                    optim_load_path = self._get_optimizer_ckpt_name(load_dir, tag, expp_rank)
+                    optim_checkpoint = self.checkpoint_engine.load(optim_load_path, map_location=torch.device('cpu'))
+                else:
+                    optim_checkpoint = checkpoint
+
                 if self.fp16_enabled():
```
clumsy, line 2729: @tjruwase do we need to add `self.bfloat16_enabled()` check here?

tjruwase: Yes, we do. Much of this code was written prior to bf16 support.

deepspeed/runtime/engine.py
tjruwase: I see. I thought you wanted just model states for inference, so never mind.

deepspeed/runtime/engine.py
```diff
@@ -2786,7 +2789,7 @@ def get_sparse_tensor_module_names(original_set, loaded_set, original_parameters
 
         client_state = {key: value for key, value in checkpoint.items() if not key in deepspeed_states}
 
-        if not load_optimizer_states and not load_module_only:
+        if optim_checkpoint is not None:
             client_state['optimizer'] = optim_checkpoint['optimizer']
```
tjruwase, line 2793: The original idea of `client_state` was to provide a convenient way for the checkpoint to contain both deepspeed-specific state, and client-specific state. Rather than the client creating a separate checkpoint file for its state, we designed `deepspeed.save_checkpoint(...)` to accept a `dict` containing client-specific state, so that deepspeed could checkpoint such state on behalf of the client, alongside the deepspeed-specific state. At load time, we identify `client_state` using this hardcoded [logic](https://github.com/microsoft/DeepSpeed/blob/7711bdbbd27c62ab4986f35c1ed01a0268fed92f/deepspeed/runtime/engine.py#L2795).  Unfortunately, at some point this [logic](https://github.com/microsoft/DeepSpeed/blob/7711bdbbd27c62ab4986f35c1ed01a0268fed92f/deepspeed/runtime/engine.py#L2797-L2798) was added to further modify the `client_state`. I am suspicious of this change since `optimizer` is almost always `deepspeed_state` and not `client_state. Nevertheless, can you please revert this test condition for backwards-compatibility concerns. I don't think it should affect the correctness of your changes, since `optim_checkpoint=None` will only be True if `load_module_only == True`. Am I correct?

clumsy: I see, reverting my change then. My understanding was that `client_state` is something we return to the user to indicate what state has been used for which part in the end.

clumsy:
Addressed your feedback, @tjruwase.

deepspeed/runtime/engine.py
tjruwase: > I see, reverting my change then. My understanding was that `client_state` is something we return to the user to indicate what state has been used for which part in the end.

Thanks! Apologies for this confusing feature. The following docs might explain a bit more: 

https://deepspeed.readthedocs.io/en/latest/model-checkpointing.html#deepspeed.DeepSpeedEngine.save_checkpoint
https://deepspeed.readthedocs.io/en/latest/model-checkpointing.html#deepspeed.DeepSpeedEngine.load_checkpoint

clumsy: Unfortunately, reverting this bit breaks several tests, @tjruwase and here's why I think. For `load_module_only=False` and `load_optimizer_states=False` it will attempt to access `optim_checkpoint`, but I'm not quite sure why. Didn't we specify not to use optimizer states - to refresh it from model weights? The intention of this change is to avoid loading the optimizer states since they are not used and hence don't return it via `client_state`.

clumsy: There are other inconsistencies by the way. I thought about including the changes for them as well, but decided to do it in a separate PR. E.g. `deepspeed_states` adds `data_sampler` (key) unconditionally, but it's only loaded into state dict if present in `checkpoint`.

tjruwase: Please undo the revert. Your original change is at least meaningful compared to the old code. The entire function needs a rewrite :(.

tjruwase (approved):

clumsy:
@tjruwase thanks for approving this PR, 3 (seemingly unrelated) HuggingFace tests still fail. Should I keep retrying, do you know if it's being looked at or a known issue?

PR #4118, opened 2023-08-09
feat(activation_checkpointing): add `non_reentrant_checkpoint` to support inputs require no grad

hughpu:
The added function is union of `torch.utils.checkpoint._checkpoint_without_reentrant` and `CheckpointFunction` in `checkpointing` module. 

 * this aim to solve the back probagation error raised from all input requires no grad, such as first layer of normal training, lora training until layer get lora injected, traditional fintune with only last layer trainable, etc... 
 * `_checkpoint_without_reentrant` has already been implemented in pytorch for a while, the solution is stable at most time except for jit script module.
 * can help to solve the issue which is hacked by `deepspeed.runtime.pipe.module.PipelineModule._is_checkpointable`
 * unit test added and proved to work, also tested with pipeline module lora tuning project as `activation_checkpoint_func` without specifing `checkpointable_layers`

PR #4128, opened 2023-08-10
Fix requires_grad of input must be true for activation checkpoint layer in pipeline train.

inkcherry:
mechanism FYI https://github.com/prigoyal/pytorch_memonger/blob/master/tutorial/Checkpointing_for_PyTorch_models.ipynb?short_path=e1e8ee7#L282C244-L282C257

For discrete input, could make activation for the embedding layer.
For image input,  no need to calculate and save grads for input.

Have verified on megatron-deepspeed train/cifar10 train ut (test_pipe. py). Under the same seed, the data of parameter and gradient updates is exactly equal to before this addition.

It can reduce  small memory on rank0  (stage0 contains embedding layer)


deepspeed/runtime/pipe/engine.py
```diff
@@ -638,7 +638,6 @@ def _exec_forward_pass(self, buffer_id):
 
         # Zero out the gradients each time we use the tensor because only the data in
         # tensor changes across batches
-        self._zero_grads(inputs)
```
tohtana, line 641: Can you explain why we can delete this? We could also delete the comment before this line.

deepspeed/runtime/pipe/module.py
```diff
@@ -21,6 +21,13 @@
 from deepspeed.runtime.state_dict_factory import SDLoaderFactory
 from deepspeed.accelerator import get_accelerator
 from deepspeed.checkpoint.utils import clone_tensors_for_torch_save
+from enum import Enum
+
+
+class CkptLayer_Enum(Enum):
+    not_ckpt_layer = 0
+    normal_ckpt_layer = 1
+    warp_ckpt_layer = 2
```
tohtana, line 30: Does warp mean wrap?

deepspeed/runtime/pipe/module.py
```diff
@@ -320,7 +343,7 @@ def forward(self, forward_input):
         # will see a different offset.
         self.micro_offset += 1
 
-        def exec_range_func(start, end):
+        def exec_range_func(start, end, warp_layer=False):
```
tohtana, line 346: Do you mean wrap, not warp?

deepspeed/runtime/pipe/module.py
```diff
@@ -342,7 +365,29 @@ def exec_func(*inputs):
                     inputs = layer(inputs)
                 return inputs
 
-            return exec_func
+            def exec_func_warp(*inputs):
```
tohtana, line 368: Can we consolidate this with exec_func?
Most of the code are duplicated.

tohtana (commented):
Thank you for submitting this PR, @inkcherry!
Is the goal of this to stop creating gradients of inputs when it is unnecessary?

Can you also clarify intentions of some changes?

inkcherry:
> Thank you for submitting this PR, @inkcherry! Is the goal of this to stop creating gradients of inputs when it is unnecessary?
> 
> Can you also clarify intentions of some changes?

thanks for your review and apologies for not explaining clearly.

Yes, currently,  requires_grads of input of the first checkpoint layer must be true. I think this is a limitation, and the current code may have 2 following logic to work around this limitation in two scenarios.  please correct me if I'm wrong.
 
1 For image input, set ```requireds_ grad=true```  for input from the dataset, this will calculate and save grads for the input image, which I think may cost memory and computing time with large size input. related issue.
https://github.com/AGI-Collective/multimodal/issues/16
 
 
2 Secondly, for LLM, notice the comment https://github.com/microsoft/DeepSpeed/blob/64c670ef021118ed42a98d5dde249f35878a07a7/deepspeed/runtime/pipe/module.py#L621, it  hooks transformer layer(force make embedding layer as a non-checkpoint layer, 
which take input with requires_grad=False and ouput tensor with requires_grad=True). 
 
For this limitation, we could pass a dummy input that requires grad but isn't necessarily used in computation  FYI.
https://github.com/prigoyal/pytorch_memonger/blob/master/tutorial/Checkpointing_for_PyTorch_models.ipynb?short_path=e1e8ee7#L282C244-L282C257  , here wrapping a ``WrapModule``` at the first layer to achieve this.

deepspeed/runtime/pipe/module.py
inkcherry: fixed.

deepspeed/runtime/pipe/module.py
inkcherry: fixed.

deepspeed/runtime/pipe/engine.py
inkcherry: https://github.com/microsoft/DeepSpeed/blob/64c670ef021118ed42a98d5dde249f35878a07a7/deepspeed/runtime/pipe/engine.py#L620-L623  here inputs has become a non-leaf tensor, and its gradient has not been saved. if ```if self.is_pipe_partitioned and not self.is_first_stage()``` is False, not create a new leaf tensor named input,  access inputs' gradient in _zero_grad will trigger a warning.

deepspeed/runtime/pipe/module.py
inkcherry: merged.

inkcherry:
@tohtana , I've made some changes to the code. Any suggestions?  would appreciate your feedback.

tohtana:
Hi @inkcherry, thank you for the fixes. Overall, this looks okay to me. 

On the other hand, I just reviewed #4118 and this might solve a part of the issue you addressed. I didn't notice this PR until very recently, and I'm sorry that I couldn't share it with you. 

We want to merge the PR, and I am wondering if we can simplify the changes in this PR using the new features proposed in #4118. Can you share your thoughts on this?

inkcherry:
@tohtana, yes I agree, https://github.com/microsoft/DeepSpeed/pull/4118 is great！
I could try based on  https://github.com/microsoft/DeepSpeed/pull/4118 merged when it is merged.

tjruwase:
@inkcherry, #4118 is now merged. Also, could you please add a unit test when you retry? Thanks!

inkcherry:
moved to https://github.com/microsoft/DeepSpeed/pull/4224
close this one

PR #4130, opened 2023-08-10
Zero infinity xpu support

Liangliang-Ma:
Currently AIO kernel has gap using on XPU device. 
This PR fixes issues that 
1. cpu buffer only detect CUDA tensor in aio threads 
2. make is_pinned() more general in all kinds of accelerator
3. io_getevents() stopped by signal on XPU

Developed and Tested on Megatron Deepspeed.


csrc/aio/py_lib/deepspeed_aio_thread.cpp
```diff
@@ -22,17 +22,44 @@ io_op_desc_t::io_op_desc_t(const bool read_op,
...
-    _contiguous_buffer = _cpu_buffer.contiguous();
+    const char *ds_acc_env_var = "DS_ACCELERATOR";
+    if (_buffer.is_cuda())
+    {
+        _cpu_buffer = _buffer.to(torch::kCPU).pin_memory();
+        _contiguous_buffer = _cpu_buffer.contiguous();
+    }
+    else if (std::getenv(ds_acc_env_var) == "xpu")
+    {
+        _cpu_buffer = _buffer.to(torch::kCPU);
```
tjruwase, line 36: Can you clarify that xpu requires 2 copies compared to 1 copy for cuda?

csrc/aio/py_lib/deepspeed_py_aio_handle.cpp
```diff
@@ -173,7 +173,7 @@ void deepspeed_aio_handle_t::_stop_threads()
 
 int deepspeed_aio_handle_t::wait()
 {
-    assert(_num_pending_ops > 0);
+    // assert(_num_pending_ops > 0);
```
tjruwase, line 176: Why disable this assert?

Liangliang-Ma:
Some change seems not necessary. I wonder if it's model related. For example, `assert(_num_pending_ops > 0);` should not raise error if thread's queue works fine. But in Megatron-deepspeed it fails. For `io_pgetevents`, I will test it more to reply you. Thanks.

csrc/aio/py_lib/deepspeed_py_aio_handle.cpp
Liangliang-Ma: I canceled this disabling. It was model related issue.

Liangliang-Ma:
Hi @tjruwase , I did a modification to current PR.  Removed env variable dependence. Let me explain some change that this PR has.

1) use `io_pgetevents` instead of `io_getevents`
It's indeed more general and it sometimes can help to block signals from interrupting the io process. The performance of these two are same. 

2)More copy on XPU
This is because Async IO requires aligned mem to be used as buffer. Currently it needs manually set to allocate an aligned place, otherwise io would fail. This could be optimized in future.

3) changes in `_swap_out_unpinned_tensors`
In `deepspeed/runtime/swap_tensor/partitioned_optimizer_swapper.py`, `def swap_in_optimizer_state` read data into buffer whose sizes are all `swap_info.numel()`. But for `def _swap_out_unpinned_tensors` in `def swap_out_optimizer_state`, write buffer sizes are numel of each tensor. This could be a mismatch in some cases.

tjruwase:
> 2)More copy on XPU This is because Async IO requires aligned mem to be used as buffer. Currently it needs manually set to allocate an aligned place, otherwise io would fail. This could be optimized in future.
> 
Requiring aligned buffers is a deliberate design decision to simplify the library and ensure high performance.  We prefer for the alignment issues to be addressed on the client side where there is more flexibility. For example, clients can easily add padding in a way that minimizes perf impact. However, I think the library needs to be improved to assert on unaligned buffers to make this requirement more explicit.

tjruwase:
> 3\. changes in `_swap_out_unpinned_tensors`
>     In `deepspeed/runtime/swap_tensor/partitioned_optimizer_swapper.py`, `def swap_in_optimizer_state` read data into buffer whose sizes are all `swap_info.numel()`. But for `def _swap_out_unpinned_tensors` in `def swap_out_optimizer_state`, write buffer sizes are numel of each tensor. This could be a mismatch in some cases.

Can you please share an example of this failure case, perhaps in the form of a unit test?

Liangliang-Ma:
> > 2)More copy on XPU This is because Async IO requires aligned mem to be used as buffer. Currently it needs manually set to allocate an aligned place, otherwise io would fail. This could be optimized in future.
> 
> Requiring aligned buffers is a deliberate design decision to simplify the library and ensure high performance. We prefer for the alignment issues to be addressed on the client side where there is more flexibility. For example, clients can easily add padding in a way that minimizes perf impact. However, I think the library needs to be improved to assert on unaligned buffers to make this requirement more explicit.

The buffers that get used in swap in and out are all created and initialized in deepspeed python codes. In client side we cannot affect it by changing model codes. By the way, I see that in [link](https://github.com/microsoft/DeepSpeed/blob/430510bfce5c8e2e8f2108b73824aef4a37fcaf1/csrc/aio/py_test/ds_aio_handle.py#L34) deepspeed test has a flag that can create aligned tensor which also calls `posix_memalign` to initialize buffer. And this flag seems not get used in deepspeed. What about adding it to let client choose how to create their buffer?

Liangliang-Ma:
> > 3. changes in `_swap_out_unpinned_tensors`
> > In `deepspeed/runtime/swap_tensor/partitioned_optimizer_swapper.py`, `def swap_in_optimizer_state` read data into buffer whose sizes are all `swap_info.numel()`. But for `def _swap_out_unpinned_tensors` in `def swap_out_optimizer_state`, write buffer sizes are numel of each tensor. This could be a mismatch in some cases.
> 
> Can you please share an example of this failure case, perhaps in the form of a unit test?

In my env of Megatron-deepspeed with Adam on xpu devices, when run to `def _update_param_state_info`, the optimizer state that get added to `swap_info` can be a shape of `[1, 22843136, 22843136]`, which will result in mismatch of swap in&out optimizer states. Because in swap_in the buffer is set to `[22843136, 22843136, 22843136]` and `[256, 22843136, 22843136]` in swap_out(aligned).  There is an assert to check this mismatch in [link](https://github.com/microsoft/DeepSpeed/blob/430510bfce5c8e2e8f2108b73824aef4a37fcaf1/csrc/aio/py_lib/deepspeed_py_aio_handle.cpp#L228).

Liangliang-Ma:
@tjruwase Any comments? Thanks!

tjruwase:
> The buffers that get used in swap in and out are all created and initialized in deepspeed python codes. In client side we cannot affect it by changing model codes.

Actually, the expectation is that the client side send aligned buffers to the library. This is why the tensor swapping utility of deepspeed has many alignment [codes](https://github.com/microsoft/DeepSpeed/blob/180dd39714a79b1d30bdecfe5c1b2ffe37b30152/deepspeed/runtime/swap_tensor/optimizer_utils.py#L374) to guarantee this condition.  Can you share a bit about your client side? Perhaps, deepspeed can expose another wapper to provide this alignment. Would that help?

tjruwase:
> By the way, I see that in [link](https://github.com/microsoft/DeepSpeed/blob/430510bfce5c8e2e8f2108b73824aef4a37fcaf1/csrc/aio/py_test/ds_aio_handle.py#L34) deepspeed test has a flag that can create aligned tensor which also calls `posix_memalign` to initialize buffer. And this flag seems not get used in deepspeed. What about adding it to let client choose how to create their buffer?

Yes, good observation. Our plan is to use this functionality within deepspeed and also expose to clients. We hope to do that soon.

tjruwase:
> In my env of Megatron-deepspeed with Adam on xpu devices, when run to `def _update_param_state_info`, the optimizer state that get added to `swap_info` can be a shape of `[1, 22843136, 22843136]`, which will result in mismatch of swap in&out optimizer states. Because in swap_in the buffer is set to `[22843136, 22843136, 22843136]` and `[256, 22843136, 22843136]` in swap_out(aligned).

I am very curious about how this mismatch could happen. It concerns me that our swapping logic is making wrong assumptions. Is it possible to share more details of the training run? Or is it possible to create unit test? Thanks!

Liangliang-Ma:
> I am very curious about how this mismatch could happen. It concerns me that our swapping logic is making wrong assumptions. Is it possible to share more details of the training run? Or is it possible to create unit test? Thanks!

I find the root cause of this mismatch now. The reason is that Deepspeed assumes that optimizer state tensors have same numel and so swap_in&swap_out can use each tensors' numel or param's numel as length. But the truth is different optimizer have different situation. It could be different numel.

For apex.optimizers.FusedAdam, the optimizer states have same numel as param: [apex optimizer states init](https://github.com/NVIDIA/apex/blob/master/apex/optimizers/fused_adam.py#L171)
For torch.optim.Adam, the optimizer states have three tensor and the first one is scalar: [torch adam](https://github.com/pytorch/pytorch/blob/main/torch/optim/adam.py#L98)

I found that my previous change is naive and not enough. So I fallback to use sgd optimizer whose state has same numel as param to bypass this issue. For this part, it can be in another PR or do you want to fix it? @tjruwase

tjruwase:
> I add the alignment in get_accelerator(), which will be easy for different platform to create or adjust their buffer and don't

This solution is fine with me. Thanks!

tjruwase:
> I found that my previous change is naive and not enough. So I fallback to use sgd optimizer whose state has same numel as param to bypass this issue. For this part, it can be in another PR or do you want to fix it? @tjruwase

Thanks for the amazing debugging effort to discover the root issue. I think it is better to fix in another PR. Are you able to help fix this?

csrc/aio/py_lib/deepspeed_aio_thread.cpp
```diff
@@ -24,7 +24,7 @@ io_op_desc_t::io_op_desc_t(const bool read_op,
       _num_bytes(num_bytes),
       _validate(validate)
 {
-    _cpu_buffer = _buffer.is_cuda() ? _buffer.to(torch::kCPU).pin_memory() : _buffer;
+    _cpu_buffer = (_buffer.is_cuda() | _buffer.is_xpu()) ? _buffer.to(torch::kCPU).pin_memory() : _buffer;
```
tjruwase, line 27: `|` -> `||`

deepspeed/runtime/swap_tensor/async_swapper.py
```diff
@@ -37,7 +38,8 @@ def has_buffers(self):
 
     def add_buffers(self, buffer_list):
         assert len(self.all_buffers) == 0
-        assert all([buffer.is_pinned() for buffer in buffer_list])
+        assert all(
+            [get_accelerator().is_pinned(buffer) or get_accelerator().is_aligned(buffer) for buffer in buffer_list])
```
tjruwase, line 42: Can you please explain why `is_aligned()` is an alternative for `is_pinned()`? We require `pinned()` buffers in the swapping logic because they are page-locked and so can be used for DMA. Are `aligned()` buffers guaranteed to work with DMA? If so, don't they have to be page-locked as well?

accelerator/abstract_accelerator.py
```diff
@@ -222,7 +222,11 @@ def LongTensor(self):
         ...
 
     @abc.abstractmethod
-    def pin_memory(self, tensor):
+    def pin_memory(self, tensor, align=False):
```
tjruwase, line 225: After further thought, I think the second API is better because it allows the clients to use other alignments besides the default choice of the accelerator.

accelerator/cpu_accelerator.py
```diff
@@ -222,9 +222,12 @@ def IntTensor(self):
     def LongTensor(self):
         return torch.LongTensor
 
-    def pin_memory(self, tensor):
+    def pin_memory(self, tensor, align=False):
```
tjruwase, line 225:
```suggestion
    def pin_memory(self, tensor, alignment_bytes=1):
```

Liangliang-Ma:
> After reviewing your alignment changes, I have shared some questions. One idea is that perhaps we could extend the existing `pin_memory()` API to take an alignment flag as well. So that way, clients can request pinned buffers that are also aligned. Also, accelerators would have flexibility in handling such requests. What do you think?
> 
> Below are two possible APIs. The second one seems more general. I am curious if this could help your work. Thanks.

@tjruwase Good idea. In this way we can implement alignment in `pin_memory `and reduce swap code complexity. I choose the first API because I think it's better get `sysconf(_SC_PAGESIZE)` as align bytes rather than let user do the setting. So then `is_align()` is not needed anymore. The code is done.

accelerator/cuda_accelerator.py
```diff
@@ -203,9 +203,12 @@ def IntTensor(self):
     def LongTensor(self):
         return torch.cuda.LongTensor
 
-    def pin_memory(self, tensor):
+    def pin_memory(self, tensor, align=False):
```
tjruwase, line 206:
```suggestion
    def pin_memory(self, tensor, alignment_bytes=1):
```

accelerator/mps_accelerator.py
```diff
@@ -186,9 +186,12 @@ def IntTensor(self):
     def LongTensor(self):
         return
 
-    def pin_memory(self, tensor):
+    def pin_memory(self, tensor, align=False):
```
tjruwase, line 189:
```suggestion
    def pin_memory(self, tensor, alignment_bytes=1):
```

accelerator/npu_accelerator.py
```diff
@@ -188,9 +188,12 @@ def IntTensor(self):
     def LongTensor(self):
         return torch.npu.LongTensor
 
-    def pin_memory(self, tensor):
+    def pin_memory(self, tensor, align=False):
```
tjruwase, line 191:
```suggestion
    def pin_memory(self, tensor, alignment_bytes=1):
```

Liangliang-Ma:
> > I found that my previous change is naive and not enough. So I fallback to use sgd optimizer whose state has same numel as param to bypass this issue. For this part, it can be in another PR or do you want to fix it? @tjruwase
> 
> Thanks for the amazing debugging effort to discover the root issue. I think it is better to fix in another PR. Are you able to help fix this?

Yes sure. I can do that.

tjruwase:
> @tjruwase Good idea. In this way we can implement alignment in `pin_memory `and reduce swap code complexity. I choose the first API because I think it's better get `sysconf(_SC_PAGESIZE)` as align bytes rather than let user do the setting. So then `is_align()` is not needed anymore. The code is done.

Sorry :). I think the second API is better so that clients have the option of different alignments, such as super pages.

Liangliang-Ma:
Another thing is that in previous workflow checks, I found that `io_pgetevents` cannot be found in your unittest env when building AIO ops. I am not sure if it's because the testing env is outdated or any other reason. Could you pls help to check with it? @tjruwase 
reference code: [libaio](https://pagure.io/libaio/blob/master/f/src/io_pgetevents.c)

tjruwase:
> It's ok:). I think we can modify it like this to let clients use default page size:
> 
> ```
>     @abc.abstractmethod
>     def pin_memory(self, tensor, align_bytes=1, sc_page_align=False):
> ```

I think it could be confusing to provide two flags to control the alignment bytes. But I also see the value of allowing users to request the default system alignment without having to know the number of bytes. So how about we adjust the second API so that `-1` value for `alignment_bytes` means to use system default such as `sysconf(_SC_PAGESIZE)`? We will document this behavior. What do you think?

Liangliang-Ma:
> I think it could be confusing to provide two flags to control the alignment bytes. But I also see the value of allowing users to request the default system alignment without having to know the number of bytes. So how about we adjust the second API so that `-1` value for `alignment_bytes` means to use system default such as `sysconf(_SC_PAGESIZE)`? We will document this behavior. What do you think?

Yes, I agree. It will be clear more.

Liangliang-Ma:
> > Another thing is that in previous workflow checks, I found that `io_pgetevents` cannot be found in your unittest env when building AIO ops.
> 
> I will pay attention to the new workflow for this issue.

@tjruwase Thanks! btw can you help to trigger workflow once again? Seems it stops when I push new commits.:)

csrc/aio/common/deepspeed_aio_common.cpp
```diff
@@ -115,10 +115,9 @@ static int _do_io_complete(const long long int min_completes,
                            std::vector<std::chrono::duration<double>>& reap_times)
 {
     const auto start_time = std::chrono::high_resolution_clock::now();
-    const auto n_completes = io_getevents(
-        aio_ctxt->_io_ctxt, min_completes, max_completes, aio_ctxt->_io_events.data(), nullptr);
+    long long int n_completes = io_pgetevents(
+        aio_ctxt->_io_ctxt, min_completes, max_completes, aio_ctxt->_io_events.data(), NULL, NULL);
```
tjruwase, line 119:
```suggestion
        aio_ctxt->_io_ctxt, min_completes, max_completes, aio_ctxt->_io_events.data(), nullptr, nullptr);
```

Will this work?

deepspeed/runtime/swap_tensor/partitioned_param_swapper.py
```diff
@@ -104,10 +104,11 @@ def _configure_aio(self, ds_config):
...
         self.reserved_buffer_ids = []
-        self.buffers = get_accelerator().pin_memory(
-            torch.empty(int(self.aligned_elements_per_buffer * self.param_buffer_count),
-                        dtype=self.dtype,
-                        requires_grad=False))
+        self.buffers = get_accelerator().pin_memory(torch.empty(int(self.aligned_elements_per_buffer *
+                                                                    self.param_buffer_count),
+                                                                dtype=self.dtype,
+                                                                requires_grad=False),
+                                                    align_bytes=-1)
```
tjruwase, line 111:
```suggestion
                                                    align_bytes=0)
```

deepspeed/runtime/swap_tensor/utils.py
```diff
@@ -184,7 +184,8 @@ def __init__(self, num_elems, count, dtype):
         self.count = count
         self.dtype = dtype
         self.all_buffers = [
-            get_accelerator().pin_memory(torch.zeros(num_elems, device='cpu', dtype=dtype)) for _ in range(count)
+            get_accelerator().pin_memory(torch.zeros(num_elems, device='cpu', dtype=dtype), align_bytes=-1)
```
tjruwase, line 187:
```suggestion
            get_accelerator().pin_memory(torch.zeros(num_elems, device='cpu', dtype=dtype), align_bytes=0)
```

csrc/aio/common/deepspeed_aio_common.cpp
Liangliang-Ma: Yes, it works. Absolutely.

Liangliang-Ma:
> @Liangliang-Ma, this PR looks good to me after addressing the final suggestions. Thanks so much for this great work.

@tjruwase I committed the change as you suggested. Thanks for your reviewing these days!

tjruwase (approved):

tjruwase:
> > Another thing is that in previous workflow checks, I found that `io_pgetevents` cannot be found in your unittest env when building AIO ops.
> 
> I will pay attention to the new workflow for this issue.

I think this is failing because the unittest env is ubuntu 18.04 which does not have `io_pgetevents` available. Is it possible for the code to use `io_getevents` or `io_pgetevents` as appropriate? I will also investigate if ubuntu version of the unittest can be updated. Thanks!

loadams:
> > > Another thing is that in previous workflow checks, I found that `io_pgetevents` cannot be found in your unittest env when building AIO ops.
> > 
> > 
> > I will pay attention to the new workflow for this issue.
> 
> I think this is failing because the unittest env is ubuntu 18.04 which does not have `io_pgetevents` available. Is it possible for the code to use `io_getevents` or `io_pgetevents` as appropriate? I will also investigate if ubuntu version of the unittest can be updated. Thanks!

This should be updated to be 20.04 now.  cc: @tjruwase and @Liangliang-Ma 

Is there somewhere we should specify a min libaio version/Ubuntu version required for this?

tjruwase:
@loadams, thanks for updating the CI ubuntu. I think the package version is tied to the OS version, so we need a way to specify minimum Ubuntu version. Unfortunately, I don't know how to do that. I am not sure we have done this previously. 

@mrwyattii, @jeffra any thoughts here? Thanks!

tjruwase:
@Liangliang-Ma, it looks like CI issues are resolved, except for formatting. Can you please address that? Thanks!

loadams:
> @loadams, thanks for updating the CI ubuntu. I think the package version is tied to the OS version, so we need a way to specify minimum Ubuntu version. Unfortunately, I don't know how to do that. I am not sure we have done this previously.
> 
> @mrwyattii, @jeffra any thoughts here? Thanks!

I think we should at a minimum put this in the README in the requirements section?  Also perhaps list the failure mode if one doesn't have this?

tjruwase:
> I think we should at a minimum put this in the README in the requirements section? Also perhaps list the failure mode if one doesn't have this? Thanks! 

Great point @loadams. @Liangliang-Ma, could please replace `io_submit` with `io_pgetevents` in the following compatibility checker function?
https://github.com/microsoft/DeepSpeed/blob/28b9d5c2313addde94a1aa17c09eeb97cb792d7e/op_builder/async_io.py#L76-L82

Liangliang-Ma:
@tjruwase Thanks for solving CI issue! Checker function and formatting modification is committed.

Liangliang-Ma:
@tjruwase seems this PR has been blocked by a cpu inference unit test in merge queue, which is out of this PR's scope. Can you please help to check it? Thanks!

loadams:
> @tjruwase seems this PR has been blocked by a cpu inference unit test in merge queue, which is out of this PR's scope. Can you please help to check it? Thanks!

Hi @Liangliang-Ma - this is a known issue, should be resolved in [this PR](4430) then we can unblock the merge queue.

Issue #4017, opened 2023-07-22
[BUG] Loss scale already at minimum - Training LlaMA2 7B via HF+deepspeed consistently fails


scorixear:
**Describe the bug**
When training the LLaMA2 7B HF Model with deepspeed on a single-node multi-gpu setup,
the loss_scale gets decreased consistently to 1 (minimum) and exits with error.

`Exception: Current loss scale already at minimum - cannot decrease scale anymore. Exiting run.`

This appears both with ZeRO Stage2 + CPU Offload and ZeRO Stage3 + CPU Offload.

**To Reproduce**
- DeepSpeed with ZeRO Stage 2 + CPU offload.
- HF Trainer (v4.32.0.dev0)

**Expected behavior**
Training Completes

**ds_report output**
```txt
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
 [WARNING]  async_io: please install the libaio-devel package with yum
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
async_io ............... [NO] ....... [NO]
cpu_adagrad ............ [NO] ....... [OKAY]
cpu_adam ............... [NO] ....... [OKAY]
fused_adam ............. [NO] ....... [OKAY]
fused_lamb ............. [NO] ....... [OKAY]
quantizer .............. [NO] ....... [OKAY]
random_ltd ............. [NO] ....... [OKAY]
 [WARNING]  please install triton==1.0.0 if you want to use sparse attention
sparse_attn ............ [NO] ....... [NO]
spatial_inference ...... [NO] ....... [OKAY]
transformer ............ [NO] ....... [OKAY]
stochastic_transformer . [NO] ....... [OKAY]
transformer_inference .. [NO] ....... [OKAY]
--------------------------------------------------
DeepSpeed general environment info:
torch install path ............... ['/software/all/staging/PyTorch/1.12.1-foss-2022a-CUDA-11.7.0/lib/python3.10/site-packages/torch']
[...]

PR #4141, opened 2023-08-13
Fixes for training models with bf16 + freshly initialized optimizer via `load_module_only`

haileyschoelkopf:
This PR makes some fixes to the case where we want to resume training from a DeepSpeed ZeRO checkpoint and initialize a new optimizer, while not using the old optimizer in the checkpoint or relying on its existence at all.

in this situation, despite passing `load_module_only=True` and `load_optimizer_states=False` to `load_checkpoint()`, the previous behavior was that:
- `self._load_zero_checkpoint` would still be called, which attempts to load from the (in this case, nonexistent) checkpoint files. This PR stops this function from being called if using `load_module_only=True` and `load_optimizer_states=False`.  Alternatively, calling this function may be alright if `"load_from_fp32_weights": true` is set in the DeepSpeed ZeRO config (reference: https://github.com/microsoft/DeepSpeed/blob/ff7d5275f2aa916cb5f320e0d817154e96f9cdb6/deepspeed/runtime/engine.py#L733) but this parameter does not seem to be documented in the docs for ZeRO config dicts.
- in `_load_checkpoint`, the following codeblock: 
```
if self.optimizer is not None and self.fp16_enabled():
    self.optimizer.refresh_fp32_params()
```
results in `self.optimizer.refresh_fp32_params()` being called only if using FP16. As a result, the FP32 optimizer state is never initialized from the 16-bit model weights. This PR removes the fp16-specific condition.

Previously reported in:
https://github.com/EleutherAI/gpt-neox/issues/947
https://github.com/EleutherAI/gpt-neox/issues/843

Should also close:
https://github.com/microsoft/DeepSpeed/issues/4017

Fixes: #4944 and #4017

This caused problems for a freshly-converted LLama checkpoint, which did not contain optimizer states, when trying to train with this model as initialization. I have confirmed the following fixes prevent this behavior.

cc @Quentin-Anthony @zhangir-azerbayev


janelu9:
my model was trained in bf16 mode ,when I load ckpt with `load_optimizer_states = False`, it still try to load it

Quentin-Anthony:
@tjruwase and @jeffra -- Want any more detail or testing from our side? This fix resolved a lot of issues on our end, and we suspect non-neox users may face it too.

tjruwase:
@haileyschoelkopf and @Quentin-Anthony, apologies for dropping the ball on this. The PR LGTM. Please resolve the conflict so we can merge. Thanks for the contribution!

tjruwase (approved):

haileyschoelkopf:
Thanks @tjruwase , merge conflicts resolved!

exnx:
How come this PR was never merged?  It fixed the finetuning bug I experienced myself, which took about 3-4 days...

Specifically, removing the extra condition to check for fp16_enabled is not needed (and hurts bf16 for example).


```
before:
if self.optimizer is not None and self.fp16_enabled():

after:
if self.optimizer is not None:
```

This should do the trick.

PR #4157, opened 2023-08-16
fix autoTP bloom inference when require_tp_fused_qkvw=True

dc3671:
### Problem

Currently, for `bloom-7b1` and `bloom-176B`, the autoTP inference is broken, because this PR https://github.com/microsoft/DeepSpeed/pull/4040 missed some split process when `require_tp_fused_qkvw=True`.

### Solution

I added back the split logic and refactored some code there, so that the logic is cleaner and simpler.

can you help review this? @mjksmith @jeffra cc @molly-smith

### DIFF
--- a/deepspeed/module_inject/auto_tp.py
+++ b/deepspeed/module_inject/auto_tp.py
@@ -335,28 +335,31 @@ class AutoTP():
             if require_tp_fused_qkvw(name, self.mp_size):
                 #for detecting fused type
                 module_str = str(self.module).strip()
-                #The copy is a regular copy, The shape of dst and src is the same
                 data = prepare_tp_fused_qkvw(module_str, child.weight.data, self.mp_size, mp_replace.gpu_index)
-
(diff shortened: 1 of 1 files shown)


inkcherry:
@dc3671 , There was a copy function here that would automatically split input when src shape is not equal to dst shape.  prepare_tp_fused_qkvw will split weight, but for bloom, I don't change its initial path using split logic in copy, but this copy seems have been removed.
so I think we'd better add a line 
```return input [gpu_index * dst_shape: (gpu_index+1) * dst_shape]```here
https://github.com/microsoft/DeepSpeed/blob/341cefd2a406f578ee05b3a9d0686553926a7e61/deepspeed/module_inject/fusedqkv_utils.py#L69


also in this line, should ```return _bloom_type_transpose```
https://github.com/microsoft/DeepSpeed/blob/341cefd2a406f578ee05b3a9d0686553926a7e61/deepspeed/module_inject/fusedqkv_utils.py#L92

PR #4167, opened 2023-08-17
do allgather only in shared optimizer states groups

inkcherry:
notice their only one way (BF16_Optimizer) for bf16+pipeline train.
pipeline parallel with bf16(use zero1 optimizer) will default call this even if dp size = 1.

some time we may only try mp/pp>1 with dp=1 , this may cost some time.


tjruwase:
@inkcherry, thanks for this PR. Note that this is a utility function is also used by zero stage 1 & 2, so this benefits more cases than bf16_optimizer.

tjruwase (approved):

PR #4172, opened 2023-08-18
Cleanup for ROCm6.0 re-org changes

BLOrange-AMD:


op_builder/cpu_adagrad.py
```diff
@@ -39,9 +39,5 @@ def include_paths(self):
         elif not self.is_rocm_pytorch():
             CUDA_INCLUDE = [os.path.join(torch.utils.cpp_extension.CUDA_HOME, "include")]
         else:
-            CUDA_INCLUDE = [
```
loadams, line 42: Hi @BLOrange-AMD - thanks for this, but this will break compatibility for anyone using ROCm < 6.0 once that is released, correct?

jithunnair-amd: > Hi @BLOrange-AMD - thanks for this, but this will break compatibility for anyone using ROCm < 6.0 once that is released, correct?

Hi @loadams, yes these changes should be backward-compatible with multiple previous ROCm versions, please refer https://rocm.docs.amd.com/en/latest/understand/file_reorg.html#wrapper-header-files

jithunnair-amd: > And how this will fit in with the changes in this PR: https://github.com/microsoft/DeepSpeed/pull/4114/files

> And how this will fit in with the changes in this PR: https://github.com/microsoft/DeepSpeed/pull/4114/files

The PR 4114 changes shouldn't be impacted by this PR at all, since that PR doesn't use any of the ROCm include paths, and the `-lcurand` doesn't apply to ROCm regardless.

loadams: I see, so this should have no impact on the MI200 tests?

loadams:
@jithunnair-amd and @BLOrange-AMD - was trying to test this on our side but having issues given the apex install mentioned here: https://github.com/ROCmSoftwarePlatform/apex/issues/115

op_builder/cpu_adagrad.py
jithunnair-amd: @loadams Sorry I missed this. Yes, this shouldn't impact the MI200 tests

loadams:
Closing as these changes were merged in #4537

PR #4192, opened 2023-08-22
Allow modification of zero partitioned parameters

### LINKED ISSUE #3830: How to modify weights during training in a deepspeed stage 3 model
This issue is more of a general question than a bug report: we are working with deepspeed stage 3 and model parallelization (training 10+ billion params LLMs). We need to be able to modify weights every N steps during training, but outside of the regular optimization process. 

I've seen multiple ways of exporting and accessing weights from deepspeed models, but I haven't found any methods to modify them directly during training. 

One option would be during saving/loading - but although I can load the weights using `get_fp32_state_dict_from_zero_checkpoint`, and possibly modify the state dict, I also can't find a way to reload this modified state dict into the model and continue training with the new set of weights. 

I am currently using deepspeed as the strategy in Fabric (pytorch-lightning). Appreciate the help or any pointers.

tjruwase:
Utilities for flexible modification of partitioned fp32 parameters and optimizer states. 

Fix #3830


tohtana (approved):
This is a great feature! The code modifications and the document are also clear. I do have one observation, though it's not immediately pressing:

Currently we have three `get_*` functions (`safe_get_full_fp32_param`, `safe_get_full_grad`, and `safe_get_full_optimizer_state`). This PR introduces `safe_set_full_fp32_param` and `safe_set_full_optimizer_state`. Is there a specific reason we're omitting safe_set_full_grad?
Maintaining consistency in the APIs can help users understand the design better.

deepspeed/runtime/zero/stage3.py
```diff
@@ -2122,36 +2123,57 @@ def get_fp32_grad_for_param(self, param) -> Tensor:
...
         if self._swappable_optimizer_subgroup(group_idx):
             self._optimizer_states_and_gradient_swap_out(group_idx)
+
         return hp_param
 
+    def set_full_hp_param(self, value, param, optim_state_key=None):
+        if not param.requires_grad:
+            return
+
+        assert value.numel() == param.ds_numel, f'{value.numel()} != {param.ds_numel}'
```
mrwyattii, line 2165:
```suggestion
        assert value.numel() == param.ds_numel, f" Number of elements do not match: {value.numel()} != {param.ds_numel}"
```

mrwyattii (approved):

tjruwase:
> This is a great feature! The code modifications and the document are also clear. I do have one observation, though it's not immediately pressing:
> 
> Currently we have three `get_*` functions (`safe_get_full_fp32_param`, `safe_get_full_grad`, and `safe_get_full_optimizer_state`). This PR introduces `safe_set_full_fp32_param` and `safe_set_full_optimizer_state`. Is there a specific reason we're omitting safe_set_full_grad? Maintaining consistency in the APIs can help users understand the design better.

@tohtana, thanks for this valid question. I am delaying support for `safe_set_full_grad` until there is explicit request for it because it is harder to implement and I have limited bandwidth to think through all the design issues :(. I will add a TODO for this. Thanks for the review.

mayank31398:
@tjruwase 
can we set gradients also this way?
This is required for using Megatron's sequence parallel instead of Ulysees.
Wanted to check if this is possible anyhow and would be super-useful.

mayank31398:
Essentialy whats needed:
```python
grad = get_grads(layernorm.weight)
dist.all_reduce(grad, group=tp_group)
safe_set_grads(grad, layernorm.weight)
```

if there is an alternative way to do it, that will also be helpful.

PR #4193, opened 2023-08-22
Fix the path in tutorial

kytimmylai:
The path in the tutorial no longer matches the current structure in DeepSpeedExamples.


kytimmylai:

conglongli (approved):

PR #4198, opened 2023-08-23
DeepSpeed Ulysses release

samadejacobs:
DeepSpeed-Ulysses release- sequence parallelism support in DeepSpeed


awan-10 (approved):

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -825,6 +826,7 @@ def __init__(
             config_dict_or_path = config
             logger.warning(
                 f'zero.Init: the `config` argument is deprecated. Please use `config_dict_or_path` instead.')
+        print(f"config_dict_or_path: {config_dict_or_path}")
```
jeffra, line 829: can we either remove this print or give more context as to why it's printed? also if we need it printed we might only want it on rank 0, right?

jeffra (approved):

PR #4203, opened 2023-08-23
fixes for 12.1 windows compiler

earonesty:
had to make these changes to get visual studio to compile deepspeed as well as some changes to the torch cpp_extension.py (which didn't include c10 lib in the linker args).  all the torch libs use int64_t and not some special torch_long_int or somethng.  imo, it sholuld probably be the latter.


earonesty:
> @microsoft-github-policy-service agree

csrc/quantization/pt_binding.cpp
```diff
@@ -9,6 +9,7 @@
 #include <vector>
 #include "quantization.h"
 
+
```
loadams, line 12: You'll want to remove this additional new line to fir our formatting guidelines

csrc/quantization/pt_binding.cpp
```diff
@@ -192,16 +192,16 @@ std::vector<at::Tensor> quantized_reduction(at::Tensor& input_vals,
                               .layout(at::kStrided)
                               .device(at::kCUDA)
                               .requires_grad(false);
-    const int scales_elems = (quantize::requires_offset(quant_type)) ? 2 : 1;
-    auto scales = torch::empty({out_groups, scales_elems}, scales_options);
+    const int64_t scales_elems = (quantize::requires_offset(quant_type)) ? 2 : 1;
+    auto scales = torch::empty({(int64_t)out_groups, scales_elems}, scales_options);
```
loadams, line 196: @earonesty - could you give samples of the errors you were hitting that caused you to change these?

loadams: And was this to build all ops other than those not suppored on Windows (aio)?

earonesty: yes, this was everything i needed.   it was mostly "template ambiguity" errors.   there are multiple torch::empty funcs in the header. i can remove them and get the exact errors.

earonesty: that being said, code should probably always use int64_t when interacting with torch tensors, as a general rule

PR #4206, opened 2023-08-24
fix lm head overriden issue, move it from checkpoint in-loop loading …

sywangyi:
…to out loop

### DIFF
--- a/deepspeed/module_inject/replace_module.py
+++ b/deepspeed/module_inject/replace_module.py
@@ -296,6 +296,15 @@ def replace_transformer_layer(orig_layer_impl, model, checkpoint_dict, config, m
 
         return new_module
 
+    def set_lm_head(module):
+        embedding_weight = None
+        for n, p in module.named_parameters():
+            if "word_embeddings." in n or "embed_tokens." in n or "wte." in n:
+                embedding_weight = p
+        if embedding_weight is not None and hasattr(module, "lm_head") and hasattr(
+                module.lm_head, "weight") and module.lm_head.weight.is_meta:
+            module.lm_head.weight = embedding_weight
+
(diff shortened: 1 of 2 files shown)


delock:
What problem caused by inner loop loading? @sywangyi

sywangyi:
r_module.lm_head.weight will be wrongly pointed to word_embeddings buffer since it's in the last sharded checkpoint. acturally in llama70b, there's lm_head weight in the checkpoint. so lm head and work embedding should point to different buffers.

yao-matrix:
This issue will lead to llama2-70B model generate wrong texts in both CPU and Habana Gaudi.

delock:
GPT-J would also have problem when checkpoints are locally sharded, and could be fixed by this PR.

delock:
@sywangyi Your PR can fix a correctness bug in Llama2-7b when loaded with from_config, can you rebase to master and resolve conflcit?

Hi @RezaYazdaniAminabadi, could this PR be merged into master?  There is a bug in master branch with Llama2 from_config loading and can be fixed by this PR.  We will open an issue soon.

deepspeed/module_inject/replace_module.py
```diff
@@ -386,6 +393,13 @@ def replace_fn(child, _policy, layer_id=0, prefix="", state_dict=None):
                                                container=container_g)
                     sds = [None for _ in sds]
                     gc.collect()
+        embedding_weight = None
```
tjruwase, line 396: Could this be refactored into a function? It looks like duplicate of other change.

sywangyi: done

tjruwase (approved):

Issue #4183, opened 2023-08-21
[BUG] llm zero2 training 'NoneType' object has no attribute 'data'


YooSungHyun:
**Describe the bug**
when i used bf16 or fp16 and zero2. error strange error accured. (for full-finetune llama-2-7b)

``` json
{
    "bf16": {
        "enabled": "auto"
    },
    "optimizer": {
        "type": "Adam",
        "params": {
            "lr": "auto",
            "adam_w_mode": true
        }
    },
    "scheduler": {
        "type": "OneCycle",
        "params": {
            "cycle_first_step_size": 100,
            "cycle_max_lr": 3e-5,
            "cycle_min_lr": 3e-6
        }
    },
    "zero_optimization": {
        "stage": 2,
        "allgather_partitions": true,
        "allgather_bucket_size": 5e7,
        "overlap_comm": true,
        "reduce_scatter": true,
        "reduce_bucket_size": 5e7,
        "contiguous_gradients": true,
        "offload_optimizer": {
            "device": "cpu",
            "pin_memory": true,
            "buffer_count": 4,
            "fast_init": false
        }
    },
    "train_micro_batch_size_per_gpu": "auto",
    "gradient_accumulation_steps": "auto",
    "wall_clock_breakdown": false
}
```

huggingface llm example,
not using peft lora. i just want to full finetune

llama-2-7b
`circulus/Llama-2-7b-orca-v1`

accured error
``` python3
Traceback (most recent call last):
  File "/data/workspace/train/train_dpo.py", line 133, in <module>
    main(model_args=model_args, dataset_args=dataset_args, training_args=training_args)
  File "/data/workspace/train/train_dpo.py", line 119, in main
    dpo_trainer.train()
  File "/data/workspace/.bart_venv/lib/python3.10/site-packages/transformers/trainer.py", line 1539, in train
    return inner_training_loop(
  File "/data/workspace/.bart_venv/lib/python3.10/site-packages/transformers/trainer.py", line 1809, in _inner_training_loop
    tr_loss_step = self.training_step(model, inputs)
  File "/data/workspace/.bart_venv/lib/python3.10/site-packages/transformers/trainer.py", line 2665, in training_step
    self.accelerator.backward(loss)
  File "/data/workspace/.bart_venv/lib/python3.10/site-packages/accelerate/accelerator.py", line 1847, in backward
    self.deepspeed_engine_wrapped.backward(loss, **kwargs)
[...]

PR #4207, opened 2023-08-24
Fix AttributeError: 'NoneType' object has no attribute 'data' in grad_accum

jp1924:
Fix: #4183

I think that when self.async_accumulate_grad_in_cpu_via_gpu(param), only grad_accum should be computed, so it would be nice to check for the presence of grad_accum in that part.


jomayeri:
Thanks for this. This issue is addressed in a separate PR https://github.com/microsoft/DeepSpeed/pull/4214

PR #4216, opened 2023-08-24
Allow TiedLayerSpec to have multiple tied weights

zphang:
Follow-up from (and directly reusing code from): https://github.com/microsoft/DeepSpeed/pull/2035

This PR allows `TiedLayerSpec` to support a list of tied weights. This will allow whole layers to be reused.

Re: https://github.com/microsoft/DeepSpeed/pull/2035#issuecomment-1196817705, `tied_weight_attrss` has two Ss intentionally, as it is a dictionary of lists. (Happy to rename to something else too.)

Credit: @thomasw21

### DIFF
--- a/deepspeed/runtime/pipe/module.py
+++ b/deepspeed/runtime/pipe/module.py
@@ -9,6 +9,7 @@ import glob
 import re as regex
 
 from functools import partial
+from typing import List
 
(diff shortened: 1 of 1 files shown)


zphang:

PR #4228, opened 2023-08-28
fix: fixed the communication problem of pp when using sequence parallel

LiuXTao:
This commit fixes the problem raised in the issue #4196. This commit adds a "disable_partitioned_tensor" parameter to control not using PartitionedTensor.

To support sequence parallelism, it is necessary to avoid using PartitionedTensor for splitting the tensor, as it would cause the value of tensor to be messed up when the activations and gradients have already been split along the sequence dimension during sequence parallel processing. 

Moreover, sequence parallelism also splits based on the tensor parallel size, so the communication overhead for activations and gradients will not increase. Additionally, this approach reduces the time required for an all_gather operation after sending the partitioned activations and gradients.


puneeshkhanna:
@LiuXTao  - Would setting is_grad_partioned and is_pipe_partitioned to False do the same for sequence parallelism?

PR #4230, opened 2023-08-29
ROCm and ROCm Apex fixes for AMD workflows

loadams:


.github/workflows/amd-mi200.yml
```diff
@@ -45,6 +45,7 @@ jobs:
         run: |
           git clone https://github.com/ROCmSoftwarePlatform/apex.git
           cd apex
+          git checkout torch_2.1_higher
```
loadams, line 48: This branch comes from the suggestion [here](https://github.com/ROCmSoftwarePlatform/apex/issues/115)

loadams: @rraminen and @jithunnair-amd - thoughts on this change?

jithunnair-amd: This branch is currently where we have the pytorch2.1-compatible changes. However, we were planning to move these changes into master branch so that it works with tip-of-tree upstream PyTorch.

jithunnair-amd: cc @pruthvistony @ramcherukuri

jithunnair-amd: @loadams We were planning on doing the branch refactor next week, but if we want to move forward here, let's merge this PR, and we can file another PR post-refactor

loadams: Makes sense, no need for this, we can wait until the changes are merged back into the master branch, thanks for the update.

loadams: @jithunnair-amd - no rush, I was just trying to make things work for the MI200 tests to be all green, and this was one thing we would need, but we can wait until these changes are back in master - I'll close this PR for now and we can re-open if we need.

PR #4251, opened 2023-09-01
Pin Triton version to >=2.0.0 and <2.1.0

lekurile:
The latest [2.1.0 release](https://pypi.org/project/triton/2.1.0/#history) of Triton includes some API changes, so this PR pins the version to `triton>=2.0.0,<2.1.0` while the API is being updated on the DeepSpeed side.


requirements/requirements-triton.txt
```diff
@@ -1 +1 @@
-triton>=2.0.0
+triton<2.1.0
```
loadams, line 1: Should we specify both, `>=2.0.0,<2.1.0`?

mrwyattii: Yes, we need to specify both

lekurile: Updated.

loadams (approved):

PR #4258, opened 2023-09-04
fix zero_sd_list judgment

zhangsheng377:
when dp_world_size is 0, zero_sd_list will return empty list [], then the judgment is wrong.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -2863,7 +2863,7 @@ class DeepSpeedEngine(Module):
                     "currently supported.")
             checkpoint_folder = None
             zero_sd_list = self._get_all_zero_checkpoints(load_dir, tag)
-            if zero_sd_list is None:
+            if not zero_sd_list:
                 return False
 
         self.optimizer.load_state_dict(state_dict_list=zero_sd_list,


tjruwase:
@zhangsheng377, can you please explain how dp_world_size can be 0?

zhangsheng377:
> 



> @zhangsheng377, can you please explain how dp_world_size can be 0?

we use megatron-deepspeed, when tp=1 and pp=1, the dp_world_size will be 0.

tjruwase:
> [self.loaded_checkpoint_dp_world_size = checkpoint['dp_world_size']](https://github.com/microsoft/DeepSpeed/blob/55d9964c59c0c6e23158b5789a5c36c28939a7b0/deepspeed/runtime/engine.py#L2760) here, actually, the checkpoint's dp_world_size is 0.

Thanks for the response. So the next question is how did the checkpoint get `dp_world_size == 0`. Are you able to recreate the checkpoint? Perhaps, this is a very old checkpoint and `dp_world_size` is actually missing. Can you please provide some context?

zhangsheng377:
> > [self.loaded_checkpoint_dp_world_size = checkpoint['dp_world_size']](https://github.com/microsoft/DeepSpeed/blob/55d9964c59c0c6e23158b5789a5c36c28939a7b0/deepspeed/runtime/engine.py#L2760) here, actually, the checkpoint's dp_world_size is 0.
> 
> Thanks for the response. So the next question is how did the checkpoint get `dp_world_size == 0`. Are you able to recreate the checkpoint? Perhaps, this is a very old checkpoint and `dp_world_size` is actually missing. Can you please provide some context?

yeah, we run a script which transform huggingface model to deepspeed.

tjruwase:
> yeah, we run a script which transform huggingface model to deepspeed.

Thanks for the response. That conversion script needs to create `dp_world_size` entry in the checkpoint similar to what DeepSpeed does [here](https://github.com/microsoft/DeepSpeed/blob/542dc0d5cb58d9aba78eceb4510296eddbaa2e1f/deepspeed/runtime/engine.py#L3199).

zhangsheng377:
> > yeah, we run a script which transform huggingface model to deepspeed.
> 
> Thanks for the response. That conversion script needs to create `dp_world_size` entry in the checkpoint similar to what DeepSpeed does [here](https://github.com/microsoft/DeepSpeed/blob/542dc0d5cb58d9aba78eceb4510296eddbaa2e1f/deepspeed/runtime/engine.py#L3199).

Thanks, but if merge the patch, it will be compatible to the old checkpoint .

tjruwase:
> Thanks, but if merge the patch, it will be compatible to the old checkpoint .

The problem with the patch is that `dp_world_size = 0` is confusing, since `dp_world_size >=1` even for single GPU.

zhangsheng377:
> confusing

ok. thanks

PR #4259, opened 2023-09-05
fix llama lm_head load wrong weight error

baodii:
Hi, 
I have found that when we do AutoTP shard loading, if there are more than one checkpoint file, we will get a wrong lm_head weight in llama model. 

Because when we replace module we will use embedding_weight as lm_head.weight in advance. In other word, we use embedding_weight as lm_head.weight before we load lm_head.weight when we have more than one checkpoint file. 

So I add is_last_cp flag to avoid this situation. 

image
image

### DIFF
--- a/deepspeed/module_inject/replace_module.py
+++ b/deepspeed/module_inject/replace_module.py
@@ -305,7 +305,8 @@ def replace_transformer_layer(orig_layer_impl, model, checkpoint_dict, config, m
                                              orig_class=orig_layer_impl,
                                              replace_fn=replace_fn,
                                              _replace_policy=config.injection_policy_tuple,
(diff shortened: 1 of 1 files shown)


baodii:
@delock @Yejing-Lai

dc3671:
I think it's kind of duplicated with https://github.com/microsoft/DeepSpeed/pull/4206 . But I don't know which way is better.

delock:
Hi @baodii  @sywangyi  can you converge these two PRs together? #4259 and #4206 ?   These two PRs fix similiar issue.  We need to have a single PR that fix all different cases.

baodii:
same as #4206

Issue #3967, opened 2023-07-15
[BUG] Difference in CLI variable expansion single-node vs multi-node


BramVanroy:
I am launching my training jobs through a shell script which in turn submits a PBS job to our cluster. This allows me to set some extra parameters and pass accept extra arguments to the shell script, e.g. `sh myscrtipt.sh -e "--arg1 hello --arg2`", which I can then pass to the pbs script as an environment variable. The pbs script then interpolates it to start a Python command. The command under scrutinty in the pbs file looks like this:

```sh
deepspeed --hostfile=$OUTPUT_DIR/hostfile train.py --deepspeed ds_config_zero2.json ${EXTRA_ARGS}
```

where `${EXTRA_ARGS}` is, as per the example, the string `--arg1 hello --arg2`.

The shell should expand this correctly into different chunks that Python can parse correctly. And this works _but only if I use more than one node_. Interestingly, when I use just a single node, my EXTRA_ARGS are not correctly interpolated - instead they are seen as a single string `"--arg1 hello --arg2". So it seems that under-the-hood deepspeed is passing the CLI arguments to different nodes differently compared to when there is just one node. But I have not figured out where that distinction is made in the code.

Potentially related: https://github.com/microsoft/DeepSpeed/issues/3961

mrwyattii:
@BramVanroy could you please try with #4007?

I believe the issue is that with multi-node, we are doing the following:
https://github.com/microsoft/DeepSpeed/blob/0a0819b78554a5130b7c1b31d18779d39421dade/deepspeed/launcher/multinode_runner.py#L64

And for single-node, we are doing this:
https://github.com/microsoft/DeepSpeed/blob/0a0819b78554a5130b7c1b31d18779d39421dade/deepspeed/launcher/runner.py#L503

The PR I linked to above match the multi-node behavior on single-node.

mrwyattii:
On closer inspection from the unit tests failures with #4007, my original fix was wrong. I've added the proper fix to #4007 now.

The problem is actually with python/argparse. We can replicate the behavior outside of DeepSpeed:
```python
# example.py
import sys
import argparse
print(sys.argv)

parser = argparse.ArgumentParser()
parser.add_argument("--arg1", type=str)
parser.add_argument("--arg2", action="store_true")
args = parser.parse_args()
print(args)
```
Then if we run the following, we can see that the `EXTRA_ARGS` are parsed as a single string:
```bash
[...]

PR #4265, opened 2023-09-05
fix user args parsing of string with spaces on runner

YudiZh:
After updating my DeepSpeed version and rerunning my program, I noticed that string arguments with spaces in user_args are not being parsed correctly as before. For example:
```shell
deepspeed --include=localhost:0,1 --master_port=12345 generate.py --arg1 "Hello world!"
```
It will cause error that `error: unrecognized arguments: world!`, i think it is caused by [#4007](https://github.com/microsoft/DeepSpeed/pull/4007). It concates all the user args then split to solve the [bug](https://github.com/microsoft/DeepSpeed/issues/3967). I have fixed the above issue by implementing a conditional check.

### DIFF
--- a/deepspeed/launcher/runner.py
+++ b/deepspeed/launcher/runner.py
@@ -390,7 +390,7 @@ def main(args=None):
     args = parse_args(args)
 
     # For when argparse interprets remaining args as a single string
-    args.user_args = shlex.split(" ".join(args.user_args))
+    args.user_args = shlex.split(" ".join(list(map(lambda x: x if x.startswith("-") else f"'{x}'", args.user_args))))
 
     if args.elastic_training:
         assert args.master_addr != "", "Master Addr is required when elastic training is enabled"


deepspeed/launcher/runner.py
```diff
@@ -390,7 +390,7 @@ def main(args=None):
     args = parse_args(args)
 
     # For when argparse interprets remaining args as a single string
-    args.user_args = shlex.split(" ".join(args.user_args))
+    args.user_args = shlex.split(" ".join(list(map(lambda x: x if x.startswith("-") else f"'{x}'", args.user_args))))
```
loadams, line 393: @mrwyattii - any reason why in #4007 we didn't copy the multinode_runner logic [here](https://github.com/microsoft/DeepSpeed/blob/0a0819b78554a5130b7c1b31d18779d39421dade/deepspeed/launcher/multinode_runner.py#L64)?

mrwyattii: Nope, but this change looks to be the proper fix

mrwyattii (approved):

loadams (approved):

PR #4279, opened 2023-09-06
Refactor unit test modeling

loadams:
Re-implementation of #2605.


loadams:
@jeffra and @mrwyattii - the CVE in the linked PR would be more impactful if the tar files weren't just for our unit tests, so curious if we want to merge this change or not.

mrwyattii (approved):

tests/unit/modeling.py
mrwyattii, line 1: Most of the code in this file and `modelingpreln` are completely unused. Let's remove unused code (including the `safe_extract`)

mrwyattii (changes requested):

PR #4288, opened 2023-09-07
Keep hpz secondary tensor in forward pass

HeyangQin:
Keep zero++ hpz secondary tensor in forward pass (when user enables zero++ hpz in json config) so we can use it if the user uses no_grad.


awan-10 (approved):

deepspeed/runtime/zero/stage3.py
```diff
@@ -1855,7 +1855,7 @@ def _post_step(self, timer_names):
         if self.swap_optimizer:
             self.optimizer_swapper.log_timers()
 
-        self.invalidate_secondary_tensor()
+        # self.invalidate_secondary_tensor() # given that we want hpz in forward pass when no_grad is set, we need to keep the secondary tensor
```
tjruwase, line 1858: I am not sure about this. Could the weight update by step not invalidate the cached secondary tensor values?

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -1501,23 +1501,22 @@ def _partition_param_sec(self, param, buffer=None, has_been_updated=False):
...
-                if not param.requires_grad and self.quantized_nontrainable_weights:
-                    secondary_partitioned_tensor, secondary_partitioned_tensor.ds_quant_scale = self.quantizer_module.quantize(
-                        secondary_partitioned_tensor)
-                secondary_partitioned_tensor.requires_grad = False
-                param.ds_secondary_tensor = secondary_partitioned_tensor
-                param.ds_secondary_tensor.ds_numel = secondary_partition_size
-                param.ds_secondary_tensor.status = PartitionedParamStatus.AVAILABLE
-                param.ds_secondary_tensor.final_location = final_location
+            final_location = None
+            secondary_partitioned_tensor = torch.empty(secondary_partition_size,
```
tjruwase, line 1505: @HeyangQin, perhaps I misunderstood our chat. I thought secondary tensor should only be created on a forward pass of a training iteration. Currently, partition_param is called on every forward and backward, which means secondary tensor is created in more cases than required. I would like to chat more about this. Thanks!

PR #4308, opened 2023-09-12
clear redundant timers

starkhu:


tjruwase:
@starkhu, this does not seem redundant as it makes it possible for client code to probe timers on other ranks.

starkhu:
> @starkhu, this does not seem redundant as it makes it possible for client code to probe timers on other ranks.
hello, tjruwase!
i just trying  to clear up out-dated events on other ranks. in more details, it is no necessary to save all previous iteration events in the current iteration , otherwise, 'len(self.times('train_batch').event_times )' is gradually increasing on other ranks。

tjruwase:
@starkhu, I see your point. You are correct. Thanks!

tjruwase (approved):

PR #4310, opened 2023-09-12
Suppress noise

tjruwase:
Suppress noisy prints.


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -389,13 +389,6 @@ def __init__(self,
             self.params_not_in_partition.append(params_not_in_partition)
             self.first_offset.append(first_offset)
 
-        for rank in range(dist.get_world_size()):
-            if dist.get_rank() == rank:
-                print(
-                    f"Rank: {rank} partition count {self.partition_count} and sizes{[(p.numel(), self.is_moe_param_group[i] if hasattr(self, 'is_moe_param_group') else False) for i,p in enumerate(self.single_partition_of_fp32_groups)]} "
-                )
-                dist.barrier()
```
loadams, line 397: The barrier isn't needed anymore and was just to synchronize the prints?

tjruwase: Yes, it was for the print. Although all ranks execute the full loop, each rank only prints once, and executes the barrier once to synchronize.

loadams (approved):

PR #4312, opened 2023-09-12
[BUG] partition_balanced return wrong result.

zjjMaiMai:
# Background

In pipeline parallelism, deepspeed uses `ds_utils.partition_balanced` to balance the partitioning of the model according to the number of parameters or class names.
https://github.com/microsoft/DeepSpeed/blob/581e44dd1ab3c409a5905335867c761d5cb4db5b/deepspeed/runtime/pipe/module.py#L380-L395

# What wrong?
```
>>> import deepspeed
>>> deepspeed.__version__
'0.10.3+542dc0d5'
>>> from deepspeed.runtime import utils as ds_utils
>>> ds_utils.partition_balanced([1, 1, 1, 1, 1], 4)
[0, 2, 4, 5, 5]
>>> 
```
the result [0, 2, 4, 5, 5] means [2, 2, 1, 0] layers for each part, which is not balanced at all. the last part will throw an exception because there are no parameters to training.

i add some unit test for this function, and i will fix it later if anyone need it.


zjjMaiMai:

zjjMaiMai:
already fixed! cc @tjruwase

ShadenSmith (approved):

ShadenSmith:
Thanks for this PR, @zjjMaiMai! 

A note on balance: the objective function in the original code minimizes the *maximum* load per partition. The maximum load on a pipeline stage determines the pipeline throughput, and so the original result is also balanced. Example paper: [Fast Optimal Load Balancing Algorithms for 1D Partitioning](https://digital.library.unt.edu/ark:/67531/metadc782652/m2/1/high_res_d/835143.pdf)

PR #4319, opened 2023-09-13
fix parser bug in launcher/runner.py

caijizhuo:
Is this a bug? It is a string instead of "None" type. If it is a str="None", it will go through in runner.py:504

```
        if args.enable_each_rank_log:
            deepspeed_launch.append(f"--enable_each_rank_log={args.enable_each_rank_log}")
```

### DIFF
--- a/deepspeed/launcher/runner.py
+++ b/deepspeed/launcher/runner.py
@@ -162,7 +162,7 @@ def parse_args(args=None):
                         "Useful when launching deepspeed processes programmatically.")
 
     parser.add_argument("--enable_each_rank_log",
-                        default="None",
+                        default=None,
                         type=str,
                         help="redirect the stdout and stderr from each rank into different log files")
 


mrwyattii:
@caijizhuo this is not a bug.
1. the dtype for `enable_each_rank_log` is `str`, so it makes sense to have it as `"None"` rather than `None`
2. The value of this arg is passed via command line to another process, and we cannot pass the `None`, so it will end up being the string `"None"` anyway: https://github.com/microsoft/DeepSpeed/blob/581e44dd1ab3c409a5905335867c761d5cb4db5b/deepspeed/launcher/runner.py#L505
3. We have a check for `args.enable_each_rank_log != "None"` in `launch.py`, rather than `runner.py`: https://github.com/microsoft/DeepSpeed/blob/581e44dd1ab3c409a5905335867c761d5cb4db5b/deepspeed/launcher/launch.py#L206

PR #4322, opened 2023-09-13
add option to disable pipeline partitioning

nelyahu:
added pipeline configuration to deepspeed config under pipeline section pipe_partitioned = "auto"
grad_partitioned = "auto"
can be used to enable or disable activations and grads partitioning in TP/PP mode


nelyahu:
@ShadenSmith @jeffra @duli2012 @tjruwase - can you please review?

nelyahu:
@tjruwase Hi :) can someone help reviewing this change?

deepspeed/runtime/config.py
```diff
@@ -446,6 +446,8 @@ def get_pipeline_config(param_dict):
         "partition": "best",
         "seed_layers": False,
         "activation_checkpoint_interval": 0,
+        "pipe_partitioned": "auto",
```
tjruwase, line 449: Why use "auto" as a value to represent True rather than simply make this a boolean config?

nelyahu: @tjruwase agreed, see updated change

tjruwase (approved):

loadams (approved):

loadams:
@tjruwase - should we add this to any docs we have?

tjruwase:
> @tjruwase - should we add this to any docs we have?

@loadams, yes, we should. But I don't believe we have any docs for DeepSpeed pipeline parallelism support.

PR #4323, opened 2023-09-13
Remove symlinks

mrwyattii:
We have seen several issues pop up (notably with Windows installation) related to the symbolic links that the DeepSpeed repository uses for `csrc`, `op_builder`, and `accelerator`. In this PR I am removing them to help avoid these issues. This PR:
- Replaces symlinks with the actual directories (in doing so, moving the directories noted above into the `deepspeed/` directory
- Adds a `prepend_path` variable to the `OpBuilder` class that allows us to easily change where `csrc` is located within the project directory structure

TODO:
- [ ] Verify all pre-compilation works
- [ ] Verify all JIT compilation works
- [ ] Verify Windows installation


mrwyattii:
@delock this PR has many changes that would affect pre-compilation and JIT. Do you have any feedback / concerns about how this would affect the abstract accelerator and intel CPU work you are doing?

FurkanGozukara:
so how do we install on Windows?

I tried like this and failed

windows 10 , Python 3.10.11

image

delock:
Hi @mrwyattii , I think one concern is op_builder is currently referenced at build time.  At build time if build script trying to import deepspeed.ops.op_builder.*** it would go through __init__.py of all levels up in which it assumes deepspeed is already installed.   This caused various issue and is the reason op_builder/accelerator/csrc has symlink.

Probably a big `if` in multiple __init__.py file may help get around this.  Another thinking is DeepSpeed build process could avoid operate on op_builder (no prebuilt mode), this could make the siutation much simpler.

> @delock this PR has many changes that would affect pre-compilation and JIT. Do you have any feedback / concerns about how this would affect the abstract accelerator and intel CPU work you are doing?

PR #4325, opened 2023-09-13
Pipe engine: Start batch_train timers only for rank 0

nelyahu:
it is being accessed only by rank0.
resources are held by all other rank, with no reason

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -348,14 +348,15 @@ class PipelineEngine(DeepSpeedEngine):
         self._compute_loss = True
 
         # Do the work
-        self.timers(TRAIN_BATCH_TIMER).start()
+        if self.global_rank == 0:
+            self.timers(TRAIN_BATCH_TIMER).start()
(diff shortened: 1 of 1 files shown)


tjruwase:
@nelyahu, thanks for the PR. You are correct that timer results are printed only on rank0, while the related resources and work are wasted on other ranks. However, this approach allows the data parallel ranks to execute similar code as much as possible which helps to keep them synchronization, load-balanced, and easier to debug. In our experience, the waste is typically negligible, so we did not try to optimize. Are you experiencing something different? Perhaps, we can make this behavior optional. Thanks!

nelyahu:
Hi @tjruwase, thanks for your response.

This resource is limited in terms of HPU backend, earlier we had relatively low number of such event and had to increase to avoid such lack of resources, but we observed that when running for a very long training we can meet a scenario where we don't have any more event handles to create, and we crash.

I see now that the updated master code calls "elapsed" API to release the resources even for non-0 ranks. which is good. 
see below PR:
https://github.com/microsoft/DeepSpeed/pull/4308

I think that it is OK for now and we can close this PR.

PR #4327, opened 2023-09-13
pipe engine _aggregate_total_loss: more efficient loss concatenation

nelyahu:
optimize _aggregate_total_loss function in order to remove dependancy of copying from device to host and back to device. This reduce the runtime on the host.


tjruwase (approved):

nelyahu:
@tjruwase can we re-trigger test workflows?

PR #4328, opened 2023-09-13
Add release version checking

loadams:
Add script that checks that the next DeepSpeed version is valid.


release/check_release_version.py
```diff
@@ -0,0 +1,29 @@
...
+
+new_version = pkg_version.parse(args.new_version)
+
+with open('./version.txt') as fd:
+    current_version = pkg_version.parse(fd.read())
+
+# Valid version are those where the major/minor/micro are incremented by no more than one from the existing release, and the less significant values are reset to 0.
+valid_major_update = pkg_version.Version(f'{current_version.major + 1}.0.0')
+valid_minor_update = pkg_version.Version(f'{current_version.major}.{current_version.minor + 1}.0')
+valid_micro_update = pkg_version.Version(
```
mrwyattii, line 23: I typically see the last number of the version referred to as `patch` rather than `micro` - Is `micro` also used?

mrwyattii (approved):

release/check_release_version.py
loadams: Me too, but from the [docs](https://packaging.pypa.io/en/stable/version.html), micro is how they refer to the third entry of the version.  I think patch is from the semantic versioning line of thought?

The same "micro" is used in the bump_patch_version script [here](https://github.com/microsoft/DeepSpeed/blob/master/release/bump_patch_version.py#L20).

PR #4336, opened 2023-09-14
Fix skipped inference tests

mrwyattii:
`HfApi().list_models()` returns a generator. We were using it twice in our setup of the creation of a model-task tuples for the inference tests. This means that the first time we used it, we got the full list. The second time we used it, we got an empty list. Looks like we have been skipping many tests for a few months now!

@loadams


loadams (approved):
Out of curiosity, how much time does this add back to the unit tests?

lekurile (approved):

PR #4337, opened 2023-09-14
adds triton flash attention2 kernel

stephen-youn:
This PR adds flash attention 2 kernel implemented in triton2.1 for inference.
Benchmarking on bert-base shows 4~13% further latency reduction (the longer the sequence length, the larger the performance gain) when compared to the case using regular attention triton kernel.

image


tests/unit/inference/test_inference.py
```diff
@@ -261,6 +261,11 @@ def validate_test(model_w_task, dtype, enable_cuda_graph, enable_triton):
 
 
 @pytest.mark.inference
+@pytest.mark.parametrize("model_w_task", [
```
lekurile, line 264: Can we make sure to remove this parametrize since it'll override the `model_w_task` fixture, which is the more comprehensive enumeration of tests?

lekurile (changes requested):
Left a comment about the `TestModelTask` unit test, other than that LGTM!

tests/unit/inference/test_inference.py
stephen-youn: yes I removed it as before

lekurile (approved):
LGTM!

PR #4340, opened 2023-09-14
Fix autotune to support Triton 2.1 

stephen-youn:
To fully support triton2.1 from deepspeed, this PR changes autotune and config pruning codes.


lekurile (approved):
LGTM!

PR #4360, opened 2023-09-19
Cache metadata for TP activations and grads

BacharL:
PartitionedTensor.from_meta will cause device to host synchronization when reading the meta tensor in 
meta = meta.tolist()
Added cpu cache for the meta tensor to avoid this synchronization in every activation and grad communication between the ranks. The meta tensor is assumed to be static since activation shape must be static. The user must call reset_activation_shape if any of the dimentions change.


hablb:
@tjruwase Could you please review?

tjruwase (approved):

BacharL:

________________________________

mrwyattii (approved):

PR #4383, opened 2023-09-21
Destroy ZeRO

jomayeri:
Create destroy function for zero stage 1 and 2. Add a destructor to the engine so that it releases memory at exit.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -362,6 +362,9 @@ class DeepSpeedEngine(Module):
         if self.optimizer is not None and hasattr(self.optimizer, 'destroy'):
             self.optimizer.destroy()
 
+    def __del__(self):
+        self.destroy()
+
     def _get_model_parameters(self):
         if self.autotuning_profile_model_info():
             self.autotuning_model_info = {}
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -519,6 +519,9 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
         self._enable_universal_checkpoint()
         self._param_slice_mappings = self._create_param_mapping()
 
+    def destroy(self):
+        self._uncreate_param_mapping()
+
     def _enable_universal_checkpoint(self):
         for lp_param_group in self.bit16_groups:
             enable_universal_checkpoint(param_list=lp_param_group)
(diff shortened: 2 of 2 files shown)


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -535,6 +538,12 @@ def _create_param_mapping(self):
 
         return param_mapping
 
+    def _uncreate_param_mapping(self):
```
tjruwase, line 541: This should be a utility function in [mixed_precision_linkage.py](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/utils/mixed_precision_linkage.py) as inverse of [link_hp_params](https://github.com/microsoft/DeepSpeed/blob/0e0748c579cff3f94dc0609677623d082c9a2132/deepspeed/utils/mixed_precision_linkage.py#L11).

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -519,6 +519,9 @@ def __init__(self,
         self._enable_universal_checkpoint()
         self._param_slice_mappings = self._create_param_mapping()
 
+    def destroy(self):
+        self._uncreate_param_mapping()
```
tjruwase, line 523: Don't we need to delete the buffers related to contiguous_gradients and gradient_partitions?

deepspeed/runtime/engine.py
```diff
@@ -362,6 +362,9 @@ def destroy(self):
         if self.optimizer is not None and hasattr(self.optimizer, 'destroy'):
             self.optimizer.destroy()
 
+    def __del__(self):
```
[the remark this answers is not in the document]
jomayeri: Will do.

deepspeed/runtime/zero/stage_1_and_2.py
jomayeri: Deleting those buffers didn't make a difference when running experiments but I can add a full delete of all attributes for completeness just as I did for ZeRO 3.

tjruwase: Sounds good. Is it possible to make the attribute deletion function a utility that lives in deepspeed.utils?

deepspeed/utils/mixed_precision_linkage.py
```diff
@@ -17,6 +16,10 @@ def link_hp_params(lp_param_list, flat_hp_partition, gradient_dict, offload_grad
                                                        offload_gradient_dict, use_offload, param_group_index,
                                                        partition_start, partition_size, partition_optimizer_state)
 
+def unlink_hp_mapping(lp_param_list):
```
tjruwase, line 19: For symmetry, please rename to `unlink_hp_params`

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -519,6 +520,10 @@ def __init__(self,
         self._enable_universal_checkpoint()
         self._param_slice_mappings = self._create_param_mapping()
 
+    def destroy(self):
```
tjruwase, line 523: This is needed in the other optimizer implementations: 
1. https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/bf16_optimizer.py
2. https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/fp16/fused_optimizer.py
3. https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/fp16/unfused_optimizer.py

PR #4384, opened 2023-09-21
Allow multiple inference engines in single script

mrwyattii:
The `InferenceEngine` relies on some global values and fails to clean up workspaces such that if we try to run multiple models via DeepSpeed-Inference in a single script (see example below), we get errors like this:
```bash
RuntimeError: The specified pointer resides on host memory and is not registered with any CUDA device.
```

This PR adds a `destroy()` method that will be automatically called on subsequent invocations of `deepspeed.init_inference()`.

example.py:
```python
import os
import torch
import deepspeed
from transformers import pipeline

local_rank = int(os.environ.get("LOCAL_RANK", 0))
world_size = int(os.environ.get("WORLD_SIZE", 1))

model1 = "bigscience/bloom-560m"
task1 = "text-generation"
pipe1 = pipeline(task1, model1, torch_dtype=torch.float16, device=local_rank)
pipe1.model = deepspeed.init_inference(
    pipe1.model,
    dtype=torch.float16,
    replace_with_kernel_inject=True,
    mp_size=world_size,
)
print(model1, pipe1("test one two"))

model2 = "gpt2"
task2 = "text-generation"
pipe2 = pipeline(task2, model2, torch_dtype=torch.float16, device=local_rank)
pipe2.model = deepspeed.init_inference(
    pipe2.model,
    dtype=torch.float16,
    replace_with_kernel_inject=True,
    mp_size=world_size,
)
print(model2, pipe2("test one two"))
```

Co-authored-by: Jeff Rasley <jeffra45@gmail.com>

### DIFF
--- a/deepspeed/inference/engine.py
+++ b/deepspeed/inference/engine.py
@@ -27,7 +27,6 @@ from ..module_inject.auto_tp import AutoTP
 
 from ..module_inject.replace_policy import generic_policies
 from ..module_inject.auto_tp_model_utils import build_bloom_alibi_tensor, build_mpt_atten_bias_tensor, build_mpt_alibi_tensor
-from ..model_implementations.transformers.ds_transformer import DeepSpeedTransformerInference, inference_module
 from ..ops.transformer.inference.ds_attention import DeepSpeedSelfAttention
 
 DS_INFERENCE_ENABLED = False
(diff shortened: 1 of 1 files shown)


RezaYazdaniAminabadi (approved):

deepspeed/inference/engine.py
```diff
@@ -174,6 +179,11 @@ def __init__(self, model, config):
         # Check if local CUDA graphs can be created in replacement modules
         self.local_cuda_graph = self._local_cuda_graph_used(self.module)
 
+    def destroy(self):
+        DeepSpeedTransformerInference.layer_id = 0
+        DeepSpeedSelfAttention.num_layers = 0
+        inference_module.release_workspace()
```
jeffra, line 185: Just in case a user wants to explicitly called destroy, might be nice to prevent an `AttributeError` in certain cases

```suggestion
        if inference_module is not None:
            inference_module.release_workspace()
```

PR #4390, opened 2023-09-23
Add condition when dimension is greater than 2

ajindal1:
When dimension is greater than 2, grad_bias should be calculated differently.


ajindal1:
@pengwa suggested this change.

tjruwase (approved):

ajindal1:
> @ajindal1, thanks for the PR. Can you please add a unit test for this case?
@tjruwase Thanks for reviewing and merging the PR, we will work on adding a unit test.

PR #4391, opened 2023-09-23
Add torch no grad condition

ajindal1:
Add torch no grad condition for param ds_tensor otherwise torch tracer will complain


ajindal1:
@pengwa suggested this change.

tjruwase (approved):

PR #4413, opened 2023-09-27
Fixup check release version script

loadams:
- Pull changes from MII

### DIFF
--- a/release/check_release_version.py
+++ b/release/check_release_version.py
@@ -8,9 +8,7 @@ from packaging import version as pkg_version
 
 parser = argparse.ArgumentParser()
 
-parser.add_argument("--release_version",
-                    type=str,
-                    help="The new version being published.")
+parser.add_argument("--release_version", type=str, help="The new version being published.")
 
 args = parser.parse_args()
 
--- a/release/check_release_version.py
+++ b/release/check_release_version.py
@@ -8,22 +8,15 @@ from packaging import version as pkg_version
 
 parser = argparse.ArgumentParser()
 
-parser.add_argument("--new_version", type=str, help="The new version being published.")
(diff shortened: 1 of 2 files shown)


release/check_release_version.py
```diff
@@ -8,22 +8,15 @@
...
 
-new_version = pkg_version.parse(args.new_version)
+release_version = pkg_version.parse(args.release_version)
 
 with open('./version.txt') as fd:
-    current_version = pkg_version.parse(fd.read())
+    repo_version = pkg_version.parse(fd.read())
 
-# Valid version are those where the major/minor/micro are incremented by no more than one from the existing release, and the less significant values are reset to 0.
-valid_major_update = pkg_version.Version(f'{current_version.major + 1}.0.0')
```
tjruwase, line 21: Why are we adopting more stricter version checking?

tjruwase (approved):

release/check_release_version.py
loadams: This was a bug we identified when testing the [new release flow](https://github.com/microsoft/DeepSpeed-MII/actions/runs/6306178296/job/17120874131) in MII.  The script was previously too strict in the versions it accepted from versions.txt to the git tag for the next release.  I discussed with @mrwyattii and @jeffra and this was what we settled on.  Basically version.txt is always set to the "next" patch version, so the valid versions for a new release are either 1) the version in version.txt if we are just updating the patch, 2) an increment in minor version by 1, zeroing the patch version, or 3) an increment in the major version by 1, zeroing the minor and patch versions.

tjruwase: Thanks for the explanation, @loadams. LGTM.

PR #4416, opened 2023-09-28
Model checkpointing with ZeRO-3 training with NVMe offload

eisene:
This is a proposal for an implementation of checkpointing models when training with ZeRO-3 with NVMe offload. It consists of several logical parts:

1. Currently, the names of the files used in the checkpoint are based on the Python id of the parameter object, which is just the parameter's address in memory. This is not stable across runs, which has two disadvantages:
   - The NVMe offloading files grow with every run of the model even if the architecture did not change. This wastes disk space and, at least for me, was a surprise when I first saw it. It is not related to checkpointing.
   - Without a way to match the file to the offloaded tensor we can't reload the checkpoint.

   We propose an alternative naming scheme. The parameters are named after their ds_id instead of their Python id, and the tensors are named after their state_name and (new) parameter id. See commit cbf83d9.

2. A model checkpoint now has to include all the offloaded tensor files. During checkpoint save/load we copy all the tensor files to/from the "offloaded_tensors" subdirectory of the checkpoint. We provide some logging on the remaining space on the file system due to the potential size of these files, especially as they accumulate in each checkpoint. See commits 0accdb8, ba7d3e6, a99daa0.

3. When loading the checkpoint, the optimizer already has prepared buffers for swapping. We need to purge them so that they are replaced with the freshly copied on-disk buffers from the checkpoint. See commit 6c6a1ec.

I'm working on a test for this feature but I wanted to finally get this code in front of the DeepSpeed team for your feedback.


tjruwase:
@eisene, this is really amazing! Thanks so much. I will review asap.

deepspeed/runtime/zero/stage3.py
```diff
@@ -2400,23 +2402,30 @@ def _rigid_load_state_dict(self, state_dict, load_optimizer_states=True):
...
-
-            for partitioned_param, q in zip(self.fp16_partitioned_groups[sub_group_id], updated_params):
-                partitioned_param.data = q.data
+        if self.swap_optimizer or self.params_in_nvme_and_cpu:
+            # Purge the swapped optimizer state, it was initialized to the freshly created model and not the checkpoint
+            for swap_info in self.optimizer_swapper.swap_params_info.values():
+                swap_info.tensors = [swap_info.tensors[0]]
+                swap_info.has_state_tensors = False
+
+        if not (self.swap_optimizer or self.params_in_nvme_and_cpu):
```
tjruwase, line 2411: Could this simply be `else` for line 2405?

deepspeed/runtime/engine.py
```diff
@@ -2701,6 +2704,17 @@ def load_checkpoint(self,
             if not success:
                 self.optimizer._restore_from_bit16_weights()
 
+        if self.zero_has_nvme_offload():
+            from shutil import copytree, disk_usage
+            offload_dir = self.optimizer.optimizer_swapper.swap_folder
+            offload_ckpt_dir = os.path.join(load_dir, tag, "offloaded_tensors")
+            _, _, free = disk_usage(offload_dir)
+            logger.info(f"Copying NVMe offload checkpoint from {offload_ckpt_dir} to {offload_dir}, {free / 1e9:,.2f} GB free on target filesystem...")
+            copytree(offload_ckpt_dir, offload_dir, dirs_exist_ok=True)
```
tjruwase, line 2713: Do you avoid copying the offloaded gradient tensor files?

deepspeed/runtime/zero/stage3.py
eisene: Yes :) I'll change it in the next commit.

eisene:
Closing in favor of #4707

PR #4427, opened 2023-09-29
Skip TestModelTask

loadams:


loadams:
Just skipping on this causes the cpu_inference tests to run for ~2 hours and still only run the same 4 tests that were running before.  So I'd recommend not completing this PR.

PR #4431, opened 2023-10-01
Small docstring fix

Jackmin801:


Jackmin801:

deepspeed/runtime/fp16/fused_optimizer.py
```diff
@@ -137,7 +137,6 @@ def zero_grad(self, set_to_none=False):
         """
         Zero FP16 parameter grads.
         """
-        # For speed, set model fp16 grad to None by default
```
loadams, line 140: Why remove this throughout and not update to match the code more closely if you don't think it does?

Jackmin801: The change that made this comment false intended to match PyTorch
https://github.com/microsoft/DeepSpeed/pull/2741

However, it does seem like PyTorch has actually changed to default `set_to_none=True` now.
https://pytorch.org/docs/stable/_modules/torch/optim/optimizer.html#Optimizer.zero_grad
https://github.com/pytorch/pytorch/pull/92731

I will raise a separate PR as this is not longer about small docstring changes and might be breaking.
Thanks for the review!

loadams (approved):

mrwyattii (approved):

PR #4436, opened 2023-10-02
optimize grad_norm calculation in stage3.py

mmhab:
reduce the synchronization between the device and the host by removing .item() from the loops that calculate the total norm.


deepspeed/runtime/zero/stage3.py
```diff
@@ -1288,10 +1288,14 @@ def complete_grad_norm_calculation_for_cpu_offload(self, params):
...
+        total_norm = total_norm_cuda[0]**(1. / norm_type)
 
-        if total_norm == float('inf') or total_norm == -float('inf') or total_norm != total_norm:
-            total_norm = -1
+        norm_is_inf = total_norm.isinf()
+        norm_is_nan = total_norm.isnan()
+        inf_or_nan = norm_is_nan.logical_or(norm_is_inf)
+
+        err = torch.tensor(-1.0, device=self.current_device, dtype=torch.float)
+        total_norm = err.where(inf_or_nan,total_norm)
```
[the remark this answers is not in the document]
mmhab: As I see it, when you use `if condition` you need to calculate the tensor on the host and synchronize between the host and the device, compared of using the `where` that in this case the two tensors `err `and `total_norm` allocated on the device so there is no need to synchronize with the host.

tjruwase (approved):

mmhab:
@tjruwase Hi, can you please help me to merge these changes?

mrwyattii:
@mmhab could you please run the formatter and commit any changes? `pre-commit run --all-files`

deepspeed/runtime/zero/stage3.py
```diff
@@ -1686,10 +1690,14 @@ def get_grad_norm_direct(self, gradients, params, norm_type=2):
...
-            total_norm = total_norm_cuda.item()**(1. / norm_type)
+            total_norm = total_norm_cuda**(1. / norm_type)
+
+        norm_is_inf = total_norm.isinf()
+        norm_is_nan = total_norm.isnan()
+        inf_or_nan = norm_is_nan.logical_or(norm_is_inf)
 
-        if total_norm == float('inf') or total_norm == -float('inf') or total_norm != total_norm:
-            total_norm = -1
+        err = torch.tensor(-1.0, device=self.current_device, dtype=torch.float)
```
mrwyattii, line 1699: I don't think there is a `self.current_device` defined

deepspeed/runtime/zero/stage3.py
```diff
@@ -1333,10 +1333,14 @@ def complete_grad_norm_calculation_for_cpu_offload(self, params):
...
-        total_norm = total_norm_cuda[0].item()**(1. / norm_type)
+        total_norm = total_norm_cuda[0]**(1. / norm_type)
 
-        if total_norm == float('inf') or total_norm == -float('inf') or total_norm != total_norm:
-            total_norm = -1
+        norm_is_inf = total_norm.isinf()
+        norm_is_nan = total_norm.isnan()
+        inf_or_nan = norm_is_nan.logical_or(norm_is_inf)
+
+        err = torch.tensor(-1.0, device=self.current_device, dtype=torch.float)
```
tjruwase, line 1342:
```suggestion
        err = torch.tensor(-1.0, device=self.device, dtype=torch.float)
```

deepspeed/runtime/zero/stage3.py
tjruwase:
```suggestion
        err = torch.tensor(-1.0, device=self.device, dtype=torch.float)
```

Great catch, Mike!

mmhab:
I made some changes in the commit. Please review @jeffra @samyam

PR #4438, opened 2023-10-02
Change default `set_to_none=true` in `zero_grad` methods

Jackmin801:
The default was changed to `false` in this PR with the intention of matching PyTorch
https://github.com/microsoft/DeepSpeed/pull/2741

However, since then, PyTorch has changed their default to `true`.
https://pytorch.org/docs/stable/_modules/torch/optim/optimizer.html#Optimizer.zero_grad
https://github.com/pytorch/pytorch/pull/92731

### DIFF
--- a/deepspeed/runtime/fp16/fused_optimizer.py
+++ b/deepspeed/runtime/fp16/fused_optimizer.py
@@ -133,7 +133,7 @@ class FP16_Optimizer(DeepSpeedOptimizer):
 
         return
 
-    def zero_grad(self, set_to_none=False):
+    def zero_grad(self, set_to_none=True):
         """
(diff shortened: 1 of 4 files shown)


Jackmin801:
Should I go around the code changing all the `self.zero_grad(set_to_none=True)` to `self.zero_grad()`?
Personally I think it's ok to just leave them be.

tjruwase (approved):

tjruwase:
> Should I go around the code changing all the `self.zero_grad(set_to_none=True)` to `self.zero_grad()`? Personally I think it's ok to just leave them be.

@Jackmin801, thanks for the PR. Yes, it is okay to leave the existing calls as is.

PR #4430, opened 2023-10-01
[Fix] Fix cpu inference UT failure

delock:
This PR fix UT test error as described in this PR and the following test job.  This PR skips `TestModelTask` if dtype is not supported by accelerator, or `InferenceBuilder` is not implemented by accelerator.
https://github.com/microsoft/DeepSpeed/pull/4419
https://github.com/microsoft/DeepSpeed/actions/runs/6341645987/job/17235544538

PR #4263, opened 2023-09-05
[CPU] Add CPU AutoTP UT.

Yejing-Lai:
This PR aims to add CPU AutoTP UT. 

We found that "Salesforce/codegen-350M-mono" model(bf16) cannot pass TestAutoTensorParallelism testcases under torch 2.0. So we skipped this test temporarily.

PR #4439, opened 2023-10-02
Test fixes + moving dtypes out of the test

loadams:


loadams:
@delock/ @Yejing-Lai - this PR just applies your fixes but also tries to reduce the overall test time.  Before, the 4 tests ran and took about 2 minutes to run.  After #4263 + #4430 it runs the same tests but because of test setup takes much longer, now [about an hour but only still runs those 4 tests](https://github.com/microsoft/DeepSpeed/actions/runs/6385697921/job/17330974721?pr=4439).  We were looking at that, but had a general question on if more tests were supposed to be added or enabled with that PR, or if we could revert to the way we had been skipping the tests before?

cc: @mrwyattii

PR #4450, opened 2023-10-03
adding 8bit dequantization kernel for asym fine-grained block quantization in zero-inference

stephen-youn:
This PR adds 8bit dequantization kernels that takes similar approach with the 4bit kernels that are already in the repo. (this was added while assessing the gemm + dequantization performance in the repo, 8bit kernels were missing but it's not difficult to add one here with this PR).


csrc/quantization/quantize_int4.cu
```diff
@@ -228,3 +228,54 @@ void launch_dequantize_int4_to_half_experimental(uint8_t* data_in,
...
+{
+    AlignedArray<half, N> ret;
+
+#pragma unroll
+    for (int idx = 0; idx < N; idx += 1) { ret[idx] = half(int(data[idx])); }
+
+    return ret;
+}
+
+__global__ void dequantize_int8_to_half(uint8_t* data_in,
```
tjruwase, line 243: It is a little confusing to have this function in a file named `*int4.cu`

stephen-youn: yeah, renamed the file name as suggested.

tjruwase:
@stephen-youn, can you please look into the failing CIs?

deepspeed/inference/quantization/utils.py
```diff
@@ -105,20 +105,24 @@ def __init__(self, config: Dict, dtype: torch.dtype) -> None:
...
-                tensor.reshape(-1, last_dimension_size), quant_scale, quant_min,
-                tensor.numel() // last_dimension_size, self.config['group_size'])
-
-            shape = list(tensor.shape)
-            if self.config['num_bits'] == 4:
+                quantized_tensor = get_quantizer_cuda_module().dequantize_int4_to_half_experimental(
+                    tensor.reshape(-1, last_dimension_size), quant_scale, quant_min,
+                    tensor.numel() // last_dimension_size, self.config['group_size'])
+                shape = list(tensor.shape)
                 shape[-1] = shape[-1] * 2
```
tjruwase, line 119: What is this line for?

tests/unit/inference/quantization/test_intX_quantization.py
```diff
@@ -326,9 +329,9 @@ def test_quantized_linear(self):
                 'weight_quantization': {
                     'post_init_quant': {
                         'layer': {
-                            'num_bits': 4,
+                            'num_bits': quantization_bits,
                             'group_size': 64,
-                            'group_dim': 0,
+                            'group_dim': 1,
```
tjruwase, line 334: Why this change?

deepspeed/inference/quantization/utils.py
stephen-youn: it's the same dequantization kernel call. the codes are the same but it's only indented and that's why it's showing as diff

tests/unit/inference/quantization/test_intX_quantization.py
stephen-youn: nice catch. fixed it.

stephen-youn:
> @stephen-youn, can you please look into the failing CIs?

tests/unit/inference/quantization/test_intX_quantization.py
stephen-youn: we need to cover group_dim along the last dimension in the tensor to invoke the cuda dequantization kernels. otherwise, it will call reference torch version([link](https://github.com/microsoft/DeepSpeed/blob/0b21637881b562ab0ca0ab9caadce0555d221912/deepspeed/inference/quantization/utils.py#L109C1-L109C72)). so i changed the code to cover both cases with group_dim 0 and 1 from this test.

tjruwase (approved):

PR #4452, opened 2023-10-04
CI fix for torch 2.1 release

mrwyattii:
Includes fix for function changes with `_coalescing_manager()` in https://github.com/pytorch/pytorch/pull/98793

@loadams


loadams (approved):

.github/workflows/nv-mii.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
       - name: Install pytorch
         run: |
-          pip3 install -U --cache-dir $TORCH_CACHE torch
+          pip3 install -U --cache-dir $TORCH_CACHE torch --index-url https://download.pytorch.org/whl/cu118
```
loadams, line 29: I don't know if we have cuda 118 runners, but should we update the runs-on here to be 118 if possible?

mrwyattii: Any of the CI workflows that we want to use "torch-latest" with will need to be updated to 11.8 - torch 2.1 is only precompiled for CUDA 11.8 and 12.1

PR #4487, opened 2023-10-09
Add AMD team to CI issue creation.

loadams:


.github/workflows/amd-mi200.yml
```diff
@@ -78,4 +78,5 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           filename: .github/ISSUE_TEMPLATE/ci_failure_report.md
+          assignees: jithunnair-amd, rraminen
```
jithunnair-amd, line 81: @loadams https://docs.github.com/en/issues/tracking-your-work-with-issues/assigning-issues-and-pull-requests-to-other-github-users#about-issue-and-pull-request-assignees

jithunnair-amd: The least permissive option seems to be to make us "organization members with read permissions to the repository"

loadams: @jithunnair-amd - I confirmed that even with the correct permissions the way that we do the issue creation I cannot auto add you.  Will manual adding work?  Since we triage them first we can tag you in them or if they are our infra side we can handle them?

jithunnair-amd: @loadams That sounds better!

PR #4388, opened 2023-09-22
[CI] use latest transformers

jeffra:
Unpin transformers version for transformers workflow

PR #4501, opened 2023-10-11
Fix transformers test skipping all tests.

loadams:
Around the release of torch 2.1, something changed that began skipping all of the tests that should have been running in nv-transformers.

Sample passing build before: https://github.com/microsoft/DeepSpeed/actions/runs/6399875223/job/17372640478#step:8:1486
Sample passing build after (skips all tests): https://github.com/microsoft/DeepSpeed/actions/runs/6476276918/job/17584733821#step:8:1355

### DIFF
--- a/.github/workflows/nv-transformers-v100.yml
+++ b/.github/workflows/nv-transformers-v100.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install pytorch
         run: |
           # use the same pytorch version as transformers CI
-          pip install -U --cache-dir $TORCH_CACHE torch==2.1.0+cu118 --index-url https://download.pytorch.org/whl/cu118
+          pip install torch==2.0.1+cu118 --index-url https://download.pytorch.org/whl/cu118
           python -c "import torch; print('torch:', torch.__version__, torch)"
           python -c "import torch; print('CUDA available:', torch.cuda.is_available())"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/nv-transformers-v100.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Install pytorch
         run: |
           # use the same pytorch version as transformers CI
```
loadams, line 29: @stas00 - could you point me to the place in transformers CI where this is set?  It looks like these are failing due to using torch 2.1.0 in place of the 2.0.1 we specify, but I wanted to make sure we were still on the same torch version.

loadams:
Closing in favor of #4388

.github/workflows/nv-transformers-v100.yml
stas00: It looks like you have sorted it out already, Logan.

Also for future similar requests please tag @pacman100 who has taken over the maintenance of the Deepspeed integration and I'm not longer with HF. Thank you!

loadams: I did - and will update who I tag in the future.  My apologies for the incorrect tag!

stas00: no problem at all, Logan - you couldn't have known ;) 

But please feel free to tag me on other non-HF-specific issues - I still care very much for Deepspeed and hope to start using it again soon (current internode speed is too slow to permit ZeRO).

PR #4508, opened 2023-10-12
add available memory check to accelerators

jeffra:
There are many scenarios where we need to get an accurate estimate of the available memory on a device. Currently we rely on the torch memory allocator stats to give us this information, however there are several cases where memory may be allocated outside the view of torch. This means that `torch.cuda.get_device_properties(device_index).total_memory - torch.cuda.memory_allocated(device_index)` is not accurate. This is usually less of a problem on data center GPUs but quite common on consumer grade GPUs that are often shared between torch and the operating system.

This PR introduces `available_memory` to the abstract accelerator interface. On CUDA devices we can rely on [`pynvml`](https://pypi.org/project/pynvml/) to get the ground truth w.r.t. available memory.

This also introduces a hard dependency on `pynvml`. I have tested on non-GPU systems and this package seems to install successfully but fails at runtime at the `nvmlInit()` call. We fall back to using torch stats for memory in cases where pynvml is not functional.

### DIFF
--- a/accelerator/cuda_accelerator.py
+++ b/accelerator/cuda_accelerator.py
@@ -22,18 +22,21 @@ pynvml = None
 class CUDA_Accelerator(DeepSpeedAccelerator):
 
     def __init__(self):
-        global pynvml
         self._name = 'cuda'
         self._communication_backend_name = 'nccl'
         if pynvml is None:
-            try:
-                import pynvml
-            except ImportError:
-                pass
-            try:
-                pynvml.nvmlInit()
-            except pynvml.NVMLError:
-                pynvml = None
+            self._init_pynvml()
+
+    def _init_pynvml(self):
+        global pynvml
+        try:
+            import pynvml
+        except ImportError:
+            return
+        try:
+            pynvml.nvmlInit()
+        except pynvml.NVMLError:
+            return
 
     def is_synchronized_device(self):
         return False
--- a/accelerator/cuda_accelerator.py
+++ b/accelerator/cuda_accelerator.py
@@ -15,12 +15,22 @@ try:
 except ImportError:
     pass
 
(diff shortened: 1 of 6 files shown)


accelerator/cuda_accelerator.py
```diff
@@ -15,12 +15,28 @@
...
+    def _init_pynvml(self):
+        global pynvml
+        try:
+            import pynvml
+        except ImportError:
+            return
+        try:
+            pynvml.nvmlInit()
+        except pynvml.NVMLError:
+            return
```
mrwyattii, line 39:
```suggestion
        global pynvml
        try:
            import pynvml as tmp_pynvml
        except ImportError:
            return
        try:
            tmp_pynvml.nvmlInit()
            pynvml = tmp_pynvml
        except tmp_pynvml.NVMLError:
            return
```

We would need to do something like this to avoid a bug when we are able to import `pynvml` by `pynvml.nvmlInit()` fails. For example, consider this:
```python
torch = None

def init_torch():
    global torch
    try:
        import torch
    except:
        return
    try:
        torch.fake_function()
    except:
        print("TORCH FAILED")
        return
        
init_torch()
if torch:
    print("TORCH IMPORTED")
```

Running this example outputs the following:
```bash
❯ python test.py  
TORCH FAILED
TORCH IMPORTED
```

This means that below in `available_memory`, we may fail to use the fallback method in the case where pynvml fails to initialize due to `if pynvml` on line 156.

tjruwase:
@delock, FYI

tjruwase (approved):

mrwyattii (approved):

delock:
> @delock, FYI

Thanks for reminding.  I think the CPU part is good.  We will add to XPU backend as well.

PR #4512, opened 2023-10-15
Don't check overflow for bf16 data type

hablb:
Always check for fp16
bf16 dynamic range is similar to fp32. don't check overflow by default.


tjruwase:
@hablb, thanks for the PR. While I understand restricting overflow checks to fp16 only, I don't see the benefit of providing a config option to enable overflow checks for bf16/fp32. Can you please explain the need for the config option? I think the restriction part is fine.

hablb:
Thanks @tjruwase 
I removed the configuration variable. Just skipping overflow check for fp16 now.

tjruwase (approved):

jinzhuer:
Dear @hablb ,Besides wasting computational resources, are there any other downsides to performing overflow checks on bf16 data types? Would performing overflow checks on bf16 data types cause any other bugs?

hablb:
No other downsides, however this check significant overhead as it reduces all gradients, communicates and synchronizes with the cpu.

PR #4516, opened 2023-10-16
Enable universal checkpoint for zero stage 1

### LINKED ISSUE #2921: [REQUEST] universal checkpoint for ZeRO - 1,2,3

I think we now have all the components ready to do universal checkpoint in ZeRO - 1,2,3, like we had for BF16Optimizer.

The need is to be able to add/remove gpus when resuming from a checkpoint with a different number of gpus once the training started.

Thank you.

Progress update:

- [x] Stage 1 https://github.com/microsoft/DeepSpeed/pull/4516
- [x] Stage 2 https://github.com/microsoft/Megatron-DeepSpeed/tree/main/examples_deepspeed/universal_checkpointing#zero-stage-2-training
- [x] Stage 3 #5475 

@tjruwase

tjruwase:
Generalize universal checkpointing in DS: 
1. Enable for zero stage 1
2. Move conversion script into DS

Tested with Megatron-DS GPT using companion PR https://github.com/microsoft/Megatron-DeepSpeed/pull/265

Fix #2921 

- [x] readthedocs
- ~~[ ] Tutorial~~: Defer


tjruwase:
@stas00, FYI

stas00:
Amazing! Thank you for starting to work on this super-essential feature, Tunji!

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -0,0 +1,298 @@
...
+    os.makedirs(param_base_path, exist_ok=True)
+
+    cnt += 1
+    counter = f"{dp_index:0>2d}"
+
+    path = os.path.join(param_base_path, f"{state_name}.{counter}")
+
+    #print(f"{param_name}: {offset}: {numel} => {path}")
+
+    t = state_flat_tensor.narrow(0, offset, numel)
```
mosheisland, line 138: do state_flat_tensor.narrow(0, offset, numel).clone()
otherwise, you keep the original space size on disk

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -0,0 +1,298 @@
...
+    vocabulary_parameters = universal_checkpoint_info.get(VOCABULARY_PARAMETER_PATTERNS, [])
+    for state in ("fp32", "exp_avg", "exp_avg_sq"):
+        slices = _merge_zero_shards(slice_base_path, state, tp_degree, shape)
+        final_path = os.path.join(param_base_path, f"{state}.pt")
+
+        #print(f"Expected shape: {shape}")
+        #print(f"Fragment sizes:", list(frag.shape for frag in slices))
+        ckpt_dict = {}
+        if any(re.match(pattern, name) for pattern in parameters_to_average):
+            param = sum(slices) / len(slices)
```
mosheisland, line 179: parameters_to_average are expected to be the same.
therefore, instead of doing average, suggest to:
1. assert all are equal
2. return first one

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -0,0 +1,298 @@
...
+                        type=int,
+                        help='How many parallel processes to extract zero shards')
+    parser.add_argument(
+        '--num_merge_workers',
+        default=2,
+        type=int,
+        help=
+        'How many parallel processes to merge tp slices (more memory intensive, use much fewer than --num_extract_workers))'
+    )
+    parser.add_argument('--for_release',
```
mosheisland, line 54: is this used?

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -0,0 +1,298 @@
...
+
+    t = state_flat_tensor.narrow(0, offset, numel)
+    _save_checkpoint(path, t)
+
+
+def _merge_zero_shards(param_base_path, state, tp_degree, slice_shape):
+    slices = []
+    for tp_index in range(tp_degree):
+        prefix_path = os.path.join(param_base_path, str(tp_index), f"{state}")
+        paths = sorted(list(glob.glob(f"{prefix_path}.0*")))
```
mosheisland, line 146: Remove leading 0. 
Change to : sorted(list(glob.glob(f"{prefix_path}.*")))
Otherwise, bug when dp >= 10

deepspeed/checkpoint/ds_to_universal.py
tjruwase: Not yet. Will remove.

deepspeed/checkpoint/ds_to_universal.py
tjruwase: Good point. I may be wrong, but I seem to recall that there was a case of averaging across the tensor slices in BLOOM, separately from the case of replicated slices. Regardless, it makes sense to separate these two cases.

@stas00, any thoughts?

stas00: It has been a very long time, and I unfortunately don't remember this exact use case, but I'm pretty sure we had to average some of the tensors. Otherwise we would have copied the first one as it simpler and clearly isn't the same as averaging.

Probably the safest approach is to give the choice to the user: so that they could define which params are to be `parameters_copy_rank0` and which are `parameters_to_average`. And assert for the former if it's not equal on all ranks to flag to the user that they might be doing the wrong thing.

stas00: oh, but wait, this is not Megatron-Deepspeed - this is ZeRO - so the logic is quite different. I thought you were asking about Meg-DS as a similar PR went there as well.

tjruwase: @stas00, this is indeed for Meg-DS. I have moved `ds_to_universal.py` into DS codebase for future support, as we discussed.   So, this is that general problem of how to reshape partitioned tensor slices.

stas00: Aha, that explains it, thank you.

So this is the problem of whether to average say TP tensor of rank 0 across all replicas, yes? i.e. to average or not is across replicas.

They should be the same, except when they aren't. And I think we saw in the case of BLOOM that they weren't the same, which I think was a bug. But again, this was long time ago to remember the details.

That's why I suggest to give the reins to the user by defining both approaches and choosing a sensible default.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2183,16 +2182,39 @@ def param_groups(self):
...
             for lp in self.bit16_groups[i]:
                 if lp._hp_mapping is not None:
                     #print(f"Loading {self.param_names[lp]} {tp_rank=} {tp_world_size=}")
                     lp.load_hp_checkpoint_state(os.path.join(checkpoint_dir, self.param_names[lp]), tp_rank,
                                                 tp_world_size)
 
+    def _load_global_state(self, sd):
+        self.loss_scaler = sd.get('loss_scaler', self.loss_scaler)
+        self.dynamic_loss_scale = sd.get('dynamic_loss_scale', self.dynamic_loss_scale)
+        self.overflow = sd.get('overflow', self.overflow)
```
mrwyattii, line 2203: It may be worthwhile to make these constants to match the style of the other attributes (e.g., `LOSS_SCALER="loss_scaler"`)

mrwyattii (approved):

deepspeed/checkpoint/ds_to_universal.py
tjruwase: Done. Now, users can specify whether slices are replicated or averaged.

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -0,0 +1,303 @@
...
+
+def dump_param_fragment(dir, tp_index, dp_index, state_name, state_flat_tensor, param_name, offset, numel):
+
+    global cnt  # temp hack
+
+    param_base_path = os.path.join(dir, param_name, str(tp_index))
+    os.makedirs(param_base_path, exist_ok=True)
+
+    cnt += 1
+    counter = f"{dp_index:0>2d}"
```
rgtjf, line 130: Dear @tjruwase ,

I'm currently examining a scenario where the maximum dp_index is 127, which, according to alphabetical order, is considered smaller than 13. This raises a question regarding the tensor sorting process in line 144 of our code. Given this context, could there potentially be any issues with how the tensors are sorted due to this ordering?

I appreciate your insight on this matter.

Best regards,
Junfeng

rgtjf: A specific example is as follows:

./temp/model.transformer_encoder.layers.19.self_attn.in_proj_weight => fp32.100, torch.Size([2187604])
./temp/model.transformer_encoder.layers.19.self_attn.in_proj_weight => fp32.99, torch.Size([48144044])

tjruwase: @rgtjf, thanks for sharing this issue. Do you mind creating a new ticket for it? I can see that line 130 may not generalize to larger scales. It would be great if you could share more details in a new ticket. Thanks!

rgtjf: @tjruwase I've opened a ticket to track the issue, #5283. Should there be any missing details or additional information required, please don't hesitate to let me know.

PR #4522, opened 2023-10-17
fix replace lmhead issue when ckpt not load

Yejing-Lai:
When the checkpoint does not load, lm_head will not replace the module. We adjusted the position of replaced lmhead module.


Yejing-Lai:
@RezaYazdaniAminabadi @delock, Pls kindly review. Thanks~

deepspeed/module_inject/replace_module.py
```diff
@@ -327,13 +327,13 @@ def set_lm_head(module):
...
             pbar.update(1)
             gc.collect()
-        replaced_module = set_lm_head(replaced_module)
     else:
         replaced_module = replace_module(model=model,
                                          orig_class=orig_layer_impl,
                                          replace_fn=replace_fn,
                                          _replace_policy=config.injection_policy_tuple)
 
+    replaced_module = set_lm_head(replaced_module)
```
Yejing-Lai, line 336: We move "set_lm_head" outside if...else so that lm_head can be replaced without loading checkpoint.

delock:
@Yejing-Lai can you elaborate a little on the use case that this PR fixes? i.e. which workload, software configuration, etc.

> When the checkpoint does not load, lm_head will not replace the module. We adjusted the position of replaced lmhead module.

delock:
Hi @RezaYazdaniAminabadi , just checked with @Yejing-Lai that this PR makes lm_head fit for use case that the weight tensor are not loaded from checkpoint.  This could happen i.e. in unit test where checkpoint loading is not mandatory.   Could this PR be reviewed?  Thanks!

loadams:
@delock - just checking, still a priority to review this PR and get it merged in?

delock:
Hi  @loadams we discussed this PR internally.  The issue behind this PR only manifests in unit test not real life workload.  So the priority of this PR is not as high as others.  We will close this PR to keep focus on real model related issues.  Thanks!

> @delock - just checking, still a priority to review this PR and get it merged in?

PR #4525, opened 2023-10-17
set device for small zeros tensor to avoid device misalignment

nelyahu:
fixes issue for non CUDA backends which device tensors cannot have CPU data storage.


deepspeed/runtime/utils.py
```diff
@@ -703,7 +703,7 @@ def full(self, device=None):
         dist.all_gather(partition_tensors, partition_tensors[self.rank], group=self.group)
 
         for i in range(len(partition_tensors)):
-            partition_tensors[i].data = torch.zeros(1)
+            partition_tensors[i].data = torch.zeros(1, device=partition_tensors[i].data.device)
```
loadams, line 706: You'll want to run formatting on this file, or just add spaces around the ` = `.

nelyahu: @loadams thanks for pointing me to the formatting failure, it appears to be because of the redundant spaces on other device argument in this commit.

tjruwase (approved):

nelyahu:
It seems like this PR has no meaning anymore, as it modified a removed code. closing.

PR #4530, opened 2023-10-17
Fix the sequence-parallelism for the dense models

RezaYazdaniAminabadi:
This PR fixes some convergence issues for when SP > 1. 
We have seen that the gradients were lower when using SP=2 for a dense model, and by further investigation, we find that the gradients were scaled with the total world size, however, they should have been summed across the SP ranks and averaged on the DP-world. Here is the initial curve comparing grad norm of SP=1 (grey) vs SP=2 (green):
image
After adding the fix for scaling the gradients using the right scale, we get parity for the grad_norm, however, it keeps gradually increasing over time, and results in inferior LM validation (orange: SP1, grey: SP-2). 
image

Fortunately, we are able to fix this by increasing the precision of the gradients before summing them up. The following curves show the LM validation loss of different cases of debugging the SP convergence issue (orange: SP1, grey: SP-2(bf16 gradient), blue: SP2(fp32 gradient)):
image
cc: @samadejacobs @tohtana

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -31,6 +31,7 @@ from deepspeed.checkpoint.constants import (DS_VERSION, GROUP_PADDINGS, PARTITIO
 from deepspeed.utils import link_hp_params
 from deepspeed.checkpoint import enable_universal_checkpoint
 
+from deepspeed.utils import groups
 # Toggle this to true to enable correctness test
 # with gradient partitioning and without
 pg_correctness_test = False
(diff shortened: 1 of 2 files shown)


deepspeed/runtime/engine.py
```diff
@@ -808,6 +808,10 @@ def communication_data_type(self):
 
         return torch.float32
 
+    @communication_data_type.setter
```
tjruwase, line 811: The communication data type should be set via ds_config rather than hardcoding for SP>1. Below is the guide. 

https://github.com/microsoft/DeepSpeed/blob/e238351101bcd23a9a0a32dba0b071685d670d6c/deepspeed/runtime/constants.py#L213-L224

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1395,15 +1397,15 @@ def allreduce_bucket(self, bucket, rank=None, log=None):
 
         tensor_to_allreduce = tensor
 
-        if pg_correctness_test:
+        if pg_correctness_test or self.sequence_parallel_size > 1:
```
tjruwase, line 1400: `self.sequence_parallel_size >1` is now redundant given the ds_config flag, right?

PR #4531, opened 2023-10-18
Turn off I_MPI_PIN for impi launcher

delock:
When using Intel MPI as launcher, Intel MPI would pin cores to ranks with even distribution of all cores in the node.  Sometimes this would be in conflict with DeepSpeed option `--bind_cores_to_rank` and `--bind_core_list`, this can happen in two situtation:
1. User want to use only a portion of cores on each node for computation by using `--bind_core_list`
2. We've been observing Intel MPI sometimes does not assign in 'natual order', i.e.  assign core 0-N to rank 1 instead of rank 0.

Besides, if we turn off `--bind_cores_to_rank` and let Intel MPI pin cores with default configuration, Intel MPI would pin rank to virtual cores instead of physical cores as DeepSpeed does, this is not desired behavior.  In short, Intel MPI pin core mechanism is not as easy to use as deepspeed `--bind_cores_to_rank` and `--bind_core_list` option.

Combine these two situation together, it is rational to set `I_MPI_PIN=0` to avoid conflict between DeepSpeed `numactl` core binding mechanism and Intel MPI core binding mechanism.  This PR turn `I_MPI_PIN=0` when launching through `impi` launcher.


delock:
FYI @tjruwase @dc3671 @Yejing-Lai @inkcherry

tjruwase (approved):

loadams (approved):

PR #4536, opened 2023-10-18
engine.py - save_checkpoint: only rank-0 should create the save dir

nelyahu:
In some NFS it may introduce a race and deadlock between the rank. We found that limiting the creation only to rank-0 can prevent this, while the following barrier ensures that other ranks are not proceeding before dir is created.


deepspeed/runtime/engine.py
```diff
@@ -2994,7 +2994,8 @@ def save_checkpoint(self, save_dir, tag=None, client_state={}, save_latest=True,
         # There seems to be issue creating them in parallel
 
         # Ensure save_dir directory exists
-        self.checkpoint_engine.makedirs(save_dir, exist_ok=True)
+        if self.global_rank == 0:
```
tjruwase, line 2997: This would be a bug when using local node storage is enabled:
https://github.com/microsoft/DeepSpeed/blob/a7358817f59c528ea567902ab8958d5ebb44f803/deepspeed/runtime/engine.py#L791

Please add a guard for this case.

nelyahu: @tjruwase you are correct, maybe letting rank == 0, which in case of local storage is the local rank 0 and in general case is global rank 0.
`rank = self.local_rank if self.use_node_local_storage() else self.global_rank`

tjruwase (approved):

PR #4538, opened 2023-10-18
[ROCm] Add rocblas header

rraminen:
Fixes #4507.
Fixes #4513.

cc: @jithunnair-amd

PR #4539, opened 2023-10-18
Switch from HIP_PLATFORM_HCC to HIP_PLATFORM_AMD

loadams:
This appears to have been deprecated a while ago.  AMD docs [here](https://cgmb-rocm-docs.readthedocs.io/en/latest/Current_Release_Notes/Deprecation.html#changed-hip-environment-variables-in-rocm-v4-1-release) and [here](https://rocm.docs.amd.com/projects/HIP/en/latest/user_guide/hip_porting_guide.html#identifying-hip-target-platform).

### DIFF
--- a/csrc/includes/conversion_utils.h
+++ b/csrc/includes/conversion_utils.h
@@ -265,7 +265,7 @@ DS_D_INLINE float2 to(__nv_bfloat162 val)
 template <>
 DS_D_INLINE __half to(double val)
 {
-#ifdef __HIP_PLATFORM_HCC__
+#ifdef __HIP_PLATFORM_AMD__
     float val_f = __double2float_rn(val);
     return __float2half(val_f);
 #else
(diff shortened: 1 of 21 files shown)


loadams:
@jithunnair-amd if you could review this to make sure I'm implementing this correctly, since I believe this is deprecated: https://rocm.docs.amd.com/projects/HIP/en/latest/user_guide/hip_porting_guide.html#identifying-hip-target-platform

loadams:
Need to wait for the additional one added in #4538.

mrwyattii (approved):

mrwyattii:
Looks like this was added in ROCM 4.1 release (which is 2+ years old). So we shouldn't run into any issues with backwards compatibility, right? I presume most if not all AMD users are on ROCM4.1+

loadams:
> Looks like this was added in ROCM 4.1 release (which is 2+ years old). So we shouldn't run into any issues with backwards compatibility, right? I presume most if not all AMD users are on ROCM4.1+

That's what it looks like to me, and it will be fully deprecated in ROCm 6+ coming later this year, it also looks like that means maybe only the MI25s would be impacted?  So I think we should be fine.

PR #4540, opened 2023-10-19
Fix for stage3 when setting different communication data type

BacharL:


deepspeed/runtime/zero/stage3.py
```diff
@@ -1158,8 +1158,7 @@ def __reduce_and_partition_ipg_grads(self, safe_mode: bool = False) -> None:
 
     @instrument_w_nvtx
     def __avg_scatter_contiguous_grads(self, buffer_to_reduce: Tensor) -> List[Tensor]:
-        dtype = buffer_to_reduce.dtype
-        if self.communication_data_type == self.dtype:
+        if self.communication_data_type != self.dtype:
```
tjruwase, line 1161:
```suggestion
        dtype = buffer_to_reduce.dtype
        if self.communication_data_type != dtype:
```

deepspeed/runtime/zero/stage3.py
```diff
@@ -1159,7 +1159,7 @@ def __reduce_and_partition_ipg_grads(self, safe_mode: bool = False) -> None:
     @instrument_w_nvtx
     def __avg_scatter_contiguous_grads(self, buffer_to_reduce: Tensor) -> List[Tensor]:
         dtype = buffer_to_reduce.dtype
-        if self.communication_data_type == self.dtype:
+        if self.communication_data_type != dtype:
```
tjruwase, line 1162:
```suggestion
        if self.communication_data_type != self.dtype:
```

tjruwase: I think line 1161 is redundant and could be deleted. What do you think?

hablb: I agree. Thanks.

tjruwase (approved):

deepspeed/runtime/zero/stage3.py
hablb: I agree.

PR #4542, opened 2023-10-19
[docs] ZeRO infinity slides and blog

jeffra:


awan-10:
We can probably remove that (2021) with all the paper bullets

PR #4563, opened 2023-10-25
Update lr_schedules.py

CoinCheung:
add cosine annealing scheduler

this scheduler is widely used in image classification task, and many llm (e.g. llama) use this also.


tjruwase:
@CoinCheung, thanks for the PR. A few items to address. 

To fix formatting issues use this [guide](https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites ).

Please add unit test: [example](https://github.com/microsoft/DeepSpeed/blob/ec029e7625ae3e6fb58c87a0483ea2c618ac3d3d/tests/unit/runtime/test_lr_schedulers.py#L311)

Inspect failing CI tests.

CoinCheung:
@tjruwase @wjessup @dfyz @manuelciosici I have no experience with pr for deepspeed, what is the status of this now? Is there any further operation that needs me to work on?

tjruwase:
@CoinCheung, thanks for making the changes. We will review and merge once the CI passes.

CoinCheung:
Hi @tjruwase , 

I have made some fixes, would you please help me launch CI test one more time?

CoinCheung:
@tjruwase Would you please launch CI one more time ?

tjruwase (approved):

CoinCheung:
Hi @jeffra @mrwyattii I think the problem is not with my fix, it is a inference error, but my fix is about training learning rate scheduler. Can this fix be merged ? Or is there other things that need me to commit?
image

tjruwase:
@CoinCheung, sorry for the delay. It seems the issue is with our CI system. Please bear with us while we resolve the problem.

CoinCheung:
Hi @tjruwase , What is the status of this thread? 
image

tjruwase:
@CoinCheung, I have restarted CI. Let's see how it goes.

CoinCheung:
Hi @tjruwase ,

Is this associated with my changes?
image

tjruwase:
@CoinCheung, no I don't think it is related to your changes.

tjruwase (approved):

kmn1024:
Should WarmupCosineLR inherit from WarmupLR?
https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/lr_schedules.py#L774

tjruwase:
> Should WarmupCosineLR inherit from WarmupLR? https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/lr_schedules.py#L774

Yes, you are correct. It should. 

@CoinCheung, are you able to refactor your changes? Thanks!

CoinCheung:
Hi @tjruwase @kmn1024 , I do not think WarmupCosineLR can be interited from this WarmupLR in this case.

SInce they use different methods to determine the learning rates. For WarmupCosineLR, I use "ratio of original lr values", which I think should be more scientific, while WarmupLR uses specific lr values.

For example, when using WarmupLR, by setting `warmup_min_lr=1e-5`, `warmup_num_steps=100`, the scheduler will set lr from `1e-5` to max lr within 100 steps. 
When using WarmupCosineLR, by setting `warmup_min_ratio=0.1`, `warmup_num_steps=100`, and we assume `lr=1e-3` when we defined the **optimizer**, the scheduler would set lr from `0,1 * lr = 1e-4` to max lr within 100 steps.  

The reason why I feel why using ratio is better: we do not need to set specific lr values everywhere in both `optimizer` and `scheduler`. When we define an `optimizer`, we need to consider the learning rate. When we define a `scheduler`, what we only need to do is to determine the shape of the learning rate curves, rather than its specific values.  When you want to keep shape of lr curve and only tune peak lr, you only need to change one place. This follows the principle of "each module only does its own work, and their settings are not impacted by each other".

From my experience of tuning models, this method is less likely to cause mistakes that I change optimizer lr but forgot to change scheduler.

tjruwase:
@CoinCheung, thanks for your response. I agree with the difference that you point out between WarmupLR and WarmupCosineLR, but these differences are to simply in the implementation and logic. However, a the high-level they are similar because of the common two phases of lr changes: (1) initial phase of warmup/increase, and (2) final phase of no change or decay. Looking more closely we observe significant similarity or duplication in many of the methods including `step`, `state_dict`, `load_state_dict`, `get_last_lr`, `_format_param`. These similarities suggest to me an opportunity to refactor and reuse code.

CoinCheung:
@tjruwase Would it be acceptable if we change args of WarmupLR ? It has only one sub-class WarmupDecayLR, and I think its usage frequency is not very high.

PR #4585, opened 2023-10-30
Added __HIP_PLATFORM_AMD__=1 for non JIT build

rraminen:
This PR is required in addition to https://github.com/microsoft/DeepSpeed/pull/4539 to define HIP_PLATFORM_AMD on ROCm.

This is required for DeepSpeed non-JIT build. For JIT build we have https://github.com/microsoft/DeepSpeed/pull/4570.

This is required for DeepSpeed extensions build in those docker images with PyTorch built before https://github.com/pytorch/pytorch/pull/111975.

cc: @jeffdaily @jithunnair-amd


loadams (approved):

op_builder/builder.py
```diff
@@ -614,6 +614,9 @@ def builder(self):
         if not self.build_for_cpu and self.enable_bf16:
             compile_args['cxx'].append("-DBF16_AVAILABLE")
 
+        if self.is_rocm_pytorch():
```
loadams, line 617: @rraminen - do we need both this change and the one in #4570?

rraminen: My understanding is that https://github.com/microsoft/DeepSpeed/pull/4585 is for non-JIT builds and https://github.com/microsoft/DeepSpeed/pull/4570 is for JIT builds. So both of them are required. Please correct me if wrong.

loadams:
The PR failure is not due to changes here, and we will get this merged as we get that other issue solved.

rraminen:
Hi @loadams, is this PR ready to be merged? Thanks.

loadams:
> Hi @loadams, is this PR ready to be merged? Thanks.

Yes, we've identified a CI issue that's unrelated we are trying to fix, will have this merged right after that.

PR #4588, opened 2023-10-31
Universal ckp fixes

mosheisland:


tjruwase:
@stas00, FYI

deepspeed/runtime/engine.py
```diff
@@ -2899,6 +2901,20 @@ def _load_zero_checkpoint(self, load_dir, tag, load_optimizer_states=True):
...
+        optimizer = self.optimizer
+        base_optimizer = optimizer.optimizer
+        state = base_optimizer.state
+        for group in optimizer.param_groups:
+            if 'step' in group:
+                group['step'] = step
+            for p in group['params']:
+                if p in state and len(state[p]) > 0 and 'step' in state[p]:
+                    if isinstance(state[p]['step'], torch.Tensor):
+                        state[p]['step'] = torch.tensor(step, dtype=torch.float, device=p.device)
```
stas00, line 2914: not sure if you want to hardcode fp32 here, perhaps this would be better?
```suggestion
                        state[p]['step'] = torch.tensor(step, dtype=p.dtype, device=p.device)
```

deepspeed/runtime/bf16_optimizer.py
```diff
@@ -75,7 +75,6 @@ def __init__(self,
         self.fp32_groups_gradient_flat_partition = []
         self.fp32_groups_has_gradients = []
 
-        self.step_count = 0
```
stas00, line 78: just checking - you removed this counter because it actually wasn't used anywhere?

deepspeed/runtime/engine.py
```diff
@@ -2899,6 +2901,20 @@ def _load_zero_checkpoint(self, load_dir, tag, load_optimizer_states=True):
             logger.info(f"loading {len(zero_sd_list)} zero partition checkpoints for rank {self.global_rank}")
         return True
 
+    def update_optimizer_step(self, step):
+        optimizer = self.optimizer
+        base_optimizer = optimizer.optimizer
+        state = base_optimizer.state
+        for group in optimizer.param_groups:
+            if 'step' in group:
+                group['step'] = step
```
tjruwase, line 2910: It might be useful to future-proof this in case `group['step']` ever becomes a tensor, as happened with `state[p]['step']`.

deepspeed/runtime/engine.py
mosheisland: state[p]['step'] information holds the step number. I don't think that the step dtype is related to the parameter dtype.
also, torch.optim.adamw uses float (see below):
```
                if len(state) == 0:
                    state['step'] = torch.zeros((1,), dtype=torch.float, device=p.device) \
                        if self.defaults['capturable'] else torch.tensor(0.)
```

deepspeed/runtime/bf16_optimizer.py
mosheisland: correct

deepspeed/runtime/engine.py
mosheisland: will do

deepspeed/runtime/engine.py
stas00: ah, good catch, yes!

There are variations of AdamW that use lower precision than fp32 and we want to make sure those are supported.

But actually I wonder why is it even fp32 in the first place when it could be `uint` - after all it's step counter which is every increasing natural number.

mosheisland: ByteTensor, assuming unsigned-8-bit-byte, will have a max value of 256 - not enough for tracking steps

stas00: heh, that wasn't a very smart suggestion of mine. Good thing you're being vigilant, Moshe.

I guess `torch.IntTensor` would be the most appropriate then, as half would be 64k max - who knows if someone will want more than 64k steps.

I wonder why is it even converted into a tensor in the first place, since `tensor + int` math works just fine.

Would anything break if we ignore the original dtype and assign a normal Python int regardless?

tjruwase: I recall that it was originally `int` and a `group` state, and then changed into tensor and `param` state (in torch or nvidia Adam), which broke DeepSpeed.

stas00: which is odd, since 
```
python -c "import torch; print(torch.ones(1)+1)"
tensor([2.])
```

unless it was expecting `step` to be a tensor and was doing some accessor call, like `step.dtype`

tjruwase: To be clear, the breakage was in checkpoint loading if I recall correctly.

stas00: I see - it must have been expecting a tensor but got an int instead.

which probably means the checkpoint came from elsewhere? Otherwise if you were to save an int, it should have been fine to load an int back.

I don't think there is any issue with how this is done, it's just odd when float is used for something that is not a float.

mosheisland:
@tjruwase, I have an additional commit that adds support for universal checkpoint for **llama** model. Should I add it to this PR? or should I wait for this PR to be merged and create a new PR? Note that it also requires changes in Megatron-DeepSpeed PR https://github.com/microsoft/Megatron-DeepSpeed/pull/276.

Also, I suspect that the unit test failures are not due to changes in this PR, but rather some instable unit tests. I know that other PRs (not mine) are failing on same unit tests.

tjruwase:
@mosheisland, so does that mean this PR is good to go? I was not sure because of the unresolved comments. But yes, it is okay for the llama changes to come in a different PR. No, need to delay this one any longer.

mosheisland:
@tjruwase, my understanding is that there are no open issues. can you please double check?

PR #4591, opened 2023-10-31
Revert "Don't check overflow for bf16 data type (#4512)"

loadams:
This reverts commit 244040c18f463d498c7bd2d7364b1e0d7e544080.


loadams:
This commit was not the root cause of the problem, just happened to pass the PR checks despite it failing them on its own too.

PR #4604, opened 2023-11-03
DeepSpeed-FastGen

cmikeh2:
DeepSpeed-FastGen is built to leverage continuous batching and non-contiguous KV caches to enable increased occupancy and higher responsivity for serving LLMs in the data center, similar to existing frameworks such as TRT-LLM, TGI, and vLLM. In order to achieve a new level of performance, DeepSpeed-FastGen introduces SplitFuse which leverages dynamic prompt and generation decomposition and unification to further improve continuous batching and system throughput.

Corresponding blog: #4607


jeffra (approved):
🚀🚀🎉🎉

PR #4620, opened 2023-11-04
Inference Checkpoints

cmikeh2:


deepspeed/inference/v2/inference_parameter.py
```diff
@@ -0,0 +1,89 @@
...
+
+        return new_tensor
+
+    @classmethod
+    def initialize(cls, core_param: torch.Tensor, **kwargs) -> 'InferenceParameter':
+        """
+        Create the inference parameter.
+        """
+        param = InferenceParameter(core_param)
+        param._aux_attrs = kwargs
```
RezaYazdaniAminabadi, line 61: @cmikeh2, can you please clarify what the aux_attr is used for?

RezaYazdaniAminabadi: are you maybe thinking about scales that are required when adding in the quantization?

deepspeed/inference/v2/model_implementations/common_parameters/embedding_parameters.py
```diff
@@ -23,7 +22,6 @@ class EmbeddingParameter(ParameterBase):
     Vocabulary parameter of shape [vocab_size, model_dim].
     """
 
-    @on_device
     def finalize(self) -> torch.Tensor:
-        return self.params
-        #return self.inference_model.transform_embed_param(self.params)
+        print("EmbeddingParameter.finalize")
```
RezaYazdaniAminabadi, line 26: Do we want to remove the debugging code here?

deepspeed/inference/v2/model_implementations/flat_model_helpers.py
```diff
@@ -0,0 +1,232 @@
...
+from ..inference_utils import elem_size
+
+
+class TensorMetadata(DeepSpeedConfigModel):
+    """
+    A class to represent a tensor specification.
+    """
+    dtype: str
+    shape: Tuple[int, ...]
+    strides: Tuple[int, ...]
```
RezaYazdaniAminabadi, line 25: what is the stride used for here?

RezaYazdaniAminabadi:
@cmikeh2, I am seeing that you are flattening the parameters and use a big allocated memory for it, but I don't see where the saving of the data is happening so that we can later load it! Is it as simple as `torch.save` the injected model with the new inference-engine and reload it back?

deepspeed/inference/v2/model_implementations/flat_model_helpers.py
```diff
@@ -0,0 +1,232 @@
...
+        l_metadata = metadata.layers[l_name]
+
+        for p_name in layer_container._annotation_attrs:
+            p_metadata = l_metadata.params[p_name]
+
+            if p_metadata.core_param.offset == -1:
+                setattr(layer_container, p_name, None)
+                continue
+
+            dummy_tensor = torch.empty([], dtype=STR_TO_DTYPE[p_metadata.core_param.dtype])
```
RezaYazdaniAminabadi, line 212: can we not pass just the string to figure out the dtype in the alloc_fn rather than creating a dummy tensor with it?

deepspeed/inference/v2/model_implementations/flat_model_helpers.py
cmikeh2: If we ever have some kind of non-contiguous parameter, this will make sure we return it correctly as is rather than silently changing the underlying storage. It shouldn't be the common case, but is low overhead to support and might remove some weird bugs in the future.

cmikeh2:
> @cmikeh2, I am seeing that you are flattening the parameters and use a big allocated memory for it, but I don't see where the saving of the data is happening so that we can later load it! Is it as simple as `torch.save` the injected model with the new inference-engine and reload it back?

The save is exposed as a `serialize` method on the engine here: https://github.com/microsoft/DeepSpeed/blob/0faea352f028865476f262bb2d0974981e28e25a/deepspeed/inference/v2/engine_v2.py#L222. It's mostly just going to save a couple of configs and the large tensor though.

The load codepath is here: https://github.com/microsoft/DeepSpeed/blob/0faea352f028865476f262bb2d0974981e28e25a/deepspeed/inference/v2/model_implementations/inference_policy_base.py#L208.

deepspeed/inference/v2/model_implementations/flat_model_helpers.py
cmikeh2: Strings would also work. The dummy tensor has no storage, and I'd really prefer to just pass the `torch.dtype`, but I couldn't figure out the correct argument for that. The whole binding around `blob` is kind of ugly, but when I tried using `UntypedStorage` which is supposedly the correct way there were phantom allocations I was never able to track down.

deepspeed/inference/v2/inference_parameter.py
cmikeh2: That's exactly right, scales for quantization or anything other metadata we create when transforming the parameter can be stored as an auxiliary attr. The declaration for something like that would be:

```python
p = InferenceParameter.initialize(param, scales=scales)

assert torch.equal(p, param)
assert torch.equal(p.scales, scales)
```

deepspeed/inference/v2/engine_factory.py
```diff
@@ -3,44 +3,66 @@
...
-        policy = OPTPolicy(checkpoint_engine, model_config)
-    elif model_config.model_type == "llama":
-        from .model_implementations.llama_v2.llama_v2_policy import Llama2Policy
-        policy = Llama2Policy(checkpoint_engine, model_config)
-    elif model_config.model_type == "mistral":
-        from .model_implementations.mistral.policy import MistralPolicy
-        policy = MistralPolicy(checkpoint_engine, model_config)
+    if os.path.exists(os.path.join(path, "ds_model_config.pkl")):
+
+        # Load metadata, for grabbing the policy name we'll have all ranks just check for
```
RezaYazdaniAminabadi, line 35: I think this part needs to be supported in a higher abstraction layer, as it is not just related to HF models and we should be able to use it with different checkpoint formats.

PR #4634, opened 2023-11-06
[MII] catch error wrt HF version and Mistral

jeffra:
If a user is trying to use Mistral with MII and running `transformers <= 4.34.0` they will get a cryptic KeyError from transformers. This is documented on the Mistral model page https://huggingface.co/mistralai/Mistral-7B-v0.1 but is an easy error to run into for users.

This PR adds a version check to error out more gracefully. This should address at least part of the issue this user reported: https://github.com/microsoft/DeepSpeed-MII/issues/258


cmikeh2 (approved):
Only comment from me is if it might make more sense to do this in the engine factory rather than the policy itself.

mrwyattii (approved):

PR #4639, opened 2023-11-06
Loadams/torch cpu mismatch

loadams:


loadams:
@jinzhen-lin - could you take a look at this change since it modifies your code from #3085?  I want to make sure I understand where the `torch.cuda.is_available()` was unavailable from.

jinzhen-lin:
@loadams I think it is ok for `jit_load`.

tjruwase (approved):

jeffra (approved):

PR #4657, opened 2023-11-08
Fix the openfold training.

cctry:
This PR removes the bias created as placeholders, which causes a crash in openfold's training pipeline.


cctry:
@conglongli

conglongli (approved):

Issue #272, opened 2020-06-21
How to reproduce BERT perf results in deepspeed blog


LiweiPeng:
Hi Deepspeed team,

The BERT perf results from blog [Microsoft DeepSpeed achieves the fastest BERT training time](https://www.deepspeed.ai/news/2020/05/27/fastest-bert-training.html) are very impressive. Can you share the detailed parameters for Figure 1 in the blog? Specifically, can you share answers to the following questions?
- When testing nvbert and huggingface bert, what mixed precision is used?
- If apex amp is used, what options are used in amp.initialize()?
- For Deepspeed bert, did you use zero optimizer? If so, which one?

Thanks.
Liwei

RezaYazdaniAminabadi:
Hi Liwei,

Thanks for trying out the deepspeed transformer kernel. 
I assume you use a 32-GB NVIDIA V100 GPU for this experiment, am I right? 
I wonder what total batch size you are using. I mean if it is with gradient accumulation of 1 or higher. Also, is it a post-ln or pre-ln BERT layer that you are running the training with? 
We use the pre-ln BERT structure and we set the gradient accumulation to 10 for all our experiments. 
One last thing is that which dataset you are using for the training?

Thanks.
Reza

LiweiPeng:
@RezaYazdaniAminabadi , Thanks for your quick response.

To answer your questions:
- I am using one 32-GB Nvidia V100 GPU. 
- I am using the post-layernorm in [nvidia's code](https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/LanguageModeling/BERT) because this is the only mode nvbert supports.
- I am using English dataset (wikipedia + bookcorpus).
- I tried 2 sets of test: 
  test1: max batch=136, gradient accumulation=1, perf is 194.57 examples/s.
  test2: max batch=132, gradient accumulation=10, perf is 200.50 examples/s.

One difference from your test is that I am using post-ln not pre-ln. I am not sure whether this can explain the max-batch size difference and perf difference or not. I'll try pre-ln later. 

Thanks.
Liwei

RezaYazdaniAminabadi:
Thanks for the reply. If you want to try pre-ln model, you may use the example in deepspeed: https://github.com/microsoft/DeepSpeedExamples/blob/8610e5e3fcce5fb247e3b85ea2bed0f2296b5443/bing_bert/nvidia/modelingpreln.py
[...]

PR #4659, opened 2023-11-09
Add safetensors support

jihnenglin:
In response to [DeepSpeed-MII issue #272](https://github.com/microsoft/DeepSpeed-MII/issues/272)

Introduce compatibility for HF repositories containing solely `.safetensors` checkpoints. In case of coexistence, prioritize `.safetensors` checkpoints over `.bin` ones.


jihnenglin:
@microsoft-github-policy-service agree company="Gigabyte Technology"

jeffra:
This is great, thank you @jihnenglin! Can you also create a PR on the MII side to update https://github.com/microsoft/DeepSpeed-MII/blob/main/requirements/requirements.txt to add `safetensors` as a dependency?

jihnenglin:
> This is great, thank you @jihnenglin! Can you also create a PR on the MII side to update https://github.com/microsoft/DeepSpeed-MII/blob/main/requirements/requirements.txt to add `safetensors` as a dependency?

Sure, will do!

deepspeed/inference/v2/checkpoint/huggingface_engine.py
```diff
@@ -51,6 +53,7 @@ def _fetch_checkpoint_files(self):
                                                                "*.bin",
                                                                "*.json",
                                                                "*.pt",
+                                                               "*.safetensors",
```
mrwyattii, line 56: In the case that a model has both `*.bin` and `*.safetensors`, will this cause `snapshot_download` to download the model weights twice?

is: Indeed there is this problem. Could we consider adding an extra parameter like hf transformers, safetensors=True/False, to adjust allow_pattern?

jihnenglin: Thank you for your feedback. I appreciate it. While developing the Safetensors model, my focus was primarily on downloading files with the `*.safetensors` extension. Upon further consideration, I have decided to introduce a new argument, `prioritize_safetensors: bool = True`, for this function. With this enhancement, the function will attempt to retrieve the prioritized format, but if unavailable, it will seamlessly fetch the alternative file type. This ensures flexibility, allowing users to choose their preferred model format.

mrwyattii: @jihnenglin thanks for updating this. I think we should try to make this a bit more automatic and seamless for the user. Rather than passing a `prioritize_safetensors` argument, we should automatically go the safetensor route if they are available and fallback to the .bin files if they are not. @jeffra thoughts?

jihnenglin: @mrwyattii Sounds great! I was not sure if you want to make it load `.safetensors` by default, so I kept both the options. But of course I'd be willing to make it more intuitive for the users. I will change it later.

jihnenglin:
I've made it automatically fetch `*.safentensors` checkpoints by default, and only fallback to download `*.bin` checkpoints if there are no `*.safentensors`.

mrwyattii (approved):

mrwyattii:
> I've made it automatically fetch `*.safentensors` checkpoints by default, and only fallback to download `*.bin` checkpoints if there are no `*.safentensors`.

Thanks @jihnenglin! I cleaned up your implementation a little bit so we could avoid some of the deeper nested if-else branching and make it a bit easier to read the code. LGTM

loadams:
The only a6000 test failing is a known failure, so we could bypass this if we need.

loadams (approved):

PR #4660, opened 2023-11-09
Fix for when prompt contains an odd num of apostrophes

oelayan7:
The apostrophes in args.user_args are mixed with the single quotes used to enclose x in the parsing code.

For example,
`deepspeed --num_nodes 1 --num_gpus 8 --no_local_rank run_generation.py --model_name_or_path tiiuae/falcon-40b --max_new_tokens 128 --prompt "I'm a student"`

instead of (which doesn't work):
`deepspeed --num_nodes 1 --num_gpus 8 --no_local_rank run_generation.py --model_name_or_path tiiuae/falcon-40b --max_new_tokens 128 --prompt 'I'm a student'`

To resolve this issue, we can make a change to enclose x in double quotes instead of single quotes.


loadams (approved):

lekurile (approved):

deepspeed/launcher/runner.py
```diff
@@ -390,7 +390,7 @@ def main(args=None):
     args = parse_args(args)
 
     # For when argparse interprets remaining args as a single string
-    args.user_args = shlex.split(" ".join(list(map(lambda x: x if x.startswith("-") else f"'{x}'", args.user_args))))
+    args.user_args = shlex.split(" ".join(list(map(lambda x: x if x.startswith("-") else f'"{x}"', args.user_args))))
```
mrwyattii, line 393: I think this would just be exchanging one problem for another. For example a `--prompt 'I am 72" tall'` would encounter the problem with these changes. I think the correct thing would be to escape any `'` characters: `{x.replace("'", "\'")}`, but I have not tested if this works.

mrwyattii (changes requested):

deepspeed/launcher/runner.py
oelayan7: The outer wrapper for prompt should be double quotation (")
But I get your point, this was my first approach before finding out that inner quotations are ignored.

In other words, it ignores double quotes in the command. seems it's a known issue. so deepspeed will never see " as part of the prompt

For example: 
--prompt "He said "Hello,"" 
will be caught as: 
['--prompt', 'He said Hello,']

mrwyattii: I see, thank you for the explanation. In that case, I think this solution is better than what we currently have.

PR #4664, opened 2023-11-10
Inference Checkpoints in V2

cmikeh2:
Add capability to snapshot an engine and resume from it, reducing load times for large models.


RezaYazdaniAminabadi (approved):

deepspeed/inference/v2/engine_factory.py
```diff
@@ -3,49 +3,104 @@
...
+from .model_implementations import (
+    OPTPolicy,
+    Llama2Policy,
+    MistralPolicy,
+)
+from .model_implementations.inference_policy_base import POLICIES, InferenceV2Policy
+from .model_implementations.flat_model_helpers import make_metadata_filename, ModelMetadata
+
+
+def buid_engine_from_ds_checkpoint(path: str,
```
jeffra, line 25: typo? buid -> build?

cmikeh2: Updated there (and the call site)

deepspeed/inference/v2/modules/implementations/linear/blas_fp_linear.py
```diff
@@ -70,14 +71,15 @@ def __init__(self, config: DSLinearConfig, implementation_config: Dict[str, Any]
...
-    def transform_param(self, param: torch.Tensor) -> torch.Tensor:
+    def transform_param(self, param: torch.Tensor) -> InferenceParameter:
         """
         Converts param to same data type as input and output.
 
         Parameters:
             param (torch.Tensor): Weight or bias tensor.
         """
-        return param.to(self._config.input_dtype)
+        param = param.to(self._config.output_dtype)
```
awan-10, line 81: Checking why we need to do self._config.output_dtype instead of self._config.input_dtype?

awan-10 (approved):

jeffra (approved):

PR #4668, opened 2023-11-11
KV Cache Improved Flexibility

cmikeh2:
This KV-cache adds the foundation for appropriately supporting two key KV-cache improvements:

1. Delineation between local/dense KV caches/models at the cache level in addition to the attention module level.
2. Support for multiple types of disjoint KV caches (such as alternating local + dense attention GPT-Neo).

Follow up item: Determine appropriate statistics for weighting local + dense KV block ratios when both are present.


awan-10 (approved):

deepspeed/inference/v2/ragged/ragged_manager.py
```diff
@@ -124,7 +116,9 @@ def flush_sequence(self, uid: int) -> None:
             return
 
         seq = self._seqs[uid]
-        self._kv_cache.free(seq.all_block_ids)
+        for i in range(self.n_kv_caches):
```
jeffra, line 119: `self.n_kv_caches` doesn't exist :(

image

PR #4670, opened 2023-11-11
fix using torch.half with transformers

abdelmoez98:
fix for this error "AttributeError: 'DeepSpeedGPTInference' object has no attribute 'dtype'" while using torch.half


loadams (approved):

deepspeed/model_implementations/transformers/ds_transformer.py
```diff
@@ -163,7 +163,7 @@ def forward(
 
         if (self.config.dtype in [torch.float16, torch.bfloat16, torch.int8]) \
             and input.dtype == torch.float:
-            target_dtype = torch.half if self.dtype == torch.int8 else self.dtype
+            target_dtype = torch.half if self.config.dtype == torch.int8 else self.config.dtype
```
loadams, line 166: @lekurile - not sure if you're the best person to tag for this or not, but I think we also have this in the ds_llama2.py file and it probably needs the same fix, FYI

lekurile: @abdelmoez98 I think this entire line can be replaced with:
``` Python
 target_dtype = self.data_type
```

Since we definte `self.data_type` as:
``` Python
data_type = torch.half if self.config.dtype == torch.int8 else self.config.dtype
```

In the constructor.

@loadams Good point on `ds_llama2.py`, that change should be made there as well. @abdelmoez98 can you please add that there?

lekurile (changes requested):

mrwyattii:
#4814 is already merged. Closing

PR #4675, opened 2023-11-14
Delete unnecessary wait event

CurryRice233:
The `cudaStreamWaitEvent` is typically used to wait for other streams in a multi-stream situation, but in this case (stopping the timer) there is only one stream,  waiting an event on the same stream is unnecessary. 

Other accelerators, such as the NPU, will force the user to use the Event actions correctly. Waiting an event on the same stream, NPU will throw an exception.

Here's an example of a cuda timer, just Record --> Synchronize --> ElapsedTime.
https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#using-cuda-gpu-timers


tjruwase:
@CurryRice233, events were added to the timers so that the timers can be seamlessly used in single-stream and multi-stream scenarios. Since this code has worked correctly in other accelerators, I am curious why NPU is throwing an exception in this case. Is it possible to adjust NPU to align with other accelerators?

CurryRice233:
> @CurryRice233, events were added to the timers so that the timers can be seamlessly used in single-stream and multi-stream scenarios. Since this code has worked correctly in other accelerators, I am curious why NPU is throwing an exception in this case. Is it possible to adjust NPU to align with other accelerators?

Correct, the better way is to align with other accelerators. Thank you very much, we will change the exception to a warning.

PR #4688, opened 2023-11-16
[Bug fix] WarmupCosineLR issues

sbwww:
Original code missing a `self.` before `warmup_num_steps` so that `warmup_num_steps` might be 0 and cause math domain error when doing `math.log(0)`

```py
        self.warmup_num_steps = max(2, warmup_num_steps)
        self.inverse_log_warm_up = 1.0 / math.log(warmup_num_steps)
```


tjruwase (approved):

sbwww:
Perhaps `WarmupCosineLR` should be refactored as a derived class of `WarmupLR` like `WarmupDecayLR` did, instead of being a derived class of `object`

tjruwase:
@sbwww, I will merge this PR. So you can do the refactor as a separate PR. Thanks!

PR #4695, opened 2023-11-16
Communication Optimization for Large-Scale Training

RezaYazdaniAminabadi:
This PR brings in some communication optimization for large-scale training systems for both dense and MoE architectures. In particular, we have focused on the backward communication collectives, such as AllReduce and AllGather, which are used for ZeRO stages 1 and 2.  Also, we have added some optimization for the sequence parallelism to reduce the All2All cost. With these optimizations, we improve the training scalability, as we see proportional boost of the training speed when increasing the number of GPUs.


blogs/comm-opt/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+</div>
+
+
+## Table of Contents
+1. [Introduction](#introduction)
+2. [Parameter-update All-Reduce Optimization](#ar-opt)
+3. [ZeRO2 Gradient All-Gather Optimization](#ag-opt)
+4. [Sequence-Parallel Optimization](#sp-opt)
+5. [Sequence-Parallel Convergence Consideration](#convergence)
```
tjruwase, line 13: I think this convergence consideration is out-of-scope for communication optimizations.

blogs/comm-opt/README.md
```diff
@@ -0,0 +1,94 @@
...
+</div>
+By changing this layout from E + D to D + E (shown in Fig 1), where we first replicate each expert by D times and then add them across expert-parallel dimension, we can reduce the all-reduce time substantially. On an A100-DGX cluster, where each node has 8 GPUs, we see about 8x reduction in cross-node communication-volume for the parameter update process. Note that by adding this optimization, we increase the cost of All-to-All happening for the MoE part of the model, however, we have seen that the preformance benfit of all-reduce overweighs this cost.
+
+After applying these two techniques for our training scenario, we reduce the all-reduce time by 2x for dense architecture and 5x - 8x for the MoE one using the multi-rank bucketing technique, and another 2x using the new D + E layout for the MoE architecture. We see higher performance gain on MoE architectures when using large number of GPUs. For the end-to-end training of a 7B-base MoE architecture, we reduce iteration-time from 13 sec to 9.5 sec on 512 GPUs (37%) and from 16 sec to 4.9 sec on 1k-GPU setup (3.3x).
```
tjruwase, line 40: I think this performance results would be better presented using tables comparing optimized vs baseline. It will also be good to provide more experimental details like GPU type and count.

blogs/comm-opt/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+## 3. ZeRO2 Gradient All-Gather Optimization <a name="ag-opt"></a>
```
tjruwase, line 42: I believe ZeRO2 does fp16/bf16 all-gather not gradient all-gather.

blogs/comm-opt/README.md
```diff
@@ -0,0 +1,94 @@
...
+2. [Parameter-update All-Reduce Optimization](#ar-opt)
+3. [ZeRO2 Gradient All-Gather Optimization](#ag-opt)
+4. [Sequence-Parallel Optimization](#sp-opt)
+5. [Sequence-Parallel Convergence Consideration](#convergence)
+
+
+## 1. Introduction <a name="introduction"></a>
+Training LLMs on large data-set can be costly both in terms of resource-budget and time-to-completion. An important step to minimize such cost is to find the right number of resources together with a scalable library that assures training under a time limit. In this post, we target a key component of the scalability feature at DeepSpeed, the communication optimization. There are several communication collectives used in the main DeepSpeed's technologies, which we go through their optimization in detail in the following sections.
+
+## 2. Parameter-update All-Reduce Optimization <a name="ar-opt"></a>
```
tjruwase, line 19: I believe this should be gradient all-reduce, not parameter-update all-reduce.

tjruwase: I think a better title would be something like "Optimizing Gradient Reduction for Mixture of Expert training".

blogs/comm-opt/README.md
tjruwase: Suggested title "Optimizing Parameter All-Gather for ZeRO2 training"

blogs/comm-opt/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+## 3. ZeRO2 Gradient All-Gather Optimization <a name="ag-opt"></a>
+
+The same as with all-reduce, all-gather takes longer as we have more partitions. As the parameters are stored in a flattened buffer for ZeRO stage-2, we can simply have a one call to all-gather the parameters into this tensor.
+
+When all-gathering the updated parameters at Zero-Stage2, the bucketing scheme uses several narrow operations and creates a list of tensors with the bucket size from each partition. We needed this scheme to align with the `all_gather` operation from PyTorch. 
+However, by adding the support for the `all_gather_into_tensor`, operation that has been added to the newer versions of PyTorch, we can simply have a kernel call to do the full-parameter all-gather. With this optimization, we see about 2x reduction in the step time for large-scale training. 
+
+## 4. Sequence-Parallel Optimization <a name="sp-opt"></a>
```
tjruwase, line 49: Suggested titles:
1. "Optimizing AlltoAll for Sequence-Parallel Training".
2. AlltoAll Optimizations for Sequence-Parallel Training".

blogs/comm-opt/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+## 2. Parameter-update All-Reduce Optimization <a name="ar-opt"></a>
+
+Before diving into this techniques, let's take a step back and show some of the case studies that motivates this work of all-reduce optimization. 
+
+All-reduce operation is an important part of the the training process. In ZeRO, we handle this in buckets, which can be configured to get good comminication throughput. As the number of GPUs increases, we encounter smaller-partition all-reduces. In this case, the current bucketing scheme cannot help with the communication overhead. This mostly becomes an issue when training smaller-scale models (like Llama-7B) with large number of GPUs. 
+
+For instance, when training a dense-7B architecture with Zero stages 1 or 2, we encounter a 1 and 2 second increase for the all-reudce time by increasing from 256 to 512 and 1024 A100 GPUs. This issue mostly arises from the fact that, the gradient-averaging happens with smaller partitions (#parameters / #GPUs) per-GPU rank. This issue gets more serious when training MoE architectures (3 - 12 second) for which the expert's parameters can be farther away due to the current parallelism layout of data and expert parallelism.
```
tjruwase, line 25: It would be better to present these results in table form showing the all-reduce time for different GPU count, and for dense and MoE models.

blogs/comm-opt/README.md
RezaYazdaniAminabadi: I think this relates to both dense and MoE architecture. Of course the second approach is only for MoE models, but the first one can apply for all the architectures.

blogs/comm-opt/README.md
RezaYazdaniAminabadi: Even though this is not really part of the optimization, however, we arrive to this issue due to the optimizations that we did. Of course, this is related to the Sequence-Parallelism optimization. So, do you think I can move it to be part of the Ulysses blog post?

blogs/comm-opt/README.md
RezaYazdaniAminabadi: I don't think there are too many data-points that make the text too verbose, however I am fine to add the table if you think this is the case.

blogs/comm-opt/README.md
RezaYazdaniAminabadi: Thanks for the suggestion @tjruwase , I will do that.

blogs/comm-opt/README.md
RezaYazdaniAminabadi: I have changed a bit the intro to show how this can be relevant in this post.

blogs/comm-opt/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Table of Contents
+1. [Introduction](#introduction)
+2. [Gradient AllReduce Optimization for ZeRO stages 1 and 2](#ar-opt)
+3. [Optimizing Parameter All-Gather for ZeRO2 Training](#ag-opt)
+4. [Optimizing AlltoAll for Sequence-Parallel Training](#sp-opt)
+
+
+## 1. Introduction <a name="introduction"></a>
+Training LLMs on large data-set can be costly both in terms of resource-budget and time-to-completion. An important step to minimize such cost is to find the right number of resources together with a scalable library that assures training under a time limit. In this post, we target a key component of the scalability feature at DeepSpeed, the communication optimization. There are several communication collectives used in the main DeepSpeed's technologies, which we go through their optimization in detail in the following sections. We made these optimization available in DeepSpeed versions >= 0.x.x.
```
tjruwase, line 16:
```suggestion
Training LLMs on large datasets can be extremely costly both in terms of hardware resources and time. An important step to minimize such costs is to carefully combine an appropriate number of resources together with a scalable library that guarantees training completion within a time limit. In this post, we discuss a key aspect of the scalability features of DeepSpeed, the communication optimization. Communication collectives (e.g., all-reduce, all-gather, etc.) are critical pieces of many popular DeepSpeed technologies (e.g., ZeRO, MoE, AutoTP, etc.), and in the following sections we discuss our new optimizations of some of these collectives. These optimizations are available in DeepSpeed versions >= 0.x.x.
```

blogs/comm-opt/README.md
```diff
@@ -0,0 +1,82 @@
+<div align="center">
+
+# Communictaion Optimization for Large-Scale Training
```
tjruwase, line 3:
```suggestion
# Communication Optimization for Large-Scale Training
```

blogs/comm-opt/README.md
```diff
@@ -0,0 +1,82 @@
...
+3. [Optimizing Parameter All-Gather for ZeRO2 Training](#ag-opt)
+4. [Optimizing AlltoAll for Sequence-Parallel Training](#sp-opt)
+
+
+## 1. Introduction <a name="introduction"></a>
+Training LLMs on large data-set can be costly both in terms of resource-budget and time-to-completion. An important step to minimize such cost is to find the right number of resources together with a scalable library that assures training under a time limit. In this post, we target a key component of the scalability feature at DeepSpeed, the communication optimization. There are several communication collectives used in the main DeepSpeed's technologies, which we go through their optimization in detail in the following sections. We made these optimization available in DeepSpeed versions >= 0.x.x.
+
+## 2. Gradient AllReduce Optimization for ZeRO stages 1 and 2 <a name="ar-opt"></a>
+
+Before diving into this techniques, let's take a step back and show some of the case studies that motivates this work of AllReduce optimization.
```
tjruwase, line 20:
```suggestion
Before diving into this optimization, let's take a step back and show some of the case studies that demonstrate the need.
```

tjruwase (approved):

jeffra (approved):
🚀

GeneZC:
Why USE_DATA_BEFORE_EXPERT_PARALLEL_DEFAULT is default to False instead of True while you have mentioned True will increase the throughput?

https://github.com/microsoft/DeepSpeed/blob/4d866bd55a6b2b924987603b599c1f8f35911c4b/deepspeed/runtime/constants.py#L441

RezaYazdaniAminabadi:
> Why USE_DATA_BEFORE_EXPERT_PARALLEL_DEFAULT is default to False instead of True while you have mentioned True will increase the throughput?
> 
> https://github.com/microsoft/DeepSpeed/blob/4d866bd55a6b2b924987603b599c1f8f35911c4b/deepspeed/runtime/constants.py#L441

that's a good question. the main reason that i didn't enable it by default is that its improvement is very much dependent on the cluster specification that you have. if you are using a multi-node job with high infini-bw communication then it is a very good optimization and should be enabled, otherwise, the all-to-all performance will be terrible and it will slow down the training process significantly. i will add a note about this in a separate PR. 
thanks @GeneZC

GeneZC:
Ok, I see. Then how about the allreduce performance across nodes if it is set as False?

RezaYazdaniAminabadi:
> Ok, I see. Then how about the allreduce performance across nodes if it is set as False?

it should be still faster than all-to-all if you don't have high dp_size for your expert parameters, however, it is going to be slower if you have larger dp_size (like if you are using 4x or 8x more gpus than your expert-parallel degree). but it would be good that you do some tests on your HW to verify this.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1405,17 +1455,18 @@ def allreduce_bucket(self, bucket, rank=None, log=None):
...
 
         if rank is None:
             #    "All Reducing"
-            dist.all_reduce(tensor_to_allreduce, group=self.dp_process_group)
+            dist.all_reduce(tensor_to_allreduce, group=process_group)
         else:
-            global_rank = dist.get_global_rank(self.dp_process_group, rank)
-            dist.reduce(tensor_to_allreduce, global_rank, group=self.dp_process_group)
+            global_rank = dist.get_global_rank(process_group, rank)
+            dist.reduce(tensor_to_allreduce, global_rank, group=process_group)
```
CurryRice233, line 1466: Never enter this branch because `rank = None` in L1443

CurryRice233: fixed https://github.com/microsoft/DeepSpeed/pull/4946

PR #4707, opened 2023-11-20
Nvme offload checkpoint

### LINKED ISSUE #2082: Saving a checkpoint when training with NVMe offloading?
I am testing NVMe offloading when training a model. When I try to save a checkpoint, I am getting (full stack trace below):
```
NotImplementedError: ZeRO-3 does not yet support checkpointing with NVMe offloading, please disable for now.
```
Is that correct? There is no checkpointing with NVMe offloading or am I missing something in my setup/config file? If there is no checkpointing, how can I save the model? 

Full stacktrace:
```
[2022-07-08 19:51:21]
[4e2c01e1] [rank=0] Traceback (most recent call last): <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/opt/conda/lib/python3.8/runpy.py", line 194, in _run_module_as_main <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     return _run_code(code, main_globals, None, <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/opt/conda/lib/python3.8/runpy.py", line 87, in _run_code <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     exec(code, run_globals) <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/run/determined/pythonuserbase/lib/python3.8/site-packages/determined/exec/harness.py", line 132, in <module> <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     sys.exit(main(args.train_entrypoint)) <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/run/determined/pythonuserbase/lib/python3.8/site-packages/determined/exec/harness.py", line 123, in main <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     controller.run() <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/run/determined/pythonuserbase/lib/python3.8/site-packages/determined/pytorch/deepspeed/_deepspeed_trial.py", line 296, in run <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     self._run() <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/run/determined/pythonuserbase/lib/python3.8/site-packages/determined/pytorch/deepspeed/_deepspeed_trial.py", line 338, in _run <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     self._save(path) <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/run/determined/pythonuserbase/lib/python3.8/site-packages/determined/pytorch/deepspeed/_deepspeed_trial.py", line 732, in _save <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     self.trial.save(self.context, path) <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/run/determined/pythonuserbase/lib/python3.8/site-packages/determined/pytorch/deepspeed/_deepspeed_trial.py", line 934, in save <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     m.save_checkpoint(path, tag=f"model{i}") <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/opt/conda/lib/python3.8/site-packages/deepspeed/runtime/engine.py", line 2842, in save_checkpoint <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     self._save_zero_checkpoint(save_dir, tag) <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/opt/conda/lib/python3.8/site-packages/deepspeed/runtime/engine.py", line 3106, in _save_zero_checkpoint <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     zero_sd = dict(optimizer_state_dict=self.optimizer.state_dict(), <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]   File "/opt/conda/lib/python3.8/site-packages/deepspeed/runtime/zero/stage3.py", line 2699, in state_dict <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0]     raise NotImplementedError( <none> [2022-07-08 19:51:21]
[4e2c01e1] [rank=0] NotImplementedError: ZeRO-3 does not yet support checkpointing with NVMe offloading, please disable for now. 
```

Config file:
```
{
  "train_batch_size": 256,
  "steps_per_print": 2000,
  "optimizer": {
    "type": "Adam",
    "params": {
      "lr": 0.001,
      "betas": [
        0.8,
        0.999
      ],
      "eps": 1e-8,
      "weight_decay": 3e-7
    }
  },
  "scheduler": {
    "type": "WarmupLR",
    "params": {
      "warmup_min_lr": 0,
      "warmup_max_lr": 0.001,
      "warmup_num_steps": 1000
    }
  },
  "gradient_clipping": 1.0,
  "prescale_gradients": false,
  "fp16": {
      "enabled": true,
      "fp16_master_weights_and_grads": false,
      "loss_scale": 0,
      "loss_scale_window": 500,
      "hysteresis": 2,
      "min_loss_scale": 1,
      "initial_scale_power": 15
  },
  "zero_optimization": {
      "stage": 3,
      "contiguous_gradients": true,
      "stage3_max_live_parameters": 1e9,
      "stage3_max_reuse_distance": 1e9,
      "stage3_prefetch_bucket_size": 1e7,
      "stage3_param_persistence_threshold": 1e5,
      "reduce_bucket_size": 1e7,
      "sub_group_size": 1e9,
      "offload_param": {
        "device": "nvme",
        "nvme_path": "nvme0n1",
        "pin_memory": false
      },
      "offload_optimizer": {
        "device": "nvme",
        "nvme_path": "nvme0n1",
        "pin_memory": false
    }
  }
}
```

eisene:
Previous PR #4416 had too many issues, closing that one and re-opening. This PR includes a passing test.

This is a proposal for an implementation of checkpointing models when training with ZeRO-3 with NVMe offload:

1. Currently, the names of the files used in the checkpoint are based on the Python id of the parameter object, which is just the parameter's address in memory. This is not stable across runs, which has two disadvantages:
   - The NVMe offloading files grow with every run of the model even if the architecture did not change. This wastes disk space and, at least for me, was a surprise when I first saw it. It is not related to checkpointing.
   - Without a way to match the file to the offloaded tensor we can't reload the checkpoint.

   We propose an alternative naming scheme. The parameters are named after their ds_id instead of their Python id, and the tensors are named after their state_name and (new) parameter id.

2. A model checkpoint now has to include all the offloaded tensor files. During checkpoint save/load we copy all the tensor files to/from the "offloaded_tensors" subdirectory of the checkpoint. We provide some logging on the remaining space on the file system due to the potential size of these files, especially as they accumulate in each checkpoint. We do not copy the gradient files.

3. When loading the checkpoint, the optimizer already has prepared buffers for swapping. We need to purge them so that they are replaced with the freshly copied on-disk buffers from the checkpoint.

The key differences between this PR and the previous one:

 - There's a test for a simple model with parameter/optimizer offload set to cpu/cpu, cpu/nvme and nvme/nvme.
 - Gradient files are not copied.
 - FP16 and FP32 parameter buffers are handled correctly during load.

Fixes #2082.


gary-young:
Thank you! I hope your PR can be merged into the master soon.

deepspeed/runtime/engine.py
```diff
@@ -2740,6 +2745,19 @@ def load_checkpoint(self,
             if not success:
                 self.optimizer._restore_from_bit16_weights()
 
+        if self.zero_has_nvme_offload():
+            from shutil import copytree, disk_usage
+            offload_dir = self.optimizer.optimizer_swapper.swap_folder
+            offload_ckpt_dir = os.path.join(load_dir, tag, "offloaded_tensors")
```
tjruwase, line 2751:
```suggestion
            offload_ckpt_dir = os.path.join(load_dir, tag, OFFLOADED_TENSORS_DIR)
```
Replace "offloaded_tensors" with symbolic, such as above, which should be defined in deepspeed/checkpoints/constants.py

tjruwase (approved):

deepspeed/runtime/engine.py
tjruwase: @eisene, the PR looks great to me as is. My only concerns are the hardcoded constants such as this one. Do you have bandwidth to address at this time? Thanks!

eisene: @tjruwase this seems very straightforward, I'll do it soon.

loadams (approved):

deepspeed/runtime/engine.py
tjruwase: @eisene, thanks. I will merge this now as it is not a reason to hold up this useful feature.

w4-jihunlorenzopark:
Hello, I have a question about one of the design details in your PR. 

There is a copytree command to copy the whole tree of nvme offload directory when [saving](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L3183)/[loading](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L2840) checkpoint, but I found `dirs_exist_ok` is true when loading the checkpoint, but it's not when saving the checkpoint.

What would be the difference between them?

I am facing the `FileExistsError` during saving `mytraining/checkpoint-8/global_step8/offloaded_tensors` and thinking it can be solved by setting `dirs_exist_ok` true.

w4-jihunlorenzopark:
@tjruwase Thank you for the response and I agree that the lossy overwrite will happen if making `dirs_exist_ok`.

In multi-rank training scenario, save_checkpoint function is called by multiple processes. Then, all processes will call [copytree line](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L3183), but the [saving logic](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L2832-L2843) for nvme offload seems all processes pointing to the same saving directory without the rank id of each process.

For my case, [the logging line](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L3178) shows
``` 
[2024-10-14 10:36:36,609] [INFO] [engine.py:3177:save_checkpoint] Copying NVMe offload files from /some/offload/dir/zero_stage_3/optimizer/rank0 to /some/model/directory/checkpoint-8/global_step8/offloaded_tensors, 5,629,499.53 GB free on target filesystem...
```
The `offload_dir` has rank id in its path while `offload_ckpt_dir` doesn't. All processes call copytree with dir_exist_ok=False to the same `offload_ckpt_dir`, and the FileExistsError occurs. This is my guess from the code and the log above.

tjruwase:
> All processes call copytree with `dirs_exist_ok=False` to the same `offload_ckpt_dir`, and the FileExistsError occurs. This is my guess from the code and the log above.

@w4-jihunlorenzopark, that looks like a bug. Are you able to provide a PR?

w4-jihunlorenzopark:
@tjruwase Sure, I will happily send the PR in few days. 

Before that, I would like to confirm the way how I will implement.
Would it be enough making the `offload_ckpt_dir` directory as follows, as the same as `offload_dir` structure?
I am thinking to insert rank id at the end of [here](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L2835) and [here](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L3175).

```
some/model/directory/checkpoint-8
│
├── offloaded_tensors
│   ├── rank0
│   ├── rank1
│   └── ...
```

During writing this, I am wondering whether `load_checkpoint` will be called by multiple processes in multi node training scenario, and if so, I got concerns that when loading checkpoint in another node with different number of ranks could be problematic.(Or would it be okay?) Maybe we can guide to users that to continue nvme offload training with checkpointed offloaded states, it should be on the same number of devices.

Any idea on this?

tjruwase:
@w4-jihunlorenzopark, your suggestion of including rank id looks good. 

Yes, this would only work if DP size is unchanged since ZeRO partitions checkpoint state across DP ranks. The following assertion should help to catch cases of DP changing:
https://github.com/microsoft/DeepSpeed/blob/6eefc3d0ead2c6360eec248daab0cae66a737ea0/deepspeed/runtime/engine.py#L3004-L3009

PR #4711, opened 2023-11-21
fix bias when calling qkv_func in DeepSpeedSelfAttention

zmzhang2000:
judge if `bias` should be `None` when calling `qkv_func` in `DeepSpeedSelfAttention`

If `self.config.set_empty_params` is `True`, `self.attn_qkvw` will be `None` and `self._merge_qkv()` is called.

The `self._attn_qkvb` will be an empty initialized tensor as `DeepSpeedSelfAttention._qkv_buffers[1]` is initialized by torch.empty() that may have `NaN`s.

This empty `self._attn_qkvb` could be input of `self.qkv_func()`, where `add_bias` is judged by `if bias is not None`.

However, `add_bias` should also be `True` if bias is an empty tensor if I understand right. Thus judgement should be made when calling `self.qkv_func`


zmzhang2000:
This bug leads to `NaN`s in the forward process of models with attention qkv bias, e.g., Llama2, when using zero stage 3.

PR #4713, opened 2023-11-21
BF16_Optimizer: add support for bf16 grad acc

nelyahu:
the default accumulation data type is fp32
by adding the below to deepspeed json file:
"data_types" : {"grad_accum_dtype": "bf16"}
gradient accumulation will be performed in bf16.


deepspeed/runtime/engine.py
```diff
@@ -1177,9 +1177,10 @@ def _do_optimizer_sanity_check(self, basic_optimizer):
         # data type checks
         elif model_dtype == grad_accum_dtype:
             if model_dtype == torch.bfloat16:
-                raise NotImplementedError(
-                    "Bfloat16 wrapper must use a gradient accumulation type of fp32, enable ZeRO to use Bfloat16 gradient accumulation"
```
tjruwase, line 1181: I am unsure this PR is needed. As this error message shows, the way to enable bf16 gradient accumulation is to set zero stage 1. Will that solution not work for you?

nelyahu: @tjruwase it seems like that DeepSpeedZeroOptimizer, is not functional with PipeEngine. I get the below failure:

`    return model.train_batch(data_iter=data_iterator)
  File "/home/nelyahu/qnpu/pt/src/deepspeed-fork/deepspeed/runtime/pipe/engine.py", line 391, in train_batch
    self._exec_schedule(sched)
  File "/home/nelyahu/qnpu/pt/src/deepspeed-fork/deepspeed/runtime/pipe/engine.py", line 1399, in _exec_schedule
    self._exec_instr(**cmd.kwargs)
  File "/home/nelyahu/qnpu/pt/src/deepspeed-fork/deepspeed/runtime/pipe/engine.py", line 803, in _exec_backward_pass
    self.optimizer.clear_lp_grads()
AttributeError: 'DeepSpeedZeroOptimizer' object has no attribute 'clear_lp_grads'`

I can try solving this out, but i have a feeling that DeepSpeedZeroOptimizer is not the native flow for pipeline users and more issues will be faced along the way.

tjruwase: @nelyahu, you are correct that bf16_optimizer is better tested for PipeEngine than DeepSpeedZeroOptimizer. In that case, can condition your change here so that the warning is triggered for PipeEngine while the original assert is triggered otherwise?

nelyahu: @tjruwase Done.

tjruwase (approved):

PR #4714, opened 2023-11-22
Fix confusing width in simd_load

yzhblind:
I found the width using in simd_load is different from simd_store.
This implementation confuses me.
The reason lies in the missing parentheses for the type conversion of x in the SIMD_LOAD2 macro definition, disrupting the intended semantics of width variable.
I try to make a quick fix for it.


ShadenSmith (approved):
Thank you!!! This is a great catch 🙂

PR #4721, opened 2023-11-23
support baichuan model:

baodii:
* fix Baichuan meta data error
   * add BaichuanLayer and DecoderLayer to glmtype when prepare tp fused qkvw
   * add get_alibi_mask function for Baichuan to enable TP


baodii:
@delock

delock:
Hi @RezaYazdaniAminabadi, this PR contains fixes for AutoTP to allow Baichuan model running with DeepSpeed AutoTP, can you review this PR for any comments?  Thanks!

deepspeed/module_inject/auto_tp_model_utils.py
```diff
@@ -60,6 +60,14 @@ def build_bloom_alibi_tensor(attention_mask: torch.Tensor, num_heads: int, dtype
     else:
         return alibi.reshape(batch_size * num_heads, 1, seq_length).to(dtype)
```
loadams, line 62: Formatting wise you should just need additional new lines above and below this function.

baodii: fixed.

deepspeed/inference/engine.py
```diff
@@ -220,6 +220,10 @@ def build_alibi_tensor(self):
             if hasattr(self.module.transformer, 'build_mpt_alibi_tensor'):
                 self.module.transformer.build_mpt_alibi_tensor_orig = self.module.transformer.build_mpt_alibi_tensor
                 self.module.transformer.__class__.build_mpt_alibi_tensor = build_mpt_alibi_tensor
+        if hasattr(self.module, 'model'):
```
loadams, line 223: @baodii - would it be possible to add a unit test for this model support?

loadams (approved):

deepspeed/inference/engine.py
baodii: > @baodii - would it be possible to add a unit test for this model support?

@delock is it necessary to add a single model's autoTP unit test?

delock: I think we should add autoTP tests with a seperate workflow in a seperate PR, there are several reasons:
1. Current autotp coverage in inference UT is not adequate (marian, codegen only), many popular model are not tested
2. Popular model has large disk/memory requirement, put them in a seperate workflow allows us to isolate the hardware requirements
3. It is suggested to test these autotp test with `deepspeed` launcher rather than `pytest` distributed launcher, this could avoid additional complexity and as close to user environment as possible.

I think we can open an issue for this task and explore in a seperate task.

PR #4722, opened 2023-11-23
params partition for skip_init

inkcherry:
Some models use ```skip_init``` to initialize weights. ```skip_init``` first initializes on a meta device in ```__init__``` of a module and then uses ```to_empty()```. This conflicts with the deepspeed hook ```module.__init__``` mechanism. it's necessary to wait for ```skip_init``` to finish before executing ```_post_init_method```. However, the ```from ... import skip_init``` behavior typically occurs outside the context, there seems to be no good way to directly hook into ```skip_init```. Hence, the approach here is to delay the execution of ```_post_init_method``` to resolve this issue. 
Known affected models include HuggingFace models like chatglm2 and chatglm3."


tohtana:
Hi @inkcherry, 
Just for clarification, it seems that `_post_init_method` runs after the `__init__` of the top-level module. Is this correct?
If so, can we reduce the peak memory footprint of parameters on initialization?

inkcherry:
> Hi @inkcherry, Just for clarification, it seems that `_post_init_method` runs after the `__init__` of the top-level module. Is this correct? If so, can we reduce the peak memory footprint of parameters on initialization?

Thanks for the review. @tohtana
It prioritizing the next non-meta module initialization completion(prioritizing a peer module below, and if None, its container module ).
I think this is ok because the module order of entering and exiting the post_init_module remains unchanged, but anyway. peak memory usage might be higher because module(which may contains several layers) of skip_init call need to be fully placed on one rank and then partition. To reduce memory(after child layer initialization is completed, partition them), I think some 
hook and restore is necessary during the skip_init lifespan.  I think I could try to maintain the code within a condition scope . What do you think.

tohtana:
Thank you for the clarification, @inkcherry.

I am wondering what happens if almost all modules are declared with `skip_init`. In this case, partitioning with zero.Init() won't work and we need the host memory of the size |all parameters| * |number of local GPUs (processes)| on a server, right?

I understand that it is difficult to set a hook in `skip_init`. But can we set one after `skip_init` as another approach?

inkcherry:
@tohtana ， Thanks for your suggestion!
Yes, if the device is not set to a GPU, the host memory behaves like this. 
I concern that adding logic after skip_init may also result in high memory usage. for example. if this module includes a ModuleList of all Transformer blocks, essentially containing almost all parameters of the model,  we also  need num_processes * all_transformer_blocks_parameters memory memory on host with to_empty() process.

currently, I have implemented a hook, without encountering such memory issues, and it provides better functionality. I decoupled it into a separate function. 

I tested full-parameters finetuning with chatglm2 6B zero3, without using skip_init and this patch, and compared it with using the default skip_init of the model with this patch. The loss is exactly the same for the first 50 steps.

inkcherry:
@tohtana  I just make some change ,could you please take a look, thanks.

inkcherry:
@tohtana I just make some change ,could you please take a look, thanks!

tohtana:
Thank you @inkcherry, the hook you implemented should work. This is an intricate and refined work!
It is good to me to merge after this change passes the tests.

tohtana (approved):

inkcherry:
@tohtana The CI has all passed. Just a reminder in case you missed it。
also, thanks for the internal  help of @delock

tohtana:
@inkcherry This PR was merged. Thank you for your great contribution!

PR #4739, opened 2023-11-28
Install `deepspeed-kernels` only on Linux

aphedges:
The package[^1] does not have sdists, and the only wheels it has are for Linux. The package therefore fails to install on any other platform. The solution to install the dev requirements on macOS is to skip the installation on incompatible OS's.

[^1]: https://pypi.org/project/deepspeed-kernels/0.0.1.dev1698255861/#files


mrwyattii (approved):

loadams (approved):

loadams:
Thanks @aphedges!

PR #4750, opened 2023-11-29
support autoTP with weight only quantization in DS inference path

ftian1:
This PR is used to make weight only quantization work with autoTP.

The sample code is like below:
```python
    model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.float16).to(device)

    ds_model = deepspeed.init_inference(model,
                                        mp_size=world_size,
                                        dtype=torch.float16,
                                        replace_with_kernel_inject=False) 

    model = ds_model.module
    from deepspeed.inference.quantization.quantization import _init_group_wise_weight_quantization
    ds_config = {
        "weight_quantization": {
            "post_init_quant": {
                '*': {
                    'num_bits': 4,
                    'group_size': 32,
                    'group_dim': 1,
                    'symmetric': False
                },
            }
        }
    }
    model = _init_group_wise_weight_quantization(model, ds_config)
```


delock:
@ftian1 if accelerator other than CUDA want to support AutoTP WOQ, which set of OpBuilder/kernels needs to be implemented?  Can you provide a link to kernel usage in the code?

delock:
It should be better to detect custom kernel existance by check attribute of the loaded ops, and call custom kernel accordingly, so any accelerator implement these kernels would be plugged.
> > @ftian1 if accelerator other than CUDA want to support AutoTP WOQ, which set of OpBuilder/kernels needs to be implemented? Can you provide a link to kernel usage in the code?
> 
> Here is the link https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/inference/quantization/utils.py#L115-L124, in which it dispatch to `deepspeed.ops.op_builder.QuantizerBuilder()` if device is `cuda` and group_size is able to be divided by 8. such implementation could be enhanced to support different h/w's opbuilder

tests/unit/inference/test_inference.py
```diff
@@ -459,6 +459,58 @@ def test(
...
+                }
+            }
+        }
+        from deepspeed.inference.quantization.quantization import _init_group_wise_weight_quantization
+        pipe.model = _init_group_wise_weight_quantization(pipe.model, ds_config)
+        pipe.device = torch.device(get_accelerator().device_name(local_rank))
+        ds_output = pipe(query, **inf_kwargs)
+
+        #print(local_rank, "baseline", bs_output)
+        print(local_rank, "deepspeed", ds_output)
```
baodii, line 510: Hi, @ftian1 I have run this test. But the result I got is 'deepspeed [{'generated_text': 'DeepSpeed is the greatest,,,,,,,,,,,,,,,'}]'. This result is not right. Can you figure out what's wrong with this test? BTW, I can pass all tests in test_intX_quantization.py.

ftian1: @baodii  may I know which device you are running on? cuda or cpu?

delock:
@ftian1 Is usage of WoQ with AutoTP similiar to with kernel injection?  Can you post a sample code show WoQ in DeepSpeed looks like withy kernel injection?

loadams:
@delock and @ftian1 - should we keep this PR active?  Or try to resolve the merge conflicts?

delock:
@loadams let me check with @ftian1 on this PR status.  Thanks for reminding!

ftian1:
@loadams  I have solved the merge conflicts. pls check it

loadams:
> @loadams I have solved the merge conflicts. pls check it

@ftian1 - I see that there are now new merge conflicts.  Could you take a look at these as well please?

PR #4766, opened 2023-12-04
allow multi-variable forward passes in Pipeline Module

bm-synth:
At the moment, `forward` passes in `PipelineModule` blocks only allow one input. This change allows the used to do forward passes without manually unfolding tuples of inputs. I.e.

Before
```
    def forward(self, input):
        (X, m) = input
        x, _ = self.block(x, m)
        return x, m
```
Now:
```
    def forward(self, x, m):
        x, _ = self.block(x, m)
        return x, m
```


bm-synth:
I'll abandon because passing variables as `layer(*args)` would break existing implementations. But having the possibility to return a `dict` that would then unfolded as input variables (`layer(**kwargs)`) would a good alternative that does not break existing implementations.

PR #4773, opened 2023-12-05
Loading FalconLinear to meta

oelayan7:
Instead of loading checkpoints to CPU,
load the weights through meta device


mrwyattii:
#4783 already merged, same change as requested here

PR #4782, opened 2023-12-07
improve the way to determine whether a variable is None

RUAN-ZX:
refactor: improve the way to decide whether a variable is None
fix: type mismatch for judging if current accelerator is in SUPPORTED_ACCELERATOR_LIST


RUAN-ZX:
@tjruwase, would you please review this pr ? I think it merely solve a bug of type mismatch and improve the way of `is None` judgement. Thank you :)

mrwyattii (approved):

Issue #4774, opened 2023-12-05
mpt model fails to run on cpu.


KepingYan:
**Describe the bug**
When I ran mpt model on the CPU, I encountered the following error.
![image](https://github.com/microsoft/DeepSpeed/assets/97155466/440d34f3-7881-4cba-aa8d-22d1c3207b9b)

**To Reproduce**
run-mpt-ds.py
```py
parser = argparse.ArgumentParser('generation script', add_help=False)
parser.add_argument("-m", "--model-id", type=str, default="EleutherAI/gpt-j-6B")
parser.add_argument("-t", "--tokenizer-id", type=str, default="EleutherAI/gpt-j-6B")
parser.add_argument('--device', type=str, default='cpu')
parser.add_argument('--max-new-tokens', default=32, type=int, help="output max new tokens")
parser.add_argument('--jit', action='store_true')
parser.add_argument('--local_rank', default=None, type=int, help="local rank")
parser.add_argument("--batch-size", default=1, type=int, help="batch size")
args = parser.parse_args()

device = torch.device(args.device)
if args.device == 'cpu':
    replace_with_kernel_inject = False
elif args.device == 'xpu':
    replace_with_kernel_inject = False
else:
    replace_with_kernel_inject = True
generate_kwargs = dict(do_sample=False, temperature=0.9, num_beams=4)
amp_enabled = True
amp_dtype = torch.bfloat16

# load model
model = AutoModelForCausalLM.from_pretrained(args.model_id, low_cpu_mem_usage=True,
                                             return_dict=not args.jit, torch_dtype=amp_dtype, trust_remote_code=True)
tokenizer = AutoTokenizer.from_pretrained(args.tokenizer_id, trust_remote_code=True)
model = model.eval().to(device)
model = model.to(memory_format=torch.channels_last)

# deepspeed engine
ds_local_rank = int(os.getenv('LOCAL_RANK', '0'))
ds_world_size = int(os.getenv('WORLD_SIZE', '0'))

if ds_world_size == 0:
    ds_world_size = 1
engine = deepspeed.init_inference(model=model, mp_size=ds_world_size, dtype=amp_dtype,
                                  replace_method="auto", replace_with_kernel_inject=replace_with_kernel_inject)
model = engine.module

prompt = "Once upon a time, there existed a little girl, who liked to have adventures. She wanted to go to places and meet new people, and have fun."
input_size = tokenizer(prompt, return_tensors="pt").input_ids.size(dim=1)
[...]

PR #4787, opened 2023-12-08
fix autoTP issue for mpt (trust_remote_code=True)

sywangyi:
to fix https://github.com/microsoft/DeepSpeed/issues/4774


sywangyi:
@delock

loadams (approved):

PR #2964, opened 2023-03-08
[DRAFT] Tentative implementation of MiCS

zarzen:
This PR drafts a tentative implementation of MiCS (https://arxiv.org/abs/2205.00119), which is first discussed in #2801 

For trying out the implementation, you can test it with a toy model at https://github.com/microsoft/DeepSpeed/blob/fdb8706a5f0b7564fe92d20cbeea460c2b569983/tests/small_model_debugging/test_mics_config.py 
 
And play with the shard configuration at here: https://github.com/microsoft/DeepSpeed/blob/fdb8706a5f0b7564fe92d20cbeea460c2b569983/tests/small_model_debugging/test_mics_config.py#L84 

Suggestions, comments are welcome, thanks!

PR #4792, opened 2023-12-09
Unit tests for MiCS

zarzen:
In response to the ask from https://github.com/microsoft/DeepSpeed/pull/2964#issuecomment-1832161865, I added three more unit tests related to MiCS. 

There are two knowledge issues:
- Testing on Torch 2.1.0 triggers `_IllegalWorker` in coalesced all gather. I made changes to ignore this condition. and Currently, I don't know the reason. 
- The MiCS implementation is not working with offloading, so the failure in `TestZeroPartialOffloadConfigSweep` is expected.


tjruwase:
@zarzen, thanks!

tests/unit/runtime/zero/test_zero.py
```diff
@@ -85,18 +85,25 @@ def test(self, zero_stage):
 
 
 # testing the fix https://github.com/microsoft/DeepSpeed/pull/1227
+@pytest.mark.parametrize("zero_stage", [3, "mics"])
```
mrwyattii, line 88:
```suggestion
@pytest.mark.parametrize("mics_enabled", [True, False])
```

tests/unit/runtime/zero/test_zero.py
```diff
@@ -85,18 +85,25 @@ def test(self, zero_stage):
...
 class TestZero3RepeatForwardLoop(DistributedTest):
     world_size = 1
 
-    def test(self, zero_stage=3):
+    def test(self, zero_stage):
```
mrwyattii, line 92:
```suggestion
    def test(self, mics_enabled, zero_stage=3):
```

tests/unit/runtime/zero/test_zero.py
```diff
@@ -85,18 +85,25 @@ def test(self, zero_stage):
...
         # force all params to be partitioned by forcing threshold=0
+        mics_shard_size = -1
+        if isinstance(zero_stage, str):
+            zero_stage = 3
+            mics_shard_size = self.world_size
+
```
mrwyattii, line 98:
```suggestion
        mics_shard_size = -1
        if mics_enabled:
            mics_shard_size = self.world_size
```

mrwyattii (changes requested):
I think it will be more clear if we parameterize a `mics_enabled` variable rather than use `zero_stage` to switch it on/off. Please update the other test with similar changes. Thanks!

zarzen:
@mrwyattii thanks for suggestions, I updated the implementation accordingly. Besides, as the mics implementation currently is not compatible with offloading, I removed the unittest for `test_zero_offloadpp.py`

loadams (approved):

mrwyattii (approved):

PR #4805, opened 2023-12-13
Switch paths-ignore to single quotes, update paths-ignore on nv-pre-compile-ops

loadams:


mrwyattii (approved):

.github/workflows/nv-pre-compile-ops.yml
```diff
@@ -7,8 +7,6 @@ on:
     paths-ignore:
       - 'docs/**'
       - 'blogs/**'
-      - 'deepspeed/inference/v2/**'
-      - "tests/unit/inference/v2/**"
```
mrwyattii, line 11: I think we want to keep these since the pre-compile-ops do not test anything related to inference v2

mrwyattii (changes requested):

.github/workflows/nv-pre-compile-ops.yml
loadams: Yep, forgot to add those back in, should be fixed now.

mrwyattii (approved):

PR #4817, opened 2023-12-14
Update imports from Transformers

loadams:
This helps resolve the following warnings:

``` 
  /tmp/actions-runner/_work/DeepSpeed/DeepSpeed/unit-test-venv/lib/python3.8/site-packages/transformers/deepspeed.py:23: FutureWarning: transformers.deepspeed module is deprecated and will be removed in a future version. Please import deepspeed modules directly from transformers.integrations
    warnings.warn(
```

### DIFF
--- a/requirements/requirements-dev.txt
+++ b/requirements/requirements-dev.txt
@@ -16,5 +16,5 @@ sphinx
 sphinx-rtd-theme
 tensorboard
 torchvision
-transformers
+transformers>=4.32.1
 wandb
--- a/requirements/requirements-inf.txt
+++ b/requirements/requirements-inf.txt
@@ -1,5 +1,5 @@
 google
 lm-eval==0.3.0
 protobuf
-transformers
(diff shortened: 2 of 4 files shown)


mrwyattii (approved):

mrwyattii:
@loadams I think we should also set a minimum `transformers` version in requirement files that have `transformers` listed (e.g., https://github.com/microsoft/DeepSpeed/blob/master/requirements/requirements-dev.txt). It looks like v4.32.1 is when the import was moved to `transformers.integrations`.

loadams:
> @loadams I think we should also set a minimum `transformers` version in requirement files that have `transformers` listed (e.g., https://github.com/microsoft/DeepSpeed/blob/master/requirements/requirements-dev.txt). It looks like v4.32.1 is when the import was moved to `transformers.integrations`.

Makes sense, I hadn't looked up the version number, but that sounds good, should be updated now.

PR #4828, opened 2023-12-16
Mixtral FastGen Support

cmikeh2:
Adds support for Mixtral with FastGen. Key features implemented:

1. Top-2 MoE support
2. Better support for RoPE thetas
3. The mistral model implementation


RezaYazdaniAminabadi:
this is amazing, i was also working on this, but i think most of what i have you already added in this pr. thanks @cmikeh2 :)

deepspeed/inference/v2/kernels/ragged_ops/moe_scatter/moe_scatter.cu
```diff
@@ -15,10 +15,11 @@ namespace scatter {
 constexpr int access_granularity = 16;
 constexpr int threads = 256;
 constexpr int warps = threads / hw_warp_size;
+constexpr int max_experts = 1024;
 
 }  // namespace scatter
 
-template <typename T, int copyUnroll>
+template <typename T, int copyUnroll, int N_TOP_K>
```
RezaYazdaniAminabadi, line 22: @cmikeh2, I know you like to generalize this function, but I was wondering if we can have two kernels here, one for top-1 and one for top-k, just so that we can remove some of the complexity added for top-1. what do u think?

cmikeh2: Are you observing any slowdown with top-1? The re-org intention here was primarily to simplify things. Previously, on block-0 we did a cumsum for the GEMM kernel and the rest of the thread blocks did max reductions. The max reduction is of similar complexity to the cumsum anyways (`log(n)`) steps and since it's necessary on all blocks for the top-N case anyways, I thought it made sense to remove the branch from the code and have a unified path.

mrwyattii (approved):
Thanks @cmikeh2!

deepspeed/inference/v2/kernels/ragged_ops/moe_scatter/moe_scatter.cu
RezaYazdaniAminabadi: i haven't but i will try to do some profiling of this in the next days. thanks for the clarification on the changes :)

deepspeed/inference/v2/kernels/ragged_ops/linear_blocked_kv_rotary/blocked_kv_rotary.cpp
```diff
@@ -51,6 +52,8 @@ void kv_trained_rotary_embeddings(torch::Tensor& kv_cache,
     TORCH_CHECK(n_tokens == k.size(0));
     TORCH_CHECK(n_tokens == v.size(0));
 
+    const float theta_base = 0.f;
```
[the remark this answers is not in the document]
cmikeh2: Yes, on this code path the theta base is irrelevant since we're reading the buffer from the model that would be derived from it. The only reason we declare it here is compatibility with the macro/single interface for the pair of rotary kernels. As the interfaces have diverged more I have wondered if it makes more sense to just split the interface entirely.

Issue #4567, opened 2023-10-26
[Feature package] Full feature support with Ascend NPU


hipudding:
## Background

**Ascend** is a full-stack AI computing infrastructure for industry applications and services based on Huawei Ascend processors and software. For more information about Ascend, see [Ascend Community](https://www.hiascend.com/en/).

**[CANN](https://www.hiascend.com/en/software/cann)** (Compute Architecture of Neural Networks), developped by Huawei, is a heterogeneous computing architecture for AI.

**Pytorch** has officially announced [support for Ascend NPU](https://pytorch.org/blog/pytorch-2-1/) (through key PrivateUse1), please see the PrivateUse1 tutorial [here](https://pytorch.org/tutorials/advanced/privateuseone.html).

## Previous work

NPU accelerator support has already been merged (see #3595, #3831), which makes it possible to use NPU as a backend accelerator for basic training and inferencing tasks. However, to achieve full support, more features need to be implemented.

## Sub tasks

Here is a list of features that need to be implemented or tested.

status|title|assigned to
--|--|--
Done|[Unit tests](https://github.com/microsoft/DeepSpeed/pull/4569)|@RUAN-ZX
Done|FP16|@minchao-sun, @wuhhu
Done|BF16|@minchao-sun, @wuhhu
Done|Gradient Accumulation|@minchao-sun, @wuhhu
Done|Data Parallelism|@minchao-sun, @wuhhu
Done|Pipeline Parallelism|@RUAN-ZX
Done|Zero1|@misstek
Done|Zero2|@misstek
Done|Zero3|@misstek
Done|Activation Checkpointing|@CurryRice233
Done|[Fused Adam](https://github.com/microsoft/DeepSpeed/pull/4343)|@CurryRice233
Done|Mixture of Experts (MoE)|@wangshuai09
Done|[RLHF](https://github.com/microsoft/DeepSpeed/pull/4343)|@wangshuai09 @CurryRice233
Done|[ZeRO Offload](https://github.com/microsoft/DeepSpeed/pull/4568)|@hipudding @CurryRice233
Done|[ZeRO Infinity](https://github.com/microsoft/DeepSpeed/pull/4809)|@misstek
Done|[1-bit Adam](https://github.com/microsoft/DeepSpeed/pull/4733)|@RUAN-ZX
Done|[1-bit LAMB](https://github.com/microsoft/DeepSpeed/pull/4733)|@RUAN-ZX
[...]

PR #4831, opened 2023-12-18
[NPU] Add NPU to support hybrid engine

CurryRice233:
Add NPU to support the hybrid engine, we tested it with the example on [DeepSpeedExample](https://github.com/microsoft/DeepSpeedExamples/tree/master/applications/DeepSpeed-Chat/training/step3_rlhf_finetuning) and the results, average reward scores, were aligned with the GPU.

For works about NPU, see https://github.com/microsoft/DeepSpeed/issues/4567


CurryRice233:
@mrwyattii hi, Could you please help to retrigger the workflow and review this PR ？

CurryRice233:
@tjruwase @mrwyattii hi, would you please help to review this PR and add to merge queue? CI has passed.

tjruwase:
@CurryRice233, thanks for this PR. Apologies for the delay. Most of the team are just returning from the holidays. We will review asap.

CurryRice233:
@tjruwase The CI failed. It seems that there is a problem with the CI server. Can you check it?

CurryRice233:
@tjruwase @loadams CI passed again, can you help to review this PR and add to merge queue?

tjruwase:
> @tjruwase @loadams CI passed again, can you help to review this PR and add to merge queue?

@CurryRice233, we will review this asap!

op_builder/npu/inference.py
```diff
@@ -0,0 +1,307 @@
...
+# DeepSpeed Team
+
+from enum import IntEnum
+import torch
+from .builder import NPUOpBuilder
+
+try:
+    import torch_npu
+except ImportError as e:
+    pass
```
mrwyattii, line 13: @CurryRice233 importing torch here causes problems for DeepSpeed install. Can you please move the `torch` and `torch_npu` imports inside of a class method in this file? (See https://github.com/microsoft/DeepSpeed/blob/9500ab7d476aaf60ec2c76653c67bf5feb758f9e/op_builder/transformer_inference.py#L20 for an example)

The reason for this is that `op_builder` is imported by `setup.py` (before any requirements are installed) and if `torch` is not available the install will fail. Please let me know if you have any questions

mrwyattii (changes requested):

op_builder/npu/inference.py
CurryRice233: Moving the `import torch' to the interior of the `try' solves this problem. If I move inside of class method, I need to import torch into each method, which I don't think is a good idea.

mrwyattii (approved):

CurryRice233:
@mrwyattii  Hi, can you add this PR to merge queue, thank you !

PR #4833, opened 2023-12-18
DeepSpeedZeroOptimizer: refactor bit16 flattening to support more accelerators

nelyahu:
The approach till today use the practice where the torch.nn.parameter data is being replaced with a new cpu data storage, to offload device memory.
All params are being flatenned on the host and moved to the device.
in some accelerators torch.nn.parameter which is a device parameter cannot be assigned with a cpu storage.
This PR copy the param data into a new cpu tensor, and shrinks the device storage. 
Later when the flat buffer is moved to the device param.data will be a view to the flat buffer.


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -316,7 +311,13 @@ def __init__(self,
 
             see_memory_usage(f"Before moving param group {i} to CPU")
             # move all the parameters to cpu to free up GPU space for creating flat buffer
-            move_to_cpu(self.bit16_groups[i])
+
+            # Create temp CPU param copies and free HPU tensors
```
tjruwase, line 315:
```suggestion
            # Create temp CPU param copies and free accelerator tensors
```

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -334,18 +335,29 @@ def __init__(self,
...
+
+            # free temp CPU params
+            for param in self.bit16_groups[i]:
+                del param.cpu_data
+
+            self.bit16_groups_flat.append(flattened_buffer.to(get_accelerator().current_device_name()))
+            del flattened_buffer
+
+            # recover temp CPU params
+            for param in self.bit16_groups[i]:
```
tjruwase, line 352: Since the params are already in flattened buffer on the device, I don't understand the need to recover or use `param.cpu_data` beyond this point. I think the `numel()` values in line 360 for padding computation could be derived earlier in line 318 by something like `param.orig_numel = param.numel()`

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -354,7 +366,11 @@ def __init__(self,
                 see_memory_usage(f"After Flattening and after emptying param group {i} cache", force=False)
 
             # set model bit16 weight to slices of flattened buffer
-            self._update_model_bit16_weights(i)
+            self._update_model_bit16_weights(i, use_cpu_data=True)
```
tjruwase, line 369: Since the weight slices are already in a flattened buffer on device, I don't understand the `use_cpu_data` here.

deepspeed/runtime/zero/stage_1_and_2.py
nelyahu: Good point, the cpu_data is required for _update_model_bit16_weights .
which calls _unflatten()_ - This function accept a flat buffer and a list of tensors, where the only use of this list is just to get a corresponding list of shapes, so the flat buffer will be unflattened according to.
At this point the each tensor in _self.round_robin_bit16_groups[group_index]_ has empty storage, only the cpu_data hold the actual shape. 
That is why we had to add _use_cpu_data_ (will be set to True only during initialization), so it will construct the tensor list according to param.cpu_data and not according to param.data.

as a possible solution i can suggest to construct a parallel list of tensors (per group) that will be placed on meta device. then we can get rid of the _orig_shape_.
self.meta_group_tensors = []
....
meta_tensor = torch.tensor(param_shape, dtype=param_dtype, device=meta)
...
self.meta_group_tensors.append(meta_tensor).

and in __update_model_bit16_weights_ we will always use it as an input for _unflatten()_

if you have something else in mind, let me know, i am open for suggestions.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -589,9 +605,12 @@ def _configure_moe_settings(self):
         assert self.expert_dp_process_group is not None, "Expert data parallel group should be configured with MoE"
         assert self.ep_process_group is not None, "Expert parallel group should be configured with MoE"
 
-    def _update_model_bit16_weights(self, group_index):
-        updated_params = self.unflatten(self.bit16_groups_flat[group_index],
-                                        self.round_robin_bit16_groups[group_index])
+    def _update_model_bit16_weights(self, group_index, use_cpu_data=False):
+        if use_cpu_data:
```
tjruwase, line 609: It seems that `use_cpu_data==True` means to unflatten from flattened device buffer into cpu buffers, is that correct? 
I am concerned that unflattening into cpu buffers will break line 615, perhaps only for non-hpu devices. This is because `p.data` will be updated from device buffer to cpu buffer (`q.data`). Can you please clarify?

deepspeed/runtime/zero/stage_1_and_2.py
nelyahu: see my comment above

deepspeed/runtime/zero/stage_1_and_2.py
nelyahu: No. see my (long) comment above. it uses only the shapes from the tensors list to know how to unflatten.

deepspeed/runtime/zero/stage_1_and_2.py
tjruwase: Yes, I prefer the idea of meta device since it was created for this kind of situation. Ideally, I would love to modernize the ZeRO code by incorporating more meta device usage, so this is a great start :)

deepspeed/runtime/zero/stage_1_and_2.py
tjruwase: You can ignore my question here. Your previous response already clarified my confusion. Thanks!

deepspeed/runtime/zero/stage_1_and_2.py
nelyahu: Sure, i will have to test it a little bit locally before submitting the change in this PR.

nelyahu: @tjruwase - see latest fixes

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -334,18 +336,31 @@ def __init__(self,
...
+            self.bit16_groups_flat.append(flattened_buffer.to(get_accelerator().current_device_name()))
+            del flattened_buffer
+
             see_memory_usage(f"After flattening and moving param group {i} to GPU", force=False)
 
             # Record padding required for alignment
             if partition_id == dist.get_world_size(group=self.real_dp_process_group[i]) - 1:
-                padding = self.bit16_groups_flat[i].numel() - sum(
-                    [t.numel() for t in self.round_robin_bit16_groups[i]])
+                padding = self.bit16_groups_flat[i].numel() - orig_group_numel
```
tjruwase, line 363: This line looks suspicious since `orig_group_numel` is hardcoded to zero.

nelyahu: Hi @tjruwase , you are correct. it was somehow dropped in my merge conflict resolution. fixed it now.

tjruwase (approved):

nelyahu:
Hi @tjruwase - i fixed a failing UT issue, can you please re-trigger workflow?

PR #4843, opened 2023-12-20
deepspeed engine: fp16 support validation on init

nelyahu:


deepspeed/runtime/engine.py
```diff
@@ -242,6 +242,9 @@ def __init__(
                     assert not self.elasticity_enabled(), ("Elasticity is not currently supported"
                                                            " with model parallelism.")
 
+        if self.fp16_enabled() and not get_accelerator().is_fp16_supported():
```
tjruwase, line 245: Can you please move this into `_do_sanity_check()`?

nelyahu: @tjruwase Done

tjruwase (approved):

PR #4849, opened 2023-12-20
Add Cache to Comm Group

cmikeh2:
This adds a global cache for creating new comm groups. Rather than returning unique objects, an identical group (same backend, same ranks) will share a single object. The motivation for including a cache like this is each creation of a group can reserve a small amount of memory that is not reclaimed when the object itself is freed. This creates a memory leak for a long-running application like MII where models may be destroyed and created many times.

Areas for feedback:
1. Does this have any side effect for training workloads? Are there any assumptions that the group objects need to be unique?
2. Does this make sense to have as an optional opt-in behavior or built with an additional wrapper, rather integrating it directly into `new_group`

### DIFF
--- a/deepspeed/comm/comm.py
+++ b/deepspeed/comm/comm.py
@@ -52,6 +52,9 @@ mpi_backend = None
 ccl_backend = None
 hccl_backend = None
 
+# Cache of previously created comms
+comms_cache = {}
+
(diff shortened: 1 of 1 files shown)


GuanhuaWang:
my major concern is this global and reuseable cache may not be correctly deleted when communication ends.

PR #4859, opened 2023-12-21
Pipeline: Add support to eval micro bs configuration

nelyahu:
When running evaluation the general memory consumption is reduced. Mainly due to absence of gradients, and hanging FWD activations. It allows to increase the micro-bs and improve the evaluation performance. This commits add the option to pass num_micro_batches to eval_batch(), as the current assumption is that same micro-bs and global-bs is used, so same number micro batches will take place.
This commit also modifies _scale_loss_by_gas in runtime/engine.py to consider number of eval micro batches for loss scaling instead of training gas.

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -402,7 +402,13 @@ class PipelineEngine(DeepSpeedEngine):
         # TODO: should return precisely what loss returned and allow others to be queried?
         return self.agg_train_loss
 
-    def eval_batch(self, data_iter, return_logits=False, compute_loss=True, reduce_output='avg', bcast_loss=True):
+    def eval_batch(self,
+                   data_iter,
+                   return_logits=False,
+                   compute_loss=True,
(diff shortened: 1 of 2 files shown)


deepspeed/runtime/pipe/engine.py
```diff
@@ -451,8 +457,11 @@ def eval_batch(self, data_iter, return_logits=False, compute_loss=True, reduce_o
         train_iterator = self.data_iterator
         self.set_dataiterator(data_iter)
 
+        # set the number micro batches incase the user chose value than training
```
loadams, line 460: The formatting check is failing on codespell [here](https://github.com/microsoft/DeepSpeed/actions/runs/7391638608/job/20108766055?pr=4859#step:5:71), "incase" -> "in case"

nelyahu: Thanks, fixed it :)

loadams (approved):

ShadenSmith (approved):
Thank you!!

PR #4865, opened 2023-12-24
Fix f-string messages

li-plus:
Fix error messages that missed the f prefix.


loadams (approved):

loadams:
Thanks for your contribution @li-plus!

PR #4878, opened 2023-12-28
Enable torch.compile with ZeRO (Experimental)

tohtana:
This PR enables `torch.compile` with ZeRO stages 1/2/3. You need to add `compile` section in your DeepSpeed config. The fields in the section are passed to `torch.compile`.

```json
  "compile": {
    "disable": false,
    "backend": "inductor"
  }
```

To enable a custom backend, you can pass the fully qualified name of the backend function. For example, if you have a backend class `my_backend` in `my_backend.py` in the current directory, you can enable it by `"backend": "my_backend.my_backend"`. You can find an example in [a unit test](https://github.com/microsoft/DeepSpeed/blob/eb9d4e06e9596f391aea305a6a5c6ec70cc28b58/tests/unit/runtime/compile/test_config.py#L116).

Currently we validated the results with Megatron-DeepSpeed. See the [example](https://github.com/microsoft/Megatron-DeepSpeed/tree/tohtana/enable_compile/examples_deepspeed/compile) for the details.

NOTICE: This PR is a draft. We will need to validate the coverage and accuracy with many more examples.


tjruwase:
@stas00, FYI

stas00:
Amazing work, @tohtana! I'm looking forward to trying it out

Here is a quick feedback:

Could we please flip `disable` to `enabled` so that the logic is consistent with other config values?
- no double negation logic
- consistent `enabled` (and not `enable`) - as all other config sections use that name.

stas00:
tried it out and the compiled engine doesn't seem to forward custom methods to the unwrapped model, e.g. it's failing:

```
[28:7]:  File "/data/env/lib/repos/retro-llama/tr043-dawn-llama-3/DeepSpeed/deepspeed/runtime/engine.py", line 468, in __getattr__
[28:7]:    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
[28:7]:AttributeError: 'DeepSpeedEngine' object has no attribute 'get_model_tflops_per_batch_per_gpu'
``` 

`get_model_tflops_per_batch_per_gpu` is a normal model's attribute and the same setup works if I set `"disable": true` for the compile section.

This method is just part of the normal model.

stas00:
I hacked around it via `model.module.method...` and then I get many warnings and errors with the `inductor` backend and then it fails. I have attached the log. This is just training Llama-2 on a single node using Accelerate with torch-nightly from last night.

```
[ds-compile.txt](https://github.com/microsoft/DeepSpeed/files/13815566/ds-compile.txt)

```

stas00:
If I disable the ds profiler than it runs despite the compilation errors/warnings - same log as in the previous comment, other than the last traceback where it crashes.

stas00:
I'm also observing a very strange behavior of performance cycling:

the tflops go like this per iteration: 196, 196, 192, 196, 196, 192, 196, 196, 192, - 2 fast one slower - very exactly

w/o compile it was a consistent 194.

tohtana:
@stas00 Thank you for your feedback! This PR is still experimental. Let me address the issues one by one.

The configuration disable is what I specifically sought feedback on. Currently, all configuration items under compile are passed to torch.compile, which accepts disable, not enable. This design was chosen for its simplicity, given the uncertainty of future changes in torch.compile. But we can define `enable` and flip it before passing it to `torch.compile.` 

Do you have any further comments on this? If not, I will switch it to enable as you suggested. Actually, it is also my personal preference.

stas00:
That's totally understandable, Masahiro. Tunji made that clear when he tagged me. If it's too early to provide feedback please ping me when you're ready for it.

----------------

`disable` vs `enabled`:

Ideally, Deepspeed users will never need to know anything about `torch.compile` specifics - many frameworks integrate this feature w/o having the user needing to interact with it directly. So its API doesn't have to impact Deepspeed's API.

Since most (all?) Deepspeed config sections use `enabled` I'd say it'd be the most consistent to continue with that convention. 

But this is an option of a single person, so please seek out opinions of others.

tohtana:
@stas00 Thank you for your quick reply. Probably it is difficult to have a clear conclusion for now. I will simply switch it to `enable`. Otherwise, many other users would have the same question as yours. 
For a clearer answer, we need more experience to know what options DeepSpeed's users need in their applications. Even the options of `torch.compile` may change.

stas00:
- please note that it's `enabled` and not `enable` that DS uses everywhere else.

- re: what other options I'd say - use the minimal amount of options - 

1. let's perhaps start with only `backend` and then pick the most sensible defaults for that option.

2. Then provide a user an API where they can preset their own `**torch_compile_kwargs` that will be passed to `torch.compile` - that way you're future proofing the Deepspeed API while allowing torch to do what they please - deepspeed will sync with the future changes to keep up with the sensible default and power-users should always be able to override the defaults. 

2a. I don't know if the current config file allows for a not predefined dict, so perhaps this could be possible:

```
  "compile": {
    "enabled": true,
    "backend": "inductor"
    "kwargs": "key1"=value, "key2"=value
  }
```
this should definitely work:

```
  "compile": {
    "enabled": true,
    "backend": "inductor"
    "kwargs": "key1=value;key2=value"
  }
```
but I don't know if all `torch.compile` kwargs could be stringified

but providing a programmatical API for power users would be the most fool-proof:

```
deepspeed_engine.set_torch_compile_kwargs(**kwargs)
```

tohtana:
@stas00, (FYI @tjruwase)
Sorry for the delayed response. I have addressed some of the issues you suggested.

- Calling custom functions
- Configuration format in a deepspeed config (You can pass `"kwargs": {"key1"=value, "key2"=value}`)
- API to set compiler options (`torch_compile_kwargs`)

I also added unit tests that verify these features. I am wondering if we should merge this PR after the API design is finalized. I know that it still has a lot of limitations and currently works only for limited model architectures, but we will need to work on the improvements incrementally.

stas00:
@tohtana, thank you for implementing my suggestions.

I haven't tested the code but looking at your tests this looks good.

I agree that doing this work incrementally is a good idea.

May I recommend to add a note in the API and config docs that this is an experimental API and a subject to change on a moment's notice - so that you're not yet committing to this new API. Let users kick the tires, send feedback and if it feels good then you could commit to that API - here the API is the config.

wconstab:
as far as API goes, i think it might be better practice to let users apply the torch.compile macro themselves on the model, and have DeepSpeed interact with that.  I suggest this for two reasons.
1) tunneling args for torch.compile through something else is a little tricky.  (another way to handle this might be to let users give you a partial where they already applied their args, and you just store their 'compiler_fn' and call it 
2) i haven't looked at how you use compile yet, but if you're compiling specific parts of models it might be nice to allow us to eventually compile larger sections after improving torch.compile.

Others may disagree with my suggestions - cc @ezyang @yf223 @wanchaol

tohtana:
Thank you @wconstab for your comment!

The simplest approach would be to allow users to run torch.compile and then pass the compiled model to DeepSpeed. However, this is not feasible because ZeRO3 sets hooks on each layer to gather parameters during initialization.

> tunneling args for torch.compile through something else is a little tricky. (another way to handle this might be to let users give you a partial where they already applied their args, and you just store their 'compiler_fn' and call it

The current configuration was designed to integrate the settings into our "DeepSpeed config". The DeepSpeed engine uses it on initialization. The config can be a dictionary, but all items must be serializable since it is often passed as a (path to) JSON. Therefore, I implemented this as the standard procedure for compile configuration in DeepSpeed.
On the other hand, I believe we can still enable users to pass compiler_fn through a new API as another choice as you suggested.

> i haven't looked at how you use compile yet, but if you're compiling specific parts of models it might be nice to allow us to eventually compile larger sections after improving torch.compile.

We currently compile the entire model provided to the DeepSpeed engine. We avoid compiling some hooks because certain communications are not yet compatible with torch.compile.
I expected users would apply torch.compile.disable to a module or a function within their model when they need.

wconstab:
> ZeRO3 sets hooks on each layer to gather parameters during initialization.

can you say which type of hooks?  we support some already, maybe not the ones you need though.

wconstab:
ok, that makes sense.  The alternate approach (maybe for later on) is that you could try to make the contents of the hook 'safe to compile' such that the torch.compile'd graph already contains the release and materialize logic in it.  But that is another level of work/design/issues.

tohtana:
@wconstab Thank you for your comment.

Regarding your last comment, currently the zero3 hooks are not safe to compile. I have added `torch.compiler.disable` to some functions to address this, but I believe we will be able to resolve it in the future. As far as I remember, I encountered an error regarding communication collectives and synchronization using CUDA events (I'm not sure about the latter). We should be able to solve the first issue by replacing the communication collective with functional ones. We can also tweak the synchronization to make it compatible with `torch.compile`.

As for your first comment, I will allow users to pass `compiler_fn` as the most flexible option. However, I will still need to maintain the current approach that *tunnels* compiler options due to the reasons I mentioned above. This is more about compatibility with DeepSpeed than `torch.compile`.

tohtana:
FYI @wconstab @tjruwase

I addressed @wconstab's comment regarding `compiler_fn` (CI seems to have some issues right now)

deepspeed/runtime/compiler.py
```diff
@@ -0,0 +1,173 @@
...
+    def __init__(self, module, compile_config: Union[CompileConfig, None] = None):
+        super().__init__()
+
+        assert is_compile_supported(), "torch.compile is not supported on this version of PyTorch."
+
+        modules = self.__dict__.get('_modules')
+        modules['wrapped'] = module
+        self.__dict__['wrapped'] = module
+        self._is_compiled = False
+        self._backend = compile_config.backend
```
tjruwase, line 98: Should it not be the following?

```suggestion
        self._backend = get_backend_fn(compile_config.backend)
```

deepspeed/runtime/compiler.py
```diff
@@ -0,0 +1,173 @@
...
+
+def get_compile_config(param_dict):
+    if COMPILE_CONFIG in param_dict:
+        compile_config_dict = param_dict[COMPILE_CONFIG]
+    else:
+        compile_config_dict = {}
+    return CompileConfig(**compile_config_dict)
+
+
+def get_backend_fn(backend_name: str):
```
tjruwase, line 33: I think this function should handle all the ways of specifying backends, including `backend_name="inductor"`

deepspeed/runtime/compiler.py
```diff
@@ -0,0 +1,173 @@
...
+        if isinstance(backend, str):
+            self._backend = get_backend_fn(backend)
+        elif callable(backend):
+            self._backend = backend
+        else:
+            raise ValueError(f"backend for torch.compile must be a string or Callable: {backend}")
+
+    def set_torch_compile_kwargs(self, kwargs: Dict[str, Union[str, Any]]) -> None:
+        """Set kwargs for torch.compile. Kwargs that are set in DeepSpeed config will be overwritten.
+        You can also pass a backend name with "backend" key to change the backend.
```
tjruwase, line 122: I think it is confusing to allow users to change backend through `set_torch_compile_kwargs`, especially since we `set_backend` already exists for that purpose.

deepspeed/runtime/compiler.py
tohtana: I think get_backend_fn runs in `validate_backend` of `CompileConfig`

deepspeed/runtime/engine.py
```diff
@@ -467,6 +469,13 @@ def __getattr__(self, name):
             return getattr(self, name)
         elif name in dir(_module):
             return getattr(_module, name)
+        elif isinstance(_module, CompiledModuleWrapper):
```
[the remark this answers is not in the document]
tohtana: This condition is necessary to make custom functions of the wrapped module work.

Assume we have the compiled module `cm`, its wrapper `w` (`CompiledModuleWrapper`), and the deepspeed engine `engine`. We want to call a custom function `f` of `cm` by `engine.f()`. However, `dir(_module)` doesn't return `f`. So `engine.f()` would fail without this condition though `w.f()` works.

I'm not sure this is the best design for the purpose. I would appreciate it if you could give my any suggestion.

deepspeed/runtime/compiler.py
tohtana: I see, you are right. We already have `set_backend`. Let me fix it.

deepspeed/runtime/compiler.py
```diff
@@ -0,0 +1,173 @@
...
+from typing import Union, Callable, Dict, Any
+import importlib
+import torch
+from ..pydantic_v1 import validator
+from .config_utils import DeepSpeedConfigModel
+
+COMPILE_CONFIG = "compile"
+
+
+def is_compile_supported():
```
tjruwase, line 15: I think this should be lifted to engine level so that usage is `deepspeed.is_compile_supported()`.

tjruwase:
@tohtana, thanks for this amazing PR. Can you please include some tests of cpu and nvme offload?

deepspeed/runtime/compiler.py
tohtana: I agree, it would be good to consolidate the initialization and error checking of `backend` to `get_backend_fn`. Let me fix as you suggested.

deepspeed/runtime/compiler.py
tohtana: We agreed on that we shouldn't overwrite `compile_config.backend` at the validator after the offline discussion. Then I removed the validator and add `get_backend_fn` to the constructor of `CompiledModuleWrapper` as suggested.

deepspeed/runtime/engine.py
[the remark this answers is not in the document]
tjruwase: Will `hasattr(_module, name)` work for both cases of compiled and not-compiled module? In other words, for non-compiled module, can we use `hasattr(_module, name)` instead of `name in dir(_module)`?

tjruwase (approved):

mrwyattii (approved):

deepspeed/runtime/compiler.py
```diff
@@ -0,0 +1,166 @@
...
+        compile_config_dict = {}
+    return CompileConfig(**compile_config_dict)
+
+
+def get_backend_fn(backend: Union[str, Callable]) -> Union[str, Callable]:
+    if isinstance(backend, Callable):
+        return backend
+
+    elif isinstance(backend, str):
+        if backend in torch._dynamo.list_backends():
```
oraluben, line 38: @tohtana The default `list_backends` call will exclude debug and experimental backends, e.g. `eager`. I think it's better to use `list_backends(exclude_tags=())` here.

tohtana: Thank you for the comment. I opened #5191.

Issue #2295, opened 2022-09-06
Question: why are overlap and contiguous grads meaningless in stage 1 and are ignored


woolpeeker:
https://github.com/microsoft/DeepSpeed/blob/80f94c10c552ec79473775adb8902b210656ed76/deepspeed/runtime/engine.py#L1384

I wonder why we cannot use overlap_comm in zero1 to reduce more latency?
Appreciate any reply.

tjruwase:
@woolpeeker, thanks for the question. This is just an implementation restriction but not a fundamental one.

ShenglongZ:
> @woolpeeker, thanks for the question. This is just an implementation restriction but not a fundamental one.
![image](https://user-images.githubusercontent.com/33069796/223333152-02bc460c-988e-4371-911c-648a2a053e44.png)

I noticed that in the DeepspeedEngine, overlap_comm is forced to be False when partitioning optimizer_states. However, if I set it True, there are no error occurred during training. By profiling with nsight sys, allreduce kernel does overlap with computational kernels as shown. 

I's wondering what is the specific restriction? Is there any walk around? 
![image](https://user-images.githubusercontent.com/33069796/223333753-17b4c419-1f3b-44f3-bd88-c77158febe71.png)

mayank31398:
@tjruwase are there any plans to support overlap for ZeRO-1?
It would be nice if this support is present.

li-plus:
I removed the restrictions of `overlap_comm` and `contiguous_gradients` in ZeRO1 in #4887. Everything works fine for me. Didn't get reviewed yet.

PR #4887, opened 2023-12-30
Release overlap_comm & contiguous_gradients restrictions for ZeRO 1

li-plus:
The `overlap_comm` and `contiguous_gradients` options have been ignored in ZeRO stage 1 since https://github.com/microsoft/DeepSpeed/pull/1246. Back in that time, ZeRO 1 and 2 are separately implemented (see https://github.com/microsoft/DeepSpeed/tree/6ae756c03f12674f17aef90622e7664a8af9d2af/deepspeed/runtime/zero). ZeRO 1 does not have gradient hooks registered to overlap backward and gradient all-reduce, so it's fine to ignore `overlap_comm` and `contiguous_gradients`. However, in the current implementation, ZeRO 1 and 2 share almost the same implementation (`stage_1_and_2.py`). Features like `overlap_comm` and `contiguous_gradients` can also be enabled for ZeRO 1 (Please correct me if I made a mistake).

With this PR, turning on `overlap_comm` and `contiguous_gradients` for ZeRO 1 on the [SFT task](https://github.com/microsoft/DeepSpeedExamples/tree/master/applications/DeepSpeed-Chat/training/step1_supervised_finetuning) produces exactly the same training curve as the latest master.

image

I also see a ~1.05x e2e speedup by overlapping backward and gradient all-reduce. I can confirm by the trace that backward and all-reduce do overlap, and the separate gradients are indeed copied to a flat buffer. These options are also effective for ZeRO 1.

image

image

Related issue: https://github.com/microsoft/DeepSpeed/issues/2295


tjruwase:
@li-plus, thanks so much for this PR. Your analysis of the problem is accurate. Previously, we did not have bandwidth to evaluate the correctness of enabling those optimizations for ZeRO-1. This is a great contribution. Thanks so much.

tjruwase (approved):

PR #4895, opened 2024-01-04
fix falcon-40b accuracy issue

Yejing-Lai:
This [PR](https://github.com/microsoft/DeepSpeed/pull/4721) added the "DecoderLayer":glmtype. It will cause the Falcon model to choose "glmtype" fused_qkv_type. Falcon model (including Falcondecoderlayer) needs to choose 'bloomtype' explicitly.


delock:
Hi @Yejing-Lai  to better understand the accuracy issue this PR intend to fix, can you post a reproducer script and add output before/after the fix?  Thanks!

Yejing-Lai:
Hi @mrwyattii. From the failure log it seems like a device space issue. Can you check this issue? Thanks!

PR #4902, opened 2024-01-05
autoTP for Qwen

inkcherry:
Enabled autoTP for the Qwen model, added some module matching, and adjusted TP-related variables. Verification was conducted on Qwen-1_8B and Qwen-72B-chat.


deepspeed/module_inject/auto_tp.py
```diff
@@ -348,15 +350,14 @@ def _replace(self, child, name, conv_linear_layer):
                 child.weight.data = child.weight.data.transpose(-1, -2).contiguous()
 
             if require_tp_fused_qkvw(name, self.mp_size):
-                #for detecting fused type
-                module_str = str(self.module).strip()
+                #Check and handle fused qkv for TP
                 #The copy is a regular copy, The shape of dst and src is the same
                 data_dc = move(
-                    prepare_tp_fused_qkvw(module_str, child.weight.data, self.mp_size, mp_replace.gpu_index),
+                    prepare_tp_fused_qkvw(self.module, child.weight.data, self.mp_size, mp_replace.gpu_index),
```
loadams, line 356: Do we not need the `str().strip()` anymore?

inkcherry: we still need it. Because Qwen needs to modify some class member variables related to TP slicing, the ```self.module``` has been passed in,  the operation to retrieve the string has been moved into the function。 ```prepare_tp_fused_qkvw```

mrwyattii (approved):

inkcherry:
It seems that the failure did not come from this PR, and I can pass it locally. Could you please check it again? Thanks! @mrwyattii

deepspeed/module_inject/auto_tp.py
loadams: Yes, thanks!

PR #4935, opened 2024-01-11
Implement some APIs of HPU accelerator

mmhab:


accelerator/hpu_accelerator.py
```diff
@@ -186,33 +186,33 @@ def replay_graph(self, graph):
...
     def IntTensor(self):
-        return torch.hpu.IntTensor
+        return self.hpu.IntTensor
 
     @property
     def LongTensor(self):
-        return torch.hpu.LongTensor
+        return self.hpu.LongTensor
 
-    def pin_memory(self, tensor, align_bytes=1):
```
tjruwase, line 215: Why drop `align_bytes`? It is part of the abstraction and used by some accelerators.

mmhab: My mistake. Added it again.

mrwyattii (approved):

Issue #4857, opened 2023-12-21
[REQUEST] [TRITON] Upgrade Sparse Attention by Using Triton > 2.1


YizhouZ:
Hi, we are looking at deepspeed.ops.sparse_attention and find out that current SA is based on triton==1.0.0, which is old version. Current triton is 2.x and our supported version is 2.x. May I know if there is any plan on upgrading triton version to 2.x and maintain the sparse attention kernel?

My error stack is mainly on [deepspeed.ops.sparse_attention.matmul:](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/ops/sparse_attention/matmul.py)
```
import triton._C.libtriton as libtriton

segmented = libtriton.superblock(layout.data_ptr(),
                                         layout.shape[0],
                                         layout.shape[1],
                                         layout.shape[2],
                                         start_width)
```

Thanks!

delock:
Is https://github.com/microsoft/DeepSpeed/pull/4071 related to this request?

YizhouZ:
> Is #4071 related to this request?

Yes, but besides changing triton version, kernel needs updates as well.

delock:
> > Is #4071 related to this request?
> 
> Yes, but besides changing triton version, kernel needs updates as well.

What specific kernel error you met?  Is it a common error that people encountered when they upgrade to triton 2.1?

A-Cepheus:
got 
`File "python3.9/site-packages/deepspeed/ops/sparse_attention/matmul.py", line 276, in make_sdd_lut
    segmented = libtriton.superblock(layout.data_ptr(), layout.shape[0], layout.shape[1], layout.shape[2],
AttributeError: module 'triton._C.libtriton' has no attribute 'superblock'`

BurhanUlTayyab:
> > > Is #4071 related to this request?
> > 
> > 
> > Yes, but besides changing triton version, kernel needs updates as well.
> 
> Hi @YizhouZ , what specific kernel error you met? Is it a common error that people encountered when they upgrade to triton 2.1?

Yes, it is. Infact Triton has dropped support for ops in 2.0. 

Reference: https://github.com/openai/triton/issues/1395#issuecomment-1483725777

wyhsleep:
[...]

PR #4937, opened 2024-01-11
Support Triton 2.2+

loadams:


YizhouZ:
Hi, may I know this upgrading triton will also adjust triton kernels? https://github.com/microsoft/DeepSpeed/issues/4857

Thanks!

loadams:
> Hi, may I know this upgrading triton will also adjust triton kernels? #4857
> 
> Thanks!

Hi @YizhouZ - the main goal here was just to add what is needed to fix our nv-inference tests (currently failing above) when using the newly published triton 2.2.  For now we have pinned to 2.1 but this effort is to allow us to use 2.2, I probably won't adjust the kernels in this PR as well.

YizhouZ:
> > Hi, may I know this upgrading triton will also adjust triton kernels? #4857
> > Thanks!
> 
> Hi @YizhouZ - the main goal here was just to add what is needed to fix our nv-inference tests (currently failing above) when using the newly published triton 2.2. For now we have pinned to 2.1 but this effort is to allow us to use 2.2, I probably won't adjust the kernels in this PR as well.

Understood. May I know if there is any plan to upgrade kernel for triton-2.2?

stephen-youn:
It would be worth to manually run the following pytests in deepspeed with the latest 2.2  triton
image

.github/workflows/nv-inference.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         run: |
           unset TORCH_CUDA_ARCH_LIST # only jit compile for current arch
           cd tests
-          #pytest $PYTEST_OPTS -m 'seq_inference' unit/ --torch_ver="2.1" --cuda_ver="12.1"
```
loadams, line 61: TODO: Need to confirm this test actually runs.

PR #4952, opened 2024-01-13
Cache KV memory requirements

tohtana:
The computing KV cache requirements is slightly expensive. This PR reuses the computed KV cache requirements.
This is useful when you frequently check schedulablity.


tohtana:
Closing because we can get better results from https://github.com/microsoft/DeepSpeed/pull/4965 and https://github.com/microsoft/DeepSpeed-MII/pull/377.

PR #4964, opened 2024-01-16
Stage_1_and_2.py: fix assert for reduce_scatter configurations combinations

nelyahu:
today it does not allow to set reduce_scatter with:
- invalid comm data type
- gradient_predivide_factor != 1.0
- postscale_gradients is False but reduce scatter is relevant only for Zero2, also reduce_scatter is set to True by default. So setting prescale_gradients=True in json will end up with assert on Zero1.


tjruwase:
@nelyahu, what invalid comm_type would you want to skip this assert?

nelyahu:
@tjruwase - Actually not sure why this check for reduce_scatter is there at all. because this is very platform specific. can maybe be offloaded to accelerator level. One accelerator can support reduce_scatter with dtype X, and other won't.
But the main point here is that reduce scatter is not relevant to stage-1, only stage-2. so even if reduce_scatter=true, it will not have any meaning on Zero1.
I am mainly interested in the postscale_gradients assertion.

tjruwase:
@nelyahu, that makes sense.

tjruwase (approved):

PR #4974, opened 2024-01-18
Optimize grad_norm calculations by reducing device/host dependency

nelyahu:
Device tensors are being cast into python variable which requires device/host transactions. while all calculations can be done on 
the device using pytorch native operations.


nelyahu:
@tjruwase can you trigger test workflows on this PR?

PR #4980, opened 2024-01-19
FastGen Jan 2024 blog

mrwyattii:


RezaYazdaniAminabadi:
Hi folks, this is really great to see all of these improvement :) Btw, I have a PR on DS-kernels side to support the moe models on H100 in case you need it: https://github.com/microsoft/DeepSpeed-Kernels/pull/17

PR #4995, opened 2024-01-23
DS#4994 : runner param typo fix

### LINKED ISSUE #4994: [BUG] Deepspeed help parameter typo
image

oushu1zhangxiangxuan1:
#4994

### DIFF
--- a/deepspeed/launcher/runner.py
+++ b/deepspeed/launcher/runner.py
@@ -74,7 +74,7 @@ def parse_args(args=None):
                         type=str,
                         default="",
                         help='''Specify hardware resources to NOT use during execution.
-                        Mutually exclusive with --include. Resource formatting
+                        Mutually exclusive with --exclude. Resource formatting
                         is the same as --include.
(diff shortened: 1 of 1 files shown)


deepspeed/launcher/runner.py
```diff
@@ -74,7 +74,7 @@ def parse_args(args=None):
                         type=str,
                         default="",
                         help='''Specify hardware resources to NOT use during execution.
-                        Mutually exclusive with --include. Resource formatting
+                        Mutually exclusive with --exclude. Resource formatting
```
loadams, line 77: @oushu1zhangxiangxuan1 I believe this is correct as it was, the --exclude option is mutually exclusive, as in it is incompatible with the --include option, so you cannot use both.  So in the --exclude option should mention the include option.

oushu1zhangxiangxuan1: @loadams Thanks, you are right. I misunderstood it‘s meaning.

Issue #4903, opened 2024-01-05
[BUG] Falcon-40 have accuracy issue when using autotp


Yejing-Lai:
**Describe the bug**
Falcon-40b autotp text-generation task will get the wrong output.

**To Reproduce**
1. Simple inference script to reproduce
[reproduce script](https://github.com/intel/intel-extension-for-pytorch/blob/main/examples/cpu/inference/python/llm/distributed/run_generation_with_deepspeed.py)
2. What packages are required and their versions
ipex: 2.1.100
torch: 2.1
oneccl 2.1.0
3. How to run the script
deepspeed --num_gpus 2 --bind_cores_to_rank run_generation_with_deepspeed.py --benchmark -m  tiiuae/falcon-40b

**Expected behavior**
input: DeepSpeed is a machine learning framework for deep neural networks (DNNs) and deep reinforcement learning (DRL). It is written in C++ and is
expected output: DeepSpeed is a machine learning framework for deep neural networks (DNNs) and deep reinforcement learning (DRL). It is written in C++ and is based on PyTorch.DeepSpeed is designed to scale DNNs and DRL to large numbers of GPUs and TPUs.
but get the wrong output: DeepSpeed is a machine learning framework for deep neural networks (DNNs) and deep reinforcement learning (DRL). It is written in C++ and is e e e e e e e e e e e e e e e e e e e e e e e e e e e e e e

**ds_report output**
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
deepspeed_not_implemented  [NO] ....... [OKAY]
deepspeed_ccl_comm ..... [NO] ....... [OKAY]
cpu_adam ............... [NO] ....... [OKAY]
fused_adam ............. [NO] ....... [OKAY]
--------------------------------------------------
DeepSpeed general environment info:
torch install path ............... ['/xxx/python3.9/site-packages/torch']
torch version .................... 2.1.2+cpu
deepspeed install path ........... ['/xxx/deepspeed']
deepspeed info ................... 0.12.7+83427253, 83427253, master
deepspeed wheel compiled w. ...... torch 2.1

delock:
[...]

PR #5007, opened 2024-01-24
Update falcon fused type order

Yejing-Lai:
The selection of fused type depends on the order of fused_type_dict.
If put “DecoderLayer” in front of “FalconDecoderLayer”, Falcon will still choose glmtype incorrectly, so need to put “DecoderLayer at” the last position of fused_type_dict.

### DIFF
--- a/deepspeed/module_inject/fusedqkv_utils.py
+++ b/deepspeed/module_inject/fusedqkv_utils.py
@@ -5,7 +5,6 @@
 import torch
 from deepspeed.utils.logging import warning_once
 from deepspeed.module_inject.tp_shard import get_shard_size, get_shard_size_list, get_num_kv_heads, get_n_embd
-import re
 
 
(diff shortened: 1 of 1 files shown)


Yejing-Lai:
@mrwyattii @delock. Please kindly review this PR. Thanks!
Related issue: [issue_4903](https://github.com/microsoft/DeepSpeed/issues/4903)

mrwyattii:
@Yejing-Lai I've updated your PR to include a change with how we selecting the fused type. It selects the `module_name` with the longest string that contains the `module_str` substring. It's not perfect and there could still be a bug here, but I think it's better than having to keep a certain order of the dictionary items.

Can you please confirm this fixes the problem you were seeing? Thanks!

loadams (approved):

Yejing-Lai:
Hi @mrwyattii. Thanks for your update! This solved the problem I was meeting. Could you please merge this PR? Thanks!

Issue #4922, opened 2024-01-09
[BUG] DeepSpeed Zero3 Inference behavior error when model.train() mode


liu-zichen:
**Describe the bug**
There are two PartitionedParameterCoordinators for train and eval respectively in one model/(DeepSpeedEngine).

When performing inference in model.train() mode with torch.no_grad() using DeepSpeed's Zero3, there is an inconsistency in how PartitionedParameterCoordinator is selected. 

Select by the model training states (model.train() or model.eval()):
https://github.com/microsoft/DeepSpeed/blob/5457363c9b4f7a98971f59a2a9c6698f0da9e13a/deepspeed/runtime/zero/parameter_offload.py#L493-L507

Select by whether context torch.is_grad_enable:
https://github.com/microsoft/DeepSpeed/blob/5457363c9b4f7a98971f59a2a9c6698f0da9e13a/deepspeed/runtime/zero/parameter_offload.py#L342-L358

So when using torch.no_grad() but model.train() to inference, deepspeed zero3 would use eval PartitionedParameterCoordinator to fetch submodule and collect submodule_order, but use train PartitionedParameterCoordinator to reset_step().
So the submodule_order would keep growing (not the only bad behavior).

**To Reproduce**
```py
from transformers import AutoModelForCausalLM
import deepspeed

model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen-1_8B-Chat", trust_remote_code=True)
model, *_ = deepspeed.initialize(model=model, config=any_zero3_config)
model.train()

gen_kwargs = {'input_ids': ...}  # Add the necessary input for inference here
with torch.no_grad():
    model.module.generate(**gen_kwargs)
```

**Expected behavior**
All use the torch.is_grad_enable() to distinguish eval or train.

**Others**
An example for model.train() with torch.no_grad() from DeepSpeed-Chat:

https://github.com/microsoft/DeepSpeedExamples/blob/ff9a0234cf22dd9af03c5c7aa8037fb9143adca6/applications/DeepSpeed-Chat/dschat/rlhf/ppo_trainer.py#L148-L160

PR #5010, opened 2024-01-25
determine coordinator mode by is_grad

ByronHsu:
### Why?

As described in https://github.com/microsoft/DeepSpeed/issues/4922, the mode of coordinator is determined by `sub_module.training` at pre and 1st post forward hook, but by `is_grad` at the 2nd post forward hook for the case of inference.

If `model.train()` and `torch.no_grad()` are both set, in pre and 1st post forward hook, the state of "train coordinator" is updated, but in 2nd post forward hook, the state of "inference coordinator" is cleared and reset. It means that the resetting is on "eval coordinator", so the state of "train coordinator" will keep accumulating. 

### Test

WIP

### DIFF
--- a/deepspeed/runtime/zero/parameter_offload.py
+++ b/deepspeed/runtime/zero/parameter_offload.py
@@ -492,29 +492,29 @@ class DeepSpeedZeRoOffload(object):
         # post backward hook
         self.backward_hooks.append(module.register_forward_pre_hook(_post_backward_module_hook))
 
-    @torch.no_grad()
     def pre_sub_module_forward_function(self, sub_module):
         see_memory_usage(f"Before sub module function {sub_module.__class__.__name__}", force=False)
-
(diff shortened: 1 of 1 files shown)


HeyangQin (approved):

HeyangQin (changes requested):
I validated this PR with https://github.com/microsoft/DeepSpeedExamples/blob/master/applications/DeepSpeed-Chat/training/step3_rlhf_finetuning/training_scripts/llama2/run_llama2_7b.sh. Unfortunately, RLHF training with this PR fails at step 61 due to gradient overflow whereas the current DeepSpeed master trains well. I guess we need a different way to handle the coordinator mode. Will investigate more when I get the bandwidth.

loadams:
@ByronHsu - ping on Heyang's comment if you are able to follow up on that.

jomayeri:
Closing due to lack response.

PR #5018, opened 2024-01-26
Fix UserWarning: The torch.cuda.*DtypeTensor constructors are no long…

ShukantPal:
> …er recommended. It's best to use methods such as torch.tensor(data, dtype=*, device='cuda') to create tensors.

By currying the returned tensor properties in CUDA_Accelerator, the above warning in PyTorch is avoided.


finlytics-hub:
Any update on this please? Does this warning  have a functional impact on deepspeed or can it be safely ignored?

tjruwase (approved):

tjruwase:
> Any update on this please? Does this warning have a functional impact on deepspeed or can it be safely ignored?

I suspect that it is safe to ignore for current torch versions, but not future ones. Either way, @ShukantPal thanks for fixing this.

PR #5026, opened 2024-01-29
Split is_synchronized_device api to multiple apis

BacharL:
Deepspeed currently calls is_synchronized_device() to decide how to use the device. 
HPU does not fit into this definition since it behaves like all streams are blocking streams, 
meaning they preserve order between each other but asynchronous to CPU. 
see cudaStreamCreateWithFlags. 

**has_data_dependency_resolving()**
HPU device is considered synchronized wrt CPU. Operations executed in the script order 
regardless of stream they were enqueued on. Tensor data is guaranteed to be valid.
No need to stream dependencies or CPU synchronizations.

**use_host_timers()**
HPU device execution is async. To measure device execution time we must use device timers.

**has_memory_backpressure()**
limiting number of inflight fetched params and number of inflight grad reduce_scatter calls 
is not necessary since HPU will stop enqueuing calls if memory is full, creating internal 
backpressure for the CPU until memory is available.


accelerator/cpu_accelerator.py
```diff
@@ -21,6 +21,15 @@ def __init__(self):
     def is_synchronized_device(self):
         return True
 
+    def use_host_timers(self):
+        return self.is_synchronized_device()
+
+    def has_data_dependency_resolving(self):
+        return self.is_synchronized_device()
+
+    def has_memory_backpressure_handling(self):
```
tjruwase, line 30:
```suggestion
    def handles_memory_backpressure(self):
```

accelerator/cpu_accelerator.py
```diff
@@ -21,6 +21,15 @@ def __init__(self):
     def is_synchronized_device(self):
         return True
 
+    def use_host_timers(self):
+        return self.is_synchronized_device()
+
+    def has_data_dependency_resolving(self):
```
tjruwase, line 27:
```suggestion
    def resolves_data_dependency(self):
```

accelerator/abstract_accelerator.py
```diff
@@ -17,6 +17,18 @@ def __init__(self):
     def is_synchronized_device(self):
         ...
 
+    @abc.abstractmethod
+    def use_host_timers(self):
+        ...
+
+    @abc.abstractmethod
+    def has_data_dependency_resolving(self):
```
tjruwase, line 25: How about simplifying name with:

```suggestion
    def can_resolve_data_dependency(self):
```

or 
```suggestion
    def resolves_dependency(self):
```

PR #5048, opened 2024-01-31
[doc] update inference related docs from `mp_size` to `tensor_parallel` for TP

yundai424:
`mp_size` field is deprecated in flavor of `tensor_parallel`/`tp` https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/inference/engine.py so update related docs that are still sticking to `mp_size`

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -286,7 +286,7 @@ def init_inference(model, config=None, **kwargs):
     .. code-block:: python
 
         generator.model = deepspeed.init_inference(generator.model,
-                                                    mp_size=world_size,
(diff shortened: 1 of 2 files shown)


mrwyattii:
@yundai424 did you run into an error when trying `mp_size`? While it's deprecated, we still allow this option (for now): https://github.com/microsoft/DeepSpeed/blob/93e9537d4ccf0e54042ce98a910dcbc125bb8485/deepspeed/inference/config.py#L277

yundai424:
@mrwyattii you're right the run didn't have any error. Updated the PR summary to reflect the actual status. Thanks!

PR #5060, opened 2024-02-02
Further refactor deepspeed.moe.utils + deepspeed.moe.layer type hints

ringohoffman:
When unpacking a `dict`, keys that appear after the unpacking can overwrite the keys of the unpacked `dict`, meaning we can avoid avoid the pattern of skipping certain keys; also use `defaultdict` to avoid having to do the boilerplate of assigning the elements of `group_moe`.

More type hints and small stylistic changes to `deepspeed.moe.layer`

### DIFF
--- a/deepspeed/moe/layer.py
+++ b/deepspeed/moe/layer.py
@@ -46,7 +46,7 @@ class MoE(nn.Module):
                  use_residual: bool = False,
                  noisy_gate_policy: Optional[str] = None,
                  drop_tokens: bool = True,
-                 use_rts=True,
+                 use_rts: bool =True,
                  use_tutel: bool = False,
                  enable_expert_tensor_parallelism: bool = False) -> None:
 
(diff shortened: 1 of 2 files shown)


deepspeed/moe/layer.py
```diff
@@ -34,20 +35,20 @@ class MoE(torch.nn.Module):
...
+                 num_experts: int = 1,
+                 ep_size: int = 1,
+                 k: int = 1,
+                 capacity_factor: float = 1.0,
+                 eval_capacity_factor: float = 1.0,
+                 min_capacity: int = 4,
+                 use_residual: bool = False,
+                 noisy_gate_policy: Optional[str] = None,
                  drop_tokens: bool = True,
                  use_rts=True,
```
mrwyattii, line 49:
```suggestion
                 use_rts: bool =True,
```

mrwyattii (approved):

mrwyattii:
@ringohoffman could you please run formatting on this branch? `pre-commit run --all-files`

PR #5084, opened 2024-02-05
Rename cpu_accelerator to xeon_accelerator and create a new x86/cpu accelerator.

loadams:
There is a need for a generic x86 CPU accelerator that isn't the pre-existing cpu_accelerator which is being renamed to xeon_accelerator to better reflect the devices that it supports.


accelerator/x86_accelerator.py
```diff
@@ -0,0 +1,298 @@
...
+
+    @property
+    def Event(self):
+        return None
+
+    # Memory management
+    def empty_cache(self):
+        return
+
+    def memory_allocated(self, device_index=None):
```
loadams, line 101: @mrwyattii - do we want to do this with psutil the same way the current cpu accelerator does?

accelerator/x86_accelerator.py
```diff
@@ -0,0 +1,297 @@
...
+        self._name = 'x86'
+
+    def is_synchronized_device(self):
+        return False
+
+    # Device APIs
+    def device_name(self, device_index=None):
+        if device_index is None:
+            return 'x86'
+        return 'x86:{}'.format(device_index)
```
tjruwase, line 31: Are we sure this is naming convention for torch on cpu?

loadams: It isn't, the naming convention is `cpu`, so this needs to be updated.  The first pass was largely find/replace, though we will also need to figure out how to handle this with the current cpu_accelerator.

accelerator/x86_accelerator.py
```diff
@@ -0,0 +1,297 @@
...
+    pass
+
+
+class x86_Accelerator(DeepSpeedAccelerator):
+
+    def __init__(self):
+        self._name = 'x86'
+
+    def is_synchronized_device(self):
+        return False
```
tjruwase, line 25: I believe CPU accelerators are synchronized devices because of sequential execution semantics, unlike GPUs with asynchronous execution that require explicit synchronization operations. 
This reminds me that proper documentation is overdue for this feature :(

accelerator/x86_accelerator.py
mrwyattii: Yes, I think that is the right thing. I don't think torch has any cpu-specific memory information available like for cuda devices

accelerator/x86_accelerator.py
tjruwase: Sounds good to me. Actually, based on my chat with Mike, this file should mostly mirror cpu_accelerator while cpu_accelerator is renamed to xeon_accelerator. I am happy to chat more about this.

accelerator/x86_accelerator.py
loadams: Correct, this has been updated.

accelerator/real_accelerator.py
```diff
@@ -154,38 +154,46 @@ def get_accelerator():
...
+
+                # Determine if we are on a GPU or x86 CPU with torch.
+                if torch.cuda.is_available():
+                    accelerator_name = "cuda"
+                else:
+                    accelerator_name = "x86"
+            except (RuntimeError, ImportError) as e:
+                pass
 
+        # What do we want to default to here?
```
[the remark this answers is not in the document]
loadams: Updated, will add the warning next

accelerator/x86_accelerator.py
loadams: Sounds good, updated.

accelerator/x86_accelerator.py
CaoZhongZ: Hi @tjruwase, What is the hurdle that we can't merge x86 and xeon accelerator? We could try to make it work for all x86, given xeon is x86.

loadams: Agreed, I've renamed the accelerator, though the git tracking of this treats it as a new file, so the review will be somewhat difficult.  If we want to rename the cpu_accelerator -> xeon_accelerator in a PR first that wil be easier for git to track.  Thoughts?

loadams:
@delock - thoughts on the rename here to make it more clear what each accelerator is for?

accelerator/x86_accelerator.py
mrwyattii: I created #5126 to better preserve git commit history

accelerator/real_accelerator.py
```diff
@@ -154,38 +154,47 @@ def get_accelerator():
...
+                if torch.cuda.is_available():
+                    accelerator_name = "cuda"
+                else:
+                    if accel_logger is not None:
+                        accel_logger.warn(
+                            "Setting accelerator to CPU. If you have GPU or other accelerator, we were unable to detect it."
+                        )
+                    accelerator_name = "cpu"
+            except (RuntimeError, ImportError) as e:
+                pass
```
delock, line 170: The behavior when pytorch is not installed should be defined.

PR #5088, opened 2024-02-06
use ninja to speed up build

jinzhen-lin:
Deepspeed have too many ops now, and it take too many time to pre-build all ops.
I notice deepspeed disabled `ninja` 4 years ago (https://github.com/microsoft/DeepSpeed/pull/298) and I think we should consider enable it now.
The issue mentioned in https://github.com/microsoft/DeepSpeed/pull/298 can be solved by resolving `include_dirs` to absolute path.


op_builder/inference_core_ops.py
```diff
@@ -60,13 +60,13 @@ def sources(self):
         sources = [
             "inference/v2/kernels/core_ops/core_ops.cpp",
             "inference/v2/kernels/core_ops/bias_activations/bias_activation.cpp",
-            "inference/v2/kernels/core_ops/bias_activations/bias_activation.cu",
+            "inference/v2/kernels/core_ops/bias_activations/bias_activation_cuda.cu",
```
loadams, line 63: @jinzhen-lin - why rename these files to contain the _cuda suffix?

jinzhen-lin: Both `bias_activation.cpp` and `bias_activation.cu` generate `bias_activation.o`. Without renaming, ninja would throw an error
```
ninja: error: build.ninja:29: multiple rules generate xxxx/bias_activation.o
```

loadams: I see - I know we previously had ninja support and that it is broken, but we should also get a review here from @mrwyattii

jinzhen-lin:
Accidentally closed PR, reopen it.

op_builder/inference_core_ops.py
mrwyattii: I think this PR looks good. Since we don't have test workflow for XPU & HPU accelerators on this repo (yet), we should verify manually that we're not breaking anything there before merging.

mrwyattii (approved):
Thanks @jinzhen-lin this is a great improvement in compile time for all the ops (~18 min --> ~10 min)!

setup.py
```diff
@@ -119,7 +119,7 @@ def get_env_if_set(key, default: typing.Any = ""):
 # For any pre-installed ops force disable ninja.
 if torch_available:
     from accelerator import get_accelerator
-    cmdclass['build_ext'] = get_accelerator().build_extension().with_options(use_ninja=False)
+    cmdclass['build_ext'] = get_accelerator().build_extension().with_options()
```
tjruwase, line 122: Can we have ninja usage control with an environment variable? It will great for users to disable ninja if needed.

jinzhen-lin: @tjruwase I have add env vars to avoid hardcoding the value of use_ninja and nvcc_threads

setup.py
```diff
@@ -119,7 +119,10 @@ def get_env_if_set(key, default: typing.Any = ""):
 # For any pre-installed ops force disable ninja.
 if torch_available:
     from accelerator import get_accelerator
-    cmdclass['build_ext'] = get_accelerator().build_extension().with_options(use_ninja=False)
+    use_ninja = True
```
tjruwase, line 122: How about

```suggestion
    use_ninja = not is_set_env("DS_DISABLE_NINJA")
```

jinzhen-lin: > How about

This is a better way.
Done it.

op_builder/inference_core_ops.py
tjruwase: @delock, @nelyahu, @hipudding FYI

Please note this sensitive PR since we could only validate on cuda/hip/cpu.

tjruwase:
@jinzhen-lin, thanks for improving the PR. Can you please resolve the formatting issues?
https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites

jinzhen-lin:
> @jinzhen-lin, thanks for improving the PR. Can you please resolve the formatting issues? https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites

Done.

tjruwase:
> Done.

Thanks! Please note that we are investigating the following CI failure, which is unrelated to your PR. 
image

mrwyattii (approved):

loadams:
> > Done.
> 
> Thanks! Please note that we are investigating the following CI failure, which is unrelated to your PR. image

This should be resolved now.

Thanks @jinzhen-lin - this speeds up our nv-pre-compile-ops builds by roughly 2x.

loadams (approved):

PR #5095, opened 2024-02-07
[NPU] replace 'cuda' with get_accelerator().device_name()

minchao-sun:
Replace 'cuda' with `get_accelerator().device_name()` to support other accelerators.

I searched the whole repo trying to fix same issue and this seems to be the only one.


loadams:
@mrwyattii - this isn't that far off from the checks for torch.cuda, could be worth extending to this case as well where the device is hardcoded to cuda?

loadams (approved):

PR #5099, opened 2024-02-08
estimate_zero2_model_states_mem_needs: fixing memory estiamtion

nelyahu:
was considering 4 bytes per model param, and 4 bytes per gradient. 
fixed it to 2 bytes - under the assumption of FP16/BF16


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2425,7 +2425,9 @@ def estimate_zero2_model_states_mem_needs(total_params,
         gpu_mem = 2 * total_params
         cpu_mem = total_params * max(4 * total_gpus, 16) * additional_buffer_factor
     else:
-        gpu_mem = 4 * total_params + int(16 * total_params / total_gpus)
+        # GPU's total_params multipliers: 2 = params_16bit,
+        # 14 = 2_grads_16bit + 4_params_32bit + 8_optimizer_states_32bit(momentum and variance)
```
tjruwase, line 2429: Interestingly, @stas00 and I were recently debating whether grads is 2 bytes or 4 bytes. I believe we have gone back-n-forth on this :). 

Anyways, the rationale for 4 bytes is that optimizer step is a 4-byte computation and so requires 4-bytes gradients. On the other hand, gradients generated in the backward are 2 bytes and so need to be upcasted for optimizer. So, the question is which more appropriate for estimation and that seems to be implementation-specific. One approach is to upcast the full gradients into HBM before optimizer computation, another is to upcast on-the-fly into register during optimizer computation. DeepSpeed generally does the former because of the ease-of-integration into existing Adam* optimizers, while the later is more memory-efficient. 

Anyways, I guess the main question, at least to me, is whether this estimation should reflect
1. Theoretical lower bounds (2 bytes), even though some implementations (e.g., DeepSpeed) may violate or 
2. DeepSpeed's implementation (4 bytes). 
 
I actually favor (1), because it is forcing function for DeepSpeed optimizations. 

Glad to get your thoughts.

tjruwase: It just occurred to me that another point for (1) is that it might motivate hardware that automatically upcasts (to relieve developer burden) or optimizes existing upcasting logic. Although, some correctness considerations come to mind, but I wonder if this thought is too far-fetched @nelyahu?

nelyahu: Thanks @tjruwase, that is good point that you are raising, Which takes me a little bit further.
I discussed with @BacharL  - The below code, occurs during the stage-2 optimizer step inside a loop that goes through all optimizer groups: https://github.com/microsoft/DeepSpeed/blob/3255569b785f4561a0bce7913f7f00c1f6fde0c6/deepspeed/runtime/zero/stage_1_and_2.py#L1858
for every group it cast the flat buffer to FP32.
So in point of time we are holding both 16bit and 32bit copies for that group gradient, which makes it 6 bytes for gradient in that group.
In worst case scenario of 1 optimizer param group we can consider 6 bytes for gradient.

What do you think?

stas00: Tunji, why would you pick a theoretical value, when this service is designed to tell the user the truth. If deepspeed uses 4 bytes then it's 4 bytes, if it's 6 then it should count it as 6, until it's re-coded to be otherwise. 

Otherwise it defeats the purpose of this feature. The user uses it to estimate things and when they come to run the program it'll take more memory, why then provide a misleading feature?

Additionally, there is fp32 comms which will too upcast to fp32 for the maximum size of the reduce_bucket, if the user configured it.

I have noticed that many people trying to measure memory usage forget to look at peak memory usage and just measure the total memory usage at the end of the iteration, which obviously doesn't work.

I usually send them to https://github.com/stas00/ipyexperiments which automatically shows them the true situation which includes peak memory reporting.

tjruwase: > Tunji, why would you pick a theoretical value, when this service is designed to tell the user the truth. If deepspeed uses 4 bytes then it's 4 bytes, if it's 6 then it should count it as 6, until it's re-coded to be otherwise.

@stas00, you are correct. My proposal would indeed lead to more confusion. 

@nelyahu, I think it is fair to use 6 bytes to handle the simultaneous existence of 16bit and 32bit versions of the gradients. I think that would also cover the instance of upcasting for gradient reduction, since that does not overlap with optimizer step.

stas00: > the simultaneous existence of 16bit and 32bit versions of the gradients

are those both static preallocated buffers or are they allocated on the fly anew for each iteration?

if static then it's definitely 6 bytes.

tjruwase: > are those both static preallocated buffers or are they allocated on the fly anew for each iteration?

No, they are allocated on the fly anew each time. So, it is somewhere between 4 bytes and 6 bytes. I am fine with conservatively using 6 bytes. As @stas00 pointed out, it is better to avoid surprising users with unexpected OOMs.

stas00: so if there is a single param group then we have a full 6 bytes per param allocation as peak memory. if there are multiple param groups is it still all grad that get to 6 bytes per param or only that of the largest param group?

the reason I'm asking, is that if it's always all params, perhaps it be possible then to create a static fp32 buffer (4 bytes) and cast it in place to fp32 or half-precision depending on the need and never need to allocate 2+4 bytes?

And, of course, once `torch.compile` works w/ DS it'd be beneficial to write a DS version of AdamW that only gets half-precision grads and any upcasting would happen in registers - saving a whooping 4 bytes per param across the board. The old mechanism would work for non-DS optims, but DS will sport a more efficient optimizer which many users would choose to save memory.

nelyahu: @stas00 

> so if there is a single param group then we have a full 6 bytes per param allocation as peak memory. if there are multiple param groups is it still all grad that get to 6 bytes per param or only that of the largest param group?

Largest, as it is allocated inside a loop iteration over the groups, so the peak will be met on the largest one. But this is not the absolute peak right? it does not take into account the activation peak memory at the beginning of the BWD.

> the reason I'm asking, is that if it's always all params, perhaps it be possible then to create a static fp32 buffer (4 bytes) and cast it in place to fp32 or half-precision depending on the need and never need to allocate 2+4 bytes?

Then you are doubling the memory consumption of the gradients during the BWD pass. and one of the incentives of working with BF16 gradients is memory (the other is performance).

tjruwase: > the reason I'm asking, is that if it's always all params, perhaps it be possible then to create a static fp32 buffer (4 bytes) and cast it in place to fp32 or half-precision depending on the need and never need to allocate 2+4 bytes?

I don't understand how this will work since torch creates `.grad` on-the-fly and we don't control the buffer it uses.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2427,7 +2427,9 @@ def estimate_zero2_model_states_mem_needs(total_params,
...
```
tjruwase, line 2431:
```suggestion
        # 18 = 2_grads_16bit + 4_grads_32bit + 4_params_32bit + 8_optimizer_states_32bit(momentum and variance)
```

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2427,7 +2427,9 @@ def estimate_zero2_model_states_mem_needs(total_params,
...
+        gpu_mem = 2 * total_params + int(14 * total_params / total_gpus)
```
tjruwase, line 2432:
```suggestion
        gpu_mem = 2 * total_params + int(18 * total_params / total_gpus)
```

tjruwase: @nelyahu, apologies for the delay in resolving this. I have made suggestions to estimate 6bytes for gradients as that more closely aligns with the current implementation.

deepspeed/runtime/zero/stage_1_and_2.py
stas00: it sounds that my suggestion is not applicable then.

nelyahu:
@tjruwase @stas00  - What do you say about the updated patch?

tjruwase (approved):

tjruwase:
> @tjruwase @stas00 - What do you say about the updated patch?

@nelyahu, LGTM. Thanks!

nelyahu:
@tjruwase  @stas00  can you please re-run validation? the failure in "cpu-torch-latest" does not seem related

nelyahu:
@tjruwase can be merged?

PR #5110, opened 2024-02-09
Skip Triton import for AMD

lekurile:
When testing DeepSpeed inference on an `AMD Instinct MI250X/MI250` GPU, the `pytorch-triton-rocm` module would break the `torch.cuda` device API. To address this, importing `triton` is skipped when the GPU is determined to be `AMD`.

This change allows DeepSpeed to be executed on an AMD GPU w/o kernel injection in the DeepSpeedExamples [text-generation example](https://github.com/microsoft/DeepSpeedExamples/tree/master/inference/huggingface/text-generation) using the following command:
```bash
deepspeed --num_gpus 1 inference-test.py --model facebook/opt-125m
```

TODO: Root-cause the interaction between `pytorch-triton-rocm` and DeepSpeed to understand why this is causing the `torch.cuda` device API to break.

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -11,9 +11,10 @@ import torch
 from torch.optim import Optimizer
 from torch.optim.lr_scheduler import _LRScheduler
 from packaging import version as pkg_version
+from .accelerator import get_accelerator
 
 # Skip Triton import for AMD due to pytorch-triton-rocm module breaking device API in DeepSpeed
-if 'AMD' not in torch.cuda.get_device_name():  #ignore-cuda
+if not get_accelerator().is_rocm_device():
     try:
         import triton  # noqa: F401 # type: ignore
         HAS_TRITON = True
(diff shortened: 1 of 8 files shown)


deepspeed/__init__.py
```diff
@@ -12,10 +12,14 @@
...
-except ImportError:
+# Skip Triton import for AMD due to pytorch-triton-rocm module breaking device API in DeepSpeed
+if 'AMD' not in torch.cuda.get_device_name():
+    try:
+        import triton  # noqa: F401 # type: ignore
+        HAS_TRITON = True
+    except ImportError:
+        HAS_TRITON = False
+else:
     HAS_TRITON = False
```
mrwyattii:
```suggestion
from deepspeed.accelerator import get_accelerator

# Skip Triton import for AMD due to pytorch-triton-rocm module breaking device API in DeepSpeed
HAS_TRITON = False

if not get_accelerator().is_rocm_device():
    try:
        import triton  #  noqa: F401 # type: ignore
        HAS_TRITON = True
    except ImportError:
        pass
```

Because of the accelerator abstraction, we cannot use `torch.cuda` calls directly. I recommend expanding the accelerator to have a `is_rocm_device` method and using it here.

mrwyattii (approved):

PR #5115, opened 2024-02-11
Check overflow on device without host synchronization for each tensor

BacharL:


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1953,24 +1953,26 @@ def _check_overflow(self, partition_gradients=True):
...
     def has_overflow_partitioned_grads_serial(self):
+        invalid_grad_count = torch.zeros([1], dtype=torch.float, device=get_accelerator().current_device_name())
         for i in range(len(self.bit16_groups)):
             for j, grad in enumerate(self.averaged_gradients[i]):
-                if grad is not None and self._has_inf_or_nan(grad.data, j):
-                    return True
-        return False
+                if grad is not None:
+                    invalid_grad_count += self._has_inf_or_nan(grad)
+        return invalid_grad_count
```
tjruwase, line 1969: I believe this should return a bool. For example, note that `self.local_overflow` is bool in line 1974.

BacharL: I intentionally changed the return value to tensor. self.local_overflow is used in the case of cpu_offload.
When not using cpu_offload this function returned result on CPU, causing device sync. The result is moved back to device in line 1975.
Is ok if I returned bool tensor instead of float tensor?

tjruwase: Yes, bool tensor works. Thanks!

BacharL: done

tjruwase (approved):

tjruwase:
@BacharL, can you please help with the formatting issues? Thanks!

PR #5120, opened 2024-02-12
Add HIP device abstraction, update Triton skip logic

lekurile:
This PR adds a `HIP_Accelerator` accelerator abstraction that inherits from `CUDA_Accelerator`, which is possible due to PyTorch for HIP reusing the existing `torch.cuda` interfaces ([see here](https://pytorch.org/docs/stable/notes/hip.html#hip-interfaces-reuse-the-cuda-interfaces)).

The motivation for a new HIP acceleration abstraction however, is to expand the flexibility when supporting AMD, allowing for accelerator function overrides if necessary, such as (see GH-5110):
```python
        return False
```
The `triton` import skip logic was also updated in this PR to source `is_triton_supported()` from the accelerator.

### DIFF
--- a/accelerator/hip_accelerator.py
+++ b/accelerator/hip_accelerator.py
@@ -1,0 +1,39 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+from .cuda_accelerator import CUDA_Accelerator
+# During setup stage torch may not be installed, pass on no torch will
+# allow op builder related API to be executed.
+try:
+    import torch.cuda
+except ImportError:
+    pass
+
+# Delay import pynvml to avoid import error when CUDA is not available
+pynvml = None
+
+
+class HIP_Accelerator(CUDA_Accelerator):
+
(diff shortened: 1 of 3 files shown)


loadams:
FYI @rraminen and @jithunnair-amd - please review if you can.

loadams (approved):

accelerator/hip_accelerator.py
```diff
@@ -0,0 +1,39 @@
...
+        global pynvml
+        try:
+            import pynvml
+        except ImportError:
+            return
+        try:
+            pynvml.nvmlInit()
+        except pynvml.NVMLError:
+            pynvml = None
+            return
```
mrwyattii, line 36: Does `pynvml` work with AMD GPUs? Either way, I think we can remove this as it's only necessary for inference V2, which is not supported on AMD yet.

lekurile: Good point, cleaned up the HIP abstraction and remove unnecessary aspects.

loadams:
Note: we need to complete !5126 and !5126 before completing this PR.

PR #5121, opened 2024-02-12
disable compile tests for torch<2.0

mrwyattii:
Tests running older version of torch will fail the compile tests added in #4878.


loadams (approved):

loadams:
Ah good catch on 2.1+

PR #5126, opened 2024-02-13
move CPU_Accelerator --> Xeon_Accelerator

mrwyattii:
Extracting work from #5084 to have a cleaner commit history. Merge this first! @loadams


loadams:
@delock - this may be a better PR to review to confirm you're okay with this rename.

loadams (approved):

delock:
> @delock - this may be a better PR to review to confirm you're okay with this rename.

Hi @loadams , thanks for the update.  As I read from this PR and also from #5084 I can capture the following summarizes.
1. CPU_Accelerator would be change to Xeon_Accelerator, and CPU_Accelerator is refractored into X86_Accelerator.
2. Xeon_Accelerator is intended to be more Intel device specialized, X86_Accelerator is more generized.
3. From real_accelerator code change, Xeon_Accelerator would be activated when Intel Extension for PyTorch is installed, and `ccl` backend is used.  And X86_Accelerator would be activated with stock PyTorch (no extension installed) while CUDA device is not presented on the system;  It is not clear which communication backend would be used for x86 accelerator.
4. Both accelerator and op_builder for Xeon_Accelerator is renamed, the kernel directory `csrc/cpu` would remain in current name.

List not clear to me:
1. What is the major difference between Xeon_Accelerator and X86_Accelerator?  I can see two differences: 1. the selection logic 2. whether to use stock pytorch only or not.   Are both difference intended?
2. Will op builder and kernel(csrc/cpu) be shared between Xeon_Accelerator and X86_Accelerator?
3. What is the intended usage scenario of each Accelerator?

We asked these question because we want to evaluate whether these two Accelerators can be aligned.  For example, if the selection logic of CPU accelerator is no PyTorch extension is installed and no CUDA device is installed, I think it's an okay scenario for Xeon Accelerator as well.  We can check alignment on other aspects and see if one Accelerator can meet the need for both Intel CPU and non-Intel CPU.  @CaoZhongZ @rogerxfeng8

delock:
@loadams There is another comment.  The reason we initially use extension detection to select CPU_Accelerator is mainly because we don't want to change existing behavior on CUDA environment.  Consider the following scenario:

1. User install DeepSpeed on login node of a cluster, which usually does not have CUDA hardware installed.  This could trigger X86_Accelerator and install a different set of op builder, following the logic in setup.py (https://github.com/microsoft/DeepSpeed/blob/083197ea1be39cf6781da4f8005eacc2ee40819e/setup.py#L172)
2. User lately get a compute node with CUDA hardware, and run DeepSpeed on it.  So user may get suprised that some CUDA op are not usable.

I think it might be a common practice for cluster users because compute node sometimes are hard to allocate.  Not sure how DeepSpeed would solve this potential issue with new selection logic.

delock:
@loadams people in Shanghai is in holiday and could get back to office starting from Feb 18.   It would be better if we can run some tests through this PR in our environment after the holidy and give you feedback from test results.

loadams:
> @loadams people in Shanghai is in holiday and could get back to office starting from Feb 18. It would be better if we can run some tests through this PR in our environment after the holidy and give you feedback from test results.

@delock - thanks for your reply while on holiday, we can certainly wait to merge this for tests to be run on your side as well, just wanted to give you a general heads up, and we will work on the answers to the other questions in the meantime, thanks!

loadams:
> > @delock - this may be a better PR to review to confirm you're okay with this rename.
> 
> Hi @loadams , thanks for the update. As I read from this PR and also from #5084 I can capture the following summarizes.
> 
> 1. CPU_Accelerator would be change to Xeon_Accelerator, and CPU_Accelerator is refractored into X86_Accelerator.
> 2. Xeon_Accelerator is intended to be more Intel device specialized, X86_Accelerator is more generized.
> 3. From real_accelerator code change, Xeon_Accelerator would be activated when Intel Extension for PyTorch is installed, and `ccl` backend is used.  And X86_Accelerator would be activated with stock PyTorch (no extension installed) while CUDA device is not presented on the system;  It is not clear which communication backend would be used for x86 accelerator.
> 4. Both accelerator and op_builder for Xeon_Accelerator is renamed, the kernel directory `csrc/cpu` would remain in current name.
> 
> List not clear to me:
> 
> 1. What is the major difference between Xeon_Accelerator and X86_Accelerator?  I can see two differences: 1. the selection logic 2. whether to use stock pytorch only or not.   Are both difference intended?
> 2. Will op builder and kernel(csrc/cpu) be shared between Xeon_Accelerator and X86_Accelerator?
> 3. What is the intended usage scenario of each Accelerator?
> 4. Is there scenario that works on X86_Accelerator but may not work on Xeon_Accelerator?
> 
> We asked these question because we want to evaluate whether these two Accelerators can be aligned. For example, if the selection logic of CPU accelerator is no PyTorch extension is installed and no CUDA device is installed, I think it's an okay scenario for Xeon Accelerator as well. We can check alignment on other aspects and see if one Accelerator can meet the need for both Intel CPU and non-Intel CPU. @CaoZhongZ @rogerxfeng8

Hi @delock 

To try and answer your questions, but overall, yes, the intent is to move the current cpu_accelerator to be renamed to xeon acceleartor, that's the accelerator for when folks have the intel extensions for pytorch/specific oneCCL software they want to use.  The new cpu_accelerator is meant to be more basic, it is meant to be for just someone on a cpu who is trying out pytorch, think of it as all they have is the pytorch+cpu version to be the most generic accelerator/last resort default.

1. Yes, the main difference is the stock pytorch and the lack of a CCL in the cpu_accelerator.
2. Yes, both will use the cpu_op builder which is now reflected in that PR.
3. The intended use scenario is to use the xeon_accelerator for all the cases where folks want that performance/setup their sytem for that, and the cpu_accelerator for the default base case where the user is just testing out DeepSpeed, perhaps has no accelerator at all.
4. I do not believe so, no.
cc: @tjruwase and @mrwyattii

mrwyattii:
> @loadams There is another comment. The reason we initially use extension detection to select CPU_Accelerator is mainly because we don't want to change existing behavior on CUDA environment. Consider the following scenario:
> 
> 1. User install DeepSpeed on login node of a cluster, which usually does not have CUDA hardware installed.  This could trigger X86_Accelerator and install a different set of op builder, following the logic in setup.py (https://github.com/microsoft/DeepSpeed/blob/083197ea1be39cf6781da4f8005eacc2ee40819e/setup.py#L172
>    )
> 2. User lately get a compute node with CUDA hardware, and run DeepSpeed on it.  So user may get suprised that some CUDA op are not usable.
> 
> I think it might be a common practice for cluster users because compute node sometimes are hard to allocate. Not sure how DeepSpeed would solve this potential issue with new selection logic.

I think this is a valid concern, but one that could be fixed by the user setting `DS_ACCELERATOR=cuda` in the environment: https://github.com/microsoft/DeepSpeed/blob/177dc14331a64e61f6dcce2c4b8071576bcb22db/accelerator/real_accelerator.py#L59

@tjruwase thoughts on this?

loadams:
> > I think this is a valid concern, but one that could be fixed by the user setting `DS_ACCELERATOR=cuda` in the environment:
> > https://github.com/microsoft/DeepSpeed/blob/177dc14331a64e61f6dcce2c4b8071576bcb22db/accelerator/real_accelerator.py#L59
> > 
> > @tjruwase thoughts on this?
> 
> Thanks for the good discussion on this issue. It seems there are two questions to address:
> 
> 1. Which accelerator ops to build at installation if detection logic fails to find any hardware? Currently default is CUDA.
> 2. Can existing cpu_accelerator work on non-Xeon hardware? We currently see CI issues on AMD.
> 
> For (1), I think the proposal of using env var to override the detection logic could work. But it will require clear documentation since it breaks backwards-compatibility. I do like the idea of moving away from implicit accelerator choice and CUDA default in recognition of the growing environment diversity. And this is not just because of accelerator diversity but also due to differences in the installation environment (no accelerator available) and the run environment (accelerator available). So I am considering two options here for discussion.
> 
> 1. Is it reasonable to require that users leverage the env var mechanism for accelerator selection in these complex cases?
> 2. Is it possible that we delay accelerator/ops compilation until deepspeed launch time when the detection logic will work?
> 
> For (2), @mrwyattii can you please share the context for the failing AMD CI with @delock to see if they can fix on their side? Will that work?

@tjruwase The failing AMD tests aren't due to anything on the abstract_accelerator side, but rather something with the underlying hardware infrastructure we use to test with.  It seems the nodes just are instantly removed while we are running on them, so not much we can fix in software.

delock:
From this link AMD CPU should be able to run with Intel Extension For PyTorch, but is not in daily validation scope.
https://github.com/intel/intel-extension-for-pytorch/issues/278

On the other hand, 'node instantly removed' sounds like a system level failure rather than compatibility issue.  Are there any more detail?  If run cpu inference workflow on AMD instance, what is the error log? There are some logging in this workflow which may tell us more information.

> > > I think this is a valid concern, but one that could be fixed by the user setting `DS_ACCELERATOR=cuda` in the environment:
> > > https://github.com/microsoft/DeepSpeed/blob/177dc14331a64e61f6dcce2c4b8071576bcb22db/accelerator/real_accelerator.py#L59
> > > 
> > > @tjruwase thoughts on this?
> > 
> > 
> > Thanks for the good discussion on this issue. It seems there are two questions to address:
> > 
> > 1. Which accelerator ops to build at installation if detection logic fails to find any hardware? Currently default is CUDA.
> > 2. Can existing cpu_accelerator work on non-Xeon hardware? We currently see CI issues on AMD.
> > 
> > For (1), I think the proposal of using env var to override the detection logic could work. But it will require clear documentation since it breaks backwards-compatibility. I do like the idea of moving away from implicit accelerator choice and CUDA default in recognition of the growing environment diversity. And this is not just because of accelerator diversity but also due to differences in the installation environment (no accelerator available) and the run environment (accelerator available). So I am considering two options here for discussion.
> > 
> > 1. Is it reasonable to require that users leverage the env var mechanism for accelerator selection in these complex cases?
> > 2. Is it possible that we delay accelerator/ops compilation until deepspeed launch time when the detection logic will work?
> > 
> > For (2), @mrwyattii can you please share the context for the failing AMD CI with @delock to see if they can fix on their side? Will that work?
> 
> @tjruwase The failing AMD tests aren't due to anything on the abstract_accelerator side, but rather something with the underlying hardware infrastructure we use to test with. It seems the nodes just are instantly removed while we are running on them, so not much we can fix in software.

delock:
I think detect ops at launch time rather than installation time should be more proper solution, this can make deepspeed more stable to HW change.  On the other hand, if user workflow require setting env variable, this would add additional complexity.  It is better if environment variable overriding is used for troubleshooting only rather than being relied on.
> > I think this is a valid concern, but one that could be fixed by the user setting `DS_ACCELERATOR=cuda` in the environment:
> > https://github.com/microsoft/DeepSpeed/blob/177dc14331a64e61f6dcce2c4b8071576bcb22db/accelerator/real_accelerator.py#L59
> > 
> > @tjruwase thoughts on this?
> 
> Thanks for the good discussion on this issue. It seems there are two questions to address:
> 
> 1. Which accelerator ops to build at installation if detection logic fails to find any hardware? Currently default is CUDA.
> 2. Can existing cpu_accelerator work on non-Xeon hardware? We currently see CI issues on AMD.
> 
> For (1), I think the proposal of using env var to override the detection logic could work. But it will require clear documentation since it breaks backwards-compatibility. I do like the idea of moving away from implicit accelerator choice and CUDA default in recognition of the growing environment diversity. And this is not just because of accelerator diversity but also due to differences in the installation environment (no accelerator available) and the run environment (accelerator available). So I am considering two options here for discussion.
> 
> 1. Is it reasonable to require that users leverage the env var mechanism for accelerator selection in these complex cases?
> 2. Is it possible that we delay accelerator/ops compilation until deepspeed launch time when the detection logic will work?
> 
> For (2), @mrwyattii can you please share the context for the failing AMD CI with @delock to see if they can fix on their side? Will that work?

tjruwase:
> I think detect ops at launch time rather than installation time should be more proper solution, this can make deepspeed more stable to HW change. On the other hand, if user workflow require setting env variable, this would add additional complexity. It is better if environment variable overriding is used for troubleshooting only rather than being used as an interface. For example, if it's an interface, then change accelerator's name will have impact on existing scripts.

@delock, I fully agree with you on this. Can we work together to enable op detection at launch time?

delock:
Yes sure.  Let's open an issue and discuss from there.  This could simplify installation for many accelerators.

> > I think detect ops at launch time rather than installation time should be more proper solution, this can make deepspeed more stable to HW change. On the other hand, if user workflow require setting env variable, this would add additional complexity. It is better if environment variable overriding is used for troubleshooting only rather than being used as an interface. For example, if it's an interface, then change accelerator's name will have impact on existing scripts.
> 
> @delock, I fully agree with you on this. Can we work together to enable op detection at launch time?

delock:
Thoughts on some actions we may possibly do:
1. Check the AMD failure log to see if its something easy to fix
2. Test whether current CPU_Accelerator can run without Intel Extension for PyTorch installed
3. I don't have better suggestion for oneCCL binding for PyTorch ('ccl' communication backend), because this seem to be the only library capable of BF16 communication.  This would be important for training with CPU.

If these actions are effective, then maybe we don't need two seperate accelerator for CPU.  This could make user less confusion (i.e. which accelerator should I use, etc.)

delock:
I ran a local workflow with Intel Extension for PyTorch uninstalled.  This seems work well.  So running current CPU_Accelerator without Intel Extension for PyTorch is possible.
https://github.com/delock/DeepSpeedSYCLSupport/actions/runs/7991757719/job/21823685139
> Thoughts on some actions we may possibly do:
> 
> 1. Check the AMD failure log to see if its something easy to fix
> 2. Test whether current CPU_Accelerator can run without Intel Extension for PyTorch installed
> 3. I don't have better suggestion for oneCCL binding for PyTorch ('ccl' communication backend), because this seem to be the only library capable of BF16 communication.  This would be important for training with CPU.
> 
> If these actions are effective, then maybe we don't need two seperate accelerator for CPU. This could make user less confusion (i.e. which accelerator should I use, etc.)

delock:
> Yes sure. Let's open an issue and discuss from there. This could simplify installation for many accelerators.
> 
> > > I think detect ops at launch time rather than installation time should be more proper solution, this can make deepspeed more stable to HW change. On the other hand, if user workflow require setting env variable, this would add additional complexity. It is better if environment variable overriding is used for troubleshooting only rather than being used as an interface. For example, if it's an interface, then change accelerator's name will have impact on existing scripts.
> > 
> > 
> > @delock, I fully agree with you on this. Can we work together to enable op detection at launch time?

https://github.com/microsoft/DeepSpeed/issues/5173 is created to track this topic.

PR #5128, opened 2024-02-13
Update nv-inference torch version

loadams:
We previously had the nv-inference tests pinned to use torch 1.13.1, this updates us to use a newer version.


loadams:
@mrwyattii - looks like we get basically the same test coverage, but we will lose I think our last test on 1.13 (and only the 1.10 tests remain not on PyTorch 2+).  Any thoughts on leaving this for coverage, or updating the 1.10 tests to 1.12/1.13?

mrwyattii (approved):

PR #5131, opened 2024-02-14
Run map reduce in one call

bm-synth:
In the map-reduce in data analysis, the run_reduce will merge several files into one. There are two open issues:

when running run_map followed by run_reduce, the run_reduce may start before all nodes finished the run_map, leading to having nodes loading files that are not populated/flush (zero-sized error)
when running run_reduce, all nodes are loading the partial result files output by all nodes, and all nodes will write the same files that result from the merge. This should be run by only one node, and all nodes should wait so that they can then feed the curriculum dataset to the deepspeed.initialize().

This PR fixes both by adding `dist.barrier`s to both steps (where barrier runs on an user-specifed `comm_group`), and by blocking non-master nodes to run the reduce operation.

### DIFF
--- a/deepspeed/runtime/data_pipeline/data_sampling/data_analyzer.py
+++ b/deepspeed/runtime/data_pipeline/data_sampling/data_analyzer.py
@@ -84,12 +84,14 @@ class DataAnalyzer(object):
                 metric_results.append({"metric_value": metric_value, "metric_value_fname": metric_value_fname})
         return metric_results
 
-    def update_metric_results(self, data, metric_types, metric_functions, metric_results):
+    def update_metric_results(self, data, metric_types, metric_dtypes, metric_functions, metric_results):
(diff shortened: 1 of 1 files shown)


conglongli (approved):
@bm-synth We will accept this solution instead of https://github.com/microsoft/DeepSpeed/pull/5130, because we still want to allow the map to be able to run on multiple machines that are not connected.

bm-synth:
> @bm-synth We will accept this solution instead of #5130, because we still want to allow the map to be able to run on multiple machines that are not connected. And would appreciate if you could fix the formatting issue for this PR. Thanks

@conglongli  makes sense.  Just pushed a commit that fixes formatting and follows the `pre-commit` warning by replacing
`import torch.distributed as dist` with ` import deepspeed.comm as dist`.

PR #5147, opened 2024-02-18
Bug fix for `batch_size`>1 with `metric_type=="accumulate_value_over_samples"`

bm-synth:
when running data analyzer with `metric_type == 'accumulate_value_over_samples'`, there's an add operation:

```
                if metric_result["metric_value"] is None:
                    metric_result["metric_value"] = metric_values
                else:
                    metric_result["metric_value"].add_(metric_values)
```

the shape of `metric_values` is a tensor of length that matches `self.batch_size`. When batch size greater than 1, this leads to strange behaviours when length of dataset does not multiply batch size:

Eg if we have `self.batch_size ==3`, the accumulated `metric_result['metric_value']` until the last data iteration is e.g: 
```
metric_result['metric_value']
tensor([3006, 3009, 3012])
```

if the last data batch, if size == 1, then operation `_add` is wrongly summing that element to all accumulators:
```
> metric_result["metric_value"].add_(  tensor([1004]) )
tensor([4010, 4013, 4016])
``` 

if size==2, then an error occurs:
```
metric_result["metric_value"].add_(torch.tensor([10,20]))
RuntimeError: The size of tensor a (3) must match the size of tensor b (2) at non-singleton dimension 0
```

And if size==3, the value written to the output file will be an array of size 3, not a single value which is the sum across all elements, as expected

This PR fixes that by summing over batch before doing that `_add`.


conglongli:
@bm-synth This metric_type is specifically for bincount the token frequency. See example at https://github.com/microsoft/Megatron-DeepSpeed/blob/d47f3cda3a9316ddc68e7f0ef904d1650ba6419d/examples_deepspeed/data_efficiency/analyze_data.py#L112. I think what this PR proposes is a different calculation.

bm-synth:
> @bm-synth This metric_type is specifically for bincount the token frequency.

Thank you @conglongli, I saw the example and I understand now. The confusion comes from the fact that you are reducing (sum) over batches and not over samples:
- I.e. you're doing `bincount(batch_1) +  bincount(batch_2) + ... + bincount(batch_B)`,
- instead of `bincount(sample_1) + bincount(sample_1) + ... + bincount(sample_S)`.

In your case it is equivalent for the `bincount` example, the sum-reduce and because you flatten the data (`data['text'].view(-1)`). This is a Megatron-specific operation.

For this to be deterministic across different batch sizes, and applicable to other use cases (not just `bincount`), then the output of `metric_values = metric_function(data)` should be:
- `[metric_function(sample_1), metric_function(sample_2), metric_function(sample_3)]`, for a batch of size 3, not
- `[metric_function(data)]`

Something like:
```
frequencies = [
    torch.bincount(
       text,
       minlength=args.padded_vocab_size+1,
       weights=mask)
       for text, mask in zip(data['text'], data['padding_mask').  ## <---- send one result per sample
    ]
```
and:
```
                elif metric_type == 'accumulate_value_over_samples':
                    for row in range(metric_values.size()[0]):  ## <---- accumulate over samples
                        if metric_results[m_idx] is None:
                            metric_results[m_idx] = row
                        else:
                            metric_results[m_idx].add_(row)
```

Does it make sense?

Anyway, I'm not using this reduce operation, so I'm fine with closing without changing.

conglongli:
@bm-synth I agree that adding `for row in range(metric_values.size()[0])` could make the metric type more general. On the other hand, for bincount if we change from per-batch to per-sample, it could greatly increase the processing time. So my suggestion is to handle this as a special case: if we detect that the output of `metric_values = metric_function(data)` is a 1-D tensor (such as bincount), we assume that this is a per-batch result.

bm-synth:
> if we detect that the output of metric_values = metric_function(data) is a 1-D tensor (such as bincount), we assume that this is a per-batch result.

@conglongli  this wouldn't fix it, because you wouldn't be able to tell, e.g. for a batch size 3:
- if `[0,1,2]` is the bin count of the batch for 3 bins (and you'd need to call `_add`), or 
- if `[0,1,2]` is the list of output values for each of the 3 input samples in the batch (and you'd call `.sum(dim=0)`. 

The only workaround is to have `metric_function` return a 2-D tensor and:
- if `len(tensor)==1` e.g. `tensor == [ [0,1,2] ]` then this is the metric value for the batch (or `batch_size==1`).
- if `len(tensor)>1` e.g. `tensor == [ [0], [1], [2] ]` this this is the metric value for each element of the batch.

I'll close this PR for now but feel free to reopen if this is something that you'd like to have in place and you're willing to make changes to the Megatron code to allow this.

Issue #5145, opened 2024-02-16
Training failing w/ certain parallelism degrees after PR #4974


lekurile:
**Describe the bug**
After the introduction of #4974, certain degrees of parallelism fail in the `scaled_global_norm` function:
https://github.com/microsoft/DeepSpeed/blob/a37e59b590a42d425ef962640762c99ced17b752/deepspeed/runtime/zero/stage_1_and_2.py#L1789C8-L1789C65

with the following error:
```bash
Traceback (most recent call last):
  File "megatron/pretrain_gpt.py", line 356, in <module>
    pretrain(train_valid_test_datasets_provider,
  File "megatron/megatron/training.py", line 226, in pretrain
    iteration = train(forward_step_func,
  File "megatron/megatron/training.py", line 1203, in train
    train_step(forward_step_func,
  File "megatron/megatron/training.py", line 669, in train_step
    loss = model[0].train_batch(data_iter=data_iterator)
  File "ds/deepspeed/runtime/pipe/engine.py", line 371, in train_batch
    self._exec_schedule(sched)
  File "ds/deepspeed/runtime/pipe/engine.py", line 1371, in _exec_schedule
    self._exec_instr(**cmd.kwargs)
  File "ds/deepspeed/runtime/pipe/engine.py", line 1173, in _exec_optimizer_step
    self._take_model_step(lr_kwargs)
  File "ds/deepspeed/runtime/engine.py", line 2073, in _take_model_step
    self.optimizer.step()
  File "ds/deepspeed/runtime/zero/stage_1_and_2.py", line 1827, in step
    scaled_global_grad_norm = self.scaled_global_norm()
  File "ds/deepspeed/runtime/zero/stage_1_and_2.py", line 1777, in scaled_global_norm
    return torch.norm(torch.stack(norm_groups), p=norm_type)
RuntimeError: stack expects each tensor to be equal size, but got [1] at entry 0 and [] at entry 1
```

The table below shows the specific parallelism degrees with failures:
| TP | PP | DP | Pass/Fail |
| -- | -- | -- | -- |
| 2 | 1 | 1 | PASS |
| 1 | 2 | 1 | PASS |
| 1 | 1 | 2 | PASS |
| 2 | 2 | 1 | PASS |
| 1 | 2 | 2 | FAIL |
| 2 | 1 | 2 | PASS |
| 2 | 2 | 2 | FAIL |

**To Reproduce**
To reproduce this behavior:
[...]

PR #5148, opened 2024-02-18
get_grad_norm_direct: fix a case of empty norm group

nelyahu:
fix for [#5145 ](https://github.com/microsoft/DeepSpeed/issues/5145)
empty norm group create a norm tensor with shape=[1], while other norms will be shapeless. torch.stack does not support such case. Fixing empty group norm to be shapless as well, instead of shape=[1].

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1686,7 +1686,7 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
             if len(all_norms) > 0:
                 total_norm = torch.stack(all_norms).square().sum().float()
             else:
-                total_norm = torch.FloatTensor([0.0]).to(self.device)
+                total_norm = torch.FloatTensor(0.0).to(self.device)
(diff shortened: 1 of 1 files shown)


lekurile (approved):

lekurile:
Hello @nelyahu,

Thanks for the PR, had to make a quick change from:
```python
total_norm = torch.FloatTensor(0.0).to(self.device)
```
to:
```python
total_norm = torch.tensor(0.0, dtype=torch.float32).to(self.device)
```
due to `torch.FloatTensor` not supporting initialization in that manner i.e.:
```bash
TypeError: new(): data must be a sequence (got float)
```

Tested the prallelism again and all looks good:
| TP	| PP	| DP	| Pass/Fail |
| -- | -- | -- | -- |
| 2	| 1	| 1	| PASS |
| 1	| 2	| 1	| PASS |
| 1	| 1	| 2	| PASS |
| 2	| 2	| 1	| PASS |
| 1	| 2	| 2	| PASS |
| 2	| 1	| 2	| PASS |
| 2	| 2	| 2	| PASS |

PR #5149, opened 2024-02-18
Add get optimizer method 

ShellyNR:
Add the get_optimizer method to accelerators for use during optimizer configuration, instead of checking for the specific accelerator in the [engine.py](http://engine.py/) code. 

In all accelerators other than hpu the current implementation returns None so that the previous flow is not affected.

In hpu_accelerator.py, the method returns an optimizer if the configured optimizer has hpu-specific implementation or requirements. Otherwise, it returns None.


ShellyNR:
> @ShellyNR please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 

accelerator/hpu_accelerator.py
```diff
@@ -294,3 +294,22 @@ def build_extension(self):
 
     def export_envs(self):
         return []
+
+    def get_optimizer(self, optimizer_name, cpu_optimization, model_parameters, **optimizer_parameters):
+        from habana_frameworks.torch.hpex.optimizers import FusedAdamW
```
tjruwase, line 299: I am not convinced that this much code is needed. 

It seems the goal is to replace AdamW on hpu accelerator as below:

`from deepspeed.ops.adam import FusedAdam` with 
`from habana_frameworks.torch.hpex.optimizers import FusedAdamW`

Is this correct?

nelyahu: @tjruwase  Yes, but in addition to prepare the ground for other fused optimizers that are addressed in this function. for example: FusedLamb, OneBit, etc...
Do you have in mind something else?

tjruwase: @nelyahu, thanks for the clarification. I do agree that deferring to accelerator to instantiate fused optimizer when available is the way to go. My concern is the duplication of optimizer selection logic here with `engine.py` will be difficult to maintain long term. 

I think the complication is that `engine.py` mixes optimizer name refinement (especially for Adam variants) and optimizer instantiation. By name refinement, I mean something like [ADAM_OPTIMIZER](https://github.com/microsoft/DeepSpeed/blob/aed599b4422b1cdf7397abb05a58c3726523a333/deepspeed/runtime/config.py#L72C1-L72C16) which is a user-facing config value, internally maps to one of many optimizers (e.g., `torch.optim.Adam`, `DeepSpeedCPUAdam`, etc.). So, my thought is to 
1. Create an internal naming convention for the various optimizer instantiations. For example, 

- `_TORCH_ADAM_OPTIMIZER` for `torch.optim.Adam`, 
- `_DS_FUSED_ADAM` for `deepspeed.ops.adam.FusedAdam`,
- `_DS_ADAM_OPTIMIZER` for `deepspeed.ops.adam.DeepSpeedCPUAdam`.  

2. Create a function that maps an external name to an internal one based on other configuration values.  For example, [ADAM_OPTIMIZER](https://github.com/microsoft/DeepSpeed/blob/aed599b4422b1cdf7397abb05a58c3726523a333/deepspeed/runtime/config.py#L72C1-L72C16) would map to one of `_TORCH_ADAM`, `_DS_FUSED_ADAM`, `_DS_CPUADAM`.

Based on the above the following changes could now be made:
1. Simplify `get_optimizer()` by accepting internal optimizer name and creating an optimizer if name matches its table of supported internal optimizer names. Also, `cpu_optimization` is no longer needed. 
2. Simplify [`_configure_basic_optimizer()`](https://github.com/microsoft/DeepSpeed/blob/aed599b4422b1cdf7397abb05a58c3726523a333/deepspeed/runtime/engine.py#L1276) to (i) convert to internal name, (ii) call `accelerator().get_optimizer()` with internal name to create optimizer, and (iii) otherwise fall through to existing optimizer creation logic. 

Looking forward to your feedback. Thanks!

ShellyNR: @tjruwase I applied the changes you suggested, however some optimizers still require specific code. Which one of the commits you think is best?

tjruwase: @ShellyNR, thanks for making the changes. This looks better to me. 

When you say 'some optimizers still require specific code`, is that referring to `_ADAM` in `hpu_accelerator` which maps to `torch.optim.Adam()`? If so, I think that is fine. My goal is to avoid duplication between accelerator codes and `engine.py`.

tjruwase:
@ShellyNR, thanks for your work on this PR. I wanted to check if you needed more clarification to address my comments?

nelyahu:
@tjruwase , sorry for not updating this PR, @ShellyNR was OOO for few days. we are considering another direction for support FusedAdam via existing HPU implementation.

tjruwase:
> @tjruwase , sorry for not updating this PR, @ShellyNR was OOO for few days. we are considering another direction for support FusedAdam via existing HPU implementation.

@nelyahu, thanks for the update. I am excited to see the new approach :).

ShellyNR:
I'm closing this PR, a new PR that addresses this issue without using a get_optimizer method will be pushed soon.
@tjruwase thanks for your review and comments (:

PR #5150, opened 2024-02-19
Fix gradient clipping

tohtana:
The gradient clipping API doesn't apply the coefficient correctly. This PR resolves the issue and adds a test case.

### DIFF
--- a/deepspeed/runtime/utils.py
+++ b/deepspeed/runtime/utils.py
@@ -407,7 +407,7 @@ def clip_grad_norm_(parameters, max_norm, norm_type=2, mpu=None):
     max_norm = torch.tensor([float(max_norm)], device=parameters[0].device)
     clip_coef = max_norm / (total_norm + 1e-6)
     tmp_tensor = torch.tensor([1.0], device=parameters[0].device)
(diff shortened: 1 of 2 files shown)


loadams (approved):

cloneofsimo:
I'm actually fascinated how did this bug go unnoticed for so long, so LONG? isn't grad clip on by default? does this mean literally all of huggingface trainer, lightning etc that leverages Deepspeed backend was faulty, yet somehow everyone succeeded on training? like HOW? is this not-as-critical?

cloneofsimo:
@tohtana thanks for clarification! According to the codebase this gets used for all engine step... so I think it effects all of deepspeed but i hope I am wrong and this pr fixes everything thank you for this!

SeunghyunSEO:
@tohtana ty for your kind explanation. my understanding is that when zero3 (or 1,2) is activated, it has no problem because we use [this function](https://github.com/microsoft/DeepSpeed/blob/5d75460640aa8f8a67baadaea23783ec708e8e1b/deepspeed/runtime/zero/stage3.py#L2108-L2117) to clip and scale the gradient, right?

PR #5154, opened 2024-02-19
ZeRO0 does not handle BF16 gradients properly

tohtana:
The combination of BF16 and ZeRO0 (no ZeRO optimization) has some issues with handling gradients. `BF16_Optimizer` seems to be designed to accumulate gradients in FP32, but this doesn't match other parts.

- DeepSpeed engine converts BF16 gradients and accumulate them in FP32 soon after the backward pass using `BF16_Optimizer`. However, the engine performs allreduce on BF16 gradients after that. As the result, gradients are not properly reduced.
- The engine calls `BF16_Optimizer`'s `backward()`. This clears BF16 gradients. So it doesn't work when the gradient accumulation step > 1.

There are two possible approaches to resolve this issue:
1. Accumulate gradients in BF16 until the boundary of gradient accumulation steps (We should clear BF16 gradient only at the gradient accumulation boundary). Then perform allreduce, conversion to FP32, and parameter update.
2. Accumulate gradients in FP32 and run allreduce.

This PR takes the first approach to resolve the issue. ZeRO 1/2/3 follows the first approach though the second one has the advantage in terms the precision for gradient accumulation.

### DIFF
--- a/deepspeed/runtime/bf16_optimizer.py
+++ b/deepspeed/runtime/bf16_optimizer.py
@@ -292,9 +292,6 @@ class BF16_Optimizer(ZeROOptimizer):
         self.clear_lp_grads()
         loss.backward(**bwd_kwargs)
 
-        if update_hp_grads:
-            self.update_hp_grads(clear_lp_grads=clear_lp_grads)
-
     @torch.no_grad()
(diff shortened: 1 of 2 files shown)


tohtana:
This PR breaks PP. Opened #5170 as another solution.

PR #5156, opened 2024-02-19
MOE gate fixes and enhancements

mosheisland:
Fixes the following issues:
- Fix capacity when using TP for non-MoE by aligning the capacity to TP
- Fix TopKGate.wg (gate weight) when using ZeRO with fp16 or bf16
- Fix top2 aux loss to be similar to top1 aux loss

Following are few configurable enhancements:
- Support top2 with disable token dropping
- Support disable top2 2nd expert sampling


deepspeed/moe/sharded_moe.py
```diff
@@ -372,13 +393,14 @@ def __init__(self,
...
-                 use_rts: bool = True) -> None:
+                 use_rts: bool = True,
+                 top2_2nd_expert_sampling: bool = True) -> None:
         super().__init__()
 
         # Only top-1 and top-2 are supported at the moment.
         if k != 1 and k != 2:
             raise ValueError('Only top-1 and top-2 gatings are supported.')
-        self.wg = torch.nn.Linear(model_dim, num_experts, bias=False).float()
+        self.wg = torch.nn.Linear(model_dim, num_experts, bias=False)
```
awan-10, line 403: @ykim362  - can you please review this part? I remember we forced .float() here because we wanted the gate weight to always be fp32 even if everything else was fp16.

deepspeed/moe/sharded_moe.py
```diff
@@ -305,17 +314,29 @@ def top2gating(logits: Tensor, capacity_factor: float, min_capacity: int) -> Tup
...
+    if drop_tokens:
+        # Calculate configured capacity and remove locations outside capacity from mask
+        capacity = _capacity(gates, torch.tensor(capacity_factor * 2), torch.tensor(min_capacity))
+        mask1 *= torch.lt(locations1, capacity)
+        mask2 *= torch.lt(locations2, capacity)
+    else:
+        # Do not drop tokens - set capacity according to current expert assignments
+        new_capacity = torch.max(exp_counts)
+        dist.all_reduce(new_capacity, op=dist.ReduceOp.MAX, group=dist.get_world_group())
+        if groups._get_expert_model_parallel_world_size() == 1:
```
awan-10, line 334: How is this check going to check for if non-expert part is tensor-parallel? Shouldn't there be an if model_parallel_world_size > 1 check or something similar?

deepspeed/moe/sharded_moe.py
```diff
@@ -305,17 +314,29 @@ def top2gating(logits: Tensor, capacity_factor: float, min_capacity: int) -> Tup
...
     locations2 += torch.sum(mask1, dim=0, keepdim=True)
 
-    # gating decisions
-    exp_counts = torch.sum(mask1, dim=0).detach().to('cpu')
-
     # Compute l_aux
     me = torch.mean(gates, dim=0)
     ce = torch.mean(mask1.float(), dim=0)
-    l_aux = torch.mean(me * ce) * num_experts * num_experts
+    l_aux = torch.mean(me * ce) * num_experts
```
awan-10, line 320: @ykim362  - FYI. I think the current value of l_aux in top-2 was giving us good convergence. I am not sure if we change it, we will need to run training experiments to verify any regression in loss. 

@mosheisland - have you trained the models using top-2 and you see that l_aux with your change gives better convergence/loss?

awan-10:
@mosheisland -- Thank you very much for working on this PR! Overall, it looks good to me but given that we don't have the bandwidth to test the convergence, it will really help us and users if you can do two things:

1. Protect any changes to l_aux, capacity padding, etc. with user-configurable flags and set their default values to be False.
      - e.g. add a flag like pad_capacity=False as a default and then only do this if pad_capacity=True is passed by the user. This is to avoid any regression as many users are using this in the current form. If we don't know if new code is going to give better results or previous code, at least, we should have the option to have new things disabled by default.

2. Share any before your changes and after your changes convergence plots for any model that you have trained with new changes. That data will help the users to refer back to this PR in case setting pad_capacity=True or using new l_aux value will change any training/loss-curves for them.

deepspeed/moe/sharded_moe.py
mosheisland: I did not run full training before and after my changes and compare.
I added this change to better aligns to original paper.
Since you saw better results with the current formulation, lets keep it as-is.
I will upload a new commit that reverts this change.
BTW, in the case of using same num_experts across all MOE layers, this can be modified by changing the ----moe-loss-coeff.

deepspeed/moe/sharded_moe.py
mosheisland: We pad (align) the calculated capacity to tp.
If non-expert part is not using tensor-parallel, then tp=1 --> align to 1 --> no alignment required
If non-expert part has tp>1, then we align to tp

mosheisland:
> 1. Protect any changes to l_aux, capacity padding, etc. with user-configurable flags and set their default values to be False.
   
>    * e.g. add a flag like pad_capacity=False as a default and then only do this if pad_capacity=True is passed by the user. This is to avoid any regression as many users are using this in the current form. If we don't know if new code is going to give better results or previous code, at least, we should have the option to have new things disabled by default.

l_aux - I will revert this change.

capacity padding - this bug is reproduced when you use drop-tokens=False and non-expert-tp>1.
The padding itself is with 0-ed rows and has no real effect on training.

> 2. Share any before your changes and after your changes convergence plots for any model that you have trained with new changes. That data will help the users to refer back to this PR in case setting pad_capacity=True or using new l_aux value will change any training/loss-curves for them.

l_aux - no need, I will revert
capacity padding - I can't share before the fix since it crashes

mosheisland:
Reverted l_aux commit

awan-10 (approved):

PR #5161, opened 2024-02-19
Added support for skipping Inference V2 tests when the accelerator is not compatible.

vshekhawat-hlab:
Added support for skipping Inference V2 tests when the accelerator is not compatible.

This commit implements the capability to skip Inference V2 tests in scenarios where the accelerator does not support Inference V2. A condition check has been added to ensure the tests are bypassed when the hardware configuration lacks support for Inference V2, mitigating potential issues.

Details:
- Implemented a condition check to assess accelerator compatibility with Inference V2.
- In cases of non-support, Inference V2 tests will be skipped, enhancing the reliability of the test suite.
- This enhancement aims to improve compatibility and stability across diverse hardware configurations.


tests/unit/inference/v2/inference_test_utils.py
```diff
@@ -44,3 +44,10 @@ def allclose(x, y, tolerances: Tuple[int, int] = None):
     else:
         rtol, atol = tolerances
     return torch.allclose(x, y, rtol=rtol, atol=atol)
+
+
+def skip_on_inference_v2():
```
loadams, line 49: Hi @vshekhawat-hlab - we already have a pytest mark for this [here](https://github.com/microsoft/DeepSpeed/blob/857584fd0c8865efaa65285ff4772b3e6142f6af/tests/pytest.ini#L7C1-L8C47) and it is already not enabled by default [here](https://github.com/microsoft/DeepSpeed/blob/857584fd0c8865efaa65285ff4772b3e6142f6af/tests/pytest.ini#L2) - can you let us know what you're trying to accomplish with this PR, or perhaps what pytest command you are running that is causing these tests to run on an HPU?

cc: @mrwyattii

vshekhawat-hlab: HI @loadams ,

On HPU, we have enabled all markers before execution. Here, I was trying to add skip checks based on devices. Yes, you are correct that these tests are disabled by default. I will fix execution command. We can close this PR.

PR #5165, opened 2024-02-20
Revert "Fix UserWarning: The torch.cuda.*DtypeTensor constructors are…

loadams:
Reverts `177dc14331a64e61f6dcce2c4b8071576bcb22db` since it breaks tests in Megatron-DeepSpeed.

### DIFF
--- a/accelerator/cuda_accelerator.py
+++ b/accelerator/cuda_accelerator.py
@@ -3,7 +3,6 @@
 
 # DeepSpeed Team
 
-import functools
 import os
 import pkgutil
 import importlib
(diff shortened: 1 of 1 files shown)


lekurile:
Hello @ShukantPal,

Appreciate the effort on the initial PR fixing the warning. However, this change causes an issue in the Megatron-DeepSpeed type check code here:
https://github.com/microsoft/Megatron-DeepSpeed/blob/d47f3cda3a9316ddc68e7f0ef904d1650ba6419d/megatron/model/module.py#L138
 
Here's the resulting error:
```bash
 File "/Megatron-DeepSpeed/megatron/model/module.py", line 139, in half_conversion
    if isinstance(val_typecheck, _FLOAT_TYPES):
TypeError: isinstance() arg 2 must be a type or tuple of types
```

Since the CUDA accelerator no longer returns a torch type, but `functools.partial`:
```bash
(Pdb) p _FLOAT_TYPES
(<class 'torch.FloatTensor'>, functools.partial(<built-in method tensor of type object at 0x7f3d941c5420>, dtype=torch.float32, device='cuda'))
```

We an issue when training using the `Megatron-DeepSpeed` repo. For now, we're looking to revert this PR and would appreciate any feedback or suggestions from your end.

Thanks,
Lev

lekurile (approved):

ShukantPal:
Hi @lekurile,

The Megatron-DeepSpeed repo would probably have to change how it detects tensor types to handle this. Probably by calling the tensor factory and checking its `dtype`; something like this:

```python
        if callable(val_typecheck) and val_typecheck([0]).dtype == torch.float:
            val = float16_convertor(val)
        return val
```

loadams:
Thanks @ShukantPal - closing in favor of @lekurile's PR in Megatron-DeepSpeed [here](https://github.com/microsoft/Megatron-DeepSpeed/pull/346).

PR #5182, opened 2024-02-23
Switch cpu-inference workflow from --extra-index-url to --index-url

loadams:


lekurile (approved):

loadams:
@delock - can you or your team register the oneccl_bind_pt package name on PyPI so that if others have this vulnerability they will be protected from malicious actors as well?

delock:
@loadams thanks for the suggestion.  I'll check with the team on this potential vulerability.

PR #5193, opened 2024-02-26
Fix param.ds_id not found bug when call 'deepspeed.runtime.swap_tensor.optimizer_utils.OptimizerSwapper._create_param_swap_info

getinglxf:
Fix param.ds_id not found bug when call 'deepspeed.runtime.swap_tensor.optimizer_utils.OptimizerSwapper._create_param_swap_info' method under condition 
 that 'offload_optimizer' and 'offload_param' set to nvme path in zero_config. 
In new version of Deepspeed(0.13.3), the deepspeed.runtime.swap_tensor.optimizer_utils.OptimizerSwapper has added 'parameter_id(param)' method to get param id, while in older version like 0.12.4 the origin method is 'id(param)'. This change causes a new bug: when call deepspeed.runtime.zero.stage3.DeepSpeedZeroOptimizer_Stage3._create_fp32_partitions method in init process，the 'self.optimizer_swapper.initialize_paraeters(parameters=[self.fp32_partitioned_groups_flat[i]], src_tensors=[unpinned_fp32_buffer])' method will use unpinned_fp32_buffer without ds_id under optimizer offload condition, which will throw a error : "AttributeError: 'Tensor' object has no attribute 'ds_id'"
ds_id error


deepspeed/runtime/zero/stage3.py
```diff
@@ -840,6 +843,7 @@ def _create_fp32_partitions(self):
                     else:
                         unpinned_fp32_buffer = torch.empty(num_elements, device=self.device, dtype=torch.float)
                         self._swap_in_sub_group_to_flat_buffer(unpinned_fp32_buffer, i)
+                        self.fp32_partitioned_groups_flat[i].ds_id = ds_id
```
tjruwase, line 846: There is currently redundancy between L846 and L872. To fix this, how about move L846 to after L831 for the `if ...` and move L846 into the `else ...`

loadams:
Closing due to fix being applied in #5824

PR #5167, opened 2024-02-21
Pydantic v2 migration

mrwyattii:
Pydantic v2 has been out for some time now. We have been relying on using the v1 API available in v2 until now. This is a refresh of #3902 to bring proper v2 support to DeepSpeed.

Corresponding DeepSpeed-MII PR [here](https://github.com/microsoft/DeepSpeed-MII/pull/423).

@loadams

PR #5204, opened 2024-02-27
Update container image in a6000 test

loadams:
The existing image used here (pytorch:23.03-py3) has versions of thinc (8.1.9), spacy (3.5.1), and confection (0.0.4) that do not support pydantic > 1.11.0.

We test updating to a newer image (pytorch:23.09-py3) that has newer versions of thinc (8.1.12), spacy (3.6.1), and confection (0.1.3).  These versions all have pydantic < 3.0.0 in their requirements.txt, so we should be safe to update to this version.


.github/workflows/nv-a6000.yml
```diff
@@ -23,7 +23,7 @@ jobs:
   unit-tests:
     runs-on: [self-hosted, nvidia, a6000]
     container:
-      image: nvcr.io/nvidia/pytorch:23.03-py3
```
loadams, line 26: This can be updated after #5167 completes.

loadams:
Re-open and complete this PR later if needed.

Issue #5173, opened 2024-02-22
[REQUEST] detect opbuilder list at launch time


delock:
**Is your feature request related to a problem? Please describe.**
Currently, DeepSpeed detects opbuilder list at installation time and put all installed ops inside `git_version_info_installed.py`.  This often works except in situation that there is HW change between installation and launch which require change of DeepSpeed Accelerator.   A typical situation is when cluster user install DeepSpeed on login node without GPU, then launch DeepSpeed on compute node with GPU.

Currently, this situation is mitigated by using Torch extension as an indicator of which Accelerator to be used.  When CPU version of Intel Extension for PyTorch is installed in the environment, DeepSpeed would deem its for CPU no matter whether a GPU device is installed on the system.  Otherwise DeepSpeed would decide which Accelerator to use depending on extension installation.  If no extension is installed (stock PyTorch), DeepSpeed would assume install and launch for CUDA_Accelerator.

Yet this is not ideal solution, this complicates DeepSpeed deploy and execution (i.e. must install PyTorch extension before install DeepSpeed).

**Describe the solution you'd like**
Ideally, DeepSpeed should adapt to HW change and act properly.  Accelerator should be tracked in git_version_info_installed.py and updated up on HW change.   A simple (and probably sufficient) solution would be discard existing info upon HW (Accelerator) change and regenerate them according to current Accelerator being used.

This change could be combine with Accelerator selection logic in `real_accelerator.py` that when using stock PyTorch (no extension), decide whether to use CPU Accelerator depending on whether CUDA device is used.   This could possibly be an alternative to PR#5084 which allows a single CPU Accelerator serve both stock PyTorch and Intel Extension for PyTorch.

delock:
Hi @tjruwase this is the request to track opbuilder detection.

delock:
One thing we should discuss is how to test this feature in a workflow.  My suggestion is as follow:
1. Install DeepSpeed on a CUDA environment with accelerator override to 'cpu', so we make sure DeepSpeed is installed with less registered op builder
2. Check `git_version_info_installed.py` to see op builder list
3. Run DeepSpeed workload involve CUDA kernel (i.e. inference with kernel injection) and observe its behavior
4. Check `git_version_info_installed.py` to see updated op builder list

This test can be a seperate workflow or embed into current nv inference workflow.

@tjruwase

delock:
Hi @tjruwase, see replies below.
[...]

PR #5206, opened 2024-02-28
Make op builder detection adapt to accelerator change

delock:
This is an WIP PR that make op builder detection adapt to accelerator change.  This is followup of https://github.com/microsoft/DeepSpeed/issues/5173
Currently, DeepSpeed generate `installed_ops` and `compatible_ops` at setup time.  If the system change to a different accelerator at DeepSpeed launch time, these two list would contain incorrect information.

This PR intend to solve this problem with more flexity ops detection.

* For `installed_ops`, DeepSpeed should disable all installed ops if accelerator detected at setup time is different from launch time.
* For `compatible_ops`, DeepSpeed should refresh the list for each launch to avoid impact of accelerator change.

In the first step, nv-inference workflow is temporary change to emulate the scenario that the system is setup with CPU_Accelerator, then launch with CUDA_Accelerator.  And CPU_Accelerator is modified to make Intel Extension for PyTorch and oneCCL binding for PyTorch not mandatory.

Starting from here we can reconstruct installed_ops and compatible_ops to follow the design above.


delock:
@tjruwase @mrwyattii @loadams FYI

Note this PR also comes with an updated CPU_Accelerator that could work without Intel Extension for PyTorch and oneCCL Binding for PyTorch.   It is designed in a way that is adaptive to whether these two packages are installed.  Hope this can help us keep one CPU_Accelerator that serve both basic and optimized user scenario.

delock:
@tjruwase `installed_ops` and `compatible_ops` are now reconstructed as this PR description.  Also removed Intel Extension for PyTorch and oneCCL binding for PyTorch installation in CPU inference workflow to test whether CPU Accelerator can work with stock PyTorch.  Can you help start the workflow?  Thanks!

delock:
Hi @tjruwase , the failure in cpu-inference and nv-inference workflow is due to missing package.  I have fixed them with latest CI.  Can you help start the workflow?  Thanks!

delock:
@tjruwase the missing dependency in nv-inference workflow had been added.

delock:
The latest error in cpu inference needs some investigation.   When CPU Accelerator is selected, some package will be used before they are installed with requirements.

delock:
Also in nv-pre-compile-ops workflow.  This command would diverge from intended purpose because it is running on a CPU node.  CPU ops instead of CUDA ops will be installed.  This might not be what users wanted and might break some users environment who use pre-built CUDA kernels.

We will need ask user to use DS_ACCELERATOR=cuda in this situation.  @tjruwase 

DS_ENABLE_NINJA=1 TORCH_CUDA_ARCH_LIST="7.0;7.5;8.0" DS_BUILD_OPS=1 DS_BUILD_SPARSE_ATTN=0  DS_BUILD_CUTLASS_OPS=0 DS_BUILD_RAGGED_DEVICE_OPS=0 DS_BUILD_EVOFORMER_ATTN=0 pip3 install .

tjruwase:
> The latest error in cpu inference needs some investigation. When CPU Accelerator is selected, some package will be used before they are installed with requirements.

I don't understand why this failure happens since `psutil` is in requirements.txt
image

tjruwase:
> Also in nv-pre-compile-ops workflow. This command would diverge from intended purpose because it is running on a CPU node. CPU ops instead of CUDA ops will be installed. This might not be what users wanted and might break some users environment who use pre-built CUDA kernels.
> 
> We will need ask user to use DS_ACCELERATOR=cuda in this situation. @tjruwase
> 
> DS_ENABLE_NINJA=1 TORCH_CUDA_ARCH_LIST="7.0;7.5;8.0" DS_BUILD_OPS=1 DS_BUILD_SPARSE_ATTN=0 DS_BUILD_CUTLASS_OPS=0 DS_BUILD_RAGGED_DEVICE_OPS=0 DS_BUILD_EVOFORMER_ATTN=0 pip3 install .

Yes, I agree with this. Can you please make the change in the PR. I think the `nv-` prefix of the workflow also indicates nvidia-specific.

delock:
The reason is CPU_Accelerator.__init__() use psutil and this is before setup.py install the requirements.  The latest code would check importError before using psutil, this works in my local environment.
> > The latest error in cpu inference needs some investigation. When CPU Accelerator is selected, some package will be used before they are installed with requirements.
> 
> I don't understand why this failure happens since `psutil` is in requirements.txt image

delock:
@tjruwase I have question on nv-torch-latest-cpu.   It looks like the intention of this test is test CUDA_Accelerator's compatibility when running on CPU only machine.  But instead CPU_Accelerator is selected.   What is user's expectation in this scenario?

loadams:
> @tjruwase I have question on workflow nv-torch-latest-cpu. It looks like the intention of this test is test CUDA_Accelerator's compatibility when running on CPU only machine. But instead CPU_Accelerator is selected. What is user's expectation in this scenario?

This workflow was part of what spurred the original discussion, but this should run on a CPU only machine but if the user has cuda installed for cpu only if they wanted to run that way, so if they installed cuda 11.6+cpu for example.

@mrwyattii - we could probably rename this file to be cpu-torch-latest since this doesn't need the nv prefix?

tjruwase:
> This workflow was part of what spurred the original discussion, but this should run on a CPU only machine but if the user has cuda installed for cpu only if they wanted to run that way, so if they installed cuda 11.6+cpu for example.
> 
> @mrwyattii - we could probably rename this file to be cpu-torch-latest since this doesn't need the nv prefix?

@loadams is absolutely correct. This workflow is meant for cpu-only environment. The `nv` prefix should be dropped from the name.

loadams:
> > This workflow was part of what spurred the original discussion, but this should run on a CPU only machine but if the user has cuda installed for cpu only if they wanted to run that way, so if they installed cuda 11.6+cpu for example.
> > @mrwyattii - we could probably rename this file to be cpu-torch-latest since this doesn't need the nv prefix?
> 
> @loadams is absolutely correct. This workflow is meant for cpu-only environment. The `nv` prefix should be dropped from the name.

PR for this is here: https://github.com/microsoft/DeepSpeed/pull/5226

delock:
The failure in cpu-torch-latest had been fixed.  On my local environment I observe 127 passes.  @loadams can you help start the workflow?  Thanks!

loadams:
> The failure in cpu-torch-latest had been fixed. On my local environment I observe 127 passes. @loadams can you help start the workflow? Thanks!

@delock - Looks like it is now failing for numactl issues?  I can't quite tell, but I think that is probably just something that needs to be added in the .yml setup.

umchand:
FYI, I had to install numactl because I ran into this error, numactl: command not found

delock:
@tjruwase @loadams @mrwyattii There are two choices on when to fallback to gloo communican backend in Pytorch.

Current solution is choose `gloo` backend in CPU Accelerator if oneCCL Binding for PyTorch is not installed.   In this case `CCLCommBuilder` will not be used.  In this case, DeepSpeed would use default PyTorch communication backend for all communications.  If oneCCL Binding for PyTorch is installed, current CCLBackend would be used which would call CCL comm kernels.

A different solution is in `ccl.py` use gloo as base communication backend if oneCCL Binding for PyTorch is not installed, but still build communication kernels from `CCLCommBuilder`.  In this case we need to install oneCCL library as indicated in 'cpu-inference' workflow and would be able to use these communication kernels written in C++.

The first option has less dependency but would have very low communication performance especially for `inference_all_reduce`.  The second option has better performance because we can leverage oneCCL performance, but will need install additional library on the system.  On a Xeon 2 socket machine, this difference can be up to 15x.  For some inference use case i.e. inference with AutoTP there will be big performance impact.

What's your preference on this?

delock:
The first solution code change is here. https://github.com/microsoft/DeepSpeed/pull/5206/files#diff-b4d7bc897d0a7ddbeda388ae8960d4d938f56dd121f3a269bdef650b2f201e8dR27

The second solution will be to change the backend to 'gloo' in here https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/comm/ccl.py#L42 if oneCCL Binding for PyTorch is not installed.

delock:
The reason for cpu-torch-latest is due  to most UTs assume fp16 data type that CPU accelerator does not support.   I changed some of the tests to use bf16 data type if accelerator support bf16 but not fp16.  It would take a while to make all the UT pass.

@loadams can you help restart the workflow?  I'll push more test changes but it would take a while.

tjruwase:
> The reason for cpu-torch-latest is due to most UTs assume fp16 data type that CPU accelerator does not support. I changed some of the tests to use bf16 data type if accelerator support bf16 but not fp16. It would take a while to make all the UT pass.

@delock, thanks for helping with this. The root cause is the following common construct in the UT.  
```python
    "fp16": {"enabled": True}
```

Rather than skipping these tests, I think the best solution is to modify to
```python
    "fp16": { "enabled": get_accelerator().is_fp16_supported()}
```

What do you think? 

@umchand, can you please try this idea in your environment? Thanks!

delock:
Also the  error in https://github.com/microsoft/DeepSpeed/actions/runs/8189457704/job/22394285394?pr=5206 indicates using `is_fp16_supported()` may cause lazy init issue on CUDA device in some UTs,  Some special handling will be needed.
https://github.com/microsoft/DeepSpeed/actions/runs/8189457704/job/22394285394?pr=5206

umchand:
> > The reason for cpu-torch-latest is due to most UTs assume fp16 data type that CPU accelerator does not support. I changed some of the tests to use bf16 data type if accelerator support bf16 but not fp16. It would take a while to make all the UT pass.
> 
> @delock, thanks for helping with this. The root cause is the following common construct in the UT:
> 
> ```python
>     "fp16": {"enabled": True}
> ```
> 
> Rather than skipping these tests, I think the best solution is to modify to
> 
> ```python
>     "fp16": { "enabled": get_accelerator().is_fp16_supported()}
> ```
> 
> What do you think?
> 
> @umchand, can you please try this idea in your environment? Thanks!

FYI, I tried this change suggested by @tjruwase and did see the 4 test which were failing pass.

delock:
> > > The reason for cpu-torch-latest is due to most UTs assume fp16 data type that CPU accelerator does not support. I changed some of the tests to use bf16 data type if accelerator support bf16 but not fp16. It would take a while to make all the UT pass.
> > 
> > 
> > @delock, thanks for helping with this. The root cause is the following common construct in the UT:
> > ```python
> >     "fp16": {"enabled": True}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Rather than skipping these tests, I think the best solution is to modify to
> > ```python
> >     "fp16": { "enabled": get_accelerator().is_fp16_supported()}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you think?
> > @umchand, can you please try this idea in your environment? Thanks!
> 
> FYI, I tried this change suggested by @tjruwase and did see the 4 test which were failing pass.

Thanks! @umchand @tjruwase.  I have changed more UTs, can you help start the workflow?

loadams:
> > > > The reason for cpu-torch-latest is due to most UTs assume fp16 data type that CPU accelerator does not support. I changed some of the tests to use bf16 data type if accelerator support bf16 but not fp16. It would take a while to make all the UT pass.
> > > 
> > > 
> > > @delock, thanks for helping with this. The root cause is the following common construct in the UT:
> > > ```python
> > >     "fp16": {"enabled": True}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Rather than skipping these tests, I think the best solution is to modify to
> > > ```python
> > >     "fp16": { "enabled": get_accelerator().is_fp16_supported()}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > What do you think?
> > > @umchand, can you please try this idea in your environment? Thanks!
> > 
> > 
> > FYI, I tried this change suggested by @tjruwase and did see the 4 test which were failing pass.
> 
> Thanks! @umchand @tjruwase. I have changed more UTs, can you help start the workflow?

Running now

delock:
@loadams the error in nv-inference is quite strange, `preferred_dtype` is a variable defined in unit/common.py

Maybe define it as a variable is not a good practice.  I changed it to a function instead.  Can you help restart the workflow?  Thanks!

delock:
@loadams The remaining 4 test failure is due to pdsh command not found.  pdsh had not been validated for CPU Accelerator, where impi launcher is recommended for multi-node.  Nevertheless, I installed pdsh in this workflow and see what happens.  Can you help restart the workflow?  Thanks!

delock:
Hi,
@tjruwase @loadams, from the error message, this machine had not  been configured for multi-node runner.  This machine needs host key verification (ssh with authenticated keys).  Otherwise we have to skip this test for CPU accelerator

```
    def test_user_args(cmd):
        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
        out, err = p.communicate()
>       assert "ARG PARSE SUCCESS" in out.decode("utf-8"), f"User args not parsed correctly: {err.decode('utf-8')}"
E       AssertionError: User args not parsed correctly: localhost: Host key verification failed.

E         pdsh@fv-az843-752: localhost: ssh exited with exit code 255
E         /home/runner/work/DeepSpeed/DeepSpeed/unit-test-venv/lib/python3.8/site-packages/transformers/utils/hub.py:124: FutureWarning: Using `TRANSFORMERS_CACHE` is deprecated and will be removed in v5 of Transformers. Use `HF_HOME` instead.
E           warnings.warn(
E         
```

delock:
@tjruwase there is also some UT skipped with CPU accelerator.  Each one may need further investigation  Will need some help looking at the real reason of the failure and fix them one by one.  Should we put this work into seperate PRs and keep these skips with CPU accelerators?

tests/unit/checkpoint/test_zero_optimizer.py
```diff
@@ -336,8 +328,8 @@ def test_immediate_save_load(self, tmpdir, zero_stage):
 
     @pytest.mark.parametrize('zero_stage', [0, 1, 2, 3])
     def test_load_immediate_save(self, tmpdir, zero_stage):
-        if not get_accelerator().is_fp16_supported():
-            pytest.skip("fp16 is not supported")
+        if zero_stage == 0 and get_accelerator().device_name() == "cpu":
```
tjruwase, line 331: Why special case of `zero_stage == 0`?

tjruwase:
> @tjruwase there is also some UT skipped with CPU accelerator. Each one may need further investigation Will need some help looking at the real reason of the failure and fix them one by one. Should we put this work into seperate PRs and keep these skips with CPU accelerators?

Yes, it is okay to put that work into a separate PR.

tests/unit/checkpoint/test_zero_optimizer.py
```diff
@@ -161,8 +153,8 @@ def test_hybrid_optimizer_state(self, tmpdir, zero_stage):
 
     @pytest.mark.parametrize('zero_stage', [0, 1, 2, 3])
     def test_load_module_only(self, tmpdir, zero_stage):
-        if not get_accelerator().is_fp16_supported():
-            pytest.skip("fp16 is not supported")
+        if zero_stage == 0 and get_accelerator().device_name() == "cpu":
```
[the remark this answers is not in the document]
delock: The root cause needs more investigation.  It seems that the failure for zero 0 relates to allgather behavior.

The skip clauses needs to  be seperated into three groups:
1. the test is inherently fp16 or intend to test fp16 behavior -- skip if fp16 is not supported
2. the test fails because a known accelerator feature is missing (i.e. missing op builder) -- skip if that feature is missing
3. the test fails because of unknown/unclear reason -- skip if accelerator is cpu and plan for fix with future PR

So skip against CPU accelerator is a marker that further investigation is needed.

tests/unit/checkpoint/test_zero_optimizer.py
delock: this test passes on zero stage 1/2/3 but not 0.   The failure is that a view is created under no_grad then modified in place under grad mode.  Some investigation will be needed.

delock:
The latest PR should fix nv-torch-latest failrues.
`checkpoint_correctness_verification` will need some further change to be more clean: remove `fp16` parameter and use `dtype` parameter instead.

loadams:
> Hi, @tjruwase @loadams, from the error message, this machine had not been configured for multi-node runner. This machine needs host key verification (ssh with authenticated keys). Otherwise we have to skip this test for CPU accelerator
> 
> ```
>     def test_user_args(cmd):
>         p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
>         out, err = p.communicate()
> >       assert "ARG PARSE SUCCESS" in out.decode("utf-8"), f"User args not parsed correctly: {err.decode('utf-8')}"
> E       AssertionError: User args not parsed correctly: localhost: Host key verification failed.
> 
> E         pdsh@fv-az843-752: localhost: ssh exited with exit code 255
> E         /home/runner/work/DeepSpeed/DeepSpeed/unit-test-venv/lib/python3.8/site-packages/transformers/utils/hub.py:124: FutureWarning: Using `TRANSFORMERS_CACHE` is deprecated and will be removed in v5 of Transformers. Use `HF_HOME` instead.
> E           warnings.warn(
> E         
> ```

@mrwyattii - what are your thoughts here?  But I think that we should probably no install pdsh here and not run those multi-node tests since we don't today?

delock:
A short summary of tests skipped if device_name is "cpu".  Total 23 skips in 12 files

checkpoint/test_lr_scheduler.py:                                                                                                                                                                                                                            26:        if get_accelerator().device_name() == 'cpu':                                                                                                                                                                                         78:        if get_accelerator().device_name() == 'cpu':                                                                                                                                                                                                                                                                                                                                                                                                                                                     checkpoint/test_zero_optimizer.py:                                                                                                                                                                                                                          156:        if zero_stage == 0 and get_accelerator().device_name() == "cpu":                                                                                                                                                                    331:        if zero_stage == 0 and get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                checkpoint/test_other_optimizer.py:                                                                                                                                                                                                                         75:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                                     common.py:                                                                                                                                                                                                                                                  88:            assert get_accelerator().device_name() == 'cpu'                                                                                                                                                                                                                                                                                                                                                                                                                                              launcher/test_user_args.py:                                                                                                                                                                                                                                 47:    if multi_node and get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      runtime/activation_checkpointing/test_activation_checkpointing.py:                                                                                                                                                                                          65:    if get_accelerator().device_name() == "cpu":                                                                                                                                                                                             87:    if get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                                         runtime/zero/test_zero.py:                                                                                                                                                                                                                                  94:        if mics_enabled and get_accelerator().device_name() == "cpu":                                                                                                                                                                        1320:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                                   runtime/zero/test_zero_tensor_fragment.py:                                                                                                                                                                                                                  147:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                                    runtime/compile/test_compile_wrapper.py:                                                                                                                                                                                                                    75:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                                     runtime/compile/test_compile_zero.py:                                                                                                                                                                                                                       33:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                                     runtime/compile/test_load_config.py:                                                                                                                                                                                                                        77:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                         85:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                         96:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                         104:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                        113:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                        122:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                                    runtime/test_ds_config_dict.py:                                                                                                                                                                                                                             251:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                                                                                                                                                                                                                                                                                    runtime/test_data_efficiency.py:                                                                                                                                                                                                                            57:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                         133:        if get_accelerator().device_name() == "cpu":                                                                                                                                                                                        178:        if get_accelerator().device_name() == "cpu":

.github/workflows/nv-inference.yml
```diff
@@ -46,7 +46,8 @@ jobs:
 
       - name: Install deepspeed
         run: |
-          pip install .[dev,1bit,autotuning,inf,triton]
+          DS_ACCELERATOR=cpu pip install .[dev,1bit,autotuning,inf]
+          #pip install .[dev,1bit,autotuning,inf,triton]
```
tjruwase, line 50:
```suggestion
```

tests/unit/checkpoint/test_zero_optimizer.py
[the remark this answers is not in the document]
tjruwase: @delock, you are correct, as always :). Can you please create an issue for separating the skip clauses, and we will handle it. Thanks!

tjruwase (approved):

PR #5211, opened 2024-02-29
Fix loading universal checkpoint for BF16_Optimizer

mosheisland:
PR#5104 (Remove optimizer step on initialization) breaks loading universal checkpoint for BF16_Optimizer.
This is since universal checkpoint attempts to load the optimizer states into lp._hp_mapping.optim_state dictionary before they are initialized (by step).

As a workaround for loading universal checkpoint, perform step and init hp params optimizer's states before loading from universal checkpoint files.


tjruwase:
@misland-habana, I think we have fixed the underlying issue. Can you please check if this PR is still needed? Thanks.

mosheisland:
> @misland-habana, I think we have fixed the underlying issue. Can you please check if this PR is still needed? Thanks.

@tjruwase, sorry for the late reply (vacation...). I have re-tested with your fix for the underlying issue and it is working for me.
I think this PR is not needed anymore.

Issue #5059, opened 2024-02-02
[BUG] ZeRO++ hpZ on llama2-7b gets zero loss


yundai424:
**Describe the bug**
With DeepSpeed built from source on latest master branch (i.e. https://github.com/microsoft/DeepSpeed/pull/4906 is in), hpZ llama2-7b and the training loss will appear 0 starting the 3rd training step:
```
{'loss': 2.1364, 'learning_rate': 0.0, 'epoch': 0.0, 'max_steps': 50, 'global_step': 1, 'current_step_time_seconds': 19.497754335403442, 'average_step_time_seconds': 19.497755765914917, 'estimated_time_to_completion_seconds': 955.3900325298309, 'estimated_total_training_time_seconds': 974.8877882957458}
{'loss': 2.0157, 'learning_rate': 2e-05, 'epoch': 0.0, 'max_steps': 50, 'global_step': 2, 'current_step_time_seconds': 24.733116388320923, 'average_step_time_seconds': 22.11543607711792, 'estimated_time_to_completion_seconds': 1061.5409317016602, 'estimated_total_training_time_seconds': 1105.771803855896}
{'loss': 0.0, 'learning_rate': 2e-05, 'epoch': 0.0, 'max_steps': 50, 'global_step': 3, 'current_step_time_seconds': 24.691705465316772, 'average_step_time_seconds': 22.974192539850872, 'estimated_time_to_completion_seconds': 1079.787049372991, 'estimated_total_training_time_seconds': 1148.7096269925437}
{'loss': 0.0, 'learning_rate': 2e-05, 'epoch': 0.0, 'max_steps': 50, 'global_step': 4, 'current_step_time_seconds': 25.519466638565063, 'average_step_time_seconds': 23.61051106452942, 'estimated_time_to_completion_seconds': 1086.0835089683533, 'estimated_total_training_time_seconds': 1180.525553226471}
```

This however cannot be reproduced on GPT2. 

**To Reproduce**
1. Training script:
```python
import transformers
import datasets
import trl
from dataclasses import dataclass, field
import peft
import torch
import time
import callbacks

MODEL_NAME = "meta-llama/Llama-2-7b"

def formatting_func(example):
    output_texts = []
    for i in range(len(example['question'])):
        choices = ""
        for j in range(len(example["choices"][i])):
            choices += f"{j+1}. {example['choices'][i][j]}\n"
[...]

PR #5216, opened 2024-03-01
[zero++] Synchronize at the end of secondary partitioning and simplify the logic

ByronHsu:
## 1. Why?

We have a very long thread investigating [the issue](https://github.com/microsoft/DeepSpeed/issues/5059). To summarize, this is because

a. The 2nd partitioning is asynchronous because it copies device-to-device from full tensor to 2nd tensor
b. When using prefetching, the all-gather of 2nd tensor can happen before 2nd partitioning ends. At that moment, the value of 2nd tensor might contain bad values.

image

Also, we found that the logic of copying is wrong and lengthy, so we simplified it to only two lines.

Kudos to @yundai424, Haowen Ning, @samadejacobs for the investigation effort.

## 2. What? 

After multiple careful tests, we found patching `get_accelerator().synchronize()` to ensure all cuda stream finished before 2nd partitioning can prevent the issue

## 3. Tests

I validated the correctness of the simplification of 2nd partition logic. The loss is "exactly" the same before and after simplification under the same random seed.

Before

```
[
  {"loss": 2.0731},
  {"loss": 2.0288},
  {"loss": 1.927},
  {"loss": 1.8347},
  {"loss": 1.8347},
  {"loss": 1.7896},
  {"loss": 1.602},
  {"loss": 1.766},
  {"loss": 1.8751},
  {"loss": 1.6776}
]

```

After

```
[
  {"loss": 2.0731},
  {"loss": 2.0288},
  {"loss": 1.927},
  {"loss": 1.8347},
  {"loss": 1.8347},
  {"loss": 1.7896},
  {"loss": 1.602},
  {"loss": 1.766},
  {"loss": 1.8751},
  {"loss": 1.6776}
]


```

## 4. TODO

We need further investigation on the issue @samadejacobs 
1) Revisit ZeRO-3 prefetch design 
2) Refactor hpz to reuse primary tensor for secondary partition.

### DIFF
--- a/deepspeed/runtime/zero/partition_parameters.py
+++ b/deepspeed/runtime/zero/partition_parameters.py
@@ -1649,7 +1649,7 @@ class Init(InsertPostInitMethodToModuleSubClasses):
                     param.ds_secondary_tensor.narrow(0, 0, elements_to_copy_sec).copy_(
                         one_dim_param.narrow(0, secondary_start, elements_to_copy_sec))
 
-            # TODO: This is a temporary fix to avoid the issue of 2nd tensor all-gather happens before 2nd tensor partitioning (copy_ is async d2d copy) is done
+            # TODO: This is a temporary fix to avoid the issue of 2nd tensor all-gather happens before 2nd tensor partition is done
             get_accelerator().synchronize()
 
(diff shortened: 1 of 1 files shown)


tjruwase:
@GuanhuaWang, FYI!

Issue #5203, opened 2024-02-27
[BUG] Deepspeed Crashes when using MoE, Stage 2 Offload with DeepSpeedCPUAdam


KyleMylonakisProtopia:
**Describe the bug**
When performing a training run with a model with Mixture of Experts (MoE) layers using stage 2 offload with the DeepSpeedCPUAdam optimizer, during the parameter update step the following runtime error is thrown.

```
│ /home/kyle/.conda/envs/llama2-chat/lib/python3.11/site-packages/lightning/fabric/wrappers.py:92 in step                                                                                                                                                                │
│                                                                                                                                                                                                                                                                        │
│ ❱  92 │   │   output = self._strategy.optimizer_step(                                                                                                                                                                                                                  │
│                                                                                                                                                                                                                                                                        │
│ /home/kyle/.conda/envs/llama2-chat/lib/python3.11/site-packages/lightning/fabric/strategies/strategy.py:206 in optimizer_step                                                                                                                                          │
│                                                                                                                                                                                                                                                                        │
│ ❱ 206 │   │   return self.precision.optimizer_step(optimizer, **kwargs)                                                                                                                                                                                                │
│                                                                                                                                                                                                                                                                        │
│ /home/kyle/.conda/envs/llama2-chat/lib/python3.11/site-packages/lightning/fabric/plugins/precision/deepspeed.py:100 in optimizer_step                                                                                                                                  │
│                                                                                                                                                                                                                                                                        │
│ ❱ 100 │   │   return optimizer.step(**kwargs)                                                                                                                                                                                                                          │
│                                                                                                                                                                                                                                                                        │
│ /home/kyle/.conda/envs/llama2-chat/lib/python3.11/site-packages/deepspeed/runtime/zero/stage_1_and_2.py:1842 in step                                                                                                                                                   │
│                                                                                                                                                                                                                                                                        │
│ ❱ 1842 │   │   scaled_global_grad_norm = self.scaled_global_norm()                                                                                                                                                                                                     │
│                                                                                                                                                                                                                                                                        │
│ /home/kyle/.conda/envs/llama2-chat/lib/python3.11/site-packages/deepspeed/runtime/zero/stage_1_and_2.py:1786 in scaled_global_norm                                                                                                                                     │
│                                                                                                                                                                                                                                                                        │
│ ❱ 1786 │   │   │   self._average_expert_grad_norms(norm_groups)                                                                                                                                                                                                        │
[...]

PR #5220, opened 2024-03-02
Fix moe cpu offload

RezaYazdaniAminabadi:
The MoE- param gradients norms don't need to be averaged when created on CPU only when using 1-DP training. However, I just moved the tensor back to GPU to get average when having data-parallel on the MoE parameters and using CPU-offload.

This PR addresses https://github.com/microsoft/DeepSpeed/issues/5203

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1782,8 +1782,8 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
             else:
                 norm_groups.append(self.get_grad_norm_direct(self.averaged_gradients[i], self.params_in_partition[i]))
 
-        if self.has_moe_layers:
-            self._average_expert_grad_norms(norm_groups)
+        if self.has_moe_layers and self.device != 'cpu':
+                self._average_expert_grad_norms(norm_groups)
 
         # calculating L2 norm
         return torch.norm(torch.stack(norm_groups), p=norm_type)
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1782,7 +1782,7 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
             else:
                 norm_groups.append(self.get_grad_norm_direct(self.averaged_gradients[i], self.params_in_partition[i]))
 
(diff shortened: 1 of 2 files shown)


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1782,7 +1782,7 @@ def scaled_global_norm(self, norm_type=2):
             else:
                 norm_groups.append(self.get_grad_norm_direct(self.averaged_gradients[i], self.params_in_partition[i]))
 
-        if self.has_moe_layers:
+        if self.has_moe_layers and self.device != 'cpu':
```
tjruwase, line 1785: @RezaYazdaniAminabadi, won't this result in different gradient norms printout for offload vs non-offload?

RezaYazdaniAminabadi: as long as the same global batch size is used between the two, I think they should stay the same. When non-offload, if one GPU is used then there is no averaging and when multi-gpu used then the batch is partitioned (assuming we have data-parallel for the expert part), and then only a partition of the data is used per rank and potentially the norm should the same result when reduced. Is my understanding correct?

tjruwase: I think the right solution would be to create a per-rank device tensor (e.g., self.scaled_gradient_norm) to represent this [value](https://github.com/RezaYazdaniAminabadi/DeepSpeed/blob/9a59df40e7b693acb968f2d14c8b65cc913eeea7/deepspeed/runtime/zero/stage_1_and_2.py#L1948) and update during accumulation, and then allreduce before step.

RezaYazdaniAminabadi: I see, you are worried that this might cause issue in the case of multi-gpu that still we have the CPU-offload and the data between ranks are different?

RezaYazdaniAminabadi: okay, will make the change and commit shortly

RezaYazdaniAminabadi: thanks @tjruwase

tjruwase: Thanks @RezaYazdaniAminabadi

tjruwase (approved):

PR #5254, opened 2024-03-11
Disable torch.nn.init when counting parmeters in initializing PipelineModule

tanconghui:
Disable any function call in torch.nn.init when counting parameters in initializing PipelineModule.

This change can greatly save the launching time of pipeline parallelism, by speeding up parameter partitioning. I have tested it with Llama2 70B, reducing the partition time from 8 minutes to less than 1 second.


tanconghui:

PR #5268, opened 2024-03-13
Average only valid part of the ipg buffer.

BacharL:
When contiguous gradients is used ipg buffer may not be fully utilized. Call average_tensor only for the slice with valid gradints

Change-Id: I760559d52c2f91e15cd6cd0b48e534ec2352802a


tjruwase (approved):

tjruwase:
@BacharL, thanks for catching this corner case.

PR #5270, opened 2024-03-13
Skip 1Bit Compression and sparsegrad tests for HPU.

vshekhawat-hlab:
HPU accelrator does not support 1bit and sparsegrad tests.


tjruwase (approved):

loadams:
@delock - agreed, I'll look into this as well as your torch.compile comment on the other PR.

tjruwase:
> Ideally a test should check for accelerator feature instead of check for accelerator name. So other accelerators could skip the test without modifying the test code again. This also gives DeepSpeed code a chance to gracefully skip some feature instead of run into error. Might be some API i.e. if "1bit" in get_accelerator().capabilities() etc.
> 
> This could apply to feature that does not require an opbuilder. For feature using opbuilder, a typical feature check could be check compatibility of the opbuilder.
> 
> @tjruwase @loadams @vshekhawat-hlab

@delock, FYI we are trying to do this in #5423. It would be great to get your thoughts. Thanks! 

@duli2012, FYI

PR #5287, opened 2024-03-16
Fix Initial Error in "WarmupCosineLR" scheduler at steps 0 and 1

richardodliu:
### Overview
This pull request addresses a specific issue within the "WarmupCosineLR" scheduler where errors could occur at the very beginning of the training process, particularly at steps 0 and 1. Such errors can cause the initial step's data to effectively not participate in the training process and lead to inconsistencies with the behavior of the "cosine" lr_scheduler_type found in the transformers library.

### Problem Description
It was identified that the WarmupCosineLR scheduler may not handle the initial steps (0 and 1) correctly, potentially causing computational inaccuracies or exceptions. This situation arises due to inadequate handling of these early steps in the learning rate computation logic. To reproduce the problem, you can follow the code below:

```
from deepspeed.runtime.lr_schedules import WarmupCosineLR

optimizer = torch.optim.Adam(model.parameters(), lr=1.0)
scheduler = WarmupCosineLR(optimizer,
                            total_num_steps = 100,
                            warmup_num_steps = 10,
                            warmup_type = "linear")


for step in range(20):

    print(f"Step: {step}, Using LR: {optimizer.param_groups[0]['lr']}")

    optimizer.zero_grad()
    outputs = model(torch.randn(64, 10))
    loss = outputs.mean()
    loss.backward()
    optimizer.step()
    scheduler.step()

    print(f"Step: {step}, Updating LR: {optimizer.param_groups[0]['lr']}")
```

The outputs corresponding to the first three steps are as follows:

```
Step: 0, Using LR: 1.0
Step: 0, Updating LR: 0.0
Step: 1, Using LR: 0.0
Step: 1, Updating LR: 0.1
Step: 2, Using LR: 0.1
Step: 2, Updating LR: 0.2
```

We can notice that the learning rates used during the updates at step 0 and step 1 are abnormal and do not meet the requirements for learning rates in actual training scenarios. We provide the code to replicate "get_cosine_schedule_with_warmup" from transformers as follows:

```
from transformers import get_cosine_schedule_with_warmup

optimizer = torch.optim.Adam(model.parameters(), lr=0.1)
scheduler = get_cosine_schedule_with_warmup(optimizer, 
                                            num_warmup_steps= 10, 
                                            num_training_steps = 100)

for step in range(20):

    print(f"Step: {step}, Using LR: {optimizer.param_groups[0]['lr']}")

    optimizer.zero_grad()
    outputs = model(torch.randn(64, 10))
    loss = outputs.mean()
    loss.backward()
    optimizer.step()
    scheduler.step()

    print(f"Step: {step}, Updating LR: {optimizer.param_groups[0]['lr']}")
```

The outputs corresponding to the first three steps are as follows:

```
Step: 0, Using LR: 0.0
Step: 0, Updating LR: 0.010000000000000002
Step: 1, Using LR: 0.010000000000000002
Step: 1, Updating LR: 0.020000000000000004
Step: 2, Using LR: 0.020000000000000004
Step: 2, Updating LR: 0.03
```

Considering that dummy batches may exist in the actual training process, the implementation provided by transformers seems more reasonable. This is because, according to the implementation of DeepSpeed, if there are dummy batches, the data at step 1 would not actually participate in the training due to the learning rate being 0. On the other hand, if there are no dummy batches, then the data at step 0 would participate in the training with the standard learning rate, but the data at step 1 would not, which contradicts the original intention of the warmup. Therefore, I believe the current version of DeepSpeed's implementation of "WarmupCosineLR" is not reasonable. This error seems likely to exist in all lr_schedulers that include a warmup phase. I hope you can take note of this.

### Solution
A simple but efficient solution (as I have submitted in this PR) is to execute `self.step()` once at the end of the `class WarmupCosineLR` initialization `__init__`, to skip the step with the standard learning rate for the first time. My local experimental results indicate that such a modification can achieve almost identical effects to get_cosine_schedule_with_warmup from transformers (since transformers.trainer seems to use step 0 as a dummy batch). I believe this modification is reasonable and effective and could be added to other learning rate schedulers that also use warmup, if feasible.

### Benefits
- Enhances the stability and reliability of the WarmupCosineLR scheduler, especially at the start of the training.
- To maintain similar effects to the implementation of this method in transformers.
- Ensures that the learning rate adjustment strategy functions as intended from the very first training step.

I believe these changes will significantly improve the user experience with DeepSpeed's WarmupCosineLR scheduler by addressing a critical issue that could affect many users. I look forward to feedback and any further suggestions for improvement.


sgunasekar:
Why is this PR marked as drafts - seems reasonable fix to me which is also done by torch's native LR schedulers! Came across this as the same issue was brought up by someone in my team.

tjruwase:
@richardodliu, are you able to revive this PR?

tjruwase:
> Why is this PR marked as drafts - seems reasonable fix to me which is also done by torch's native LR schedulers! Came across this as the same issue was brought up by someone in my team.

@sgunasekar, this PR is from 3P contributor, so not sure why they marked it draft. I already pinged them above. Let's wait a bit for a response.

tjruwase:
@richardodliu, since this is a one-liner, I have marked it ready for review. I suspect the following unit test might need updating
https://github.com/microsoft/DeepSpeed/blob/8411816583ee5fe56d95f2241e6e54199b879849/tests/unit/runtime/test_lr_schedulers.py#L447

@sgunasekar, FYI

richardodliu:
> @richardodliu, since this is a one-liner, I have marked it ready for review. I suspect the following unit test might need updating
> 
> https://github.com/microsoft/DeepSpeed/blob/8411816583ee5fe56d95f2241e6e54199b879849/tests/unit/runtime/test_lr_schedulers.py#L447
> 
> @sgunasekar, FYI

@sgunasekar Thanks for your attention!

The reason this PR was marked as a draft is that I noticed my PR could not pass the unit test at the time. I reviewed the code again today, and now the issue is clearer: The problem I encountered can still be reproduced with the code I submitted above. Then, I reproduced the unit test function used in the method:
https://github.com/microsoft/DeepSpeed/blob/8411816583ee5fe56d95f2241e6e54199b879849/tests/unit/runtime/test_lr_schedulers.py#L447
specifically, I used the following code:

```
import torch
import deepspeed

def random_dataset(total_samples, hidden_dim, device, dtype=torch.float16):
    train_label = torch.empty(total_samples, dtype=torch.long, device=device).random_(hidden_dim)
    return train_dataset

def random_dataloader(model, total_samples, hidden_dim, device, dtype=torch.float16):
    batch_size = model.train_micro_batch_size_per_gpu()
    train_dataset = random_dataset(total_samples, hidden_dim, device, dtype=dtype)
    train_loader = torch.utils.data.DataLoader(train_dataset, batch_size=batch_size)
    return train_loader

class SimpleModel(torch.nn.Module):

    def __init__(self, hidden_dim, empty_grad=False, nlayers=1):
        super(SimpleModel, self).__init__()
        self.linears = torch.nn.ModuleList([torch.nn.Linear(hidden_dim, hidden_dim) for i in range(nlayers)])
        if empty_grad:
            self.linear2 = torch.nn.Linear(hidden_dim, hidden_dim)
        self.empty_grad = empty_grad

    def forward(self, x, y):
        if len(self.linears) == 1:
            x = self.linears[0](x)
        else:
            for i, l in enumerate(self.linears):
                x = self.linears[i // 2](x) + l(x)
        return self.cross_entropy_loss(x, y)

config_dict = {
    "train_batch_size": 1,
    "steps_per_print": 1,
    "optimizer": {
        "type": "Adam",
        "params": {
            "lr": 1.0
        },
    },
    "scheduler": {
        "type": 'WarmupCosineLR',
        "params": {
            'warmup_type' : "linear",
            'total_num_steps': 20,
            'warmup_min_ratio': 0,
            'warmup_num_steps': 10,
            'cos_min_ratio': 0,
        }
    },
    "gradient_clipping": 1.0
}
hidden_dim = 10

model = SimpleModel(hidden_dim, empty_grad=False)
model, _, _, lr_scheduler = deepspeed.initialize(config=config_dict,
                                                    model=model,

data_loader = random_dataloader(model=model,
                                total_samples = 20,
                                hidden_dim=hidden_dim,
                                device=model.device,
                                dtype=torch.float)

step_lrs = []
for _, batch in enumerate(data_loader):
    model.backward(loss)
    model.step()
    step_lrs.extend(lr_scheduler.get_lr())

print(step_lrs)
```

And we can have the following result:
```
[0.0,
 0.1,
 0.2,
 0.3,
 0.4,
 0.5,
 0.6,
 0.7,
 0.8,
 0.9,
 0.9755282581475768,
 0.9045084971874737,
 0.7938926261462366,
 0.6545084971874737,
 0.5,
 0.34549150281252633,
 0.2061073738537635,
 0.09549150281252633,
 0.024471741852423234,
 0.0]
```

It seems that the initial value error we mentioned appears to have disappeared, but this is not actually the case. We notice that the maximum learning rate is only 0.976, not the desired 1.0. Overall, this PR does not seem to completely resolve the initial value error I encountered. This is because using this learning rate scheduler alone behaves differently from when it is wrapped within the DeepSpeed framework (specifically, whether it skips the first maximum learning rate step). This error is systemic since it is inherited and used by other methods in DeepSpeed (otherwise, our PR modifying the init method would not fail the unit test). Additionally, the learning rate scheduler seems unable to pass the second statement in the test under any circumstances:
```
# Verify peak lr
assert abs(step_lrs[warmup_num_steps - 1] - opt_lr) < 1e-7
```
Overall, the errors I am currently encountering can be summarized into two categories. One is the inconsistency in the behavior of the lr scheduler when called by different methods, and the other is the maximum learning rate not meeting the expected value. These two issues seem to be caused by some detailed errors in the implementation. Due to time constraints, I am unable to provide a comprehensive solution at the moment, so I cannot completely fix this problem for now.

PR #5291, opened 2024-03-17
Parallel map step for `DistributedDataAnalyzer` map-reduce

bm-synth:
- adds multi CPU-processing to the `DistributedDataAnalyzer` map operation (parallelism set with parameter `num_workers`). Works with a `SharedMemory` / `Manager's` queue per metric, written concurrently by processes.
- much faster `write_buffer_to_file` in `DistributedDataAnalyzer` reduce operation by copying to cpu and "detaching"  output tensor.


bm-synth:
@loadams @conglongli  what's holding this PR?

conglongli (approved):

PR #5292, opened 2024-03-17
Bug fix in `split_index` method

bm-synth:
Bug description: on a dataset of 20 samples, when running 4 workers with 8 threads per worker, then the `split_dataset` would return for worker id `1`: 

```
self.worker_splits
[[0, 5], [5, 10], [10, 15], [15, 20]]


self.thread_splits
[[5, 6], [6, 7], [7, 8], [8, 9], [9, 10], [10, 10], [11, 10], [12, 10]]
```

`thread_splits` is wrong and causes a crash in the `DataAnalyzer`: the end sample id is lower than the initial one on the last 2 threads.
This PR fixes that by fixing the behaviour of `split_index`


bm-synth:
@loadams @conglongli  what's holding this PR?

loadams:
> @loadams @conglongli what's holding this PR?

We need the correct folks to review

loadams (approved):

PR #5296, opened 2024-03-18
Fix dtype-related RuntimeError when computing scaled global norm

potamides:
`complete_grad_norm_calculation_for_cpu_offload` can return a python int even though `scaled_global_norm` expects it to always return a python float which makes `torch.norm` fail with a RuntimeError.


tjruwase:
@potamides, thanks for this PR. However, it is a duplicate of #5252.

PR #5298, opened 2024-03-19
Supporting custom backend for all accelerator.

umchand:


deepspeed/runtime/zero/offload_config.py
```diff
@@ -14,6 +14,7 @@ class OffloadDeviceEnum(str, Enum):
     none = "none"
     cpu = "cpu"
     nvme = "nvme"
+    hpu = "hpu"
```
tjruwase, line 17: `hpu` is not a memory target for offloading. 
```suggestion
```

accelerator/abstract_accelerator.py
```diff
@@ -287,3 +287,7 @@ def build_extension(self):
     @abc.abstractmethod
     def export_envs(self):
         ...
+
+    @abc.abstractmethod
+    def get_accelerator_backend(self, device_index=None):
```
tjruwase, line 292: 1. Compile backend should be global rather than per device. 
2. We are returning compilation backend, so `compile_backend` is a more apt name. 
```suggestion
    def get_compile_backend(self):
```

tests/unit/runtime/compile/test_compile_zero.py
```diff
@@ -24,7 +24,9 @@ class TestZeRO(DistributedTest):
 
     @pytest.mark.parametrize('dtype', [torch.bfloat16, torch.float16, torch.float32])
     @pytest.mark.parametrize('zero_stage', [1, 2, 3])
-    @pytest.mark.parametrize('offload_device', [OffloadDeviceEnum.none, OffloadDeviceEnum.cpu, OffloadDeviceEnum.nvme])
+    @pytest.mark.parametrize(
+        'offload_device',
+        [OffloadDeviceEnum.none, OffloadDeviceEnum.cpu, OffloadDeviceEnum.nvme, OffloadDeviceEnum.hpu])
```
tjruwase, line 29: `hpu` is not an offload device.

tjruwase:
@umchand, I think we can close this because of #5299. Please take a look and share comments/questions.

PR #5301, opened 2024-03-20
fix pagable h2d memcpy

GuanhuaWang:
ZeRO offload case

Fix the issue of pageble h2d memcpy in step process. Now h2d memcpy uses pinned memory.

Speedup h2d memcpy by 6x on single GPU and 4-5x on 8GPU node.

cc @tjruwase


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1863,7 +1863,7 @@ def step(self, closure=None):
                 #    bit16_partitions[partition_id].data.copy_(fp32_partition.data)
                 bit16_partitions = self.parallel_partitioned_bit16_groups[i]
                 fp32_partition = self.single_partition_of_fp32_groups[i]
-                bit16_partitions[partition_id].data.copy_(fp32_partition.data)
+                bit16_partitions[partition_id].data.copy_(fp32_partition.to('cuda').data)
```
tjruwase, line 1866: We should use `get_accelerator()` instead of `cuda` references. 
```suggestion
                bit16_partitions[partition_id].data.copy_(fp32_partition.to(get_accelerator().current_device_name()).data)
```

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -390,11 +390,19 @@ def __init__(self,
             # Note that the params in single_partition_of_fp32_groups is cloned and detached
             # from the origin params of the model.
             if not fp16_master_weights_and_gradients:
-                self.single_partition_of_fp32_groups.append(self.parallel_partitioned_bit16_groups[i][partition_id].to(
-                    self.device).clone().float().detach())
+                weights_partition = self.parallel_partitioned_bit16_groups[i][partition_id].to(
```
tjruwase, line 393: The new change seems unduly complicated with the duplication of the offload condition in the `if not fp16_master_weights_and_gradients` branch. 

Is there a reason why something like below does not work?

```python
            if not fp16_master_weights_and_gradients:
                weights_partition = self.parallel_partitioned_bit16_groups[i][partition_id].to(
                    self.device).clone().float().detach()
            else:
                weights_partition = self.parallel_partitioned_bit16_groups[i][partition_id].to(
                    self.device).clone().half().detach()
           
            if self.cpu_offload:
               weights_partition = get_accelerator().pin_memory(weights_partition)
             
            self.single_partition_of_fp32_groups.append(weights_partition)
``

tjruwase (approved):

PR #5317, opened 2024-03-27
Correct typo in checking on bf16 unit test support

loadams:
Co-Authored-By: Du Li <du.li@microsoft.com>

### DIFF
--- a/tests/unit/checkpoint/test_lr_scheduler.py
+++ b/tests/unit/checkpoint/test_lr_scheduler.py
@@ -53,7 +53,7 @@ class TestLRSchedulerCheckpoint(DistributedTest):
         }
         if get_accelerator().is_fp16_supported():
             config_dict["fp16"] = {"enabled": True}
-        elif get_accelerator().is_fp16_supported():
+        elif get_accelerator().is_bf16_supported():
             config_dict["bf16"] = {"enabled": True}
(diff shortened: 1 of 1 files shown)


tests/unit/checkpoint/test_lr_scheduler.py
```diff
@@ -53,7 +53,7 @@ def test_checkpoint_lr_scheduler(self, tmpdir, zero_stage, use_cpu_offload):
         }
         if get_accelerator().is_fp16_supported():
             config_dict["fp16"] = {"enabled": True}
-        elif get_accelerator().is_fp16_supported():
+        elif get_accelerator().is_bf16_supported():
```
loadams, line 56: fyi @delock

duli2012: @delock ln54-56 was added by you in another PR. Why do we need it?

tjruwase (approved):

tests/unit/checkpoint/test_lr_scheduler.py
delock: @duli2012 the intention is to allow the test run on CPU which currently does not support FP16 datatype, but BF16 datatype is supported.  Thanks for fix the bug!

PR #5328, opened 2024-03-28
BF16 optimizer: Clear lp grads after updating hp grads in hook

YangQun1:
This fix is to solve:

- Previous iteration's lp grads will still alive during the next iteration's forward. This increases the memory footprint.
- The hook behavior is not aligned to its name accumulate_hp_grads_and_remove_lp


YangQun1:
Hi @mrwyattii @tjruwase , could you help to review this change?

tjruwase (approved):

PR #5329, opened 2024-03-28
Remove dtype(fp16) condition check for residual_add unit test

raza-sikander:
When the dtype is bf16 or fp32 the if condition is not satisfied and it continues execution instead of skipping when triton is not installed.


raza-sikander:
@loadams  could you please review.

tjruwase (approved):

loadams (approved):

PR #5333, opened 2024-03-29
Fix the FP6 kernels compilation problem on non-Ampere GPUs.

JamesTheZ:
Refine the guards of FP6 kernel compilation. Fix the `undefined symbol` problem of FP6 kernels on non-Ampere architectures.

Related issue: https://github.com/microsoft/DeepSpeed-MII/issues/443.


tests/unit/inference/v2/modules/test_quantized_linear_module.py
```diff
@@ -145,10 +145,9 @@ def _fp6_quantized_linear_helper(tokens: int,
     ActivationType.ReGLU,
     ActivationType.SiGLU,
 ]
-all_tokens = [1, 37]
```
loadams, line 148: Do we just not support these cases anymore or should we make them conditional on the hardware they are running on?

JamesTheZ: This is to reduce the number of tests for FP6 kernel. Currently, the UTs of FP6 kernels are too time consuming.

loadams: Sounds good, thanks!

mrwyattii (approved):

PR #5336, opened 2024-03-29
FP [6,8,12] quantizer op

jeffra:
Flexible-bit quantizer-dequantizer library with fp6/fp12/fp8 support

Requires Ampere+ architecture, this is due to the initial focus of this op only on `bfloat16` input types.

Co-authored-by: Reza Yazdani <reza.yazdani@snowflake.com>

### DIFF
--- a/csrc/fp_quantizer/quantize.cu
+++ b/csrc/fp_quantizer/quantize.cu
@@ -1,0 +1,427 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0
+
+// DeepSpeed Team
+
+#include <stdexcept>
+#include "context.h"
+#include "memory_access_utils.h"
+#include "quantize.h"
+#include "reduction_utils.h"
+
+#include <cuda.h>
+#include <stdint.h>
+
+#include <cuda_fp16.h>
+#include <curand_kernel.h>
+
+#include <cuda_bf16.h>
+#include <cuda_runtime_api.h>
+
+using ROp = reduce::ROpType;
+
+namespace quantization {
+
+constexpr int access_granularity = 16;
+constexpr int quanitzed_access_granularity = 4;
+constexpr int quanitzed_access_granularity_6bits = 2;
+constexpr int threads = 256;
+constexpr int warps = threads / 32;
+
+}  // namespace quantization
+
+template <int _mantisa_bits, int q_mantisa_bits, int stochastic_rounding>
+__device__ void round(uint32_t& mantisa, uint32_t& dst_exponent, curandStatePhilox4_32_10_t* state)
+{
+    constexpr uint32_t mantisa_mask = (1 << (_mantisa_bits - q_mantisa_bits)) - 1;
+    uint32_t offset = stochastic_rounding ? (curand_poisson(state, 10) & mantisa_mask)
+                                          : 1 << (_mantisa_bits - q_mantisa_bits - 1);
+    mantisa += offset;
+    dst_exponent += (((mantisa & ~mantisa_mask) == (1 << _mantisa_bits)) ? 1 : 0);
+}
+
+template <int _mantisa_bits, int _exponent_bits, int q_mantisa_bits, int q_exponent_bits>
+__device__ void clip(uint32_t& exponent, uint32_t& mantisa)
+{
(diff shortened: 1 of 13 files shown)


jeffra:
Regarding the `nv-pre-compile-ops` test failure. I just did some debugging on this and I see that for some reason if you set `DS_ENABLE_NINJA=1` this fails but if you don't use ninja the pre-compile works fine. I remember we used to have lots of issues with using ninja but then it was re-enabled in https://github.com/microsoft/DeepSpeed/pull/5088/. Not sure if this PR introduces something sublte that doesn't work w. ninja?

@mrwyattii @loadams any ideas?

jeffra:
@microsoft-github-policy-service agree company="Snowflake"

loadams:
> Regarding the `nv-pre-compile-ops` test failure. I just did some debugging on this and I see that for some reason if you set `DS_ENABLE_NINJA=1` this fails but if you don't use ninja the pre-compile works fine. I remember we used to have lots of issues with using ninja but then it was re-enabled in #5088. Not sure if this PR introduces something sublte that doesn't work w. ninja?
> 
> @mrwyattii @loadams any ideas?

@jeffra - No immediate guesses, but I'll take a look

jeffra:
Thanks @loadams! I believe this is resolved now, I figured out the issue with ninja enabled. I have to disable the pre-compile for this op for the test since it includes V100 compute target which is not compatible with this op (since it only works with bf16). I was able to get all the CI tests to pass (before updating with latest master right now). 

I think now it just needs a review from folks? :)

loadams:
> Thanks @loadams! I believe this is resolved now, I figured out the issue with ninja enabled. I have to disable the pre-compile for this op for the test since it includes V100 compute target which is not compatible with this op (since it only works with bf16). I was able to get all the CI tests to pass (before updating with latest master right now).
> 
> I think now it just needs a review from folks? :)

That makes sense, I hadn't reviewed which is probably why I missed that you already identified the issue :)

mrwyattii (approved):
@arashashari and/or @JamesTheZ please add your review as well, thank you

deepspeed/ops/fp_quantizer/quantize.py
```diff
@@ -0,0 +1,79 @@
...
+        if fp_quant_module is None:
+            fp_quant_module = FPQuantizerBuilder().load()
+
+        self.group_size = group_size
+        self.orig_dtype = None
+
+    def quantize(self,
+                 input,
+                 q_bits=8,
+                 q_mantisa_bits=3,
```
JamesTheZ, line 26: NIT: `q_mantisa_bits` seems only used for `q_bits == 8`. This parameter can be moved and directly assign its value in L37? Or moving the predefined `q_mantisa_bits ` values in L38-43.

sfc-gh-reyazda: that's a good point @JamesTheZ, the reason we are receiving as the parameter for FP8 is that we can switch between E5M2 and E4M3, but for the rest we hard-code the bit-ratios the kernel supports.

csrc/fp_quantizer/quantize.cu
```diff
@@ -0,0 +1,427 @@
...
+                    dst_mantisa, dst_exponent, &state);
+                if (cur_exponent != 0)
+                    clip<_mantisa_bits, _exponent_bits, q_mantisa_bits, q_exponent_bits>(
+                        dst_exponent, dst_mantisa);
+
+                dst_mantisa = (dst_mantisa & mantisa_mask) >> (_mantisa_bits - q_mantisa_bits);
+
+                if (dst_exponent != (1 << q_exponent_bits) - 1)
+                    dst_exponent = (dst_exponent - ((1 << (_exponent_bits - 1)) - 1)) +
+                                   (1 << (q_exponent_bits - 1)) - 1;
```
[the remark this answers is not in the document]
sfc-gh-reyazda: Yes, it is handled in the clip and round functions and also using the condition on line 136.

sfc-gh-reyazda:
> @arashashari and/or @JamesTheZ please add your review as well, thank you

@mrwyattii, probably you meant @arashb :) btw, @arashashari is also welcome to review this ;)

JamesTheZ (approved):
LGTM

PR #5343, opened 2024-04-01
Improve parallel process of universal checkpoint conversion

tohtana:
The conversion script from a regular checkpoint to the universal one runs the followings in parallel.

1. extracts zero sharded optimizer states
2. merge the shards

However, it passes `map()` a set of only a few tasks (the number specified as workers). Thus it needs to wait for the slowest tasks to finish for every set. 
This PR submits all the tasks to the pool and wait until the futures get ready. We can keep all workers running.


tjruwase (approved):

tjruwase:
@tohtana, this is an amazing improvement. Did you observe any conversion speedups that you can share in this PR?

tohtana:
@tjruwase I converted data on a blob storage. It was totally limited by IO but I still observed 10-20% speed up in total.
At the very initial stage of conversion, it was 2x faster but slowed down soon. I thought it was the limitation of the blog.

PR #5352, opened 2024-04-02
Fix redundant seq data parallel grp argument in Z3/MiCS

samadejacobs:
Deprecate redundant sequence_data_parallel_group argument. Users/client code will control across which process group Z3 parameters will be partitioned from one of [None, data_parallel_group, sequence_data_parallel].


deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -789,24 +789,21 @@ class Init(InsertPostInitMethodToModuleSubClasses):
...
+                 remote_device=None,
+                 pin_memory=False,
+                 config_dict_or_path=None,
+                 config=None,
+                 enabled=True,
+                 dtype=None,
+                 mpu=None,
+                 zero_param_parallel_group=None,
+                 zero_quantized_weights=False,
+                 zero_quantized_nontrainable_weights=False,
```
mrwyattii, line 805:
```suggestion
                 zero_quantized_nontrainable_weights=False,
                 sequence_data_parallel_group=None,
```
Let's properly deprecate this over the next few versions. Below, in the `__init__` method, do something like:
```python
if sequence_data_parallel_group is not None:
    print_rank_0("WARNING: 'sequence_data_parallel_group' is deprecated and will be removed. Use 'data_parallel_group' instead.", force=True)
    if data_parallel_group is not None:
        raise ValueError("Both 'data_parallel_group' and 'sequence_data_parallel_group' were specified. Please provide only one of these arguments.")
    data_parallel_group = sequence_data_parallel_group
```

mrwyattii (changes requested):

tohtana (approved):
Looks good to me, thank you!

PR #5354, opened 2024-04-02
Auto convert moe param groups

jeffra:
When using frameworks like HF Accelerate with MoE models in HF there's an issue when DeepSpeed is creating the optimizer where we have no way to automatically create the compatible MoE param groups. This PR detects if no client optimizer is set and model_parameters are passed to DeepSpeed that they are either MoE compatible or makes them MoE compatible automatically.

This was never an issue previously since (1) MoE hasn't really been tested outside MDS and (2) MDS manually converts the weight-decay param groups into being MoE compatible before deepspeed.initialize.

The error that is triggered if the param groups are not MoE compatible is triggered here: https://github.com/microsoft/DeepSpeed/blob/cc897ecf15fdac5437fa4a2743154dc6c1749da4/deepspeed/runtime/zero/stage_1_and_2.py#L610-L612

### DIFF
--- a/deepspeed/moe/utils.py
+++ b/deepspeed/moe/utils.py
@@ -150,3 +150,29 @@ def split_params_into_different_moe_groups_for_optimizer(
 
 def is_moe_param_group(param_group):
     return param_group.get('moe', False)
+
+
+def configure_moe_param_groups(model_parameters: List):
+    # peak at the first element to determine how to proceed
+    first = model_parameters[0]
+
+    # match torch.optim.Optimizer expectations
+    if not isinstance(first, (torch.Tensor, dict)):
+        raise TypeError("param argument that would be given to the optimizer should be "
+                        f"an iterable of Tensors or dicts, but got {type(first)}")
+
+    # Case 1: model_parameters is a list of torch.nn.Parameter
+    #   -> need to create moe compatible param groups
+    if isinstance(first, torch.nn.Parameter):
+        param_group = {'params': model_parameters, 'name': 'dense-params'}
+        return split_params_into_different_moe_groups_for_optimizer(param_group)
+
+    # Case 2: model_parameters is a list of param groups List[dict]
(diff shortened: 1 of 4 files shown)


deepspeed/moe/utils.py
```diff
@@ -150,3 +150,29 @@ def split_params_into_different_moe_groups_for_optimizer(
 
 def is_moe_param_group(param_group):
     return param_group.get('moe', False)
+
+
+def configure_moe_param_groups(model_parameters: List):
```
tohtana, line 155: Is `model_parameters` always a list? Can it be other Iterable? I wonder if [0] always works.

deepspeed/moe/utils.py
```diff
@@ -150,3 +150,29 @@ def split_params_into_different_moe_groups_for_optimizer(
...
+    # peak at the first element to determine how to proceed
+    first = model_parameters[0]
+
+    # match torch.optim.Optimizer expectations
+    if not isinstance(first, (torch.Tensor, dict)):
+        raise TypeError("param argument that would be given to the optimizer should be "
+                        f"an iterable of Tensors or dicts, but got {type(first)}")
+
+    # Case 1: model_parameters is a list of torch.nn.Parameter
```
tohtana, line 164: Can we briefly validate that all elements in `model_parameters` are the same class?

deepspeed/moe/utils.py
jeffra: Good question, I was also curious if we could assume this. But if you trace back where model_parameters is coming from you can see we force it to be a list here: 

https://github.com/microsoft/DeepSpeed/blob/731fd682996a0af56879c033ae3baec3db844e71/deepspeed/runtime/engine.py#L304

jeffra: added an assert here just in case the function is used outside of engine for some reason later

tohtana (approved):

PR #5355, opened 2024-04-02
OptimizedLinear implementation

jeffra:
Optimized version of `nn.Linear` that adds features such as:
      * LoRA w. base weight sharding
      * FP [6,8,12] quantization

Depends on #5336 being merged first

Co-authored-by: @rajhans
Co-authored-by: @aurickq

### DIFF
--- a/deepspeed/linear/config.py
+++ b/deepspeed/linear/config.py
@@ -1,0 +1,21 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+from dataclasses import dataclass
+
+
+@dataclass
+class LoRAConfig:
+    lora_r: int = 64
+    lora_alpha: float = 16
+    base_weight_sharding: int = 1
+
+
+@dataclass
+class QuantizationConfig:
+    q_bits: int = 8
+    rounding: str = "nearest"
+    mantissa_bits: int = 3
+    group_size: int = 512
--- a/deepspeed/linear/optimized_linear.py
+++ b/deepspeed/linear/optimized_linear.py
@@ -1,0 +1,141 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+import torch
+import math
+import torch.nn as nn
+
+from deepspeed.accelerator import get_accelerator
+import deepspeed.comm as dist
+
+from .config import LoRAConfig, QuantizationConfig
+from .quantization import QuantizedParameter, QuantizedLinear
+
+
+class OptimizedLinear(nn.Module):
+    """
+    Optimized version of nn.Linear that adds features such as:
+      * LoRA w. base weight sharding
+      * FP [6,8,12] quantization
+
+    Arguments:
+        input_dim: Required: size of each input sample
+        output_dim: Required: size of each output sample
+        bias: Optional: If set to False, the layer will not learn an additive bias. Default: False
+        lora_config: Optional: LoRAConfig defining lora features and base-weight-sharding degree
+        quantization_config: Optional: QuantizationConfig defining quantization features
+        dtype: Optional: parameter dtype, only supports bfloat16 currently
+
+    Returns:
+        Returns a new nn.Module depending on the input config. Either native
+        torch.nn.Linear, QuantizedLinear, or the full-featured DSOptimizedLinear.
+    """
+
+    def __new__(self,
+                input_dim: int,
+                output_dim: int,
+                bias: bool = False,
+                lora_config: LoRAConfig = None,
+                quantization_config: QuantizationConfig = None,
+                dtype=torch.bfloat16):
+
+        if lora_config is None and quantization_config is None:
+            # Everything disabled, fall back to normal nn.Linear
+            self = nn.Linear(input_dim, output_dim, bias=bias)
+        elif lora_config:
+            # lora enabled w. optional quant
+            self = DSOptimizedLinear(input_dim=input_dim,
+                                     output_dim=output_dim,
+                                     bias=bias,
+                                     lora_config=lora_config,
+                                     quantization_config=quantization_config,
+                                     dtype=dtype)
+        elif quantization_config:
+            # only quantization enabled, no lora
+            self = QuantizedLinear(input_dim=input_dim,
+                                   output_dim=output_dim,
+                                   bias=bias,
+                                   quantization_config=quantization_config,
+                                   dtype=dtype)
+        return self
+
+
+class DSOptimizedLinear(nn.Module):
+
+    def __init__(self,
+                 input_dim: int,
+                 output_dim: int,
+                 bias: bool = False,
+                 lora_config: LoRAConfig = None,
+                 quantization_config: QuantizationConfig = None,
+                 device=None,
+                 dtype=torch.bfloat16):
+        super().__init__()
+        self.input_dim = input_dim
+        self.output_dim = output_dim
+        self.bias = bias
+        self.lora_config = lora_config
+        self.quantization_config = quantization_config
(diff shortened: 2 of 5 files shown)


deepspeed/linear/optimized_linear.py
```diff
@@ -0,0 +1,141 @@
...
+        torch.nn.Linear, QuantizedLinear, or the full-featured DSOptimizedLinear.
+    """
+
+    def __new__(self,
+                input_dim: int,
+                output_dim: int,
+                bias: bool = False,
+                lora_config: LoRAConfig = None,
+                quantization_config: QuantizationConfig = None,
+                dtype=torch.bfloat16):
```
tjruwase, line 42: If only bfloat16 is supported then I think the following is required to avoid confusing errors to users:
1. dtype should not have a default value to avoid injecting unexpected dtype (i.e., bfloat16) into execution.
2. Assertion that dtype is bfloat16, or a supported type.

deepspeed/linear/optimized_linear.py
```diff
@@ -0,0 +1,141 @@
...
+
+        if lora_config is None and quantization_config is None:
+            # Everything disabled, fall back to normal nn.Linear
+            self = nn.Linear(input_dim, output_dim, bias=bias)
```
tjruwase, line 46: Since dtype is not required for this baseline case, the comments should reflect this. 

Alternatively, since dtype seems to apply only for quantization, I wonder if it should be moved to `LoRAConfig` and `QuantizationConfig` data types to make this connnection explicit.

deepspeed/linear/optimized_linear.py
```diff
@@ -0,0 +1,141 @@
...
+        elif lora_config:
+            # lora enabled w. optional quant
+            self = DSOptimizedLinear(input_dim=input_dim,
+                                     output_dim=output_dim,
+                                     bias=bias,
+                                     lora_config=lora_config,
+                                     quantization_config=quantization_config,
+                                     dtype=dtype)
+        elif quantization_config:
```
tjruwase, line 55: It seems that the intention is for user to specify at most one of `lora_config` or `quantization_config`, but not both. However, the current logic allows users to specify both, and future reordering of the conditionals will change code execution in unexpected ways. I feel that a cleaner design would be to pass exactly one quantization config (or None for baseline behavior), and for the logic to figure out which kind of quantization is needed. Happy to chat more in case I am missing something.

deepspeed/linear/optimized_linear.py
```diff
@@ -0,0 +1,141 @@
...
+        self.quantization_config = quantization_config
+        device = get_accelerator().current_device() if device is None else device
+        assert self.lora_config is not None, "DSOptimizedLinear requires a LoRA config"
+
+        self.zero_shards = self.lora_config.base_weight_sharding
+        self.sharded_weight_size = int(float(self.input_dim) // self.zero_shards)
+        w = torch.nn.Parameter(torch.empty((self.output_dim, self.sharded_weight_size), dtype=dtype))
+        torch.nn.init.xavier_uniform_(w)
+        if self.quantization_config is not None:
+            self.base_weight = QuantizedParameter(w)
```
tjruwase, line 89: Should `self.quantization_config` not be passed to `QuantizedParameter(...)? In general, it seems `self.quantization_config` is not actually used in this class. Specifically, it is only tested as not being None.

deepspeed/linear/optimized_linear.py
```diff
@@ -0,0 +1,141 @@
...
+            # only quantization enabled, no lora
+            self = QuantizedLinear(input_dim=input_dim,
+                                   output_dim=output_dim,
+                                   bias=bias,
+                                   quantization_config=quantization_config,
+                                   dtype=dtype)
+        return self
+
+
+class DSOptimizedLinear(nn.Module):
```
tjruwase, line 65: I think that `LoRAOptimizedLinear` would be a more appropriate and descriptive name than `DSOptimizedLinear`

deepspeed/linear/config.py
```diff
@@ -0,0 +1,21 @@
...
+
+@dataclass
+class LoRAConfig:
+    lora_r: int = 64
+    lora_alpha: float = 16
+    base_weight_sharding: int = 1
+
+
+@dataclass
+class QuantizationConfig:
```
tjruwase, line 17: Is it possible to have a more descriptive name instead of `QuantizationConfig`? Afterall, LoRA is a form of quantization, so why is it treated differently?

deepspeed/linear/optimized_linear.py
tjruwase: On further reading, it seems that:
1. lora_config requires (uses) quantization_config 
2. quantization_config is self-contained. 

Can you please correct/confirm my understanding?

deepspeed/linear/quantization.py
```diff
@@ -0,0 +1,115 @@
...
+        if quantizer is not None:
+            self.quantizer = quantizer
+        else:
+            self.quantizer = FP_Quantize(group_size=quantization.group_size)
+        self._ensure_quantized(self)
+        return self
+
+    def _ensure_quantized(self, tensor: torch.Tensor):
+        # If the tensor is on a cuda device and is not quantized, then quantize it in-place.
+        if tensor.device.type == "cuda" and tensor.dtype != torch.int8:
```
tjruwase, line 46: What is expected behavior on non-cuda devices?

tests/unit/linear/test_linear.py
```diff
@@ -0,0 +1,120 @@
...
+        batch_size = 5  # Number of samples in a batch
+
+        lora_config = LoRAConfig(lora_r=16, lora_alpha=16, base_weight_sharding=base_weight_sharding)
+        quant_config = QuantizationConfig(q_bits=q_bits)
+
+        linear_layer = OptimizedLinear(input_dim=input_features,
+                                       output_dim=output_features,
+                                       lora_config=lora_config,
+                                       quantization_config=quant_config)
+        linear_layer = linear_layer.to(f"cuda:{rank}")
```
tjruwase, line 117: Do we expect this UT to run only on cuda devices?

tests/unit/linear/test_linear.py
```diff
@@ -0,0 +1,120 @@
...
+        batch_size = 5  # Number of samples in a batch
+
+        lora_config = None
+        quant_config = QuantizationConfig(q_bits=q_bits)
+
+        linear_layer = OptimizedLinear(input_dim=input_features,
+                                       output_dim=output_features,
+                                       lora_config=lora_config,
+                                       quantization_config=quant_config)
+        linear_layer = linear_layer.to(f"cuda:{rank}")
```
tjruwase, line 90: Is this a cuda-device only UT?

deepspeed/linear/quantization.py
```diff
@@ -0,0 +1,115 @@
...
+        new_instance.data = copy.deepcopy(state["data"])
+        return new_instance
+
+    def __copy__(self):
+        new_instance = type(self).__new__(type(self))
+        state = self.__getstate__()
+        new_instance.__setstate__(state)
+        return new_instance
+
+    def cuda(self, device=None, non_blocking=False):
```
tjruwase, line 85: Since our CI now also runs on non-CUDA devices, we need to clarify expectation of this logic on non-CUDA devices.

deepspeed/linear/optimized_linear.py
jeffra: The if-condition here is a bit strange, I updated the comment around the `LoRAOptimizedLinear` to hopefully better clarify. In general though `OptimizedLinear` should support all 4 combos of lora_config or quant_config being enabled or disabled (TT,TF,FT,FF). In the case of `LoRAOptimizedLinear` it can support LoRA with or without a quant_config. Hope this helps clarify along with the rest of the updates I've just pushed.

tjruwase (approved):

PR #5358, opened 2024-04-03
Increase CI coverage for Gaudi2 accelerator.

vshekhawat-hlab:
This change increase CI coverage for HPU accelerator.


.github/workflows/hpu-gaudi2.yml
```diff
@@ -5,10 +5,11 @@ on:
   schedule:
     - cron: "0 0 * * *"
   pull_request:
-    paths:
-      - ".github/workflows/hpu-gaudi2.yml"
-      - "accelerator/hpu_accelerator.py"
-
+    paths-ignore:
```
loadams, line 8: Hi @vshekhawat-hlab - I'm not sure we want to run on every commit, we run nightly for this reason, to reduce the total scope of tests run on each PR, and should be able to identify any regression that way.  Do you feel this needs to be run on all PRs?  My concern is there isn't enough capacity (1 machine) supporting this test, so we will queue on waiting for these tests to pass.

vshekhawat-hlab: Thank you, @loadams , for notifying us of the issues for running Gaudi2 CI with every commit. Let me check and get back to you on this. We aim to increase coverage for Gaudi2 CI to reduce the pain of bisection.

loadams: Certainly, I understand that bisecting commits can be annoying, though we don't tend to merge that many commits per day.  But we can certainly add more path filters too

.github/workflows/hpu-gaudi2.yml
```diff
@@ -8,7 +8,23 @@ on:
...
+      - "deepspeed/runtime/zero/partitioned_param_coordinator.py"
+      - "deepspeed/runtime/zero/parameter_offload.py"
+      - "deepspeed/runtime/pipe/engine.py"
+      - "deepspeed/runtime/utils.py"
+      - "deepspeed/inference/engine.py"
+      - "deepspeed/module_inject/auto_tp.py"
+      - "deepspeed/module_inject/replace_module.py"
+      - "deepspeed/module_inject/load_checkpoint.py"
+      - "deepspeed/module_inject/inject.py"
+      - "op_builder/hpu/**"
```
loadams, line 25: Hi @vshekhawat-hlab - thanks, this looks better to me.  One comment, could you move the `op_builder/hpu` directory up with `hpu_accelerator.py` file near the top, that will make it more readable by having the HPU specific files at the top, then the general DeepSpeed files next?

loadams (approved):

.github/workflows/hpu-gaudi2.yml
loadams: And could you add a second new line between the paths list and the concurrency section in the YML?

loadams: Nevermind, we don't need that second newline, and I updated the order.

PR #5363, opened 2024-04-04
Add throughput timer configuration

deepcharm:
The new "timers" section describes configuration for different timers.

Specifically, in the "throughput" section, it is possible to disable the throughput timer (enabled by default). This allows to avoid the performance degradation whenever the throughput measurement is not needed, for example in production environment.

No device synchronize() is invoked when "synchronized" is set to False (default is True). This allows to produce approximate throughput measurements with minimal performance penalty.


deepcharm:
We've discovered the following issues in the current implementation of the Throughput timer:

- The timer invokes synchronize() twice on each step at [start](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/utils/timer.py#L240) and  [stop](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/utils/timer.py#L252).
  
- Calling synchronize() ensures that all workloads running on the device complete.
This makes sure that the measured time includes both the elapsed CPU and the device times.
 

- However, being an expensive operation, synchronize() causes the performance drop of up to 20-25%
on some accelerators (e.g. HPU). On the other hand, that avoiding the call to synchronize()
still produces a meaningful performance estimate without performance drop.
    
- Since there is no configuration flag to disable the timer, today a user pays consistent performance penalty,
even in cases when the throughput measurement may be unnecessary (e.g. production).

The purpose of this PR is to add config options to control the above behavior.

deepspeed/utils/config.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+from deepspeed.runtime.config_utils import get_scalar_param, DeepSpeedConfigObject
+from deepspeed.utils.constants import *
+
+
+class DeepSpeedThroughputTimerConfig(DeepSpeedConfigObject):
```
mrwyattii, line 10: This should use the `DeepSpeedConfigModel` rather than `DeepSpeedConfigObject` to take advantage of config validation via pydantic: https://github.com/microsoft/DeepSpeed/blob/4d4c99d9e51afa01a8f4fc7d59a0bc09128abf3e/deepspeed/runtime/config_utils.py#L16

mrwyattii (changes requested):

deepspeed/utils/config.py
deepcharm: @mrwyattii, the requested changes have been made. Thanks.

deepspeed/runtime/config.py
```diff
@@ -66,6 +66,8 @@
 from .data_pipeline.config import get_data_efficiency_enabled, get_data_efficiency_config, get_curriculum_enabled_legacy, get_curriculum_params_legacy
 from .data_pipeline.constants import *
 
+from ..utils.config import get_timers_config
```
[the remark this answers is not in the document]
deepcharm: @loadams  - thanks, done.

lekurile (approved):

nelyahu:
@loadams @mrwyattii @lekurile can this change be merged?

loadams:
> @loadams @mrwyattii @lekurile can this change be merged?

Sorry for the delay @nelyahu - I will make sure this is merged shortly.

mrwyattii (approved):

loadams (approved):

PR #5368, opened 2024-04-05
Switch to shared function to check torch version

loadams:


deepspeed/comm/torch.py
```diff
@@ -9,6 +9,7 @@
 from .backend import *
 from .comm import *
 from ..runtime import compiler
+from ..runtime.utils import required_torch_version
```
loadams, line 12: Evaluate moving this to utils to resolve the circular import issues.

PR #5376, opened 2024-04-08
Add CI for Intel XPU/Max1100

Liangliang-Ma:
Add basic workflow for tests on intel xpu. Currently we have part of tests enabled. We will add more tests in later PRs.


.github/workflows/xpu-max1100.yml
```diff
@@ -0,0 +1,59 @@
+name: xpu-max1100
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
+  pull_request:
```
loadams, line 7: @Liangliang-Ma - I'm not sure we would want to run this on every PR at the start?  Perhaps if we can add it to just run on PRs when certain files are modified?

Also it appears the runner isn't created yet.

Liangliang-Ma: Thanks for review. I have added a initial scope for the CI.

loadams (approved):

PR #5386, opened 2024-04-09
Update path name on xpu-max1100.yml, add badge in README

loadams:
- Fixes xpu-max1100 not running on PR because of incorrect yml name.


.github/workflows/xpu-max1100.yml
```diff
@@ -6,7 +6,7 @@ on:
     - cron: "0 0 * * *"
   pull_request:
     paths:
-      - ".github/workflows/xpu.yml"
+      - ".github/workflows/xpu-max1100.yml"
```
loadams, line 9: FYI @Liangliang-Ma this is why it didn't run on your PR originally but it is now.

PR #5390, opened 2024-04-10
Add custom reshaping for universal checkpoint

tohtana:
This PR adds more flexibility to define weight tensor reshaping for universal checkpointing.

Currently universal checkpointing assumes a few patterns of partitioning for tensor parallelism, such as column/row wise partitioning of a 2-dim tensor. However, these are not flexible enough to define partitioning for more complex usages. Here are some examples:

1) MoE: The user may define the weight tensor for MoE's FFN as [n_experts * hidden_out, hidden_in]. For TP, we need to *view* this tensor as 3-dim tensor and partition it along `hidden_out` dimension. 
2) GQA: The weights for QKV are often represented as one tensor and we may have Q, K and V with different sizes. The tensor shape will be [q_size + k_size + v_size, hidden]. We partition this along first dimension but for each Q, K, and V. In this case, we first need to partition Q, V, and V separately and then concatenate them to get a shard for TP.

We propose a new pattern `PARAMETER_WITH_SUB_PARAMS` to support this. Here is the usage to cover the above use cases. You can define the view of the weight tensor and specify the dimension for partitioning based on the view.
```python
from deepspeed.checkpoint import PARAMETER_WITH_SUB_PARAMS, SubparamShape

info[PARAMETER_WITH_SUB_PARAMS] = [
    asdict(SubparamShape(patterns=[layers_prefix + r"\d+moe.fc1.weight"],
                shape=(num_experts, hidden_out, hidden_in), partition_dim=1)),
    asdict(SubparamShape(patterns=[layers_prefix + r"\d+.qkv.weight"],
                shape=((q_size, k_size, v_size), hidden_size), partition_dim=0)),
...
]
```

The conversion script (`ds_to_universal.py`) merges TP-sharded weight tensors and the loader of universal checkpoints also partitions them following the information.


deepspeed/checkpoint/constants.py
```diff
@@ -74,10 +74,14 @@
...
+SUB_PARAM_SHAPE = "sub_param_shape"
+
 # Regex list of parameters that require special handling
 VOCABULARY_PARAMETER_PATTERNS = 'vocabulary_parameter_patterns'
 PIPELINE_REPLICATED_PARAMETER_PATTERNS = 'pipeline_replicated_parameter_patterns'
 PARAMETER_TO_AVERAGE_PATTERNS = 'parameter_to_average_patterns'
 PARAMETER_WITH_ROW_PARALLELISM_PATTERNS = 'parameter_with_row_parallelism_patterns'
 TP_REPLICATED_PARAMETER_PATTERNS = 'tp_replicated_parameter_patterns'
 PARAMETER_WITH_2_SUB_PARAMS_CAT_DIM_0 = 'parameter_with_2_sub_params_cat_dim_0'
+PARAMETER_WITH_SUB_PARAMS = 'parameter_with_sub_params'
```
tjruwase, line 86: @tohtana, I really like these new constructs: `PARAMETER_WITH_SUB_PARAMS` and `PARAMETER_WITH_SUB_PARAMS`. It seems to me that they can capture a broad range of patterns. For example, it seems they can represent the following: 
1. `PARAMETER_WITH_ROW_PARALLELISM_PATTERNS`,
2. `PARAMETER_WITH_2_SUB_PARAMS_CAT_DIM_0`

Am I correct?

stas00:
Thank you for the heads up, Tunji. 

I'm yet to experiment with these new types of models so I'm probably not the right person to ask. If and when I get to use those I will gladly send feedback back.

tjruwase:
@minjiaz, @xylian86, @zhangninja FYI

mosheisland:
> @mosheisland, @stas00, FYI, in this PR, @tohtana introduces new partitioning specification that seems to work for MoE and GQA. I feel this approach could generalize to more patterns including some of the existing ones. We will really appreciate your feedback. Thanks!

@tjruwase - I agree. PARAMETER_WITH_SUB_PARAMS is flexible and can replace both patterns that you mentioned. Good idea.

tohtana:
@tjruwase Sorry for my late reply. As this PR was already merged, let me answer here.

>  It seems to me that they can capture a broad range of patterns. For example, it seems they can represent the following existing patterns: `PARAMETER_WITH_ROW_PARALLELISM_PATTERNS`, `PARAMETER_WITH_2_SUB_PARAMS_CAT_DIM_0`

Yes, you are right. These patterns are represented by using this new pattern `PARAMETER_WITH_SUB_PARAMS`.
But I think we should keep these patterns as a simple way to cover major use cases.

Let me also add a visualization of these use cases.
```
info[PARAMETER_WITH_SUB_PARAMS] = [
    asdict(SubparamShape(patterns=[layers_prefix + r"\d+moe.fc1.weight"],
                shape=(num_experts, hidden_out, hidden_in), partition_dim=1)),
    asdict(SubparamShape(patterns=[layers_prefix + r"\d+.qkv.weight"],
                shape=((q_size, k_size, v_size), hidden_size), partition_dim=0)),
...
]
```

These figures show an MoE exmpale with 4 experts/TP size=2 and a GQA example.

image

PR #5391, opened 2024-04-10
[CPU] Support SHM based inference_all_reduce in TorchBackend

delock:
This PR adds SHM based `inference_all_reduce` kernel to `TorchBackend` communication backend.  When inference on CPU server, this path replaces default `torch.distributed.all_reduce` which eventurally use gloo backend.   This PR will improve inference performance with AutoTP when only stock PyTorch is installed without Intel Extension for PyTorch.

Compared with gloo backend.  SHM based inference_all_reduce kernel is a more directed path and perform much better on single node.

| message size | gloo all_reduce(ms) | SHM all_reduce(ms) |
| --- | --- | --- |
| 32MB | 30.7 | 0.65 |
| 64KB | 0.23 | 0.028 |

In text generation of bloom-3b with AutoTP, average token latency improved 1.45x with this PR on 2S Xeon node.


delock:
Hi @loadams  the formatting error had been fixed, thanks!

csrc/cpu/comm/ccl.cpp
```diff
@@ -263,11 +263,6 @@ void inference_all_reduce(torch::Tensor& data, py::object op, bool async_op)
     auto start = std::chrono::system_clock::now();
 #endif
 
-    static py::object ReduceOp = py::module_::import("deepspeed.comm").attr("ReduceOp");
```
tjruwase, line 266: I think it is confusing to drop this assert since `inference_all_reduce` only works for sum even though it is not clear from the name. Users can easily assume that this function is similar to more common ones like nccl allreduce that supports other user-defined reduction ops like MAX. I see two possible solutions:
1. Restore this assert so that users are notified of unsupported ops, and developers can eventually support more ops without changing API.
2. Rename function to something like `inference_all_reduce_sum` to better reflect the intention.

op_builder/cpu/comm.py
```diff
@@ -42,3 +42,30 @@ def extra_ldflags(self):
             return []
         else:
             return ['-lccl', f'-L{ccl_root_path}/lib']
+
+
+class SHMCommBuilder(CPUOpBuilder):
```
tjruwase, line 47: @delock, please feel free to disregard this optional feedback. I think it would be more readable to rename `SHM` to something like `SharedMem`. The first thing that comes to mind when reading the code is `shim` :) https://stackoverflow.com/questions/2116142/what-is-a-shim

I am probably the only person affected like this, so please feel free to disregard.

delock: Hi @tjruwase , thanks for the comment.   I'll change the class name to make it more clear.

csrc/cpu/comm/ccl.cpp
delock: Hi @tjruwase I think leave some space for developer to extend might be a better option.  Let me add the assertion back.

tjruwase (approved):

PR #5396, opened 2024-04-11
add device config env for the accelerator 

shiyuan680:
Thank you for [pr](https://github.com/microsoft/DeepSpeed/pull/5369) and @delock  contribution of ideas.
As mentioned in this [pr](https://github.com/microsoft/DeepSpeed/pull/5369), each device has its own environmental variables.
We create visible_devices_envs() and set_visible_devices_envs() methods on the accelerator class to enable each accelerator to implement env settings within the interface , which is more generic to other accelerators.

this commit has tested on npu, each one has 8 ascend npus


shiyuan680:

eigen2017:
great thanks to HW and Microsoft，as ascend npu customer, we'll continuously provide contributions.

CurryRice233:
> great thanks to HW and Microsoft，as ascend npu customer, we'll continuously provide contributions.

Thanks to you too :)

CurryRice233:
@delock @nelyahu @mrwyattii we added an accelerator method to get the env of the visible device, we don't know the env of other accelerators.

delock:
For XPU, the name of env is `ZE_AFFINITY_MASK`.

I think it would be better to define interface as `set_visible_devices(list)` and let each accelerator implement setting of env inside the interface.   This will give each accelerator flexibility to set visible devices in different ways.

tjruwase:
> I think it would be better to define interface as `set_visible_devices(list)` and let each accelerator implement setting of env inside the interface. This will give each accelerator flexibility to set visible devices in different ways.

@shiyuan680, thanks for creating this PR. Can you please add this interface suggested by @delock?

@delock and @BacharL, thanks for assisting with this PR.

accelerator/cpu_accelerator.py
```diff
@@ -320,3 +320,10 @@ def build_extension(self):
 
     def export_envs(self):
         return []
+
+    def visible_devices_envs(self):
+        return ['CUDA_VISIBLE_DEVICES']
```
tjruwase, line 325: @delock, I suspect that `CUDA_VISIBLE_DEVICES` is not correct for cpu_accelerator?

accelerator/mps_accelerator.py
```diff
@@ -258,3 +258,10 @@ def build_extension(self):
 
     def export_envs(self):
         return []
+
+    def visible_devices_env(self):
+        return ['CUDA_VISIBLE_DEVICES']
```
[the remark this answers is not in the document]
loadams: I'm not able to find one in their docs, but for most of their devices so far, they only have 1 GPU so a warning/TODO probably makes the most sense.

CurryRice233:
@loadams CI has passed, would you please help to review this PR and add to merge queue ?

tjruwase:
> @loadams CI has passed, would you please help to review this PR and add to merge queue ?

@CurryRice233, can you help to address my questions? Thanks!

eigen2017:
> > @loadams CI has passed, would you please help to review this PR and add to merge queue ?
> 
> @CurryRice233, can you help to address my questions? Thanks!

yes it is unknown that how to set cpu and mps env，both setting CUDA_VISIBLE_DEVICES, i suppose it is means to keep the same logic to the old version of deepspeed for mps and cpu. 

so is it better to add annotations over the code to explain why set CUDA_VISIBLE_DEVICES ?

tjruwase:
> so is it better to add annotations over the code to explain why set CUDA_VISIBLE_DEVICES ?

@eigen2017, yes annotation is reasonable for this. Thanks for your explanation.

eigen2017:
> > so is it better to add annotations over the code to explain why set CUDA_VISIBLE_DEVICES ?
> 
> @eigen2017, yes annotation is reasonable for this. Thanks for your explanation.

i don't have write access to microsoft/DeepSpeed, could anyone help to add new commits to branch [shiyuan680:master], and add annotations to [mps_accelerator.py] and [cpu_accelerator.py] ? @tjruwase @CurryRice233 @loadams @shiyuan680 

explainations can like:
"TODO: before mps' and cpu's visible envs are confirmed, keep the same old env setting as CUDA_VISIBLE_DEVICES "

accelerator/cpu_accelerator.py
```diff
@@ -322,3 +322,10 @@ def build_extension(self):
 
     def export_envs(self):
         return []
+
+    def visible_devices_envs(self):
```
tjruwase, line 326:
```suggestion
    TODO: cpu's visible envs is confirmed, keep as CUDA_VISIBLE_DEVICES
    def visible_devices_envs(self):
```

accelerator/mps_accelerator.py
```diff
@@ -258,3 +258,11 @@ def build_extension(self):
 
     def export_envs(self):
         return []
+
+    def visible_devices_envs(self):
```
tjruwase, line 262:
```suggestion
    TODO: mpu's visible envs is confirmed, keep as CUDA_VISIBLE_DEVICES
    def visible_devices_envs(self):
```

tjruwase (approved):

eigen2017:
@tjruwase thank you for helping us~~thank you~

CurryRice233:
@eigen2017 thank you for your explanation, and @tjruwase thank you for the annotation. :)

PR #5408, opened 2024-04-12
Update engine.py to avoid torch warning

etiennebonnafoux:
The state_dict function of module.py from torch write a warning if arguments are positional arguments and not keyword arguments


deepspeed/runtime/engine.py
```diff
@@ -2542,7 +2542,7 @@ def all_gather_scalar(self, value, dp_group):
         return tensor_list
 
     def module_state_dict(self, destination=None, prefix="", keep_vars=False, exclude_frozen_parameters=False):
-        sd = self.module.state_dict(destination, prefix, keep_vars)
+        sd = self.module.state_dict(destination = destination, prefix = prefix,keep_vars = keep_vars)
```
loadams, line 2545: Hi @etiennebonnafoux - thanks for this change, if you run our formatter, it will remove the spaces between the variables and the = it will pass that check and we can get this merged.  Thanks!

loadams (approved):

deepspeed/runtime/engine.py
etiennebonnafoux: I am new to PR (this is my second) so it tooks me some time.

loadams: Not a problem, thanks for submitting the fix, we will ge this merged shortly.

Issue #5276, opened 2024-03-14
[REQUEST] Democratizing Present Day LLM Fine-Tuning To The GPU Poor


NicolasMejiaPetit:
Dear DeepSpeed developers,

[...]

PR #1310, opened 2021-08-17
DeepSpeed MoE 

awan-10:
This PR introduces DeepSpeed Mixture of Experts (MoE) support. The code has been written in collaboration with many contributors at Microsoft including the Z-code team.

PR #5409, opened 2024-04-14
CPUAdam fp16 and bf16 support

BacharL:
Hi.
Please review the following changes
I added support for BF16 to cpu adam. BF16, FP16 and float are supported at compilation time. the correct template is called at runtime according to input params dtype.


csrc/includes/cpu_adagrad.h
```diff
@@ -12,21 +12,29 @@
...
 #include <cuda_runtime_api.h>
 #include "cuda.h"
 #include "custom_cuda_layers.h"
-typedef __half ds_half_precision_t;
 #elif defined(__ENABLE_CANN__)
 #include "acl/acl.h"
 #include "torch_npu/csrc/core/npu/NPUStream.h"
-typedef c10::Half ds_half_precision_t;
 #else
 typedef unsigned short ds_half_precision_t;
```
tjruwase, line 31: Does line 34 makes this dead code?

csrc/includes/cpu_adagrad.h
```diff
@@ -130,6 +138,9 @@ void Adagrad_Optimizer::Step_AVX(size_t* rounded_size,
                                  ds_half_precision_t* dev_params,
                                  bool half_precision)
 {
+#if !defined(__AVX512__)
+    if (std::is_same_v<T, c10::BFloat16>) { return; }
```
tjruwase, line 142: Why is it safe to return early and skip computation?

csrc/includes/cpu_adagrad.h
```diff
@@ -12,21 +12,29 @@
...
-typedef __half ds_half_precision_t;
 #elif defined(__ENABLE_CANN__)
 #include "acl/acl.h"
 #include "torch_npu/csrc/core/npu/NPUStream.h"
-typedef c10::Half ds_half_precision_t;
 #else
 typedef unsigned short ds_half_precision_t;
 #endif
 
+typedef HALF_DTYPE ds_half_precision_t;
```
tjruwase, line 34: Does this line 34 make line 31 above dead?

deepspeed/ops/adagrad/cpu_adagrad.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, model_params, lr=1e-2, eps=1e-10, weight_decay=0, amsgrad=Fal
         self.opt_id = DeepSpeedCPUAdagrad.optimizer_id
         DeepSpeedCPUAdagrad.optimizer_id = DeepSpeedCPUAdagrad.optimizer_id + 1
         self.fp32_optimizer_states = fp32_optimizer_states
-        self.ds_opt_adagrad = CPUAdagradBuilder().load()
+        self.ds_opt_adagrad = CPUAdagradBuilder().set_dtype(self.param_groups[0]['params'][0].dtype).load()
```
tjruwase, line 22: I am unconvinced that the new `set_dtype` is needed. Could `dtype` not simply be passed into constructor to set `self.dtype`?

op_builder/cpu/cpu_adam.py
```diff
@@ -12,6 +12,13 @@ class CPUAdamBuilder(CPUOpBuilder):
 
     def __init__(self):
         super().__init__(name=self.NAME)
+        self.dtype = None
+
+    def set_dtype(self, dtype):
+        import torch
+        assert (dtype in [torch.bfloat16, torch.half, torch.float32])
```
tjruwase, line 19: I think this check would be better in `cxx_args()` and other compilation code paths, even if repeated.

csrc/includes/cpu_adam.h
```diff
@@ -13,22 +13,29 @@
...
 #elif defined(__ENABLE_CANN__)
 #include "acl/acl.h"
 #include "torch_npu/csrc/core/npu/NPUStream.h"
-typedef c10::Half ds_half_precision_t;
 #else
 #include <cmath>
-typedef unsigned short ds_half_precision_t;
 #endif
 
+typedef HALF_DTYPE ds_half_precision_t;
```
tjruwase, line 35: @BacharL, this amazing PR of yours has made me realize that `ds_half_precision_t` was not well thought out for optimizer offloading. It did not anticipate that device type could be anything other than `fp16` and `bf16`. Consequently, we now require users to set the following confusing compiler option: `-DHALF_DTYPE=float`. 
In retrospect, it seems `ds_device_precision_t` is a better name. 

Similarly, the `half_precision` bool variables scattered around the code is also redundant. Thanks for removing some of them in this PR.  

@BacharL, what do you think?

op_builder/cpu/fused_adam.py
```diff
@@ -12,12 +12,31 @@ class FusedAdamBuilder(CPUOpBuilder):
...
+        self.dtype = dtype
+        return self
 
     def absolute_name(self):
         return f'deepspeed.ops.adam.{self.NAME}_op'
 
     def sources(self):
         return ['csrc/cpu/adam/fused_adam.cpp', 'csrc/adam/cpu_adam_impl.cpp']
 
+    def cxx_args(self):
```
tjruwase, line 29: Given the identical nature of `cxx_args` across multiple ops, is it possible to refactor to avoid duplication?

tjruwase:
@BacharL, thanks for this incredible improvement to the offloading optimizers and op builders. I left a few comments and questions, but overall looks good to me.

csrc/includes/cpu_adagrad.h
BacharL: Before returning, Step_AVX sets rounded_size=new_rounded_size so the remaining part of the buffer will be handled in non AVX fashion. returning early here would make the whole buffer fallback to non AVX path.

csrc/includes/cpu_adam.h
BacharL: I will fix the naming of these, however, I think ds_half_precision_t causes discrepancy in buffer handling.
Step_AVX will handle the buffer with on cpu with FP16 of BF16 as implemented in AVX. The cast to fp32 is done in simd.h SIMD_LOAD_XX macros. 
But the remaining part of the buffer will be handled in Step_1 with dtype from the device, which may be __half or __bfloat16. This dtype may not provide bit exact results as CPU implementation. The cast to fp32 is done in Step_1 inside the main loop.
I think even AVX/Native C++ cast may not be compatible. So if we already have this issue, why do we need to add device dtypes here? we can limit this code to c10::Half or c10::BFloat16 only. 

We can also remove more half_precision varaibles by passing _params and dev_params with correct type instead of float* and fix their usage inside cuda and avx code

op_builder/cpu/fused_adam.py
BacharL: removed all cxx_args changes. code now define BF16 and FP16 dtype in header.

op_builder/cpu/cpu_adam.py
BacharL: removed set_dtype. no need to pass dtype during build time.

deepspeed/ops/adagrad/cpu_adagrad.py
BacharL: removed set_dtype function

csrc/includes/cpu_adagrad.h
BacharL: fixed

BacharL:
@tjruwase Thanks for reviewing this change. I have made changes to address your comments.
Now there is no need to pass HALF_DTYPE as compiler define. all functions will be templated according to ds_device_precision_t.
removed all half_precision parameters.

csrc/includes/cpu_adam.h
tjruwase: Thanks for raising questions about the discrepancies in this code base. I am aligned that the code is unduly complicated to the point of being bug prone. I think this occurred because of the uncontrolled evolution of the code. Hopefully, some measure of cleanliness can be restored through careful usage of templates, and torch data type support. In the meantime, I will answer your question below. 

> I think even AVX/Native C++ cast may not be compatible. So if we already have this issue, why do we need to add device dtypes here? we can limit this code to c10::Half or c10::BFloat16 only

I think adding device dtype is useful for making the data type conversions of offloading computation explicit. 

1. Forward/backward computation on the device using device_dtype (currently fp16, bf16, or fp32) with output of gradients in device_dtype. 
2. Conversion step of gradients from device_dtype to host_dtype which is input for subsequent optimizer computation on host.
3. Optimizer step computation on host using host_dtype (currently fp32) and returning param of host_dtype.  
4. Conversion of (updated) host_dtype param into device_dtype param which is input for subsequent forward/backward computation on device.

csrc/adagrad/cpu_adagrad.cpp
```diff
@@ -21,26 +21,19 @@ static std::unordered_map<int, std::shared_ptr<void>> s_optimizers;
 
 // C++ interface
 
-void Adagrad_Optimizer::Step_1(float* _params,
-                               float* grads,
+template <typename T, typename ds_device_precision_t>
+void Adagrad_Optimizer::Step_1(T* _params,
+                               T* grads,
                                float* _exp_avg_sq,
```
tjruwase, line 27: This for consistency. 
```suggestion
                               T* _exp_avg_sq,
```

BacharL: Thanks for catching this.
Since fp32_optimizer_states may be controlled from configuration independently of params and grads dtype I think I need to add a third template argument T2 for state['exp_avg_sq']. Now the template will be controlled by 3 types, I think I need to find a way to call it at runtime without checking for each tensor type at runtime in ds_adagrad_step.
Also, the code in ds_adagrad_step_plus_copy is even more complicated and not even used since this path is disabled in stage_1_and_2.py::_optimizer_step()

BacharL: These are all the possible type 
```
params+grads, opt_states, fp16_params
float       , float     , half
float       , float     , bfloat
half        , half      , half
bfloat      , bfloat    , bfloat
half        , float     , half
bfloat      , float     , bfloat
```
Since the number is limited, I think I will expand the if statements in ds_adagrad_step (and matching functions for adam and lion) in order to call the correct template.
Is it ok?

tjruwase: The requirements are actually simpler than you think for the following reasons. 
1. `params`, `grads`, and `opt_states` must have the same dtype. 
2. Clarification from @RezaYazdaniAminabadi confirms that template usage should eliminate `half_precision` argument and associated logic. 
3. As you observed, `ds_adagrad_step_plus_copy` is not used, as we were never able to realize the potential perf benefits. I am happy to deprecate this API if that makes things easier for you. 

So, the requirements can be expressed in terms of host vs device states:
1. host-side states = `params`, `grads`, and `opt_states`
2. device-side state = `fp16_params`

One caveat is that the above is true only for mixed-precision training, since DS also supports fp32 training. 

Furthermore, the host-side states are of the same type, i.e., `ds_host_precision_t`, while the device state is of `ds_device_precision_t`. These two typenames can be instantiated as follows (for mixed precision training):

```python
  params+grads+opt_states,      fp16_params
  float                                half
  float                                bfloat
  half                                 half
  bfloat                             bfloat                                         
```

Please let me know your thoughts and whether the above is helpful. 

@RezaYazdaniAminabadi, please share your thoughts as well.

RezaYazdaniAminabadi: All good here, thanks @tjruwase for the clarification. Just wanna confirm that even though the optimizer states are sometimes saved in bit16 precision, we still wanna do the computation in fp32. So, maybe we can have a template argument as `compute_type`.

RezaYazdaniAminabadi (commented):
the PR looks good to me, thanks a lot @BacharL for working on this and templating different parts. Adding bfloat16 is definitely important for CPU-Adam :)

csrc/adagrad/cpu_adagrad.cpp
BacharL: @tjruwase 
Regarding 1, I see there is an option to set cpu_offload=True, fp_32_optimizer_states=True and fp16_master_weights_And_gradients=True. In this mode params+grads will be half and opt_states will be float. A different template type is required for each.

Regarding 3, if we deprecate ds_adagrad_step_plus_copy we don't need ds_device_precision_t anymore, since device_params is always passed as nullptr. Can we also remove all related code under __ENABLE_CUDA__ and __ENABLE_CANN__ ?

BacharL:
Added templated invoker to help selecting the implementation
The map stores function pointers to templated functions, the key is the type enum. At initialization all supported dtypes are templated and inserted into the map.
I didn't clean ds_adagrad_step_plus_copy and related code under `__ENABLE_CUDA__ ` but also couldn't test it.

csrc/adagrad/cpu_adagrad.cpp
tjruwase: > Regarding 1, I see there is an option to set cpu_offload=True, fp32_optimizer_states=True and fp16_master_weights_And_gradients=True. In this mode params+grads will be half and opt_states will be float. A different template type is required for each.

@BacharL, I was unaware of this option, even though @RezaYazdaniAminabadi alluded to it above. My search suggests it was introduced in the MoE PR #1310. In my opinion, we should deprecate this option as the costs outweigh the benefits. I would like to be convinced otherwise. I will now discuss the cost/benefit of this option as I see it. 

**Benefit**
The main benefit is reducing per-parameter optimizer footprint in CPU memory from 16 bytes to 12 bytes, i.e., 25% savings. There are also DtoH/HtoD b/w reductions, but I think those can also be realized in the baseline mode. Before discussing the downsides below, @RezaYazdaniAminabadi, @jeffra, and @awan-10 can you please share your thoughts on other benefits.

**Costs**
1. Use case is narrow: mixed-precision + offloading. Also, CPU memory savings is rarely critical compared to GPU memory.
2. Increased code complexity since the implementation has two manage 3 data types instead of 2 or 1 (if we further deprecate the `*_step_plus_copy()` API.
3. It complicates integration of other cpu-based optimizer implementations, such as torch Adam (works on both CPU and GPU). Ideally, CPUAdam, torch.Adam, (and others) should be drop-in replacements differing only in implementation details such as performance. However, introduction of [fp32_optimizer_states](https://github.com/microsoft/DeepSpeed/blob/0b224edcf7d83713b95ad6b989694a8bdf01809e/deepspeed/ops/adam/cpu_adam.py#L25) to support this option has deviated from torch API. 
4. It complicates our plans to support optimizers of arbitrary data types, besides fp32, such as 4-bit optimizers requested [here](https://github.com/microsoft/DeepSpeed/issues/5276). The problem is our deviation from the torch [approach](https://github.com/pytorch/pytorch/blob/82b7b59d2acd849cb16f2167fa3eaf7cabf95dcf/torch/optim/adam.py#L153-L159) of inferring optimizer state type from param type. 
5. It complicates model checkpointing, since [self.single_partition_of_fp32_groups](https://github.com/microsoft/DeepSpeed/blob/0b224edcf7d83713b95ad6b989694a8bdf01809e/deepspeed/runtime/zero/stage_1_and_2.py#L389-L402) which points to the master weights is widely used in checkpoint [save/load](https://github.com/microsoft/DeepSpeed/blob/0b224edcf7d83713b95ad6b989694a8bdf01809e/deepspeed/runtime/zero/stage_1_and_2.py#L2399). Right now, I am not sure how this option works with model checkpointing. 

Apologies for the very long post, I think it is important that we get this right. Thanks!

tjruwase: > Regarding 3, if we deprecate ds_adagrad_step_plus_copy we don't need ds_device_precision_t anymore since device_params is always passed as nullptr. Can we also remove all related code under `__ENABLE_CUDA__` and `__ENABLE_CANN__` ?

Yes, I generally love code removal whenever possible.

op_builder/hpu/fused_adam.py
```diff
@@ -4,10 +4,85 @@
 
 # DeepSpeed Team
 
-from .builder import CPUOpBuilder
+try:
```
tjruwase, line 7: @BacharL, is it possible to have a separate PR for this, since it is unrelated to CPUAdam? Splitting will greatly help future searches and references to the respective topics.

BacharL: yes, sorry, it went in in error. I will revert
https://github.com/microsoft/DeepSpeed/pull/5500

csrc/adagrad/cpu_adagrad.cpp
BacharL: > > Regarding 3, if we deprecate ds_adagrad_step_plus_copy we don't need ds_device_precision_t anymore since device_params is always passed as nullptr. Can we also remove all related code under `__ENABLE_CUDA__` and `__ENABLE_CANN__` ?
> 
> Yes, I generally love code removal whenever possible.

Cleaned up all device code from cpu optimizers
Now left with 2 template types, params+grads and optimizer state.

tjruwase: > Now left with 2 template types, params+grads and optimizer state.

Thanks @BacharL for the updates. Please note that if I get alignment on my proposal regarding 1, then we would be left with one template type for params+grads+optimizer state. However, I think that could be a separate PR, so as not to delay this any further. 

One request that I do have is for performance benchmarking of the new code to uncover any regressions. The code [here](https://github.com/microsoft/DeepSpeed/tree/master/tests/perf) might be useful starting point.

BacharL: Thanks for unblocking this PR

Benchmarked Adam and Adagrad using Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz
CPUAdam fp32  0.400s -> 0.399s
CPUAdam fp16  0.290s -> 0.283s
CPUAdam bf16  NA      -> 0.298s

CPUAdagrad fp32  0.291s -> 0.298s
CPUAdagrad fp16  NA      -> 0.180s
CPUAdagrad bf16  NA      -> 0.191s

tests/unit/ops/adam/test_cpu_adam.py
```diff
@@ -63,6 +63,9 @@ class TestCPUAdam(DistributedTest):
 
     @pytest.mark.skipif(not get_accelerator().is_available(), reason="only supported in CUDA environments.")
     def test_fused_adam_equal(self, dtype, model_size):
```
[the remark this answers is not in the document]
tjruwase: Yes, please skip the following for fp16 
image

With that change, I think this PR is should be ready for merge :)

csrc/adagrad/cpu_adagrad.cpp
tjruwase: @BacharL, thanks for sharing these results. Did you use the existing perf script? If not, do you mind sharing yours?

tests/unit/ops/adam/test_cpu_adam.py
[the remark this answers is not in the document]
BacharL: This time other configs failed, so I disabled all fp16 configs
https://github.com/microsoft/DeepSpeed/actions/runs/9061344866/job/24892917705

csrc/adagrad/cpu_adagrad.cpp
BacharL: Yes, I used the perf script above. just changed param dtype to the desired one.

tjruwase (approved):

PR #5412, opened 2024-04-15
[XPU] Use non_daemonic_proc by default on XPU device

ys950902:
Set non_daemonic_proc=True by default on XPU Device, using non_daemonic_proc for unit test.


loadams (approved):

tests/unit/common.py
```diff
@@ -248,6 +248,11 @@ def _launch_procs(self, num_procs):
                 f"Skipping test because not enough GPUs are available: {num_procs} required, {get_accelerator().device_count()} available"
             )
 
+        if get_accelerator().device_name() == 'xpu':
+            print("use non_daemonic_procs for xpu")
```
loadams, line 252: Hi @ys950902 - could you remove the print, since the code below is quite readable?  And I tested this and it works, but could you let us know why this needs to be set or is recommended to be set?

loadams:
Passing test here: https://github.com/microsoft/DeepSpeed/actions/runs/8693423957

tests/unit/common.py
ys950902: Hi @loadams, I have removed the print, yes, because some issues for ipex when using daemonic_proc, this can be a WA for XPU device to cover more unit test for multi-processing scenarios.

loadams (approved):

PR #5423, opened 2024-04-16
Adding capability API in accelerator

duli2012:
This PR is a prototype of adding API for capabilities in accelerators including:
1. define capabilities in abstract_accelerator
2. set capabilities in cuda_accelerator 

Welcome hardware vendors to define capabilities for their own hardware.


accelerator/abstract_accelerator.py
```diff
@@ -12,7 +12,12 @@ class DeepSpeedAccelerator(ABC):
     def __init__(self):
         self._name = None
         self._communication_backend_name = None
-
+        self._capabilities: dict[str, bool] = {
+            "zero1": False,
```
tjruwase, line 16: These hardcoded constants, such as "zero1", should be defined as symbolic constants in a global location.

delock:
Hi @duli2012 thanks for adding this interface.  I have always been worring accelerator interface# may grow too big when we propose more and more capabilities into it, this interface is a good way to put all capabilities into one interface.  My comments below:

1. For zero1/2/3, it sounds like accelerator agnostic features.  Basically it should be supported if accelerator runtime and communication collectives is well implemented.
2. For sparse_attn, it is in the category of whether certain OpBuilder is implemented.  Previous way of telling whether ops is implemented is through compatible ops as the following link.   I think its better to move this capability group into capability interface since the new interfacve look more intuitive to use.   Is it possible to make it auto reflect accelerator OpBuilder implementation so there will be less maintentance work?  A minor suggestion is name capability in this group with same prefix i.e. `op.sparse_attn`
https://github.com/microsoft/DeepSpeed/blob/b22706a7211366abf2df98a0d118ea1d3a837e21/tests/unit/inference/test_inference.py#L336
3. For 1-bit adam, I agree this is a case that could be covered by this interface.

What should we do with already existing accelerator interface that falls into category of capabilities?  Should they be added to the new inferface or just keep them that way?  I think that's an open to discuss.

duli2012:
> Hi @duli2012 thanks for adding this interface. I have always been worring accelerator interface# may grow too big when we propose more and more capabilities into it, this interface is a good way to put all capabilities into one interface. My comments below:
> 
> 1. For zero1/2/3, it sounds like accelerator agnostic features.  Basically it should be supported if accelerator runtime and communication collectives is well implemented.
> 2. For sparse_attn, it is in the category of whether certain OpBuilder is implemented.  Previous way of telling whether ops is implemented is through compatible ops as the following link.   I think its better to move this capability group into capability interface since the new interfacve look more intuitive to use.   Is it possible to make it auto reflect accelerator OpBuilder implementation so there will be less maintentance work?  A minor suggestion is name capability in this group with same prefix i.e. `op.sparse_attn`
>    https://github.com/microsoft/DeepSpeed/blob/b22706a7211366abf2df98a0d118ea1d3a837e21/tests/unit/inference/test_inference.py#L336
> 3. For 1-bit adam, I agree this is a case that could be covered by this interface.
> 
> What should we do with already existing accelerator interface that falls into category of capabilities? Should they be added to the new inferface or just keep them that way? I think that's an open to discuss.

Thanks @delock for your comments. I integrated most of them. please take a look.

accelerator/abstract_accelerator.py
duli2012: @tjruwase  As we discussed, I collected the names in *OpBuilder and put it in constants.py. Do we need add other features like optimizers as ds features too?

accelerator/constants.py
```diff
@@ -0,0 +1,31 @@
...
+OP_CPU_ADAGRAD = "cpu_adagrad"
+OP_CPU_ADAM = "cpu_adam"
+OP_CPU_LION = "cpu_lion"
+OP_EVOFORMER_ATTN = "evoformer_attn"
+OP_FP_QUANTIZER = "fp_quantizer"
+OP_FUSED_ADAM = "fused_adam"
+OP_FUSED_LAMB = "fused_lamb"
+OP_FUSED_LION = "fused_lion"
+OP_INFERENCE_CORE_OPS = "inference_core_ops"
+OP_CUTLASS_OPS = "cutlass_ops"
```
rogerxfeng8, line 19: A general name is needed to cover the non-cuda devices?

loadams: Hi @rogerxfeng8 - is there a specific one you're referencing?  I believe we call all devices (cuda and non-cuda) accelerators.

PR #5424, opened 2024-04-16
uniform deepspeed overflow check

GuanhuaWang:
**Before**: Overflow check is scattered and duplicated in all places.

**This PR:** 

- Single interface as CheckOverflow class, abstract and uniform overflow check among ZeRO, ZeRO-Offload, Pipeline Parallelism, BF16_optimizer. 
- Skip step() operation if detect gradients overflow in BF6_optimizer. (avoid pollute checkpoint, etc)

cc @tjruwase

### DIFF
--- a/deepspeed/runtime/bf16_optimizer.py
+++ b/deepspeed/runtime/bf16_optimizer.py
@@ -13,9 +13,10 @@ from deepspeed.runtime.base_optimizer import ZeROOptimizer
 from packaging import version as pkg_version
 from deepspeed.git_version_info import version
 from deepspeed.runtime.utils import (get_global_norm_of_tensors, clip_tensors_by_global_norm, DummyOptim,
-                                     align_dense_tensors, all_gather_dp_groups, is_model_parallel_parameter,
-                                     see_memory_usage, graph_process, get_norm_with_moe_layers)
-from deepspeed.utils import link_hp_params, lazy_init_hp_params_optimizer_state, fragment_address, groups
+                                     CheckOverflow, align_dense_tensors, all_gather_dp_groups,
+                                     is_model_parallel_parameter, see_memory_usage, graph_process,
+                                     get_norm_with_moe_layers)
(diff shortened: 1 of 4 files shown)


Anhelor:
Why not using tensor.isnan() and tensor.isinf()?

deepspeed/runtime/utils.py
```diff
@@ -181,12 +181,13 @@ def get_norm_with_moe_layers_fast(all_groups_norm, group):
 class CheckOverflow(object):
     '''Checks for overflow in gradient across parallel process'''
 
-    def __init__(self, param_groups=None, mpu=None, zero_reduce_scatter=False, deepspeed=None):
+    def __init__(self, param_groups=None, mpu=None, zero_reduce_scatter=False, deepspeed=None, partition_grads=False):
```
tjruwase, line 184: Passing deepseed engine into a submodule is not a good design and create all sorts of cyclic reference issues. It is better to pass the specific attributes that are needed, such as `enable_backward_allreduce`

deepspeed/runtime/engine.py
```diff
@@ -1473,6 +1473,7 @@ def _configure_bf16_optimizer(self, optimizer):
         timers = self.timers if self.wall_clock_breakdown() else NoopTimer()
         optimizer = BF16_Optimizer(optimizer,
                                    self.param_names,
+                                   deepspeed=self,
```
tjruwase, line 1476: Don't pass `self` into submodules.

deepspeed/runtime/bf16_optimizer.py
```diff
@@ -92,6 +94,10 @@ def __init__(self,
         if self.using_real_optimizer:
             self._setup_for_real_optimizer()
 
+        # Overflow check init
+        self.overflow = False
```
tjruwase, line 98: Should `self.overflow` be a class member since it seems to be only used once?

Issue #5175, opened 2024-02-23
[BUG]Training speed of deepspeed>=0.12.5 becomed slower than before!


janelu9:
about 5% slowdown when training with pipeline parallelism or megatron-deepspeed

loadams:
Hi @janelu9 - could you let us know what baseline DeepSpeed version you are comparing with?

janelu9:
> Hi @janelu9 - could you let us know what baseline DeepSpeed version you are comparing with?  And that this is the same "issue" as #4881 ?

this version is faster https://github.com/microsoft/DeepSpeed/files/13825076/deepspeed-0.12.4.zip

loadams:
Thanks, so something between 0.12.4 and 0.12.5, I'll git bisect between those and see if I can tell, can you share your torch version as well so I'm running the same tests?  As well as if you have a small repro case or one from DeepSpeed Examples?

janelu9:
> Thanks, so something between 0.12.4 and 0.12.5, I'll git bisect between those and see if I can tell, can you share your torch version as well so I'm running the same tests? As well as if you have a small repro case or one from DeepSpeed Examples?

My torch version==2.1.2+cu118, I don't have a repro published, I test them with Llama-7B of 8 stages in pipeline engine on one node of 8 A100-40G.

loadams:
@janelu9 - I'll have to test a repro with V100s probably and a smaller model, and it is possible that it won't repro there in the same way.

If it is easy for you to run your test, could you try with `pip install deepspeed git+https://github.com/microsoft/DeepSpeed.git@a7900bcc3d2f7789cc734aa28a11d2f3b3d8b04f` to help bisect the search space?

janelu9:
> @janelu9 - I'll have to test a repro with V100s probably and a smaller model, and it is possible that it won't repro there in the same way.
> 
> If it is easy for you to run your test, could you try with `pip install git+https://github.com/microsoft/DeepSpeed.git@a7900bcc3d2f7789cc734aa28a11d2f3b3d8b04f` to help bisect the search space?

Well, this version is the faster one which was 10.3 samples/sec while the latest one was 9.9 samples/sec. Otherwise, I quoted flash-attn in my model codes.

loadams:
The one installed from the git link was also faster?  If so, could you also try this one?

pip install git+https://github.com/microsoft/DeepSpeed.git@b83b1c2e1c4dc4c91c4ad78773dc2232ca9f7070

[...]

PR #5434, opened 2024-04-18
set the default to use set_to_none for clearing gradients in BF16 optimizer.

inkcherry:
as discussed in #5175,  set the default to use set_to_none for clearing gradients in BF16 optimizer.
Additionally, for the case of zero clearing, use foreach_zero.
Verified correctness with mega-ds llama 7B training.

FYI @loadams

### DIFF
--- a/deepspeed/runtime/bf16_optimizer.py
+++ b/deepspeed/runtime/bf16_optimizer.py
@@ -341,7 +341,7 @@ class BF16_Optimizer(ZeROOptimizer):
 
         # clear gradients
         if clear_lp_grads:
-            lp.grad._zero()
+            lp.grad.zero_()
 
     @torch.no_grad()
     def _update_hp_grads_func(self, clear_lp_grads=False):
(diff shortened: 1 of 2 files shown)


deepspeed/runtime/bf16_optimizer.py
```diff
@@ -441,11 +441,20 @@ def clear_hp_grads(self):
             self.fp32_groups_has_gradients[i] = [False] * len(group)
 
     def clear_lp_grads(self):
+
+        # using zero_() fixed memory address for graph replay
+        set_to_none = set_to_none = False if self.graph_harvesting else True
```
tjruwase, line 446:
```suggestion
        set_to_none = False if self.graph_harvesting else True
```

inkcherry: fixed and reverified

deepspeed/runtime/bf16_optimizer.py
```diff
@@ -341,7 +341,7 @@ def _update_hp_grad(self, lp, group_idx, param_idx, clear_lp_grads):
 
         # clear gradients
         if clear_lp_grads:
-            lp.grad._zero()
+            lp.grad.zero_()
```
tjruwase, line 344: @inkcherry, will this break on older pytorch versions?

inkcherry: It shouldn't, referring to the definition of ```zero_grad``` in version 1.10 at https://pytorch.org/docs/1.10/_modules/torch/optim/optimizer.html#Optimizer

tjruwase (approved):

PR #5436, opened 2024-04-19
enable phi2 autotp

Yejing-Lai:
This PR aims to enable phi2 model autotp.


deepspeed/module_inject/auto_tp.py
```diff
@@ -306,6 +306,8 @@ def tp_parser(model):
                 # Mixtral-7x8b used w2*act(w1*w3) linear. need to replace w2 to linearallreduce.
                 elif 'w2' in layer and 'Mixtral' in str(type(module)):
                     gem_list = gem_list + [layer]
+                elif "self_attn.dense" in layer:
```
mrwyattii, line 309: Should this also have a check for `and "Phi" in str(type(module))`?

Yejing-Lai: Added it~

PR #5461, opened 2024-04-24
Revert "stage3: efficient compute of scaled_global_grad_norm (#5256)"

lekurile:
This reverts commit 54c06872647ca60699f752e60ac1643bd05aa63c due to #5256 causing bugs when the ZeRO3 + ZeRO Offload features are enabled.

This bug was discovered due to failures in the DS Chat CI workflow. Failing tests across CI failures:
| Failing Test Name |
| --- |
| test_ds_chat[zero3--offload-] |
| test_ds_chat[zero3--offload-lora] |
| test_ds_chat[zero3-he-offload-] |
| test_ds_chat[zero3-he-offload-lora] |

Error message:
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:3 and cpu!
```

It seems that `torch.stack()` or `torch.norm()` is having issues when the offload feature is enabled and tensors are split between CPU/GPU, however this is just an initial guess and would require more investigation.

@nelyahu Since you are the original author of the PR, if you have some bandwidth, any help here is greatly appreciated!

After reverting this commit, all tests pass in the DS Chat CI workflow:
https://github.com/microsoft/DeepSpeed/actions/runs/8824064414/job/24225802763

@tjruwase for context.

PR #5462, opened 2024-04-25
Update ds-chat CI workflow paths to include zero stage 1-3 files

lekurile:
This PR updates the ds-chat CI workflow to run when ZeRO stage 1-3 files are updated.

### DIFF
--- a/.github/workflows/nv-ds-chat.yml
+++ b/.github/workflows/nv-ds-chat.yml
@@ -10,6 +10,10 @@ on:
         required: false
         default: 'master'
         type: string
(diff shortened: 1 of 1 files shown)


loadams (approved):
Should we wait until we have the workflow passing again to merge this?

lekurile:
> Should we wait until we have the workflow passing again to merge this?

Yes, I think we should wait until the revert PR goes in. I've done a manual test of that PR, so it will work one that goes in. Could you please review https://github.com/microsoft/DeepSpeed/pull/5461?

PR #5465, opened 2024-04-25
fix: swapping order of parameters in create_dir_symlink method.

alvieirajr:
Order of parameters in create_dir_symlink method looks wrong. Because this we get the error "PermissionError: [WinError 5] Denied access: '.\\deepspeed\\ops\\csrc'" when install deepspeed >= 0.4.0 on Windows enviroment.

Please check this out @eltonzheng and @jeffra.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -208,7 +208,7 @@ else:
     git_branch = "unknown"
 
 
-def create_dir_symlink(src, dest):
+def create_dir_symlink(dest,src):
     if not os.path.islink(dest):
         if os.path.exists(dest):
             os.remove(dest)
--- a/setup.py
+++ b/setup.py
@@ -208,7 +208,7 @@ else:
     git_branch = "unknown"
 
 
-def create_dir_symlink(dest,src):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -208,7 +208,7 @@ def op_enabled(op_name):
     git_branch = "unknown"
 
 
-def create_dir_symlink(src, dest):
+def create_dir_symlink(dest,src):
```
loadams, line 211: Perhaps keep these parameters in this order since that matches the os.symlink(src, dest) convention?  And switch the paths below where this is used?

alvieirajr: I did your sugestion here (commit 1f5384f6f3bf2ed40add713d073c28d6ad018ee7) @loadams.

loadams:
@alvieirajr - were you able to validate that swapping these resolved your issues?

alvieirajr:
> @alvieirajr - were you able to validate that swapping these resolved your issues?

@loadams  - yes, it worked too.

loadams (approved):

PR #5473, opened 2024-04-28
Add Compressedbackend for Onebit optimizers

Liangliang-Ma:
In the process of adding onebit optimizers support for XPU devices, we have noticed that for different accelerator, the main difference of implementation of `compressed_allreduce` lies on `packbits` and `unpackbits`. CUDA uses cupy and NPU uses torch_npu. Instead of replace these to xpu only functions, we provided a CompressedBackend to do the `compressed_allreduce` work where users can add their own packbits/unpackbits kernels, which is a general path for all kinds of accelerators.

In this PR, we:
1. Add CompressedBackend for onebitAdam, onebitLamb and zerooneAdam
2. Add XPU implement of packbits/unpackbits with SYCL, built in PackbitsBuilder
3. Add tests for onebit with CompressedBackend


csrc/xpu/packbits/packing.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    c10::impl::VirtualGuardImpl impl(device.type());
+    c10::Stream _stream = impl.getStreamFromGlobalPool(device, /*isHighPriority=*/false);
+    sycl::queue queue = xpu::get_queue_from_stream(_stream);
+    return queue;
+}
+
+at::Tensor packbits(at::Tensor tensor, int input_size, int rank)
+{
+    /*
```
delock, line 42: @Liangliang-Ma the function documentation needs to be moved to line 39 right before the function def line.

Liangliang-Ma: Updated.

delock:
@tjruwase this PR is an approach to abstract the generic part of 1bit-adam and implment accelerator dependent part with DeepSpeed custom op builder.  So 1bit-adam does not need to depend on accelerator specific libraries.

@inkcherry I remember you investigated in 1bit adam portability before, FYI this PR implement a portable version of 1bit adam support.

Liangliang-Ma:
Hi @tjruwase , could you please help to review this PR? Thanks!

tests/onebit/test_compressed_backend.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
Liangliang-Ma: @tjruwase Thanks for review. README added.

tjruwase (approved):

Liangliang-Ma:
@tjruwase I have noticed that in [onebit unit test](https://github.com/microsoft/DeepSpeed/blob/master/tests/unit/runtime/half_precision/onebit/test_onebit.py), the onebit comm backend is assigned like this `"comm_backend_name": get_accelerator().communication_backend_name()`. But in fact, it's not same as the common communication backend, which is under `deepspeed/comm/`, doing allreduce, broadcast, etc. Currently deepspeed supported NCCL, MPI, HCCL, and this PR added Compressed backend, are not corresponding to accelerator's comm backend(Most of them are actually using torchbackend now). How about we add a interface under accelerator like `get_accelerator().onebit_backend()` or change unit test to a more general way? Thanks.

Liangliang-Ma:
@tjruwase Hi, May I ask if you could help to review my last comment or merge this one first? Thanks

tjruwase:
@Liangliang-Ma, apologies for delay. I am still thinking about your last comment, but will not delay this PR.

PR #5481, opened 2024-04-30
Enable dynamic shapes for pipeline parallel engine inputs

tohtana:
This PR enables dynamic shapes for inputs to pipeline parallel (PP) engine.

Currently PP engine checks tensor shapes and allocates communication buffer at the first forward/backward passes. This causes a tensor shape mismatch error when input tensor shapes changed. 

This PR adds an option to check tensor shapes at every iteration and allocate buffer based on the shapes. As shown below, you can enable this feature by passing `dynamic_shape=True` to `PipelineModule`. 
Note that this might have a performance impact and the option is set to False as default.

```python
model = PipelineModule(
...
   dynamic_shape=True
)
```

This will increase the overhead of buffer allocation and communication for tensor metadata. To mitigate the overhead, this PR also includes these improvements:
- Consolidate multiple communication calls to send/recv tensor shapes 9f96ad4049b1fb63d38a1a090480dbef61dc0490 
- Reuse (extend) communication buffer instead of creating a new one b3c07504be05b08772f6297a864ec6c27b5eeca3


deepspeed/runtime/pipe/engine.py
```diff
@@ -926,15 +934,17 @@ def _send_tensor_meta(self, buffer, recv_stage):
...
+            meta_buf_list = [
+                0,  # type of data (0: tensor, 1: list, 2: tuple)
+                self.DTYPE_TO_ID[buffer.dtype],  # dtype
+                len(buffer.size())  # ndims
+            ]
+            meta_buf_list.extend(buffer.size())
+            meta_buffer[:len(meta_buf_list)].copy_(torch.tensor(meta_buf_list, dtype=torch.int32))
+            p2p.send(meta_buffer, recv_stage)
+
         elif isinstance(buffer, list):
```
tjruwase, line 948: Why is meta buffer not applied to `list` option?

deepspeed/runtime/pipe/engine.py
```diff
@@ -926,15 +934,17 @@ def _send_tensor_meta(self, buffer, recv_stage):
                 * ndims
                 * shape
         """
-        send_bytes = 0
+        meta_buffer = torch.empty(TENSOR_META_SIZE, dtype=torch.int32, device=self.device)
```
tjruwase, line 937: How can we guard against overflow if entries into `meta_buffer` is more than `TENSOR_META_SIZE`

deepspeed/runtime/pipe/engine.py
```diff
@@ -40,6 +42,8 @@
 PIPE_RECV_INPUT_TIMER = 'pipe_recv_input'
 PIPE_RECV_GRAD_TIMER = 'pipe_recv_grad'
 
+TENSOR_META_SIZE = 256
+
```
duli2012, line 46: I believe 256 is an estimate of max meta_buffer size. However, we need a fallback path if the actual meta_buffer size exceeds the 256, which should be very rare though. 
An alternative is to send the actual meta buffer size before send the meta buffer. This way, it adds one more communication for just one integer, not sure how much overhead it introduces, but it's a safer approach.

deepspeed/runtime/pipe/engine.py
```diff
@@ -1305,6 +1295,18 @@ def _allocate_buffers(self, shapes_and_dtypes, requires_grad=False, num_buffers=
             buffers.append(buffer)
         return buffers
 
+    def _allocate_or_extend_buffers(self, idx, shape, dtype):
+        numel = reduce(mul, shape) if len(shape) > 0 else 1
+        if len(self._grad_layer_buf) <= idx or self._grad_layer_buf[idx].numel() < numel:
+            new_buf = self._allocate_buffer(shape, dtype=dtype, num_buffers=1)[0]
+            if len(self._grad_layer_buf) <= idx:
+                self._grad_layer_buf.append(new_buf)
```
duli2012, line 1303: why could len(self._grad_layer_buf) be smaller then idx?  Dynamic shape inputs shouldn't change the number of gradients, right?

deepspeed/runtime/pipe/engine.py
tohtana: We have `assert (False)` in the branch for `list`. I think the author left it for debugging or development purpose.
As it is four years old now, I will delete it.

deepspeed/runtime/pipe/engine.py
tohtana: I want to avoid adding another communication because it can be the overhead. I added the assertions to detect the overflow. Given the current buffer size, we can send metadata for more than 80 dimensions. I don't think this can happen as long as other parts are properly working.

deepspeed/runtime/pipe/engine.py
tohtana: I added assertions to check the buffer size

deepspeed/runtime/pipe/engine.py
tohtana: The default number `256` is far more than the necessary size in most cases. Typically, it should be less than 10. I want to keep this a constant to avoid the overhead by the additional communication call. (Assuming the size of the communication doesn't matter but adding another call can be the overhead)
Just to make it easier to identify the error caused by the buffer overflow, I added assertion and the detailed error message.

deepspeed/runtime/pipe/engine.py
tohtana: The original code also extended the buffer for gradients (`_allocate_buffers`). I added `_allocate_or_extend_buffers` to pass shape and dtype one by one, instead of having a loop inside.
I found that `_allocate_buffers` is now unused. Let me just delete it.

tohtana:
I addressed comments from @tjruwase and @duli2012 and added a unit test.

tjruwase (approved):

Issue #5422, opened 2024-04-16
[BUG] Tensors are on different devices when model.step()


Heathcliff-Zhao:
**Describe the bug**
The behavior is same to what is reported in #4565 . When model.step() with zero3, Tensors are on different devices. I modified [stage3.py#L2117](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/zero/stage3.py#L2117) to self.fp32_partitioned_groups_flat[sub_group_id].grad.mul_(1. / combined_scale.item()), and it seems to be right. I am not sure whether this is a bug or I made a mistake in my training script.

**My code**
My training code is taken from here: https://github.com/liucongg/ChatGLM-Finetuning/blob/master/train.py, and I run this with `CUDA_VISIBLE_DEVICES=0 deepspeed train.py --train_path data/spo_0.json --model_name_or_path ./opensource --per_device_train_batch_size 8 --max_len 4096 --max_src_len 2048 --learning_rate 1e-4 --weight_decay 0.1 --num_train_epochs 2 --gradient_accumulation_step 4 --warmup_ratio 0.1 --mode glm3 --train_type all --seed 1234 --ds_file ds_zero3_offload.json --gradient_checkpointing --show_loss_step 10 --output_dir ./output-glm3`

Thanks very much for your precious time!

cloudwaysX:
I have the same issue here when running the following script. https://github.com/allenai/open-instruct/blob/main/scripts/dpo_train_with_accelerate.sh. I notice that version 0.14.0 does not have this issue.

wuxb45:
I have the same issue with 0.14.1 when running a similar training script. Version 0.14.0 works.
I cross-checked with torch 2.2.1, 2.2.2 and transformers 4.39.0, 4.39.3. The issue is with 0.14.1 across all combinations.

Here is a snippet of the backtrace (but I'm sorry that I cannot provide the python code. Hope it helps):
```
  File "/home/x/work/.venv/lib/python3.11/site-packages/deepspeed/runtime/engine.py", line 2169, in step
    self._take_model_step(lr_kwargs)
  File "/home/x/work/.venv/lib/python3.11/site-packages/deepspeed/runtime/engine.py", line 2075, in _take_model_step
    self.optimizer.step()
  File "/home/x/work/.venv/lib/python3.11/site-packages/deepspeed/utils/nvtx.py", line 15, in wrapped_fn
    ret_val = func(*args, **kwargs)
[...]

PR #5504, opened 2024-05-07
Z3: optimizations for grad norm calculation and gradient clipping

nelyahu:
This PR add the below functionality:
1. complete_grad_norm_calculation_for_cpu_offload: move total_norm to CPU, as expected device in such case is CPU..
2. repalce get_global_norm() with torch.linalg.norm for better performance.
3. unscale_and_clip_grads: replace clipping based on if statement to use torch.clamp for better performance.

change (3) is taken from https://github.com/microsoft/DeepSpeed/pull/5547 (which was closed)


deepspeed/runtime/zero/stage3.py
```diff
@@ -2027,7 +2027,7 @@ def step(self, closure=None):
             return
 
         norm_groups = self._get_norm_groups()
-        scaled_global_grad_norm = get_global_norm(norm_list=norm_groups)
+        scaled_global_grad_norm = torch.norm(torch.stack(norm_groups))
```
tjruwase, line 2030: I just noticed that `torch.norm` is deprecated. Is there any reason not to use the recommended replacements? 
image

Anhelor: torch.linalg.norm is fine.

nelyahu: @Anhelor @tjruwase Done

tjruwase (approved):

jomayeri:
Changing this line has been associated with several bugs https://github.com/microsoft/DeepSpeed/issues/5422, https://github.com/microsoft/DeepSpeed/issues/5538

loadams:
> Changing this line has been associated with several bugs #5422, #5538

@nelyahu - thoughts on this comment, seems last time this line was modified users ran into issues?

nelyahu:
> > Changing this line has been associated with several bugs #5422, #5538
> 
> @nelyahu - thoughts on this comment, seems last time this line was modified users ran into issues?

@loadams, Yes - this optimization was already pushed and reverted due to ds-chat (failures in cpu-offload configurations).
i did offline debugging of those failure and improved the code change so it will pass. Since then ds-chat tests where added to DeepSpeed repo CI and it is now passing. 
Are there any other tests (full model training for example), that does not exists in the CI, which can be manually ran?

tjruwase:
> i did offline debugging of those failure and improved the code change so it will pass

@nelyahu, it great that you narrowed this down. Do you think a unit test can be added for this case?

loadams:
> > i did offline debugging of those failure and improved the code change so it will pass
> 
> @nelyahu, it great that you narrowed this down. Do you think a unit test can be added for this case?

@nelyahu - we've stabilized the CI, thoughts on adding this test?

nelyahu:
@loadams Oh, Sorry- I missed the last comment. 
Sure, yes we can add such UT that will cover it.
But i cannot address it immediately. I will update this PR once we have a unit test.

nelyahu:
@loadams / @tjruwase  as request i make sure the regression was discussed here will be covered by a unit tests. I used TestZeroPartialOffloadConfigSweep, and added it gradient_clipping so it will go though the problematic flow.
i reproduced the issue using the UT, and made sure it is fixed.

nelyahu:
@loadams can you re-run workflows? i suspect that the failure are not related to this PR

Issue #5430, opened 2024-04-17
[BUG] No `universal_checkpoint_info` in the Accelerate+Deepspeed Checkpoint


Orion-Zheng:
I trained model using Accelerate+Deepspeed ZeRO-2 and got a ZeRO-2 checkpoint. The checkpoint structure is listed below. And this is the Google Drive [link](https://drive.google.com/drive/folders/1ey6zLPQh0IsX66FQby6qYBnno1un6jol) to my checkpoint. 
```
checkpoint-3/
├── config.json
├── generation_config.json
├── global_step3
│   ├── bf16_zero_pp_rank_0_mp_rank_00_optim_states.pt
│   ├── bf16_zero_pp_rank_1_mp_rank_00_optim_states.pt
│   ├── bf16_zero_pp_rank_2_mp_rank_00_optim_states.pt
│   ├── bf16_zero_pp_rank_3_mp_rank_00_optim_states.pt
│   └── mp_rank_00_model_states.pt
├── latest
├── model.safetensors
├── rng_state_0.pth
├── rng_state_1.pth
├── rng_state_2.pth
├── rng_state_3.pth
├── scheduler.pt
├── trainer_state.json
├── training_args.bin
└── zero_to_fp32.py
```
I tried to convert this ZeRO-2 checkpoint to the universal format using `ds_to_universal.py` but encountered errors:
```
args = Namespace(input_folder='experiment_ckpts/tinyllama_expanded_frez_embed-2024-04-16-010251/checkpoint-3', output_folder='experiment_ckpts/tinyllama_expanded_frez_embed-2024-04-16-010251/checkpoint-3_universal', num_extract_workers=10, num_merge_workers=10, keep_temp_folder=False, strict=True)
Convert DeepSpeed Checkpoint to Universal Checkpoint
Converting DeepSpeed checkpoint in experiment_ckpts/tinyllama_expanded_frez_embed-2024-04-16-010251/checkpoint-3 to Universal checkpoint in experiment_ckpts/tinyllama_expanded_frez_embed-2024-04-16-010251/checkpoint-3_universal
Traceback (most recent call last):
  File "dist_env_tools/ds_to_universal.py", line 363, in <module>
    main(args)
  File "dist_env_tools/ds_to_universal.py", line 320, in main
    _check_for_required_state(ds_checkpoint)
  File "dist_env_tools/ds_to_universal.py", line 311, in _check_for_required_state
[...]

PR #5526, opened 2024-05-13
create mininal universal checkpoint info for client state

xylian86:
This PR solves the [Issue-5430](https://github.com/microsoft/DeepSpeed/issues/5430).

The PR enables the universal checkpoint feature for other platforms like HuggingFace Trainer without requiring changes to the HuggingFace code. It does this by creating a minimal universal checkpoint info, specifically the version, as a default action for the client state.


tjruwase:
@xylian86, thanks for this great work. Can you please add convergence curves of an HF model as demo?

deepspeed/runtime/engine.py
```diff
@@ -3319,6 +3320,7 @@ def _save_checkpoint(self, save_dir, tag, client_state={}, exclude_frozen_parame
                      ds_config=self.config,
                      ds_version=version)
         state.update(client_state)
+        inject_universal_info(state)
```
tohtana, line 3323: Shoudn't we show a warning when we don't have the necessary info?
It will silently produce an incorrect checkpoint if the checkpoint is loaded for TP or PP.
We can say that the converted checkpoint is only for pure DP scaling.

tohtana: After I discussed with Tunji, we are considering another approach for this. He will share a new approach. I keep this comment but just disregard it.

tjruwase: @xylian86, the new approach is to do the injection in the conversion script rather than during saving. The injection should be done into `ds_checkpoint` before this [assertion](https://github.com/microsoft/DeepSpeed/blob/ebf82e8f3ad6d51d49d115e54a11ae4597ff36fb/deepspeed/checkpoint/ds_to_universal.py#L358).  Furthermore, the injection should be enabled by command-line argument (disabled by default) so that users are fully aware of what is going on. The command-line arg could be called `--inject-missing-state`.

deepspeed/checkpoint/utils.py
```diff
@@ -5,7 +5,7 @@
 
 import os
 import torch
-from .constants import (MODEL_FILE_PREFIX, MODEL_FILE_SUFFIX, OPTIM_FILE_SUFFIX, ZERO_FILE_PREFIX)
+from .constants import (MODEL_FILE_PREFIX, MODEL_FILE_SUFFIX, OPTIM_FILE_SUFFIX, ZERO_FILE_PREFIX, UNIVERSAL_CHECKPOINT_INFO, UNIVERSAL_CHECKPOINT_VERSION_KEY, UNIVERSAL_CHECKPOINT_VERSION_VALUE)
```
loadams, line 8: FYI @xylian86 - can you run the precommit formatter on this branch so it will pass our Formatting check?

```
pre-commit run --all-files
```

PR #5532, opened 2024-05-14
[manifest] update mainfest to add hpp file in deepspeed

ys950902:
Update the mainfest to cover hpp file in deepspeed.


ys950902:
Hi @loadams, could you please help to review the changes?
The op csrc is in deepspeed, so should also add the hpp file

PR #5571, opened 2024-05-27
[CPU] SHM based allreduce improvement for small message size

delock:
On CPU server, when running SHM based allreduce for small messages, the performance is pretty much dominated by synchronization latency.  These latency includes the following two situations:
1. Wait for status change from other ranks.
2. Use `#pragma omp parallel for` to accelerator memory bandwidth bound operations such as `parallel_memcpy` or `reduce`.

Each synchronization add a little time to allreduce latency.  In current implementation, for small messages, 5 syncs on rank 0 are needed.  This includes: 1) copy-in; 2) wait for other ranks done copy; 3) reduce; 4) copy-out; 5) wait for other ranks finish copy-out

We redesign the algorithm for small message allreduce (called `symmetric_naive_allreduce`) to have only three syncs, each rank do exactly the same steps: 1) copy-in; 2) wait for other ranks done copy; 3) reduce to output buffer directly.   We use double buffer so we can skip the last wait and go directly to next call using another buffer.  We have a carefully designed state check to avoid using global barrier among ranks.

Test shows for message size < 1MB, allreduce latency will reduce 30% to 50%.  This is especially helpful for tensor parallel decoding with small batch size, where the tensor size is usually a few 10s of KBytes.

|message size(bytes)|new method latency(us)|old method latency(us)|
|---|---|---| 
|             2      |    13.34|20.39
|             4      |    13.44|19.57
|             8      |    13.70|19.76
|            16     |     13.27|20.43
|            32     |     13.42|19.75
|            64     |     13.38|19.80
|           128    |      13.70|19.44
|           256    |      13.99|20.33
|           512    |      13.91|20.28
|          1024   |       15.00|22.86
|          2048   |       15.82|20.93
|          4096   |       16.00|21.08
|          8192   |       16.31|21.50
|         16384  |        16.27|22.95
|         32768  |        16.13|25.17
|         65536  |        18.92|25.90
|        131072 |         21.12|27.42
|        262144 |         23.09|32.36
|        524288 |         32.78|42.80

Because the new method would compute same reduce value on all ranks.  Caution needs to be taken to ensure the result is identical on all ranks.   We use the test in the link https://github.com/delock/ds_allreduce_bench/blob/main/ds_comm_bench.py#L70 to ensure the implementation is correct.   https://github.com/delock/ds_allreduce_bench/blob/main/validate.sh is a test script for better coverage.


tjruwase:
> Hi @awan-10 , this PR is ready for review, can this PR be reviewed? Thanks!

@delock, we are reviewing now. Thanks for the PR!

csrc/cpu/comm/shm.cpp
```diff
@@ -597,59 +593,78 @@ void distributed_naive_reduce(char* data_ptr,
...
+    workspace[world_rank]->states[state_group] = copy_current;
 
 #ifdef DO_PROFILE
     auto t1 = std::chrono::system_clock::now();
 #endif
 
     for (int i = 0; i < world_size; i++) {
         // wait until all the other ranks copy the buffer
-        wait_buffer_state_until_range(i, coll_allreduce_naive__copy_in_done, 2);
+        wait_buffer_state_until_2(i, copy_current, reduce_current, state_group);
```
adk9, line 630:
```suggestion
        if (i != world_rank) { wait_buffer_state_until_2(i, copy_current, reduce_current, state_group); }
```

csrc/cpu/comm/shm.cpp
```diff
@@ -597,59 +593,78 @@ void distributed_naive_reduce(char* data_ptr,
...
 #endif
 
     for (int i = 0; i < world_size; i++) {
-        int rank = (i + world_rank) % world_size;
-        // wait until the other rank reduce the buffer
-        wait_buffer_state_until_range(rank, coll_allreduce_naive__reduce_done, 2);
-        parallel_memcpy(slice_data(data_ptr, chunk_el, data_size, rank),
-                        slice_data(workspace[rank]->buffer, chunk_el, chunk_size / chunk_el, rank),
-                        slice_size(chunk_el, rank) * data_size);
+        wait_buffer_state_until_2(i, reduce_current, copy_next, state_group);
```
adk9, line 652:
```suggestion
        if (i != world_rank) { wait_buffer_state_until_2(i, reduce_current, copy_next, state_group); }
```

csrc/cpu/comm/shm.cpp
```diff
@@ -539,46 +457,124 @@ size_t slice_el_start(size_t chunk_el, int slice_idx)
...
+                                size_t chunk_el)
 {
-    parallel_memcpy(workspace[world_rank]->buffer, data_ptr, chunk_size);
-    std::atomic_thread_fence(std::memory_order_release);
-    workspace[world_rank]->state = coll_allreduce_naive__copy_in_done;
+#ifdef DO_PROFILE
+    static double total_t1_t0 = 0.0;
+    static double total_t2_t1 = 0.0;
+    static double total_t3_t2 = 0.0;
+    static double total_t4_t3 = 0.0;
```
adk9, line 477: Not needed.

csrc/cpu/comm/shm.cpp
```diff
@@ -539,46 +457,124 @@ size_t slice_el_start(size_t chunk_el, int slice_idx)
...
+    static double total_t5_t4 = 0.0;
```
adk9, line 478: This is also not needed. It's unused.

adk9 (changes requested):

csrc/cpu/comm/shm.cpp
delock: Will remove, thanks for the catch!

csrc/cpu/comm/shm.cpp
delock: Will change according to suggestion.

delock:
Hi @adk9 , code updated according to comments, thanks!

adk9 (approved):

adk9:
> Hi @adk9 , code updated according to comments, thanks!

Hi @delock, thanks for your changes! The formatting check seems to fail for this PR. Could you run `pre-commit` or `clang-format` on your changes?

delock:
> > Hi @adk9 , code updated according to comments, thanks!
> 
> Hi @delock, thanks for your changes! The formatting check seems to fail for this PR. Could you run `pre-commit` or `clang-format` on your changes?

Hi @adk9 , formatting had been fixed in latest CI.  Thanks!

PR #5572, opened 2024-05-27
Reuse KV cache of prefixes

tohtana:
This PR implements reusing KV cache prefixes.
See https://github.com/microsoft/DeepSpeed-MII/pull/484 for details.


deepspeed/inference/v2/ragged/prefix_block_map.py
```diff
@@ -0,0 +1,56 @@
...
+            else:
+                break
+        return cached_blocks
+
+    def extend(self, tokens: torch.Tensor, new_block_ids: List[int], num_already_cached_blocks: int) -> None:
+        n_blocks = len(tokens) // self.block_size
+        for i in range(num_already_cached_blocks, n_blocks):
+            chunk = tokens[:(i + 1) * self.block_size]
+            hash = token_ids_to_hash(chunk)
+            if hash not in self.tokens_to_blocks:
```
umchand, line 46: Do we still need this if statement if in the for loop you are already starting from num_already_cached_blocks.

tohtana: I'm not sure if it is necessary but there might be a complicated case. Assume we are running two requests.
Request 1 is using cached blocks 0, 1, 2, and just generated the last token of the current block. Then it saves the generated sequence and a hash.
Request 2 is also using the cached block 0, 1, 2, but the generation is a few steps later than Request 1. They are not sharing the last block. But the request may generate exact same tokens for the last block. So, it will try to update the cache with the same hash.
In this case, I didn't want to overwrite the cache.

umchand: sounds good, so this is to ensure we are sharing the recently generated blocks.

PR #5578, opened 2024-05-29
reduce cpu host overhead when using moe

ranzhejiang:
The operation `.to('cpu') `is not necessary for exp_counts, and it will cause device to host synchronization which damage  performance.


deepspeed/moe/sharded_moe.py
```diff
@@ -366,7 +366,7 @@ def top2gating(logits: Tensor,
     combine_weights = combine1_sec + combine2_sec
     dispatch_mask = combine_weights.bool()
 
-    return l_aux, combine_weights, dispatch_mask, exp_counts.detach().to('cpu')
+    return l_aux, combine_weights, dispatch_mask, exp_counts
```
tohtana, line 369: Currently `exp_counts` is unused at `forward()` of any of MoE classes, right?

deepspeed/moe/sharded_moe.py
```diff
@@ -322,7 +322,7 @@ def top2gating(logits: Tensor,
     l_aux = torch.mean(me * ce) * num_experts * num_experts
 
     # gating decisions
-    exp_counts = torch.sum(mask1 + mask2, dim=0)
+    exp_counts = torch.sum(mask1 + mask2, dim=0).detach().to(logits.device)
```
tohtana, line 325: Can the device of `mask1` and `mask1` be different from `logits`?

tohtana (commented):
@ranzhejiang Thank you for your contribution! I have a few questions about your changes. Can you clarify them?

deepspeed/moe/sharded_moe.py
ranzhejiang: > Can the device of `mask1` and `mask1` be different from `logits`?

From[ line 296 to 301](https://github.com/microsoft/DeepSpeed/pull/5578/files#diff-7d28ff32544f36339d99fdf142c2ab63d7332305758be0ff7ccc94e6350e1476R296-R301), we can find that , the calculation of `mask1` depends on `logits`, and all torch operation will keep the original device,  so the device of `mask1` and `logits` be the same one. The same to `mask1` and `mask2` [line 309 to 311](https://github.com/microsoft/DeepSpeed/pull/5578/files#diff-7d28ff32544f36339d99fdf142c2ab63d7332305758be0ff7ccc94e6350e1476R309-R311)

deepspeed/moe/sharded_moe.py
ranzhejiang: > Currently `exp_counts` is unused at `forward()` of any of MoE classes, right?

Yes, I have test it in Megatron-deepspeed and find that `exp_counts` is unused at `forward()` of any of MoE classes.

ranzhejiang: I have test these changes on my testing GPU platform, and work fine, no error and the loss keeps same to the original way.

ranzhejiang:
Hi, @tohtana  I have clarified the modifications you mentioned and retest this PR with Megatron-Deepspeed on GPU platform(8xA800). It runs well and loss remains consistent with the original method, Could you please help review it again? Thanks!

ranzhejiang:
https://github.com/microsoft/DeepSpeed/pull/5881 also adopts this plan to reduce cpu time

deepspeed/moe/sharded_moe.py
inkcherry: This causes no harm, and the loss remains consistent. By removing this CPU sync, it will improve performance by 1-2% on the GPU and 5%+ on the HPU. Could you please take a look again ?  @tohtana Thanks

tohtana: Thank you for the clarification!

tohtana (approved):

PR #5581, opened 2024-05-29
Remove compile wrapper to simplify access to model attributes

tohtana:
Having a wrapper of a compiled module brings various restrictions about accessing attributes of the compiled model.
This PR removes the wrapper of compiled module to simplify the access to the compiled model.


deepspeed/runtime/engine.py
```diff
@@ -3600,3 +3616,64 @@ def empty_partition_cache(self):
...
+            raise ValueError("Cannot change compiler_fn after compiling the module.")
+        self._compiler_fn = compiler_fn
+
+    @property
+    def is_compiled(self) -> bool:
+        return self._is_compiled
+
+    @property
+    def backend(self) -> Union[str, Callable]:
+        return self._backend
```
BacharL, line 3671:
```suggestion
        return self._compiler_backend
```

deepspeed/runtime/engine.py
```diff
@@ -361,8 +361,10 @@ def __init__(self,
         self.flatten = _flatten_dense_tensors
         self.unflatten = _unflatten_dense_tensors
 
-        if self._config.compile_config.enabled:
-            self._set_client_model(CompiledModuleWrapper(self.module, self._config.compile_config))
+        self._is_compiled = False
+        self._compiler_backend = None
```
BacharL, line 365:
```suggestion
        self._compiler_backend = get_backend_fn(self._config.compile_config.backend)
```

deepspeed/runtime/engine.py
```diff
@@ -90,7 +90,7 @@
 
 from .pipe.module import PipelineModule
 from .utils import get_ma_status
-from .compiler import CompiledModuleWrapper
+from .compiler import get_backend_fn
```
BacharL, line 93:
```suggestion
from .compiler import get_backend_fn, is_compile_supported
```

deepspeed/runtime/engine.py
```diff
@@ -1790,6 +1792,20 @@ def forward(self, *inputs, **kwargs):
...
+            if self._compiler_fn is None:
+                compiled_model = torch.compile(self.module,
+                                               backend=self._compiler_backend,
+                                               **self._config.compile_config.kwargs)
+            else:
+                compiled_model = self._compiler_fn(self.module)
+
+            self._set_client_model(compiled_model)
+            self._is_compiled = True
+
```
BacharL, line 1808:
```suggestion
        if self._config.compile_config.enabled not self._is_compiled and is_compile_supported():
            backend = self._compiler_fn if self._compiler_fn is not None else self._compiler_backend
	    self.module.compile(backend = backend, **self.torch_compile_kwargs)
	    self._is_compiled = True
```
Reasons for the above:

1. Allows passing user defined kwargs even when the user provides custom compiler_fn
2. Type of self.module does not changes since torch.nn.Module.compile is called instead of torch.compile. The module is compiled in-place.
I used to fail on [this line](https://github.com/microsoft/DeepSpeed/blob/2fc702ed9f84f3c88179cb713bb9f4effbbdba33/deepspeed/runtime/pipe/engine.py#L70). It should pass now as we only compile on forward and not in __init__, nevertheless we should consider keeping the type of self.module.

tohtana: Using `torch.nn.Module.compile` is a great idea. Thank you for this suggestion.

`compiler_fn` is not a backend. The intention of `compiler_fn` is to enable something that can't be done just by setting backend and kwargs for `torch.compile` (e.g. compiling part of the model). We need to fix that part carefully to make it consistent with in-place compile.

deepspeed/runtime/engine.py
tohtana: Fixed, thank you!

deepspeed/runtime/engine.py
tohtana: Thank you, I also simplified setting _compiler_backend in forward

deepspeed/runtime/engine.py
BacharL: I think it is possible to achieve this with `compiler_fn`. `torch.nn.Module.compile` just calls this function for compilation.
example in [this custom function](https://github.com/microsoft/DeepSpeed/blob/2fc702ed9f84f3c88179cb713bb9f4effbbdba33/tests/unit/runtime/compile/test_load_config.py#L29) one can iterate over module and call `compile()` inplace for every part to compile

```
    def custom_compiler_fn(module: torch.nn.Module, example_inputs):
        global custom_compler_fn_called
        custom_compler_fn_called = True
        module.l1.compile(backend=get_accelerator().get_compile_backend())
```

deepcharm: > Reasons for the above:
> 
> 1. Allows passing user defined kwargs even when the user provides custom compiler_fn
> 2. Type of self.module does not changes since torch.nn.Module.compile is called instead of torch.compile. The module is compiled in-place.
>    I used to fail on [this line](https://github.com/microsoft/DeepSpeed/blob/2fc702ed9f84f3c88179cb713bb9f4effbbdba33/deepspeed/runtime/pipe/engine.py#L70). It should pass now as we only compile on forward and not in **init**, nevertheless we should consider keeping the type of self.module.

On the other hand:

1. The model returned by `torch.compile(user_model)` is actually a smart wrapper of the `user_model`. Torch keeps the  `user_model` inside, such that user can add new attributes or change the existing ones in their original model.
2. User may not want their module be compiled by DS.
`compiled_model = torch.compile(self.module,..` solves that.

tohtana: @BacharL I think I get your points in your last comment. Can you clarify the intention of this? This seems that `_compiler_fn` has no difference with `_compiler_backend`.

```python
            backend = self._compiler_fn if self._compiler_fn is not None else self._compiler_backend
	    self.module.compile(backend = backend, **self._compile_kwargs)
```

@deepcharm You can pass a compiled module to deepspeed's init but it will make a slight difference. DeepSpeed sets some hooks but I thought it can't once the model is compiled. I don't think it is harmful to keep `compiler_fn` while such use cases might not be very popular.

tohtana: @BacharL We already allow users to set str or a function to `backend`. Do you mean we can simply remove `_compiler_fn`?
I thought `backend` is given a compiled (but not optimized) graph. I'm not sure if we can keep the flexibility we intended without `_compiler_fn`. My understanding might be wrong. I would appreciate it if you could point out what I might have missed.

BacharL: > I thought backend is given a compiled (but not optimized) graph

I missed that, you are correct. In that case we must keep both, `backend `and `compiler_fn`.
Do you think it is acceptable to ask the user that `compiler_fn `will compile in place, similar to `nn.Module.compile()`?

```
        if self._config.compile_config.enabled not self._is_compiled and is_compile_supported():
            if self._compiler_fn is None:
	        self.module.compile(backend = backend, **self._compile_kwargs)
            else:
                self._compiler_fn(self.module)
	    self._is_compiled = True
```
Alternatively we can verify the returned type and instance is not changed after calling compiler_fn
```
assert type(compiled_model) == type(self.module)
assert compiled_model == self.module
```

tohtana: @BacharL I pushed changes to use the in-place `compile()`.

The verification doesn't seem to work because `compiler_fn` cannot change `self.module` and the assertions always hold. What do you think?

tohtana: @BacharL @deepcharm, I had a discussion with @tjruwase about the design.

He suggested having `compile` API of DeepSpeed instead of running compilation at the first forward pass. The behavior would be easier for users to understand. In this design, we don't need ds config for compile any more. At least for now, it will be a simple wrapper of engine.module.compile().
Do you have any thought on this? I think I can briefly test the feasibility.

BacharL: This suggestion seems good. 
Thanks.

tohtana: @BacharL I implemented the approach. Would it work for you?

BacharL: Great!. Thanks you.

BacharL: Great!, Thank you.

deepcharm: @tohtana This approach is simple and explicit, thanks.

deepspeed/runtime/engine.py
```diff
@@ -3600,3 +3599,20 @@ def empty_partition_cache(self):
...
+        """Compile the module using the specified backend and kwargs.
+        If a compiler_fn is set, it will be used instead of torch.compile().
+        """
+        if not is_compile_supported():
+            raise RuntimeError("compile is not supported in your version of PyTorch.")
+
+        if self.is_compiled:
+            return
+
+        self.module = self.module.compile(backend=backend, **compile_kwargs)
```
BacharL, line 3613:
```suggestion
        self.module.compile(backend=backend, **compile_kwargs)
```
nn.Module.compile returns None

tohtana: Good catch, thank you!

tjruwase (approved):

loadams (approved):

AlongWY:
The Pull Requese make the torch 2.1 and 2.2 compile failed.

PR #5584, opened 2024-05-29
Update profiler.py

gameofdimension:
assume one flop equals two macs


loadams:
@gameofdimension  can you elaborate more on this PR?

loadams:
> @gameofdimension can you elaborate more on this PR and why one flop is 2 macs, usually 1 mac is 2 flops?

Hi @gameofdimension - any thoughts on this PR?

PR #5590, opened 2024-05-30
WA for Torch-compile-Z3-act-apt accuracy issue from the Pytorch repo

NirSonnenschein:
We have been encountered an accuracy issue when running Torch compile + zero3 + activation checkpointing. Specifically some grads gets is zeroed (running without torch compile, this issue is not encountered). This issue was also reproduced by Umesh Chand from the DS team. We found that in the Pytorch repo torch compile has been specifically disabled using the label: @torch._disable_dynamo()
reference to the WA in the Pytorch repo (https://github.com/pytorch/pytorch/blob/ec8b254ef49b4a057cf89c2ae64520fb7b423a3e/torch/utils/checkpoint.py#L324) this indicates that there is some issue with torch compile and checkpointing (not necessarily DS related).

given that the checkpointing function in DeepSpeed is based on the Pytorch function, We propose to adopt this WA to ensure correct behavior (it can be removed later if the underlying issue is fixed) 
Note: this shouldn't impact non-troch compile cases.


tjruwase (approved):

tjruwase:
@umchand, @tohtana FYI

umchand (approved):

tohtana:
@NirSonnenschein Thank you for the PR.

It seems old versions of PyTorch doesn't have `@torch._disable_dynamo()`. Can you try `deepspeed.runtime.compiler.disable` instead? 
https://github.com/microsoft/DeepSpeed/blob/2fc702ed9f84f3c88179cb713bb9f4effbbdba33/deepspeed/runtime/compiler.py#L19C5-L19C12

NirSonnenschein:
> @NirSonnenschein Thank you for the PR.
> 
> It seems old versions of PyTorch don't have `@torch._disable_dynamo()`. Can you try `deepspeed.runtime.compiler.disable` instead? https://github.com/microsoft/DeepSpeed/blob/2fc702ed9f84f3c88179cb713bb9f4effbbdba33/deepspeed/runtime/compiler.py#L19C5-L19C12

thanks for the comment, I've uploaded a new version which should fix this.

NirSonnenschein:
Hi @tohtana , 
would it be possible to re-run CI to check that the new version passes?

NirSonnenschein:
Hi @mrwyattii, 
Would it be possible to review this PR when you have time ?

NirSonnenschein:
thanks @tjruwase, 
the failure in the merge queue tests doesn't seem related to this commit. 
the error seems to be a HTTP error getting a model from huggingface:
E   huggingface_hub.utils._errors.HfHubHTTPError: 429 Client Error: Too Many Requests for url: https://huggingface.co/api/models?cursor=eyJfaWQiOnsiJGd0IjoiNjY0ZTIxNTAyZmI1MWJjZjFlNzdkZjBkIn19

PR #5594, opened 2024-05-31
Feature branch

gvram13541:
-> added the file: deepspeed_aio_base.h which contains base nvme functionality
-> edited files trampoline.cpp, trampoline.h, and py_dsaio, and files in plugins folder
-> i haven't deleted the previous code but't still you can see it commented at the bottom if you scroll.


Yathin-KN:
Added param

PR #5601, opened 2024-06-02
state_dict_factory: llama checkpoint - support SWIGLU

nelyahu:
DeepSpeed supports loading a checkpoint for inference with different DP/TP/PP. This requires to split/merge parameters based on their TP attributes. Currently, this is done by using model specific parameter names. This is not a good practice and should be modified.

This commit handles the required changes to support MDS LLaMA model. There are 2 changes:
- Support for lm_head.weight
- Support for mlp.h_to_4h.weight for SWIGLU

SWIGLU requires different handling, however there is no meta data available that identifies mlp.h_to_4h.weight as SWIGLU. Therefore, for now we use a hack to detect it.


deepspeed/runtime/state_dict_factory.py
```diff
@@ -335,10 +335,26 @@ def merge_state_dict(self,
                         new_client_sd[key] = torch.cat(value_list, axis=0)
                     else:
                         new_client_sd[key] = self.merge_query_key_value(value_list, ckpt_ver)
-            elif "mlp.dense_h_to_4h.weight" in key or "word_embeddings.weight" in key or "mlp.dense_h_to_4h.bias" in key:
-                if quantize and "mlp.dense_h_to_4h.weight" in key:
-                    value_list = quantizer.Quantize(value_list, quantize_bits, groups, key=key)
+            elif "word_embeddings.weight" in key or "mlp.dense_h_to_4h.bias" in key or "lm_head.weight" in key:
                 new_client_sd[key] = torch.cat(value_list, axis=0)
+            elif "mlp.dense_h_to_4h.weight" in key:
```
tjruwase, line 340: @nelyahu, this logic is very old and hacky. Could it not be replaced with Universal Checkpointing?

tohtana (commented):
As Tunji commented, we now have a more flexible solution introduced in #5390. Can you it?

tohtana:
@nelyahu Do we have any update?

loadams:
> @nelyahu Do we have any update?

ping @nelyahu

loadams:
@nelyahu - closing this as stale for now, happy to come back to it, just re-open or tag us

PR #5611, opened 2024-06-04
Add compile backend arg for test_set_compiler_fn

vshekhawat-hlab:
Fix the test_set_compiler_fn UT, current test_set_compiler_fn uses inductor as a backend, which is incompatible with HPU. Added backend arg to torch.compile.


tohtana:
Hi, @vshekhawat-hlab,
we are considering to remove `set_compiler_fn` and its test in #5581. I think we can merge it within this week. Can you wait for a while?

vshekhawat-hlab:
Thanks @tohtana for information.

I am closing this.

PR #5612, opened 2024-06-04
[INF] Enable torch compile for inference

oelayan7:


delock:
Hi @oelayan7 do you have sample code using this feature?  Or does it turn on by default?  I wonder if I can try it with CPU accelerator.

oelayan7:
Hi @delock, 
No, it isn't run by default. The function .compile() needs to be called. 
This changed was adopted in a similar way that was done in zero and for training, which used CompileWrapper that was removed in https://github.com/microsoft/DeepSpeed/pull/5581 and a new approach was introduced.
This change was adopted from the all old PRs before.

oelayan7:
@tjruwase the failing tests are because an HTTP failure with HF. 
Can you please retrigger? (couldn't find a way to do that)

oelayan7:
@tjruwase @tohtana @umchand Can anyone trigger the CI please?

tohtana (approved):
Thank you @oelayan7 !

deepspeed/inference/engine.py
```diff
@@ -6,13 +6,15 @@
 import torch
 import time
 import os
+import deepspeed.utils.nvtx
```
loadams, line 9: @oelayan7 - looks like this is imported but unused, and that's what is causing the formatting failures, could you check and remove this line if so?

loadams: Seems we may need this merged first?

#5697

tohtana: Yes, you are right. Sorry for the confusion

umchand (approved):

deepspeed/inference/engine.py
loadams: That PR is merged now.

loadams (approved):

oelayan7:
@loadams Thanks! 
I fixed the formatting issue, can you please retrigger?

oelayan7:
@loadams I don't think the failures are related.

loadams:
> @loadams I don't think the failures are related.

Correct, working on getting this merged despite some hiccups in our tests not related to your PR.

loadams:
> > @loadams I don't think the failures are related.
> 
> @oelayan7 Correct, working on getting this merged despite some hiccups in our tests not related to your PR.

@oelayan7 - we believe we should be merging all the CI fixes today and will get this merged shortly after that.

PR #5621, opened 2024-06-05
Test just updating HPU docker image

loadams:


vshekhawat-hlab:
This PR not required. Driver update required at Node.

PR #5633, opened 2024-06-09
Monitor was always enabled causing performance degradation

deepcharm:
The Boolean expression for the monitor to be enabled was incorrect, as instead of using the `enabled` field, it used the comet configuration object, making the expression always True.

This caused performance degradation (we've observed ~10% drop) as it erroneously invoked the events logging flow along with the expensive calculation of `loss.mean().item()`.


tjruwase (approved):

nelyahu:
http connection issue caused cpu-torch-latest to fail

alexkuzmik:
@deepcharm 
Hi! It's Alex from Comet. Thanks for noticing and fixing the bug, we appreciate it.

loadams (approved):

Issue #5496, opened 2024-05-02
[BUG] Memory Leak in Stage 2 Optimizer


chiragjn:
**Describe the bug**
I am using transformers Trainer + accelerate to tune language models. What I noticed was post training when I call gc.collect() and torch.cuda.empty_cache() the trainable layers stick around on gpu memory (all ranks)

On adding a debugger and weakref to torch parameters I was able to at least trace it down to _hp_mapping and added methods on top of the params

https://github.com/microsoft/DeepSpeed/blob/f32ad3e1c562be80bd8be6b7b2246dc3041b5bfd/deepspeed/utils/mixed_precision_linkage.py#L33-L37

![image](https://github.com/microsoft/DeepSpeed/assets/10295418/9db0cf0e-7a88-457f-a043-6d22b0c37525)

These references never go to zero even when the optimizer is completely destroyed.
I am a bit puzzled why and wondering if the memory is leaking because torch.tensor / torch.nn.Parameter are not entirely python implementations and adding methods on top can cause unaccounted reference leaks

**To Reproduce**
1. Train a model with Deepspeed 2 Config - no offload
```
{
  "fp16": {
    "enabled": false,
    "auto_cast": false,
    "loss_scale": 0,
    "loss_scale_window": 1000,
    "initial_scale_power": 32,
    "hysteresis": 2,
    "min_loss_scale": 1
  },
  "bf16": {
    "enabled": true
  },
  "optimizer": {
    "type": "AdamW",
    "params": {
      "lr": "auto",
      "betas": "auto",
      "eps": "auto",
      "weight_decay": "auto"
    }
  },
  "activation_checkpointing": {
    "partition_activations": false,
    "cpu_checkpointing": false
  },
  "zero_optimization": {
    "stage": 2,
    "overlap_comm": true,
    "contiguous_gradients": true,
    "sub_group_size": 1e9,
    "reduce_bucket_size": "auto",
    "allgather_partitions": true,
    "allgather_bucket_size": 2e8,
    "reduce_scatter": true,
    "round_robin_gradients": true
  },
  "gradient_accumulation_steps": "auto",
  "gradient_clipping": "auto",
  "steps_per_print": 2000,
  "train_batch_size": "auto",
  "train_micro_batch_size_per_gpu": "auto",
  "wall_clock_breakdown": false,
  "zero_allow_untested_optimizer": true
}
```

Note: some of the auto values are filled in by the HF Trainer. I'll try and get these values soon

[...]

PR #5643, opened 2024-06-11
Fix memory leak from _hp_mapping

chiragjn:
See #5496 
I don't really know if this is a good solution

### DIFF
--- a/deepspeed/runtime/bf16_optimizer.py
+++ b/deepspeed/runtime/bf16_optimizer.py
@@ -94,6 +94,15 @@ class BF16_Optimizer(ZeROOptimizer):
 
         see_memory_usage('end bf16_optimizer', force=True)
 
+    def destroy(self):
+        for i, _ in enumerate(self.optimizer.param_groups):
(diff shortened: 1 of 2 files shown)


deepspeed/runtime/bf16_optimizer.py
```diff
@@ -94,6 +94,15 @@ def __init__(self,
 
         see_memory_usage('end bf16_optimizer', force=True)
 
+    def destroy(self):
+        for i, _ in enumerate(self.optimizer.param_groups):
+            for p in self.bit16_groups[i]:
```
loadams, line 99: FYI @chiragjn - could you run the pre-commit formatter, that will resolve the formatting tests.  Thanks!

chiragjn: Done :)

tjruwase (approved):

chiragjn:
Apologies, I had made some mistakes in bf16 optimizer and couldn't get to fixing them sooner, I should have tested things locally 😅

PR #5666, opened 2024-06-17
fix IDEX dependence in xpu accelerator

Liangliang-Ma:
We don't use IDEX in xpu accelerator from Deepspeed.
Fix this hardcode.


Liangliang-Ma:
@tjruwase Please help to review it. Thanks

PR #5669, opened 2024-06-17
[CPU] add fp16 support to shm inference_all_reduce

delock:
This PR adds FP16 support to DeepSpeed SHM inference_all_reduce.  Previously only FP32 and BF16 is supported.  This is to align with PyTorch CPU support on FP16 datatype.


adk9:
Could you consider adding some unit tests to perhaps `test_dist.py` to test support for the different data types?

delock:
> Could you consider adding some unit tests to perhaps `test_dist.py` to test support for the different data types?

Let me see if I can add some tests.

delock:
Hi @adk9, `TestDistInferenceAllReduce` is modified to test fp32, bf16 and fp16.  Can you help start the workflow?   Thanks!

adk9 (approved):

delock:
I'm seeing test for `TestDistInferenceAllReduce` dtype1 and dtype2 passed, but dtype0 didn't pass.  This is kind wierd because dtype0 supposed to be FP32.

I'll add a sanity check to the workflow to see whether the target machine support all three datatypes.   Will get back to you later @adk9

delock:
@adk9  I wonder what is the pytest used in DeepSpeed workflow.  I tried to execute `pytest $PYTEST_OPTS -n 4 unit/ --torch_ver="2.3"` on my system manually but was told pytest does not recognize argument `-n`.   Is it aliased to something else in the runner system?

delock:
Hi @adk9 the failure for FP32 allreduce is due to modified UT test `world_size=1, 2, or 4` and there was an unnecessary assertion for world_size == 1.   Now this assertion had been removed given the code can handle this situation well.   Can you help restart the workflow?  Thanks!

PR #5680, opened 2024-06-18
Fix numpy upgrade to 2.0.0 BUFSIZE import error

Yejing-Lai:
Numpy 2.0.0 removed the BUFSIZE API, replacing it with 'numpy._core.umath.BUFSIZE'.

The minimum numpy version supported by the new code is 1.26.1. If you need to support the lower numpy version, please tell me and we can add another path.

### DIFF
--- a/deepspeed/autotuning/scheduler.py
+++ b/deepspeed/autotuning/scheduler.py
@@ -5,7 +5,11 @@
 
 import copy
 
-from numpy import BUFSIZE
+import numpy
+if numpy.__version__ < '2.0.0':
+    from numpy import BUFSIZE
+else:
+    from numpy._core.umath import BUFSIZE
 import json
 import subprocess
 import sys


tjruwase:
@Yejing-Lai, thanks for the PR. Please add support for older numpy versions.

Yejing-Lai:
> @Yejing-Lai, thanks for the PR. Please add support for older numpy versions.

Added older version support. please kindly review. Thanks~

deepspeed/autotuning/scheduler.py
```diff
@@ -5,7 +5,11 @@
 
 import copy
 
-from numpy import BUFSIZE
+import numpy
+if numpy.__version__ < '2.0.0':
```
loadams, line 9: @Yejing-Lai - we are working to determine if we need BUFSIZE from this file or from deepspeed/autotuning/constants.py (which is used in deepspeed/autotuning/autotuner.py), if we do need this from np, we will merge this.

mrwyattii (approved):

loadams (approved):

adk9 (approved):

PR #5688, opened 2024-06-20
ENV var added for recaching in INF Unit tests

raza-sikander:
FORCE_UPDATE_HF_CACHE set to True to recache.
HF_CACHE_EXPIRY_DAYS to set the number of cache expiry days


tjruwase (approved):

tjruwase:
@raza-sikander, thank for this PR. Can you please see the following link to fix the formatting issue?
https://github.com/microsoft/DeepSpeed/blob/master/CONTRIBUTING.md#contributing

tests/unit/inference/test_inference.py
```diff
@@ -97,7 +98,8 @@ def _hf_model_list() -> List[ModelInfo]:
     current_time = time.time()
 
     # Update the cache if it has expired
-    if (model_data["cache_time"] + cache_expiration_seconds) < current_time:
+    if ((model_data["cache_time"] + cache_expiration_seconds) < current_time) or os.getenv("FORCE_UPDATE_HF_CACHE",
```
loadams, line 101: Thanks for this @raza-sikander - not sure if we should document this anywhere though?

loadams (approved):

tests/unit/inference/test_inference.py
loadams: This is the only place it is used, so this should be fine to leave.  Thanks for this change.

PR #5691, opened 2024-06-21
sequence parallel with communication overlap

inkcherry:
SP is a fantastic piece of work, it is very elegant and concise， at the current stage, a transformer layer's forward and backward passes involve 8 all-to-all operations, with 5 opportunities for overlapping communication:

Forward pass: The QKV matrix operations can be pipelined alongside some of the all-to-all communications.
Backward pass: DQ, DK, DV all-to-all communications can be pipelined alongside matrix operations.
Backward pass: DO_w can be parallel with DO_input, involving matrix operations and all-to-all communications. Similar overlap-comm strategies are used in Megatron for TP/TP-sp parallelism.
I tested under conditions of  1N8C zero1, disabled activation checkpointing, ds-sp=8, and gbs=16:
1B 64K
7B 16K
They showed over 10% improvement (where I found that for mega-ds, using split QKV itself can also enhance performance due to reducing slice + cat operations in fwd/bwd), despite some TFLOPs already performing at a relatively good level.
co-work with https://github.com/microsoft/Megatron-DeepSpeed/pull/415


Edenzzzz:
overlapping only happens when computation doesn't depend on communication?

inkcherry:
> overlapping only happens when computation doesn't depend on communication?
@Edenzzzz 
Yes, manual control of some dependencies is required。

samadejacobs:
@inkcherry, many thanks for this excellent contribution to DeepSpeed codebase. To help with our review, could you please add (1) unit test(s) and (2) numbers on parallel performance improvements (throughput and latency) to the pull request? Your continuous and remarkable contributions to DeepSpeed are appreciated.

tohtana (approved):

Edenzzzz:
@inkcherry Thanks for your insight! Can I ask why we need sp_stream here, as it seems to be never used, e.g. by torch.cuda.stream(sp_stream)?

inkcherry:
> @inkcherry Thanks for your insight! Can I ask why we need sp_stream here, as it seems to be never used, e.g. by torch.cuda.stream(sp_stream)?
hi @Edenzzzz 
apology for missing your comments,   I noticed that DeepSpeed's sequence is designed in a modular way, which means we can't freely insert communication calls (comm) anywhere we want to use async_op. When computation happens before communication, since the collective communication library (CCL) will automatically sync with the default stream, we need to use a custom stream or even an event to achieve parallelism between computation and communication. It's also crucial to maintain the dependencies between them properly. The setup code for this is in Megatron's DeepSpeed implementation.

PR #5697, opened 2024-06-25
Disable nvtx decorator to avoid graph break

tohtana:
`instrument_w_nvtx` breaks a graph as `range_push` and `range_pop` return a non-tensor int.
This PR disables the decorator to avoid the break graph.

This actually impacts the performance. In my environment, the training iteration time using Llama-3-8B/4GPUs/ZeRO1 is improved  from 3.02s -> 2.54s.


deepspeed/runtime/engine.py
```diff
@@ -3608,6 +3608,9 @@ def compile(self, backend=get_accelerator().get_compile_backend(), compile_kwarg
         """Compile the module using the specified backend and kwargs.
         If a compiler_fn is set, it will be used instead of torch.compile().
         """
+        # Avoid graph breaks
+        deepspeed.utils.nvtx.enable_nvtx = False
```
umchand, line 3612: @tohtana Do you want to set this based on the config - "compiler enabled =True"?

deepspeed/utils/nvtx.py
```diff
@@ -5,15 +5,19 @@
 
 from deepspeed.accelerator import get_accelerator
 
+enable_nvtx = True
```
umchand, line 8: if we set it based on the config as I mentioned above, you will not have to set it to True here

deepspeed/runtime/engine.py
tohtana: We don't have the config any more

umchand (approved):

loadams (approved):

PR #5699, opened 2024-06-27
Add Windows scripts (deepspeed, ds_report).

costin-eseanu:


bin/deepspeed.bat
```diff
@@ -0,0 +1,2 @@
+@echo off
+python "%~dp0\ds" %*
```
loadams, line 2: Looks like the formatter wants a newline at the end of these .bat files, that shouldn't cause any issues, right?

costin-eseanu: No, that would work fine, let me fix it.

loadams (approved):

PR #5700, opened 2024-06-27
Fix memory leak for pipelined optimizer swapper

mauryaavinash95:
We identified a memory leak when training with NVMe offloaded optimizer states. The issue occurs when `pipeline_write=true`, where the tensors that have swapped out and written to NVMe are not deallocated, leading to a memory leak.

This PR resolves the issue by deallocating the unused tensors which have swapped out to NVMe.

### DIFF
--- a/deepspeed/runtime/swap_tensor/pipelined_optimizer_swapper.py
+++ b/deepspeed/runtime/swap_tensor/pipelined_optimizer_swapper.py
@@ -8,6 +8,7 @@ Functionality of swapping optimizer tensors to/from (NVMe) storage devices.
 
 from deepspeed.ops.op_builder import AsyncIOBuilder
 from deepspeed import comm as dist
+import torch
 
 from deepspeed.runtime.swap_tensor.constants import *
 from deepspeed.runtime.swap_tensor.utils import swap_in_tensors, swap_out_tensors, print_object
@@ -154,6 +155,8 @@ class PipelinedOptimizerSwapper(OptimizerSwapper):
 
     def _complete_swap_out(self, swap_out_type):
         self.swap_ops[swap_out_type].wait()
+        for buffer in self.swap_ops[swap_out_type].state_buffers:
+            buffer = torch.Tensor()
         self.swap_buffer_manager.free(self.swap_ops[swap_out_type].allocated_buffers)
         self.swap_ops[swap_out_type] = None
 


deepspeed/runtime/swap_tensor/pipelined_optimizer_swapper.py
```diff
@@ -154,6 +154,8 @@ def swap_out_gradients(self, parameter, gradient_offsets, gradient_tensors):
 
     def _complete_swap_out(self, swap_out_type):
         self.swap_ops[swap_out_type].wait()
+        for i in range(len(self.swap_ops[swap_out_type].state_buffers)):
```
adk9, line 157: Minor nit, but you can rewrite this in a more readable form as:
```
for buffer in self.swap_ops[swap_out_type].state_buffers:
    buffer = torch.Tensor()
```

for in range(len()) is a bit of anti-pattern when the index is not strictly needed.

adk9 (approved):

deepspeed/runtime/swap_tensor/pipelined_optimizer_swapper.py
tjruwase: @mauryaavinash95, please make this change.

tjruwase:
@mauryaavinash95, thanks for fixing this leak. As this is not a commonly used feature, I am curious if you are seeing any benefits from the pipelined swapper?

deepspeed/runtime/swap_tensor/pipelined_optimizer_swapper.py
mauryaavinash95: Thanks for the suggestion @adk9 and @tjruwase. I've made the change (76b63c888ceaa37e5be0f61149b44b08b3aa486f).

mauryaavinash95:
@tjruwase: We see a 20-30% improvement in the optimizer update phase when using the pipelined swapper-- most likely because of the full-duplex PCIe through which the NVMes are connected. We are running some extensive experiments on this and can share some results in a few weeks if you're interested.

tjruwase:
> @tjruwase: We see a 20-30% improvement in the optimizer update phase when using the pipelined swapper-- most likely because of the full-duplex PCIe through which the NVMes are connected. We are running some extensive experiments on this and can share some results in a few weeks if you're interested.

@mauryaavinash95, wow! It is great to learn that you are seeing speedups from pipelined swapper. Thanks for sharing. 

The back story is that we actually never got a chance to properly evaluate the perf benefits, but released the code anyways in hope that someday somebody might find it useful :). And so, I am very curious about your work and results in this line, if you don't mind sharing.

mauryaavinash95:
Sorry, I accidentally made some other commits to my master branch which caused this PR to get closed and I had to open it again.

PR #5702, opened 2024-06-28
update xpu fusedadam opbuilder for pytorch 2.3

baodii:
update the way to get queue for FusedAdam OpBuilder.


loadams (approved):

loadams:
@baodii - do we need to test this with pytorch 2.3?  Should we update the container [here](https://github.com/microsoft/DeepSpeed/blob/master/.github/workflows/xpu-max1100.yml#L39C7-L39C58) to one with 2.3?

loadams:
> @baodii - do we need to test this with pytorch 2.3? Should we update the container [here](https://github.com/microsoft/DeepSpeed/blob/master/.github/workflows/xpu-max1100.yml#L39C7-L39C58) to one with 2.3?

It appears one isn't published yet, so we can merge this

https://github.com/intel/intel-extension-for-pytorch/releases

PR #5715, opened 2024-07-01
Unit Test: Add error handling for rate limit exceeded in model list

HeyangQin:
This PR fixes the random failure in our unit test due to HTTP 429


loadams (approved):

adk9:
Thanks for the fix. I think this will definitely make `test_inference` less flaky.

I've two high-level concerns though about what's happening here:

1. What's the reason for the default model cache time to be 1 day? The models/tasks we're testing barely update so we could perhaps set a much longer cache time?
2. We get rate-limited because we're making ~750k requests (getting model information for all models on HF) to test only 24 models. I think a better approach is to pass a filter to `api.list_models()` to get information of only the models we're interested in testing?

loadams:
> Thanks for the fix. I think this will definitely make `test_inference` less flaky.
> 
> I've two high-level concerns though about what's happening here:
> 
> 1. What's the reason for the default model cache time to be 1 day? The models/tasks we're testing barely update so we could perhaps set a much longer cache time?
> 2. We get rate-limited because we're making ~750k requests (getting model information for all models on HF) to test only 24 models. I think a better approach is to pass a filter to `api.list_models()` to get information of only the models we're interested in testing?

It was re-set here, though we had it set to 1 day before this: https://github.com/microsoft/DeepSpeed/pull/5688/files

adk9 (approved):

loadams (approved):

PR #5756, opened 2024-07-08
Update node16 check on self-hosted runners and remove python 3.6

loadams:
With changes from GitHub [finally deprecating](https://github.com/actions/checkout/issues/1474) [node16 based runners](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/) (which the checkout@v3 action uses) we need to make changes to support this.  

To do this, there are two changes.  First we remove the python 3.6 check as with the changes in pydantic v2 that will be merged soon, we will be removing this check there, so we can more easily remove it now so that future PRs are cleaner and it is clear why some changes have been made.

Second, node16 is the default on some of our self-hosted runners.  To work around tests failing on these, we [set the GitHub env var to override this check](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/).

Other relevant links:
https://github.com/actions/checkout/issues/1474
https://github.com/easybuilders/easybuild-framework/pull/4574/files
https://github.com/actions/checkout/issues/1809
https://github.com/actions/runner/issues/3373
https://github.com/actions/checkout/issues/1809

### DIFF
--- a/.github/workflows/cpu-inference.yml
+++ b/.github/workflows/cpu-inference.yml
@@ -24,6 +24,8 @@ jobs:
   unit-tests:
     runs-on: [self-hosted, cpu]
 
+    env: {ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true} # Allow using Node16 actions
+
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/nv-lightning-v100.yml
+++ b/.github/workflows/nv-lightning-v100.yml
@@ -21,6 +21,8 @@ jobs:
   unit-tests:
     runs-on: [self-hosted, nvidia, cu111, v100]
 
+    env: {ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true} # Allow using Node16 actions
+
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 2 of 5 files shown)


loadams:
Merged #5755 into this PR to have a single PR that fixes both issues.

loadams:
Workflow actions for nv-lightning and python are tested here, linked below are builds for nv-torch110-v100 and nv-torch110-p40:

https://github.com/microsoft/DeepSpeed/actions/runs/9845125845
https://github.com/microsoft/DeepSpeed/actions/runs/9845127511

adk9 (approved):

.github/workflows/cpu-inference.yml
```diff
@@ -24,6 +24,8 @@ jobs:
   unit-tests:
     runs-on: [self-hosted, cpu]
 
+    env: {ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true} # Allow using Node16 actions
```
adk9, line 27: Would this also work with checkout@v4, i.e. force it to use node16? In that case, should we consider upgrading to v4 actions for all our workflows?

loadams: I think checkout@v4 supports only node20, which is why we couldn't upgrade those workflows yet.  I think we should just work on getting those runners to use images that contain node20/node22/updated containers so we can remove this envvar entirely

loadams:
nv-torch-latest-v100 is failing and known, bypassing to get these fixes merged.

PR #5759, opened 2024-07-09
move is_checkpointable call reducing torch.compile Graph breaks

NirSonnenschein:
We have encountered a performance issue when running torch compile on a model utilizing
the pipeline engine (Mixtral). 
The issue was found to be the is_checkpointable function which is called in the engine's forward function.
This function creates a graph break when using torch.compile leading to decreased performance (particularly since this happens in every forward call). We propose a change in the way is_checkpointable is checked by precomputing and storing its value before the forward call and accessing the stored values in the forward function.
given this change the graph break in the forward call is avoided which should lead to better performance for torch compile.

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -213,6 +213,8 @@ class PipelineEngine(DeepSpeedEngine):
                 self.module.activation_checkpoint_func = ds_checkpointing.non_reentrant_checkpoint
                 if self.grid.get_global_rank() == 0:
                     logger.info(f'CONFIG: activation_checkpoint_func=non_reentrant_checkpoint')
+        if self.module.activation_checkpoint_interval > 0:
+            self.module._precompute_checkpointable_values()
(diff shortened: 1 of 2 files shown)


NirSonnenschein:
Hi @duli2012, 
when you have a moment would it be possible to review this change.
thanks

tjruwase:
> Hi @duli2012, when you have a moment would it be possible to review this change. thanks

@tohtana, will help review.

NirSonnenschein:
Hi @tohtana,
when you have a moment would it be possible to review this change. 
thanks

tohtana (approved):

tohtana:
@NirSonnenschein This is a great improvement. Sorry for the delay. I just approved.

loadams:
The current test failures will be resolved after https://github.com/microsoft/DeepSpeed/pull/5797 completes.

NirSonnenschein:
Thanks @tohtana and @loadams

PR #5763, opened 2024-07-10
Add DataStates-LLM: Asynchronous Checkpointing Engine Support

mauryaavinash95:
We are a team at Argonne National Laboratory working on low-overhead asynchronous checkpointing approaches for LLMs and transformers. As part of these efforts, we have developed DataStates-LLM, a library that we would like to contribute to the DeepSpeed community:
https://github.com/datastates/datastates-llm

The key idea we leverage is to allow non-blocking tensor copies during the forward pass and backward pass from the GPU to the host. Only if these copies do not finish until the update phase, then we block. Meanwhile, from the host memory, the tensors are flushed asynchronously to durable storage (parallel file systems, local SSDs, etc). 

To enable this capability, our initial implementation makes the scheduler aware of checkpointing, calling a ckpt.wait() primitive before starting the update phase. We illustrated this with the pipeline scheduler. We are also considering a scheduler-independent solution that integrates with DeepSpeed/Megatron and provides a hook for the start of the update phase, which we can leverage to run ckpt.wait().

We appreciate your feedback and look forward to a collaboration in this space.


.github/workflows/cpu-inference.yml
```diff
@@ -24,6 +24,8 @@ jobs:
   unit-tests:
     runs-on: [self-hosted, cpu]
 
+    env: {ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true} # Allow using Node16 actions
```
loadams, line 27: @mauryaavinash95 - can you merge the latest DeepSpeed master branch into this dev branch so we can see these merge artifacts resolved now that they are merged into DeepSpeed?

loadams:
Hi @mauryaavinash95 - following up on this, would you be able to resolve the merge conflicts with this PR if you are still planning on completing it?

loadams:
Hi @mauryaavinash95 - I'm going to close this PR now as stale, but we would appreciate the contribution if you are able to come back to this and resolve the merge conflicts?

PR #5764, opened 2024-07-11
Add fp8-fused gemm kernel

sfc-gh-reyazda:
This PR adds the new fused kernel for the Dense GeMM using fp8-quantized weight.

### DIFF
--- a/csrc/fp_quantizer/fp_quantize.cpp
+++ b/csrc/fp_quantizer/fp_quantize.cpp
@@ -108,9 +108,22 @@ void selective_dequantize(torch::Tensor& val,
 #endif
 }
 
+at::Tensor get_scales(torch::Tensor& out, int num_groups)
+{
+    auto options = at::TensorOptions()
+                       .dtype(torch::kFloat)
+                       .layout(at::kStrided)
+                       .device(at::kCUDA)
+                       .requires_grad(false);
+    auto scales =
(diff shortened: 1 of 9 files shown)


HeyangQin (approved):

jeffra:
One thing that needs to be resolved before merging this, this kernel requires `triton==2.3.0`. This should be checked at runtime and communicated to users somehow.

HeyangQin:
Hi @jeffra. To clarify, does this kernel require exactly `triton==2.3` or `triton>=2.3`?

jeffra:
> Hi @jeffra. To clarify, does this kernel require exactly `triton==2.3` or `triton>=2.3`?

@sfc-gh-reyazda would know better, I am not sure if we've tested with newer triton than 2.3.0. I have not personally tested this at least.

sfc-gh-reyazda:
> Hi @jeffra. To clarify, does this kernel require exactly `triton==2.3` or `triton>=2.3`?

It needs that specific version, unfortunately triton keeps changing/improving and their APIs change too so it is hard to track it properly. That's also another motivation to move to cutlass soon and have a more solid implementation to work independent of other libraries. On the other hand, Triton gives the flexibility to run on various hardwares. So, it is always a tradeoff. I think we need to have some more discussions on such dependencies later in a different discussion.
Best,
Reza

PR #5780, opened 2024-07-18
Fix CPU Adam JIT compilation

lekurile:
This PR fixes CPU Adam JIT compilation by including the `CUDA_LIB64` path in the `extra_ldflags` list before calling `load()`.

### DIFF
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -552,13 +552,6 @@ class OpBuilder(ABC):
             os.environ["PYTORCH_ROCM_ARCH"] = self.get_rocm_gpu_arch()
             cxx_args.append('-DROCM_WAVEFRONT_SIZE=%s' % self.get_rocm_wavefront_size())
 
-        print(f"{self.name=}")
-        print(f"{sources=}")
-        print(f"{extra_include_paths=}")
-        print(f"{cxx_args=}")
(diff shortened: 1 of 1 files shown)


op_builder/builder.py
```diff
@@ -772,25 +772,32 @@ def libraries_args(self):
...
+
     def extra_ldflags(self):
         if self.build_for_cpu:
             return ['-fopenmp']
 
         if not self.is_rocm_pytorch():
-            return ['-lcurand']
+            ld_flags = ['-lcurand']
+            if not self.build_for_cpu:
+                ld_flags.append(f'-L{self.get_cuda_lib64_path()}')
```
loadams, line 792: Do you want a space in this before the append?

lekurile: Since we're appending to a list, I don't think we need a space here, since it'll just be a list of string args that get passed into the `load()` function.

lekurile:
Manual ds-chat run:
https://github.com/microsoft/DeepSpeed/actions/runs/9995112919

op_builder/builder.py
loadams: Yep, nevermind

loadams (approved):

shawnccx:
When will the next release be cut to have this fix? @loadams

PR #5781, opened 2024-07-18
Non-reentrant checkpointing hook fix

ic-synth:
This PR adds an extra condition to attach backward pass hooks to leaf nodes only if Synchronisation or Profiling is enabled, as otherwise these hooks are not necessary. Hook code below:

```
    def after_backward_hook(_nonuse_grads):
        """the hook registered to all leaf tensors"""
        nonlocal leaf_tensors, backward_visited_leaf_nodes
        backward_visited_leaf_nodes += 1

        if backward_visited_leaf_nodes == len(leaf_tensors):
            see_memory_usage("After backward checkpointing code after backward", force=False)

            if PROFILE_TIME:
                timers('backward').stop()
                timers.log(['backward'])
            if SYNCHRONIZE:
```

see_memory_usage is nevel used, as `force` is hardcoded to `False`. Thus this hook only does any real work only when  PROFILE_TIME or SYNCHRONIZE is True. Otherwise it creates unnecessary function calls


tjruwase (approved):

ic-synth:
Is there a way to restart the ci check without pushing code? It seems like the runner has failed

ic-synth:
@loadams hey! i think some other tests have started to fail, even though i didn't change the parts of the code related to them

ic-synth:
Screenshot 2024-07-18 at 21 31 40
For additional context this small fix has a pretty substantial effect on execution. Our model got 50% slower after 1/20 of the training, but with this fix there is no degradation of the execution speed.

loadams:
> @loadams hey! i think some other tests have started to fail, even though i didn't change the parts of the code related to them

Thanks, we know we have some instability in CI, I'll monitor this PR, but we are working on resolving things and will work on getting this PR merged.

PR #5789, opened 2024-07-22
Fixes for latest Huggingface_hub changes on modelId -> id

loadams:
PRs in huggingface_hub that mirror this:

https://github.com/huggingface/huggingface_hub/pull/2405

### DIFF
--- a/tests/unit/inference/test_inference.py
+++ b/tests/unit/inference/test_inference.py
@@ -77,7 +77,7 @@ _test_tasks = [
 
 @dataclass
 class ModelInfo:
-    modelId: str
+    id: str
     pipeline_tag: str
     tags: List[str]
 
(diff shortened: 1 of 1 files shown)


lekurile (approved):

loadams:
Tested locally, the failure appears to be with the cached versions we keep, so going to force merge this PR.

julien-c:
sorry about the breaking change on our side, and thanks for quickly pushing this fix ❤️ 

(we have reverted the change server side for now but we will again remove `modelId` in the future)

loadams:
> sorry about the breaking change on our side, and thanks for quickly pushing this fix ❤️
> 
> (we have reverted the change server side for now, but we plan to re-apply it in the future)

Thanks @julien-c, we have made all changes to support the new field now, so we should be fine in the future.

PR #5797, opened 2024-07-23
Update torch version in cpu-torch-latest and nv-torch-latest-v100 tests to 2.4

loadams:
Now that the tests have moved to using torch 2.4, we need to update the tests or they will fail.


lekurile (approved):

HeyangQin (approved):

loadams:
Both failures were OOM, re-running to test.

PR #5802, opened 2024-07-26
fix: handle exception when loading cache file in test_inference.py

HeyangQin:
This PR is to fix CI failures such as https://github.com/microsoft/DeepSpeed/actions/runs/10085903860/job/27887546470#step:8:3616
cc @tjruwase

### DIFF
--- a/tests/unit/inference/test_inference.py
+++ b/tests/unit/inference/test_inference.py
@@ -94,7 +94,10 @@ def _hf_model_list() -> List[ModelInfo]:
     model_data = {"cache_time": 0, "model_list": []}
     if os.path.isfile(cache_file_path):
         with open(cache_file_path, 'rb') as f:
-            model_data = pickle.load(f)
+            try:
+                model_data = pickle.load(f)
+            except Exception as e:
+                print(f"Error loading cache file {cache_file_path}: {e}")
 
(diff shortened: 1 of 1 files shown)


tests/unit/inference/test_inference.py
```diff
@@ -94,7 +94,10 @@ def _hf_model_list() -> List[ModelInfo]:
     model_data = {"cache_time": 0, "model_list": []}
     if os.path.isfile(cache_file_path):
         with open(cache_file_path, 'rb') as f:
-            model_data = pickle.load(f)
+            try:
+                model_data = pickle.load(f)
+            except Exception as e:
```
tjruwase, line 99: Can you expand on how catching and printing the exception improves robustness?

HeyangQin: This is to prevent corrupted cache file from crashing all following runs. From my point of view, this caching is not really needed after the previous fix https://github.com/microsoft/DeepSpeed/pull/5715 that reduced the number of model info from thousands to <100. But as I am very occupied by the benchmarking effort, I just want to roll in a quick fix to unblock the CI

tjruwase (approved):

loadams (approved):

PR #5806, opened 2024-07-29
Fix op_builder for CUDA 12.5

keshavkowshik:
std lib needed to be updated to C++ version 20 for CUDA 12.5 to fix compilation issues in the op_builder.

TODO:
Fix may need to be extended to CUDA 12.4, needs testing.

### DIFF
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -756,11 +756,11 @@ class CUDAOpBuilder(OpBuilder):
                 else:
                     std_lib = '-std=c++17'
             else:
-                std_lib='-std=c++14'
+                std_lib = '-std=c++14'
             args += [
-                '-allow-unsupported-compiler' if sys.platform == "win32" else '', '--use_fast_math',
-                std_lib, '-U__CUDA_NO_HALF_OPERATORS__',
-                '-U__CUDA_NO_HALF_CONVERSIONS__', '-U__CUDA_NO_HALF2_OPERATORS__', f'--threads={nvcc_threads}'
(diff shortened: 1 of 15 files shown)


op_builder/builder.py
```diff
@@ -749,10 +749,17 @@ def nvcc_args(self):
             except ValueError:
                 nvcc_threads = min(os.cpu_count(), 8)
 
-            cuda_major, _ = installed_cuda_version()
+            cuda_major, cuda_minor = installed_cuda_version()
+            if cuda_major > 10:
+                if cuda_major==12 and cuda_minor==5:
```
loadams, line 754: If this is a breaking change going forward, should this be `cuda_minor >= 5`

lekurile: Agreed, good catch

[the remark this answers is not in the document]
keshavkowshik: Thanks for pointing it out, Logan :)

lekurile (approved):

op_builder/builder.py
loadams: @keshavkowshik - this doesn't look to be updated?

keshavkowshik: Now you should see it reflected.

loadams: Yes thanks!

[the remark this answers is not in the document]
loadams: You can run to check locally by running `pre-commit run --all-files`

loadams (approved):

Issue #5682, opened 2024-06-18
[BUG] Logs full of FutureWarning when training with nightly PyTorch


rosario-purple:
**Describe the bug**
When training with nightly PyTorch, the logs are full of deprecation warnings like this:

```
/home/alyssavance/miniforge3/envs/brr/lib/python3.10/site-packages/deepspeed/runtime/zero/linear.py:49: FutureWarn\
ing: `torch.cuda.amp.custom_fwd(args...)` is deprecated. Please use `torch.amp.custom_fwd(args..., device_type='cu\
da')` instead.
[860302:0]:/home/alyssavance/miniforge3/envs/brr/lib/python3.10/site-packages/deepspeed/runtime/zero/linear.py:67:\
 FutureWarning: `torch.cuda.amp.custom_bwd(args...)` is deprecated. Please use `torch.amp.custom_bwd(args..., devi\
ce_type='cuda')` instead.
```

**To Reproduce**
Fine-tune a model with Zero-3 while using nightly PyTorch.

**Expected behavior**
The deprecation warning shouldn't appear.

**ds_report output**
[2024-06-18 08:59:11,766] [INFO] [real_accelerator.py:203:get_accelerator] Setting ds_accelerator to cuda (auto detect)
/home/alyssavance/miniforge3/envs/brr/lib/python3.10/site-packages/deepspeed/runtime/zero/linear.py:49: FutureWarning: `torch.cuda.amp.custo\
m_fwd(args...)` is deprecated. Please use `torch.amp.custom_fwd(args..., device_type='cuda')` instead.
  def forward(ctx, input, weight, bias=None):
/home/alyssavance/miniforge3/envs/brr/lib/python3.10/site-packages/deepspeed/runtime/zero/linear.py:67: FutureWarning: `torch.cuda.amp.custo\
m_bwd(args...)` is deprecated. Please use `torch.amp.custom_bwd(args..., device_type='cuda')` instead.
  def backward(ctx, grad_output):
Warning: The default cache directory for DeepSpeed Triton autotune, /home/alyssavance/.triton/autotune, appears to be on an NFS system. Whil\
e this is generally acceptable, if you experience slowdowns or hanging when DeepSpeed exits, it is recommended to set the TRITON_CACHE_DIR e\
nvironment variable to a non-NFS path.
^[[93m [WARNING] ^[[0m Please specify the CUTLASS repo directory as environment variable $CUTLASS_PATH
[...]

PR #5811, opened 2024-07-30
Update linear.py compatible with torch 2.4.0

terry-for-github:
deepspeed/runtime/zero/linear.py:67: FutureWarning: `torch.cuda.amp.custom_bwd(args...)` is deprecated. Please use `torch.amp.custom_bwd(args..., device_type='cuda')` instead.

Fixes: #5682


loadams (approved):

loadams:
Thanks @terry-for-github!  I'm running some tests on non-cuda hardware and we will merge this after tests complete.

deepspeed/runtime/zero/linear.py
```diff
@@ -33,8 +34,8 @@ def print_rank_0(message, debug=False, force=False):
 
 
 try:
-    autocast_custom_fwd = get_accelerator().amp().custom_fwd
-    autocast_custom_bwd = get_accelerator().amp().custom_bwd
+    autocast_custom_fwd = functools.partial(torch.amp.custom_fwd, device_type=get_accelerator().device_name())
```
loadams, line 37: Thanks @terry-for-github - It looks like torch.amp is introduced in torch 1.11, @tjruwase thoughts on keeping support for older torch versions?

loadams: @terry-for-github 

Could you add something like

```
if hasattr(torch, 'amp'):
    autocast_custom_fwd = functools.partial(torch.amp.custom_fwd, device_type=get_accelerator().device_name())
    autocast_custom_bwd = functools.partial(torch.amp.custom_bwd, device_type=get_accelerator().device_name())
else:
    autocast_custom_fwd = get_accelerator().amp().custom_fwd
    autocast_custom_bwd = get_accelerator().amp().custom_bwd
```

That way we can support older torch versions too (like 1.11)

terry-for-github: sure! I will try it.

terry-for-github: torch.amp.custom_f/bwd only exist in torch 2.4 even though torch.amp is introduced after torch 1.11(not included).

So what I understand is that the original code supports torch 1.11~2.3(or maybe older, I'm not sure) and my code only supports 2.4. What I need to do is fix my code so that it supports 1.11~2.4.

Am I right?

terry-for-github: See ab7ab617c1aec835a15759ea5bf90ec0aa7d7fad

terry-for-github: I think there is something wrong with the unit test.

loadams: Yes, we've fixed the unit tests now, thanks!

PR #5840, opened 2024-08-05
move pynvml install to setup.py

Rohan138:
Only install pynvml on nvidia gpus; not all accelerators


loadams (approved):

Rohan138:

loadams:
Hi @Rohan138 - we're not able to re-sync master with your branch, could you please do that, that way the tests will re-run with the new changes we have merged?  Thanks!

PR #5877, opened 2024-08-08
Use `torch.nan_to_num` replace numpy wrapper one

jinyouzhi:


tjruwase:
@jinyouzhi, thanks for the PR. How can we avoid breaking backwards-compatibility with older pytorch versions >= 1.12?

jinyouzhi:
> @jinyouzhi, thanks for the PR. How can we avoid breaking backwards-compatibility with older pytorch versions >= 1.12?

Great advice! I checked torch doc and find nan_to_num function added at torch 1.8, if 1.12 is better one please let me know, Thank you for review!

tjruwase (approved):

loadams (approved):

PR #5878, opened 2024-08-08
fix environment variable export bug for MultiNodeRunner

TideDra:
In some multi-node environment like SLURM，there are some environment vars that contain special chars and can trigger errors when being exported.

For example, there is a var `SLURM_JOB_CPUS_PER_NODE=64(x2)` when requesting two nodes with 64 cpus using SLURM.
Using `runner.add_export` to export this var will add a command `export SLURM_JOB_CPUS_PER_NODE=64(x2)` when launching subprocesses, while this will cause a bash error since `(` is a key word of bash, like:
```
[2024-08-07 16:56:24,651] [INFO] [runner.py:568:main] cmd = pdsh -S -f 1024 -w server22,server27 export PYTHONPATH=/public/home/grzhang/code/CLIP-2;  export SLURM_JOB_CPUS_PER_NODE=64(x2); ...
server22: bash: -c: 行 0: 未预期的符号“(”附近有语法错误
```
This PR simply wrap the environment vars with a pair of `"` to make sure they are treated as string.


deepspeed/launcher/multinode_runner.py
```diff
@@ -34,7 +34,7 @@ def get_cmd(self, environment, active_resources):
         """Return the command to execute on node"""
 
     def add_export(self, key, var):
-        self.exports[key.strip()] = var.strip()
+        self.exports[key.strip()] = f"\"{var.strip()}\""
```
loadams, line 37: @TideDra have you tested this?

I think an issue we have seen previously is if the key is supposed to contain quotes, we run into issues with this solution, right?

TideDra: I test this in my use case and it solves the bug I meet. I think this is a safe operation, since the `var` here is already treated as string, and using quotes just makes sure it is treated as string too when exported. If there was any issue caused by quotes previously, please let me know😀

loadams (approved):

deepspeed/launcher/multinode_runner.py
qiuosier: Hi @TideDra @loadams, actually I run into the an issue after this fix. The value of my environment variable is a valid JSON containing double quotes (for example, `KEY='{"key": "val"}'`). With this update, the export command becomes `export KEY="'{"key": "val"}'"`, which essentially removed double quotes and make the value as invalid JSON `'{key: val}'`.

PR #5881, opened 2024-08-08
add moe topk(k>2) gate support

inkcherry:
Notice some users need to use topk > 2 to train MoE models. For example: https://huggingface.co/Qwen/Qwen1.5-MoE-A2.7B/blob/main/config.json, this PR adds support for topk (k > 2) gates.

- add topk (k>2) support
- add drop token policy based on position and probabilities.
- unit tests


tohtana (approved):

tohtana:
Thank you @inkcherry for the great contribution! I approved and scheduled merging.

PR #6002, opened 2024-08-15
Pin transformers version on nv-nightly

loadams:
nv-nightly was failing due to updates in transformers, we will need to introduce a real fix for these, but this at least gets the test passing and we need to update transformers support for MII too.

### DIFF
--- a/.github/workflows/nv-nightly.yml
+++ b/.github/workflows/nv-nightly.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Install pytorch
         run: |
-          pip install -U --cache-dir $TORCH_CACHE torch==1.13.1 torchvision --index-url https://download.pytorch.org/whl/cu117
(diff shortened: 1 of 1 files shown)


.github/workflows/nv-nightly.yml
```diff
@@ -25,7 +28,7 @@ jobs:
 
       - name: Install pytorch
         run: |
-          pip install -U --cache-dir $TORCH_CACHE torch==1.13.1 torchvision --index-url https://download.pytorch.org/whl/cu117
+          pip install -U --cache-dir $TORCH_CACHE torch torchvision --index-url https://download.pytorch.org/whl/cu118
```
loadams, line 31: Switch to nightly feed

loadams: Nevermind, that's not the original meaning of nightly in this, these are nightly tests, not to keep up with torch nightly releases.

tjruwase (approved):

Issue #5620, opened 2024-06-05
[REQUEST] Moving a trainable model with an optimiser between GPU and CPU


kfertakis:
**Is your feature request related to a problem? Please describe.**
When a deespeed model is initialised with an optimiser, the `torch.nn.module.to()` functionality for moving the model between devices breaks as the optimiser holds references to the model parameters and thus GPU memory is not cleared when trying to move it to CPU for example. 

**Describe the solution you'd like**
Functionality that is similar to `torch.nn.module.to()` for moving both model and optimiser between devices which de-allocates the previously occupied memory.

**Describe alternatives you've considered**
The alternative is to destroy the model instance and recreate it from a checkpoint but this has a much higher time cost.

tohtana:
Hi @kfertakis, 
Is #6011 useful for your purpose? It is not merged yet but we validated that it reduces the memory footprint.

kfertakis:
Hey @tohtana,

Thank you very much for bringing it to my attention. I will be testing it and report back but it does seem promising for my use case. Thanks a lot.

kfertakis:
#6011 addresses this request.

PR #6011, opened 2024-08-16
Add APIs to offload states of model, optimizer, and engine

tohtana:
This PR adds the following APIs to offload model, optimizer, and engine states.

```pytyon
def offload_states(self,
                   include: Container[OffloadStateTypeEnum] = None,
                   device: OffloadDeviceEnum = OffloadDeviceEnum.cpu,
                   pin_memory: bool = True,
                   non_blocking: bool = False) -> None:
    """Move the ZeRO optimizer buffers to the specified device.

    Arguments:
        include: Optional. The set of states to offload. If not provided, all states are offloaded.
        device: Optional. The device to move the ZeRO optimizer buffers to.
        pin_memory: Optional. Whether to pin the memory of the offloaded states.
        non_blocking: Optional. Whether to offload the states asynchronously.
...
def offload_states_back(self, non_blocking: bool = False) -> None:
```

Here is the typical usage.
```python
# Offload after forward, backward, and step
model.offload_states()
# Do something requiring a lot of device memory
...
# Load states back to device memory
model.offload_states_back()
```

You can selectively offload states to balance the offloading overhead and memory saving.
```python
model.offload_states(include=set([OffloadStateTypeEnum.hp_params, OffloadStateTypeEnum.opt_states], device=OffloadDeviceEnum.cpu)
```

Performance (4.3B parameters / 4x A100)
- Environment (4x A100, [benchmark script](https://gist.github.com/tohtana/05d5faba5068cf839abfc7b1e38b85e4))
  - Average Device to Host transfer time: 2.45 GB/s, aggregated: 9.79 GB/s
  - Average Host to Device transfer: 11.05 GB/s, aggregated: 44.19 GB/s
- Mem (allocated by PyTorch)
  - Before offload 18.2GB
  - After offloading 17.7MB
- Time ([benchmark script](https://github.com/microsoft/DeepSpeedExamples/tree/tohtana/offload_states/training/offload_states), offloading time/loading time)

python output_table.py 
|   |pin_memory=0 non_blocking=0|pin_memory=0 non_blocking=1|pin_memory=1 non_blocking=0|pin_memory=1 non_blocking=1|
|--:|---------------------------|---------------------------|---------------------------|---------------------------|
|  1|4.34 / 3.42                |4.99 / 2.37                |6.5 / 2.42                 |6.0 / 2.39                 |
|  2|9.9 / 3.28                 |5.1 / 2.34                 |6.21 / 2.42                |6.25 / 2.45                |
|  3|9.92 / 3.19                |6.71 / 2.35                |6.33 / 2.38                |5.93 / 2.42                |
|  4|9.55 / 2.82                |7.11 / 2.39                |6.9 / 2.38                 |6.5 / 2.43                 |
|  5|4.4 / 3.35                 |6.04 / 2.41                |6.26 / 2.41                |6.32 / 2.47                |
|  6|4.4 / 3.57                 |6.58 / 2.42                |6.88 / 2.4                 |6.35 / 2.43                |
|  7|9.51 / 3.12                |6.9 / 2.39                 |6.9 / 2.39                 |6.46 / 2.4                 |
|  8|4.77 / 3.64                |6.69 / 2.39                |7.39 / 2.42                |6.56 / 2.46                |
|  9|9.5 / 3.07                 |7.18 / 2.42                |6.67 / 2.39                |7.38 / 2.46                |

TODO:
- Enable offloading to a NVMe storage -> NVMe support is non-trivial. I suggest adding the support in another PR
- [DONE] Discard buffer (and recreate it) instead of offloading. We don't need to restore the contiguous buffer for reduce.
- [DONE] Check pin_memory improves performance or not


deepspeed/runtime/engine.py
```diff
@@ -3681,3 +3681,40 @@ def compile(self, backend=get_accelerator().get_compile_backend(), compile_kwarg
...
+        if device == OffloadDeviceEnum.none:
+            logger.warning("No device specified for offloading states.")
+            return
+
+        if device == OffloadDeviceEnum.nvme:
+            raise ValueError("NVMe offload is not supported for offloading states.")
+
+        self.optimizer.offload_states(include=include, device=device, pin_memory=pin_memory, non_blocking=non_blocking)
+
+    def offload_states_back(self, non_blocking: bool = False) -> None:
```
tjruwase, line 3712: Let's improve name?

```suggestion
    def restore_offloaded_states(self, non_blocking: bool = False) -> None:
```

tjruwase, line 3712: Another naming option

```suggestion
    def reload_states(self, non_blocking: bool = False) -> None:
```

tohtana: Let's do this `reload_states`.

tohtana: Changed to `reload_states` as suggested in another comment.

tohtana: Fixed as suggested

deepspeed/runtime/zero/offload_config.py
```diff
@@ -98,3 +98,12 @@ def set_pipeline(self):
         pipeline = self.pipeline_read or self.pipeline_write
         self.__dict__["pipeline"] = pipeline
         return self
+
+
+class OffloadStateTypeEnum(str, Enum):
+    """ Enum for internal buffer types """
+    opt_states = "opt_states"
```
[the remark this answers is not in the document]
tohtana: Fixed

deepspeed/runtime/zero/stage3.py
```diff
@@ -2782,6 +2790,151 @@ def checkpoint_event_epilogue(self):
...
+                for src, dest in zip(self.hp_params_pin_buffers, self.fp32_partitioned_groups_flat):
+                    dest.data = src.to(device, non_blocking=non_blocking)
+            else:
+                for buf in self.fp32_partitioned_groups_flat:
+                    buf.data = buf.data.to(device, non_blocking=non_blocking)
+            self.offloaded_states.remove(OffloadStateTypeEnum.hp_params)
+
+        # LP Param
+        if OffloadStateTypeEnum.lp_params in self.offloaded_states:
+            self.lp_param_buffer.data = self.lp_param_buffer.data.to(device, non_blocking=non_blocking)
```
tjruwase, line 2898: What about case of `self.lp_param_contiguous_pin_buffer`?

deepspeed/runtime/zero/stage3.py
```diff
@@ -2782,6 +2790,151 @@ def checkpoint_event_epilogue(self):
...
+            else:
+                for buf in self.fp32_partitioned_groups_flat:
+                    buf.data = buf.data.to(device, non_blocking=non_blocking)
+            self.offloaded_states.add(OffloadStateTypeEnum.hp_params)
+
+        # LP param
+        if needs_offload(OffloadStateTypeEnum.lp_params):
+            if pin_memory:
+                if not hasattr(self, "lp_param_contiguous_pin_buffer"):
+                    self.lp_param_contiguous_pin_buffer = get_accelerator().pin_memory(
```
tjruwase, line 2829: I don't understand the need for two pinned buffers `self.lp_param_contiguous_pin_buffer` and `self.lp_params_pin_buffers`. Let's chat about this.

deepspeed/runtime/zero/stage3.py
```diff
@@ -2782,6 +2790,151 @@ def checkpoint_event_epilogue(self):
...
+
+        def needs_offload(target):
+            # return True
+            return target not in self.offloaded_states and (include == None or target in include)
+
+        # HP param
+        if needs_offload(OffloadStateTypeEnum.hp_params):
+            if pin_memory:
+                if not hasattr(self, "hp_params_pin_buffers"):
+                    self.hp_params_pin_buffers = [
```
tjruwase, line 2812: I think it would be faster and reduce internal fragmentation to pin one large buffer instead of many smaller buffers. We could then use `narrow` to partition the large buffer.

tohtana: For `hp_params`, I thought the above code already handled large buffers. Is this correct?

deepspeed/runtime/utils.py
```diff
@@ -1065,3 +1065,38 @@ def to_tensor(v):
...
 
     return total_norm
+
+
+def offload_adam_states(optimizer, device, pin_memory: bool = False, non_blocking: bool = False):
+    """Move optimizer states to device. Note that this assumes the state structure of DeepSpeed Adam."""
+
+    def move_key(state, key):
+        if pin_memory:
+            pin_mem_key = f"{key}_pin_memory"
```
tjruwase, line 1075: Since we never free the offload buffer, I wonder if we need this much difference between pin and non-pin options. For example, would something like this work
```python
  offload_buffer = torch.empty_like(state[key], device=device)
  if pin_memory:
      state['offload_buffer'] = get_accelerator().pin_memory(offload_buffer)
  else:
    state['offload_buffer'] = offload_buffer
 state['offload_buffer'].copy_(state[key,],  ...)
 state[key].data = state['offload_buffer']
```

tests/unit/runtime/zero/test_offload_states.py
```diff
@@ -0,0 +1,108 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
```
tjruwase, line 3: Need unit test to verify that 
1. tensor.device == cpu after `offload_states`
2. tensor.device == accelerator after `reload_states`

deepspeed/runtime/zero/stage3.py
tohtana: Great catch, changed it to keep only one of the two.

deepspeed/runtime/zero/stage3.py
tohtana: Fixed

tests/unit/runtime/zero/test_offload_states.py
tohtana: Added validation

deepspeed/runtime/utils.py
tohtana: Simplified the logic as suggested

deepspeed/runtime/zero/stage3.py
tjruwase: You are correct. I forgot that logic.

tjruwase (approved):

tjruwase:
@tohtana, need to add offload APIs to readthedocs, similar to 
https://deepspeed.readthedocs.io/en/latest/zero3.html#gpu-memory-management

tohtana:
@tjruwase Added the document.

kfertakis:
Hi,

Thank you for your work. I've been trying the new APIs to test model offloading in a multi-model deployment (e.g., deepspeed-chat) as part of #5620 . Although the API works in offloading a model and reducing GPU memory, after bringing the model back and completing the first training iteration (i.e., optimiser states have been updated), I get a `RuntimeError: param {} still in flight` exception when trying to offload the model again. I thus wanted to ask whether you think this has something to do with a misuse of the API or if you could provide some further context. I'm providing the relevant stack trace below: Thank you again. 

```
[rank0]: Traceback (most recent call last):
[rank0]:   File "training_script.py", line 173, in gen_function
[rank0]:     self.model_engine.offload_states()
[rank0]:   File "/home/user/DeepSpeed/deepspeed/runtime/engine.py", line 3710, in offload_states
[rank0]:     self.optimizer.offload_states(include=include, device=device, pin_memory=pin_memory, non_blocking=non_blocking)
[rank0]:   File "/home/user/DeepSpeed/deepspeed/runtime/zero/stage3.py", line 2794, in offload_states
[rank0]:     self.empty_partition_cache()
[rank0]:   File "/home/user/DeepSpeed/deepspeed/runtime/zero/stage3.py", line 2785, in empty_partition_cache
[rank0]:     self.parameter_offload.empty_partition_cache()
[rank0]:   File "/home/user/DeepSpeed/deepspeed/runtime/zero/parameter_offload.py", line 181, in empty_partition_cache
[rank0]:     self.partition_all_parameters()
[rank0]:   File "/home/user/DeepSpeed/deepspeed/utils/nvtx.py", line 18, in wrapped_fn
[rank0]:     ret_val = func(*args, **kwargs)
[rank0]:   File "/home/user/DeepSpeed/deepspeed/runtime/zero/parameter_offload.py", line 159, in partition_all_parameters
[rank0]:     self.get_param_coordinator(training=self.module.training).release_and_reset_all(self.module)
[rank0]:   File "/home/user/DeepSpeed/deepspeed/utils/nvtx.py", line 18, in wrapped_fn
[rank0]:     ret_val = func(*args, **kwargs)
[rank0]:   File "/home/user/venv/lib/python3.9/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
[rank0]:     return func(*args, **kwargs)
[rank0]:   File "/home/user/DeepSpeed/deepspeed/runtime/zero/partitioned_param_coordinator.py", line 412, in release_and_reset_all
[rank0]:     raise RuntimeError(f"param {param.ds_summary()} still in flight")
[rank0]: RuntimeError: param {'id': 1, 'status': 'INFLIGHT', 'numel': 4198400, 'ds_numel': 4198400, 'shape': (2050, 2048), 'ds_shape': (2050, 2048), 'requires_grad': True, 'grad_shape': None, 'persist': False, 'active_sub_modules': set(), 'ds_tensor.shape': torch.Size([4198400])} still in flight

```

tohtana:
Thank you for reporting, @kfertakis!

I have an [example script](https://github.com/microsoft/DeepSpeedExamples/tree/tohtana/offload_states/training/offload_states) showing the usage of the APIs. Can you try this?
I suspect that ZeRO3 fails to clean the partitioning status for some models. I would like to clarify that your issue is model specific or not.

kfertakis:
So I tested the issue again with various models and it seems the problem is model-size related as it does not seem to occur for smaller models (i.e., <= 1B params, e.g., gpt2, gpt2-medium) and it does for bigger ones(i.e., OPT-1.3B, mistral-7B). Is there anything I could do to investigate it further and debug it?  By the way, I should mention that I'm testing this in a single node, single GPU configuration (i.e., single worker) thus ZeRO3 partitioning should not have to partition data across other workers. I will also test the benchmark you referenced with an artificially larger model size setting.

Thanks again.

tohtana:
Hi @kfertakis, I tried this [example](https://github.com/microsoft/DeepSpeedExamples/tree/tohtana/offload_states/training/offload_states) with a 4B model but it worked. Can you try this on your environment?
It would be also great if you could offer us a simple repo.

tjruwase:
> `in flight` exception when trying to offload the model again. I thus wanted to ask whether you think this has something to do with a misuse of the API from my end or if you could provide some further context. I'm providing the relevant stack trace below: Thank you again.

@tohtana, I wonder if it is useful to expose `validate_device()` functionality as a deepspeed utility, so that clients can check/confirm the offload status at arbitrary points in their code? 
https://github.com/microsoft/DeepSpeed/blob/8f81634e45a4f28dfa1abe6ee9c6eec5ac184618/tests/unit/runtime/zero/test_offload_states.py#L20

Similar to how `see_memory_usage` enables inspection of HBM/DRAM usage at any point, we could provide mechanisms for offload status. Perhaps we need something like `see_offload_status` that displays the mapping of params, grads, and optimizer to {HBM, DRAM, NVMe}. 

@kfertakis, would love to get your thoughts as well on whether any of the above would be useful? Thanks!

kfertakis:
Hey, thanks for the comments.

@tohtana, I've tried the example you provided and it does seem to work so I'm sharing a [fork](https://github.com/kfertakis/DeepSpeedExamples/tree/kfertakis/test_offload_states) of the DeepSpeed-Examples repo to showcase the problem. I've modified the DeepSpeed-Chat code to use `offload_states`. After you prepare an environment with the right deepspeed version for the new API and also install DeepSpeed-Chat, you can then run the following:

```
deepspeed --num_gpus=1 ./applications/DeepSpeed-Chat/training/step3_rlhf_finetuning/main.py --actor_model_name_or_path facebook/opt-1.3b --critic_model_name_or_path facebook/opt-1.3b --actor_zero_stage 3 --critic_zero_stage 3 --num_padding_at_beginning 1 --data_path Dahoas/rm-static --per_device_generation_batch_size 2 --per_device_training_batch_size 2 --generation_batches 1 --ppo_epochs 1 --max_answer_seq_len 512 --max_prompt_seq_len 512 --gradient_accumulation_steps 1 --actor_dropout 0.0 --deepspeed --dtype bf16 --enable_hybrid_engine --offload_test
```
this should lead to the `RuntimeError: param {} still in flight` that I mentioned. Any thoughts on this would be much appreciated. 

@tjruwase thanks for the reference. Current problem aside, I can see how the helper functions can be useful in the future for ensuring consistency. thanks.

tohtana:
Hi @kfertakis, thank you for sharing the repro. It seems that the actual issue is related to ZeRO3's prefetching.

I opened #6557 as a workaround to address this issue. Can you try the branch `tohtana/clean_up_prefetch_param`? It also includes the offloading APIs. You can just switch to it.

kfertakis:
Hi @tohtana,

thank you for your work. I tried your branch and the issue seems to be fixed. I will continue testing and raise any new issues but for now, the `offload_states` API seems to be working as expected. Many thanks.

kfertakis:
I also wanted to ask if the offloading functionality could be extended to support `DeepSpeedCPUAdam` optimiser, besides `FusedAdam`, in the future for offloading a model with an optimizer which is already offloaded to the CPU? Thank you

tohtana:
> I wonder if it is useful to expose validate_device() functionality as a deepspeed utility, so that clients can check/confirm the offload status at arbitrary points in their code?

@tjruwase Let me address this by another PR after this one is merged.

tohtana:
Thank you @kfertakis for validating the fix. 

> I also wanted to ask if the offloading functionality could be extended to support DeepSpeedCPUAdam optimiser, besides FusedAdam, in the future for offloading a model with an optimizer which is already offloaded to the CPU? Thank you

Let me consider how to do this. Please feel free to open a new issue to track it as I am going to merge this PR first.

kfertakis:
Thank you @tohtana for completing and merging the feature. I've opened two additional requests #6595 , #6596 to track the relevant extensions we discussed above. thanks.

PR #6360, opened 2024-08-19
Dtype support check for accelerator in UTs

raza-sikander:
Check if the dtype is supported by the accelarator if not then skip


tests/unit/ops/transformer/inference/inference_test_utils.py
```diff
@@ -26,8 +26,10 @@ def get_tolerances():
 def get_dtypes():
     global DTYPES
     if DTYPES is None:
-        DTYPES = [torch.float16, torch.float32]
+        DTYPES = [torch.float32]
         try:
+            if get_accelerator().is_fp16_supported():
```
tjruwase, line 31:
```suggestion
            if get_accelerator().is_fp16_supported():
```

tests/unit/ops/transformer/inference/inference_test_utils.py
```diff
@@ -26,12 +26,7 @@ def get_tolerances():
 def get_dtypes():
     global DTYPES
     if DTYPES is None:
-        DTYPES = [torch.float16, torch.float32]
-        try:
-            if get_accelerator().is_bf16_supported():
-                DTYPES.append(torch.bfloat16)
-        except (AssertionError, AttributeError):
-            pass
+        DTYPES = get_accelerator().supported_dtypes()
```
tjruwase, line 29: Thanks for making this change. I checked and it seems that `torch.float32` may not be included in the `.supported_dtypes()` of the accelerators. Can you please double check that?

tjruwase: Thanks for checking.

tjruwase (approved):

PR #6387, opened 2024-08-20
Unpin tests that previously used a pinned version of transformers

loadams:


.github/workflows/nv-transformers-v100.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           git clone https://github.com/huggingface/transformers
           cd transformers
           # if needed switch to the last known good SHA until transformers@master is fixed
-          git checkout e7e9261a2
+          #git checkout e7e9261a2
```
loadams, line 41: This test is currently disabled and will need to be re-enabled in order to even run

PR #6390, opened 2024-08-21
Store/Load CIFAR from local/offline

raza-sikander:
CIFAR10_DATASET_PATH -> Path where the dataset is stored
STORE_CIFAR10        -> Store the dataset 1/0
CIFAR10_OFFLINE      -> To use offline dataset 1/0
MISC:
Added getDeviceId to get device if by name in case of accelerator


tests/unit/inference/test_inference.py
```diff
@@ -298,6 +298,14 @@ def verify_injection(module):
     verify_injection(model)
 
 
+# Used to Get Device name
+def getDeviceId(local_rank):
+    device = local_rank
+    if get_accelerator().device_name() != 'cuda':
```
tjruwase, line 304: I think either `get_accelerator().device(locak_rank)` or `get_accelerator().device_name(local_rank)` should used here for all accelerators including `cuda`. Did you run into problems with those?

tests/unit/alexnet_model.py
```diff
@@ -102,11 +102,24 @@ def cifar_trainset(fp16=False):
     if local_rank != 0:
         dist.barrier()
 
-    data_root = os.getenv("TEST_DATA_DIR", "/tmp/")
-    trainset = torchvision.datasets.CIFAR10(root=os.path.join(data_root, "cifar10-data"),
-                                            train=True,
-                                            download=True,
-                                            transform=transform)
+    if os.getenv("CIFAR10_OFFLINE", default=None):
```
tjruwase, line 105: I am unfamiliar with these CIFAR10 env vars. Are they standard, and can you please provide doc links? Thanks!

tests/unit/inference/test_inference.py
raza-sikander: Updated as per comment. We faced issue when we used local rank for accelerator so we modified device as per name but kept the cuda to be same as local_rank. 
Raised new PR https://github.com/microsoft/DeepSpeed/pull/6397 for the same.

tests/unit/alexnet_model.py
raza-sikander: These are custom env vars. I have updated the logic in latest commit.

tests/unit/inference/test_inference.py
tjruwase: Thanks for applying the proper fix. I approved #6397.

tests/unit/alexnet_model.py
raza-sikander: We were facing issue to download CIFAR Dataset everytime so  we had stored it in local and were using the same for runs.

tests/unit/alexnet_model.py
```diff
@@ -101,11 +101,11 @@ def cifar_trainset(fp16=False):
     dist.barrier()
     if local_rank != 0:
         dist.barrier()
-
     data_root = os.getenv("TEST_DATA_DIR", "/tmp/")
-    trainset = torchvision.datasets.CIFAR10(root=os.path.join(data_root, "cifar10-data"),
+    download = False if os.getenv("CIFAR10_DATASET_PATH") else True
```
tjruwase, line 105: I recommend simplification like the following to improve readability. I have approved regardless. 
```suggestion
    if os.getenv("CIFAR10_DATASET_PATH"):
       data_root = os.getenv("CIFAR10_DATASET_PATH")
       download = False
    else:
       data_root = os.path.join(os.getenv("TEST_DATA_DIR", "/tmp"), "cifar10-data")
       download = True
```

tjruwase (approved):

loadams (approved):

tests/unit/alexnet_model.py
loadams: @raza-sikander - does this resolve your issues, could you confirm before we commit these changes or just merge the branch as is?

raza-sikander: Updating the commit as per suggestion for simplification

loadams (approved):

PR #6443, opened 2024-08-26
[CCL] fix condition issue in ccl.py

YizhouZ:
previous condition check is not right, it would cause this condition always be True.


loadams (approved):

loadams (approved):

loadams:
I wanted to check this on XPU and HPU, 

XPU: https://github.com/microsoft/DeepSpeed/actions/runs/10586606972/job/29335722768
HPU seems to have another issue, debugging that now.

PR #6449, opened 2024-08-27
DeepNVMe tutorial

tjruwase:


docs/_tutorials/deepnvme.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+## Putting it together
+We hope that the above material is sufficient to get you started with DeepNVMe. You can visit [here](https://github.com/microsoft/DeepSpeedExamples/tree/master/deepnvme) for end-to-end examples, including [simple file operations](https://github.com/microsoft/DeepSpeedExamples/blob/master/deepnvme/file_access/README.md).
+
+
+## Appendix
+
+### Pinned Tensors
+A key part of DeepNVMe optimization is the use of direct memory access (DMA) for I/O operations, which requires that the host or device tensor be pinned. To pin host tensors, you can use mechanisms provided by [Pytorch](https://pytorch.org/docs/stable/generated/torch.Tensor.pin_memory.html) or [DeepSpeed Accelerators](/tutorials/accelerator-abstraction-interface/#tensor-operations). On the other hand,`gds_handle` provides `new_pinned_device_tensor()` and `pin_device_tensor()` functions for pinning CUDA tensors.
```
jomayeri, line 118: Can I add an example here showing how to pin a tensor?

tjruwase: > Can I add an example here showing how to pin a tensor?

Sure, go for it.

jomayeri (approved):

docs/_tutorials/deepnvme.md
```diff
@@ -0,0 +1,201 @@
...
+>>> from deepspeed.ops.op_builder import AsyncIOBuilder
+>>> h = AsyncIOBuilder().load().aio_handle(num_threads=4)
+>>> r=h.async_pwrite(t,'/local_nvme/test_1GB.pt')
+>>> h.wait()
+1
+>>> os.path.isfile('/local_nvme/test_1GB.pt')
+True
+>>> os.path.getsize('/local_nvme/test_1GB.pt')
+1073741824
+```
```
GuanhuaWang, line 114: this part seems duplicated to me, the only change is num_thread=4 comparing with async write above. maybe we can just re-use async write code sample (e.g., async write we have num_thread =1, then in parallel write section we mention parallel write just change num_thread = N depending on how parallel user want), in parallel write section we only highlight the diff code line as 
`h = AsyncIOBuilder().load().aio_handle(num_threads=4)`

maybe also discuss a bit pros. and cons. of parallel write (e.g., file chunk per thread should not be too small, etc.)

docs/_tutorials/deepnvme.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Putting it together
+We hope that the above material is sufficient to get you started with DeepNVMe. You can visit [here](https://github.com/microsoft/DeepSpeedExamples/tree/master/deepnvme) for end-to-end examples, including [simple file operations](https://github.com/microsoft/DeepSpeedExamples/blob/master/deepnvme/file_access/README.md).
+
+
+## Appendix
+
+### Pinned Tensors
```
GuanhuaWang, line 122: I think pinned tensor maybe necessary for good performance with gds? shall we put it here or as "prerequisite" at the beginning of this blog.

GuanhuaWang (commented):
just added a few comments.

docs/_tutorials/deepnvme.md
tjruwase: Thanks for the feedback, @GuanhuaWang. The duplication is deliberate so that each snippet is self-contained.

docs/_tutorials/deepnvme.md
tjruwase: This is a good point. I am also conflicted about whether putting pinning in main section hurts the simplicity of the tutorial or not.

docs/_tutorials/deepnvme.md
```diff
@@ -0,0 +1,201 @@
...
+>>> h.wait()
+1
+>>> os.path.isfile('/local_nvme/test_1GB.pt')
+True
+>>> os.path.getsize('/local_nvme/test_1GB.pt')
+1073741824
+```
+
+### Parallel File Write
+An important DeepNVMe optimization is the ability to parallelize individual I/O operations. This optimization is enabled by specifying the desired parallelism degree when constructing a DeepNVMe handle. Subsequent I/O operations with that handle are automatically parallelized over the requested number of host or device threads, as appropriate. I/O parallelism is composable with either the blocking or non-blocking I/O APIs. The example below illustrates 4-way parallelism of a file write using `async_pwrite`. Note the use of `num_threads` argument to specify the desired parallelism degree in handle creation.
```
tohtana, line 97: In this example, we write a single file. Does multi-threading still gain the performance? Or it works only for multiple `async_pwrite` calls?

tjruwase: Yes, this optimization is for a single I/O operation from one client thread.

PR #6450, opened 2024-08-27
Add workflow to build DS without torch to better test before releases

loadams:
- Adds a nightly workflow that tests to confirm we can build DeepSpeed without torch as a dependency, as this often only surfaces when doing a release.


sfc-gh-mwyatt (commented):
LGTM

jomayeri (approved):

PR #6458, opened 2024-08-28
Fix transformers/torch errors on nn.RMSNorm by pinning transformers.

loadams:
HPU pipeline was failing due to updated transformers package that seems to have a dependency on torch 2.4: https://github.com/microsoft/DeepSpeed/actions/runs/10586654075/job/29335873815

Opened an issue with transformers to investigate: https://github.com/huggingface/transformers/issues/33176

CC: @nelyahu - we can also fix this when/if there is a package with torch 2.4

### DIFF
--- a/.github/workflows/hpu-gaudi2.yml
+++ b/.github/workflows/hpu-gaudi2.yml
@@ -111,7 +111,7 @@ jobs:
         run: |
           git clone https://github.com/huggingface/transformers
           cd transformers
-          git checkout v4.44.1 # Pin to specific version before we can update torch version to 2.4+ for nn.RMSNorm
+          git checkout v4.44.2 # Pin to specific version before we can update torch version to 2.4+ for nn.RMSNorm
           git rev-parse --short HEAD
           pip install .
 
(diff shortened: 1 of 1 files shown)


loadams:
Should be fixed in latest PR to HF Transformers - won't need to complete this PR but leaving until transformers PR is merged.

nelyahu:
@loadams on 1.18.0 we will upgrade to pt2.4.0.
CC: @vshekhawat-hlab

loadams:
> @loadams on 1.18.0 we will upgrade to pt2.4.0. CC: @vshekhawat-hlab

Thanks @nelyahu - turns out this is a bug in transformers and we will need that fixed regardless, but good to know, thanks!

loadams:
PR in transformers is here: https://github.com/huggingface/transformers/pull/33177

nelyahu:
@loadams it seems that the Gaudi2 pipeline is still broken.
Are you planning to wait till PR for transformers will be merged and included in a version? 
i think it might take a while. no?

loadams:
The PR has merged into transformers (took longer than anticipated) but it looks to be running again, so no need to pin anything.

PR #6487, opened 2024-09-04
print warning if actual triton cache dir is on NFS, not just for default

### LINKED ISSUE #6486: [BUG] deepspeed prints a warning when the user's home directory is on NFS even when TRITON_CACHE_DIR is set to a non-NFS directory
When the user's home directory is on NFS, deepspeed prints a warning such as this one:
```
Warning: The default cache directory for DeepSpeed Triton autotune, /home/ubuntu/.triton/autotune, appears to be on an NFS system. While this is generally acceptable, if you experience slowdowns or hanging when DeepSpeed exits, it is recommended to set the TRITON_CACHE_DIR environment variable to a non-NFS path.
```

However, deepspeed prints this warning even when `TRITON_CACHE_DIR` is set.  Although technically the warning is correct as it says "The default cache directory" it is also very misleading as it is irrelevant when `TRITON_CACHE_DIR` is set to a non-NFS directory. Furthermore, the warning is _not_ be printed when the home directory is not on NFS but `TRITON_CACHE_DIR` is explicitly set to an NFS directory. 

I suggest refactoring the logic that checks for the NFS directory and prints the warning to do so after the actual cache dir lookup is performed. 

I.e. the code in https://github.com/microsoft/DeepSpeed/blob/359c12e62973a35f2d04af8c04fb76677a588a8b/deepspeed/ops/transformer/inference/triton/matmul_ext.py#L41-L44
```
    if is_nfs_path(tmp_path):
        print(
            f"Warning: The default cache directory for DeepSpeed Triton autotune, {tmp_path}, appears to be on an NFS system. While this is generally acceptable, if you experience slowdowns or hanging when DeepSpeed exits, it is recommended to set the TRITON_CACHE_DIR environment variable to a non-NFS path."
        )
```

Should be moved to operate on `self.cache_dir` (at https://github.com/microsoft/DeepSpeed/blob/359c12e62973a35f2d04af8c04fb76677a588a8b/deepspeed/ops/transformer/inference/triton/matmul_ext.py#L78) and the message can just be changed to reflect that it refers to the active cache directory rather than the default. Something like:
```
    if is_nfs_path(self.cache_dir):
        print(
            f"Warning: The cache directory for DeepSpeed Triton autotune, {self.cache_dir}, appears to be on an NFS system. While this is generally acceptable, if you experience slowdowns or hanging when DeepSpeed exits, it is recommended to set the TRITON_CACHE_DIR environment variable to a non-NFS path."
        )
```

jrandall:
move the logic that prints a warning when triton cache dir is on NFS to act on the actual calculated cache_dir rather than on the default.

this means that:
 - when the default directory (in the user's home directory) is on NFS but `TRITON_CACHE_DIR` is set to a non-NFS directory, no warning will be printed whereas prior to this change a spurious and confusing warning was printed
  - when the user's home directory is not on NFS but `TRITON_CACHE_DIR` is set to an NFS directory, a warning will be printed whereas prior to this change no warning would be printed
 
fixes #6486

### DIFF
--- a/deepspeed/ops/transformer/inference/triton/matmul_ext.py
+++ b/deepspeed/ops/transformer/inference/triton/matmul_ext.py
@@ -40,13 +40,17 @@ class TritonCacheDir:
     _warning_printed = False
 
     @staticmethod
-    def default_cache_dir():
-        tmp_path = os.path.join(Path.home(), ".triton", "autotune")
-        if is_nfs_path(tmp_path) and not TritonCacheDir._warning_printed:
+    def warn_if_nfs(cache_dir):
+        if is_nfs_path(cache_dir) and not TritonCacheDir._warning_printed:
             print(
-                f"Warning: The default cache directory for DeepSpeed Triton autotune, {tmp_path}, appears to be on an NFS system. While this is generally acceptable, if you experience slowdowns or hanging when DeepSpeed exits, it is recommended to set the TRITON_CACHE_DIR environment variable to a non-NFS path."
+                f"Warning: The cache directory for DeepSpeed Triton autotune, {cache_dir}, appears to be on an NFS system. While this is generally acceptable, if you experience slowdowns or hanging when DeepSpeed exits, it is recommended to set the TRITON_CACHE_DIR environment variable to a non-NFS path."
             )
             TritonCacheDir._warning_printed = True
+        return
+
+    @staticmethod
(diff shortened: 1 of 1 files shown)


tohtana:
@jrandall Thank you for the contribution!

tohtana (approved):

loadams (approved):

PR #6488, opened 2024-09-04
Handle an edge case where `CUDA_HOME` is not defined on ROCm systems

amorehead:
* Handles an edge case when building `gds` where `CUDA_HOME` is not defined on ROCm systems

### DIFF
--- a/op_builder/gds.py
+++ b/op_builder/gds.py
@@ -49,6 +49,11 @@ class GDSBuilder(AsyncIOBuilder):
             return False
 
         CUDA_HOME = torch.utils.cpp_extension.CUDA_HOME
+        if CUDA_HOME is None:
+            if verbose:
+                self.warning("Please install torch CUDA if trying to pre-compile GDS with CUDA")
+            return False
+
         CUDA_LIB64 = os.path.join(CUDA_HOME, "lib64")
         gds_compatible = self.has_function(funcname="cuFileDriverOpen",
                                            libraries=("cufile", ),


op_builder/gds.py
```diff
@@ -49,6 +49,11 @@ def is_compatible(self, verbose=False):
             return False
 
         CUDA_HOME = torch.utils.cpp_extension.CUDA_HOME
+        if CUDA_HOME is None:
+            if verbose:
+                self.warning("Please install torch CUDA if trying to pre-compile GDS with CUDA")
```
loadams, line 54: Shouldn't this be covered on ROCm systems in line 39 above from [this PR](https://github.com/microsoft/DeepSpeed/pull/6456/files)?

amorehead: Oddly enough, on my system, in the `builder.py` script `self.is_rocm_pytorch()` evalutes to `True`, while here it evaluates to `False`. In this case, the code continues into the CUDA-specific logic, where a crash then follows.

jomayeri: Why is `self.is_rocm_pytorch` getting different values in different parts of the code?

amorehead: I sadly haven't had a chance to isolate why. I just know that without this patch, I have to set `CUDA_HOME` to a bogus value in my Dockerfile (which works around this issue but then causes another issue downstream since I'm using AMD ROCm cards).

jomayeri: This is a good secondary check regardless. To be clear even with this change you see an issue building CPU Adam?

amorehead: Nope, with this change, I no longer have issues building CPU Adam. That's because I no longer have to set a bogus `CUDA_HOME` environment variable in my (ROCm-targeting) Dockerfile for model training.

jomayeri (approved):

PR #6489, opened 2024-09-04
DS_BUILD_OPS should build only compatible ops

tjruwase:
Currently DS_BUILD_OPS=1 fails on incompatible ops. This is a deviation from [documentation](https://www.deepspeed.ai/tutorials/advanced-install/#pre-install-deepspeed-ops) which states that only compatible ops are built. 

image


loadams (approved):

loadams:
We will definitely want to call this out in the release notes, since I think we've technically not been matching this documentation for a while.  But this is a better way to handle this.

PR #6490, opened 2024-09-04
Safe usage of popen

tjruwase:
Avoid shell=True security issues with Popen

### DIFF
--- a/deepspeed/utils/numa.py
+++ b/deepspeed/utils/numa.py
@@ -49,8 +49,8 @@ def check_for_numactl_pkg():
         flag, lib, tool = data
         path = distutils.spawn.find_executable(pkgmgr)
         if path is not None:
-            cmd = f"{pkgmgr} {flag} {lib}"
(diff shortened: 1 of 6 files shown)


tohtana (commented):
Great fixes @tjruwase,
I think `subprocess.run` and `subprocess.check_output` have the same security risk. Shall we fix all? We can grep all with `shell=True`.

loadams (approved):

tjruwase:
> I think `subprocess.run` and `subprocess.check_output` have the same security risk. Shall we fix all? We can grep all with `shell=True`.

@tohtana. good point. I will address others in separate PR since this one needs to be released immediately.

@loadams, can we please create a release once this is merged? Thanks!

loadams:
> > I think `subprocess.run` and `subprocess.check_output` have the same security risk. Shall we fix all? We can grep all with `shell=True`.
> 
> @tohtana. good point. I will address others in separate PR since this one needs to be released immediately.
> 
> @loadams, can we please create a release once this is merged? Thanks!

@tjruwase - yes, I'll ensure this PR is merged and we will do a release shortly after.

PR #6491, opened 2024-09-04
Remove all other instances of shell=True

loadams:
Todo:
- [ ] Test changes to Windows builds.
- [ ] Test changes to ds_bench.
- [ ] Ensure other tests still function properly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -204,9 +204,9 @@ git_hash_cmd = "git rev-parse --short HEAD"
 git_branch_cmd = "git rev-parse --abbrev-ref HEAD"
 if command_exists('git') and not is_env_set('DS_BUILD_STRING'):
     try:
-        result = subprocess.check_output(git_hash_cmd)
+        result = subprocess.check_output(git_hash_cmd, shell=True)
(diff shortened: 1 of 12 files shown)


setup.py
```diff
@@ -204,9 +204,9 @@ def op_enabled(op_name):
 git_branch_cmd = "git rev-parse --abbrev-ref HEAD"
 if command_exists('git') and not is_env_set('DS_BUILD_STRING'):
     try:
-        result = subprocess.check_output(git_hash_cmd, shell=True)
```
loadams, line 207: These instances are required in order to build deepspeed and get the git commit.

tohtana (commented):
I don't think just removing `shell=True` works. We can change the command string to a list for most cases. We also need a bit more for some commands such as ones including pipes.

loadams:
> I don't think just removing `shell=True` works. We can change the command string to a list for most cases. We also need a bit more for some commands such as ones including pipes.

Agreed, @tjruwase will share a more full PR fixing this soon.

PR #6496, opened 2024-09-05
add option to disable logger while compiling to avoid graph breaks

ShellyNR:
adding an option to disable calls for logger while compiling to avoid graph breaks. Here I used an environment variable to determine whether to activate this option, but it can also be determined using the json config file or any other way you see fit.


tohtana:
Hi @ShellyNR, thank you for the great catch!
Can we also show a warning to recommend setting `DISABLE_LOGS_WHILE_COMPILING=1` when the compiler is enabled but the variable is not set?

SNahir:
Hi, I added a warning as you suggested, I used a logger filter for this as it runs every time a log record is created
@tohtana

tohtana (approved):

tohtana:
Hi @ShellyNR, it looks like the PyTorch version we use for tests doesn't have `torch.compiler`. We already have the function to check it. Can you add a condition before calling `torch.compiler.is_compiling()`?

https://github.com/microsoft/DeepSpeed/blob/d45cfd34551537ce6f8317504bd520d7a2a1a588/deepspeed/runtime/compiler.py#L9

@tjruwase @loadams Can we discuss which PyTorch version we should use for testing? We probably want to avoid increasing the test time, though.

SNahir:
Hi @tohtana , I added is_compile_supported() to the condition before accessing torch.compiler, also I changed the wrapper function to call is_compiling() from torch.compiler instead of _dynamo.

PR #6498, opened 2024-09-05
Avoid security issues of subprocess shell

tjruwase:
Avoid security issues of `shell=True` in subprocess


deepspeed/comm/comm.py
```diff
@@ -697,8 +697,9 @@ def mpi_discovery(distributed_port=TORCH_DISTRIBUTED_DEFAULT_PORT, verbose=True)
 
     master_addr = None
     if rank == 0:
-        hostname_cmd = ["hostname -I"]
-        result = subprocess.check_output(hostname_cmd, shell=True)
+        import shlex
+        hostname_cmd = shlex.split("hostname -I")
```
tohtana, line 701: We may not need `shlex.split` when the command has no placeholder for injection. The same goes for some other fixes.

tohtana: For `hostname -I`, using `socket.gethostname` and `socket.gethostbyname_ex` can be safer.

tjruwase: @tohtana, thanks for the feedback. I will remove the `shlex.split()` here. However, I am not getting the same output if I use `socket.gethostbyname_ex(socket.gethostname())`

tjruwase: Actually, I think the `shlex.split()` is still need to pass args as a list since we are moving `shell=True`. Alternatively, I could manually construct the list. However, I think I will keep the `shlex.split()` to future-proof for arg changes.

tohtana: How about this code?  This is a general proposal to make the code robust against malicious modifications of system commands. But I don't think this is crucial because it won't be a typical attack that can harm our users. We can just keep `hostname -I` if this doesn't work.

```
>>> import socket
>>> hostname = socket.gethostname()
>>> ip_addresses = socket.gethostbyname_ex(hostname)[2]
>>> ip_addresses
['172.17.0.2']
```

loadams: This is my vote, since it also helps resolve issues where we've seen `hostname -I` not available on all systems, such as Mac and latest inetutils on ArchLinux

tjruwase: This is discrepancy that I get on my lambda box.  
```bash
>>> subprocess.check_output(["hostname", "-I"]).decode("utf-8").split()[0]
'192.168.1.104'
>>> socket.gethostbyname_ex(socket.gethostname())[2][0]
'127.0.1.1'
```

Maybe it is a problem with my setup. Can you guys test in your environment? I am happy to switch. Thanks.

loadams: I see that as well, it looks like we would need to modify the command to be:

```
>>> socket.gethostbyname_ex(socket.gethostname()+".local")[2][0]
```

Thought it is slower to run on my system with that change.

tohtana: I think that is because `hostname -I` got the info from network interfaces like `eth0` while `socket` APIs relies on name resolving.
It's natural to expect DNS and/or `/etc/hosts` are properly set up on a multi-node run environment, but this definitely has an impact on many use cases.

tohtana: > I see that as well, it looks like we would need to modify the command to be:
> 
> ```
> >>> socket.gethostbyname_ex(socket.gethostname()+".local")[2][0]
> ```
> 
> Thought it is slower to run on my system with that change.

I wonder if `hosts` or DNS in your system (Windows?) has `[HOSTNAME].local` but it doesn't work on my env.

tjruwase: > ```
> >>> socket.gethostbyname_ex(socket.gethostname()+".local")[2][0]
> ```
> 
This works and gives matching results on my Linux lambda box. 

@tohtana, do you mean this does not work for you?

tohtana: It doesn't work on my wsl.

```
$ python -c 'import socket; socket.gethostbyname_ex(socket.gethostname()+".local")[2][0]'
socket.gaierror: [Errno -5] No address associated with hostname
```

loadams: Interesting, I'd say lets leave it as hostname -I for now, and we can make another PR to update where we can more strenuously test Windows and other OSs?

tohtana: I agree to address this in another PR. This one is urgent and focuses on security.
But I'm not sure it is a good idea to get the IP from the first entry from `hostname -I`. It is not simple to control it even for the administrator. It is easier to tell users to configure `/etc/hosts` properly.
After a quick look on the usage, probably it can also be a hostname, not an IP.

tjruwase:
@loadams, do you have any idea how to repro these py UT failures?
https://github.com/microsoft/DeepSpeed/actions/runs/10723587021/job/29737199498?pr=6498

loadams:
> @loadams, do you have any idea how to repro these py UT failures? https://github.com/microsoft/DeepSpeed/actions/runs/10723587021/job/29737199498?pr=6498

Yes, @tjruwase - I believe this should just be taking an environment with torch installed and cloning your branch and running `pip install .`

tjruwase:
> Yes, @tjruwase - I believe this should just be taking an environment with torch installed and cloning your branch and running `pip install .`

Unfortunately, I am unable to repro this failure following those steps. 

> Though these look to be caused by the changes in setup.py not being able to find git to get the commit hash to build the wheel with name 0.15.1+commithash.

Yes, I think that is the problem. This work with `shell=True`, probably due to some env or PATH settings.

loadams:
> > Yes, @tjruwase - I believe this should just be taking an environment with torch installed and cloning your branch and running `pip install .`
> 
> Unfortunately, I am unable to repro this failure following those steps.
> 
> > Though these look to be caused by the changes in setup.py not being able to find git to get the commit hash to build the wheel with name 0.15.1+commithash.
> 
> Yes, I think that is the problem. This work with `shell=True`, probably due to some env or PATH settings.

This does run in a docker container and you can run using the same container if needed to test as well, but looks like it is largely resolved as well

tohtana (approved):

jagadish-amd:
We are hitting few issues on ROCm environment (Ubuntu ROCm Pytorch container) due to this PR.
1.
 [rank7]:     raise child_exception_type(errno_num, err_msg, err_filename)
[rank7]: FileNotFoundError: [Errno 2] No such file or directory: "/opt/rocm/bin/rocminfo | grep -Eo -m1 'Wavefront Size:[[:space:]]+[0-9]+' | grep -Eo '[0-9]+'"
[rank0]: Traceback (most recent call last):

2.
Building extension module transformer_inference...
Using envvar MAX_JOBS (32) as the number of workers...
Using /root/.cache/torch_extensions/py39_cpu as PyTorch extensions root...
Using /root/.cache/torch_extensions/py39_cpu as PyTorch extensions root...
Using /root/.cache/torch_extensions/py39_cpu as PyTorch extensions root...
ninja: error: build.ninja:8: unexpected '='

[rank4]: Traceback (most recent call last):
[rank4]:   File "/opt/conda/envs/py_3.9/lib/python3.9/site-packages/torch/utils/cpp_extension.py", line 2107, in _run_ninja_build
[rank4]:     subprocess.run(
[rank4]:   File "/opt/conda/envs/py_3.9/lib/python3.9/subprocess.py", line 528, in run
[rank4]:     raise CalledProcessError(retcode, process.args,
[rank4]: subprocess.CalledProcessError: Command '['ninja', '-v', '-j', '32']' returned non-zero exit status 1.

Please let me know if we need to create new issue to track it.
@loadams @jithunnair-amd @pruthvistony @jeffdaily

tjruwase:
> We are hitting few issues on ROCm environment (Ubuntu ROCm Pytorch container) due to this PR. 1. [rank7]: raise child_exception_type(errno_num, err_msg, err_filename) [rank7]: FileNotFoundError: [Errno 2] No such file or directory: "/opt/rocm/bin/rocminfo | grep -Eo -m1 'Wavefront Size:[[:space:]]+[0-9]+' | grep -Eo '[0-9]+'" [rank0]: Traceback (most recent call last):
> 
> 
> Building extension module transformer_inference... Using envvar MAX_JOBS (32) as the number of workers... Using /root/.cache/torch_extensions/py39_cpu as PyTorch extensions root... Using /root/.cache/torch_extensions/py39_cpu as PyTorch extensions root... Using /root/.cache/torch_extensions/py39_cpu as PyTorch extensions root... ninja: error: build.ninja:8: unexpected '='
> 
> [rank4]: Traceback (most recent call last): [rank4]: File "/opt/conda/envs/py_3.9/lib/python3.9/site-packages/torch/utils/cpp_extension.py", line 2107, in _run_ninja_build [rank4]: subprocess.run( [rank4]: File "/opt/conda/envs/py_3.9/lib/python3.9/subprocess.py", line 528, in run [rank4]: raise CalledProcessError(retcode, process.args, [rank4]: subprocess.CalledProcessError: Command '['ninja', '-v', '-j', '32']' returned non-zero exit status 1.
> 
> Please let me know if we need to create new issue to track it. @loadams @jithunnair-amd @pruthvistony @jeffdaily

@jagadish-amd, yes please create a new ticket.

PR #6502, opened 2024-09-07
Handle when `backend` is also in compile_kwargs

oraluben:
cc @tohtana

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -3675,7 +3675,11 @@ class DeepSpeedEngine(Module):
         if self.is_compiled:
             return
 
-        self.module.compile(backend=backend, **compile_kwargs)
+        if 'backend' in compile_kwargs:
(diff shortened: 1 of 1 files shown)


tohtana:
Hi @oraluben, Can you elaborate your goal with this PR?

tohtana:
Hi @oraluben,
Thank you for the clarification. It makes sense. At this point, it's not very clear how we benefit from treating backend separately from other compiler options, but we want to keep the option for compatibility.

I agree with the fix, but the method for merging dictionaries is not correct, as the tests have failed. Could you address this?
It would also be good if we could show a warning when backend is specified in both ways and one of them is overwritten.

oraluben:
Is there anything wrong with the CI? My PRs keep timeout in v100.

loadams:
> Is there anything wrong with the CI? My PRs keep timeout in v100.

Yes this looks to be an issue on our side, I will take care of this.

tohtana (approved):

PR #6508, opened 2024-09-09
Revert "BF16 optimizer: Clear lp grads after updating hp grads in hook"

nelyahu:
Reverts microsoft/DeepSpeed#5328


YangQun1:
@nelyahu Thanks. This change looks good to me.

tjruwase (approved):

PR #6528, opened 2024-09-12
add bfloat16 to inference support dtypes

nelyahu:
to allow running inference tasks using bfloat16


tjruwase (approved):

deepspeed/inference/engine.py
```diff
@@ -324,7 +324,7 @@ def _validate_args(self, mpu, replace_with_kernel_inject):
         if self._config.checkpoint is not None and not isinstance(self._config.checkpoint, (str, dict)):
             raise ValueError(f"checkpoint must be None, str or dict, got {type(self._config.checkpoint)}")
 
-        supported_dtypes = [None, torch.half, torch.int8, torch.float]
+        supported_dtypes = [None, torch.half, torch.int8, torch.float, torch.bfloat16]
```
tjruwase, line 327: Do you need to adjust this as well?
https://github.com/microsoft/DeepSpeed/blob/4e74b8d066b27d34a0061d62dede2ce4b4697d1e/deepspeed/inference/engine.py#L76-L77

nelyahu: @tjruwase I can generalize it, so int8 would also be checked, via accelerate.supported_dtype() method.
but i see that no accelerator support int8.

nelyahu: @tjruwase  Done

deepspeed/inference/engine.py
```diff
@@ -73,8 +73,8 @@ def __init__(self, model, config):
         if hasattr(self.module, "config"):
             TransformerPolicy.hf_model_config = self.module.config
 
-        if config.dtype == torch.half and not get_accelerator().is_fp16_supported():
-            raise ValueError("Type fp16 is not supported.")
+        if config.dtype not in get_accelerator().is_fp16_supported():
```
loadams, line 76: @nelyahu - I believe this is failing because it should check 

```
if config.dtype not in get_accelerator().supported_dtypes():
```

loadams: @nelyahu - figured I could just make this change now and wait for you to confirm if this is acceptable.

loadams (approved):

Issue #5818, opened 2024-07-31
[BUG] Launcher does not honor CUDA_VISIBLE_DEVICES


arnold-jr:
**Describe the bug**
The documentation [here](https://www.deepspeed.ai/getting-started/#resource-configuration-single-node) implies that CUDA_VISIBLE_DEVICES is not supported, but the launcher script does attempt to [handle that case](https://github.com/microsoft/DeepSpeed/blob/afe1b9ede1794b0bf221607be8051728cf4afa18/deepspeed/launcher/runner.py#L407).
Given that CUDA_VISIBLE_DEVICES is used so commonly, I think this still qualifies as a bug. 

On a single node, when `CUDA_VISIBLE_DEVICES=0,2` for example, when I launch deepspeed I get the error.
 ```bash
ValueError: No slot '2' specified on host 'localhost'
```

There is a simple, if clunky, workaround.
```bash
INCLUDE_STR="localhost:$CUDA_VISIBLE_DEVICES"
unset CUDA_VISIBLE_DEVICES
deepspeed --include $INCLUDE_STR ...
```
**To Reproduce**
1. Simple inference script to reproduce
```python
import collections
import os

import pytest


class FakeAccelerator:
    def __init__(self, num_devices: int = 2):
        self.num_devices = num_devices

    def device_count(self) -> int:
        return self.num_devices


def test_main_with_cuda_visible_devices(monkeypatch):
    fake_acc = FakeAccelerator(2)
    from deepspeed.launcher import runner

    monkeypatch.setattr(runner, "get_accelerator", lambda: fake_acc)

    cvd = "0,1"
    os.environ["CUDA_VISIBLE_DEVICES"] = cvd

    runner.main()

    cvd = "0,2"
    os.environ["CUDA_VISIBLE_DEVICES"] = cvd
    with pytest.raises(ValueError):
        runner.main()


def test_parse_resource_filter():
    from deepspeed.launcher.runner import parse_resource_filter

    resource_pool = collections.OrderedDict({"localhost": list(range(2))})
    parse_resource_filter(resource_pool, include_str="localhost:0,1", exclude_str="")

    with pytest.raises(ValueError):
        parse_resource_filter(
            resource_pool, include_str="localhost:0,2", exclude_str=""
        )


def test_parse_inclusion_exclusion():
    from deepspeed.launcher.runner import parse_inclusion_exclusion

    resource_pool = collections.OrderedDict({"localhost": 2})
[...]

PR #6530, opened 2024-09-12
Fix device selection using CUDA_VISIBLE_DEVICES

tohtana:
This PR addresses #5818.
Instead of contiguous numbers based on the device count, this PR uses device indices in `--include`.

### DIFF
--- a/deepspeed/launcher/runner.py
+++ b/deepspeed/launcher/runner.py
@@ -21,6 +21,7 @@ from copy import deepcopy
 import signal
 import time
 from typing import Tuple, List, Dict
+from collections import defaultdict
 import shlex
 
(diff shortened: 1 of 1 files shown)


tjruwase (approved):

loadams (approved):

loadams:
We might need to check the docs for what we say about CUDA_VISIBLE_DEVICES on the deepspeed.ai website to see if we need to update that with these changes?

tohtana:
@loadams I believe this PR aligns the launcher's behavior with the description in the document.

rlanday:
The Getting Started guide says:

https://www.deepspeed.ai/getting-started/
> Also note that CUDA_VISIBLE_DEVICES can’t be used with DeepSpeed to control which devices should be used.

Is this correct?

PR #6532, opened 2024-09-13
[XPU] Support DeepNVMe new code structure

Liangliang-Ma:
In DeepNVMe GDS update, many functions are changed into a more abstract way. Also added some files. These change break zero-infinity on XPU. To bring this feature back, we have this PR:
1. modify the aio opbuilder for new files.
2. Add custom cpu_op_desc_t for xpu users. (XPU don't handle buffer aligned here)


csrc/xpu/aio/deepspeed_cpu_op.cpp
```diff
@@ -0,0 +1,60 @@
...
+    // XPU don't handle buffer here. See XPU Accelerator pin_memory.
+    _contiguous_buffer = _cpu_buffer.contiguous();
+}
+
+char* cpu_op_desc_t::data_ptr() const { return (char*)_contiguous_buffer.data_ptr(); }
+
+void cpu_op_desc_t::finish()
+{
+    if (_read_op) {
+        if (_buffer.is_cuda()) { _buffer.copy_(_cpu_buffer.to(torch::kCUDA)); }
```
tjruwase, line 29: I think this logic could be simplified since this extension will only be built on xpu. So there is no need to handle `cuda`, `npu` or any other accelerator? What do you think?

Liangliang-Ma: Yes that's reasonable to me. I modified the code.

tjruwase (approved):

Liangliang-Ma:
@tjruwase  Hi, the nv CI tests seems break for env issues. Could you please help to take a look? thanks!

loadams:
> @tjruwase Hi, the nv CI tests seems break for env issues. Could you please help to take a look? thanks!

@Liangliang-Ma - I've got this fixed now, I'll get this merged.

loadams (approved):

PR #6544, opened 2024-09-16
[INF] Add config var to enable keeping checkpoints on host

oelayan7:
Using keep_module_on_host config var will let us control if the loaded checkpoints to model parameters will be moved to the device or stay on cpu


oelayan7:
Closing because im observing issues with the PR. 
Will reopen when issues are clear

PR #6553, opened 2024-09-18
Enabled configurable auto Tensor Parallelism (TP) for the inference of diverse models

gyou2021:
Auto TP in auto_tp.py handles linear type modules in emerging complex models. 1) The result of some linear modules in a model should operate all reduce operation after running on multiple HPU/GPU cards; The name of those linear modules may be different from those in the method tp_parser(). 2) The weight of some linear modules in a model CANNOT be split to multiple HPU/GPU cards; 3) The weight of some linear modules in a model should NOT be split to multiple HPU/GPU cards to avoid decreasing performance because of afterward all gather operation (gather result from all cards). In case 1) the Linear type should change to LinearAllreduce type in DeepSpeed. In cases 2) and 3) the linear modules should keep Linear type. A configurable auto TP was proposed to handle those cases easily. The method tp_parser() will add the linear modules in case 1) (Here module name list was stored in the environment variable 'DS_ALL_REDUCE_LINEAR_ITEMS') and the method _replace() will add the linear modules in case 2) and 3) (Here module name list was stored in the environment variable 'DS_KEEP_LINEAR_ITEMS').  Those environment variables are configurable. They can be configured directly in environment variables or a configuration file.

Take the Mixtral 8x7B model as an example:
We will add 'w2' to LinearAllreduce, and keep 'gate' as Linear. 'o_proj' is the default deepspeed LinearAllreduce layer.
Add the following into some main code.
import os
os.environ["DS_ALL_REDUCE_LINEAR_ITEMS"] = "{'w2':'mixtral'}"  
os.environ["DS_KEEP_LINEAR_ITEMS"] = "{'gate':'mixtral'}"

Origin Mixtral model:
image

Mixtral model of auto_TP:
image


delock:
@gyou2021 I like the goal to avoid repetition of same logic from L296 to L315, but I also have concern that models enabled by these lines will not be able to run out-of-box with this PR.  This may not be friendly to self-helping users without access to proper BKC documentation to various models.

Could `allReduceLinearItems` have an initial value as a built-in list, then pre-pend with `os.environment` to get runtime configurability?  I think if the model to be enabled by environment is a public model, it should be contributed to the built-in list to provide OOB experience, right?

gyou2021:
@microsoft-github-policy-service agree [company="Intel"]

gyou2021:

deepspeed/module_inject/auto_tp.py
```diff
@@ -279,40 +282,27 @@ def tp_parser(model):
...
+
+        allReduceLinearItems = os.environ['allReduceLinearItems']
+        allReduceLinearItems = ast.literal_eval(allReduceLinearItems)
+
         for module in module_list:
             for key, submodule in module._modules.items():
                 if isinstance(submodule, nn.Linear):
                     layer_list = layer_list + ["." + key]
-                elif isinstance(submodule, nn.LayerNorm) or key in norm_layer_name_list:
+                elif isinstance(submodule, nn.LayerNorm) or key == 'LayerNorm' or key == 'layer_norm':
```
inkcherry, line 293: Can this be replaced equivalently？

deepspeed/module_inject/auto_tp.py
```diff
@@ -279,40 +282,27 @@ def tp_parser(model):
         module_list = AutoTP.get_module_list(model)
         assert AutoTP.supported(model), "AutoTP not supported for model. Please use kernel injection since container policy for model exists." \
         if AutoTP.kernel_supported(module_list) else "AutoTP not supported for model. Please provide policy."
-        norm_layer_name_list = ['LayerNorm', 'layer_norm', 'ln_1', 'ln_2']
-        #ln_1 , ln_2 for Qwen
+
+        allReduceLinearItems = os.environ['allReduceLinearItems']
```
delock, line 286: Could the environment keep DeepSpeed as `DS_ALL_REDUCE_LINEAR_ITEMS`?  This will make the naming more consistent with existing DeepSpeed knobs.

gyou2021:

loadams:
Hi @delock  and @gyou2021 - what more needs to be done to complete this PR?  Just a review/approval?  Any other changes?

delock:
@loadams let me check with gyou on this PR status.

deepspeed/module_inject/auto_tp.py
gyou2021: Sure. It has been modified.

deepspeed/module_inject/auto_tp.py
gyou2021: Yes, I think so. However, it is not in the scope of this PR.

gyou2021:
Sure. I updated the code to enable it to run out-of-box.  Thank you for your comments.

> Hi @gyou2021 I like the goal to avoid repetition of same logic from L296 to L315, but I also have concern that models enabled by these lines will not be able to run out-of-box with this PR. This may not be friendly to self-helping users without access to proper BKC documentation to various models.
> 
> Could `allReduceLinearItems` have an initial value as a built-in list, then pre-pend with `os.environment` to get runtime configurability? I think if the model to be enabled by environment is a public model, it should be contributed to the built-in list to provide OOB experience, right?

deepspeed/module_inject/auto_tp.py
```diff
@@ -291,10 +291,35 @@ def tp_parser(model):
...
+        default_ds_common_reduceLinear_keys = ['out_proj', 'o_proj', 'down_proj']
+        #the different model names of the same key are concat with comma
+        predefined_ds_common_reduceLinear_items = {
+            'attention.dense': 'GPTNeoX',
+            'self_attention.dense': 'falcon,ChatGLM,Phi',
+            'w2': 'Mixtral',
+            'dense_4h_to_h': 'ChatGLM'
+        }
+
+        ds_user_reduceLinear_items = os.environ.get('DS_ALL_REDUCE_LINEAR_ITEMS')
```
delock, line 305: Suggest to add a line of comment how how a typical `DS_ALL_REDUCE_LINEAR_ITEMS` looks like.

deepspeed/module_inject/auto_tp.py
```diff
@@ -336,9 +368,27 @@ def _replace(self, child, name, conv_linear_layer):
...
-            ('mlp.shared_expert_gate' == name or 'mlp.gate' == name) and 'qwen2_moe' in str(type(self.module))):
-            return child
+        predefined_keep_Linear_Items = {
+            'q_a_proj': 'DeepseekV2',
+            'kv_a_proj_with_mqa': 'DeepseekV2',
+            'block_sparse_moe.gate': 'DeepseekV2',
+            'mlp.shared_expert_gate': 'qwen2_moe',
+            'mlp.gate': 'qwen2_moe'
+        }
+        user_keep_Linear_Items = os.environ.get('DS_KEEP_LINEAR_ITEMS')
```
delock, line 378: Same here, suggest to add a line of comments to give an example

deepspeed/module_inject/auto_tp.py
```diff
@@ -291,10 +291,35 @@ def tp_parser(model):
...
+            'w2': 'Mixtral',
+            'dense_4h_to_h': 'ChatGLM'
+        }
+
+        ds_user_reduceLinear_items = os.environ.get('DS_ALL_REDUCE_LINEAR_ITEMS')
+        if ds_user_reduceLinear_items:
+            ds_user_reduceLinear_items = ast.literal_eval(ds_user_reduceLinear_items)  #dict
+
+        ds_reduceLinear_items = predefined_ds_common_reduceLinear_items
+        ds_reduceLinear_items.update(ds_user_reduceLinear_items)
```
delock, line 310: what happends if DS_ALL_REDUCE_LINEAR_ITEMS is not defined?  A little experiement shows there will be TypeError.

deepspeed/module_inject/auto_tp.py
```diff
@@ -291,10 +291,35 @@ def tp_parser(model):
...
+        ds_reduceLinear_keys = ds_reduceLinear_items.keys()
+
+        ds_user_remove_reduceLinear_keys = os.environ.get('DS_REMOVED_COMMON_REDUCE_LINEAR_KEYS')
```
delock, line 313: Same here, add an example in comments

deepspeed/module_inject/auto_tp.py
gyou2021: Added the explanation and examples and updated the codes.

deepspeed/module_inject/auto_tp.py
gyou2021: Added the explanation and examples and updated the codes.

deepspeed/module_inject/auto_tp.py
gyou2021: If  DS_ALL_REDUCE_LINEAR_ITEMS is not defined, default LinearAllreduce items will be used. default LinearAllreduce items are those defined previously. The bug was fixed.

deepspeed/module_inject/auto_tp.py
gyou2021: Added explanations.
#'DS_REMOVED_COMMON_REDUCE_LINEAR_KEYS' is a list. The layer name in the list will be removed from those of default common LinearAllReduce.
#import os
# os.environ["DS_ALL_REDUCE_LINEAR_ITEMS"] = "['layer_name_1', 'layer_name_2',...]"
#for example: os.environ["DS_ALL_REDUCE_LINEAR_ITEMS"] = "['o_proj']"

delock:
@loadams my questions are all resolved and I have no further question to @gyou2021 , thanks!

delock:
Hi @loadams , is this PR under review?

loadams:
> Hi @loadams , is this PR under review? Thanks!

Hi @delock - sorry for the delay in this, we will work on getting this reviewed.

delock:
Close due to long time no update.

Issue #5828, opened 2024-08-05
[BUG] In deepspeed Zero3, RuntimeError: still have inflight params


XuyaoWang:
**Describe the bug**

When using zero stage 3, an issue occurs during training where some model parameters are selected based on data content. The error message is: RuntimeError: still have inflight params.

**To Reproduce**
1. Create a python file
```
touch test_inflight.py
```
2. Paste the following code into test_inflight.py.
```
import argparse
import deepspeed
import torch
import torch.distributed as dist
import torch.nn as nn
from torch.optim import AdamW
from transformers import get_scheduler
from transformers.deepspeed import HfDeepSpeedConfig


class MyNetwork(nn.Module):
    def __init__(self):
        super(MyNetwork, self).__init__()
        self.fc1 = nn.Linear(1024, 1024)
        self.fc2 = nn.Linear(1024, 1024)
        self.fc3 = nn.Linear(1024, 1)
        self.selu = nn.SELU()
        self.sigmoid = nn.Sigmoid()
    
    def forward(self, x, use_fc2):
        x = self.fc1(x)
        x = self.selu(x)
        if use_fc2:
            x = self.fc2(x)
        x = self.selu(x)
        x = self.fc3(x)
        x = self.sigmoid(x)
        return x

def main() -> None:
    parser = argparse.ArgumentParser()
    parser.add_argument('--local_rank', type=int, default=-1)
    parser = deepspeed.add_config_arguments(parser)
    args = parser.parse_args()

    model = MyNetwork()

    deepspeed.init_distributed()

    torch.cuda.set_device(args.local_rank)
    device = torch.device('cuda', args.local_rank)
    args.device = device
    args.global_rank = dist.get_rank()

    dist.barrier()

    ds_config = {
        'train_batch_size': None,
        'train_micro_batch_size_per_gpu': 8,
        'gradient_accumulation_steps': 1,
        'steps_per_print': 10,
        'zero_optimization': {
            'stage': 3,
            'offload_param': {
                'device': 'none',
            },
            'offload_optimizer': {
                'device': 'none',
            },
            'param_persistence_threshold': 1e4,
            'max_live_parameters': 3e7,
            'prefetch_bucket_size': 3e7,
            'memory_efficient_linear': False,
            'gather_16bit_weights_on_model_save': True,
        },
        'gradient_clipping': 1.0,
        'prescale_gradients': False,
        'wall_clock_breakdown': False,
    }


[...]

PR #6557, opened 2024-09-21
Clean up prefetched parameters

tohtana:
Parameters prefetched by ZeRO3 are sometimes not used. This occurs when the actual sub-module execution differs from previous tracing. As a result, the state of the allgather handle for such a parameter remains `INFLIGHT`, causing functions like `empty_partition_cache` to detect it and throw an error.
This PR resolves the issue by ensuring that communication finishes and the parameters are freed.

As this issue was mentioned in #6011, this includes the change of the branch. We need to merge #6011 first.


tjruwase:
Please check if this PR fixes #5828.

tohtana:
> Please check if this PR fixes #5828.

@tjruwase Using this PR branch, the repro in #5828 shows the message below but exits without throwing an error. I think this is expected as the model has a conditional branch and the execution order of modules changes.

```
Invalidate trace cache @ step 3: expected module 2, but got module 4
```

tjruwase (approved):

tohtana:
@tjruwase I added the cleaning of the inflight parameter registry in `_invalidate_trace` as you suggested. This allows us to free the gathered (but unused) parameters earlier. However, I also kept it in `reset_step`.
This is why we don't detect deviations from the trace when some modules at the end of the trace remain unvisited. The original assertion in `reset_step` will still be triggered in that case.

loadams (approved):

PR #6570, opened 2024-09-25
[COMPILE] workflow for deepspeed + torch.compile

YizhouZ:
We use simple model + deepspeed zero 3 + torch.compile and count graph break numbers to demonstrate current status of combing deepspeed + torch.compile.


.github/workflows/xpu-compile.yml
```diff
@@ -0,0 +1,58 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  compile-tests:
+    runs-on: [self-hosted, intel, xpu]
```
loadams, line 21: @YizhouZ - does it make more sense to add this to its own workflow, or to add it to the existing xpu workflow?

tests/torch_compile/test_compile.py
```diff
@@ -0,0 +1,103 @@
...
+    # runs backpropagation
+    model_engine.backward(loss)
+    # weight update
+    model_engine.step()
+
+dynamo_stats = get_dynamo_stats()
+dynamo_stats.subtract(start_stats)
+
+if comm.get_rank() == 0:
+    print(dynamo_stats)
```
loadams, line 103: @YizhouZ - could you run the pre-commit formatter on the PR, that will resolve the formatting/python issues.

delock:
I also have a question for discussion.  Initially I thought of reference test result of this workflow (as number of graph breaks) on a page, a little bit like workflow badges on DeepSpeed README page.  But I didn't find proper tool allowing us to do that.  Is there a pointer or direction where we should look at?  Thanks!

tests/torch_compile/test_compile.py
YizhouZ: > @YizhouZ - could you run the pre-commit formatter on the PR, that will resolve the formatting/python issues.

sure!

.github/workflows/xpu-compile.yml
YizhouZ: > @YizhouZ - does it make more sense to add this to its own workflow, or to add it to the existing xpu workflow?

you mean adding a new worker for this workflow?

delock:
xpu_compile workflow now passed.  Now the graph break number shows in test summary
https://github.com/microsoft/DeepSpeed/actions/runs/11044896580?pr=6570

tohtana (approved):
Thank you @YizhouZ, looks good to me. It is great to have utilities like this in our workflows.

PR #6577, opened 2024-09-27
Add llama3.2 vision autotp

Yejing-Lai:
Llama3.2-11b and llama3.2-90b including vision model and text model, these two models have different num_kv_heads, so we need to set num_kv_heads dynamically.


loadams (approved):

loadams:
@Yejing-Lai - this should be able to be merged now, I just needed to fix one more thing in the CI,.

PR #6578, opened 2024-09-27
[DO NOT MERGE] Debug leaked semaphore

tohtana:
Our CI tests frequently get stuck after showing the error below. You can find an example at [here](https://github.com/microsoft/DeepSpeed/actions/runs/11065075925/job/30743885716?pr=6557).

```
/opt/conda/envs/ptca/lib/python3.8/multiprocessing/resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 6 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

As this doesn't happen on my local env, I added logging to see what happens on the CI env.


tohtana:
I think the cause of the memory issue in tests has been identified as `reuse_dist_env` and it was disabled by #6623. We still see tests sometimes stuck and opened #6627 to track it down.

PR #6472, opened 2024-09-02
[Accelerator] Cambricon MLU support

Andy666G:
### Description
This PR includes Cambricon MLU accelerator support. 
With this PR, DeepSpeed supports MLU as backend for training and inference tasks.

PR #6584, opened 2024-09-27
Fix torch include in `op_builder/mlu/fused_adam.py` and update no-torch workflow triggers

loadams:
Changes from #6472 caused the no-torch workflow that is an example of how we build the DeepSpeed release package to fail (so we caught this before a release, see more in #6402).   These changes also copy the style used to include torch in other accelerator op_builder implementations, such as npu [here](https://github.com/microsoft/DeepSpeed/blob/master/op_builder/npu/fused_adam.py#L8) and hpu [here](https://github.com/microsoft/DeepSpeed/blob/828ddfbbda2482412fffc89f5fcd3b0d0eba9a62/op_builder/hpu/fused_adam.py#L15).

This also updates the no-torch workflow to run on all changes to the op_builder directory.  The test runs quickly and shouldn't add any additional testing burden there.

Resolves: #6576

### DIFF
--- a/op_builder/mlu/fused_adam.py
+++ b/op_builder/mlu/fused_adam.py
@@ -5,7 +5,11 @@
 # DeepSpeed Team
 
 from .builder import MLUOpBuilder
-import torch
+
+try:
+    import torch
+except ImportError as e:
+    pass
 
 
 class MLUFusedAdam:
--- a/.github/workflows/no-torch.yml
+++ b/.github/workflows/no-torch.yml
@@ -5,6 +5,7 @@ on:
   pull_request:
     paths:
       - '.github/workflows/no-torch.yml'
+      - 'op_builder/**'
   schedule:
     - cron: "0 0 * * *"
 


op_builder/mlu/fused_adam.py
```diff
@@ -5,7 +5,11 @@
 # DeepSpeed Team
 
 from .builder import MLUOpBuilder
-import torch
```
loadams, line 8: FYI @Andy666G - this change broke the no-torch workflow.  See more info in the PR comment, but please let us know if you have any concerns with this change,

op_builder/mlu/fused_adam.py
```diff
@@ -5,7 +5,11 @@
...
+
+try:
```
tohtana, line 9: Can we import `torch` in `multi_tensor_adam` if it is the only place where you need torch?
Then you can throw a more specific error when torch is unavailable and `multi_tensor_adam` is called.

jomayeri (approved):

PR #6590, opened 2024-09-29
Add API for updating ZeRO gradients

tjruwase:


deepspeed/utils/tensor_fragment.py
```diff
@@ -207,6 +216,26 @@ def safe_get_full_grad(param):
...
+
+        Args:
+            param (``torch.nn.Parameter``): A model parameter
+            value (``torch.Tensor``): New value
+    """
+    if param.grad is not None:
+        param.grad.copy_(value)
+        return
+
+    # ZeRO stage 3 param
```
tohtana, line 230: Should we clarify that only one case is fired from `if hasattr(param, 'ds_id')` and `if hasattr(param, '_hp_mapping')` by if-elif? Will this silently be nop when Z0 or caught by `if param.grad is not None`?

deepspeed/utils/tensor_fragment.py
```diff
@@ -207,6 +216,26 @@ def safe_get_full_grad(param):
     return None
 
 
+def safe_set_full_grad(param, value):
+    """Update the partitioned gradient of a low-precision (e.g., fp16) parameter.
+
+        Args:
+            param (``torch.nn.Parameter``): A model parameter
+            value (``torch.Tensor``): New value
```
tohtana, line 224: Can we clarify `value` is a partitioned value for ZeRO3? I think we pass the entire grad for ZeRO1. How about ZeRO2?

deepspeed/utils/tensor_fragment.py
```diff
@@ -207,6 +216,26 @@ def safe_get_full_grad(param):
...
+        param.grad.copy_(value)
+        return
+
+    # ZeRO stage 3 param
+    if hasattr(param, 'ds_id'):
+        param._z3_optimizer.set_fp32_grad_for_param(value, param)
+
+    # ZeRO stage 1, 2, and bf16_optimizer params
+    if hasattr(param, '_hp_mapping'):
+        param.set_full_hp_grad(value)
```
tohtana, line 236: The docstring says this API updates *low-precision* but it actually calls `set_full_hp_grad`. Is this correct?

deepspeed/utils/tensor_fragment.py
```diff
@@ -223,6 +252,22 @@ def safe_get_local_grad(param):
     return None
 
 
+def safe_set_local_grad(param, value):
+    """Update the gradient of a partitioned parameter.
+        Args:
+            param (``torch.nn.Parameter``): A model parameter
+            value (``torch.Tensor``): New value
+    """
+    if param.grad is not None:
```
tohtana, line 261: Does this condition cover all Z0,1,2? We have the same condition in `safe_set_full_grad` but it covers only Z0 there?

tohtana (commented):
I left some comments for clarification.

deepspeed/utils/tensor_fragment.py
tjruwase: Great question! The general principle of these APIs is that output values returned by `get_` and input values of `set_` are the full unpartitioned value. This is because we don't want users to worry about partitioning when manipulating these tensors. I will carefully update the docs globally to provide this clarification.

deepspeed/utils/tensor_fragment.py
tjruwase: Yes, you are correct, and I will improve this for clarity.  

Below is my current thinking - glad to get your thoughts.  
```python
  if param.grad is not None: 
  # Z0 and Z1-default
  elif hasattr(param, 'ds_id'):
 # Z3
  elif hasattr(param, '_hp_mapping'):
 # Z2, Z1+hooks (e.g., fp32 grad accum)
 else:
  Noop or maybe assert? 
```

deepspeed/utils/tensor_fragment.py
tjruwase: Yes, there might be some inconsistency here. The `_local_` APIs were contributed by a user because of their needs, and I didn't spend too much time thinking about it. The `_local_` APIs were created specifically for Z3 since that is the case where each rank holds an equal-sized partition of every parameter/gradient. So, it might be better to throw an assert for all other cases to provide user clarity.

deepspeed/utils/tensor_fragment.py
tjruwase: This is a great question. Stas and I decided to delay on this because there are multiple options. 
https://github.com/microsoft/DeepSpeed/blob/8cded575a94e296fee751072e862304676c95316/deepspeed/utils/tensor_fragment.py#L189

Let's have some brainstorming on this.

deepspeed/runtime/zero/stage3.py
```diff
@@ -2298,6 +2298,24 @@ def get_fp32_grad_for_param(self, param) -> Tensor:
...
+            return
+
+        if not get_accelerator().resolves_data_dependency():
+            self.reduce_and_partition_stream.synchronize()
+
+        if self.offload_optimizer:
+            group_idx, dest_offset, num_elements = self.grad_position[self.get_param_id(param)]
+            fp32_grad = self.fp32_partitioned_groups_flat[group_idx].grad.narrow(0, dest_offset, num_elements)
+        else:
+            fp32_grad = self.__param_id_to_grad_partition[param.ds_id]
```
tohtana, line 2312: The dtype of `__param_id_to_grad_partition` is `gradient_accumulation_dtype`. It is confusing to name the result `fp32_grad`.

tohtana: The same goes for `get_fp32_grad_partitions`.

tohtana: Can we be clear about the precision of the gradients regarding this? A typical use case is `safe_get_full_grad` -> (Modify gradients) -> `safe_set_full_grad`. 
Although `safe_get_full_grad` returns FP32 but it is actually converted from FP16/BF16 for the non-offloading setting (It is taken from FP32 for the offloading setting). It is not clear for users what the original precision is. In addition, `safe_set_full_grad` silently converts the FP32 gradients potentially modified by the user to FP16/BF16 in the non-offloading scenario, but it is hard for the user to know he might have lost the precision.

tohtana (approved):
Thank you @tjruwase, this looks good to me. The docstrings you added are very helpful.

PR #6592, opened 2024-09-30
AIO CPU Locked Tensor

jomayeri:
Restoring the functionality of the cpu locked tensor in the AIO library.
Make async_io operator available for CPU accelerator, i.e., CPU only environment.


tjruwase:
@jomayeri, need to re-enable these unit tests
https://github.com/microsoft/DeepSpeed/blob/8cded575a94e296fee751072e862304676c95316/tests/unit/ops/aio/test_aio.py#L81

tests/unit/ops/aio/test_gds.py
```diff
@@ -54,7 +54,7 @@ def _get_test_write_file_and_device_buffer(tmpdir, ref_buffer, gds_handle, index
 def _validate_handle_state(handle, single_submit, overlap_events):
     assert handle.get_single_submit() == single_submit
     assert handle.get_overlap_events() == overlap_events
-    assert handle.get_thread_count() == IO_PARALLEL
+    assert handle.get_thread_count() == 1
```
tjruwase, line 57: Why is `IO_PARALLEL` no longer correct?

jomayeri: For GDS the number of threads for `IO_PARALLEL` are set in the `cufile.json`. From the handle's point of view there is still only 1 thread. This [PR](https://github.com/microsoft/DeepSpeed/pull/6459/files) has the corresponding change in the GDS handle.

csrc/aio/py_lib/py_ds_aio.cpp
```diff
@@ -27,7 +27,7 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
              "queue_depth"_a = 128,
              "single_submit"_a = false,
              "overlap_events"_a = false,
-             "num_threads"_a = 1)
+             "intra_op_parallelism"_a = 1)
```
tjruwase, line 30: Let's also change the API `get_thread_count()` -> `get_intra_op_parallelism()`

tjruwase: Also need to update the tutorial 
https://github.com/microsoft/DeepSpeed/blob/master/docs/_tutorials/deepnvme.md#parallel-file-write

csrc/aio/py_lib/deepspeed_pin_tensor.cpp
```diff
@@ -19,17 +19,18 @@ deepspeed_pin_tensor_t::~deepspeed_pin_tensor_t()
...
     auto pinned_buffer = ds_page_aligned_alloc(num_bytes, true);
     assert(nullptr != pinned_buffer);
 
     _locked_tensors[pinned_buffer] = num_bytes;
 
-    auto options = torch::TensorOptions().dtype(elem_type).device(torch::kCPU);
+    auto options = torch::TensorOptions().dtype(elem_type).device(torch::kCPU).requires_grad(false);
 
-    return at::from_blob(pinned_buffer, static_cast<long int>(num_bytes), options);
+    return at::from_blob(pinned_buffer, static_cast<long long int>(num_elem), options);
```
tjruwase, line 33: Can you please explain the change of `num_bytes` -> `num_elem`?

jomayeri: If the ScalarType (data type) input is anything other than `uint8` num bytes will fail, because torch will create `num_bytes*elementSize` tensor; `numel` works in all cases.

tjruwase (approved):

csrc/aio/common/deepspeed_aio_utils.cpp
```diff
@@ -103,7 +103,7 @@ int get_file_size(const char* filename, long long int& size)
     return 0;
 }
 
-void* ds_page_aligned_alloc(const size_t size, const bool lock)
+void* ds_page_aligned_alloc(const long long int size, const bool lock)
```
tohtana, line 106: You can consider `int64_t`, which is used in PyTorch, instead of `long long int`

tohtana (approved):

csrc/aio/common/deepspeed_aio_utils.cpp
tjruwase: Done

PR #6597, opened 2024-10-01
Add first Step in LR Schedulers

jomayeri:
Some (not all) of the LR schedulers in runtime were missing the initialization of the optimizer group lr.


deepspeed/runtime/lr_schedules.py
```diff
@@ -675,11 +675,16 @@ def __init__(self,
         self.warmup_type = warmup_type
         self.inverse_log_warm_up = 1.0 / math.log(self.warmup_num_steps)
         self.last_batch_iteration = last_batch_iteration
+        # Initialize lr in optimizer
```
tohtana, line 678: These changes are duplicated. Can we extract the code as a function and reuse it?

deepspeed/runtime/lr_schedules.py
```diff
@@ -851,15 +851,17 @@ def get_lr_ratio(self):
         ratio = max(0.0, self.cos_min_ratio + ratio_delta * ratio)
         return ratio
 
+    def update_lr(self):
```
tohtana, line 854: Can we also consolidate this with the previous one?

tohtana (approved):

PR #6601, opened 2024-10-06
[compile] Show breakdown of graph break

delock:
This PR extends https://github.com/microsoft/DeepSpeed/pull/6570 by showing a breakdown of graph breaks.  So we can see how graph breaks are distributed among different reasons.  An example of graph break output can be seen from the following workflow run https://github.com/microsoft/DeepSpeed/actions/runs/11199157962


tests/torch_compile/test_compile.py
```diff
@@ -14,22 +14,11 @@
 
 torch._dynamo.config.cache_size_limit = 100
 
-import collections
-
 
 def get_dynamo_stats():
     # TODO: consider deepcopy'ing the entire counters struct and
     # adding a helper to do subtraction on it
```
tohtana, line 20: Is this comment still valid?

tohtana (approved):
@delock Thank you @delock, this is a very useful feature! I left a comment, but it looks good to me.

tests/torch_compile/test_compile.py
delock: Thanks for the comments, I think they should be removed.  Will update.

delock:
@tohtana the unneeded comments had been removed.  Also make the workflow to count zero2, there is no graph breaks for zero2 in this simple case.

https://github.com/microsoft/DeepSpeed/actions/runs/11269655884?pr=6601

PR #6621, opened 2024-10-11
Ensure exclusive allocation of master port

tohtana:
The base class for distributed tests allocates master port after checking availability of ports and `PYTEST_XDIST_WORKER`. This potentially leads to conflicts of ports:
- Anther process may allocate the port after checking the availability of ports
- The scope of `PYTEST_XDIST_WORKER` is a single workflow but we run multiple workflows in parallel.

This PR maintains the use of ports by keeping records in a file. We lock the file when updating the records to ensure exclusive allocation of master port.


tohtana:
Closing as we don't run multiple workflows in parallel on the same machine.

PR #6622, opened 2024-10-11
[Bug Fix] Support threads_per_head < 64 for wavefront size of 64

jagadish-amd:
When launching apply_rotary_pos_half kernel, only threads_per_head of 64 is supported for wavefront size of 64.
This change adds support for threads_per_head < 64 such as 4, 8, 16.

Fixes the issue introduced in https://github.com/microsoft/DeepSpeed/pull/5402


jagadish-amd:
ping @jithunnair-amd @jeffdaily @loadams

jagadish-amd:
@loadams any comments on this PR?

csrc/transformer/inference/csrc/apply_rotary_pos_emb.cu
```diff
@@ -116,10 +108,11 @@ __global__ void apply_rotary_pos_half(T* mixed_query,
         LAUNCH_ROT_POS_EMB_HALF(16, ALIGNMENT); \
     } else if (threads_per_head == 32) {        \
         LAUNCH_ROT_POS_EMB_HALF(32, ALIGNMENT); \
+    } else if (threads_per_head == 64) {        \
+        LAUNCH_ROT_POS_EMB_HALF(64, ALIGNMENT); \
```
tjruwase, line 112: @jagadish-amd, are you confident this logic will work on non-amd since it only previously ran on amd? Would it be safer to condition this to amd?

jithunnair-amd: Actually that brings up a good question: what would be the right set of deepspeed unit tests or workloads to test the various cases in this logic block?

sfc-gh-reyazda: there are several unit tests here for the different operations we support: https://github.com/microsoft/DeepSpeed/tree/master/tests/unit/ops/transformer/inference. The RoPE one is missing though :) I think it would be nice to add one, @jagadish-amd wound you mind adding one to test these different dimensions? Thanks

sfc-gh-reyazda: @jagadish-amd, as Tunji mentioned too this new condition is not valid in Nvidia GPUs, as the Warp size is 32 vs the 64 for AMD's wavefront. So, the kind of blocks we are creating [here](https://github.com/microsoft/DeepSpeed/blob/master/csrc/transformer/inference/csrc/apply_rotary_pos_emb.cu#L42) can result into two warps and thus we cannot safely run the shfl instruction that is needed for [this part](https://github.com/microsoft/DeepSpeed/blob/master/csrc/transformer/inference/csrc/apply_rotary_pos_emb.cu#L75). I think we have a better implementation for this here: https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/inference/v2/kernels/ragged_ops/linear_blocked_kv_rotary/blocked_kv_rotary_cuda.cu which reads the data in larger granularity which ensures the warp-size is not exceeded here: https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/inference/v2/kernels/ragged_ops/linear_blocked_kv_rotary/blocked_kv_rotary_cuda.cu#L49

jagadish-amd: Thanks all for the inputs. I added code only for warpsize 64 and retained old code for warpsize 32.

tjruwase (approved):

csrc/transformer/inference/csrc/apply_rotary_pos_emb.cu
```diff
@@ -113,6 +114,20 @@ __global__ void apply_rotary_pos_half(T* mixed_query,
     } else {                                    \
```
jithunnair-amd, line 114: @jagadish-amd Looks like there's more repetition here than necessary. If we really need to limit the 64 case to HIP/AMD, we can have just that condition ifdef'd. Right, @tjruwase?

tjruwase: I see your point @jithunnair-amd. But I am fine either way.

jagadish-amd: cannot have ifdef inside define.

jithunnair-amd (commented):
LGTM, but let's add a unit test to ensure this functionality can be tested on ROCm (and CUDA)

loadams (approved):

loadams:
> LGTM, but let's add a unit test to ensure this functionality can be tested on ROCm (and CUDA)

@jagadish-amd - thoughts on adding unit tests for this?

jagadish-amd:
> > LGTM, but let's add a unit test to ensure this functionality can be tested on ROCm (and CUDA)
> 
> @jagadish-amd - thoughts on adding unit tests for this?

I will add the unit tests. Thanks

jagadish-amd:
> > > LGTM, but let's add a unit test to ensure this functionality can be tested on ROCm (and CUDA)
> > 
> > 
> > @jagadish-amd - thoughts on adding unit tests for this?
> 
> I will add the unit tests. Thanks

@loadams I have added the test case to test the threads_per_head ,warp size alignment issue. 
Unfortunately, I lost access to the AI model / node for which the "Assertion `false' failed" error had triggered on warp_size 64 node. Hence the values in the test cases are assumed here, but it still tests the intended fix. I will add the exact values in the future.
I noticed that the files in unit/ops/transformer/inference were refactored. I have used the InferenceBuilder().load() way to test the function apply_rotary_pos_emb. If this is not right, plz let me know, we can merge the PR without the newly added test. (and later folks can add the test cases, this change affects only warp_size 64 case / AMD Instinct device).

These are the results.
On warp_size = 32, test cases pass regardless of the changes in apply_rotary_pos_emb.cu
==================================== 4 passed, 2 warnings in 25.24s ====================================
On warp_size =64, with the fix.
==================================== 4 passed, 2 warnings in 5.58s =====================================
On warp_size =64, without the fix.
python: /opt/conda/envs/py_3.10/lib/python3.10/site-packages/deepspeed/ops/csrc/transformer/inference/csrc/apply_rotary_pos_emb.hip:169: void launch_apply_rotary_pos_emb(T *, T *, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int, float, hipStream_t, int) [T = float]: Assertion `false' failed.
Fatal Python error: Aborted

The test run is aborted (as expected) due to the error in kernel. Not sure if there is better way to handle this?

loadams:
> > > > LGTM, but let's add a unit test to ensure this functionality can be tested on ROCm (and CUDA)
> > > 
> > > 
> > > @jagadish-amd - thoughts on adding unit tests for this?
> > 
> > 
> > I will add the unit tests. Thanks
> 
> @loadams I have added the test case to test the threads_per_head ,warp size alignment issue. Unfortunately, I lost access to the AI model / node for which the "Assertion `false' failed" error had triggered on warp_size 64 node. Hence the values in the test cases are assumed here, but it still tests the intended fix. I will add the exact values in the future. I noticed that the files in unit/ops/transformer/inference were refactored. I have used the InferenceBuilder().load() way to test the function apply_rotary_pos_emb. If this is not right, plz let me know, we can merge the PR without the newly added test. (and later folks can add the test cases, this change affects only warp_size 64 case / AMD Instinct device).
> 
> These are the results. On warp_size = 32, test cases pass regardless of the changes in apply_rotary_pos_emb.cu ==================================== 4 passed, 2 warnings in 25.24s ==================================== On warp_size =64, with the fix. ==================================== 4 passed, 2 warnings in 5.58s ===================================== On warp_size =64, without the fix. python: /opt/conda/envs/py_3.10/lib/python3.10/site-packages/deepspeed/ops/csrc/transformer/inference/csrc/apply_rotary_pos_emb.hip:169: void launch_apply_rotary_pos_emb(T *, T *, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int, float, hipStream_t, int) [T = float]: Assertion `false' failed. Fatal Python error: Aborted
> 
> The test run is aborted (as expected) due to the error in kernel. Not sure if there is better way to handle this?

@jagadish-amd - this should be fine for now.  I believe the only remaining thing for this PR is the CLA agreement, you should just need to reply to it with accept and company as AMD.

loadams (approved):

jagadish-amd:
> @jagadish-amd please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 

Issue #6620, opened 2024-10-11
[BUG] failed to find frozen {param} in named params


ssklzx:
**Describe the bug**
failed to find frozen {param} in named params

**To Reproduce**
 use accerate deepspeed to train flux. 1

 accelerator = Accelerator()
 model, optimizer, data = accelerator.prepare(model, optimizer, data)
 device_map = {}
  model = accelerate.dispatch_model(model, device_map=device_map)
accelerator.save_state(save_path)

When I use accelerate.dispatch_model after accelerator.prepare, there will be an error when saving the model

Traceback (most recent call last):
  File "/root/paddlejob/workspace/env_run/x-flux/train_flux_lora_deepspeed.py", line 391, in <module>
    main()
  File "/root/paddlejob/workspace/env_run/x-flux/train_flux_lora_deepspeed.py", line 311, in main
    accelerator.save_state(save_path)
  File "/root/paddlejob/workspace/env_run/xflux_train_python3/lib/python3.10/site-packages/accelerate/accelerator.py", line 2952, in save_state
    model.save_checkpoint(output_dir, ckpt_id, **save_model_func_kwargs)
  File "/root/paddlejob/workspace/env_run/xflux_train_python3/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 3164, in save_checkpoint
    self._save_checkpoint(save_dir,
  File "/root/paddlejob/workspace/env_run/xflux_train_python3/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 3365, in _save_checkpoint
    frozen_param_shapes=self._get_zero_frozen_param_attributes(self._get_param_shape_func)
  File "/root/paddlejob/workspace/env_run/xflux_train_python3/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 3432, in _get_zero_frozen_param_attributes
    raise ValueError(f"failed to find frozen {param} in named params")
ValueError: failed to find frozen Parameter containing:
tensor([[-0.0520,  0.0588,  0.0430,  ..., -0.0391,  0.0083, -0.0104],
        [-0.0320,  0.0408,  0.0112,  ..., -0.0121, -0.0264, -0.0081],
        [ 0.0055,  0.0493,  0.0488,  ..., -0.0088, -0.0187,  0.0135],
        ...,
        [-0.0957,  0.0220,  0.0087,  ..., -0.0021, -0.0474, -0.0645],
[...]

PR #6626, opened 2024-10-15
modify_load_save_model

ssklzx:
accelerator = Accelerator()
model, optimizer, data = accelerator.prepare(model, optimizer, data)
device_map = {}
model = accelerate.dispatch_model(model, device_map=device_map)
accelerator.save_state(save_path)

When I use accelerate.dispatch_model after accelerator.prepare, there will be an error when saving the model

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -2678,12 +2678,10 @@ class DeepSpeedEngine(Module):
 
         if checkpoint.get(FROZEN_PARAM_FRAGMENTS, None) is not None:
             saved_frozen_params = checkpoint[FROZEN_PARAM_FRAGMENTS]
-            for param in self.module.parameters():
+            for name, param in self.module.named_parameters():
                 if param.requires_grad:
                     continue
-                if param not in self.param_names:
-                    raise ValueError(f"failed to find frozen {param} in named params")
-                name = self.param_names[param]
+
                 if hasattr(param, 'ds_id'):
                     param.ds_tensor.data.copy_(saved_frozen_params[name].data)
                 else:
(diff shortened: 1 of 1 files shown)


tjruwase:
@ssklzx, thanks for creating this PR. However, I think you misunderstood my response 
https://github.com/microsoft/DeepSpeed/issues/6620#issuecomment-2412593947. 

What I meant is that we need to debug further to understand why some parameters are missing from `self.param_names`. Are you able to provide a full repro?

ssklzx:
> @ssklzx, thanks for creating this PR. However, I think you misunderstood my response [#6620 (comment)](https://github.com/microsoft/DeepSpeed/issues/6620#issuecomment-2412593947).
> 
> What I meant is that we need to debug further to understand why some parameters are missing from `self.param_names`. Are you able to provide a full repro?

Because after the initialization of 'self. param_name', I will change the position of the parameters, such as moving them from 'CUDA: 0' to 'CUDA: 1', so I will not be able to find these transferred parameters

for example:
model, optimizer, data = accelerator.prepare(model, optimizer, data). # initialization  'self. param_name'
model = accelerate.dispatch_model(model, device_map=device_map)  # change parameter position
accelerator.save_state(save_path) # report errors

tjruwase:
> Because after the initialization of 'self. param_name', I will change the position of the parameters, such as moving them from 'CUDA: 0' to 'CUDA: 1', so I will not be able to find these transferred parameters

@ssklzx, thanks for the clarification. I think the correct solution here is for accelerate and DeepSpeed to coordinate to ensure that DeepSpeed is aware of new parameter locations, including updating `self.param_names`

ssdevtech:
hi

PR #6645, opened 2024-10-21
[XPU] [DeepNVMe] use same cpu_op_desc_t with cuda

Liangliang-Ma:
We have found that #6592 uses `_pinned_tensor_mgr` to create cpu bounce buffer, which is same with what our xpu accelerator currently doing. 
So no need to use xpu device specific cpu_op_desc_t.
In this PR:
1. remove custom csrc/xpu/aio/deepspeed_cpu_op.cpp
2. modify xpu async_io opbuilder.

This issue cannot be easily done with revert #6532 , for we added some source file as last time GDS feature going in DS. So file this new PR :)


tjruwase (approved):

Liangliang-Ma:
@tjruwase Could you please help to re-trigger the checks for xpu-max1100 workflow updation#6646 ?
Thanks!

tjruwase:
> @tjruwase Could you please help to re-trigger the checks for xpu-max1100 workflow updation#6646 ? Thanks!

@delock, are you able to trigger workflows?

PR #6648, opened 2024-10-21
Update gaudi2 docker version to latest release (1.18)

raza-sikander:
Updated docker version to 1.18.0-latest

Note: for this update the firmware on the Gaudi2 node had to be updated to use firmware version 1.18.

### DIFF
--- a/deepspeed/module_inject/tp_shard.py
+++ b/deepspeed/module_inject/tp_shard.py
@@ -24,7 +24,9 @@ def set_n_embd(num):
 
 def get_num_kv_heads():
     global num_kv_heads
-    return num_kv_heads
+    if 'num_kv_heads' in globals():
+        return num_kv_heads
+    return None
 
 
 def get_num_attention_heads():
--- a/deepspeed/profiling/flops_profiler/profiler.py
+++ b/deepspeed/profiling/flops_profiler/profiler.py
@@ -115,7 +115,7 @@ class FlopsProfiler(object):
                 get_accelerator().synchronize()
                 module.__start_time__ = time.time()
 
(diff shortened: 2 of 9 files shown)


loadams (approved):

.github/workflows/hpu-gaudi2.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     # The type of runner that the job will run on
     runs-on: [self-hosted, intel, gaudi2]
     container:
-      image: vault.habana.ai/gaudi-docker/1.17.0/ubuntu22.04/habanalabs/pytorch-installer-2.3.1:latest
+      image: vault.habana.ai/gaudi-docker/1.18.0/ubuntu22.04/habanalabs/pytorch-installer-2.4.0:latest
```
loadams, line 42: FYI @raza-sikander it appears that this test is running too long and timing out, do you know if this version is expected to be slower or how this could impact the test speed?

raza-sikander: The tests shouldnt be impacted. Let me check.

loadams: It appears to work fine with 1.17.1 with either torch 2.3.1 or 2.4, it seems we might need to update the underlying software on the node itself.

raza-sikander: Yes, On local machine with latest image it works fine. On checking both hl-smi difference we can see firmware differences.

loadams: It appears to work once we update the underlying node's firmware to 1.18 (from 1.17).  I've done that and will merge this PR shortly, thanks!

PR #6658, opened 2024-10-23
A faster and more memory-efficient implementation of `zero_to_fp32`

xu-song:
It is a faster and more memory-efficient implementation of `zero_to_fp32`.

The previous version double the memory usage, which cause cpu OOM for very large models (e.g. llama 405B).
https://github.com/microsoft/DeepSpeed/blob/b647fb2470f8f6fefe5cab0ea84a2d89696eb898/deepspeed/utils/zero_to_fp32.py#L438-L441

## How does it work?

1. **Lazy loading**: Load checkpoint with `mmap=True`, thus the weights are mmaped rather than loading all the storages into memory.
2. **Lazy merge**: `GatheredTensor` contains the mmaped weights and tensor offset. It is a memory-efficient pseudo tensor. Only when `tensor.contiguous()` is called, it starts to load related weights to memory and merge into a single tensor.
3. **Release memory in time**: Save checkpoints shard by shard, and release the memory once a shard is saved.

Throughout the process, only one shard of tensors are keeped in memory.

## How much benefit in speed and memory ?

Experiments were conducted on a linux host with 1TB of memory. Here is a detailed comparision
|                      | world size | peak memory(GB)  | elapsed time(h:mm:ss) |
|----------------------|------------|--------------|--------------------|
| llama3-8B(old->new)  | 8          | 90 -> 41 | 0:02:17 -> 0:01:10 |
| llama2-13B(old->new)  | 8        | 146 -> 54 | 0:02:30 -> 0:01:47  |
| llama2-70B(old->new)  | 16        | 789 -> 159 | 0:20:47 -> 0:20:45 |
| qwen1.5-110B(old->new)  | 32       | OOM -> 217 | ? -> 0:34:21 |
| llama3-405B(old->new)  | 192      | OOM -> 262 | ? -> 2:09:59 |

You can reproduce with the following scripts
```sh
# 1. install requirments
apt-get install time
# 2. prepare zero-3 checkpoints
# 3. convert zero to fp32 checkpoints
/usr/bin/time -v python zero_to_fp32.py . output_dir/ --safe_serialization
```

- **memory**: Theoretically, this PR reduces the memory cost from `2M` to `(1/n)M`, where `M` is the memory cost of the full weights, `n` is num_shards.
- **speed**: The speed gain mainly comes from avoiding extra tensor copying. The benifit may be slight.

## Impl history

- [v1](https://github.com/xu-song/DeepSpeed/commit/19712a1c75bfc1da4a7f3ecca6915a86af671568#diff-6a2ca3427fa608c387b7351359f98cfc1313be6e960cee86344ff246bf1b8326R441-R447) :  a hf_hub compatible approach. 
    It has been discarded due to the controversial implementation of `data_ptr().`
- [v2](https://github.com/microsoft/DeepSpeed/pull/6658/files): a simple approach with `torch.empty`


deepspeed/utils/zero_to_fp32.py
```diff
@@ -305,6 +303,7 @@ def _zero2_merge_trainable_params(state_dict, world_size, fp32_flat_groups, zero
             if debug:
                 print(f"{name} full shape: {shape} unpartitioned numel {unpartitioned_numel} ")
             state_dict[name] = full_single_fp32_vector.narrow(0, offset, unpartitioned_numel).view(shape)
+            # tensor = GatheredTensor(fp32_flat_groups, offset, partitioned_numel, shape)
```
tjruwase, line 306: Remove comment

deepspeed/utils/zero_to_fp32.py
```diff
@@ -398,9 +397,71 @@ def _zero3_merge_frozen_params(state_dict, world_size, zero_model_states):
...
+                    end_part = i
+                    break
+            for i in range(start_part, end_part + 1):
+                flat_tensor = flat_group[i]
+                start_offset = self.offset - sum_idx[i]
+                end_offset = min(end_idx, sum_idx[i + 1]) - sum_idx[i]
+                tensor_slice.append(flat_tensor[start_offset:end_offset])
+
+            pad_flat_param_chunks.append(torch.concat(tensor_slice, 0))
+        pad_flat_param = torch.cat(pad_flat_param_chunks, dim=0).to(torch.float16)
```
tjruwase, line 437: Why is this cast to fp16? The checkpoint could be bf16.

deepspeed/utils/zero_to_fp32.py
```diff
@@ -435,10 +496,9 @@ def _zero3_merge_trainable_params(state_dict, world_size, fp32_flat_groups, zero
                 f"Trainable params: {total_params} {name} full shape: {shape} partition0 numel={partitioned_numel} partitioned_padding_numel={partitioned_padding_numel}"
             )
 
-        # XXX: memory usage doubles here
-        state_dict[name] = torch.cat(
-            tuple(fp32_flat_groups[i].narrow(0, offset, partitioned_numel) for i in range(world_size)),
-            0).narrow(0, 0, unpartitioned_numel).view(shape)
+        # memory efficient tensor
+        tensor = GatheredTensor(fp32_flat_groups, offset, partitioned_numel, shape)
```
tjruwase, line 500: This type change from `torch.Tensor` to a pseudo tensor seems problematic, especially for non-HF usage. Can you please explain the rationale and implications?

tjruwase: Okay. now I see that pseudo tensor is converted to `torch.Tensor` by `contiguous()`.

deepspeed/utils/zero_to_fp32.py
```diff
@@ -398,9 +397,71 @@ def _zero3_merge_frozen_params(state_dict, world_size, zero_model_states):
...
+                end_offset = min(end_idx, sum_idx[i + 1]) - sum_idx[i]
+                tensor_slice.append(flat_tensor[start_offset:end_offset])
+
+            pad_flat_param_chunks.append(torch.concat(tensor_slice, 0))
+        pad_flat_param = torch.cat(pad_flat_param_chunks, dim=0).to(torch.float16)
+        param = pad_flat_param[:self.shape.numel()].view(self.shape).contiguous()
+        return param
+
+    # The following part make it compatible with `huggingface_hub.split_torch_state_dict_into_shards`
+    # https://github.com/huggingface/huggingface_hub/blob/src/huggingface_hub/serialization/_torch.py
```
tjruwase, line 442: Can you clarify why this HF_Hub APIs are needed since pseudo tensor is never exported into the output checkpoint file?

deepspeed/utils/zero_to_fp32.py
```diff
@@ -398,9 +397,71 @@ def _zero3_merge_frozen_params(state_dict, world_size, zero_model_states):
...
+    def storage(self):
+        return self
+
+    def data_ptr(self):
+        return self.offset * 100000 + self.partitioned_numel
```
tjruwase, line 447: What does the hardcoded constant, `100000`, mean?

tjruwase:
@xylian86, FYI

deepspeed/utils/zero_to_fp32.py
xu-song: https://github.com/microsoft/DeepSpeed/blob/6e6563d3c8d7527713cc48d4a3adce51f22e83a2/deepspeed/utils/zero_to_fp32.py#L565-L566

HF_Hub APIs is used to split wights to shards.

deepspeed/utils/zero_to_fp32.py
xu-song: `tensor.storage().data_ptr()` is part of torch_storage_id, which is used to distinguish different tensor storage.


https://github.com/huggingface/huggingface_hub/blob/v0.26.1/src/huggingface_hub/serialization/_torch.py#L455

deepspeed/utils/zero_to_fp32.py
xu-song: more comments have been added to the script.

deepspeed/utils/zero_to_fp32.py
xu-song: to fp16 has been removed

deepspeed/utils/zero_to_fp32.py
xu-song: done

deepspeed/utils/zero_to_fp32.py
tjruwase: My question is why is this compatibility a requirement? To my understanding, the output checkpoint file will contain `torch.Tensor` and not pseudo tensor. Am I wrong?

deepspeed/utils/zero_to_fp32.py
tjruwase: Thanks for the explanation. I will mark this as resolved, but my original question about the need for this logic remains.

tjruwase:
@xu-song, just to clarify, we greatly appreciate this PR. The memory and speed benefits are very useful. My only concern are the HF_Hub related changes, so hopefully those can be clarified. 

Can you please add the observed speed and memory benefits of this optimizations? Such details are generally useful for readers to better appreciate the value. Thanks!

xu-song:
> @xu-song, just to clarify, we greatly appreciate this PR. The memory and speed benefits are very useful. My only concern are the HF_Hub related changes, so hopefully those can be clarified.
> 
> Can you please add the observed speed and memory benefits of this optimizations? Such details are generally useful for readers to better appreciate the value. Thanks!

Is there any alternative approach of sharding checkpoint in deepspeed?

tjruwase:
> @tjruwase Is there any alternative approach to sharding torch state_dict?
> 
> If any, the compatible feature to `huggingface_hub.split_torch_state_dict_into_shards` can be discarded.

Sorry, but I am a bit confused about the objective of this PR. The goal of [zero_to_fp32](https://deepspeed.readthedocs.io/en/latest/model-checkpointing.html#zero-checkpoint-fp32-weights-recovery) is to create a consolidated checkpoint state from the sharded checkpoints of ZeRO-* training, so I don't understand why state_dict sharding is a consideration here. 

It seems that there are two parts of this PR.
1. Speed and memory optimizations
2. HF_hub compatibility involving state_dict sharding

Am I correct?

xu-song:
> > @tjruwase Is there any alternative approach to sharding torch state_dict?
> > If any, the compatible feature to `huggingface_hub.split_torch_state_dict_into_shards` can be discarded.
> 
> Sorry, but I am a bit confused about the objective of this PR. The goal of [zero_to_fp32](https://deepspeed.readthedocs.io/en/latest/model-checkpointing.html#zero-checkpoint-fp32-weights-recovery) is to create a consolidated checkpoint state from the sharded checkpoints of ZeRO-* training, so I don't understand why state_dict sharding is a consideration here.
> 
> It seems that there are two parts of this PR.
> 
> 1. Speed and memory optimizations
> 2. HF_hub compatibility involving state_dict sharding
> 
> Am I correct?

https://github.com/microsoft/DeepSpeed/blob/54903e09eb131bb7b69bfc154e3970d4958131b9/deepspeed/utils/zero_to_fp32.py#L565-L567

To save memory, the tensors in state_dict is pesudo tensor instead of torch tensor.

xu-song:
> > @tjruwase Is there any alternative approach to sharding torch state_dict?
> > If any, the compatible feature to `huggingface_hub.split_torch_state_dict_into_shards` can be discarded.
> 
> Sorry, but I am a bit confused about the objective of this PR. The goal of [zero_to_fp32](https://deepspeed.readthedocs.io/en/latest/model-checkpointing.html#zero-checkpoint-fp32-weights-recovery) is to create a consolidated checkpoint state from the sharded checkpoints of ZeRO-* training, so I don't understand why state_dict sharding is a consideration here.
> 
> It seems that there are two parts of this PR.
> 
> 1. Speed and memory optimizations
> 2. HF_hub compatibility involving state_dict sharding
> 
> Am I correct?

https://github.com/microsoft/DeepSpeed/blob/54903e09eb131bb7b69bfc154e3970d4958131b9/deepspeed/utils/zero_to_fp32.py#L565-L567

To save memory, the tensors in state_dict is pesudo tensor instead of torch tensor.

xu-song:
> > @tjruwase Is there any alternative approach to sharding torch state_dict?
> > If any, the compatible feature to `huggingface_hub.split_torch_state_dict_into_shards` can be discarded.
> 
> Sorry, but I am a bit confused about the objective of this PR. The goal of [zero_to_fp32](https://deepspeed.readthedocs.io/en/latest/model-checkpointing.html#zero-checkpoint-fp32-weights-recovery) is to create a consolidated checkpoint state from the sharded checkpoints of ZeRO-* training, so I don't understand why state_dict sharding is a consideration here.
> 
> It seems that there are two parts of this PR.
> 
> 1. Speed and memory optimizations
> 2. HF_hub compatibility involving state_dict sharding
> 
> Am I correct?

deepspeed/utils/zero_to_fp32.py
xu-song: Loading state_dict of `torch.Tensor` costs large memory.

tjruwase:
> Besides, our implementation exactly follows the the goal of zero_to_fp32. As the [document](https://deepspeed.readthedocs.io/en/latest/model-checkpointing.html#zero-checkpoint-fp32-weights-recovery) says

Apologies for not being clear. The reason that I referred to that doc was to show that `zero_to_fp32` designed to generate a consolidated checkpoint and not a sharded one. Does this clarify the intention of `zero_to_fp32`?

deepspeed/utils/zero_to_fp32.py
```diff
@@ -483,6 +530,7 @@ def get_fp32_state_dict_from_zero_checkpoint(checkpoint_dir, tag=None, exclude_f
         - ``checkpoint_dir``: path to the desired checkpoint folder
         - ``tag``: checkpoint tag used as a unique identifier for checkpoint. If not provided will attempt to load tag in 'latest' file. e.g., ``global_step14``
         - ``exclude_frozen_parameters``: exclude frozen parameters
+        - ``lazy_merge``: a more memory-efficient feature
```
tjruwase, line 533: Provide brief description of why more memory-efficient, and perhaps mention important usage concepts like pseudo tensors and `contiguous()`.
 
Also, please add a unit test.

xu-song:
> > Besides, our implementation exactly follows the the goal of zero_to_fp32. As the [document](https://deepspeed.readthedocs.io/en/latest/model-checkpointing.html#zero-checkpoint-fp32-weights-recovery) says
> 
> Apologies for not being clear. The reason that I referred to that doc was to show that `zero_to_fp32` is meant to generate a consolidated checkpoint and not a sharded one. Does this clarify the intention of `zero_to_fp32`?

截屏2024-10-29 09 06 33

You may confused the intension of `zero_to_fp32.py` and `get_fp32_state_dict_from_zero_checkpoint`

tjruwase:
> If your point is `no-sharding is allowed`, then ignore and close this pr.

No, my point is not `no-sharding is allowed`. Like I said earlier, this PR is very useful, and we are appreciative. I notice that new PR does not contain HF_hub compatibility logic, and so my concern is addressed. Thanks for that.

deepspeed/utils/zero_to_fp32.py
tjruwase: This would be a good place for a unit test: https://github.com/microsoft/DeepSpeed/blob/0e11b081be237f9b1867a3af88479a23de11345f/tests/unit/checkpoint/test_shared_weights.py#L48

xu-song: thanks

xu-song:
a unit test has been added

tjruwase (approved):

xu-song:
formatting and unit test issue have been resolved.

tjruwase:
> formatting and unit test issue have been resolved.
@xu-song, apologies that our CI issues is delayed merge. Thanks for your patience.

NicholasCao:
I have a question that's a bit off-topic. Why is it necessary to convert zero to fp32 instead of directly converting it to fp16 or bf16? I don't understand this. Is there a zero_to_bf16.py script available?

PR #6661, opened 2024-10-23
Clean all param coordinators

tohtana:
Clean the status of param coordinators for both training/non-training modes.


tohtana:
Closing to try a simpler solution in #6662.

PR #6694, opened 2024-10-31
Reduce the device bubble introduced by heavy loop synchronization in coalesced fetch/release(z3_leaf_module)

inkcherry:
depend on https://github.com/microsoft/DeepSpeed/pull/6649

When performing fetch/release operations on Z3 leaf modules, the loop time is excessively long in fine-grained module. Compared to non-leaf modules, Z3 leaf modules may include a larger number of parameters. Although each loop unit does not consume much time, the overall loop length can be significant.
image
**The fetch time is impacted by:**

Post-allgather operations (narrow， slice ，cat, difficult to avoid)
Memory pressure（record_stream/fetch event create&sync）
**The release time is impacted by:**
slice
Free parameter record_stream

Considering the fine-grained leaf modules, where each parameter is relatively small, we can treat the parameters within each leaf module as a unified entity to handle memory pressure. This approach can approximately halve the CPU time required for fetch/release operations.


inkcherry:
Hi , @tjruwase. Some models use a large number of MoE experts, but the total number of parameters for MoE remains unchanged. This PR handles such cases as a whole to optimize memory management and reduce bubbles caused by long loops. when ```zero_module_granularity_threshold``` is enabled.

Do you have any suggestions?

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -704,24 +706,30 @@ def wait(self) -> None:
                     partitions.append(part_to_copy)
             param.data = instrument_w_nvtx(torch.cat)(partitions).view(param.ds_shape)
             param.ds_status = ZeroParamStatus.AVAILABLE
-
-            for part_to_copy in partitions:
-                if not get_accelerator().is_synchronized_device():
-                    part_to_copy.record_stream(get_accelerator().current_stream())
+            if handle_dependency:
```
tjruwase, line 709: Would the following not work?

```suggestion
            if handle_dependency and not get_accelerator().is_synchronized_device() :
```

inkcherry: Thanks, it's better to move the condition outside the loop, modified.

deepspeed/runtime/zero/parameter_offload.py
```diff
@@ -145,6 +145,16 @@ def __init__(
             module.ds_inflight_param_registry = InflightParamRegistry()
         self.__inflight_param_registry = module.ds_inflight_param_registry
 
+        self.fast_fetch_for_leaf_module = False
```
tjruwase, line 148: @inkcherry, since this great feature accelerates both fetch and release, I think it would be helpful for the naming to reflect this. The current naming suggests that only fetch is accelerated. So, can we consider more general naming, such as below?

```suggestion
        self.fast_sharding_for_leaf_module = False
```

inkcherry: thanks~ yes, I’ve updated the naming .

tjruwase (approved):

inkcherry:
hi, @tjruwase ,It seems that the OOM in the CI is not caused by this PR. Could you please help retrigger it? Thanks!

PR #6717, opened 2024-11-06
Update GH hosted workflows to 24.04

loadams:
`ubuntu-latset` is moving to be 24.04, so we should test updating as well to ensure it doesn't break any of our workflows.


.github/workflows/formatting.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
   # formatting and basic install on cpu-only machine
   unit-tests:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
loadams, line 21: Need to work around `ModuleNotFoundError: No module named 'lib2to3'` since this was removed in python 3.13

loadams: Need to update yapf to work around this.

loadams: yapf changes are included in #6721

loadams: PR merged.

PR #6726, opened 2024-11-08
Fix potential memory issues when use deepspeed Z3

wenbinc-Bin:
I had OOM problem when doing DPO training using zero3. It needs to call module twice in one training step, and second call is with no_grad().
The problem is caused by two bugs:
1. "__n_available_params", which helps to control fetched parameters, becomes negative after release_and_reset_all() function.
2. module.ds_grads_remaining becomes negative in backward() if we call module more than once in one training step.

I tried to create two patches to fix these issues.

### DIFF
--- a/deepspeed/runtime/zero/parameter_offload.py
+++ b/deepspeed/runtime/zero/parameter_offload.py
@@ -392,7 +392,8 @@ class DeepSpeedZeRoOffload(object):
                                                                _run_after_backward_hook, inputs)
 
         def _post_backward_module_hook(module, inputs):
-            module.ds_grads_remaining = 0
+            if not hasattr(module, "ds_grads_remaining"):
+                module.ds_grads_remaining = 0
 
             if not hasattr(module, "post_bwd_fn"):
 
--- a/deepspeed/runtime/zero/partitioned_param_coordinator.py
+++ b/deepspeed/runtime/zero/partitioned_param_coordinator.py
@@ -252,7 +252,6 @@ class PartitionedParameterCoordinator:
         self.__most_recent_step_id_param_fetched_for = collections.defaultdict(lambda: int(-1e10))
         self.__step_id_module_fetched_for = collections.defaultdict(lambda: collections.deque())
         self.__step_id = 0
-        self.__n_available_params = 0
         self.__profiler.reset_events()
 
     def _dump_params(self, tag, sub_module, params, step_id=None):
@@ -430,7 +429,7 @@ class PartitionedParameterCoordinator:
             # there's a hook execution issue
             param.ds_active_sub_modules.clear()
             self.__release_param(param)
-
+        self.__n_available_params = 0
         for param in iter_params(module, recurse=True):
             if param.ds_status != ZeroParamStatus.NOT_AVAILABLE:
                 raise RuntimeError(f"{param.ds_summary()} expected to be released")
--- a/tests/unit/runtime/zero/test_zero_multiple_run.py
+++ b/tests/unit/runtime/zero/test_zero_multiple_run.py
@@ -20,10 +20,6 @@ class TestZ3MultipleModelCall(DistributedTest):
             "zero_optimization": {
                 "stage": 3
             },
-            "fp16": {
-                "enabled": True,
-                "initial_scale_power": 8
-            },
             "optimizer": {
                 "type": "Adam",
                 "params": {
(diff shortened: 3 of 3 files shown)


wenbinc-Bin:
@microsoft-github-policy-service agree [company="Intel"]

wenbinc-Bin:
@microsoft-github-policy-service agree company=Intel

wenbinc-Bin:
@microsoft-github-policy-service agree company=Intel

deepspeed/runtime/zero/partitioned_param_coordinator.py
```diff
@@ -430,6 +430,7 @@ def release_and_reset_all(self, module: Module) -> None:
             # there's a hook execution issue
             param.ds_active_sub_modules.clear()
             self.__release_param(param)
+        self.__n_available_params = 0
```
tjruwase, line 433: This looks like a hack to me. I think it would better to understand why the following decrement is insufficient
https://github.com/microsoft/DeepSpeed/blob/12189feb4f0377dbdbd9337ec14a590b58714f23/deepspeed/runtime/zero/partitioned_param_coordinator.py#L500

deepspeed/runtime/zero/parameter_offload.py
```diff
@@ -392,7 +392,8 @@ def _run_before_forward_function(input):
                                                                _run_after_backward_hook, inputs)
 
         def _post_backward_module_hook(module, inputs):
-            module.ds_grads_remaining = 0
+            if not hasattr(module, "ds_grads_remaining"):
```
tjruwase, line 395: Can you please explain why `hasattr` is needed here?

tjruwase:
@wenbinc-Bin, thanks for the PR. Are you able to provide unit tests for this?

deepspeed/runtime/zero/partitioned_param_coordinator.py
wenbinc-Bin: I explained in detail in my commit log. I copy it here:
"__n_available_params" is set to 0 in reset_step() which is called in
backward(). All parameter are released in release_and_reset_all() which
is called in step(). "__n_available_params" is reduced when parameter is
released. These mean if step() is called after backward(),
"__n_available_params" will be reduced to a negative number.
"__n_available_params" is used to restrict fetched parameters, so
negative value leads to a problem that fetched parameter will be larger
than upper bound ("__max_n_available_params").

deepspeed/runtime/zero/parameter_offload.py
wenbinc-Bin: Also copy to here:
"ds_grads_remaining" is used to triger post_backward_function(). If the
module is called more than once in one training step, this variable will
be initialized every time. If backward() is also called multiple times,
"ds_grads_remaining" will be reduced to a negative number.
post_backward_function() will not be called as expected. This leads to
extra fetch operation or extra memory usage.

wenbinc-Bin: Use "hasattr" to avoid repeat initialization.

wenbinc-Bin:
> @wenbinc-Bin, thanks for the PR. Are you able to provide unit tests for this?

Ok, I try to add a unit test.

deepspeed/runtime/zero/partitioned_param_coordinator.py
wenbinc-Bin: Generally speaking, "__n_available_params" is set to 0 in reset_step() and this leads to wrong value for following release operation.

tjruwase: Thanks for your explanation. Yes, setting `_n_available_params` to 0 in `reset_step()` is a bug since it is inconsistent with the fetched parameter count. Strictly speaking, `_n_ available_params` should be modified by gather (increment) and release (decrement) operations. Since this a pre-existing bug, it is not a requirement that you fix. But it would be greatly appreciated if you could do so. Thanks!

wenbinc-Bin: Ok, I will update the PR later with unit test together.

wenbinc-Bin: I found there is a reason to set `__n_available_params=0`. This variable becomes negative somehow after second step. I need to investigative.

wenbinc-Bin: There is persistent_parameters, which are gathered outside of Class "PartitionedParameterCoordinator". "__n_availiable_params" cannot recode these information, so it becomes negative after release all parameters. That is why there is a line of code to set "__n_availiable_params" to 0. If that is the case, then we need to move "`__n_availiable_params = 0`" to release_and _reset_all() instead of reset_step()

wenbinc-Bin:
I saw unit-tests workflow failed. I fix the issue and update the PR.

tjruwase (approved):

PR #6728, opened 2024-11-08
Increase CI coverage for Gaudi2

raza-sikander:


loadams (approved):

loadams:
Tagging @raza-sikander for the failures on the newly added tests.  Let us know if we need to modify anything on the node.

.github/workflows/hpu-gaudi2.yml
```diff
@@ -49,51 +49,64 @@ jobs:
       TORCHINDUCTOR_COMPILE_THREADS: 1
       TEST_LIST: |
         test_accelerator.py
+        test_activation_checkpointing.py
```
loadams, line 52: @raza-sikander - it looks like the tests pass now but take 2.5 hours to run.  Should we switch some of the tests to a cron job if they take that long?  Since triggering on as many PRs as this yml file does means we may see queueing there.

raza-sikander: Hi @loadams, We can split it as you mentioned. 
Can you share the process for creation of cron job? So i can split the content two.
And also it would mean we would have two jobs one for CI and other like nightly?

loadams: Hi @raza-sikander - yes, we would need to split into 2 yml files, one for PRs and one for cron jobs, since they would run a different set of tests.  

Here is a sample of the nv-nightly test that only runs on PRs that modify that file, and otherwise only runs once per day, set via the `cron` option. https://github.com/microsoft/DeepSpeed/blob/877aa0dba673c2aa2157029c28363b804d6ee03d/.github/workflows/nv-nightly.yml#L3C1-L9C24

Otherwise we would just remove the `schedule` section from the current yml file and it would continue to run on PRs, though with a smaller set of tests to ensure it ran more quickly, ideally the ~1 hour it takes now?

raza-sikander: @loadams thanks for link.
In that case ill add the other tests to `cron` in a seperate file as suggested with that the content would be split into 1hr and 1hr 20mins approx.

raza-sikander: Closing this PR. will raise the additional tests into the cron job pr.

PR #6750, opened 2024-11-14
Add the missing view operations from sequence parallel(async).

inkcherry:
FYI @loadams 

a view operation was missing in some updates compared to the original version https://github.com/microsoft/DeepSpeed/blob/17ed7c77c58611a923a6c8d2a3d21d359cd046e8/deepspeed/sequence/layer.py#L56

add missing view operation.
The shape required for the view cannot be easily obtained in the current function， so refactor layout params code.


loadams:
> @loadams The CI encountered a "no space left" issue, which doesn't seem to be caused by this patch. Could you please retrigger it? thanks！

Hi @inkcherry - yes, sorry that does seem to be an intermittent failure, I'll re-trigger them, but it looks like there are merged conflicts now, could you take a look?

inkcherry:
@loadams Thank you for the reminder, I have resolved and verified the merge conflicts(both ds+megads)   : )

tohtana (approved):

PR #6753, opened 2024-11-15
Gaudi2 Nightly job for daily check

raza-sikander:


.github/workflows/hpu-gaudi2-nightly.yml
```diff
@@ -0,0 +1,103 @@
+name: hpu-gaudi2-nightly
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
+  pull_request:
+    paths:
+      - ".github/workflows/hpu-gaudi2.yml"
```
loadams, line 9: To make this a nightly only PR, and only have the PR trigger when we touch this file, we should remove all the paths from this section except for the ".github/workflows/hpu-gaudi2.yml", though it should be updated to match the name of this file.

loadams: Updated

loadams (approved):

PR #6764, opened 2024-11-19
Use `json_schema_extra` instead of extra keyword in `Field`

qgallouedec:
> Using extra keyword arguments on `Field` is deprecated and will be removed. Use `json_schema_extra` instead. (Extra keys: 'new_param'). Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.9/migration/


loadams (approved):

loadams:
Thanks @qgallouedec - looks like we missed this before, thanks for the PR!

Issue #6718, opened 2024-11-06
[BUG] DeepSpeed accuracy issue for torch.compile if activation checkpoint function not compiler disabled


jerrychenhf:
**Describe the bug**
When running DeepSpeed with torch.compile and activation checkpointing on LLAMA2 model training, if activation checkpoint function is not compiler disabled, we found accuracy issues for the training (fine tuning).

DeepSpeed implemented a WA for resolving this accuracy issue:
https://github.com/microsoft/DeepSpeed/pull/5590

But the root cause was never identified and a proper fix from the root cause should be provided instead of a WA.

**To Reproduce**
The following condition need to be satisfied to reproduce:
1. We used LLAMA2 7B model
2. torch.compile is used for compiling the model being trained
3. Peft LoRA optimization applied to the model
4. DeepSpeed ZeRO3 training
5. Activation checkpointing is enabled
6. remove the compiler.disable decorator on the DeepSpeed checkpoint function.
**Expected behavior**
The accuracy with or without compiler.disable decorator on the DeepSpeed checkpoint function should be the same. 
But actually, without compiler.disable decorator on the DeepSpeed checkpoint function, the end accuracy achieved is much lower.

We recently have a in-depth investigation and analysis to this DeepSpeed issue and found the root cause of the issue. This bug entry is to track this problem we found behind the scene so that we can provide a PR to fix which can be understood and accepted.

What is happening:
1. LoRA and activation checkpointing is in use, the "x = x.to(lora_A.weight.dtype)" code line the triggered a special path to ZeRO3 parameter offloading. Which is ZeROOrderedDict getitem method.

```
result = self.base_layer(x, *args, **kwargs)

lora_A = self.lora_A[active_adapter]
lora_B = self.lora_B[active_adapter]
dropout = self.lora_dropout[active_adapter]
scaling = self.scaling[active_adapter]
x = x.to(lora_A.weight.dtype)
result = result + lora_B(lora_A(dropout(x))) * scaling
```
2.  ZeROOrderedDict getitem calls into param.all_gather if the param data is not avaiable.

```
        if hasattr(param, "ds_status") and getattr(param, "ds_status") == ZeroParamStatus.NOT_AVAILABLE:
            if self._parent_module._parameters._in_forward:
                register_external_parameter(FWD_MODULE_STACK[-1], param)
                param.all_gather()
                print_rank_0(f'Registering external parameter from getter {key} ds_id = {param.ds_id}', force=False)
```
[...]

PR #6773, opened 2024-11-21
Stage3: Use new torch grad accumulation hooks API

deepcharm:
* This commit addresses a Deepspeed issue [#6718](https://github.com/microsoft/DeepSpeed/issues/6718)
* The existing code has been using the grad_acc node hook to reduce params grads.
   The constructs such as `param.data = replicated_tensor.data` used in `allgather_params(..)`
   are compiled into `param.set()` causing the hook assigned to the grad_acc node not being called.
* The above caused accuracy issues and could be temporarily solved by simply disabling the torch compile when activation checkpointing is used.
* This commit provides a clean solution by replacing the hook on a grad_acc node to a hook using a new and robust hook API on a param itself: `param.register_post_accumulate_grad_hook(..)`

### DIFF
--- a/deepspeed/runtime/zero/stage3.py
+++ b/deepspeed/runtime/zero/stage3.py
@@ -1156,7 +1156,6 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
 
     def create_reduce_and_remove_grad_hooks(self):
         print_rank_0(f'[Begin] Create gradient reduction hooks')
-        self.grad_accs = []
         self.leaf_parameters = defaultdict(list)
         for i, param_group in enumerate(self.fp16_groups):
             for param in param_group:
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/zero/stage3.py
```diff
@@ -1169,15 +1168,13 @@ def create_reduce_and_remove_grad_hooks(self):
...
-                        grad_acc = param_tmp.grad_fn.next_functions[0][0]
 
                         @instrument_w_nvtx
                         def reduce_partition_and_remove_grads(*notneeded):
                             self.reduce_ready_partitions_and_remove_grads(param)
 
-                        self._grad_acc_hooks.append(grad_acc.register_hook(reduce_partition_and_remove_grads))
-                        self.grad_accs.append(grad_acc)
+                        self._grad_acc_hooks.append(
+                            param.register_post_accumulate_grad_hook(reduce_partition_and_remove_grads))
```
tjruwase, line 1177: Which pytorch version introduced this API? How should we handle older versions?

deepcharm: Hi @tjruwase

- This API was introduced starting pytorch v2.1.
- We can add check what version and use the older API when needed.
- However, compilation should be disabled for the older versions to avoid the accuracy issues (if activation checkpointing is enabled).

Please advice what should be the approach. Thanks.

deepcharm: In fact, in PT 2.0 the compile feature was still quite unstable. So, maybe we discourage compilation on version earlier than 2.1 by displaying a warning in the engine.compile().
@tjruwase what are your thoughts on this?

tjruwase: @deepcharm, apologies for the delayed response. I am aligned with your suggestion, specifically, we should enable for >= 2.1 and disable for older versions. In other words, no need to warn for 2.0 given the unstable nature. Thanks!

deepcharm: @tjruwase thanks for your comments. I've updated the patch which now:
1) Uses proper grad accumulation hook API depending on the PyTorch version
2) Disables compile for PyTorch < 2.1

Let me know if anything is missing. Thanks

tjruwase (approved):

deepspeed/runtime/zero/stage3.py
tjruwase: @deepcharm, looks great. Thanks!

tohtana (approved):

deepcharm:
I can see other pending PRs are consistently failing on the unit-tests.
@loadams is that a known issue?

loadams:
> I can see other pending PRs are consistently failing on the unit-tests. @loadams is that a known issue?

@deepcharm - this is a known issue we are working on fixing.

PR #6803, opened 2024-11-28
Zero2: avoid graph breaks in torch.compile by using param_idx

nelyahu:
inside reduce_independent_p_g_buckets_and_remove_grads and in reduce_ipg_grads which are being executed during the BWD hook in zero2, the model param is being stored inside params_in_ipg_bucket. torch.compile has hard time tracing parameters.
By using the param's static index inside the group the same logic can be maintain with less complexity.


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1067,7 +1068,7 @@ def average_tensor(self, tensor):
 
             process_group = self.dp_process_group
             # count = 0
-            for i, param, param_id in self.params_in_ipg_bucket:
+            for i, param_idx_in_group, param_id in self.params_in_ipg_bucket:
```
tjruwase, line 1071:
```suggestion
            for i, param_idx_in_group, param_id in self.params_in_ipg_bucket:
                 param = self.bit16_groups[group_idx][param_idx_in_group]
```

tjruwase: @nelyahu, do you have any thoughts on this suggestion?

nelyahu: @tjruwase Done. sorry for the delay

tjruwase (approved):

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1067,7 +1068,8 @@ def average_tensor(self, tensor):
...
+                param = self.bit16_groups[group_idx][param_idx_in_group]
```
loadams, line 1072: @nelyahu - this is failing on group_idx just fyi

nelyahu: fixed

PR #6812, opened 2024-12-02
Support pure meta model lm_head tp

Yejing-Lai:
Add lm_head tp support when checkpoint not provided to deepspeed.init_inference().


delock:
Hi @Yejing-Lai can you describe the high level scenarios that this PR enables?  Thanks!

loadams (approved):

deepspeed/module_inject/replace_module.py
```diff
@@ -399,6 +396,7 @@ def conv2d_parallel_shard_weights(model, rank, world_size):
                                          orig_class=orig_layer_impl,
                                          replace_fn=replace_fn,
                                          _replace_policy=config.injection_policy_tuple)
+    replaced_module = set_lm_head(replaced_module)
```
loadams, line 399: @Yejing-Lai - thanks for the PR, sorry the review took so long.  I understand the above changes to `set_lm_head`, but could you explain the reasoning for the changes here or if it is tied into the above changes?

Yejing-Lai: This change is for setting lm_head tp when checkpoint_dict is None. It may be more clear to add another logical judgment.

loadams: Thanks @Yejing-Lai, that makes sense.

PR #6828, opened 2024-12-06
Add MLP/lm_head tp grain size setting.

Yejing-Lai:
This PR aims to add MLP/lm_head tp size granularity setting to deepspeed.init_inference() API. It will be more flexible to set the MLP/lm_head sharding grain size. 

DNN library favors tensor size in granularity of power of 2, we pick 64 as a default size.

We aim to be able to set the MLP/lm_head tp grain size flexibly. This is a preliminary solution. If there is a better solution, we can discuss it together. Thanks~


tjruwase (approved):

tjruwase:
@delock, @Yejing-Lai, thanks for the PR. I agree that updating `init_inference` is right choice. However, I am curious whether it is better to make `grain_size` a sub-field of the existing `tensor_parallel` field.
https://github.com/microsoft/DeepSpeed/blob/1b58ba5ec04493a112fae10d9cc9c824dfbd40ca/deepspeed/inference/config.py#L132-L136

Did you guys consider this choice? Either way, I have approved the PR.

Yejing-Lai:
> @delock, @Yejing-Lai, thanks for the PR. I agree that updating `init_inference` is right choice. However, I am curious whether it is better to make `grain_size` a sub-field of the existing `tensor_parallel` field.
> 
> https://github.com/microsoft/DeepSpeed/blob/1b58ba5ec04493a112fae10d9cc9c824dfbd40ca/deepspeed/inference/config.py#L132-L136
> 
> Did you guys consider this choice? Either way, I have approved the PR.

Thank you for your advice. I updated the code. Please review~

tjruwase:
> Thank you for your advice. I updated the code. Please review~

@Yejing-Lai, thanks for updating the code. LGTM.

PR #6846, opened 2024-12-10
[inf] Add config var to enable keeping module on host

oelayan7:
Using keep_module_on_host config var will let us control if the loaded checkpoints to model parameters will be moved to the device or stay on host


deepspeed/module_inject/auto_tp.py
```diff
@@ -17,9 +17,11 @@
 from deepspeed.module_inject.tp_shard import get_shard_size, get_shard_size_list
 
 
-def move(tensor, device):
+def move(tensor, device, keep_module_on_host=False):
```
tjruwase, line 20: Won't it be simpler to modify callers to pass device='cpu' when `keep_module_on_host=True`?

oelayan7: Done

oelayan7:
@tjruwase Can you please rereview and retrigger the CI?

tests/unit/inference/test_inference.py
```diff
@@ -554,6 +554,7 @@ def test(self, model_w_task, injection_policy, query, inf_kwargs, assert_fn, dty
 
 
 @pytest.mark.seq_inference
+@pytest.mark.parametrize('keep_module_on_host', [True, False])
```
tjruwase, line 557: Is it useful to validate that tensors are on `cpu` when `keep_module_on_host=True`?

oelayan7: You're right. 
A check was added.

oelayan7: @tjruwase @loadams 
The changes were made as requested, can someone re-review it please?

tjruwase (approved):

Issue #6772, opened 2024-11-20
[BUG] [Fix-Suggested] ZeRO Stage 3 Overwrites Module ID Attribute Causing Incorrect Expert Placement on GPUs


traincheck-team:
## Description

We experienced wrong GPU placement when doing MoE with ZeRO Stage 3. We use `module.id` to control which expert to be loaded onto which GPU for finegrained controlm and we find out that `module.id` got corrupted after `deepspeed.initialize`.

## Suspected Root Cause

DeepSpeed uses `.id` in ZeRO Stage 3 optimization to manage states, as seen in [`runtime/zero/parameter_offload.py:L271`](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/zero/parameter_offload.py#L269-L271). 

This practice is very brittle in that:
1. `id` is an overly generic attribute name, might get easilly collided with some user-defined attributes.
2. There's no special check on `.id` attribute before setting it, this allows for accidental overwrites of the attribute, causing hard-to-diagnose problems.

In the specific bug we've encountered (bug.py provided below), each expert module is identified by the `.id` attribute, but during initialization, the `.id` is overwritten by the `_register_hooks_recursively` function in `deepspeed/runtime/zero/stage3.py`, leading to a mess on expert-GPU placement.

### To reproduce

The following code in ZeRO Stage 3 is responsible for overwriting the `.id` attribute:

1. Install deepspeed `0.15.4`

2. run `bug.py` using `deepspeed --num_gpus=2 bug.py` (num_gpus argument here doesn't matter, use 1 if you don't have multigpu nodes.)

```python
import torch
import deepspeed
from torch.nn import Module, Linear

# Define a simple expert module
class Expert(Module):
    def __init__(self, id):
        super().__init__()
        self.id = id  # ID for custom GPU placement
        self.fc = Linear(128, 128)

    def forward(self, x):
        return self.fc(x)

# Create a model with 60 experts
class MoEModel(Module):
    def __init__(self):
        super().__init__()
        self.experts = torch.nn.ModuleList([Expert(i) for i in range(60)])
    def forward(self, x, expert_id):
        return self.experts[expert_id](x)

# Helper function to log expert ids
def log_expert_ids(model, rank):
    loaded_experts = [e.id for e in model.experts]

def main():
    deepspeed.init_distributed()
    rank = torch.distributed.get_rank()

    # Create model
    model = MoEModel()
[...]

Issue #6771, opened 2024-11-20
[BUG] [Fix-Suggested] Checkpoint Inconsistency When Freezing Model Parameters Before `deepspeed.initialize`


traincheck-team:
Related Issue: https://github.com/microsoft/DeepSpeed/issues/5489

## Describe the Bug

If the model has frozen parameters (`requires_grad == False`) before calling `deepspeed.initialize`, the frozen layers are excluded from the parameters managed by the DeepSpeed optimizer. However, the original (wrapped) optimizer still retains these parameters.

As a result, when calling the `deepspeed_engine` to save the model checkpoint, the optimizer state checkpoint (`zero_pp_rank_0_mp_rank_00_optim_states.pt`) only contains parameters managed by the DeepSpeed optimizer, leading to incomplete or inconsistent checkpointing.

## Suspected Root Cause

When model parameters are modified (e.g., set `requires_grad=False` to freeze layers) before `deepspeed.initialize`, the DeepSpeed optimizer only manages the parameters that are still trainable. This inconsistency occurs because the frozen parameters are no longer part of the DeepSpeed-managed parameter group but remain covered by the original optimizer. Consequently, during checkpointing, only the active parameters are saved, resulting in a partial optimizer state and potential errors when attempting to resume training.

## Expected Behavior / Suggested Fix

To address this issue, we propose the following fixes:

1. **Parameter Modification Enforcement**: Ensure that `deepspeed.initialize` raises an error / warning if modifications to parameters (e.g., changing `.requires_grad`) are detected before the initialization.
2. **Consistent Checkpointing**: Add a validation step during checkpoint saving to verify that all parameters in the optimizer match the model parameters, thereby ensuring consistency and providing a clear warning if a discrepancy is found.

## Steps to Reproduce the Bug

Below is an example that reproduces the issue by freezing the model parameters before initializing DeepSpeed:

```python
import torch
import deepspeed

# Define a model
def initialize_model():
    model = torch.nn.Linear(10, 1)
    return model

def freeze_parameters(model):
    for param in model.parameters():
        param.requires_grad = False

# Define an optimizer
def expose_bug():
    model = initialize_model()
    optimizer = torch.optim.Adam(filter(lambda p: p.requires_grad, model.parameters()), lr=1e-3)

    # Freeze model parameters
    freeze_parameters(model)

    # Initialize DeepSpeed
    ds_config_fp16 = {
        "train_micro_batch_size_per_gpu": 1,
[...]

Issue #6770, opened 2024-11-20
[BUG] [Fix-Suggested] KeyError in stage_1_and_2.py Due to Optimizer-Model Parameter Mismatch


traincheck-team:
## Describe the bug

related to https://github.com/microsoft/DeepSpeed/issues/3718

An KeyError is thrown inside `deepspeed.initialize` at [`runtime/zero/stage_1_and_2.py", line 574, in _create_param_mapping`](https://github.com/microsoft/DeepSpeed/blob/83e4364fbd921ad4fe808d31eb107ef080228a7a/deepspeed/runtime/zero/stage_1_and_2.py#L574), due to inconsistent usage of model parameters and parameters managed by the optimizer.

<details>
<summary><b>Full Traceback (Click to Show)</b></summary>
<br>
<pre><code>```log
Traceback (most recent call last):
  File "/home/yuxuan/gitrepos/machine-learning-issues/DS-3718/bug.py", line 53, in <module>
    expose_bug()
  File "/home/yuxuan/gitrepos/machine-learning-issues/DS-3718/bug.py", line 50, in expose_bug
    model_engine, _, _, _ = deepspeed.initialize(model=model, optimizer=optimizer, config_params=ds_config_fp16)
  File "/home/yuxuan/miniconda3/envs/ml-daikon/lib/python3.10/site-packages/deepspeed/__init__.py", line 193, in initialize
    engine = DeepSpeedEngine(args=args,
  File "/home/yuxuan/miniconda3/envs/ml-daikon/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 313, in __init__
    self._configure_optimizer(optimizer, model_parameters)
  File "/home/yuxuan/miniconda3/envs/ml-daikon/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 1302, in _configure_optimizer
    self.optimizer = self._configure_zero_optimizer(basic_optimizer)
  File "/home/yuxuan/miniconda3/envs/ml-daikon/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 1560, in _configure_zero_optimizer
    optimizer = DeepSpeedZeroOptimizer(
  File "/home/yuxuan/miniconda3/envs/ml-daikon/lib/python3.10/site-packages/deepspeed/runtime/zero/stage_1_and_2.py", line 553, in __init__
    self._param_slice_mappings = self._create_param_mapping()
[...]

PR #6848, opened 2024-12-10
Fix: ensure `deepspeed.initialize` can only be called once.

traincheck-team:
Partially Fixes: #6772 #6771 #6770 by forbidding repeated initialization.
- Introduced a module-level variable `_deepspeed_initialized` to track initialization state.
- Raised a `RuntimeError` if `initialize` is called more than once.
- Ensured initialization logic is executed only on the first call.

**This is a draft issue as we have not included the unit test for this PR**

@tjruwase 
We currently added a global variable to enforce that `deepspeed.initialize` can only be called once in one single interpreter.
However, this seems to have interfered with existing tests that might intentionally call this API multiple times.
image

A solution might be to record `id` of the models and optimizers that have been used, to enforce the semantics only for individual models and optimizers, but that might get trickier and complicated, thus we'd like to hear some feedback on how this should be implemented before proceeding.

### DIFF
--- a/accelerator/real_accelerator.py
+++ b/accelerator/real_accelerator.py
@@ -125,10 +125,9 @@ def get_accelerator():
         if accelerator_name is None:
             try:
                 import intel_extension_for_pytorch as ipex
+
                 if ipex._C._has_xpu():
                     accelerator_name = "xpu"
-                else:
-                    accelerator_name = "cpu"
             except ImportError as e:
                 pass
         if accelerator_name is None:
@@ -162,7 +161,6 @@ def get_accelerator():
             except ImportError as e:
                 pass
         if accelerator_name is None:
-            # borrow this log from PR#5084
             try:
                 import torch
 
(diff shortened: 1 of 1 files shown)


tjruwase:
@traincheck-team, thanks for creating this PR so quickly. Unfortunately, it seems there is some miscommunication of the expectation. We want to maintain support for multiple `deepspeed.initialize()` calls as it is a widely used feature such as in [DeepSpeed-Chat](https://github.com/microsoft/DeepSpeedExamples/tree/master/applications/DeepSpeed-Chat) and [Accelerate](https://huggingface.co/docs/accelerate/en/usage_guides/deepspeed_multiple_model). Instead, what we want to prevent is repeated `deepspeed.initialize()` on a specific model, optimizer, lr_scheduler, etc. instance. And so, the deepspeed-specific attribute indicating prior `deepspeed.intialize()` should be attached to those objects. 

What do you think?

traincheck-team:
That makes sense and clear. Will change the code soon.

PR #5257, opened 2024-03-12
Enable mixtral 8x7b autotp

Yejing-Lai:
This PR aims to enable mixtral 8x7b (MoE model) autotp.

PR #6856, opened 2024-12-11
Add arctic model support by adding w2 to all_reduce

pi314ever:
As title says. 

Default behavior of arctic model produces shape issues with AutoTP due to the MLP layer performing `w2 * act(w1*w3)`. However, method provided to fix Mixtral-7x8b in #5257 does not work since the MLP for Arctic is also used within a ModuleList for the MoE. This results in MLP weights hiding behind individual experts as layers `#.w#`, which is not caught by the fix in #5257. This adds the check directly within replace, where it can check for actual layer names for the `w2` key in the model to patch with `all_reduce`.


pi314ever:

jeffra:
@RezaYazdaniAminabadi @sfc-gh-reyazda can you take a look?

deepspeed/module_inject/auto_tp.py
```diff
@@ -346,11 +346,15 @@ def _replace(self, child, name, conv_linear_layer):
                 weight, bias = shard_value_with_share_qk(child.weight.data, child.bias, dist.get_rank(),
                                                          dist.get_world_size(), False)
                 return LinearAllreduce(weight, bias, self.mp_group)
+        # For Arctic model, bypass to all_reduce replacement for w2 weights
+        arctic_w2_all_reduce_linear = False
+        if 'Arctic' in str(self.module) and 'w2' in name:
+            arctic_w2_all_reduce_linear = True
```
sfc-gh-reyazda, line 352: Thanks for catching this :) May I ask why you don't add these layers to the all_reduce_linears rather than introducing a new argument here? I am afraid the logic here gets so messy with adding more and more options for specific models.

sfc-gh-reyazda: I see, then in that case, I agree with your solution. Thanks @pi314ever for the clarification.

tjruwase (approved):

PR #6874, opened 2024-12-16
Fix: forbid repeated deepspeed.initialize on training objects

traincheck-team:
Previously closed PR:
https://github.com/microsoft/DeepSpeed/pull/6848

Partially Fixes: https://github.com/microsoft/DeepSpeed/issues/6772 https://github.com/microsoft/DeepSpeed/issues/6771 https://github.com/microsoft/DeepSpeed/issues/6770 by forbidding repeated initialization.

**What are changed:**
1. Marking 'model', 'optimizer' and 'lr_scheduler' in the arguments of `deepspeed.initialize` with the flag `ds_is_inited = True`.
2. Marking 'engine', 'engine.optimizer' and 'engine.lr_scheduler' in the return values of `deepspeed.initialize` with the flag `ds_is_inited = True`.
3. When calling `deepspeed.initialize`, raise an exception if detected `ds_is_inited == True` in the input `model`, `optimizer` or `lr_scheduler`

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -66,6 +66,50 @@ __git_branch__ = git_branch
 dist = None
 
 
+def _mark_initialized(trainobj: Union[torch.nn.Module, Optimizer, _LRScheduler]):
+    """Mark a trainobj as initialized by setting the ds_is_inited attribute to True."""
+    # we shouldn't hit the assert below, but just in case
+    assert not hasattr(
+        trainobj, 'ds_is_inited'
+    ), "Model has already been initialized, please make sure to only call deepspeed.initialize on a model once."
+    trainobj.ds_is_inited = True
+
+
+def _is_initialized(trainobj: Union[torch.nn.Module, Optimizer, _LRScheduler]):
+    """Check if a trainobj has been initialized by checking the ds_is_inited attribute."""
+    if hasattr(trainobj, 'ds_is_inited'):
+        # we shouldn't hit the assert below, but just in case
+        assert trainobj.ds_is_inited, "Not expecting the model has `ds_is_inited` to be False if it exists, make sure you didn't set it to False or called deepspeed.initialize on the model more than once."
+        return True
+    return False
+
+
+def _assert_trainobjs_not_inited(model: torch.nn.Module, optimizer: Optional[Optimizer],
+                                 lr_scheduler: Optional[_LRScheduler]):
+    """Enforce the model, optimizer, and lr_scheduler have not been used in a previous deepspeed.initialize call."""
+    if _is_initialized(model):
+        raise ValueError(
+            "Model has already been initialized, please make sure to only call deepspeed.initialize on a model once.")
+    if optimizer is not None and _is_initialized(optimizer):
+        raise ValueError(
+            "Optimizer has already been initialized, please make sure to only call deepspeed.initialize on an optimizer once."
+        )
+    if lr_scheduler is not None and _is_initialized(lr_scheduler):
+        raise ValueError(
+            "LR scheduler has already been initialized, please make sure to only call deepspeed.initialize on an LR scheduler once."
+        )
+
+
+def _mark_trainobjs_initialized(model: torch.nn.Module, optimizer: Optional[Optimizer],
+                                lr_scheduler: Optional[_LRScheduler]):
+    """Mark the model, optimizer, and lr_scheduler as initialized."""
+    _mark_initialized(model)
+    if optimizer is not None:
+        _mark_initialized(optimizer)
+    if lr_scheduler is not None:
+        _mark_initialized(lr_scheduler)
+
+
 def initialize(args=None,
                model: torch.nn.Module = None,
                optimizer: Optional[Union[Optimizer, DeepSpeedOptimizerCallable]] = None,
@@ -137,6 +181,10 @@ def initialize(args=None,
     zero.partition_parameters.shutdown_init_context()
 
     assert model is not None, "deepspeed.initialize requires a model"
+    # enforce that model, optimizer, and lr_scheduler have not been used in a previous deepspeed.initialize call
+    _assert_trainobjs_not_inited(model, optimizer, lr_scheduler)
+    # mark model, optimizer, and lr_scheduler as initialized
+    _mark_trainobjs_initialized(model, optimizer, lr_scheduler)
 
     global dist
     from deepspeed import comm as dist
@@ -220,6 +268,9 @@ def initialize(args=None,
 
     # Restore zero.Init context if necessary
     zero.partition_parameters.restore_init_context()
+
+    # mark engine, optimizer, and lr_scheduler as initialized
+    _mark_trainobjs_initialized(engine, engine.optimizer, engine.lr_scheduler)
 
     return_items = [
         engine,
--- a/tests/unit/runtime/test_ds_initialize.py
+++ b/tests/unit/runtime/test_ds_initialize.py
@@ -21,6 +21,7 @@ from deepspeed.runtime.utils import see_memory_usage
 from deepspeed.utils.torch import required_torch_version
 from deepspeed.accelerator import get_accelerator
 from deepspeed.ops.op_builder import FusedAdamBuilder
+from deepspeed import _assert_trainobjs_not_inited, _is_initialized
 
 
 @pytest.mark.parametrize('zero_stage', [0, 3])
@@ -434,3 +435,41 @@ class TestClientLrSchedulerInit(DistributedTest):
         else:
             # callable
             assert isinstance(ds_lr_scheduler, OneCycleLR)
+
+
+# https://github.com/microsoft/DeepSpeed/issues/6770
+class TestNoRepeatedInitializationAllowed(DistributedTest):
+    world_size = 1
+
+    def test_no_repeated_init(self):
+        hidden_dim = 10
+        model = SimpleModel(hidden_dim)
+        client_optimizer = torch.optim.Adam(model.parameters(), lr=0.01)
+
+        model = SimpleModel()
+        # Initialize DeepSpeed configurations for fp16
+        config_dict = {'train_batch_size': 1}
+
+        client_optimizer = torch.optim.Adam(filter(lambda p: p.requires_grad, model.parameters()), lr=1e-3)
+        # Initialize DeepSpeed engine
+        _assert_trainobjs_not_inited(model=model, optimizer=client_optimizer, lr_scheduler=None)
+        model_engine, optim, dataloader, scheduler = deepspeed.initialize(model=model,
+                                                                          optimizer=client_optimizer,
+                                                                          config_params=config_dict)
+
+        # arguments should be marked as initialized now
+        assert _is_initialized(model), "Client model should be marked as initialized"
+        assert _is_initialized(client_optimizer), "Client optimizer should be marked as initialized"
+
+        # return values should also be marked as initialized
+        assert _is_initialized(model_engine), "Model engine should be marked as initialized"
+        assert _is_initialized(optim), "Optimizer should be marked as initialized"
+        assert _is_initialized(scheduler), "Scheduler should be marked as initialized"
+
+        exception_raised = False
+        try:
+            deepspeed.initialize(model=model, optimizer=client_optimizer, config_params=config_dict)
+        except ValueError:
+            exception_raised = True
+
+        assert exception_raised, "Repeated initialization should raise an exception"


traincheck-team:

traincheck-team:
This fix still has interference with existing unit tests. Let me double check before we proceed.

traincheck-team:
The unit tests in `tests/unit/runtime/test_ds_initialize.py` all passed. 
I am not able to check other unit tests due to GPU memory constraint.

deepspeed/__init__.py
```diff
@@ -66,6 +66,50 @@ def _parse_version(version_str):
...
+def _mark_initialized(trainobj: Union[torch.nn.Module, Optimizer, _LRScheduler]):
+    """Mark a trainobj as initialized by setting the ds_is_inited attribute to True."""
+    if hasattr(trainobj, 'ds_is_inited'):
+        assert trainobj.ds_is_inited, "Not expecting the training object has `ds_is_inited` to be False if it exists, make sure you didn't set it to False or called deepspeed.initialize on the model more than once."
+        return
+
+    trainobj.ds_is_inited = True
+
+
+def _is_initialized(trainobj: Union[torch.nn.Module, Optimizer, _LRScheduler]):
```
tjruwase, line 78: Make name a bit deepspeed-specific to avoid collisions
```suggestion
def _is_ds_initialized(trainobj: Union[torch.nn.Module, Optimizer, _LRScheduler]):
```

deepspeed/__init__.py
```diff
@@ -66,6 +66,50 @@ def _parse_version(version_str):
...
+    """Check if a trainobj has been initialized by checking the ds_is_inited attribute."""
+    if hasattr(trainobj, 'ds_is_inited'):
+        # we shouldn't hit the assert below, but just in case
+        assert trainobj.ds_is_inited, "Not expecting the training object has `ds_is_inited` to be False if it exists, make sure you didn't set it to False or called deepspeed.initialize on the model more than once."
```
tjruwase, line 82: It seems odd for a boolean function to also assert or fail. Also, shouldn't function be something like 

```python
  return getattr(trainobj, 'ds_is_inited', False)
```

deepspeed/__init__.py
```diff
@@ -66,6 +66,50 @@ def _parse_version(version_str):
...
+    return False
+
+
+def _assert_trainobjs_not_inited(model: torch.nn.Module, optimizer: Optional[Optimizer],
+                                 lr_scheduler: Optional[_LRScheduler]):
+    """Enforce the model, optimizer, and lr_scheduler have not been used in a previous deepspeed.initialize call."""
+    if _is_initialized(model):
+        raise ValueError(
+            "Model has already been initialized, please make sure to only call deepspeed.initialize on a model once.")
+    if optimizer is not None and _is_initialized(optimizer):
```
tjruwase, line 93: Note that optimizer could be a Callable, not an object
https://github.com/microsoft/DeepSpeed/blob/4cd1d97460b677563d57f07a293724bdc02e0ef5/deepspeed/__init__.py#L71

deepspeed/__init__.py
```diff
@@ -66,6 +66,50 @@ def _parse_version(version_str):
...
+        raise ValueError(
+            "Optimizer has already been initialized, please make sure to only call deepspeed.initialize on an optimizer once."
+        )
+    if lr_scheduler is not None and _is_initialized(lr_scheduler):
```
tjruwase, line 97: Ditto for lr_scheduler
https://github.com/microsoft/DeepSpeed/blob/4cd1d97460b677563d57f07a293724bdc02e0ef5/deepspeed/__init__.py#L74

deepspeed/__init__.py
```diff
@@ -137,6 +181,10 @@ def initialize(args=None,
     zero.partition_parameters.shutdown_init_context()
 
     assert model is not None, "deepspeed.initialize requires a model"
+    # enforce that model, optimizer, and lr_scheduler have not been used in a previous deepspeed.initialize call
+    _assert_trainobjs_not_inited(model, optimizer, lr_scheduler)
```
tjruwase, line 185: I think this call should be moved into `_mark_trainobjs_initialized()

traincheck-team:
Thanks for the review @tjruwase. 

- I added handling for callable types for optimizer and lr_scheduler. The handling is to only mark `is_ds_inited` for objects instead of callables, as the callables are not stateful and reuse should be allowed.

Regarding,
> I think this call should be moved into `_mark_trainobjs_initialized()`
I think `_assert_trainobjs_not_inited` should still be separated from `_mark_trainobjs_initialized` as `_mark_trainobjs_initialized` is also called before `deepspeed.initialize` return the wrapped model and optimizers. The wrapped models may already have `is_ds_inited` being `True` since in DeepSpeedEngine all model flags will be passed through on the wrapper.

tests/unit/runtime/test_ds_initialize.py
```diff
@@ -434,3 +435,71 @@ def _lr_scheduler_callable(optimizer) -> _LRScheduler:
...
+        # arguments should be marked as initialized now
+        assert _is_ds_initialized(model), "Client model should be marked as initialized"
+        if optimizer_type is Optimizer:
+            assert _is_ds_initialized(client_optimizer), "Client optimizer should be marked as initialized"
+
+        # return values should also be marked as initialized
+        assert _is_ds_initialized(model_engine), "Model engine should be marked as initialized"
+        assert _is_ds_initialized(optim), "Optimizer should be marked as initialized"
+
+        exception_raised = False
```
tjruwase, line 476: Please use pytest [feature](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions) for expected assertions instead of `try...except`.

tjruwase:
> If we still want to keep `_assert_trainobjs_not_inited` inside `_mark_trainobjs_initialized`, we can do either of the three:

Thanks for raising this important design issue. Below are my thoughts and/or clarifications. 

1. DeepSpeed is designed to optimize PyTorch, and currently does this by wrapping PyTorch constructs, e.g., [nn.Module](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module), [optim.Optimizer](https://pytorch.org/docs/stable/optim.html#torch.optim.Optimizer), [optim.lr_scheduler](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.LRScheduler.html#torch.optim.lr_scheduler.LRScheduler), etc. 
2. To avoid nesting issues, DeepSpeed should not wrap DeepSpeed data structures, such as [DeepSpeedEngine](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/engine.py), [DeepSpeedOptimizer](https://github.com/microsoft/DeepSpeed/blob/cc03c76d57f41752d8cfb84c2e45b8e0da8083da/deepspeed/runtime/base_optimizer.py#L14), [DeepSpeedDataLoader](https://github.com/microsoft/DeepSpeed/blob/cc03c76d57f41752d8cfb84c2e45b8e0da8083da/deepspeed/runtime/dataloader.py#L41), etc.


Based on the above, I am aligned with your 3rd suggestion (below) of leveraging type information to simplify this PR.  
> 3\. check whether inited use type information as well, i.e. for types of DeepSpeedEngine we directly return `inited == True` instead of checking for flags.

What do you think?

traincheck-team:
Got it. Yeah 3 is technically what should be implemented instead of letting flags just flying around. Will get the update EOD.

traincheck-team:
@tjruwase Thanks for the suggestion. Please check the updated PR.

traincheck-team:
@tjruwase Hi Olatunji, are there any further modifications that you'd like me to do?

loadams:
> @tjruwase Hi Olatunji, are there any further modifications that you'd like me to do?

@traincheck-team - could you look into the failures on the nv-torch-latest-v100 test?  We've stabilized the CI and these look to be real failures.

traincheck-team:
@loadams Thanks for getting back so quickly and I apologize for the late reply. 

I inspected the 9 tests that failed. It appears to me that these tests failed because they initialized deepspeed using the same model multiple times, which is the exact behavior that this PR is trying to forbid. The PR itself is not buggy.

For example:

1. in `tests/unit/checkpoint/test_other_optimizer.TestOtherOptimizerCheckpoint::test_checkpoint_unfused_optimizer` at https://github.com/microsoft/DeepSpeed/blob/c08e69f21238f15bfe0e3779170fefa2f75d4c7e/tests/unit/checkpoint/test_other_optimizer.py#L61-L69. Two `checkpoint_correctness_verification` were invoked with the same `models` argument.
2. in `tests/unit/runtime/half_precision/onebit/test_onebit.TestZeroOneAdamCheckpointing::test` at https://github.com/microsoft/DeepSpeed/blob/c08e69f21238f15bfe0e3779170fefa2f75d4c7e/tests/unit/runtime/half_precision/onebit/test_onebit.py#L589-L617. Two `deepspeed.initialize` were invoked with exactly the same parameter.

As far as I see there are two solutions:
1. **Modify the unit tests to use separate model copies.** But I don't think you guys may want to risk doing this, but I feel this is the "right" thing to do.
2. **Optionally suppress the exceptions by env or config vars". This might be an easier and more practical solution.

deepspeed/__init__.py
```diff
@@ -65,6 +67,45 @@ def _parse_version(version_str):
...
+def _ensure_and_mark_trainobjs_inited(
+    model: torch.nn.Module,
+    optimizer: Optional[Union[Optimizer, DeepSpeedOptimizerCallable]],
+    lr_scheduler: Optional[Union[_LRScheduler, DeepSpeedSchedulerCallable]],
+    ensures_not_inited: bool = False,
+):
+    trainobjs = {"model": model, "optimizer": optimizer, "lr_scheduler": lr_scheduler}
+
+    for name, trainobj in trainobjs.items():
+        print(f"Checking {name}")
```
tjruwase, line 96: Use [log_dist](https://github.com/microsoft/DeepSpeed/blob/7f3d669b40f8d29010efd9578d4a2cdd0f16b20e/deepspeed/utils/logging.py#L107) or remove

traincheck-team: Removed it.

deepspeed/__init__.py
```diff
@@ -65,6 +67,45 @@ def _parse_version(version_str):
...
+        if trainobj is None:
+            continue
+        if name in ("optimizer", "lr_scheduler") and not isinstance(trainobj, (Optimizer, _LRScheduler)):
+            # skipping DeepSpeedOptimizerCallable and DeepSpeedSchedulerCallable
+            continue
```
tjruwase, line 100: Why not use the following simplification?

```suggestion
        if  isinstance(trainobj, (DeepSpeedOptimizerCallable , DeepSpeedSchedulerCallable)):
            continue
```

deepspeed/__init__.py
```diff
@@ -221,6 +263,9 @@ def initialize(args=None,
     # Restore zero.Init context if necessary
     zero.partition_parameters.restore_init_context()
 
+    # mark engine, optimizer, and lr_scheduler as initialized
+    _ensure_and_mark_trainobjs_inited(engine, engine.optimizer, engine.lr_scheduler, ensures_not_inited=False)
```
tjruwase, line 267: Since deepspeed objects are not marked, won't this be a no-op?

tests/unit/runtime/test_ds_initialize.py
```diff
@@ -434,3 +435,52 @@ def _lr_scheduler_callable(optimizer) -> _LRScheduler:
...
             # callable
             assert isinstance(ds_lr_scheduler, OneCycleLR)
+
+
+# https://github.com/microsoft/DeepSpeed/issues/6770
+class TestNoRepeatedInitializationAllowed(DistributedTest):
+    world_size = 1
+
+    @pytest.mark.parametrize('optimizer_type', [None, Optimizer, Callable])
+    def test(self, optimizer_type):
```
tjruwase, line 445: It would be better to split this unit test into two:
1. Validate correct marking (lines 466 -- 473)
2. Detect nested initialization (lines 475 -- 486)

deepspeed/__init__.py
traincheck-team: I've tried the suggested line previously, but isinstance checks does not work for fancy type annotations for callables. Using `if callable(trainobj)` here is also not ideal, as instances of `Optimizer` and `_LRScheduler` have a `__call__` method defined.

traincheck-team: >>> isinstance(print, DeepSpeedSchedulerCallable)
  File "<stdin>", line 1, in <module>
  File "/home/xxx/miniconda3/envs/env_test_ds_bug/lib/python3.10/typing.py", line 994, in __instancecheck__
    return self.__subclasscheck__(type(obj))
  File "/home/xxx/miniconda3/envs/env_test_ds_bug/lib/python3.10/typing.py", line 997, in __subclasscheck__
    raise TypeError("Subscripted generics cannot be used with"
TypeError: Subscripted generics cannot be used with class and instance checks

deepspeed/__init__.py
traincheck-team: `engine.optimizer` and `engline.lr_scheduler` are not necessarily deepspeed objects in my tests. Also, since optimizer and lr_scheuler are optional args to `deepspeed.initialize`, in cases where a caller does not provide an optimizer or scheduler, it is also necessary to mark objects as initialized here.

After removing this line, test[None] and test[Callable] will fail too.

deepspeed/__init__.py
tjruwase: Got it. Thanks for trying the suggestion. No need for change.

deepspeed/__init__.py
tjruwase: Got it. Thanks!

tests/unit/runtime/test_ds_initialize.py
traincheck-team: Good suggestion!

tjruwase (approved):

loadams:
FYI @traincheck-team - there still appear to be some unit test errors if you could investigate.

traincheck-team:
Thanks for the heads up. Will do that soon.

traincheck-team:
@loadams I tried to amend the six problematic commits using git rebase, but since there were previous merges from main into this branch, it caused conflicts and made things messy (e.g. see the requested reviewers of this PR as code owners).

	To fix this, I plan to create a new local branch from main, reapply the necessary changes, and then force-push it to replace this branch. This should give us a clean history while preserving the required commits.

	Let me know if you have any concerns or if there’s a better way to proceed!

traincheck-team:
@loadams Hi Logan, I apologize for the late reply. I’ve reviewed the 9 unit test failures in the recent workflow run: https://github.com/deepspeedai/DeepSpeed/actions/runs/13205140637/job/36866442471. 

My understanding is that these failures are caused by repeated model initialization, which this PR now detects and raises exceptions for.

## Evidences:
- Inspecting the affected locations confirms that the exception is raised precisely where repeated initialization occurs.

## Impacted Tests:
These failed unit tests span 3 files 
- `test_onebit.py:249`, `test_onebit.py:616`, `test_onebit.py:997`, 
- `test_other_optimizer.py:68`, `test_other_optimizer.py:107`  and 
- `test_zero_optimizer.py:365`.

## Suggested Fix:
The solution is straightforward and manageable as—ensure a new model object is created before the second initialization, which does not require modification to any critical logic. I suggest handling this in a separate PR to fix these tests.

If there are any additional failures beyond these 9, please let me know.

PR #6881, opened 2024-12-17
Fix `checkpointable_layers` Logic

Quentin-Anthony:
**Problem**

There's an edge-case in DeepSpeed, where if all three of the following are true: 
1. Deepspeed activation checkpointing is applied 
2. The user passes `checkpointable_layers` (e.g. https://github.com/EleutherAI/gpt-neox/blob/f5325805678c2b9e35aae4528283e0132c5f5bbc/megatron/model/gpt2_model.py#L175)
3. The user's model class contains `GPT2ModelPipe` or GPTModelPipe`

Then the `checkpointable_layers` will not be activation checkpointed. 

**Reason**

This is because in the current logic, `_is_checkpointable` will short-circuit to just return layers matching `ParallelTransformerLayerPipe` in the case of `self.__class__.__name__ in ('GPTModelPipe', 'GPT2ModelPipe')`. See https://github.com/microsoft/DeepSpeed/blob/da771ed42e41a44d5047813ca4672f1cfe9d1731/deepspeed/runtime/pipe/module.py#L653

**Proposed Fixes**

I think that `checkpointable_layers` should always be checked for, and added logic to this effect. I also found the documentation for `checkpointable_layers` confusing and contradictory, so I updated the docstring. Lastly, I added a unit test for `checkpointable_layers`.


tjruwase (approved):

tests/unit/runtime/activation_checkpointing/test_activation_checkpointing.py
```diff
@@ -259,3 +260,52 @@ def test_ckpt_non_tensor_output_ordering(self, non_tensor_output):
         else:
             ordering += [torch.is_tensor(non_tensor_output)]
         _test_activation_checkpoint_ordering(module, ordering, inputs)
+
+
+class TestCheckpointableLayersConfig(DistributedTest):
```
loadams, line 265: Thanks for the PR/test @Quentin-Anthony - looks like the added test is failing, could you take a look?

Quentin-Anthony: Yep on it

Quentin-Anthony: Fixed in https://github.com/microsoft/DeepSpeed/pull/6881/commits/b21217739e1d5c21e3572c6ba0ed3b19c53c82d1. Was a silly mistake where I forgot to specify `num_dp=1` for `PipeModelDataParallelTopology`. Test is passing on my end.

PR #6885, opened 2024-12-17
Fix `no-torch` workflow and update real_accelerator

loadams:
This reverts #6845, since this breaks the `no-torch` workflow that we require in order to do releases where we do not require torch to be in the environment when building an sdist.

FYI @keiwoo (we want to make the change, but this will revert and at least trigger this test when we re-apply these changes.


loadams:
Closing in favor of #6886

PR #6886, opened 2024-12-18
Don't error out when cpu accelerator doesn't have torch (as default for whl building)

loadams:
This fixes a bug introduced in #6845, which breaks the `no-torch` workflow that we require in order to do releases where we do not require torch to be in the environment when building an sdist.  This adds the same logic to the cpuaccelerator that the cudaaccelerator had where we don't require torch to be installed to build the whl.


sfc-gh-mwyatt (commented):
LGTM

PR #6909, opened 2024-12-24
[BUG FIX]:fix get torch.version.cuda error when cuda is None in rocm

hj-wei:
HI, I found some error when using deepspeed with rocm-torch
```
torch_cuda_version = ".".join(torch.version.cuda.split('.')[:2]) 
```
will raise an AttributeError when torch.version.cuda is None. This occurs because the CUDA version in rocm-torch/version.py is set to always be None, leading to potential runtime errors in environments where ROCm is being used.

### DIFF
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -415,10 +415,11 @@ class OpBuilder(ABC):
             return '-mcpu=native'
         return '-march=native'
 
-    def is_cuda_enable(self):
+    def get_cuda_compile_flag(self):
         try:
-            assert_no_cuda_mismatch(self.name)
-            return '-D__ENABLE_CUDA__'
+            if not self.is_rocm_pytorch():
+                assert_no_cuda_mismatch(self.name)
+                return "-D__ENABLE_CUDA__"
         except MissingCUDAException:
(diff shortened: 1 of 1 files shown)


op_builder/builder.py
```diff
@@ -839,7 +839,9 @@ def cxx_args(self):
 
         CPU_ARCH = self.cpu_arch()
         SIMD_WIDTH = self.simd_width()
-        CUDA_ENABLE = self.is_cuda_enable()
+        CUDA_ENABLE = (
+            "-D__DISABLE_CUDA__" if self.is_rocm_pytorch() else self.is_cuda_enable()
```
tjruwase, line 843: Thanks for this PR. Can you please help with the following changes to improve the codebase beyond your fix?

1. Rename the method `is_cuda_enable()` -> to `get_cuda_compile_flag()`, which is more meaningful. 
2. Handle `self.is_rocm_pytorch()` case in `get_cuda_compile_flag()`, returning `-D__DISABLE_CUDA__` as appropriate. 

Are you able to help these changes? Thanks!

hj-wei: fixed

loadams (approved):

PR #6922, opened 2025-01-02
Autotp training

inkcherry:
FYI @tjruwase  @GuanhuaWang @delock @skyshine102   context: https://github.com/microsoft/DeepSpeed/pull/5445
**changes/support**
- auto tensor parallel training for HF model(zero compatible. I only tested zero1 currently)
- distributed ckpt save(UCP is not supported).
- HF model files save(set ```gather_16bit_weights_on_model_save=True``` in ds config).
- Dataloader check.
- Uts.
- tp layer refactor by abstract layer design.

**HF trainer dependency:**
   transformer: https://github.com/inkcherry/transformers/tree/ds_tp
   accelerate: https://github.com/inkcherry/accelerate/tree/ds_tp
   I could send them once ds support these api.

**Usage:**
Users do not need to modify the client code, they only need to configure the settings in the config file to achieve the desired functionality.
Below is an example of code for fine-tuning a LLaMA 2 model (SFT). It supports Zero3/FSDP training and enables TP training by simply adjusting the configuration

https://github.com/inkcherry/stanford_alpaca/commits/tp_demo_1127/
This branch contains three commits, with the last two commits added for quick experiments and logging purposes.
**results**
loss curve（gbs=16）:
zero3(baseline)
image
tp(this)
image

zero1 with zero1+tp(zero compatible)
image

performance（For your reference only.）:
zero3(not enabled any acceleration.) : 18GB     2.3s/it
zero1：38GB      1.30s/it          
zero1+tp: 24GB   1.66s/it
**extension:**
I think async-TP/domino .etc.  can be implemented by inheriting a class and overriding the fwd/bwd methods. The logic for gather/partition can be reused to achieve this.(please correct me if I am wrong)

Complex sharding can also be achieved through independent partitioning and gathering. Partitioning is mandatory, while gathering is required for training.
**TODO:**
embedding vocab parallel
Currently, the parallelism for embeddings is primarily based on hidden_dim parallel combined with allreduce. This approach takes advantage of efficient reduction kernels. and it is not forced to use. 
In training, however, the more common method is vocab parallelism. Enabling by default can save a certain amount of GPU memory.

thanks for @delock guidance.
I also verified inference with cpu-inference workloads(Optimized Model List in https://github.com/intel/intel-extension-for-pytorch/tree/main).
many thanks for @xuguangxin @ikurtchen  @rogerxfeng8 ,@Yejing-Lai ,@ys950902 .etc. Help review and address matters related to inference.


delock:
@tjruwase @GuanhuaWang 
We had internal review of @inkcherry 's PR.  This PR allows train HF models with tensor parallel without need for megatron.  Which is very friendly to user.

Let us know your plan for Domino integration.  @inkcherry 's memory data looks good.  With Domino we think it can have less impact on performance since TP communication can overlap with computation.

@inkcherry by design should autotp training work with ZeRO3 as well?

inkcherry:
> @tjruwase @GuanhuaWang We had internal review of @inkcherry 's PR. This PR allows train HF models with tensor parallel without need for megatron. Which is very friendly to user.
> 
> Let us know your plan for Domino integration. @inkcherry 's memory data looks good. With Domino we think it can have less impact on performance since TP communication can overlap with computation.
> 
> @inkcherry by design should autotp training work with ZeRO3 as well?

for Zero3 + TP: Currently, the logic to combine the saving of HF weights for TP & DP has not been implemented, but it is entirely feasible. If needed, it can be implemented in the future.

tjruwase:
> We had internal review of @inkcherry 's PR. This PR allows train HF models with tensor parallel without need for megatron. Which is very friendly to user.

Bravo @inkcherry, this is an excellent technology and massive usability benefit for users. This is really exciting! 

> Let us know your plan for Domino integration. @inkcherry 's memory data looks good. With Domino we think it can have less impact on performance since TP communication can overlap with computation.

In terms of Domino integration, @GuanhuaWang will take the lead on that. 

> for Zero3 + TP: Currently, the logic to combine the saving of HF weights for TP & DP has not been implemented, but it is entirely feasible. If needed, it can be implemented in the future.

I would love to prioritize enabling UCP support sooner than later. @inkcherry, can you please share the work needed here?

tests/unit/model_parallelism/test_autotp_training.py
```diff
@@ -0,0 +1,534 @@
...
+        else:
+            raise e
+
+
+class TestTpParallelStates(DistributedTest):
+    world_size = 4
+
+    def test(self):
+        set_autotp_mode(training=True)
+        tp_size = 4
```
tjruwase, line 60: Can you parametrize `tp_size` to improve coverage?

deepspeed/runtime/zero/config.py
```diff
@@ -339,7 +340,10 @@ class DeepSpeedZeroConfig(DeepSpeedConfigModel):
     """
     Override nn.Module apply function, for Stage 3.
     """
-
+    autotp_size: int = Field(0, ge=0, new_param="autotp_size")
```
tjruwase, line 343: Why is `autotp_size` defined as a subfield of zero, instead of a top-level field in ds_config? Is there a dependency on zero logic?

deepspeed/runtime/engine.py
```diff
@@ -3571,6 +3644,49 @@ def _save_zero_checkpoint(self, save_path, tag):
...
+        """
+        Get a full non-partitioned state_dict with fp16 weights on cpu.
+        Important: this function must be called on all ranks and not just rank 0.
+        This is similar to nn.Module.state_dict (modelled after _save_to_state_dict)
+        This method is used for tensor parallel training.
+
+        Returns:
+        OrderedDict: The consolidated state dictionary if the current process rank is 0, otherwise None.
+        """
+        #TODO: If we use both Zero3 and tensor parallel simultaneously
```
tjruwase, line 3657: Can you clarify what is meant by the gather mechanism of tensor parallelism?

deepspeed/module_inject/auto_tp.py
```diff
@@ -339,77 +349,37 @@ def _replace(self, child, name, conv_linear_layer):
...
+
             elif 'o_proj' in name:
-                weight, bias = shard_value_with_share_qk(child.weight.data, child.bias, dist.get_rank(),
-                                                         dist.get_world_size(), False)
-                return LinearAllreduce(weight, bias, self.mp_group)
-        # For Arctic model, bypass to all_reduce replacement for w2 weights
+                return Yuan_LinearALlreduce(child, self.mp_group)
+
+        # For MLP including chunk layer.
+        if 'gate_up_proj' in name or ('dense_h_to_4h' in name and 'GLM' in str(self.module)):
```
skyshine102, line 358: This additional code block is trying to deal with "MLP including chunk layer" (general case), but the returned module/object is in the name of GLM prefix. 
It could be better to rename the `GLM_LinearLayer` to sth like `GateUpPack_LinearLayer`.

deepspeed/module_inject/auto_tp.py
```diff
@@ -11,10 +11,12 @@
 from typing import Optional
 import torch
 from deepspeed import comm as dist
-from .layers import LinearAllreduce, LinearLayer, LmHeadLinearAllreduce
+from .layers import LinearAllreduce, LinearLayer, LmHeadLinearAllreduce, Yuan_LinearALlreduce, Yuan_LinearLayer, GLM_LinearLayer, Conv_LinearALlreduce, fused_LinearLayer, conv_LinearLayer
```
skyshine102, line 14: Original coding style is `LinearAllreduce` instead of `LinearALlreduce`.

deepspeed/runtime/engine.py
```diff
@@ -411,6 +413,74 @@ def _optimized_linear_offload_setup(self):
...
+                                     device=get_accelerator().current_device())
+                    ), "Data inconsistency within the TP group. Please check the Dataloader implementation to ensure consistency."
+
+            bcast_rank = self.mpu.get_tensor_model_parallel_src_rank()
+            bcast_group = self.mpu.get_tensor_model_parallel_group()
+
+            broadcast_and_check(args, bcast_rank, bcast_group)
+            broadcast_and_check(kwargs, bcast_rank, bcast_group)
+
+            print(f"RANK[{dist.get_rank()}]:The Dataloader has passed the TP group consistency check.")
```
skyshine102, line 476: maybe use the logger instead of print at rank 0.

deepspeed/runtime/engine.py
skyshine102: Same question. 
I could somehow understand as it's a similar function to https://github.com/microsoft/DeepSpeed/blob/f2cc80909b8bdd399a029acf3ed0c894279974b6/deepspeed/runtime/engine.py#L3574 but specific for TP. The function name can be improved.

skyshine102 (commented):
Thanks @inkcherry for this contribution. I have spent some time to read this PR and I'm happy to be involved in this discussion. (I'm not from deepspeed team but deepspeed user. My comments are relatively minor though.)

skyshine102:
> @tjruwase @GuanhuaWang We had internal review of @inkcherry 's PR. This PR allows train HF models with tensor parallel without need for megatron. Which is very friendly to user.
> 
> Let us know your plan for Domino integration. @inkcherry 's memory data looks good. With Domino we think it can have less impact on performance since TP communication can overlap with computation.
> 
> @inkcherry by design should autotp training work with ZeRO3 as well?

I have the same question. Does this PR support the flow like https://pytorch.org/tutorials/intermediate/TP_tutorial.html#combine-tensor-parallel-with-fully-sharded-data-parallel-together ?  (TP to shared weight $W$ to $W_{tp_i}$, then further shard $W_{tp_i}$ by ZeRO-3 to $W_{tp_i, dp_j}$)

deepspeed/runtime/engine.py
tjruwase: @skyshine102, thanks for the comment. A key difference between is zero3 and TP is that partitioned zero3 modules materialized using allgather before compute, whereas TP modules compute in a partitioned manner. So, it is unclear to me what requires gathering for TP.

deepspeed/inference/config.py
```diff
@@ -31,6 +31,11 @@ class MoETypeEnum(str, Enum):
     standard = "standard"
 
 
+class AUTOTP_MODE(Enum):
```
tjruwase, line 34: This logic should ideally be outside of the inference module. For example, in deepspeed/runtime/tensor_parallel module?

inkcherry: Yes, it is currently not invoked during the calculation. In my opinion, the ultimate goal of training such models is to save HF format weights for inference, primarily focusing on saving the original HF model (not TP-partitioned). Additionally, it provides the flexibility for customized complex sharding to enable specifically designed gathers in the TP layer

deepspeed/runtime/engine.py
inkcherry: Yes, it is currently not invoked during the calculation. 
In my opinion, the ultimate goal of training such models is to save HF format weights for inference, primarily focusing on saving the original HF model (not TP-partitioned). Additionally, it provides the flexibility for customized complex sharding to enable specifically designed gathers in the TP layer

skyshine102: The tradition of TP is saving a checkpoint for each rank (related: universal format from ds team) while HF format is usually consolidated so a function is needed.

tests/unit/model_parallelism/test_autotp_training.py
```diff
@@ -0,0 +1,534 @@
...
+    torch_linear.zero_grad()
+    return torch_linear, torch_out, torch_norm
+
+
+class TestTpLayerFwdBwd(DistributedTest):
+    world_size = 4
+    reuse_dist_env = True
+
+    def testRowParallel(self):
+        tp_size = 4
```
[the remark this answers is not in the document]
inkcherry: fixed

deepspeed/inference/engine.py
```diff
@@ -247,6 +248,11 @@ def _post_forward_hook(self, module, input, output):
         self._model_times.append(elapsed_time)
 
     def _create_model_parallel_group(self, config):
+
+        if is_autotp_training_mode():
```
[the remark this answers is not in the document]
inkcherry: yes, it seems there was no suitable file currently, so I created a runtime/tensor_parallel folder now~

deepspeed/module_inject/auto_tp.py
inkcherry: thanks for the comments, modified：）

deepspeed/module_inject/auto_tp.py
inkcherry: fixed

deepspeed/inference/config.py
inkcherry: yes, moved to runtime/tensor_parallel folder

deepspeed/runtime/engine.py
inkcherry: yes, modified.

tests/unit/model_parallelism/test_autotp_training.py
```diff
@@ -0,0 +1,558 @@
...
+    return model, base_model
+
+
+class TestSave(DistributedTest):
+
+    world_size = 4
+    reuse_dist_env = True
+
+    def test_save_original_weight(self):
+        tp_size = 4
```
GuanhuaWang, line 360: same here, could we parameterize both tp_size 2 and 4?

deepspeed/utils/groups.py
```diff
@@ -64,6 +62,127 @@ def _ensure_divisibility(numerator, denominator):
...
+
+    global _DATA_PARALLEL_GROUP
+    global _MODEL_PARALLEL_GROUP
+    global _TENSOR_MODEL_PARALLEL_GROUP
+
+    if _TENSOR_MODEL_PARALLEL_GROUP is not None:
+        return
+
+    if data_parallel_size is None:
+        data_parallel_size = dist.get_world_size() // tensor_model_parallel_size
```
GuanhuaWang, line 91: do we need to consider pipeline_parallel_size?

deepspeed/runtime/utils.py
```diff
@@ -1101,3 +1101,46 @@ def move_back_key(state, key):
...
+            elif val1 != val2:
+                return False
+        return True
+
+    elif isinstance(inputs1, dict) and isinstance(inputs2, dict):
+        if inputs1.keys() != inputs2.keys():
+            return False
+        for key in inputs1:
+            val1 = inputs1[key].to(get_accelerator().current_device())
+            val2 = inputs2[key].to(get_accelerator().current_device())
```
GuanhuaWang, line 1138: do we really need this `compare_tensors_in_structures` function? there are plenty of easier ways to compare lists or dicts are equal. One concern is here it did a lot small memcpy as `val.to(device)` which is inefficient.

deepspeed/runtime/tensor_parallel/tp_manager.py
```diff
@@ -0,0 +1,67 @@
...
+        """Apply the given injection policy to a client module."""
+        if isinstance(self.module, torch.nn.Module):
+            replace_transformer_layer(client_module, self.module, None, self.config, self.model_config)
+
+    def _initialize_tp_config(self, tp_size):
+        """Perform TP configuration initialization."""
+        self.tp_config = TPConfig()
+        self.tp_config.tp_size = tp_size
+        if tp_size <= 1:
+            self.tp_config.enabled = False
```
GuanhuaWang, line 60: I don't see anywhere this flag is used (i.e. there seems no design/code if enabled flag == False)? is this needed?

deepspeed/runtime/engine.py
GuanhuaWang: I also don't see why need to gather weights/params in TP training/inference. If it is only used for re-collecting weights for single point checkpoint write, then you can use our universal checkpoint feature to convert model parallel strategy after training.

deepspeed/runtime/tensor_parallel/config.py
```diff
@@ -0,0 +1,64 @@
...
+class TPConfig(DeepSpeedConfigModel):
+    """ Configure tensor parallelism settings """
+
+    enabled: bool = True
+    """ Turn tensor parallelism on/off. """
+
+    tp_size: int = 1
+    """ Number of devices to split the model across using tensor parallelism. """
+
+    tp_grain_size: int = 64
```
GuanhuaWang, line 27: this argument I also did not see any use case

deepspeed/module_inject/layers.py
```diff
@@ -7,10 +7,570 @@
...
+
+class Yuan_LinearLayer(LinearLayer):
+    #Yuan2
+    @torch.no_grad()
+    def partition(self, params_list):
+        weight, bias = shard_value_with_share_qk(params_list[0].data, params_list[1], self.tp_index,
+                                                 self.tp_world_size, True)
+        params_list[0].data = move(weight, get_accelerator().current_device_name()).detach()
+        if bias is not None:
+            params_list[1].data = move(bias, get_accelerator().current_device_name()).detach()
```
GuanhuaWang, line 517: is it possible to make an abstraction of `partition` method with arguments passed-in for different models? if doing this, we can avoid create 2 new classes (e.g., Yuan_linear & Yuan_linear+allreduce) for every new model structure.

deepspeed/module_inject/layers.py
```diff
@@ -7,10 +7,570 @@
...
+            if key == 'mp_group':
+                new_obj.mp_group = self.mp_group
+            else:
+                setattr(new_obj, key, deepcopy(value, memo))
+
+        memo[id(self)] = new_obj
+        return new_obj
+
+
+class GatherReplacedLayerParams:
```
GuanhuaWang, line 230: do we need gather TP params during training or inference?

GuanhuaWang (commented):
Hi Sorry for the delay. I just left some comments. Thanks

deepspeed/utils/groups.py
inkcherry: Currently, this feature does not support the pipeline and pipeline-related logic will not reach this part. Perhaps we can consider adding pipeline support in the future.

deepspeed/runtime/utils.py
inkcherry: Yes, you are it will affect performance. however, this check only occurs during the first iterations, and the check hook will remove itself afterward. This helps users avoid some basic mistakes. otherwise, even if an error occurs, it won't cause a crash but might increase the burden of troubleshooting.

deepspeed/module_inject/layers.py
```diff
@@ -7,10 +7,570 @@
...
+        """
+        Forward pass.
+        """
+        ctx.group = group
+        if group == None:
+            return input
+        if is_inference_mode:
+            dist.inference_all_reduce(input, group=group)
+        else:
+            dist.all_reduce(input.contiguous(), group=group)
```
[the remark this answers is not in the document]
inkcherry: It seems that adding this makes it safer, potentially helping to avoid discontinuity introduced by ```transpose/permute```.
FYI: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/tensor_parallel/mappings.py#L23

I am not very clear on the implementation detail of ```inference_all_reduce```, so I have kept the original ```dist.inference_all_reduce``` code path.

deepspeed/module_inject/layers.py
```diff
@@ -7,10 +7,570 @@
...
+
+class ColumnParallel(torch.autograd.Function):
+    """
+    Custom autograd function for column-wise parallelism.
+    """
+
+    @staticmethod
+    def symbolic(graph, input):
+        """Symbolic function for tracing."""
+        return dist.all_reduce(input.contiguous(), dist.get_tensor_model_parallel_group())
```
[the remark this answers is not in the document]
inkcherry: It is consistent with the previous situation.

deepspeed/runtime/tensor_parallel/tp_manager.py
inkcherry: Thank you for pointing that out. It's not necessary, I was referring to the inference config. I have removed it now.

deepspeed/runtime/tensor_parallel/config.py
inkcherry: The variable is used in the autoTP parser to set tile boundaries to accelerate GEMM.  https://github.com/microsoft/DeepSpeed/blob/05eaf3d1cab0f42f130a153802c7b94d86ecc872/deepspeed/module_inject/replace_module.py#L308
it has not been activated in training yet, as it requires support for uneven gather. I have added clearer comments for better understanding.

tests/unit/model_parallelism/test_autotp_training.py
```diff
@@ -0,0 +1,558 @@
...
+
+
+@pytest.mark.sequential
+class TestParamsGather(DistributedTest):
+    world_size = 4
+    reuse_dist_env = True
+
+    @pytest.mark.parametrize("layer_type", ["linear", "linearallreduce"])
+    def test(self, layer_type):
+        tp_size = 4
```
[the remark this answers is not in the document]
inkcherry: Thanks for the reminder， added

deepspeed/runtime/engine.py
inkcherry: Thanks for your suggestion, UCP is a great feature for recovering training from parallel topology changes.

In this patch, my plan is to support two basic Huggingface save APIs:

```trainer.save_model()```, which does not rely on any checkpoints. After calling this (without post-processing), it saves in the HF format for other purposes. The interface is user-friendly.
```trainer.save_state()```, which is used for resuming training from checkpoints.
To implement the basic ```trainer.save_model()```, I referred to the current Zero3 strategy. For safety, it’s possible that ```save_model()``` will be called more than once, so it is implemented via context. This is similar to Zero3’s gather, but it will not be used in the forward and backward passes.

Regarding checkpoints, currently, it is non-UCP. The reasons are:

- I noticed that UCP still requires some local post-processing (according to the documentation).

- It has mainly been validated on Mega-ds, so I think integrating and validating it with HF-related code would require additional effort. Considering my bandwidth, I lean towards we my complete this enhancement in another PR.
Regarding the gather code, the logic for restoring shards (reverse of partition) can't be omitted (even as post-processing). Managing partition and gather via a class might be more convenient. I think we may  use a config switch to directly save in a universal format during runtime if we support it.

deepspeed/module_inject/layers.py
inkcherry: no, the reason are integrated into the comments above.

deepspeed/module_inject/layers.py
inkcherry: Yes, they currently only have one method. every specific shard logic should have a corresponding reverse gather logic. The current shard method hasn’t implemented the corresponding gather. I think  using a class might help reserve a potential placeholder and make the code more consistent.

inkcherry:
@tjruwase @GuanhuaWang  Thank you for your review. I’ve added modifications or explanations. Could you take another look? Thanks!

deepspeed/module_inject/layers.py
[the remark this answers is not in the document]
hwchen2017: If it is already contiguous, `.contiguous()` will not launch additional memory copy kernel. Or it won't hurt performance.

hwchen2017 (commented):
Hi @inkcherry, thanks for contributing. Just a heads-up, all the all_reduce call in domino is supposed to be asynchronous, and current `LinearAllreduce` and `LinearLayer` need to be updated to work with Domino. 
For example, in the `LinearAllreduce`, we'd like to get the handle from asynchronous all reduce, and synchronize it later to overlap computation. 
The Domino work is still in progress, and it's not finalized yet. So, you don't need to worry about the compatibility with Domino at this point. But one thing you can easily support is the async TP, similar to Megatron [here](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/tensor_parallel/layers.py#L487).  Maybe it can be your next PR. 
Thanks for your help!

deepspeed/runtime/engine.py
tjruwase: @inkcherry, thanks for your explanation. It is reasonable.

tjruwase (approved):

inkcherry:
hi, @tjruwase . I noticed the CI updated the DCO check. Using the suggested rebase method would reintroduce many conflicts https://github.com/deepspeedai/DeepSpeed/pull/6922/checks?check_run_id=36694512120, so I opted for a squash merge with sign-off instead.

PR #6932, opened 2025-01-08
generalize deepspeed linear and implement it for non cuda systems

oelayan7:


op_builder/fp_quantizer.py
```diff
@@ -9,6 +9,7 @@
     pkg_version = None
 
 from .builder import CUDAOpBuilder, installed_cuda_version
+import torch
```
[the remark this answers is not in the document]
oelayan7: @loadams I moved it inside the function, and imported it locally. 
When this function is called, there will always be torch.

oelayan7:
@loadams  I think the failure in cpu-torch isn't related to the changes, as it passed yesterday (in the first commit) 

Thanks in advance

loadams:
> @loadams I think the failure in cpu-torch isn't related to the changes, as it passed yesterday (in the first commit)
> 
> Thanks in advance

@oelayan7 - yes, I've not yet debugged this failure since it only occurs on some of the GitHub provided runners, but it is intermittent and I'll re-run.

op_builder/fp_quantizer.py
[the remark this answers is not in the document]
loadams: Thanks, that should work, and at least the test should catch it if we need to import torch for other functions there in the future.

oelayan7:
@loadams tests are passing now. Looks like it was a momentary issue

nelyahu:
@loadams Hi, can you please help review this PR? i see there was one round of review, are there any leftovers?

loadams:
> @loadams Hi, can you please help review this PR? i see there was one round of review, are there any leftovers?

Sorry for the delay @nelyahu and @oelayan7, working on getting the right set of eyes on this.  We will prioritize it this week.

deepspeed/ops/fp_quantizer/quantize.py
```diff
@@ -42,12 +42,18 @@ def dequantize(self, input_q, fp_out=None, q_bits=8, q_mantisa_bits=3, scale=Non
...
 class FP_Quantize(Quantizer):
 
-    def __init__(self, group_size=512) -> None:
+    def __init__(self, quantization_config) -> None:
         global fp_quant_module
-        super().__init__(group_size=group_size)
+        super().__init__(group_size=quantization_config.group_size)
         if fp_quant_module is None:
             fp_quant_module = FPQuantizerBuilder().load()
+        self.is_python_impl = getattr(fp_quant_module, "PYTHON_IMPL", False)
```
tjruwase, line 50: It seems the goal here is to handle `cuda` specially, so I think it is make that explicit with something like:

```suggestion
        self.is_cuda_impl = getattr(fp_quant_module, "CUDA_IMPL", False)
```

deepspeed/ops/fp_quantizer/quantize.py
```diff
@@ -73,15 +79,27 @@ def quantize(self,
...
-                                  dtype=torch.uint8,
-                                  device=input.device)
-        out = fp_quant_module.quantize(self.input_q, input, self.group_size, stochastic_mode, q_bits, q_mantisa_bits)
+
+        # Adding (group_size - 1) is for padding
+        self.num_groups = (input.numel() + self.q_config.group_size - 1) // self.q_config.group_size
+        # group_size should be the minimal number between the defined group size and number of elements in tensor.
+        group_size = int(min(self.q_config.group_size, input.numel()) * q_bits) // 8
+        # CUDA quantization kernel saves the scale as (fp32) inside the quantized tensor for each group
+        if not self.is_python_impl:
```
tjruwase, line 88: Can be make explicit as special handling for cuda
```suggestion
        if self.is_cuda_impl:
```

op_builder/hpu/fp_quantizer.py
```diff
@@ -0,0 +1,86 @@
...
+        import habana_frameworks.torch.utils.experimental as htexp
+        if htexp._get_device_type() == htexp.synDeviceType.synDeviceGaudi2:
+            dtype = torch.float8_e4m3fnuz
+        else:
+            dtype = torch.float8_e4m3fn
+        return torch.finfo(dtype).max
+
+
+class FPQuantizer:
+    PYTHON_IMPL = True
```
tjruwase, line 49: Why not
```suggestion
    HPU_IMPL = True
```

tests/unit/ops/fp_quantizer/test_fp8_gemm.py
```diff
@@ -25,7 +26,11 @@
 def test_fp_quant(dtype, q_bits, M):
     device_name = get_accelerator().device_name()
     quantization_group_size = 128
-    fpq = FP_Quantize(group_size=quantization_group_size)
+
+    quant_config = QuantizationConfig()
```
tjruwase, line 30: Will the following work?
```suggestion
    quant_config = QuantizationConfig(group_size = quantization_group_size, q_dtype = FPQuantizerBuilder.get_default_quant_dtype())
```

jeffra (changes requested):
This all looks good to me, thanks for this generalization! However, please address @tjruwase's suggestions.

tests/unit/ops/fp_quantizer/test_fp8_gemm.py
oelayan7: Done

op_builder/hpu/fp_quantizer.py
oelayan7: I don't think we need a variable for HPU, the old variable (python) was added to handle the special implementation introduced for CUDA kernels.

deepspeed/ops/fp_quantizer/quantize.py
oelayan7: Done (changed it in all appearances as well)

deepspeed/ops/fp_quantizer/quantize.py
oelayan7: @tjruwase you are right. It is better to do it that way.

tjruwase (approved):

tjruwase:
> This all looks good to me, thanks for this generalization! However, please address @tjruwase's suggestions.

@jeffra, the merging seems to be blocked by your requested changes. How to unblock?

jeffra (approved):

PR #6953, opened 2025-01-15
Pin numpy version

BLOrange-AMD:
This PR is to fix incompatible numpy version of pyt_deepspeed_megatron_gpt2 and pyt_train_deepspeed_megatron_gpt2 with ROCm PyTorch release/2.5 branch.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -2,7 +2,7 @@ einops
 hjson
 msgpack
 ninja
-numpy
+numpy<=1.26.4
 packaging>=20.0
 psutil
 py-cpuinfo


loadams:
Hi @BLOrange-AMD - I recall there being a dependency between the torch version and numpy version for us, could you share more info on the error you are seeing?

BLOrange-AMD:
@loadams With ROCm PyTorch 2.5, on pyt_deepspeed_megatron_gpt2 and pyt_train_deepspeed_megatron_gpt2 models, newer numpy-2.2.1 is downloaded and used instead of using cached numpy-1.26.4, which causes "RuntimeError: Could not infer dtype of numpy.int64". So fixed numpy version from DeepSpeed could be a more stable way to solve the issue.

loadams:
> @loadams With ROCm PyTorch 2.5, on pyt_deepspeed_megatron_gpt2 and pyt_train_deepspeed_megatron_gpt2 models, newer numpy-2.2.1 is downloaded and used instead of using cached numpy-1.26.4, which causes "RuntimeError: Could not infer dtype of numpy.int64". So fixed numpy version from DeepSpeed could be a more stable way to solve the issue.

@BLOrange-AMD - I see, however the issue is that DeepSpeed doesn't strictly require a lower numpy version.  Here is a [sample workflow ](https://github.com/microsoft/DeepSpeed/actions/runs/12814592404/job/35731368457#step:7:75) that uses numpy>2.0.0.  Is there another way to pin the numpy version?  Or do you believe this will be fixed soon in torch 2.6?

loadams:
@BLOrange-AMD - this doesn't seem to be needed currently, so I'm going to close this PR for now but we can re-open as needed.

PR #6960, opened 2025-01-17
Explicitly use the linalg.vector_norm call in `comm/`

loadams:
ds-chat workflow: https://github.com/microsoft/DeepSpeed/actions/runs/12835483030


GuanhuaWang:
I may misunderstood, is there a reason for switching from `norm` to `vec_norm`? To me, `norm` seems to cover the functionality of `vec_norm`?

loadams:
> Hi @loadams I may misunderstood, is there a reason for switching from `norm` to `vec_norm`? To me, `norm` seems to cover the functionality of `vec_norm`?

@GuanhuaWang - this is from the pytorch docs, it was mostly to make it more clear, but we don't "need" to update these like we needed to update from `torch.norm` to `torch.linalg.norm`

image

PR #6937, opened 2025-01-09
Add deepseek autotp

Yejing-Lai:
Deepseek including Multi-Head Latent Attention(MLA) and MoE.

For MLA TP, we need to skip two low-rank layers("q_a_proj" and "kv_a_proj_with_mqa)
For Deepseek MoE, tp_parse gets this moe layer name is layer_idx.down_proj, it is hard to add the policy, so we set the down_proj  layer to all_reduce_linears default.

PR #6964, opened 2025-01-21
Enabled high-performance Automatic Tensor Parallelism (auto TP) for the MoE models on multiple GPUs/HPUs

gyou2021:
Reduced the routed experts' AllReduce operation times per MoE layer to ONCE for the MoE models such as Qwen2-MoE and DeepSeek-V2. The results of all selected routed experts per layer on GPU/HPU cards will be gathered ONCE using the AllReduce operation, instead of gathering each selected routed expert individually or by the number of selected routed experts. This change will greatly increase performance.
In addition to modifying auto_tp.py, the following files should be updated: modeling_qwen2_moe.py and modeling_deepseek_v2.py. Add the following code after the weighted sum of the output of the selected experts per MoE layer. 
if is_deepspeed_available():
            if dist.is_initialized():
                dist.all_reduce(final_hidden_states, op=dist.ReduceOp.SUM)
Notes: final_hidden_states is the result of the weighted sum of the output of the selected experts per MoE layer.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ pre-commit install
 Afterwards, our suite of formatting tests run automatically before each `git commit`. You
 can also run these manually:
 ```bash
-pre-commit run --all-files
+pre-commit run --files  $(git diff --name-only master)
 ```
 If a formatting test fails, it will fix the modified code in place and abort
 the `git commit`. After looking over the changes, you can `git add <modified files>`
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 [![Downloads](https://static.pepy.tech/badge/deepspeed)](https://pepy.tech/project/deepspeed)
 [![Build](https://badgen.net/badge/build/check-status/blue)](#build-pipeline-status)
 [![OpenSSF Best Practices](https://www.bestpractices.dev/projects/9530/badge)](https://www.bestpractices.dev/projects/9530)
-[![Twitter](https://img.shields.io/twitter/follow/MSFTDeepSpeed)](https://twitter.com/intent/follow?screen_name=MSFTDeepSpeed)
-[![Japanese Twitter](https://img.shields.io/badge/%E6%97%A5%E6%9C%AC%E8%AA%9ETwitter-%40MSFTDeepSpeedJP-blue)](https://twitter.com/MSFTDeepSpeedJP)
+[![Twitter](https://img.shields.io/twitter/follow/DeepSpeedAI)](https://twitter.com/intent/follow?screen_name=DeepSpeedAI)
+[![Japanese Twitter](https://img.shields.io/badge/%E6%97%A5%E6%9C%AC%E8%AA%9ETwitter-%40DeepSpeedAI_JP-blue)](https://twitter.com/DeepSpeedAI_JP)
 [![Chinese Zhihu](https://img.shields.io/badge/%E7%9F%A5%E4%B9%8E-%E5%BE%AE%E8%BD%AFDeepSpeed-blue)](https://www.zhihu.com/people/deepspeed)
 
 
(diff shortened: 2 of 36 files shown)


delock:
Hi @gyou2021 , There is another PR for DeepSeek autotp in this link https://github.com/microsoft/DeepSpeed/pull/6937  What is relationship of your PR to this previous PR regarding on functionality?

@Yejing-Lai can you take a look at this PR?

deepspeed/module_inject/auto_tp.py
```diff
@@ -258,7 +260,9 @@ def update_policy_list(policy_list, new_module, new_gems):
             for i, policy in enumerate(policy_list):
                 # if module already exists in policy, combine gems and remove duplicates
                 if policy[0] == type(new_module):
-                    new_gems = set(new_gems + policy[1])
+                    #new_gems = set(new_gems + policy[1])
+                    print('new_gems ,policy[1] type', type(new_gems), type(policy[1]))
+                    new_gems = (new_gems + policy[1])
```
delock, line 265: Why does `new_gems+policy[1]`  change from a set  (ensures no duplicate item) to a normal concat operation?

gyou2021:
> Hi @gyou2021 , There is another PR for DeepSeek autotp in this link #6937 What is relationship of your PR to this previous PR regarding on functionality?
> 
> @Yejing-Lai can you take a look at this PR?

The difference lies in how the results of the weighted sum of routed experts per layer in the MoE are gathered. In my understanding, each selected routed expert per layer was gathered individually in https://github.com/microsoft/DeepSpeed/pull/6937, meaning the gathering time was proportional to the number of selected routed experts per layer. In this PR, the results are gathered once per layer, regardless of the number of selected routed experts.

deepspeed/module_inject/auto_tp.py
gyou2021: Because _"TypeError: can only concatenate list (not "set") to list"_ exists in new_gems = set(new_gems + policy[1]).  
Fixed it with new_gems = list(set(new_gems + policy[1])). It can ensure no duplicate items and no errors. 
The revised code was submitted.

delock:
Hi @gyou2021 , thanks for this optimization.  As we reviewed internally, need to figure out a way to turn on this optimization through knobs in model config.json file.  In that way this optimization won't break workloads that does not utilize this optimization.

Plus: I think it will be even more user friendly if the colasced allredcuce could be injected by AutoTP.  But as I reviewed with @gyou2021 , this seems tricky because no proper injection point can be captured by process nn.module.   @tohtana @loadams @tjruwase @Yejing-Lai if you have an idea we can discuss in this thread.

gyou2021:
> Hi @gyou2021 , thanks for this optimization. As we reviewed internally, need to figure out a way to turn on this optimization through knobs in model config.json file. In that way this optimization won't break workloads that does not utilize this optimization.
> 
> Plus: I think it will be even more user friendly if the colasced allredcuce could be injected by AutoTP. But as I reviewed with @gyou2021 , this seems tricky because no proper injection point can be captured by process nn.module. @tohtana @loadams @tjruwase @Yejing-Lai if you have an idea we can discuss in this thread.

Added support for the environment variable DS_MOE_EXPERTS_REDUCE_ONCE, which indicates whether the MoE expert all-reduce once optimization was enabled. If the value of DS_MOE_EXPERTS_REDUCE_ONCE is true, that is the optimization was enabled, the Auto_TP will execute corresponding operations to some linear layers.

delock:
Hi @loadams, this PR is now ready to review.  This PR allows user to explicitly turns off allreduce in experts for AutoTP tensor parallel inference, inorder to delay the allreduce until all experts are reduced locally.  Thanks!

deepspeed/module_inject/auto_tp.py
```diff
@@ -17,6 +17,11 @@
 from deepspeed.module_inject.tp_shard import get_shard_size, get_shard_size_list
 from deepspeed.utils import groups
 from deepspeed.module_inject.layers import is_autotp_training_mode
+<<<<<<< HEAD
+import os
+import ast
+=======
+>>>>>>> origin/master
```
hwchen2017, line 24: Line 20, 23 and 24 make all unit tests failed.

PR #6918, opened 2024-12-29
fix: `RuntimeError` for UCP large DP

saforem2:
We encountered a strange bug when attempting to convert checkpoints (created with DP=768) to universal format.

An overview of the bug as well as a detailed description of the proposed fix is written up in:

[argonne-lcf / Megatron-DeepSpeed / ALCF / notes / `universal_checkpoint_bug.md`](https://github.com/argonne-lcf/Megatron-DeepSpeed/blob/main/ALCF/notes/universal_checkpoint_bug.md)

PR #6974, opened 2025-01-26
Remove assumption that padding only occurs on last rank

xylian86:
As discussed in [PR-6918](https://github.com/microsoft/DeepSpeed/pull/6918), padding can occur on multiple ranks with large DP degrees.

For example, with:
- Flattened tensor size: 266240
- DP degree: 768
- Alignment: 1536 
- Required padding: 1024 (1536 * 174 - 266240)
- Per-rank partition size: 348 (1536 * 174 / 768)
- The padding occurs on last three ranks. 

This PR removes the single-rank padding assumption for more general cases.


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -384,6 +377,18 @@ def __init__(self,
...
             self.parallel_partitioned_bit16_groups.append(data_parallel_partitions)
 
+            # Record padding required for alignment
+            left_boundary = sum([t.numel() for t in data_parallel_partitions[:partition_id]])
+            curr_partition_size = data_parallel_partitions[partition_id].numel()
+
+            if orig_group_numel <= left_boundary:
+                padding = curr_partition_size
+            elif orig_group_numel < left_boundary + curr_partition_size:
+                padding = orig_group_numel - left_boundary
```
tjruwase, line 387: Please double check that it shouldn't be 
```suggestion
                padding = (left_boundary + curr_partition_size) - orig_group_numel
```

xylian86: Yes, you are right, the padding should be the right side of this partition.

tjruwase:
@xylian86, thanks for the quick solution! 

@saforem2, can you please test this PR?

saforem2:
yes will work on testing this today, thanks!

saforem2:
This appears to be fixed.

I've added a [new comment with details](https://github.com/deepspeedai/DeepSpeed/pull/6918#issuecomment-2643247653) in the original PR

PR #6989, opened 2025-02-01
import triton files when triton is supported and installed

oelayan7:
Those files have code that gets run when importing them, so in systems that doesn't support triton but have triton installed this causes issues. 

In general, I think it is better to import triton when it is installed and supported.

### DIFF
--- a/deepspeed/model_implementations/transformers/ds_transformer.py
+++ b/deepspeed/model_implementations/transformers/ds_transformer.py
@@ -14,7 +14,7 @@ from deepspeed.ops.transformer.inference.ds_attention import DeepSpeedSelfAttent
 from deepspeed.ops.transformer.inference.op_binding.workspace import WorkspaceOp
 from deepspeed.accelerator import get_accelerator
 import deepspeed
-if deepspeed.HAS_TRITON:
+if deepspeed.HAS_TRITON and get_accelerator().is_triton_supported():
     from deepspeed.ops.transformer.inference.triton.mlp import TritonMLP
     from deepspeed.ops.transformer.inference.triton.attention import TritonSelfAttention
 


deepspeed/model_implementations/transformers/ds_transformer.py
```diff
@@ -14,7 +14,7 @@
 from deepspeed.ops.transformer.inference.op_binding.workspace import WorkspaceOp
 from deepspeed.accelerator import get_accelerator
 import deepspeed
-if deepspeed.HAS_TRITON:
+if deepspeed.HAS_TRITON and get_accelerator().is_triton_supported():
```
loadams, line 17: Is it worth cleaning up the `HAS_TRITON` flag if it isn't useful in this context, which it seems like it should have been.

loadams (approved):

oelayan7:
@loadams  some CI job is failing on this 
image

deepspeed/model_implementations/transformers/ds_transformer.py
oelayan7: In other places it is checked that the system has triton installed and the use_triton flag is passed. 
So I thought those other places are protected, as it is the user's responsibility to pass the use_triton flag. 

But in this import, no other protection is done.

loadams:
> @loadams some CI job is failing on this image

This should be transient, but taking a look.

loadams:
Hi @oelayan7 - could you add the DCO that's now required as a result of the repo move?  More information about fixing this check can be found [here](https://github.com/deepspeedai/DeepSpeed/pull/6989/checks?check_run_id=36804084817), and going forward we will need to sign all commits with DCO.

oelayan7:
@loadams  Done, sorry for delay. 
Can you please retrigger the CI?

PR #6991, opened 2025-02-02
Allow NVIDIA Blackwell

fabiendupont:
NVIDIA Blackwell GPU generation has number 10. The SM code and architecture should be `100`, but the current code generates `1.`, because it expects a 2 characters string.

This change modifies the logic to consider it as a string that contains a `.`, hence splits the string and uses the array of strings.

### DIFF
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -612,8 +612,8 @@ class CUDAOpBuilder(OpBuilder):
 
         - `TORCH_CUDA_ARCH_LIST` may use ; or whitespace separators. Examples:
 
-        TORCH_CUDA_ARCH_LIST="6.1;7.5;8.6" pip install ...
-        TORCH_CUDA_ARCH_LIST="6.0 6.1 7.0 7.5 8.0 8.6+PTX" pip install ...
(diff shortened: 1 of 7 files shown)


fabiendupont:
@microsoft-github-policy-service agree company="Red Hat"

op_builder/builder.py
```diff
@@ -651,9 +651,9 @@ def compute_capability_args(self, cross_compile_archs=None):
         args = []
         self.enable_bf16 = True
         for cc in ccs:
-            num = cc[0] + cc[2]
+            num = cc[0] + cc[1].split('+')[0]
```
tjruwase, line 654: Could you please update the comment lines 615 & 616 with Blackwell compute? This will help understanding of this code. Thanks!

tjruwase (approved):

loadams (approved):

op_builder/builder.py
fabiendupont: @tjruwase, fair point. I added the Hopper and Blackwell compute capabilities in the comments at lines 615 & 616.

PR #6993, opened 2025-02-03
Enable torch.autocast with ZeRO

tohtana:
This PR supports `torch.autocast` with ZeRO. You need to enable `torch.autocast` in DeepSpeed config.

```json
"torch_autocast": {
  "enabled": true,
  "dtype": "bfloat16"
}
```

You don't need to explicitly call `torch.autocast` in your code. The grad scaler is also applied in the DeepSpeed optimizer.

All the parameters are maintained in FP32 but certain operators and layers are computed in the specified dtype. With ZeRO enabled, the communication (allreduce/reduce_scatter) for specified layers are also done in the specified dtype ([List of modules](https://github.com/microsoft/DeepSpeed/blob/0e52edb6a3cc940a9d8c53f7343f3473ebd69bbb/deepspeed/runtime/torch_autocast.py)).
You cannot enable `fp16` or `bf16` with `torch.autocast`.

(Currently working on ZeRO3)

### DIFF
--- a/deepspeed/runtime/constants.py
+++ b/deepspeed/runtime/constants.py
@@ -203,6 +203,23 @@ AMP_ENABLED = "enabled"
 AMP_ENABLED_DEFAULT = False
 
 #########################################
+# Torch AMP support
+#########################################
+TORCH_AUTOCAST_FORMAT = '''
+PyTorch autocast config should be of the format:
+"torch_autocast": {
+  "enabled": true,
+  "dtype": "bfloat16",
+}
+'''
+TORCH_AUTOCAST = "torch_autocast"
+
+TORCH_AUTOCAST_ENABLED = "enabled"
+TORCH_AUTOCAST_ENABLED_DEFAULT = False
+TORCH_AUTOCAST_DTYPE = "dtype"
+TORCH_AUTOCAST_DTYPE_DEFAULT = None
+
+#########################################
 # Gradient clipping
 #########################################
 # Gradient clipping. By default, this feature is not enabled.
--- a/deepspeed/runtime/config.py
+++ b/deepspeed/runtime/config.py
@@ -156,6 +156,24 @@ def get_amp_params(param_dict):
         return False
 
 
+def get_torch_autocast_enabled(param_dict):
+    if TORCH_AUTOCAST in param_dict.keys():
+        return get_scalar_param(param_dict[TORCH_AUTOCAST], TORCH_AUTOCAST_ENABLED, TORCH_AUTOCAST_ENABLED_DEFAULT)
+    else:
+        return False
+
+
+def get_torch_autocast_dtype(param_dict):
+    if TORCH_AUTOCAST in param_dict:
+        if TORCH_AUTOCAST_DTYPE in param_dict[TORCH_AUTOCAST]:
+            try:
+                return DtypeEnum(param_dict[TORCH_AUTOCAST][TORCH_AUTOCAST_DTYPE]).value
+            except KeyError:
+                raise ValueError(
+                    f"Invalid dtype for torch autocast: {param_dict[TORCH_AUTOCAST][TORCH_AUTOCAST_DTYPE]}")
+    return None
+
+
 def get_fp16_enabled(param_dict):
     if FP16 in param_dict.keys():
         return get_scalar_param(param_dict[FP16], FP16_ENABLED, FP16_ENABLED_DEFAULT)
@@ -835,6 +853,8 @@ class DeepSpeedConfig(object):
         self.fp16_master_weights_and_gradients = get_fp16_master_weights_and_grads_enabled(param_dict)
         self.amp_enabled = get_amp_enabled(param_dict)
         self.amp_params = get_amp_params(param_dict)
+        self.torch_autocast_enabled = get_torch_autocast_enabled(param_dict)
+        self.torch_autocast_dtype = get_torch_autocast_dtype(param_dict)
         self.loss_scale = get_loss_scale(param_dict)
         self.initial_dynamic_scale = get_initial_dynamic_scale(param_dict)
         self.dynamic_loss_scale_args = get_dynamic_loss_scale_args(param_dict)
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -374,6 +374,11 @@ class DeepSpeedEngine(Module):
 
         self._is_compiled = False
 
+        # Verify autocast setting
+        if self.torch_autocast_enabled():
+            assert not self.fp16_enabled(), "Cannot enable both torch autocast and fp16"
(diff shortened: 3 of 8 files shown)


stas00:
I have a question about `lower_precision_safe_modules`

https://pytorch.org/docs/stable/amp.html#torch.autocast doesn't have an option to specify `lower_precision_safe_modules` - why do we then put the onus on the deepspeed user? (and I'm aware that it's optional and there is a default list). 

My question is - can we automatically retrieve that safe list from pytorch? Or is this because what's pytorch considers safe isn't necessarily what deepspeed considers safe?

deepspeed/runtime/constants.py
```diff
@@ -202,6 +202,23 @@
...
 
+#########################################
+# Torch AMP support
+#########################################
+TORCH_AUTOCAST_FORMAT = '''
+PyTorch autocast config should be of the format:
+"torch_autocast": {
+  "enabled": true,
+  "dtype": "bfloat16",
+}
```
stas00, line 213: add `...` here as in other sections as it's incomplete.

stas00: or update it with the rest of flags?

stas00 (commented):
If the training loop uses `torch.autocast` 
1. could deepspeed detect that and probably assert if ds config isn't done right?
2. or actually should it assert if this happens because then the behavior is unsupported?

tohtana:
Hi @stas00, sorry I have missed your replies.

> My question is - can we automatically retrieve that safe list from pytorch? Or is this because what's pytorch considers safe isn't necessarily what deepspeed considers safe?

Great question. Ideally, we would retrieve the list directly from PyTorch. I found it in the following files:
- https://github.com/pytorch/pytorch/blob/18a7a04c4adecda3be17dd364d48d484fd1dcdba/aten/src/ATen/autocast_mode.cpp
- https://github.com/pytorch/pytorch/blob/18a7a04c4adecda3be17dd364d48d484fd1dcdba/aten/src/ATen/autocast_mode.h#L828

It does seem challenging to determine which module parameters are safe to use in lower precision. Hopefully, we can refine the default list for this feature based on real-world usage.

tohtana:
> 1. could deepspeed detect that and probably assert if ds config isn't done right?
> 2. or actually should it assert if this happens because then the behavior is unsupported?

I think it should be harmless to include `torch.autocast` in the training loop, though we should display a warning message. Let me double-check the behavior and add that message.

One confusing aspect is that it's valid to use PyTorch’s `torch.autocast` while disabling "torch_autocast" in the DeepSpeed config. In this case, DeepSpeed treats it as a pure FP32 setting. The main issue is that all communication operations (e.g., gradient reduction and all-gather) will also be performed in FP32. Since some modules are computed in BF16/FP16, this leads to inefficiencies in many cases.

stas00:
> One confusing aspect is that it's valid to use PyTorch’s torch.autocast while disabling "torch_autocast" in the DeepSpeed config. In this case, DeepSpeed treats it as a pure FP32 setting.

Dare I say this is not valid - nobody does fp32 training these days.

Such situations should assert IMHO.

And fp32 comms is a huge problem as well, not just because of time overhead, but because the comms buffer is 2x larger and this happens 2 times - so overall it'll consume 4x more memory for all comms. I happen to know this because I discovered this when porting Ulysses SP, where it automatically enables fp32 comms and I had huge memory spikes in backward because of that.

stas00:
I'm talking about peak memory usage.

if you run torch memory profiler around the reduction call you will see that it'll use `reduce_bucket_size * 4 bytes * 2 copies` - I think it's because one goes from 2b to 4b (1 copy) then torch/nccl makes another copy?

I measured the default `reduce_bucket_size` 5e8 to consume 4GB peak memory usage when comms are in fp32. and only 1GB in bf16.

since those reduction ops memory peaks happen during backward they often lead to OOM if you want to push bs/seqlen.

deepspeed/runtime/constants.py
tohtana: Thank you @stas, fixed to include the other config item.

sfc-gh-sbekman:
Thank you for looking into it,  Masahiro. No problem doing it elsewhere.

Using torch mem profiler will be very helpful to see the reduction memory spikes

https://pytorch.org/blog/understanding-gpu-memory-1/ - it's very easy to set up - if you need help please let me know.

sfc-gh-sbekman:
> Good, I can an assertion to detect that torch.autocast is enabled outside of DeepSpeed but ds_config doesn't set torch_autocast's enabled. Or it might be better to automatically enable it.

If it has to be on and it breaks nothing then automatically enabling it is probably a better idea to help with ease of use.

tohtana:
Hi @stas, 
Thank you for your feedback!

> If it has to be on and it breaks nothing then automatically enabling it is probably a better idea to help with ease of use.

After reviewing the design, I now feel automatically enabling it wouldn't be straightforward. This autocast feature sets some flags to parameters before the optimizer is initialized. However, we only know whether `torch.autocast` is enabled or not just before a forward pass call as `with torch.autocast(...)` is placed to wrap a forward call. Reinitializing parts of the optimizer at that point would complicate the code.

Given that, I think it’s better to [throw an error with the explanation](https://github.com/deepspeedai/DeepSpeed/blob/bb855b762300146f5549bbe98f027438d14c1894/deepspeed/runtime/torch_autocast.py#L97).

stas00:
Then assert is the way to go, Masahiro

tohtana:
> Then assert is the way to go, Masahiro

Thank you @stas00, then can you approve this PR?

stas00:
Hmm, I can't just hit approve, that would be defeat the purpose of doing the review.

We have only discussed one small aspect of this PR, which has been resolved, but the rest of the PR I don't know and currently rushing to finish the porting of Ulysses so until that is done I won't have time to do a serious review.

tjruwase (approved):

Vervious:
I'm running into an issue where turning on this feature results in massive grad norms, using zero 2; have you seen this before?

```
[deepspeed.torch_autocast]
enabled = true  # NOTE: turning this on makes grad norms explode in stage 2 (but not stage 3)
dtype = "bfloat16"

# [deepspeed.bf16]
# enabled = true  # this works properly

[deepspeed.zero_optimization]
stage = 2
allgather_partitions = true
overlap_comm = false
reduce_scatter = true
contiguous_gradients = true
stage3_prefetch_bucket_size = 0
stage3_max_live_parameters = 0
stage3_max_reuse_distance = 0
stage3_gather_16bit_weights_on_model_save = true

```

Grad norms were reported via `model_engine.get_global_grad_norm()` and also observed via `safe_get_full_grad` after a backward call (but before step). Stage 3 seems to have reasonable grad norms. For some reason the loss curves also don't match exactly between stage 2 and 3 (but they match exactly using deepspeed.bf16 instead)

PR #6999, opened 2025-02-04
config torch to avoid graph breaks caused by logger

ShellyNR:
Following changes in Pytorch trace rules , my previous PR to avoid graph breaks caused by logger is no longer relevant. So instead I've added this functionality to torch dynamo - https://github.com/pytorch/pytorch/commit/16ea0ddcdbe49f5560bfbbc6ab331c3422e27b2e 
This commit allows the user to config torch to ignore logger methods and avoid associated graph breaks.

To enable ignore logger methods - os.environ["DISABLE_LOGS_WHILE_COMPILING"] = "1"
To ignore logger methods except for a specific method / methods (for example, info and isEnabledFor) - os.environ["DISABLE_LOGS_WHILE_COMPILING"] = "1"
 and os.environ["LOGGER_METHODS_TO_EXCLUDE_FROM_DISABLE"] = "info, isEnabledFor"

### DIFF
--- a/deepspeed/utils/logging.py
+++ b/deepspeed/utils/logging.py
@@ -7,7 +7,8 @@ import functools
 import logging
 import sys
 import os
-from deepspeed.runtime.compiler import is_compile_supported, is_compiling
+import torch
+from deepspeed.utils.torch import required_torch_version
 
 log_levels = {
     "debug": logging.DEBUG,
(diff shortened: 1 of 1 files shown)


loadams:
> Thank you @ShellyNR, this is a very clean approach. I would like to approve this, but DCO prevents tests from running because of the migration to `deepspeedai`. Can you rebase based on the [explanation](https://github.com/deepspeedai/DeepSpeed/pull/6999/checks?check_run_id=36669743198)?

I believe I've fixed this for now.

tjruwase (approved):

PR #7000, opened 2025-02-04
Update workflows to cuda 12.4

loadams:
- Update existing workflows that use cu121 to cu124.  Note, this means that where we download torch latest, we will now be getting torch 2.6 rather than the torch latest 2.5 provided with cuda 12.1.
- Note, nv-nightly is failing in master currently due to unrelated errors, so this could be ignored in this PR (nv-nightly tested locally, where it passes with 12.1 and it also passes with 12.4).


.github/workflows/nv-ds-chat.yml
```diff
@@ -49,7 +49,7 @@ jobs:
 
       - name: Install deepspeed-chat
         run: |
-          BRANCH="master"
+          BRANCH="loadams/torch-load-weights-only"
```
tjruwase, line 52: Is this change needed?

tjruwase (approved):

.github/workflows/nv-ds-chat.yml
loadams: This was for testing until the PR was merged into DSE, but its reverted now.

PR #7004, opened 2025-02-05
autotp training(fix dco)

inkcherry:
Same as [this PR](https://github.com/deepspeedai/DeepSpeed/pull/6922). [affeb88](https://github.com/deepspeedai/DeepSpeed/pull/6922/commits/affeb884576936b3a0efc5ce86435da37db1d87e)
I noticed the CI updated the DCO check recently. Using the suggested rebase method for sign-off would reintroduce many conflicts, so I opted for a squash merge with sign-off instead. thanks: )

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -37,7 +37,7 @@ from .runtime.lr_schedules import add_tuning_arguments
 from .runtime.config import DeepSpeedConfig, DeepSpeedConfigError
 from .runtime.activation_checkpointing import checkpointing
 from .ops.transformer import DeepSpeedTransformerLayer, DeepSpeedTransformerConfig
-from .module_inject import replace_transformer_layer, revert_transformer_layer
+from .module_inject import replace_transformer_layer, revert_transformer_layer, set_autotp_mode
 
 from .utils import log_dist, OnDevice, logger
 from .comm.comm import init_distributed
@@ -364,3 +364,34 @@ def init_inference(model, config=None, **kwargs):
     engine = InferenceEngine(model, config=ds_inference_config)
 
     return engine
+
+
+def tp_model_init(model, tp_size, dtype):
+    """
+    Initialize the model for tensor parallelism.
+
+    Args:
+        model (torch.nn.Module): The model to be initialized.
+        tp_size (int): The tensor parallelism size.
+        dtype (torch.dtype): The data type to be used for the model.
+
+    Returns:
+        torch.nn.Module: The initialized model with tensor parallelism.
+    """
+    # avoid re-entry
+    assert not hasattr(
+        model, 'ds_autotp_parsed'), "ds_autotp_parsed' attribute already exists in the model, re-entry is not allowed."
+
+    set_autotp_mode(training=True)
+
+    from deepspeed.runtime.tensor_parallel import TpTrainingManager
+    # The expected usage here is for it to be invoked by transformers package.
+
+    #TODO: We should provide a custom TP mapping solution without using autoTP
+    #as modifying the autoTP logic may be more difficult for users compared to configuring it
+
+    model = TpTrainingManager(model=model, tp_size=tp_size, dtype=dtype).module
+
+    setattr(model, 'ds_autotp_parsed', True)
+
+    return model
--- a/deepspeed/comm/comm.py
+++ b/deepspeed/comm/comm.py
@@ -225,6 +225,12 @@ def broadcast(tensor, src, group=None, async_op=False, prof=False, log_name='bro
 
 
 @timed_op
+def broadcast_object_list(object_list, src, group=None, device=None):
+    global cdb
+    return cdb.broadcast_object_list(object_list=object_list, src=src, group=group, device=device)
+
+
+@timed_op
 def all_gather(tensor_list,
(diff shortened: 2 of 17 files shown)


tjruwase (approved):

delock:
Kudos @inkcherry for contributing AutoTP training!  It's a nice feature make tensor parallel training/finetuning more available to HF model users.

I think a tutorial page would help user discover and learn how to use this feature in DeepSpeed.   Is it possible to write a tutorial and add it under https://github.com/deepspeedai/DeepSpeed/tree/master/docs/_tutorials introducing steps how to use this feature?  I remember you have an example training alpaca with DeepSpeed AutoTP.

inkcherry:
> Kudos @inkcherry for contributing AutoTP training! It's a nice feature make tensor parallel training/finetuning more available to HF model users.
> 
> I think a tutorial page would help user discover and learn how to use this feature in DeepSpeed. Is it possible to write a tutorial and add it under https://github.com/deepspeedai/DeepSpeed/tree/master/docs/_tutorials introducing steps how to use this feature? I remember you have an example training alpaca with DeepSpeed AutoTP.

Yes, I will add some document soon~

tjruwase:
@inkcherry, I think a blog would be appropriate to publicize this amazing technology. Although blogs can be a bit work, we will be glad to collaborate and jointly advertise.

oelayan7:
@inkcherry @loadams @tjruwase 
cc @nelyahu 
Please notice that this PR has removed keep_module_on_host which was added in https://github.com/deepspeedai/DeepSpeed/pull/6846/ 

keep_module_on_host is currently ignored and not used in auto_tp. 
Please align the current code to the old version that was in auto_tp

tjruwase:
@inkcherry, can you please help address issues raised by @oelayan7. It seems we need to re-apply the lost changes from #6846.

inkcherry:
@oelayan7  
Sorry for the break, I actually noticed this issue as well.
You can use this as a temporary workaround for now:
in ```deepspeed/module_inject/layers ```
```_partition = move(_partition, 'cpu').detach()```

I will fix it later. Thanks!

**Comment by inkcherry** [comment]
hi @tjruwase @stas00  Thanks for the information! In my local environment with PyTorch 2.7, I noticed that the unit test passes, but the pytest process doesn't exit cleanly. Using ```reuse_dist_env=True``` resolve the issue. However, it appears that ```set_autotp_mode(training=False)``` still causes a hang. I've submitted a fix here: https://github.com/deepspeedai/DeepSpeed/pull/7321.

**Comment by stas00** [comment]
it hangs here in starmap
```
    def _close_pool(self, pool, num_procs, force=False):
        if force or not self.reuse_dist_env:
            msg = pool.starmap(self._dist_destroy, [() for _ in range(num_procs)])
            pool.close()
            pool.join()
```
the corresponding part of the trace is:
```
    wait (threading.py:320)
    wait (threading.py:607)
    wait (multiprocessing/pool.py:765)
    get (multiprocessing/pool.py:768)
    starmap (multiprocessing/pool.py:375)
    _close_pool (unit/common.py:349)
    pytest_runtest_teardown (conftest.py:80)
```

stas00:
should we open an Issue about it? Otherwise this will get forgotten and then someone will have to diagnose this again. It'll happen as soon as someone tries it with pt-2.7.x

delock:
> should we open an Issue about it? Otherwise this will get forgotten and then someone will have to diagnose this again. It'll happen as soon as someone tries it with pt-2.7.x

Did https://github.com/deepspeedai/DeepSpeed/pull/7321 fixed the issue?  Otherwise we should open an issue to track.

inkcherry:
@stas00 @delock yes, created : https://github.com/deepspeedai/DeepSpeed/issues/7334

stas00:
I suppose the thing to figure out is whether you just need to make the tests pass, or is there an actual bug introduced in pt-2.7 that needs to be diagnosed, reported and fixed in the pt-core?

Swiping the problem under the carpet but making the tests pass often can rear its head later in a real user application.

But this is not the project I was part of, I just reported the problem, so it's up to you what you do with it.

delock:
> I suppose the thing to figure out is whether you just need to make the tests pass, or is there an actual bug introduced in pt-2.7 that needs to be diagnosed, reported and fixed in the pt-core?
> 
> Swiping the problem under the carpet but making the tests pass often can rear its head later in a real user application.
> 
> But this is not the project I was part of, I just reported the problem, so it's up to you what you do with it.

Thanks @stas00 !   We will take a look at this.   I think it depends on whether this bug can only be reproduced by pytest or we can build a standalone reproducer without using pytest.   Will need some deeper look to decide.

inkcherry:
thanks @delock and @stas00, Yes, we've observed an apparent conflict between the process exit logic in DistributedTest (which was designed independently) and PyTorch's destroy_process_group. We'll conduct a deeper analysis to: Isolate whether this is solely a PyTorch issue, or A design mismatch in our test infrastructure.
Will sync findings promptly

rookie-ryan:
Hey @inkcherry , I am confused that the function tp_model_init is not called in the initialization. Does this function need to be explicitly called when I want to enable autoTP training?

PR #7007, opened 2025-02-06
Enable python 3.11 and 3.12 tests

loadams:


tjruwase (approved):

tohtana:
PyTorch doesn't support Python 3.8 now. I think we can also stop supporting 3.8.

tohtana:
Removed Python 3.8 and 3.9 (The latest PyTorch version works with 3.10+). I also removed sudo as it caused error on the test environment. Please open a new PR if you need to include it. @loadams 
UPDATE: Instead of removing temporal files for installed packages, I replaced the images with smaller ones.

PR #7010, opened 2025-02-06
test new CI trigger HPU job

oelayan7:


loadams:
hpu-gaudi2 workflow passed.

PR #7016, opened 2025-02-07
Update A6000 tests transformers version

loadams:


.github/workflows/nv-a6000.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         run: |
           git clone https://github.com/huggingface/transformers
           cd transformers
-          git checkout v4.47.1
+          #git checkout v4.47.1
```
stas00, line 45:
```suggestion
          # if you need to use an older transformers version temporarily in case of breakage
          # git checkout v4.47.1
```

stas00: proposing to future proof it ;)

stas00 (changes requested):
LGTM

.github/workflows/nv-a6000.yml
loadams: Makes sense to me

stas00 (approved):

PR #7018, opened 2025-02-08
Training multiple models

tjruwase:
Support training multiple models, such as in [HF](https://huggingface.co/docs/accelerate/en/usage_guides/deepspeed_multiple_model)

Here is some update on supporting multiple DS engines with single loss.backward(). The main message is that I think we can support this. First, some context. Backward pass in ZeRO is complicated because the optimizations/features require special handling of gradients, such as:

1. Gradient partitioning
2. Overlapping backward and reduction
3. Upcasting for fp32 grad accumulation

So, we created engine.backward(loss)  as a wrapper function to provide us fine-grained control over backward as below

```python
def backward(loss):
 backward_prologue() # setup logic for special gradient handling
 loss.backward()
 backward_epilogue() # cleanup/teardown logic
```

As demonstrated by @muellerzr, this approach breaks down when loss originates from multiple DS engines. Our proposed solution is to use backward hooks on the module to launch backward_prologue()  and backward_epilogue() . Specifically,

1. backward pre hook on engine.module to launch backward_prologue() before any module gradient is created.
2.  backward post hook on engine.module to launch backward_epilogue() after all module gradients are created.

We plan for this solution to preserve BC, i.e., engine.backward() will remain correct  for single engine scenarios.
The current status is that (1) is completed, while (2) is in progress. To unblock e2e testing for multi-engine scenarios, since there are probably other issues, we have a temporarily added engine._backward_prologue() . You can try this out via the following artifacts.

1. Simple multi-engine test code: https://gist.github.com/tjruwase/f1adccf087b8fa269ffce2ab91c4f1c6#file-multi_engine-py
2. DS branch: https://github.com/microsoft/DeepSpeed/tree/olruwase/zero_multi_models


stas00 (commented):
I am missing context of what this PR is doing (other than it tries to do something wrt training multiple models).

But don't you need new tests?

tjruwase:
> I am missing context of what this PR is doing (other than it tries to do something wrt training multiple models).
> 
> But don't you need new tests?

@stas00, thanks for the feedback. I have updated the OP with some background from earlier discussions. 

I will work on converting the gist codes into UTs.

stas00:
That's much better after your OP expansion, Tunji.

The gists look good, please ping me once they are tests and would be happy to review again.

That's a very important feature for quite a few users.

tjruwase:
> The gists look good, please ping me once they are tests and would be happy to review again.

@stas00, the code is ready for review. I am still having CI issues though.

tests/unit/runtime/test_multiple_models.py
```diff
@@ -0,0 +1,99 @@
...
+    data_loader = random_dataloader(model=models[0],
+                                    total_samples=4,
+                                    hidden_dim=hidden_dim,
+                                    device=models[0].device,
+                                    dtype=dtype)
+    dist.barrier()
+    for _, batch in enumerate(data_loader):
+        losses = [m.module(batch[0], batch[1]) for m in models]
+        loss = sum(l / (i + 1) for i, l in enumerate(losses))
+        loss.backward()
```
stas00, line 34: should the case with each model driving its own `backward` be tested as well? 

I'm not sure independent models will want to be connected and impact each other via autograd as is the case here.

probably could be done in this same test as another loop?

deepspeed/runtime/engine.py
```diff
@@ -2136,30 +2143,50 @@ def backward(self, loss, release_loss=False, retain_graph=False, scale_wrt_gas=T
...
-        if do_gradient_reduction:
-            # Traditional code path that allreduces the module parameter grads
-            self.allreduce_gradients()
+    @contextmanager
+    def no_sync(self):
+        r"""
+            Context manager to disable gradient reduction during backward pass.
+            This context manager has the following effects on other DeepSpeed features.
+            1. Incompatible with ZeRO stage 2/3 which rely on reduction for gradient partitioning.
+            2. It is illegal to  call engine.step() within the context manager.
```
stas00, line 2160:
```suggestion
            2. It is illegal to call engine.step() within the context manager.
```

stas00: perhaps it's slightly more clear to say: "within this context context manager"
```suggestion
            2. It is illegal to call engine.step() within this context manager.
```

deepspeed/runtime/engine.py
```diff
@@ -2136,30 +2143,50 @@ def backward(self, loss, release_loss=False, retain_graph=False, scale_wrt_gas=T
...
-        self._start_timers(self.engine_timers.backward_reduce_timers)
-
-        if do_gradient_reduction:
-            # Traditional code path that allreduces the module parameter grads
-            self.allreduce_gradients()
+    @contextmanager
+    def no_sync(self):
+        r"""
+            Context manager to disable gradient reduction during backward pass.
+            This context manager has the following effects on other DeepSpeed features.
```
stas00, line 2158:
```suggestion
            This context manager has the following effects on other DeepSpeed features:
```

tests/unit/runtime/test_multiple_models.py
```diff
@@ -0,0 +1,99 @@
...
+
+        for m in models:
+            m._backward_epilogue()
```
stas00, line 37: so will this be now a requirement for users wanting to use 2+ models? If so it needs to be documented, no?

tests/unit/runtime/test_multiple_models.py
tjruwase: Can you clarify a bit more, because I think the case of each model driving its own backward is different from this case of composing individual losses into a `global` loss.  

More specifically, are you referring to a case like below? 
```python
    models = [...]
    for batch in dataset:
       losses = [m(batch) for m in models]
       for m, loss in zip(models, losses):
          m.backward(loss)
  ```

tests/unit/runtime/test_multiple_models.py
tjruwase: Yes, it needs documenting as experimental since I view it as a temporary hack. The actual solution is a hook that triggers epilogue after the last grad creation of the model. Althought Z3 has similar code, I have not found time to adapt it :(

tests/unit/runtime/test_multiple_models.py
stas00: yes, that would be one of them. I surely can't think of all use-cases, but there will be definitely some that will want to independent engines, running at the same time.

tests/unit/runtime/test_multiple_models.py
stas00: yeah, let's document that - otherwise what's the point of investing so much energy if others then don't know they can use it. Does it resonate?

Even a super basic example or even pointing to a test would be better than nothing, IMHO.

stas00 (approved):
Other than the discussions we are having about docs/tests this in general looks good, though I don't know all the nuances, so ideally someone else should review. 

Thank you for working on this, Tunji!

tjruwase:
> ideally someone else should review.

@sfc-gh-jrasley or @muellerzr can you help review?

tests/unit/runtime/test_multiple_models.py
stas00: Thank you for adding the docs, Tunji!

PR #7023, opened 2025-02-11
AIO on ROCM

jomayeri:
Adding compile support for AIO library on AMD GPUs.


op_builder/async_io.py
```diff
@@ -65,8 +65,11 @@ def extra_ldflags(self):
 
         import torch.utils.cpp_extension
         CUDA_HOME = torch.utils.cpp_extension.CUDA_HOME
-        CUDA_LIB64 = os.path.join(CUDA_HOME, "lib64")
-        ldflags = [f'-L{CUDA_HOME}', f'-L{CUDA_LIB64}', '-laio', '-lcuda', '-lcudart']
+        if CUDA_HOME is None:
```
loadams, line 68: Perhaps better to use the `is_rocm_pytorch()` function to check?

loadams (approved):

op_builder/async_io.py
tjruwase: I think it is better to check for CUDA vs non-CUDA in general since we support other accelerators. I believe rocm is more similar to non-CUDA for this case.

PR #7024, opened 2025-02-11
[Ulysses tutorial] typos

stas00:
Fix typos

### DIFF
--- a/docs/_tutorials/ds-sequence.md
+++ b/docs/_tutorials/ds-sequence.md
@@ -49,7 +49,7 @@ def forward():
 
 ```
 
-* **Add sequence parallel communication group**:  Note that DistributedAttention takes `local_attn` and `sequence_parallel_group` as the parameters, where local_attn can be your original attention block. You also need to build the sequence parallel nication group and pass that the DistributedAttention. One way to do this is to build the sequence parallel group at the model initialization stage.
+* **Add sequence parallel communication group**:  Note that DistributedAttention takes `local_attn` and `sequence_parallel_group` as the parameters, where local_attn can be your original attention block. You also need to build the sequence parallel communication group and pass that the DistributedAttention. One way to do this is to build the sequence parallel group at the model initialization stage.
 
 
 ```python
@@ -94,7 +94,7 @@ DeepSpeed's sequence parallelism can be combined with different types of attenti
 
 `FlashAttention`: the implementation from [FlashAttention: Fast and Memory-Efficient Exact Attention with IO-Awareness](https://arxiv.org/abs/2205.14135). Enabled by `--use-flash-attn`.
 
-`FlashAttention + Triton`: a of FlashAttention in Triton (tested with triton==2.0.0.dev20221202). Enabled by `--use-flash-attn-triton`.
+`FlashAttention + Triton`: FlashAttention in Triton (tested with triton==2.0.0.dev20221202). Enabled by `--use-flash-attn-triton`.
 
 For the best performance, we recommend using FlashAttention + Triton. Below are the installation steps. Note that FlashAttention is compatible only with NVIDIA Turing, Ampere, Ada, or Hopper GPUs.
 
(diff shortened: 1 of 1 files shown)


stas00:
What is this new DCO thing? 

I created a PR via github directly so I'm not sure what is missing as the instructions don't apply.

loadams:
> What is this new DCO thing?
> 
> I created this PR via github directly so I'm not sure what is missing as the instructions don't apply to gh workflow.

I've had issues with DCO with using GH to resolve merge conflicts too, all commits have to be signed off, so it needs to include the `-s` with the commit, which I haven't found a way to enable in the GH GUI.  But [DCO](https://github.com/dcoapp/app?tab=readme-ov-file) replaces the CLA when moving from Microsoft to LF.

tjruwase:
I typically override as follows:

Click details 
image

Click "Set to Pass"
image

stas00:
so basically DCO isn't compatible with github-based commits?

stas00:
do we really need a CLA? It's no longer a company owned project - perhaps it can be just dropped, no?

tjruwase:
> so basically DCO isn't compatible with github-based commits?

The docs claim that `-s` on `git commit` avoids this. I have been trying to do that, but still forget quite often. 
image

loadams:
> do we really need a CLA? It's no longer a company owned project - perhaps it can be just dropped, no?

@stas00 This is a requirement from the Linux Foundation for their projects.

loadams:
> > so basically DCO isn't compatible with github-based commits?
> 
> The docs claim that `-s` on `git commit` avoids this. I have been trying to do that, but still forget quite often. image

Yes, when remembering this is fine, otherwise you have to rebase and sign the commits that way.  I've not found a way to aadd the -s when using the GitHub interface though

stas00:
> @stas00 This is a requirement from the Linux Foundation for their projects.

but then how can you override it? https://github.com/deepspeedai/DeepSpeed/pull/7024#issuecomment-2652302500

stas00:
but my commit is signed as far as github goes! it has the verified check mark next to my commit. 

I suppose it just doesn't bother to add the header and so non-native GH apps fail to match the expectations. 

A bug in GH?

loadams:
> > @stas00 This is a requirement from the Linux Foundation for their projects.
> 
> but then can you override it if it's a requirement? [#7024 (comment)](https://github.com/deepspeedai/DeepSpeed/pull/7024#issuecomment-2652302500)

Having DCO as a policy is the requirement.  I think the fact that you can override the DCO bot means roughly the same thing as signing off on the commit but I'm not sure.

loadams:
> but my commit is signed as far as github goes! it has the verified check mark next to my commit.
> 
> I suppose it just doesn't bother to add the header and so non-native GH apps fail to match the expectations.
> 
> A bug in GH?

I think it is that it doesn't pass the -s flag.  I suspect" if your commit message on GH contained the "Signed-off-by: Random J Developer <random@developer.example.org>" at the end, it might work?

stas00:
we don't know how it performs the online commits, probably not using cli? don't know.

but no, if you look at this, there is no signed-off header
https://github.com/deepspeedai/DeepSpeed/commit/c1283f0d8c3c088a4bb5567f7fc557109ae3d333.patch
it doesn't bother with it, since it already knows it's signed off as the commit was done via an authenticated online account. 

but the problem is that this breaks other non-native apps that want this header.

loadams:
I think there is a difference between the actual signing of the commit and the "signed off" for DCO though, since [this commit](https://[github.com/deepspeedai/DeepSpeed/commit/8f9c4669f02dc0135a387252eddfc76c7992ef44.patch](https://github.com/deepspeedai/DeepSpeed/commit/8f9c4669f02dc0135a387252eddfc76c7992ef44.patch)) was signed off on the command line, but I think all the DCO bot is looking for is the "Signed-off-by" in the commit body?

PR #7027, opened 2025-02-12
[ROCm] Enable fp_quantizer on ROCm

rraminen:
This change is required to successfully build fp_quantizer extension on ROCm.


op_builder/fp_quantizer.py
```diff
@@ -95,6 +95,10 @@ def sources(self):
 
     def extra_ldflags(self):
         return ['-lcurand']
```
loadams, line 97: @rraminen - line 97 should be removed as well?

loadams (approved):

Issue #7031, opened 2025-02-13
Fix - Update DeepSpeed to be PEP517 compliant, update to `pyproject.toml`


loadams:
We need to update DeepSpeed to be compliant with [PEP 517](https://peps.python.org/pep-0517/) as well as for handling of [future changes to pip](https://github.com/pypa/pip/issues/11457).

felker:
FYI with `pip` 25.3 released today, there will probably be more Issues filed about source builds failing due to DeepSpeed's lack of PEP517 compliance 

the docs should be updated [here](https://www.deepspeed.ai/tutorials/advanced-install/) to change
```
DS_BUILD_OPS=1 pip install deepspeed --global-option="build_ext" --global-option="-j8"
```
to
```
DS_BUILD_OPS=1 pip install deepspeed -C="--global-option=build_ext" -C="--build-option=-j8"--no-build-isolation
```
no?

sowndappan5:
I tested the documented installation command on Windows using Python 3.10 and pip 26.1.2:

```pip install deepspeed --global-option build_ext --global-option -j8```

and pip immediately returns:

```no such option: --global-option```

Interestingly, `pip install --help` still lists `--global-option`, but the command is not accepted during execution in my environment.

This may be another data point supporting the need to revisit the installation instructions for newer pip versions.

PR #7032, opened 2025-02-13
Update to support pyproject.toml

loadams:
- [ ] Test all build cases this needs to support.
- [ ] Test Windows builds/support.
- [ ] Confirm pre-compiling ops

This PR was quickly updated with `hatch` to build the pyproject.toml as a starting point.

Fixes: #7031


pyproject.toml
```diff
@@ -0,0 +1,166 @@
+[build-system]
+requires = ["hatchling"]
```
loadams, line 2: Switch to setuptools

PR #7033, opened 2025-02-13
Add pyproject.toml

loadams:
- [ ] Test all build cases this needs to support.
- [ ] Test Windows builds/support.
- [ ] Confirm pre-compiling ops

This PR was quickly updated with `hatch` to build the pyproject.toml as a starting point.

Fixes: #7031

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,9 @@
+[build-system]
+requires = [
+    "setuptools",
+    "wheel",
+    "ninja",
+    "typing-extensions>=4.10.0",
+]
+# Use legacy backend to import local packages in setup.py
+build-backend = "setuptools.build_meta:__legacy__"


pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[build-system]
+requires = [
+    "setuptools",
```
stas00, line 3:
```suggestion
    "setuptools>=64",
```

as suggested at https://github.com/pypa/pip/issues/11457 - scroll down to What should I do?

jeffra:
@mrwyattii we just went through some of this with arctic training. If it’s helpful @loadams let’s discuss on slack a bit. There’s a ton that’s currently happening in setup.py, this could be a big lift? But I agree, needs to happen!

loadams:
The current problem is that the logic inside setup.py aside from the call to `setup()` isn't run.  This means that we don't append cupy into the requirements file, and tests fail as a result.  This impacts other parts of the build experience, so we will need to do more work to switch to a modern build backend.

sfc-gh-mwyatt (approved):

pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[build-system]
+requires = [
+    "build",
+    "setuptools>=64",
+    "torch",
+    "wheel"
```
agronholm, line 6: If you depend on setuptools 70.1 or later, you won't need `wheel`.
```suggestion
    "setuptools>=70.1",
    "torch"
```

PR #7050, opened 2025-02-18
Bug Fix for offload_states API

U-rara:
@fukun07 and I discovered a bug when using the `offload_states` and `reload_states` APIs of the Zero3 optimizer. When using grouped parameters (for example, in weight decay or grouped lr scenarios), the order of the parameters mapping in `reload_states` ([here](https://github.com/deepspeedai/DeepSpeed/blob/14b3cce4aaedac69120d386953e2b4cae8c2cf2c/deepspeed/runtime/zero/stage3.py#L2953)) does not correspond with the initialization of `self.lp_param_buffer` ([here](https://github.com/deepspeedai/DeepSpeed/blob/14b3cce4aaedac69120d386953e2b4cae8c2cf2c/deepspeed/runtime/zero/stage3.py#L731)), which leads to misaligned parameter loading. This issue was overlooked by the corresponding unit tests ([here](https://github.com/deepspeedai/DeepSpeed/blob/master/tests/unit/runtime/zero/test_offload_states.py)), so we fixed the bug in our PR and added the corresponding unit tests.


tests/unit/runtime/zero/test_offload_states.py
```diff
@@ -124,5 +124,8 @@ def test_offload_states(self, included_state, pin_memory, non_blocking):
         with deepspeed.zero.Init(config_dict_or_path=config_dict):
             model = SimpleModel(hidden_dim, nlayers=4)
 
+        params = list(model.parameters())
+        param_groups = [{"params": params[::2], "weight_decay": 0.0}, {"params": params[1::2], "weight_decay": 0.1}]
```
tjruwase, line 128: I recommend improving readability by making things more explicit like so:

```python
  param_groups = [
      {"params": [p for n, p in model.named_parameters() if not 'bias' in n], "weight_decay" : 0.1}, 
      {"params": [p for n, p in model.named_parameters() if 'bias' in n], "weight_decay" : 0.0}
  ]
```

On a minor note, I also flipped the weight decay to weights instead of biases :).

U-rara:
@tjruwase Good suggestion! My previous fix seemed a bit hasty, so I optimized it according to your advice.

tohtana (approved):
Great catch, thank you @U-rara!

PR #7055, opened 2025-02-19
Fix, pipeline model with moe cause error when send grad 

wukong1992:
Fix, pipeline model with moe, send and recv grad use same buffer will cause error when moe loss not in position 1. Solution, activation and grad use diff buffer

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -992,7 +992,7 @@ class PipelineEngine(DeepSpeedEngine):
             recv_dtype = self.ID_TO_DTYPE[buffer[1].item()]
             recv_ndims = buffer[2].item()
             recv_shape = buffer[3:3 + recv_ndims].tolist()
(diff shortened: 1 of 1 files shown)


wukong1992:
@tohtana Could you review the merge requests?

loadams:
Hi @wukong1992 - looks like your changes are failing the formatting check - could you either run the pre-commit formatter locally or take a look at [the errors](https://github.com/deepspeedai/DeepSpeed/actions/runs/13441878629/job/37558109511?pr=7055#step:5:70)?  It looks like your changes mean we can remove an import and flake8 caught that.

tohtana:
Hi @wukong1992, Thank you for this great fix, and sorry for my late response.

I wonder if some behaviors might change with this PR other than solving the issue. `_allocate_or_extend_buffers` reuses the buffer when possible. Your change seems to allocate memory every time. Honestly, I'm not sure how much it affects the performance, but afraid that it can increase memory fragmentation. Can you share any thought on this?

PR #7062, opened 2025-02-20
Avoid graph breaks in torch.compile caused by inner classes in the backward hooks

deepcharm:
This PR is part of the effort to improve Deepspeed performance when using PyTorch compile.

There is a known [bug](https://github.com/pytorch/pytorch/issues/128942) in torch.compile which causes a graph break when an inner class is defined within
a method that is being compiled. The following would then appear in the log:

`[__graph_breaks] torch._dynamo.exc.Unsupported: missing: LOAD_BUILD_CLASS`

This is the case with the inner classes `PreBackwardFunctionForModule` and `PostBackwardFunctionModule`.

While there is an open PyTorch [PR#133805 ](https://github.com/pytorch/pytorch/pull/133805) for this, we can solve the issue by moving the inner classes into the initialization code.

No graph breaks and the corresponding logs are produced anymore.


tjruwase (approved):

tjruwase:
@deepcharm, thanks for the PR. Please resolve the DCO 
https://github.com/deepspeedai/DeepSpeed/pull/7062/checks?check_run_id=37557944610

tjruwase:
@deepcharm, gentle ping on DCO blocker

deepcharm:
> @deepcharm, gentle ping on DCO blocker

@tjruwase DCO should be fine now, sorry for the delay

PR #7069, opened 2025-02-21
Replace calls to `python setup.py sdist` with `python -m build --sdist`

loadams:
With future changes coming to pip/python/etc, we need to modify to no longer call `python setup.py ...` and replace it instead: https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#should-setup-py-be-deleted

image

This means we need to install the build package which is added here as well.

Additionally, we pass the `--sdist` flag to only build the sdist rather than the wheel as well here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,7 @@ jobs:
     - name: Build DeepSpeed
       run: |
         pip install setuptools
+        pip install build
         DS_BUILD_STRING=" " python -m build --sdist
     - name: Publish to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Build DeepSpeed
       run: |
         pip install setuptools
-        DS_BUILD_STRING=" " python -m build
+        DS_BUILD_STRING=" " python -m build --sdist
     - name: Publish to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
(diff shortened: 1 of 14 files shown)


.github/workflows/release.yml
```diff
@@ -26,7 +26,8 @@ jobs:
     - name: Build DeepSpeed
       run: |
         pip install setuptools
-        DS_BUILD_STRING=" " python setup.py sdist
+        pip install build
+        DS_BUILD_STRING=" " python -m build --sdist
```
loadams, line 30: Todo: test the release workflow to ensure this package is building things as expected.

build_win.bat
```diff
@@ -11,6 +11,6 @@ set DS_BUILD_GDS=0
 set DS_BUILD_RAGGED_DEVICE_OPS=0
 set DS_BUILD_SPARSE_ATTN=0
 
-python setup.py bdist_wheel
+python -m build --wheel
```
loadams, line 14: Todo: test on Windows

loadams: This builds and runs on Windows just fine.

sfc-gh-mwyatt:
I think there are a few places in docs that need to be updated as well (e.g., https://github.com/deepspeedai/DeepSpeed/blob/9d820e4c053160af0136a9cfc3a3a55d4297a023/docs/_tutorials/ds-sequence.md?plain=1#L114). Otherwise LGTM

sfc-gh-mwyatt (approved):

.github/workflows/release.yml
loadams: DS_BUILD_STRING works to pass the string.

PR #7073, opened 2025-02-24
Fix meta load tensor imcompatible issue

Yejing-Lai:
The partition tensor doesn't need to move to the current device when meta load is used.

### DIFF
--- a/deepspeed/module_inject/layers.py
+++ b/deepspeed/module_inject/layers.py
@@ -48,7 +48,8 @@ def move(tensor, device):
     # to save host resources when DP > 1。
 
     if tensor.is_meta:
-        return torch.empty_like(tensor, device=device)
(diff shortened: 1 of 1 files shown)


Yejing-Lai:
@loadams @inkcherry Please review~

deepspeed/module_inject/layers.py
```diff
@@ -48,7 +48,8 @@ def move(tensor, device):
     # to save host resources when DP > 1。
 
     if tensor.is_meta:
-        return torch.empty_like(tensor, device=device)
+        # Keep tensor in meta device if tensor is meta.
```
inkcherry, line 51: LGTM thanks

inkcherry (commented):
LGTM, thanks!

Issue #7082, opened 2025-02-26
[BUG] Bad code sample in Inference tutorial


riedgar-ms:
**Describe the bug**

The code in the section:
https://www.deepspeed.ai/tutorials/inference-tutorial/#initializing-for-inference
appears to be incorrect.

Specifically, in the first code block of that section
```python
model = ds_engine.module
output = model('Input String')
```
should probably be:
```python
model = ds_engine.module
pipe = pipeline("text-generation", model=model, tokenizer=tokenizer)
print(f"{pipe('Hello')}")
```
(or something similar). The code in
https://www.deepspeed.ai/tutorials/inference-tutorial/#end-to-end-gpt-neo-27b-inference
is closer to what's required.

**To Reproduce**

I attempted to use the code sample as the basis for my own script.

**Expected behavior**

Code sample to work

**ds_report output**
Please run `ds_report` to give us details about your setup.

```
torch install path ............... ['/azureml-envs/azureml_28a03b9f9b7c5f401fdd179f0d3ee4d8/lib/python3.10/site-packages/torch']
torch version .................... 2.6.0+cu124
deepspeed install path ........... ['/azureml-envs/azureml_28a03b9f9b7c5f401fdd179f0d3ee4d8/lib/python3.10/site-packages/deepspeed']
deepspeed info ................... 0.16.4, unknown, unknown
torch cuda version ............... 12.4
torch hip version ................ None
nvcc version ..................... 12.1
deepspeed wheel compiled w. ...... torch 0.0, cuda 0.0
shared memory (/dev/shm) size .... 2.00 GB
```

**Screenshots**

**System info (please complete the following information):**
 - OS: Ubuntu
 - GPU count and types [e.g. two machines with x8 A100s each]: Single machine with 4xA100
 - (if applicable) what [DeepSpeed-MII](https://github.com/deepspeedai/deepspeed-mii) version are you using
 - (if applicable) Hugging Face Transformers/Accelerate/etc. versions
 - Python version: 3.10
 - Any other relevant info about your setup

**Docker context**
Are you using a specific docker image that you can share?

**Additional context**

Running in AzureML.

PR #7083, opened 2025-02-26
Improve inference tutorial docs

loadams:
Fixes: #7082

### DIFF
--- a/docs/_tutorials/inference-tutorial.md
+++ b/docs/_tutorials/inference-tutorial.md
@@ -21,18 +21,22 @@ if args.pre_load_checkpoint:
     model = model_class.from_pretrained(args.model_name_or_path)
 else:
     model = model_class()
+
+# create the tokenizer
+tokenizer = model_class.from_pretrained(args.model_name_or_path)
 ...
 
 import deepspeed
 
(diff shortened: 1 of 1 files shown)


jomayeri (approved):

docs/_tutorials/inference-tutorial.md
```diff
@@ -21,18 +21,22 @@ if args.pre_load_checkpoint:
...
 
 import deepspeed
 
 # Initialize the DeepSpeed-Inference engine
 ds_engine = deepspeed.init_inference(model,
-                                 tensor_parallel={"tp_size": 2},
-                                 dtype=torch.half,
-                                 checkpoint=None if args.pre_load_checkpoint else args.checkpoint_json,
-                                 replace_with_kernel_inject=True)
+                                     tensor_parallel={"tp_size": 2},
```
riedgar-ms, line 33: Small point, but if you're making changes, should this '2' become something like `world_size` ? It wouldn't make the example runnable, but would at least make the expected value explicit.

loadams: Yes, plus that better matches the other examples in this file.

PR #7088, opened 2025-02-27
Use transformers latest on v100 tests

loadams:
Reverts #7085 which pinned the transformers version due to likely breaking changes from https://github.com/huggingface/transformers/pull/36335, fixes introduced in  https://github.com/huggingface/transformers/pull/36455

Error:
```
RuntimeError: Failed to import transformers.models.vision_encoder_decoder.modeling_vision_encoder_decoder because of the following error (look up to see its traceback)
```

```
FAILED unit/runtime/zero/test_zero_nesting_init.py::TestNestedParallelInit::test_nested_parallel_init - RuntimeError: Error(s) in loading state_dict for VisionEncoderDecoderModel:
```

After these changes, we still see errors on some tests with all_close


loadams:
nv-a6000 test appears to be failing because transformers calls `from torch.distributed.tensor import DTensor` which doesn't exist prior to torch 2.5.

tohtana (approved):

PR #7093, opened 2025-02-28
Update gaudi2 nightly,ci to latest 1.20.0 build

raza-sikander:


loadams (approved):

loadams:
Hi @raza-sikander - could you fix the DCO or let us know if you agree and we can sign it off for you?  

Also it might be failing/hanging on the HPU-nightly tests?

raza-sikander:
@loadams  thanks i have fixed the DCO.
It shouldnt be failing on the nighlty tests. Tested them before pushing.

loadams:
> @loadams thanks i have fixed the DCO. It shouldnt be failing on the nighlty tests. Tested them before pushing.

Thanks, re-running the tests now.  Though we might need to update the firmware on the HPU that we're testing on?  Could that lead to hangs?

raza-sikander:
> Thanks, re-running the tests now. Though we might need to update the firmware on the HPU that we're testing on? Could that lead to hangs?

Yes we may need to update the firmware on HPU if hangs are observed.

loadams:
> > Thanks, re-running the tests now. Though we might need to update the firmware on the HPU that we're testing on? Could that lead to hangs?
> 
> Yes we may need to update the firmware on HPU if hangs are observed.

@raza-sikander - the nightly test looks to be on its way to hanging, its been running for 170 mins already and will probably hit our 300 minute timeout default.  

Do you know what version we need to update the firmware on the runner to?  1.20?

loadams:
> > > Thanks, re-running the tests now. Though we might need to update the firmware on the HPU that we're testing on? Could that lead to hangs?
> > 
> > 
> > Yes we may need to update the firmware on HPU if hangs are observed.
> 
> @raza-sikander - the nightly test looks to be on its way to hanging, its been running for 170 mins already and will probably hit our 300 minute timeout default.
> 
> Do you know what version we need to update the firmware on the runner to? I only see 1.18 (which is the current test machine is running) and 1.19 and 1.19.1.

@raza-sikander - updating to 1.19.1 seems to have resolved the issue.

PR #7103, opened 2025-03-03
[XPU]Support xccl on xpu device

ys950902:
For our XPU software stack will using XCCL on torch path, and no need to import ipex in the future, so add this feature to support XCCL.


loadams:
@ys950902 - could you also resolve the DCO or let us know if you agree and we can fix it?

PR #7113, opened 2025-03-06
[XPU] Support XCCL on deepspeed side

ys950902:
XCCL will be used for XPU device in the future, we will also reserve the old path for torch-CCL enable.


loadams:
Hi @ys950902 - thanks for fixing the DCO errors.  Could you run the pre-commit formatter on the PR?  `pre-commit run --all-files`?  That will fix the formatting issues.

For the no-torch and python tests, I believe we are running those tests with too old of a version of torch, before xpu support was added there.  We can try to update these tests, but should we check has_attr xpu first?

loadams:
@ys950902 - let us know if you want to discuss more on the backwards compat or have questions on the no-torch test failures.

delock:
@ys950902  can you take a look at no-torch failures?

loadams:
> Hi @ys950902 can you take a look at no-torch failures?

@ys950902 - reminder on the no-torch failure here.

PR #7115, opened 2025-03-07
async tp allreduce 

inkcherry:


hwchen2017 (commented):
Please fix the format error by running pre-commit and DCO error.

deepspeed/module_inject/layers.py
```diff
@@ -382,11 +411,15 @@ def __init__(self, module, mp_group=None, skip_partition=False, **kwargs):
             self.config_tp_params(self.bias)
 
     def forward(self, input):
-        if getattr(self, 'mp_group', None) is not None:
-            input = ColumnParallel.apply(self.mp_group, input)
-        output = torch.matmul(input, self.weight.transpose(-1, -2))
-        if self.bias is not None:
-            output += self.bias
+        if True:
```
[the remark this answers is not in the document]
inkcherry: thanks, added.

inkcherry:
> Please fix the format error by running pre-commit and DCO error.

fixed.

inkcherry:
It seems that the UT failure is not related to this PR, could you please take a look. @hwchen2017  Thanks!

deepspeed/module_inject/layers.py
```diff
@@ -124,11 +153,17 @@ class TensorParallel_Layer(nn.Module, ABC):
...
 
-    # keep_module_on_host is used to keep the module on the host. Checkpoints are loaded to the host first (in some
-    # cases it can be done from the disk even to prevent filling host's memory), thus no need to create a new copy.
+    # keep_module_on_host determines whether to keep the module on the host.
+    # Checkpoints are first loaded to the host (sometimes directly from disk to avoid filling host memory),
+    # so an additional copy is unnecessary.
     keep_module_on_host: bool = False
 
+    ##### Runtime Parameter List #####
+    overlap_comm: bool = False
```
hwchen2017, line 164: I believe that deepspeed already use this flag for computation communication overlap in zero series. So it makes more sense to have a slightly different flag like `tp_overlap_comm` or `tp_bwd_overlap_comm`. I plan to make the forward all reduce asynchronous, but haven't test it yet.

inkcherry: Thanks, I have modified ```tp_overlap_comm```. I also gave it some thought, and if we want to achieve batch-level overlap without rewriting the model, we indeed need to further split inputs like ```attn_mask``` and ```position```. These tensors should typically change with the batch partitioning.

hwchen2017: More changed are needed for the integration of AutoTP with Domino because the rowlinear and colmnlinear layers have different implementation. 

Can you contribute your training examples to the DeepspeedExample repo [here](https://github.com/deepspeedai/DeepSpeedExamples/tree/master/training)? It will be very helpful. Thanks!

inkcherry: sure, for this repo https://github.com/inkcherry/stanford_alpaca, all you need to do is 

1. update the transformer/accelerate to master branch
2.  then change the config like this https://github.com/inkcherry/stanford_alpaca/blob/tp_demo_1127/run.sh

 I will add it  to DeepspeedExample later~

hwchen2017 (approved):

PR #7139, opened 2025-03-14
Unpin transformers version for most workflows

loadams:
Unpin transformers version for all workflows except `nv-torch-latest-v100` as this still has a tolerance issue with some quantization tests.


loadams:
`nv-torch-nightly` isn't running currently so we will bypass that test.

PR #7154, opened 2025-03-19
DeepCompile for enhanced compiler integration

tohtana:
This PR introduces *DeepCompile*, a new feature that efficiently integrates compiler optimizations with other DeepSpeed features. DeepCompile utilizes torch's dynamo to capture the computation graph and modifies it to incorporate DeepSpeed’s optimizations seamlessly.

Currently, DeepCompile supports ZeRO-1 and ZeRO-3, with enhancements such as proactive prefetching and selective unsharding to improve performance.
(More details will be added later.)


delock:
Hi @tohtana , fantastic work on the DeepCompile feature in DeepSpeed! 🚀
I noticed DeepCompile leverages the DeepSpeed OpBuilder for C++ kernels. Would there be plans to provide guide detailing the kernel interfaces? This could greatly help accelerator owners port this feature to their hardware.
Thanks for building this awesome capability!

tohtana:
Hi @delock, Thank you for your comment!

Yes, we definitely plan to extend support beyond CUDA to other accelerators. A good starting point would be defining proper dispatch to accelerator-specific operators [here](https://github.com/deepspeedai/DeepSpeed/blob/584c801dc20e7c3220973c5228cb0ef061856f76/csrc/compile/init.cpp#L30), although some operators are common across multiple platforms.

This project is still in its very early stages, and we would be happy to collaborate with you all on designing multi-platform support.

.github/workflows/nv-pre-compile-ops.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             #python -c "import torch; print('CUDA available:', torch.cuda.is_available())"
         - name: Compile DeepSpeed Ops
           run: |
-            DS_ACCELERATOR=cuda DS_ENABLE_NINJA=1 TORCH_CUDA_ARCH_LIST="7.0;7.5;8.0" DS_BUILD_OPS=1 DS_BUILD_SPARSE_ATTN=0 DS_BUILD_FP_QUANTIZER=0 DS_BUILD_CUTLASS_OPS=0 DS_BUILD_GDS=0 DS_BUILD_RAGGED_DEVICE_OPS=0 DS_BUILD_EVOFORMER_ATTN=0 pip3 install .
+            DS_ACCELERATOR=cuda DS_ENABLE_NINJA=1 TORCH_CUDA_ARCH_LIST="7.0;7.5;8.0" DS_BUILD_OPS=1 DS_BUILD_SPARSE_ATTN=0 DS_BUILD_FP_QUANTIZER=0 DS_BUILD_CUTLASS_OPS=0 DS_BUILD_GDS=0 DS_BUILD_RAGGED_DEVICE_OPS=0 DS_BUILD_EVOFORMER_ATTN=0 DS_BUILD_DEEP_COMPILE=0 pip3 install .
```
loadams, line 39: Could you also set `DS_BUILD_DEEP_COMPILE=0` in the build_win.bat script (unless this is supported on Windows?)

tohtana: Thank you for pointing this out! I've added the flag.

deepspeed/runtime/engine.py
```diff
@@ -3865,10 +3889,50 @@ def compile(self, backend=get_accelerator().get_compile_backend(), compile_kwarg
...
+                        assert callable(p) or p in opt_passes, f"Unknown pass {p}"
+                    return [p if callable(p) else opt_passes[p] for p in passes]
+
+                schedule = [(step, passes_name_to_fn(passes)) for step, passes in schedule]
+
+            assert backend in ['inductor', 'eager'], f"Backend {backend} is not supported for DeepCompile."
+
+            compile_config = self._config.compile_config
+            if (("zero_optimization" in self.config and "offload_optimizer" in self.config["zero_optimization"]
+                 and "offload_param" in self.config["zero_optimization"])
```
tjruwase, line 3912: Does this mean, offloading is only possible if both param and opt_state are offloaded to cpu in ds_config?

tjruwase: What is the downside of enabling offloading when only opt_state is offloaded to cpu?

deepspeed/compile/util.py
```diff
@@ -0,0 +1,429 @@
...
+
+
+def log_rank0(msg: str, enable: bool = False):
+    if dist.get_rank() == 0 and enable:
+        print(msg)
+
+
+def get_no_copy_ops():
+    # Need to compile custom ops
+    get_deepcompile_handle()
```
tjruwase, line 81: Is this a no-op?

deepspeed/comm/torch.py
```diff
@@ -409,6 +409,13 @@ def init_device_mesh(self, mesh_shape, mesh_dim_names):
                                                                   mesh_shape,
                                                                   mesh_dim_names=mesh_dim_names)
 
+    def enable_symm_mem_for_group(self, group_name):
+        if not required_torch_version(min_version=2.5):
+            raise RuntimeError(f"Current torch version does not have enable_symm_mem_for_group"
```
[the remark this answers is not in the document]
tohtana: Thank you, updated the error message to include the required version.

deepspeed/compile/util.py
tohtana: `get_deepcompile_handle()` adds operators under `torch.ops.dc` at runtime. We need to make sure `get_deepcompile_handle()` is called before we reference any operators in `torch.ops.dc`.

deepspeed/runtime/engine.py
tohtana: Setting `compile_config.offload_parameters = True` enables a behavior similar to ZeRO-3’s parameter offloading, which isn’t featured in this initial release. It works independently of the optimizer state offloading described in the blog and paper.
In theory, enabling both optimization passes (one for parameters and one for optimizer states) should behave like ZeRO-3’s full offloading setup. However, this hasn’t been tested yet.

deepspeed/compile/passes/offload_parameters.py
```diff
@@ -0,0 +1,54 @@
...
+                                        torch.ops.dc.offload_parameter.default,
+                                        args, {},
+                                        name="offload_parameter")
+
+    return new_node
+
+
+def add_reload_parameter(graph_id: int, gm: GraphModule, node: Node, ds_id: int):
+    new_node = None
+    with gm.graph.inserting_after(node):
```
tjruwase, line 30: For my education, should this not be `graph.inserting_before()`?

tohtana: In this case, arg `node` is an argument of allgather, which is ds parameter in the partitioned state. The path in the graph will be: ds_param -> [inserted reload op] -> allgather.
You could also do the same with `inserting_before(allgather_node)`.
@AliZafar120 Please leave a comment if you have something to add.

tjruwase (approved):

deepspeed/compile/util.py
tjruwase: This seems like a tricky side-effect situation that developers can miss. It would be great to either have informative asserts and/or docs to help.

tohtana: Agreed. It should be helpful to provide a helper function to access custom ops and encourage users to use it in a doc. Let's discuss more later.

deepspeed/compile/passes/offload_parameters.py
AliZafar120: inserting_before works in backend=eager, 
However, in backend=inductor there is an issue.

In inductor settings,
when I insert before the allgather_node and use the parameter node as an argument to show its relationship with the graph, it causes an error. It might be because the node tensor has not yet been moved to the GPU and the reload_node will transfer the tensor to GPU. It therefore gives the following error.

"ValueError: Pointer argument (at 3) cannot be accessed from Triton (cpu tensor?)".

From my understanding in Inductor, allgather_node should be in GPU so inserting_before expects the inserted node and its argument tensor to be in GPU.

deepspeed/compile/init_z3.py
```diff
@@ -0,0 +1,94 @@
...
+    # Unpatch linear
+    if hasattr(InsertPostInitMethodToModuleSubClasses, "linear_bk"):
+        torch.nn.functional.linear = InsertPostInitMethodToModuleSubClasses.linear_bk
+
+    if compile_config.symmetric_memory:
+        group_name = engine.data_parallel_group.group_name
+        dist.enable_symm_mem_for_group(group_name)
+
+    for p in engine.module.parameters():
+        grad_buffer = optimizer._DeepSpeedZeroOptimizer_Stage3__param_id_to_grad_partition[p.ds_id]
```
bsochack, line 51: @tohtana sorry for asking questions on the old PR.
I'm trying to enable DeepCompile on one of configurations and I run into error:
image

based on the code analysis, I see that __param_id_to_grad_partition is initialized based on a slightly different parameters.
Should the loop iterate over list(itertools.chain.from_iterable(optimizer.fp16_groups)) instead of engine.module.parameters()?

PR #7163, opened 2025-03-21
fixed: Modified the topkgating function and modified the test_moe file for testing

xiongjyu:
Since the previous PR encountered the DCO problem and could not be solved for some reason, I resubmitted a completely identical PR but without the problem.


tjruwase:
@xiongjyu, please fix the formatting errors.

tohtana (approved):

Issue #6709, opened 2024-11-04
"__nv_bfloat162" has already been defined


wolfljj:
system:windows 11
build deepspeed use build_win.bat

error: class "__nv_bfloat162" has already been defined (previous definition at line 4787 of C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.6/include\cuda_bf16.h)
  using __nv_bfloat162 = __half2;
        ^
1 error detected in the compilation of "csrc/transformer/inference/csrc/gelu.cu".
error: command 'C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v12.6\\bin\\nvcc' failed with exit code 4294967295

loadams:
Hi @wolfljj - can you share more information about your system?  Torch version, pip list, nvcc version, msvc version?

vinodkvijayan:
- system : windows 11
- Python version : 3.12.7
- Torch : 2.5.1+cu124
- nvcc :  cuda_12.6.r12.6/compiler.34841621_0
- msvc/cl : 19.42.34433 for x64
- - build deepspeed use build_win.bat

I too had the same error

error: class "__nv_bfloat162" has already been defined (previous definition at line 4787 of C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.6/include\cuda_bf16.h)
using __nv_bfloat162 = __half2;

I could get the wheel built by commenting out the __nv_bfloat162 definition in both 

- DeepSpeed\csrc\transformer\inference\csrc\gelu.cu`  and 
- DeepSpeed\csrc\transformer\inference\csrc\transform.cu, 

like this

// only used to avoid compilation error due to lack of definition.
// #ifndef BF16_AVAILABLE
// using __nv_bfloat162 = __half2;
// #endif

I'm not sure whether this will work for everyone.

vinodkvijayan:
- system : windows 11
- Python version : 3.12.7
- Torch : 2.5.1+cu124
- nvcc :  cuda_12.6.r12.6/compiler.34841621_0
- msvc/cl : 19.42.34433 for x64
- build deepspeed use build_win.bat

I too had the same error

error: class "__nv_bfloat162" has already been defined (previous definition at line 4787 of C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.6/include\cuda_bf16.h)
using __nv_bfloat162 = __half2;

[...]

PR #7168, opened 2025-03-22
Fix pre-compile on cpu-only machines

AlongWY:
+ Fix pre-compile on cpu-only machines


accelerator/cuda_accelerator.py
```diff
@@ -242,6 +242,8 @@ def communication_backend_name(self):
         return self._communication_backend_name
 
     def is_triton_supported(self):
+        if not torch.cuda.is_available():
```
tjruwase, line 245:
```suggestion
        if not self.is_available():
```

tjruwase: Please change as above

tjruwase:
@AlongWY, thanks for the PR. Can you please share the motivation to help us understand your use case?

AlongWY:
> @AlongWY, thanks for the PR. Can you please share the motivation to help us understand your use case?

I used GitHub Actions to precompile DeepSpeed, but after upgrading to v0.16.4, the precompilation failed. The repository is here: https://github.com/AlongWY/deepspeed_wheels

loadams:
> > @AlongWY, thanks for the PR. Can you please share the motivation to help us understand your use case?
> 
> I used GitHub Actions to precompile DeepSpeed, but after upgrading to v0.16.4, the precompilation failed. The repository is here: https://github.com/AlongWY/deepspeed_wheels
> 
> And the [compile result](https://github.com/AlongWY/deepspeed_wheels/actions/runs/14005965046/job/39219890195#step:10:5216) here show that tranformer inference is not support cuda 12.6, which is also shown in #6709.

@AlongWY - just so I follow, you're compiling on Windows in a GitHub action that has cuda installed but no GPU (so that's why you are hitting the CPU only path here) but with DS_ACCELERATOR=cuda?  Just wanted to make sure I understood the repo correctly?

AlongWY:
Yes, the GitHub Action is compiling in an Ubuntu environment with CUDA installed but without a GPU.  
The GitHub Action runs only on Ubuntu, but the error still occurs.

AlongWY:
Yes, the triton has been installed in the environment.

# Error for triton
image

# Error for type
image

AlongWY:
Any update?

sfc-gh-truwase:
> Any update?

@AlongWY apologies for the delay. This overall looks good to me. I will appreciate a couple more clarifications. 
1. I am curious why [accelerator detection logic](https://github.com/deepspeedai/DeepSpeed/blob/cda3f9628cfe1dd90c7af10ddc5b8d3b16aa459c/accelerator/real_accelerator.py#L51) is incorrectly [identifying as cuda env.](https://github.com/deepspeedai/DeepSpeed/blob/cda3f9628cfe1dd90c7af10ddc5b8d3b16aa459c/accelerator/real_accelerator.py#L201). Or are you hardcoding DS_ACCELERATOR=cuda` in your code?
2. Are you aligned with the suggested code change?

AlongWY:
@sfc-gh-truwase Hi, i have update the code.

For the two questions:
1. I use GitHub Actions to automatically build wheel packages for different environments, so I manually set the environment variable DS_ACCELERATOR=cuda.  
2. I have aligned with the suggested code change.

Issue #5242, opened 2024-03-08
[BUG] grad_norm and loss is nan when deepspeed==0.13.5 but ok with deepspeed==0.10.2


Chandler-Bing:
**Describe the bug**
when fine-tuning my model using deepspeed==0.13.5, and huggingface trainer, loss and grad_norm will be nan at step 2
![image](https://github.com/microsoft/DeepSpeed/assets/29994840/cf2d7a6b-91df-43d6-9706-aa82c2dbf074)

but 2 ways below could solve the problem
1. deepseed==0.10.2
2. add this to my deepspeed config, (which slow down my training speed)
```
"comms_logger": {
  "enabled": true,
  "verbose": false,
  "prof_all": true,
  "debug": false
}
```
why this happen? 

**To Reproduce**
1. my run script
```
deepspeed \
    pretraining.py \
    --model_type auto \
    --model_name_or_path /app/nfs_share_dir/3/llm_model/Baichuan2-7B-Base \
    --train_file_dir /app/nfs_share_dir/1/archive/v2/token-baichuan/tmp \
    --validation_file_dir /app/nfs_share_dir/1/archive/v2/token-baichuan/tmp \
    --lazy_mode True \
    --per_device_train_batch_size 8 \
    --per_device_eval_batch_size 8 \
    --do_train \
    --do_eval \
    --seed 3 \
    --warmup_ratio 0.01 \
    --num_train_epochs 1 \
    --learning_rate 2e-5 \
    --lr_scheduler_type cosine \
    --weight_decay 1e-4 \
    --logging_strategy steps \
    --logging_steps 1 \
    --save_steps 1000 \
    --save_strategy steps \
    --save_total_limit 10 \
    --gradient_accumulation_steps 1 \
    --block_size 4096 \
    --torch_compile True \
    --output_dir outputs_qwen \
    --overwrite_output_dir \
    --ddp_timeout 30000 \
    --logging_first_step True \
    --log_on_each_node 0 \
    --torch_dtype bfloat16 \
    --report_to tensorboard \
    --ddp_find_unused_parameters False \
    --gradient_checkpointing True \
    --deepspeed ./config/ds_2_config.json \
    --bf16 \
    --bf16_full_eval
```
2.  deepspeed config
```
{
    "optimizer": {
        "type": "AdamW",
        "params": {
            "lr": "auto",
            "weight_decay": "auto",
            "torch_adam": true,
            "adam_w_mode": true
        }
    },
    "bf16": {
[...]

PR #7171, opened 2025-03-25
Fix issue #5242 grad_norm and loss is nan

Glaceon-Hyy:
This PR addresses a regression introduced in commit [61daaa1](https://github.com/deepspeedai/DeepSpeed/commit/61daaa1ea211e3a114250a7cf7117cd0c8fecd5e) that affects gradient clipping when handling infinite values.

The modified NaN/Inf handling logic in total_norm calculation leads to unexpected behavior:

Original logic ([v0.10.3](https://github.com/deepspeedai/DeepSpeed/blob/v0.10.3/deepspeed/runtime/zero/stage_1_and_2.py#L1233)): Converted both NaN and Inf to -1 before entering unscale_and_clip_grads
Post-commit behavior: When total_norm is Inf, inf_or_nan.logical_not() * total_norm produces NaN instead of 0, causing gradient clipping to fail

Here is a minimal reproducible example comparing gradient clipping behavior across implementations.
```python
import torch
import numpy as np
import copy

def test(total_norm):
    test_old_deepspeed(total_norm)
    test_deepspeed(total_norm)
    test_torch(total_norm)
    test_deepspeed_fix(total_norm)

def test_old_deepspeed(total_norm_tensor):
    total_norm = copy.deepcopy(total_norm_tensor)
    # https://github.com/deepspeedai/DeepSpeed/blob/v0.10.3/deepspeed/runtime/zero/stage_1_and_2.py#L1233
    if total_norm == float('inf') or total_norm == -float('inf') or total_norm != total_norm:
        total_norm = torch.tensor(float(-1))
        
    # https://github.com/deepspeedai/DeepSpeed/blob/v0.10.3/deepspeed/runtime/zero/stage_1_and_2.py#L1848
    clip_grad = float(1.0)
    loss_scale = float(1.0)
    combined_scale = loss_scale
    clip = ((total_norm / loss_scale) + 1e-6) / clip_grad
    if clip > 1:
        combined_scale = clip * loss_scale
    print(f"old_deepspeed: {1. / combined_scale}")

def test_deepspeed(total_norm_tensor):
    total_norm = copy.deepcopy(total_norm_tensor)
    # https://github.com/deepspeedai/DeepSpeed/blob/v0.16.4/deepspeed/runtime/zero/stage_1_and_2.py#L1710
    norm_is_inf = total_norm.isinf()
    norm_is_nan = total_norm.isnan()
    inf_or_nan = norm_is_nan.logical_or(norm_is_inf)

    err = torch.tensor(-1.0, dtype=torch.float)
    total_norm = inf_or_nan * err + inf_or_nan.logical_not() * total_norm

    # https://github.com/deepspeedai/DeepSpeed/blob/v0.16.4/deepspeed/runtime/zero/stage_1_and_2.py#L1970
    clip_grad = float(1.0)
    loss_scale = float(1.0)
    clip = ((total_norm / loss_scale) + 1e-6) / clip_grad
    clip = torch.clamp(clip, min=1.0)
    combined_scale = clip * loss_scale
    print(f"test_deepspeed: {1. / combined_scale}")
    
def test_torch(total_norm_tensor):
    # https://github.com/pytorch/pytorch/blob/v2.6.0/torch/nn/utils/clip_grad.py#L155
    total_norm = copy.deepcopy(total_norm_tensor)
    max_norm = float(1.0)
    clip_coef = max_norm / (total_norm + 1e-6)
    clip_coef_clamped = torch.clamp(clip_coef, max=1.0)
    print(f"torch: {clip_coef_clamped}")

def test_deepspeed_fix(total_norm_tensor):
    total_norm = copy.deepcopy(total_norm_tensor)
    if total_norm.isinf() or total_norm.isnan():
        total_norm = torch.tensor(-1.0, dtype=torch.float)

    # https://github.com/deepspeedai/DeepSpeed/blob/v0.16.4/deepspeed/runtime/zero/stage_1_and_2.py#L1970
    clip_grad = float(1.0)
    loss_scale = float(1.0)
    clip = ((total_norm / loss_scale) + 1e-6) / clip_grad
    clip = torch.clamp(clip, min=1.0)
    combined_scale = clip * loss_scale
    print(f"test_deepspeed_fix: {1. / combined_scale}")
    
if __name__ == '__main__':
    print("*****NAN*****")
    test(torch.tensor(float('nan')))
    print("*****INF*****")
    test(torch.tensor(float('inf')))
    print("*****positive*****")
    test(torch.tensor(float(2.0)))

```
Result:

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1987,8 +1987,10 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
         if self.clip_grad > 0.:
             # norm is in fact norm*scale
             clip = ((total_norm / self.loss_scale) + 1e-6) / self.clip_grad
-            clip = torch.clamp(clip, min=1.0)
-            combined_scale = clip * self.loss_scale
+
+            # handle total_norm invalid value -1
+            if clip > 1:
+                combined_scale = clip * self.loss_scale
 
         for grad in grad_groups_flat:
             if isinstance(grad, list):
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1725,12 +1725,9 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
 
             total_norm = total_norm.pow(1. / norm_type)
 
-        norm_is_inf = total_norm.isinf()
-        norm_is_nan = total_norm.isnan()
-        inf_or_nan = norm_is_nan.logical_or(norm_is_inf)
+        if total_norm.isinf() or total_norm.isnan():
+             total_norm = torch.tensor(-1.0, device=self.device, dtype=torch.float)
 
-        err = torch.tensor(-1.0, device=self.device, dtype=torch.float)
-        total_norm = inf_or_nan * err + inf_or_nan.logical_not() * total_norm
         return total_norm
 
     # creates a flat fused tensor from the tensor list starting at the first_offset
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1725,8 +1725,11 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
 
             total_norm = total_norm.pow(1. / norm_type)
 
-        if total_norm.isinf() or total_norm.isnan():
-             total_norm = torch.tensor(-1.0, device=self.device, dtype=torch.float)
+        norm_is_inf = total_norm.isinf()
+        norm_is_nan = total_norm.isnan()
+
+        if norm_is_inf or norm_is_nan:
+            total_norm = torch.tensor(-1.0, device=self.device, dtype=torch.float)
 
         return total_norm
 


tjruwase:
@Glaceon-Hyy, thanks for this PR. Is it possible to convert the repro into a unit test somewhere [here](https://github.com/deepspeedai/DeepSpeed/tree/master/tests/unit/runtime/half_precision)?

tjruwase:
@Glaceon-Hyy, also do you know if setting `overlap_comm` to False has any effect on this?

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1725,12 +1725,9 @@ def get_grad_norm_direct(self, gradients, params, norm_type=2):
 
             total_norm = total_norm.pow(1. / norm_type)
 
-        norm_is_inf = total_norm.isinf()
-        norm_is_nan = total_norm.isnan()
-        inf_or_nan = norm_is_nan.logical_or(norm_is_inf)
+        if total_norm.isinf() or total_norm.isnan():
```
loadams, line 1728: @Glaceon-Hyy - could you run the pre-commit formatter to fix the formatting error?

Glaceon-Hyy:
I noticed that in commit 61daaa1, even when total_norm produced a NaN instead of the expected -1, the clip calculation (total_norm / self.loss_scale + 1e-6)/self.clip_grad still resulted in NaN. However, the condition nan > 1 evaluates to False, which coincidentally handled the invalid value.

However, in commit 1ef9b02, the torch.clamp(clip, min=1.0) introduced a new issue: when clip is NaN, torch.clamp() returns NaN unchanged. This NaN value then propagates to combined_scale, causing subsequent gradient scaling grad.data.mul_(1. / combined_scale) to produce NaN.

My latest commit addresses this by adding an explicit check to convert NaN values in clip to 1.0 before applying the clamp operation. This prevents NaN propagation while maintaining the desired gradient scaling behavior, ensuring numerical stability in cases where total_norm might become invalid.

deepspeed/runtime/zero/stage_1_and_2.py
Glaceon-Hyy: fixed

Glaceon-Hyy:
@loadams  I just force-pushed to fix DCO (Developer Certificate of Origin) issues in the commits. I noticed that my development environment using Magit did not have signed-off-by configured by default.

tjruwase:
@nelyahu, FYI for any perf impact.

nelyahu:
@tjruwase  @Glaceon-Hyy  @loadams  @hwchen2017  can you please review the fix in https://github.com/deepspeedai/DeepSpeed/pull/7184 ?
i replaced the logical equation with torch.where.

PR #7191, opened 2025-03-31
Remove definition that causes error in cuda 12.6

loadams:


tjruwase (approved):

csrc/transformer/inference/csrc/gelu.cu
```diff
@@ -11,11 +11,6 @@ namespace cg = cooperative_groups;
 #define MAX_CAP 4
 #define MAX_SEQ 2048
 
-// only used to avoid compilation error due to lack of definition.
-#ifndef BF16_AVAILABLE
```
loadams, line 15: This seems to be needed for 11.6 as this fails in the pre-compile-ops test.  For now we can either update that workflow to use a newer cuda version or guard this in a better way.

PR #7208, opened 2025-04-08
Support complicated use cases with TiedLayerSpec

limjcst:
I want to reuse a composed module in the pipeline. For example, the following `MyModule` has a member `linear`, which is also a module.

```python
class MyModule(torch.nn.Module):
    def __init__(self, n_in: int, n_out: int):
        super().__init__()
        self.linear = torch.nn.Linear(n_in, n_out)
        self.layer_norm = torch.nn.LayerNorm(n_out)

    def forward(self, data: torch.Tensor) -> torch.Tensor:
        hidden = self.linear(data)
        hidden = self.layer_norm(hidden)
        return hidden
```

`MyModule.linear.weight` should be synchronized among related ranks. As a result, I add `linear.weight` to `TiedLayerSpec.tied_weight_attr`.
BTW, I generate the whole `tied_weight_attr` by the following instruction.

```python
tied_weight_attr = [name for name, p in layer.named_parameters() if p.numel() > 1]
```

However, the builtin `getattr` used by `PipelineModule` fails to find a nested attribute like `linear.weight`.
Hence, this PR first extends the builtin `getattr` to a recursive version `PipelineModule._recursive_getattr`, accessing each attribute segment one by one.

Meanwhile, the order of tied weights matters in synchronization. This PR suggests to sort tie_keys in `PipelineModule._index_tied_modules` to avoid hanging.

### DIFF
--- a/deepspeed/runtime/pipe/module.py
+++ b/deepspeed/runtime/pipe/module.py
@@ -443,18 +443,26 @@ class PipelineModule(nn.Module):
 
         self._set_bounds(start=self.parts[stage_id], stop=self.parts[stage_id + 1])
 
+    @staticmethod
+    def _recursive_getattr(module: torch.nn.Module, attr_name: str) -> torch.Tensor:
+        '''Allow getting an attribute like "linear.weight"'''
+        weight = module
+        for item in attr_name.split("."):
+            weight = getattr(weight, item)
+        return weight
+
     def allreduce_tied_weight_gradients(self):
         '''All reduce the gradients of the tied weights between tied stages'''
         for key, comm in self.tied_comms.items():
             for attr_name in comm['weight_attr']:
-                weight = getattr(self.tied_modules[key], attr_name)
+                weight = self._recursive_getattr(self.tied_modules[key], attr_name)
                 dist.all_reduce(weight.grad, group=comm['group'])
 
     def get_tied_weights_and_groups(self):
         weight_group_list = []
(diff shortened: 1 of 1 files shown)


tohtana:
Thank you @limjcst for the contribution! This is a significant improvement.

tohtana (approved):

loadams:
> > [nv-accelerate-v100 failed](https://github.com/deepspeedai/DeepSpeed/actions/runs/14346042861/job/40215911601), raising "invalid command 'bdist_wheel'". However, this job succeeded in [another run](https://github.com/deepspeedai/DeepSpeed/actions/runs/14340182591/job/40197254667).
> > Note that the failed job used "cached wheel-0.46.1-py3-none-any.whl.metadata". @agronholm suggested `setuptools >= 70.1` for `wheel==0.46.0`. Is this suggestion meaningful for DeepSpeed?
> > [pypa/wheel#660 (comment)](https://github.com/pypa/wheel/issues/660#issuecomment-2775668964)
> 
> Thanks @limjcst - I saw this failure on another PR and will take a look and merge the fixes into your PR when ready.

@limjcst - it looks like the `wheel` team has yanked the problemabic wheel version so tests should be passing again.

agronholm:
Nevertheless, upgrading to `setuptools >= 70.1` should prevent any future issues. I noticed that this project lacks a standard `pyproject.toml` which would be able to specify build requirements.

Issue #7229, opened 2025-04-17
[BUG] DeepCompile in ZeRO-1 fails to do the forward pass


HollowMan6:
**Describe the bug**
It looks like DeepCompile might have some compatibility issues with flash attention, when doing the forward pass.

Error logs:

```logs
MemoryProfiling error /pytorch/build/aten/src/ATen/RegisterCUDA.cpp:18503: SymIntArrayRef expected to contain only concrete integers

While executing %view : [num_users=2] = call_function[target=torch.ops.aten.view.default](args = (%arg1_1, [%arg0_1]), kwargs = {})
Original traceback:
  File "flash_attn/bert_padding.py", line 114, in torch_dynamo_resume_in_unpad_input_at_114
    indices = torch.nonzero(all_masks.flatten(), as_tuple=False).flatten()

```

**To Reproduce**
I'm using [OpenRLHF](https://github.com/OpenRLHF/OpenRLHF) with DeepCompile, but I think this should not be directly related to code in OpenRLHF and should be general enough to reproduce.

Basically, you may be able to reconstruct a reproduction script that, during training, run a forward pass when doing `.eval()` in ZeRO-1 with DeepCompile, for a huggingface model (`meta-llama/Llama-3.2-1B-Instruct`) with flash attention enabled (`transformers==4.51.1`, `flash_attn==2.7.4.post1`).

**Expected behavior**
No error thrown

**ds_report output**

```logs
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
 [WARNING]  async_io: please install the libaio-devel package with yum
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
async_io ............... [NO] ....... [NO]
fused_adam ............. [NO] ....... [OKAY]
cpu_adam ............... [NO] ....... [OKAY]
cpu_adagrad ............ [NO] ....... [OKAY]
cpu_lion ............... [NO] ....... [OKAY]
[...]

PR #7243, opened 2025-04-24
Fix issue with symint input

tohtana:
This PR fixes an issue with symint input in backend. (See #7229)


loadams (approved):

hijkzzz:
The MR does not solve the bug "RuntimeError: 'weight' must be 2-D
" for flash-attn 2.7.4.post1

PR #7244, opened 2025-04-24
Fixed the bug of memory consumption when using ZeRO-Offload

LaosGAmin:
Fixed the bug of memory consumption when using ZeRO-Offload with FP16/BF16. Before the fix, the memory usage about 3x params_FP16. Now it is 1x params_FP16.
(Something is also not sure, such as how custom CUDA stream used in deepspeed, is that can used on multi-machines? )


loadams:
@LaosGAmin - could you fix the DCO or let us know and we can set it to passing?

LaosGAmin:
@loadams Sorry about that, I am in online git environment. And the Signed-off-by: 51F1y <1790690514@qq.com>
Besides, there is still a formatting task failed.

Issue #7190, opened 2025-03-31
Compiling DeepSpeed with CUDA 12.6


r-sarma:
I am trying to compile DeepSpeed on the HPC system JUWELS Booster. The CUDA version on the system is 12.6. Detailed specs of the system available here:
https://apps.fz-juelich.de/jsc/hps/juwels/booster-overview.html

I first installed Torch 2.6 with cu12.6 wheels. Then I am trying to install DeepSpeed 0.15.0. I am exporting these variables:
```
export DS_BUILD_CCL_COMM=1
export DS_BUILD_UTILS=1
export DS_BUILD_AIO=1
export DS_BUILD_FUSED_ADAM=1
export DS_BUILD_FUSED_LAMB=1
export DS_BUILD_TRANSFORMER=1
export DS_BUILD_STOCHASTIC_TRANSFORMER=1
export DS_BUILD_TRANSFORMER_INFERENCE=1
```

But I get an error during compilation with the message:
```
csrc/transformer/inference/csrc/gelu.cu(16): error: class "__nv_bfloat162" has already been defined (previous definition at line 4787 of /p/software/jurecadc/stages/2025/software/CUDA/12/include/cuda_bf16.h)
        using __nv_bfloat162 = __half2;
              ^
      
      1 error detected in the compilation of "csrc/transformer/inference/csrc/gelu.cu".
      error: command '/p/software/jurecadc/stages/2025/software/CUDA/12/bin/nvcc' failed with exit code 255
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for deepspeed
  Running setup.py clean for deepspeed
Failed to build deepspeed
ERROR: Could not build wheels for deepspeed, which is required to install pyproject.toml-based projects
```

So basically the CUDA 12.6 stack already has this bf16 definition, which is leading to the issue. When I check the corresponding file, is looks like:
```
#ifndef BF16_AVAILABLE
using __nv_bfloat162 = __half2;
#endif
```
If I comment out the corresponding line, it works fine. But is there a better way to handle this?

loadams:
Hi @r-sarma - thanks for reporting this.  We've seen this on Windows builds before, but not on Linux builds.  I'll try with the specific 12.6 repro case, but in the meantime, could you please try installing a newer DeepSpeed version as well?  I am not sure if we added support for 12.6 when 0.15.0 was released.

r-sarma:
[...]

PR #7248, opened 2025-04-24
Fix compile error for nv_bloat162

loscrossos:
some systems seem not to have the __nv_bfloat162 definition so a placeholder was introduced. newer CUDA libs have that definition, which breaks the compile process. this patch adds the official cuda_bf16.h guard while keeping the old code and a safety assert in case the definition should change in the future. see #7190 for reference


loscrossos:
hmm i just see that there is a similar that tried to remove the original definition: https://github.com/deepspeedai/DeepSpeed/pull/7191

@loadams does this cancel the other out?

loadams:
> hmm i just see that there is a similar that tried to remove the original definition: #7191
> 
> @loadams does this cancel the other out?

Thanks @loscrossos - this PR is fine, I hadn't had time to work on the other yet, though I will still want to update the nv-pre-compile test.

csrc/transformer/inference/csrc/gelu.cu
```diff
@@ -12,9 +12,14 @@ namespace cg = cooperative_groups;
 #define MAX_SEQ 2048
 
 // only used to avoid compilation error due to lack of definition.
-#ifndef BF16_AVAILABLE
+#ifndef BF16_AVAILABLE
```
loadams, line 15: @loscrossos - could you please run the pre-commit formatter on your branch?  That will resolve the remaining CI error.

Also could you fix the DCO, or let us know and we can resolve that on our side as well.  Thanks.

loscrossos:
oh dear.. i am a bit at wits end today on my machine.. i rant the formatter and signed off but am a bit confused on merges.. could you please tell the if this is the right commit? i thought i did the DCO but the bot seems to be complaining about it still

loadams:
> oh dear.. i am a bit at wits end today on my machine.. i rant the formatter and signed off but am a bit confused on merges.. could you please tell the if this is the right commit? i thought i did the DCO but the bot seems to be complaining about it still. I also see that i signed with the old nick. so please could you resolve it? else can close and redo the PR

@loscrossos - formatting looks good, thanks.  For DCO, the commands should be [here](https://github.com/deepspeedai/DeepSpeed/pull/7248/checks?check_run_id=41132994917) but we can also manually take care of that if you want rather than trying another merge?  Just let me know if you want me to do that part.

loscrossos:
@loadams thanks! i signed off the branch now... actually this story threw me into a dark rabbit hole where i now re-installed my developing setup in windows AND linux.. :D it was due for a while anyway.. phew what a day

loadams (approved):

PR #7253, opened 2025-04-27
Fix the memory usage of ZeRO-Offload (update stage_1_and_2.py)

arminzhu:
Fix the memory usage of ZeRO-Offload with stage 1 and 2. Before the fix, the memory usage about 3x params_FP16. Now it is 1x params_FP16.
Signed-off-by: arminzhu <mingzhengzhu1998@gmail.com>

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1888,7 +1888,7 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
                 bit16_partitions = self.parallel_partitioned_bit16_groups[i]
                 fp32_partition = self.single_partition_of_fp32_groups[i]
                 bit16_partitions[partition_id].data.copy_(
-                    fp32_partition.to(dtype=bit16_partitions[partition_id].data.dtype).data, non_blocking=True)
+                    fp32_partition.to(dtype=bit16_partitions[partition_id].data.dtype, non_blocking=True).data, non_blocking=True)
 
                 self.timers(OPTIMIZER_STEP_TIMER).stop()
             else:
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1888,7 +1888,7 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
                 bit16_partitions = self.parallel_partitioned_bit16_groups[i]
                 fp32_partition = self.single_partition_of_fp32_groups[i]
                 bit16_partitions[partition_id].data.copy_(
-                    fp32_partition.to(get_accelerator().current_device_name()).data)
+                    fp32_partition.to(dtype=bit16_partitions[partition_id].data.dtype).data, non_blocking=True)
 
                 self.timers(OPTIMIZER_STEP_TIMER).stop()
             else:


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1888,7 +1888,7 @@ def step(self, closure=None):
                 bit16_partitions = self.parallel_partitioned_bit16_groups[i]
                 fp32_partition = self.single_partition_of_fp32_groups[i]
                 bit16_partitions[partition_id].data.copy_(
-                    fp32_partition.to(get_accelerator().current_device_name()).data)
+                    fp32_partition.to(dtype=bit16_partitions[partition_id].data.dtype).data, non_blocking=True)
```
sfc-gh-truwase, line 1891: Unfortunately, this change will hurt performance by introducing pageable h2d copies. Please the previous PR that fixed this perf problem: #5301

sfc-gh-truwase: A possible solution is to use an intermediate pinned `bit16` buffer to ensure page-locked h2d. This intermediate buffer can be created in constructor to be as large as the largest partition and thus reused for all partitions.

arminzhu: > Unfortunately, this change will hurt performance by introducing pageable h2d copies. Please the previous PR that fixed this perf problem: #5301

You were right. And maybe we have another solution. In _tensor.to()_ 'non_blocking=True' will get a pined memory tensor, so it can transfer tensors in asynchronous manner by change it into 'fp32_partition.to(dtype=bit16_partitions[partition_id].data.dtype, non_blocking=True).data, non_blocking=True)' . If we do this, there will also be a little bit hurt performance because of the data type transfer.

arminzhu: > A possible solution is to use an intermediate pinned `bit16` buffer to ensure page-locked h2d. This intermediate buffer can be created in constructor to be as large as the largest partition and thus reused for all partitions.

The best solution I can think of is that the data type convert should be in the fused AdamW. If the inputs are `g_16, m_32, v_32, p_32`, the outputs will be `m_32, v_32, p_32, p_16`. We can reuse the `g_16` as the `.copy_()` destination and make it become `p_16`. If we do so, we can minimize the data read and write between CPU and CPU memory.

sfc-gh-truwase: @arminzhu, one issue with your suggestion is that it requires modification or assumptions about AdamW or other optimizers. Remember that is situation only arises for offloading not for general ZeRO usage. For these reasons, it is better to handle this inside ZeRO. 

Did you see a problem with my proposal of creating an intermediate pinned `bit16` buffer?

arminzhu: > @arminzhu, one issue with your suggestion is that it requires modification or assumptions about AdamW or other optimizers. Remember that is situation only arises for offloading not for general ZeRO usage. For these reasons, it is better to handle this inside ZeRO. 
> 
> Did you see a problem with my proposal of creating an intermediate pinned `bit16` ? 

Yes, create an intermediate pinned `bit16` buffer is good. I just try to get a better performance version, haha.

If we don't consider the .pin_memory() overhead, we also can use it.

sfc-gh-truwase: > No, creating an intermediate pinned `bit16` buffer is good. I just try to get a better performance version, haha.

I approve of the relentless pursuit for more performance. That is true DeepSpeed spirit :)


> If we don't consider the .pin_memory() overhead, we also can use it.

Since the page-locked buffer is created during ZeRO initialization it should negligible in end-to-end training time.

arminzhu: Yeah. Add memory buffers is good enough. And it will increase the CPU memory usage, too.
In fact, we recently analyzed the bottleneck of ZeRO-Offload. And we find it needs very high CPU memory bandwidth to keep performance. So, I think if we can reduce the data movement between CPU and DRAM, we can get better performance ; )

I have added the memory buffers, and tested it on my local machine. It seems OK. I think I will PR the code tomorrow. Because it is 23:10 in my time zone now :)

PR #7267, opened 2025-04-30
add `Makefile` to ease maintenance

stas00:
adding `Makefile` with `make format` and `make test` to make things easier to maintain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ help: ## this help
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[0-9a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-22s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 	echo $(MAKEFILE_LIST)
 
-test: ## run cpu and gpu tests
+test: ## run tests
 	pytest --forked tests/unit/
 
 format: ## fix formatting
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,23 @@
+# usage: make help
+
+.PHONY: help test format
+.DEFAULT_GOAL := help
+
+help: ## this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[0-9a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-22s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,23 @@
...
+.DEFAULT_GOAL := help
+
+help: ## this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[0-9a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-22s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+	echo $(MAKEFILE_LIST)
+
+test: ## run tests
+	pytest --forked tests/unit/
+
+format: ## fix formatting
```
loadams, line 13: This is great, @stas00, thanks!  Would it be worth adding a link to this in the docs anywhere?

loadams (approved):

Makefile
stas00: that's a good suggestion, Logan. Added.

stas00:
@loadams, what do we do with those CI jobs that got cancelled? and why did they get cancelled?

loadams:
> @loadams, what do we do with those CI jobs that got cancelled? and why did they get cancelled?

@stas00 - the v100 runners we used went down for maintenance, and we are working on getting them back up/I was preoccupied with other work late last week.  The jobs were cancelled because they could not find a runner within 24 hours whereas they will just queue otherwise if there are runners available that are just busy.  They should just need a re-run once we get the CI nodes back.

stas00:
so what's the CI rerun protocol - it doesn't look like it can restart jobs that were cancelled.

Do I have to push an empty commit?

stas00:
Actually in this particular case - adding a `Makefile` can't possibly break the CI, can it? The formatting job is a success - I think that's all that matters. 

In general I think the massive requirements of DS CI can be diminished by not running real tests on doc PRs - only formatting jobs.

loadams:
> so what's the CI rerun protocol - it doesn't look like it can restart jobs that were cancelled.
> 
> Do I have to push an empty commit?

I'm able to click on them and see a "Re-Run" option, so perhaps that's a permissions thing.  I re-ran some when we got runners back temporarily late last week but then we had to shut them down for maintenance as well.  I'll re-run all when they are back I suppose, that or merging master once we get commits in.

loadams:
> Actually in this particular case - adding a `Makefile` can't possibly break the CI, can it? The formatting job is a success - I think that's all that matters.
> 
> In general I think the massive requirements of DS CI could be reduced by not running real tests on doc PRs - only formatting jobs.

Agreed, these tests don't need to run here.  We just haven't perfected the path include/exclude filters, right now we exclude docs, but not the docs at the root of the repo.  We can force complete this one now though if you agree?

stas00:
we have the same perms. specifically I went into the CI and told it to re-run all tests, but the cancelled tests didn't get requeued.

yes, let's force complete on this one indeed as we both agree on it.

PR #7268, opened 2025-04-30
Ulysses SP for HF Integration

stas00:
This is the Deepspeed counterpart of https://github.com/snowflakedb/ArcticTraining/pull/45 - as the new feature(s) require changes on both sides.

For PR reviewers: 

Readiness status:
- [x] Code
- [x] Tests
- [ ] Docs - working on it

Features:

- [x] add support for delaying grad addition via `param.ds_grad_is_ready` flag (used when performing tiled compute in an autograd function)
- [x] add light sp-only mpu version (Jeff Rasley)
- [x] improved debug
- [x] added `all_gather_object` to `dist`
- [x] `UlyssesSPAttentionHF` (port of UlyssesAttention from Megatron-Deepspeed plus modern MHA-variations)
- [x] `UlyssesSPDataLoaderAdapter` - DL adapter to shard the normal DL batches to be used by `UlyssesSPAttentionHF`
- [x] `SequenceTiledCompute` - generic autograd function to perform compute after tiling on the sequence dimension
- [x] `TiledMLP` - a specific autograd function to perform tiled MLP (it's much easier to understand before trying to grok `SequenceTiledCompute`)
- [x] added a differentiable `_DimZeroAllToAll` (Samyam Rajbhandari)
- [x] torch-dist-check now allows `torch.distributed.nn` (which is needed since deepspeed's dist is not up to date with `torch.distributed.nn`)


deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -0,0 +1,1225 @@
+# Copyright (c) The DeepSpeed Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+# XXX: the copyright above is wrong - needs to be either Snoflake or something else but not MSFT, but CI is requiring MSFT for now
```
sfc-gh-truwase, line 6: I thought this is now resolved, no?

deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -0,0 +1,1225 @@
...
+    @staticmethod
+    def backward(ctx, *grads) -> torch.Tensor:
+        fn = ctx.fn
+        (x, ) = ctx.saved_tensors
+        self = ctx.self
+        shards = ctx.shards
+        compute_params = ctx.compute_params
+        x_requires_grad = x.requires_grad
+        x = x.detach()
+        x.requires_grad_(x_requires_grad)
```
sfc-gh-truwase, line 818: The following sequence looks like a no-op. What does it do?

```suggestion
        x_requires_grad = x.requires_grad
        x = x.detach()
        x.requires_grad_(x_requires_grad)
```

deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -0,0 +1,1225 @@
...
+            if compute_params is not None:
+                if i + 1 < shards:
+                    for param in compute_params:
+                        param.ds_grad_is_ready = False
+                else:
+                    # last shard, can add the grad
+                    for param in compute_params:
+                        param.ds_grad_is_ready = True
+
+            x_shard.requires_grad_(x_requires_grad)
```
sfc-gh-truwase, line 837: Since `x_shard.grad` is referenced below, is it possible for `x_shard.requires_grad` to be `False`?

deepspeed/runtime/sequence_parallel/ulysses_sp.py
stas00: Please see:

https://github.com/pytorch/pytorch/blob/134179474539648ba7dee1317959529fbd0e7f89/torch/utils/checkpoint.py#L73-L74

yes/no/better way?

deepspeed/runtime/sequence_parallel/ulysses_sp.py
stas00: that's the thing, it wasn't set until I explicitly set it for the shard.

I tested doing `chunk` on a requires_grad tensor alone and the shards inherited this property, but for some reason in this code it wasn't the case.

deepspeed/runtime/sequence_parallel/ulysses_sp.py
stas00: Thank you, will fix!

deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -0,0 +1,1225 @@
...
+        fn,
+        self,
+        x,
+        shards,
+        compute_params,
+    ) -> torch.Tensor:
+        ctx.fn = fn
+        ctx.self = self
+        ctx.shards = shards
+        ctx.compute_params = compute_params
```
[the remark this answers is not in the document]
stas00: good call, Tunji! Thank you!

deepspeed/runtime/sequence_parallel/ulysses_sp.py
stas00: here is why:

```
In [1]: import torch

In [2]: x = torch.zeros((1))

In [3]: x.requires_grad
Out[3]: False

In [4]: x = torch.zeros((1), requires_grad=True)

In [5]: x.requires_grad
Out[5]: True

In [6]: y = x.detach()

In [7]: y.requires_grad
Out[7]: False

In [8]: x = x.detach()

In [9]: x.requires_grad
Out[9]: False
```

as you can see `detach` drops `requires_grad` flag.

stas00: so my code gets the original value and then restores it, probably could write it as:

```
x_requires_grad = x.requires_grad
x = x.detach().requires_grad_(x_requires_grad)
```
so it's more intuitive?

deepspeed/runtime/sequence_parallel/ulysses_sp.py
stas00: I tested again and discovered a very strange thing. `x_shard.requires_grad` was already `True`, but... if I don't do:
```
x_shard.requires_grad_(x_requires_grad)
```
my correctness test against the baseline fails!

Could it be that the `view` that is an output of `torch.chunk` is somehow buggy and it shows `.requires_grad` to be True, but it's not?

sfc-gh-truwase: Hmm, interesting. Could be bug. Would it be better to explicitly set `requires_grad=True` since that is desired behavior?

sfc-gh-truwase: I am fine either way.

stas00: I was just following the precendent set by pytorch of preserving whatever the incoming grad requiring is.

for TiledMLP it's probably safe to hardcode it, but for the generic tiledcompute class we don't know.

stas00:
@tjruwase, this is madness, I have been retrying `nv-torch-latest-v100` 6 times already - it fails on a new flakey test each time. None related to this PR.

stas00:
phoooph! it's finally in!

we can start rolling out a new version?

sfc-gh-truwase:
@loadams, can you please help with a release on Monday?

PR #7283, opened 2025-05-15
fix asymmetric in dequantize

pencil-hub:
框架中反量化算子（dequantize）在非对称量化产生结果时只将float类型转成half类型，漏掉了对float类型的转换，导致在输出是float类型时，会产生精度误差。
企业微信截图_1747294273387


patch-file.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/csrc/includes/quantization_utils.h b/csrc/includes/quantization_utils.h
+index 61630d0a..94958fb4 100644
+--- a/csrc/includes/quantization_utils.h
++++ b/csrc/includes/quantization_utils.h
```
loadams, line 4: @pencil-hub - this came through as a diff rather than as a change to the file.  Would it be possible to re-submit your PR?

pencil-hub: ok, I have resubmitted.

loadams: Thanks

csrc/includes/quantization_utils.h
```diff
@@ -123,7 +123,7 @@ class Params<Type::Asymmetric, numBits> {
     DS_D_INLINE T dequantize(int8_t val)
     {
         const float val_deq_f = ((conversion::to<float>(val)) * scale) + offset;
-        return conversion::to<__half>(val_deq_f);
+        return conversion::to<T>(val_deq_f);
```
sfc-gh-truwase, line 126: I am not familiar with this logic, but I can concerned whether it is correct to dequantize into arbitrary type since the original data that was quantized is of `__half` type. I understand that this was existing code. So I will appreciate if you can educate me a bit about your use case that motivated this PR. Thanks!

sfc-gh-truwase (approved):

loadams (approved):

PR #7288, opened 2025-05-16
Temporarily skip AIO tests due to an issue with runners

loadams:


sfc-gh-truwase (approved):

.github/workflows/nv-torch-latest-v100.yml
```diff
@@ -55,5 +55,5 @@ jobs:
         run: |
           unset TORCH_CUDA_ARCH_LIST # only jit compile for current arch
           cd tests
-          pytest $PYTEST_OPTS --forked -n 8 unit/ --torch_ver="2.6" --cuda_ver="12.4"
+          pytest $PYTEST_OPTS --forked -n 8 unit/ --torch_ver="2.6" --cuda_ver="12.4" --ignore=unit/ops/aio/
```
sfc-gh-truwase, line 58: Since all the nvme unit tests are affected, I think a more robust option is to uninstall libaio-dev to make the ops incompatible.

loadams: Agreed, removing this and going with your changes.

PR #7289, opened 2025-05-16
Modal CI

sfc-gh-truwase:
Test

### DIFF
--- a/.github/workflows/modal_torch_latest.yml
+++ b/.github/workflows/modal_torch_latest.yml
@@ -1,0 +1,98 @@
+name: deepspeed-torch-latest-modal
+
+# This CI is running on modal.com's GPUs.
+#
+# It's set up here on github actions and then the cloned repo is sent to modal and everything
+# happens on their hw - see arctic_training/ci.py for where the actual vm is loaded, updated and the tests are
+# run.
+#
+# Both files are annotated to what's important and how one might change or update things if needed.
+#
+# Note that since this is a Required job we can't use `on.push.path` file filter - we are using
+# collect-tests job to do the filtering for us so that the job can be skipped and satisfy the
+# Required status for PRs to pass.
+#
+
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+
+  pull_request:
+    types: [draft, opened, ready_for_review, synchronize]
+    branches:
+      - master
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  collect-tests:
+    name: Collect tests to run
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      deepspeed: ${{ steps.filter.outputs.deepspeed }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          lfs: true
+
+      - name: Filter changed files
+        uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          filters: |
+            deepspeed:
+              - 'deepspeed/**/*.py'
+              - '.github/workflows/deepspeed.yml'
+
+  deploy:
+    name: DeepSpeedAI CI
+    runs-on: ubuntu-latest
+    needs: collect-tests
+    env:
+      # these are created at https://modal.com/settings/deepspeedai/tokens
+      # they are then added to the repo's secrets at https://github.com/deepspeedai/deepspeed/settings/secrets/actions
+      MODAL_TOKEN_ID: ${{ secrets.MODAL_TOKEN_ID }}
+      MODAL_TOKEN_SECRET: ${{ secrets.MODAL_TOKEN_SECRET }}
+      # this one comes from https://huggingface.co/settings/profile of the bot user
+      # and it too is then updated at https://github.com/deepspeedai/deepspeed/settings/secrets/actions
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+
+    if: needs.collect-tests.outputs.deepspeed == 'true'
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          lfs: true
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+          cache: 'pip' # caching pip dependencies
+
+      - name: Install build dependencies
+        run: |
+          pip install uv==0.4.0 # much faster than pip
+          uv pip install --system modal
+          pip install .[dev,1bit,autotuning,deepcompile]
+          ds_report
+          # time uv pip compile arctic_training/setup.py --extra all -o arctic_training/ci-requirements.txt
+          # # add vllm manually to deps since it fails pip compile w/o CUDA_HOME being set in github actions
+          # # if changing the version here also change it in setup.py to match
+          # echo 'vllm==0.6.2' >> arctic_training/ci-requirements.txt
+
+      - name: Run tests
+        run: |
+          modal run deepspeed.torch_latest_ci
+
--- a/deepspeed/torch_latest_ci.py
+++ b/deepspeed/torch_latest_ci.py
@@ -1,0 +1,26 @@
+from pathlib import Path
+
+import modal
+
+ROOT_PATH = Path(__file__).parent.parent
+
+image = (modal.Image.from_registry(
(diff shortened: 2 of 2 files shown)


.github/workflows/modal-accelerate.yml
```diff
@@ -0,0 +1,103 @@
+name: modal-accelerate
+
+# This CI is running on modal.com's GPUs.
+#
+# It's set up here on github actions and then the cloned repo is sent to modal and everything
+# happens on their hw - see arctic_training/ci.py for where the actual vm is loaded, updated and the tests are
```
sfc-gh-sbekman:
```suggestion
# happens on their hw - see deepspeed/modal_ci/accelerate.py for where the actual vm is loaded, updated and the tests are
```

.github/workflows/modal-torch-latest.yml
```diff
@@ -0,0 +1,103 @@
+name: modal-torch-latest
+
+# This CI is running on modal.com's GPUs.
+#
+# It's set up here on github actions and then the cloned repo is sent to modal and everything
+# happens on their hw - see arctic_training/ci.py for where the actual vm is loaded, updated and the tests are
```
sfc-gh-sbekman:
```suggestion
# happens on their hw - see deepspeed/modal_ci/torch_latest.py  for where the actual vm is loaded, updated and the tests are
```

deepspeed/modal_ci/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) DeepSpeed Team.
```
sfc-gh-sbekman: why is `modal_ci` inside `deepspeed`? it doesn't belong to the deepspeed package, no? do you want it to be installed on package install?

deepspeed/modal_ci/accelerate.py
```diff
@@ -0,0 +1,49 @@
...
+
+import modal
+
+ROOT_PATH = Path(__file__).parent.parent.parent
+
+# yapf: disable
+image = (modal.Image
+         .from_registry("pytorch/pytorch:2.7.0-cuda12.6-cudnn9-devel", add_python="3.10")
+         .run_commands("apt update && apt install -y libaio-dev")
+        .apt_install("git")
```
sfc-gh-sbekman:
```suggestion
         .apt_install("git")
```

deepspeed/modal_ci/accelerate.py
```diff
@@ -0,0 +1,49 @@
...
+         .run_commands("uv pip install --system --compile-bytecode datasets==3.6.0")
+         .run_commands(
+                "git clone https://github.com/huggingface/accelerate && \
+                uv pip install --system --compile-bytecode accelerate[testing]"
+            )
+         .run_commands("uv pip install --system --compile-bytecode protobuf")
+         .run_commands("pip list")
```
sfc-gh-sbekman:
```suggestion
         .run_commands("uv pip list")
```

deepspeed/modal_ci/torch_latest.py
```diff
@@ -0,0 +1,44 @@
...
+
+import modal
+
+ROOT_PATH = Path(__file__).parent.parent.parent
+
+# yapf: disable
+image = (modal.Image
+         .from_registry("pytorch/pytorch:2.6.0-cuda12.4-cudnn9-devel", add_python="3.10")
+         .run_commands("apt update && apt install -y libaio-dev")
+         .run_commands("pip list")
```
sfc-gh-sbekman:
```suggestion
         .run_commands("uv pip list")
```

deepspeed/modal_ci/torch_latest.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) Snowflake.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+from pathlib import Path
+
+import modal
+
+ROOT_PATH = Path(__file__).parent.parent.parent
```
sfc-gh-sbekman:
```suggestion
ROOT_PATH = Path(__file__).parents[2]

File: deepspeed/modal_ci/accelerate.py [hunk 8]
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) Snowflake.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+from pathlib import Path
+
+import modal
+
+ROOT_PATH = Path(__file__).parent.parent.parent
```
sfc-gh-sbekman:
```suggestion
ROOT_PATH = Path(__file__).parents[2]
```

deepspeed/modal_ci/accelerate.py
```diff
@@ -0,0 +1,49 @@
...
+@app.function(
+    gpu="l40s:1",
+    # gpu="a10g:2",
+    # secrets=[modal.Secret.from_local_environ(["HF_TOKEN"])],
+    timeout=1800,
+)
+def pytest():
+    import subprocess
+    subprocess.run(
+        "pytest --verbose /accelerate/tests/deepspeed".split(),
```
sfc-gh-sbekman: probably activate `pytest-xdist` (`-n 4`) to massively speed up the completion and lowering costs! - the tests usually need very little memory and can reuse the same gpus at the same time. need to install `pytest-xdist` if it's not there already.

I normally try a few -n values and use the lowest value that gives the fastest completion time. i.e. no point doing `-n 10` if `-n 4` gives about the same completion time - a very high number will create a lot of cpu/io stress and won't actually benefit the completion time.

same for the next workflow.

deepspeed/modal_ci/__init__.py
sfc-gh-truwase: I am open to suggestions here. My thinking is that CI assets reside in the package. In this case, it looked like we are transferring logic from `.github` to `deepspeed.modal_ci`.

sfc-gh-sbekman: `/modal_ci` on the top level would work just fine.

Alternatively, you could also do: `/deepspeed_ci` or just `/ci` and have all the ci-related modules there - same placement logic as `/op_builder`.

deepspeed/modal_ci/accelerate.py
```diff
@@ -43,7 +43,7 @@
 def pytest():
     import subprocess
     subprocess.run(
-        "pytest --verbose /accelerate/tests/deepspeed".split(),
+        "pytest -n 4 --verbose /accelerate/tests/deepspeed".split(),
```
sfc-gh-sbekman: so when everything is ready to merge, let's try a few of these and see which one gives us the fastest completion - we should log the total tests runtime for each so that we have a reference point.

As the amount of tests changes we might want to periodically re-measure.

Also if we change the gpu type this could change as well, as higher gpu memory will allow a higher concurrency.

deepspeed/modal_ci/__init__.py
sfc-gh-truwase: Thanks. I will go with `/ci`.

deepspeed/modal_ci/accelerate.py
sfc-gh-sbekman: also `xdist` has additional flags to further optimize runtime - e.g. if there is a shared test setup in the module it'll group the tests belonging to the same module together, so that the `setup` is run only once. If there are independent then it makes no difference to the bottom line.

.github/workflows/modal-accelerate.yml
```diff
@@ -0,0 +1,103 @@
...
+          lfs: true
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+          cache: 'pip' # caching pip dependencies
+
+      - name: Install build dependencies
+      # TODO: Need to determine whether these installs are redundant by subesquent modal.Image call.
```
sfc-gh-truwase: @stas00 any thoughts on this concern.

sfc-gh-sbekman: as discussed on slack, the only thing you need to install on GH actions is what `modal` requires to launch itself - it doesn't care for deepspeed install at this point.

so no deepspeed installing.

ci/accelerate.py
```diff
@@ -0,0 +1,43 @@
...
+
+from pathlib import Path
+
+import modal
+
+ROOT_PATH = Path(__file__).parents[1]
+
+# yapf: disable
+image = (modal.Image
+         .from_registry("pytorch/pytorch:2.6.0-cuda12.4-cudnn9-devel", add_python="3.10")
```
loadams, line 14: I'd love to make this configurable in the future but for now this makes sense as the best way to clarify the versions we are using.

loadams (approved):

PR #7299, opened 2025-05-21
[XPU] Support XCCL on deepspeed side

ys950902:
XCCL will be used for XPU device on Pytorch-2.8, with this support will remove torch-ccl on XPU device, and we will also reserve the old path for torch-CCL enable.


ys950902:
Hi @loadams, sorry for late response, cause xccl official support is based on pytorch-2.8, so fix is based on this. https://github.com/deepspeedai/DeepSpeed/pull/7113 this pr is too old based on new deepseed, I opened a new pr for xccl support.

accelerator/real_accelerator.py
```diff
@@ -136,6 +136,20 @@ def get_accelerator():
...
                 pass
+        if accelerator_name is None:
+            try:
+                import torch
+
+                # torch.xpu supported in upstream pytorch.
+                # Currently we can run on xpu device only using pytorch,
+                # also reserve the old path using ipex when the torch version is old.
+                if hasattr(torch, 'xpu'):
+                    if torch.xpu.device_count() > 0 and torch.xpu.is_available():
```
delock, line 147: I'm thinking about the following scenario: A desktop/laptop with an integrated XPU device and a dicrete CUDA device.  In this case following the logic xpu will be selected.  Intuitively, CUDA device should be default.

To be more friendly to desktop users, could add one more check `torch.cuda.device_count == 0`?  This would be more safe for desktop CUDA device users.

ys950902: Got it, I have fixed it.

accelerator/real_accelerator.py
```diff
@@ -136,6 +136,21 @@ def get_accelerator():
                     accelerator_name = "xpu"
             except ImportError as e:
                 pass
+        if accelerator_name is None:
+            try:
+                import torch
+
+                # torch.xpu supported in upstream pytorch.
```
loadams, line 143: Could you add what version of pytorch xpu is supported in?

ys950902: Done, has added, is for pytorch-2.8.

delock (approved):

loadams (approved):

PR #7301, opened 2025-05-21
disable license check until the new license situation has been sorted…

stas00:
Until we sort out the new license situation disable this check so that new code not owned by MSFT could be added

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
         args: []
 
 # disabled until the new license situation has been sorted out:
-# 1. the old copyrights need to remain 
+# 1. the old copyrights need to remain
 # 2. probably need to modify the check to check for the new copyright's existence only
 # -   repo: local
 #     hooks:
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -33,14 +33,17 @@ repos:
...
--   repo: local
-    hooks:
-    -   id: check-torchdist
-        name: check-torchdist
-        entry: ./scripts/check-torchdist.py
-        language: python
-        exclude: ^(deepspeed/comm/|docs/|benchmarks/|scripts/check-torchdist.py|deepspeed/moe/sharded_moe.py|deepspeed/runtime/comm/coalesced_collectives.py|deepspeed/elasticity/elastic_agent.py|deepspeed/launcher/launch.py|tests/unit/comm/test_dist.py)
-        # Specific deepspeed/ files are excluded for now until we wrap ProcessGroup in deepspeed.comm
+# disabled until the new license situation has been sorted out:
+# 1. the old copyrights need to remain
```
loadams, line 37: Would we be not just be better with [this change ](https://github.com/deepspeedai/DeepSpeed/pull/7289/files#diff-4afee83e1c470278002402f04d6cceaffda6bc68208eaf6e3f0e654136b17486)to do the same thing?  Since that leaves the check for DeepSpeed Team which seems like a fine thing to keep for now?

stas00: Yes, it is better. Pushed the change you proposed. Thanks Logan.

loadams (approved):

PR #7307, opened 2025-05-23
Fix the GPU memory usage of ZeRO-Offload (update stage_1_and_2.py)

arminzhu:
Signed-off-by: Armin Zhu <mingzhengzhu1998@gmail.com>
Fix the memory usage of ZeRO-Offload with stage 1 and 2.
Before the fix, the memory usage is about 3x that of params_FP16. This is caused by the H2D data copy is using different data type.
Now the GPU memory usage is about 1x params_FP16. And the H2D memory copy needs a 16bit pinned memory buffer.

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -258,6 +258,9 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
         # that this process will update
         self.single_partition_of_fp32_groups = []
 
+        # a single 16-bit CPU pinned memory buffer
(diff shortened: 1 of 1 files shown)


arminzhu:
`cpu-torch-latest` reports error: `Need to provide pin_memory allocator to use pin memory.` 
@sfc-gh-truwase How to create a pinned memory tensor in DeepSpeed?

arminzhu:
cpu-torch-latest reports error: Need to provide pin_memory allocator to use pin memory.
@sfc-gh-truwase Do we have to use the `get_accelerator().pin_memory(Tensor)` to create a pinned memory tensor in DeepSpeed?

PR #7316, opened 2025-05-28
anchor transformers version

stas00:
some features require minimal transformers versions so let's start anchoring.

and fixing tests that break with recent transformers.

I need this fixed to be able to merge https://github.com/deepspeedai/DeepSpeed/pull/7268 which requires `transformers>=4.51.3`

### DIFF
--- a/requirements/requirements-dev.txt
+++ b/requirements/requirements-dev.txt
@@ -17,5 +17,5 @@ sphinx
 sphinx-rtd-theme
 tensorboard
 torchvision
-transformers>=4.39.0
+transformers>=4.51.3
 wandb
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -9,3 +9,4 @@ py-cpuinfo
 pydantic>=2.0.0
 torch
 tqdm
+transformers>=4.51.3


requirements/requirements-dev.txt
```diff
@@ -17,5 +17,5 @@ sphinx
 sphinx-rtd-theme
 tensorboard
 torchvision
-transformers>=4.39.0
+transformers>=4.51.3
```
loadams, line 20: I think a few of the test pipelines have issues with transformers > 4.48 maybe?  I believe we pinned nv-torch-latest for this reason

stas00: note I also added the anchor to the normal `requirements.txt` since deepspeed is now tightly connected with transformers (has been for a long time).

requirements/requirements.txt
```diff
@@ -9,3 +9,4 @@ py-cpuinfo
 pydantic>=2.0.0
 torch
 tqdm
+transformers>=4.51.3
```
sfc-gh-jrasley, line 12: this concerns me

stas00: removed

requirements/requirements-dev.txt
stas00: I adapted the CI job to use an older transformers, by reversing the install order. That should fix things.

PR #7327, opened 2025-06-02
Move pytest pinning from individual tests to requirements-dev.txt until fixed.

loadams:
pytest 8.4.0 seems to break a number of our tests, rather than pinning in each individually, we should just put this in the requirements file until we resolve the issue.


loadams:
Also need to unpin HPU tests after fix from #7337

PR #7362, opened 2025-06-16
Enable torch version dependent compilation of record_module and iter_params

deepcharm:
Dynamo breaks graphs because currently compile is disabled for a number of functions such as `iter_params` and `record_module`.

The above functions compile successfully for at least PyTorch version 2.7.0.

We enable the compilation based on the user PyTorch version using a new `compiler.enable(min_version=None)` decorator.
This should avoid the corresponding graph breaks and improve the performance.


tohtana (approved):
Hi @deepcharm,
Thank you for the improvement. I confirmed that this works with the latest PyTorch (v2.7).
I think it is good to merge this PR now, but I remember that the compiler of the older version threw an error. I'm not sure which version fixed it. We might need to turn on/off compiling this function depending on the PyTorch version.

deepcharm:
> Hi @deepcharm, Thank you for the improvement. I confirmed that this works with the latest PyTorch (v2.7). I think it is good to merge this PR now, but I remember that the compiler of the older version threw an error. I'm not sure which version fixed it. We might need to turn on/off compiling this function depending on the PyTorch version.

hi @tohtana, thanks for pointing this out - let me try to find the failing compiler version and then turn on/off that respectively.

loadams (approved):

loadams:
@deepcharm - is this ready to be merged?

deepcharm:
> @deepcharm - is this ready to be merged?

Hi @loadams, @tohtana 
I've added PyTorch version dependent `compiler.enable` decorator to ensure backward compatibility with older releases (< 2.7). Also enabled compilation for `iter_params`.

Kindly review, thanks.

PR #7371, opened 2025-06-19
fix #7188 

### LINKED ISSUE #7188: [BUG]when use 'overlap_comm:True' w/ 'contiguous_gradients:True', grad_norm is NaN
**Describe the bug**
As title, when enable overlap_comm and contiguous_gradients together, grad_norm will be nan (or be a constant float value in the latest master code, w/ this pr:https://github.com/deepspeedai/DeepSpeed/pull/7171 , seems still not fixed the root cause of nan). 
But it works fine when 'overlap_comm:false' w/ 'contiguous_gradients:true' OR  'overlap_comm:true' w/ 'contiguous_gradients:false' . It seems some bug behind contiguous_gradients, maybe some memory copy conflict? 

**To Reproduce**
1. in my private dataset, i can always reproduce it.
2. I think we can discuss on this issue, more likely to have some code reviews to find the bug? because it is strongly have a connection between 'overlap_comm' and 'contiguous_gradients'.

**Expected behavior**

**ds_report output**

**Screenshots**

 - OS: [e.g. Ubuntu 18.04]
 - GPU count and types [e.g. two machines with x8 A100s each]
 - Python version

**Launcher context**

**Docker context**

**Additional context**

lpnpcs:
I found that when using DeepSpeed Zero2 for my training task, the loss becomes 0 at the third step with a grad_norm of 1.414. This issue doesn't occur when using Zero3. I found the same issue #7188. After conducting a series of experiments, I identified the cause: there's a synchronization problem when using double ipg_buffer swapping. The issue was resolved after making modifications.

before 
image

after
image


tjruwase:
@lpnpcs, thanks for contributing this fix. I am a bit concerned of the perf impact of synchronizing the device. Are you able to measure the perf before/after the fix. This will help guide whether to pursue finer-grained synchronization on streams instead of device.

tjruwase:
@lpnpcs, please use the following to fix the formatting issues
https://github.com/deepspeedai/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites

lpnpcs:
> @lpnpcs, thanks for contributing this fix. I am a bit concerned of the perf impact of synchronizing the device. Are you able to measure the perf before/after the fix. This will help guide whether to pursue finer-grained synchronization on streams instead of device.

Thank you for your review. I conducted the following experiments to illustrate the impact on performance.

lpnpcs:
> @lpnpcs, thanks for contributing this fix. I am a bit concerned of the perf impact of synchronizing the device. Are you able to measure the perf before/after the fix. This will help guide whether to pursue finer-grained synchronization on streams instead of device.

Thank you for your review. I conducted the following experiments to illustrate the impact on performance.

I trained the Qwen2.5-vl-7b model using 8 A100 GPUs with 1,000 samples for 3 epochs. Below are the performances under several cases.

1  Original code
image

2 device synchronize
image

3 streams synchronize
image

Overall, adding synchronization makes the code slightly slower than the original, but it avoids bugs. Stream-level synchronization shows some improvement compared to device-level synchronization. Stream-level synchronization might be more precise and can also solve the issue, so I made some changes to the code.

hwchen2017:
Hi @lpnpcs , can you share your full repo - including source code, dataset, and launch script?   I’d be happy to help investigate further if I can reproduce the issue.

lpnpcs:
> Hi @lpnpcs , can you share your full repo - including source code, dataset, and launch script? I’d be happy to help investigate further if I can reproduce the issue.

Sorry, our dataset is somewhat sensitive. Everything except the dataset is public, and we used llamafactory to fine-tune qwen2.5-vl-7b. One characteristic of my dataset is that it contains a few pieces of dirty data, which might cause the grad_norm to become extremely large. At this point, the gradient turns into NaN, and then deepspeed processes it as -1. However, using zero3 or disabling overlap_comm or contiguous_gradients resolves the issue.

hwchen2017:
Can you enable comm_overlap and run your code with CUDA sanitizer and share the output if any?
`TORCH_CUDA_SANITIZER=1 python your_code.py`

FYI: https://pytorch.org/docs/stable/cuda._sanitizer.html

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -972,7 +972,10 @@ def reduce_independent_p_g_buckets_and_remove_grads(self, param, i):
         if self.elements_in_ipg_bucket + param.numel() > self.reduce_bucket_size:
             self.report_ipg_memory_usage("In ipg_remove_grads before reduce_ipg_grads", param.numel())
             self.reduce_ipg_grads()
-            if self.contiguous_gradients and self.overlap_comm:
+            if self.contigoous_gradients and self.overlap_comm:
+                if not get_accelerator().resolves_data_dependency():
+                    self.reduction_stream.wait_stream(get_accelerator().current_stream())
```
sfc-gh-truwase, line 999: Can you please explain why this mutual waiting between the streams is necessary? I notice the pattern is used elsewhere. But it seems like an overkill for managing double buffer concurrency between cooperating streams.

sfc-gh-truwase:
@lpnpcs, please fix conflict. Thanks!

jhwei:
I found similar issue recently while training Qwen2.5-VL-&B too. My solution is pretty similar.

I think the key issue is discovered in some issues like https://github.com/deepspeedai/DeepSpeed/issues/5545 and https://github.com/deepspeedai/DeepSpeed/pull/5606, that is the default stream should wait for the reduction stream.

PR 5606 claimed to have fixed this issue but it didn't. It does fix in some cases. but I think the default stream should wait for the reduction stream at the end of [reduce_ipg_grads](https://github.com/deepspeedai/DeepSpeed/blob/c2bb53f20fa32d6cbf472c08a42959a287dd9049/deepspeed/runtime/zero/stage_1_and_2.py#L1438) because the latter code in [reduce_independent_p_g_buckets_and_remove_grads](https://github.com/deepspeedai/DeepSpeed/blob/c2bb53f20fa32d6cbf472c08a42959a287dd9049/deepspeed/runtime/zero/stage_1_and_2.py#L1013) modifies the buffer at the same time. 

This PR fix the issue as well as it syncronize after `reduce_ipg_grads`. I think the reduction stream does not need to wait for the default stream(current stream) at this place as the default stream does not modify the buffer at the same time.

loadams:
@lpnpcs - would you be able to resolve the conflicts on this and we can get it merged?

lpnpcs:
> @lpnpcs - would you be able to resolve the conflicts on this and we can get it merged?

Done!

loadams:
> > @lpnpcs - would you be able to resolve the conflicts on this and we can get it merged?
> 
> Done!

Thanks, @lpnpcs - could you resolve the formatting fixes as well then we can merge?

lpnpcs:
> > > @lpnpcs - would you be able to resolve the conflicts on this and we can get it merged?
> > 
> > 
> > Done!
> 
> Thanks, @lpnpcs - could you resolve the formatting fixes as well then we can merge?

Done.

deepspeed/runtime/zero/stage_1_and_2.py
loadams: @lpnpcs - could you take a look at this comment?

sfc-gh-truwase: @lpnpcs and @jhwei I will appreciate your thoughts on the above question? It would be good to avoid excessive synch overhead. Thanks!

lpnpcs: Thanks for the suggestion! I'll run some tests later to see if those two streams really need to wait for each other.

hwchen2017: @sfc-gh-truwase The reduction stream needs to wait for the default stream, which is easy to understand because the communication should not begin until the computation on that buffer is done. 

But should the default stream wait for the reduction stream? Yes, in case that the communication is longer than the computation. To make the overlapping safe, mutual wait between the two stream is needed. 

 The third figure in #5606 shows the logic I explained. But I haven't investigated further on the implementation that can fix this issue.

tjruwase: @hwchen2017 thanks for the explanation.

GuCarpenter:
In this way, we have a synchronize point before reduction and after reduction, which means overlap_comm not works anymore, is this the proper solution?

sfc-gh-truwase:
@anyinlover thanks for raising concerns about overlap_comm perf. Do you have any numbers that show degradation? 

@lpnpcs and @jhwei I wonder if you have any thoughts on this? Did you measure with overlap_comm in your experiments?

Issue #7260, opened 2025-04-29
[BUG] AttributeError: 'UnembedParameter' object has no attribute 'dtype'


lambda7xx:
my code 
```
model_name_or_path = "meta/llama2-3.1B"

 mii_engine = mii.pipeline(model_name_or_path = args.model_type, tensor_parallel = 1)

mii.engine(prompt="hello, paris")

```
the error is

```
[rank0]:     deepseedfastgen_engine = get_engine(args, arrival_rate)
[rank0]:   File "/root/xxxx/DeepSpeed-MII/hybridserve/serve.py", line 60, in get_engine
[rank0]:     mii_engine = mii.pipeline(model_name_or_path = args.model_type, tensor_parallel = args.tensor_parallel)
[rank0]:   File "/root/xxxx/DeepSpeed-MII/mii/api.py", line 231, in pipeline
[rank0]:     inference_engine = load_model(model_config)
[rank0]:   File "/root/xxxx/DeepSpeed-MII/mii/modeling/models.py", line 17, in load_model
[rank0]:     inference_engine = build_hf_engine(
[rank0]:   File "/root/xxxx/.python/vllm/lib/python3.10/site-packages/deepspeed/inference/v2/engine_factory.py", line 135, in build_hf_engine
[rank0]:     return InferenceEngineV2(policy, engine_config)
[rank0]:   File "/root/xxxx/.python/vllm/lib/python3.10/site-packages/deepspeed/inference/v2/engine_v2.py", line 83, in __init__
[rank0]:     self._model = self._policy.build_model(self._config, self._base_mp_group)
[rank0]:   File "/root/xxxx/.python/vllm/lib/python3.10/site-packages/deepspeed/inference/v2/model_implementations/inference_policy_base.py", line 157, in build_model
[rank0]:     self.populate_model_parameters()
[rank0]:   File "/root/xxxx/.python/vllm/lib/python3.10/site-packages/deepspeed/inference/v2/model_implementations/inference_policy_base.py", line 201, in populate_model_parameters
[rank0]:     buffer, metadata = flatten_inference_model(container_map.transformer_params,
[rank0]:   File "/root/xxxx/.python/vllm/lib/python3.10/site-packages/deepspeed/inference/v2/model_implementations/flat_model_helpers.py", line 191, in flatten_inference_model
[rank0]:     total_size = process_layer(non_transformer_container, l_name, total_size)
[...]

PR #7378, opened 2025-06-22
fix(inference): Add missing dtype attribute to ParameterBase setter

Flink-ddd:
### Description

This PR fixes an `AttributeError: 'UnembedParameter' object has no attribute 'dtype'` that occurs in the Inference V2 engine. The issue is triggered when using a high-level interface like [DeepSpeed-MII](https://github.com/microsoft/DeepSpeed-MII) to run inference on models with tied input/output embeddings, such as Llama 2.

**Resolves: #7260**

### Root Cause Analysis

The root cause is that while the `ParameterBase` metaclass correctly creates property setters for parameter tensors, the setter function (`param_setter`) only assigns the tensor value itself. It does not propagate the tensor's `dtype` to the container instance.

Downstream functions, such as `flatten_inference_model`, expect every parameter container to have a `.dtype` attribute. When they encounter a custom container like `UnembedParameter` that lacks this attribute, an `AttributeError` is raised.

### The Fix

The solution is to modify the `param_setter` function within `make_param_setter` located in `deepspeed/inference/v2/model_implementations/parameter_base.py`.

I have added the line `self.dtype = value.dtype` immediately after the parameter tensor is assigned. This simple change ensures that any object inheriting from `ParameterBase` will now correctly expose the `dtype` of the tensor it wraps, resolving the error.

### Verification

This fix has been thoroughly verified in a containerized GPU environment (RunPod with PyTorch 2.1). The verification process involved:
1.  Cloning both the `deepspeed` and `DeepSpeed-MII` repositories from source.
2.  Installing the modified `deepspeed` library from this branch.
3.  Installing the `DeepSpeed-MII` library to trigger the bug.
4.  Running an end-to-end inference script with `mii.pipeline` and the `meta-llama/Llama-2-7b-hf` model.

The logs confirm that with this fix, the program successfully executes past the original point of failure. The `AttributeError` is completely resolved, and the DeepSpeed engine proceeds correctly to the model loading and C++/CUDA op compilation phase.

*(Note: The full end-to-end run in the test environment was ultimately blocked by a separate, pre-existing build issue in DeepSpeed's op builder (`ModuleNotFoundError: dskernels`), which is unrelated to this logic fix. The successful progression past the original error point serves as definitive proof of this fix's effectiveness.)*

### Related Context

This bug is primarily triggered via the [**DeepSpeed-MII**](https://github.com/microsoft/DeepSpeed-MII) project, which acts as a high-level wrapper for the DeepSpeed inference engine. A separate PR addressing a packaging issue in that repository was also created to facilitate this end-to-end verification.


tohtana (approved):
Thank you @Flink-ddd!

Flink-ddd:
Hi @tohtana ,
Thank you so much for your time in reviewing and merging this PR! I truly appreciate it.

I just wanted to provide a little more context on the companion PR for the DeepSpeed-MII project, as it was a necessary prerequisite for verifying this fix. The PR is:  **[deepspeedai/DeepSpeed-MII#567](https://github.com/deepspeedai/DeepSpeed-MII/pull/567)**.

The change to setup.py in that PR was essential. During testing in a clean containerized environment, the original find_packages() call in DeepSpeed-MII's setup script unreliably failed to discover the mii package, leading to a persistent ModuleNotFoundError. Making the package discovery explicit with include=['mii', 'mii.*'] resolves this build issue, ensuring a robust installation from source and improving the developer experience.

If you have a moment, I would be grateful if you could take a look. No rush at all, and thanks again for maintaining this great project!

PR #7379, opened 2025-06-22
Add support for ws=1 scenario

NirSonnenschein:
add support for ws=1 scenario and make sure
fields are always initialized for robustness.


inkcherry (commented):
LGTM thanks!

loadams (approved):

PR #7385, opened 2025-06-24
add blog link

stas00:
add a blog link to Arctic Long Sequence Training (ALST) with DeepSpeed


stas00:
@loadams, why do we run code CI on README.md changes?

loadams (approved):

loadams:
> @loadams, why do we run code CI on README.md changes?

@stas00 - we probably shouldn't, but we've just never added the README to the list of file paths to exclude.  We really should only run the formatter there I'd think.

stas00:
perhaps in addition to excluding paths, it'd be more sustainable to also filter by file type? e.g. never trigger code CI jobs on `.md` and `.rst` files?

PR #7387, opened 2025-06-25
Add RISCV64 CPU Support

heyujiao99:
Hi, 
This patch adds riscv64 support for the deepspeed_shm_comm operator，enabling DeepSpeed to perform pure CPU training/inference on RISCV64 hosts, for research purposes. 

Unit tests related have passed：

`
tests/unit/comm/test_dist.py::TestDistInferenceAllReduce::test[dtype0] PASSED                                                   [ 33%]
tests/unit/comm/test_dist.py::TestDistInferenceAllReduce::test[dtype1] PASSED                                                   [ 66%]
tests/unit/comm/test_dist.py::TestDistInferenceAllReduce::test[dtype2] PASSED                                                   [100%]
`


loadams:
@heyujiao99 - would it be worth adding this as a unique accelerator?  Or do you foresee this as the only difference compilation wise?

heyujiao99:
> @heyujiao99 - would it be worth adding this as a unique accelerator? Or do you foresee this as the only difference compilation wise?

In the long run, a unique accelerator is worthwhile because RISC-V CPUs may have different acceleration/computation backends, which could differ from Intel CPUs. By the way, does the DeepSpeed community have plans to support different architecture CPUs?

loadams:
> > @heyujiao99 - would it be worth adding this as a unique accelerator? Or do you foresee this as the only difference compilation wise?
> 
> In the long run, a unique accelerator is worthwhile because RISC-V CPUs may have different acceleration/computation backends, which could differ from Intel CPUs. By the way, does the DeepSpeed community have plans to support different architecture CPUs?

@heyujiao99 - We are certainly happy to support them if users are able to contribute/validate them.  We don't have any to test with ourselves.  Also @tjruwase this is similar to our discussion before on switching cpu_accelerator to xeon_accelerator and making the existing cpu_accelerator more generic?

delock:
Hi @heyujiao99 , do you know whether in PyTorch, RISCV64 shares the same 'cpu' device type with Intel device?   Do you see RISCV64 difference only in OpBuilder, or do you forsee difference in other pytorch code as well?   If most code is the same and only OpBuilder is different, then an abstraction in OpBuilder and create sub-directory under `csrc/cpu/` such as `csrc/cpu/comm/x86_64`, `csrc/cpu/comm/riscv64` might be better choice.

delock:
@heyujiao99 @loadams @tjruwase following up this discussion.  From current PR, I see that main difference between x86-64 CPU and riscv CPU are how C++ kernel looks like and how to build them.  For building the kernel, OpBuilder base class should have a way to choose the right build flag.   For kernel itself, there are two situations:
1. Kernel code is different between x86_64 and riscv, in that case, we need a _naming convention_ to help OpBuilder select the right source code to build.  The idea is to give a 'base' name and let OpBuilder add the right prefix or suffix to pick the right kernel code.
2. Kernel code is same between x86_64 and riscv, in that case, the same source code can be used.
3. That said, we could introduce a mechanism that pick kernel source code with following logic: a. pick 'machine' specific kernel code if possible.  b. pick generic kernel code if no machine specific code is found.  c. raise error if either not found.

With this mechanis we should be able to catch most difference between x86_64 and riscv, and also extendable to other CPU ISA if needed.

tjruwase:
@delock thanks for the analysis. I am aligned with your proposal that we should reuse as much as possible, even if some refactoring is needed. 

@heyujiao99, it would be great to get your thoughts on how to minimize code duplication in this PR. In particular, `shm-riscv64.cpp` appears to duplicate `shm.cpp`.

heyujiao99:
> @delock thanks for the analysis. I am aligned with your proposal that we should reuse as much as possible, even if some refactoring is needed.
> 
> @heyujiao99, it would be great to get your thoughts on how to minimize code duplication in this PR. In particular, `shm-riscv64.cpp` appears to duplicate `shm.cpp`.

Thanks for all the discussions. I totally agree with these, and I should find a more extendable way to support riscv cpu. But I found that some dependencies like PyTorch and cpuinfo also lack mature support for RISC-V CPUs, and I'll explore it further and attempt later when I have a deeper understanding.

delock:
> > @delock thanks for the analysis. I am aligned with your proposal that we should reuse as much as possible, even if some refactoring is needed.
> > @heyujiao99, it would be great to get your thoughts on how to minimize code duplication in this PR. In particular, `shm-riscv64.cpp` appears to duplicate `shm.cpp`.
> 
> Thanks for all the discussions. I totally agree with these, and I should find a more extendable way to support riscv cpu. But I found that some dependencies like PyTorch and cpuinfo also lack mature support for RISC-V CPUs, and I'll explore it further and attempt later when I have a deeper understanding.

Thanks @heyujiao99 !  Looking forward for your PR.  If you need discussion ping me on this thread or via email, I'll be happy to see new architecture running DeepSpeed.

PR #7394, opened 2025-06-27
add support for CUDAtk12.9

loscrossos:
CUDA Toolkit 12.9 has been out for a while. The build currently fails when it is installed as the builder checks against hardcoded values. this PR adds the value 12.9. a better mechanism would be to check dynamically that the major number is the same... maybe next time when CUDA13 comes out :)


stas00:
@tjruwase, @loadams - perhaps we should switch to not check minor versions - so far 10.x, 11.x and 12.x we have had to make a PR to hardcode every CUDA minor release version - so every new CUDA release users suffer.

Perhaps if the history tells us something we should switch to a reverse logic and not accept the mismatch situation only if we know that some minor version is actually not compatible with the rest of the major version array.

p.s. a new release is needed. thank you!

loadams:
> @tjruwase, @loadams - perhaps we should switch to not check minor versions - so far 10.x, 11.x and 12.x we have had to make a PR to hardcode every CUDA minor release version - so every new CUDA release users suffer.
> 
> Perhaps if the history tells us something we should switch to a reverse logic and not accept the mismatch situation only if we know that some minor version is actually not compatible with the rest of the major version array.
> 
> p.s. a new release is needed. thank you!

@stas00  - agreed on this, we probably should - I think this came from problems we saw in Cuda 11.1 and 11.2 early on if I recall.  But those were fixed with the official versions.

And I missed this, but the new release should be out now.

stas00:
the minor case started with 10.x already. I was the one who proposed to ignore the minor version diff, since it was often in a conflict with pre-installed system-wide cuda - vs cuda torch was built with. 

So you're saying cuda 11.1 and 11.2 actually did need to be matched? If so what about my proposal to not require matching minor versions unless we know a particular minor version has to be matching?

PR #7402, opened 2025-06-30
fix: DeepCompile for torch 2.8

HollowMan6:
In torch v2.8.0, all symm mem code are moved into a dedicated folder https://github.com/pytorch/pytorch/commit/ffc6cbfaf78ca219092ce64dcf113377ae698300

So this PR tries to address this change by checking if we have it located under `torch/csrc/distributed/c10d/symm_mem/SymmetricMemory.hpp` (new location). If not, we fall back to the original place for backward compatibilities.

This PR also clean up some includes in `z1/2/3.cpp` that has already been included in `deepcompile.h`

### DIFF
--- a/csrc/compile/z1.cpp
+++ b/csrc/compile/z1.cpp
@@ -6,17 +6,6 @@
 #include "z1.h"
 #include "deepcompile.h"
 
-#define USE_C10D_NCCL
-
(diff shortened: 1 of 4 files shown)


tohtana:
Looks good to me. Thank you for your contribution, @HollowMan6!

tohtana (approved):

Issue #7251, opened 2025-04-25
[BUG] deepspeed offload states is not compatible with ZeRO2


hijkzzz:
In OpenRLHF, we want to offload/reload the states of the Adam optimizer, and at the same time use auto tensor parallelism for RL training. However, I found that Zero2 is not compatible with offloading/reloading optimizer states. Could you please provide support for this?
@loadams

loadams:
Tagging @tjruwase for FYI as well

tohtana:
Hi @hijkzzz,
Since some parts will be shared with ZeRO-3, could you help us add support for ZeRO-2 as well? The entry points are already defined in the engine: [offload](https://github.com/deepspeedai/DeepSpeed/blob/0e741714f5a708b75a4db0a8140f9016f3aa5fd3/deepspeed/runtime/engine.py#L3943), [reload](https://github.com/deepspeedai/DeepSpeed/blob/0e741714f5a708b75a4db0a8140f9016f3aa5fd3/deepspeed/runtime/engine.py#L3977). The actual logic currently exists only in the ZeRO-3 optimizer. What we need is to remove the assertion checking the ZeRO stage and add `offload_states` and `reload_states` in [ZeRO1/ZeRO2 optimizer](DeepSpeedZeroOptimizer).

You can refer to the ZeRO-3 implementations of [`offload_states`](https://github.com/deepspeedai/DeepSpeed/blob/0e741714f5a708b75a4db0a8140f9016f3aa5fd3/deepspeed/runtime/zero/stage3.py#L2939) and [`reload_states`](https://github.com/deepspeedai/DeepSpeed/blob/0e741714f5a708b75a4db0a8140f9016f3aa5fd3/deepspeed/runtime/zero/stage3.py#L3024) in the ZeRO3 optimizer. The offloading/reloading of Adam will be similar.

If you can draft a PR, we'd be happy to review it!

PR #7421, opened 2025-07-08
Support DeepSpeed offload and reload states with ZeRO1 and ZeRO2

LYMDLUT:
Resolves: #7251


tohtana:
Hi @LYMDLUT, thank you for the great work!

Does this code actually work for ZeRO1 as well? If it doesn't, can we keep the assertions changing the condition?
We have [unit tests](https://github.com/deepspeedai/DeepSpeed/blob/master/tests/unit/runtime/zero/test_offload_states.py) for this feature. Can you add stage 2 (and 1 if it works) to the test argument?

LYMDLUT:
> Hi @LYMDLUT, thank you for the great work!
> 
> Does this code actually work for ZeRO1 as well? If it doesn't, can we keep the assertions changing the condition? We have [unit tests](https://github.com/deepspeedai/DeepSpeed/blob/master/tests/unit/runtime/zero/test_offload_states.py) for this feature. Can you add stage 2 (and 1 if it works) to the test argument?

Thank you for your review. I have added a test unit for zero1 and zero2. However, I am not very familiar with DeepSpeed. Could you please carefully check the code and provide other guidance? I will do my best to cooperate with you.

LYMDLUT:
Looking forward to your guidance.

tohtana:
Hi @LYMDLUT,
Sorry for my late response and thank you for adding the test! 

1. This actually works for ZeRO1 when I tried. How about add a parameter like this?
```python
@pytest.mark.parametrize("zero_stage", [1, 2])
...
    def test_offload_states_zero2(self, included_state, pin_memory, non_blocking, zero_stage):
        hidden_dim = 1024
        config_dict = {
            "optimizer": {"type": "Adam", "params": {"lr": 1e-6}},
            "zero_optimization": {"stage": zero_stage},
            "bf16": {"enabled": True}
        }
```

2. As the behavior of memory release is often tricky, can you check if the memory usage is actually reduced? You can find the example [here](https://github.com/deepspeedai/DeepSpeed/blob/c2bb53f20fa32d6cbf472c08a42959a287dd9049/tests/unit/runtime/zero/test_offload_states.py#L65).

3. Can we consolidate the new test into `test_offload_states.py`? The initialization and validation of values are a bit different from ZeRO3, I think we can just have another function in `TestOffloadStates`. If you can even consolidate functions (e.g. `run_model` in `test_offload_states` and `run_model_zero2`), it would be great. We don't even need to have the logic to check the memory usage in both `run_model` and `run_model_zero2`.

LYMDLUT:
> @tohtana , I really need your guidance and help!

> 2. As the behavior of memory release is often tricky, can you check if the memory usage is actually reduced? You can find the example [here](https://github.com/deepspeedai/DeepSpeed/blob/c2bb53f20fa32d6cbf472c08a42959a287dd9049/tests/unit/runtime/zero/test_offload_states.py#L65).

When I was performing the second step, I noticed that the memory allocated (as indicated by memory_allocated()) for hp_param, lp_param, and lp_grads remained unchanged, with only the memory for optim_states showing a decrease. However, the corresponding tensors had indeed been transferred from CUDA to the CPU. 

Despite debugging, I couldn't figure out the reason. I will continue to conduct in-depth research on this issue. However, if you know how to solve it, it will be of great help to me. Thank you very much.

tohtana:
Hi @LYMDLUT,

Thank you for enhancing the test!

I think we still have a reference to the buffer somewhere. PyTorch's memory allocator frees the memory only when there is no reference to the buffer. Some operators like `view` and `narrow` creates a view of the buffer, which refer to the buffer. We need to make sure we don't have any these views.

Can you first check if we create such views of the buffer? We need to free the views when offloading and reconstruct it when reloading.

LYMDLUT:
Hi，@tohtana ，
Thank you for your guidance! I have already resolved the issues with optim_states, hp_param, and lp_param, but I still can't figure out what is wrong with lp_grads.Could you please provide further guidance?

LYMDLUT:
Hi @tohtana,
We may have resolved most of the issues. Before I delete tests/unit/runtime/zero/test_offload_states_zero2.py, I hope you can review the update changes to ensure the correctness of the code. I look forward to your further guidance.

LYMDLUT:
The contiguous_grad_buffer should not need to be implemented in zero1 and zero2. I believe most of the work has already been completed, and I hope to receive further guidance. Meanwhile, it should be noted that for the lp_param function, some tensors that do not require gradients (such as inv_freq_expanded in ROPE) will still remain on the GPU's cuda instead of being offloaded to the CPU.

LYMDLUT:
image
image
image

Based on OpenRLHF version 0.8.10 and DeepSpeed, the following operations including offload, reload_optim_states, contiguous_grad_buffer, hp_params, and lp_grads were performed on zero2 and zero3. From the result curves, it seems that zero2 can work properly.

sfc-gh-truwase:
@LYMDLUT thanks sharing those curves. That looks great. Can you please address the formatting issues via 
https://github.com/deepspeedai/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites

sfc-gh-truwase:
Hmm, that is strange since the CI still fails: 

https://github.com/deepspeedai/DeepSpeed/actions/runs/17046165288/job/48322449749?pr=7421. 

Perhaps your versions are different from expected:  `pre-commit== 4.3.0` and `clang-format== 18.1.3`.

LYMDLUT:
I tested the results of OpenRLHF on zero1, zero2, zero3, and autotp with zero1 and ds_tensor=2. From the curves, the effect is normal.
image
image
image
image

LYMDLUT:
tests/unit/runtime/zero/test_offload_states_zero2.py

Please remember to delete this intermediate test file before merging.

sfc-gh-truwase:
> tests/unit/runtime/zero/test_offload_states_zero2.py
> 
> Please remember to delete this intermediate test file before merging.

@LYMDLUT since you are the author, can you not delete this file now?

LYMDLUT:
> > tests/unit/runtime/zero/test_offload_states_zero2.py
> > Please remember to delete this intermediate test file before merging.
> 
> @LYMDLUT since you are the author, can you not delete this file now?

done！

sfc-gh-truwase:
@LYMDLUT thanks for the quick delete. Can you check formatting as it seems to have broken again?.

LYMDLUT:
> @LYMDLUT thanks for the quick delete. Can you check formatting as it seems to have broken again?.

image
I don't know what kind of format error it is. Could you offer some help?

sfc-gh-truwase (approved):

sfc-gh-truwase:
> I don't know what kind of format error it is. Could you offer some help?

Similar to earlier [discussion](https://github.com/deepspeedai/DeepSpeed/pull/7421#issuecomment-3197683840), can you run `pre-commit run --all-files` using `pre-commit==4.3.0` and `clang-format==18.1.3`?

sfc-gh-truwase:
@LYMDLUT actually I just made the formatting changes locally and pushed into your branch. Let's see if that works.

tohtana:
Hi @LYMDLUT,
I am glad this PR was merged and really appreciate your great effort! Sorry that I wasn't actively help this.

Can I ask about the details for the future improvement? I see some similarity between `offload_optimizer_states` and `offload_adam_states` (`reload_optimizer_states` and `reload_optimizer_states` as well). Do you think there is a challenge to refactor them to consolidate the functions?

PR #7424, opened 2025-07-10
trying to fix nv-accelerate-v100.yml CI job

stas00:
trying a day old accelerate from the day before https://github.com/huggingface/accelerate/commit/1ac8643df70273fcf2e2f0605a6bd8e64a4e49d7

### DIFF
--- a/.github/workflows/nv-accelerate-v100.yml
+++ b/.github/workflows/nv-accelerate-v100.yml
@@ -57,4 +57,4 @@ jobs:
           # force protobuf version due to issues
           pip install "protobuf<4.21.0"
           pip list
-          pytest $PYTEST_OPTS --color=yes --durations=0 --verbose tests/deepspeed      
+          pytest $PYTEST_OPTS --color=yes --durations=0 --verbose tests/deepspeed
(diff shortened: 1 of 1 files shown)


stas00:
I diffed the installed package lists for good/bad and I can now repro, the problem comes from the just released `datasets==4.0.0` which is the source of this breakage. 

if someone needs this to work right away, force `datasets==3.6.0`

stas00:
So it's comparing:

```
def simple_accuracy(preds, labels):
    print(preds, labels)
    print(f"{preds==labels}")
    return float((preds == labels).mean())
```
data:
```
Column([1, 0, 0, 1, 1]) Column([1, 0, 0, 1, 0])
False
```
and fails there as it returns a bool, and bool has no `mean`

stas00:
reported it here https://github.com/huggingface/datasets/issues/7679 and https://github.com/huggingface/accelerate/issues/3676

stas00:
so we can merge the workaround to force datasets==3.6.0 in this PR until upstream fixes it.

stas00:
update: they are working on fixing the `evaluate` package.

Issue #7423, opened 2025-07-09
[REQUEST] Expose TP/DP/PP Rank as Environment Variable for integration with low Level Profiling Tools


WoosungMyung:
First, thank you for providing such a powerful and scalable framework for distributed training.

I would like to propose a small but practical feature that would help with system-level profiling using external tracing tools like CUPTI.

While torchrun already sets RANK, LOCAL_RANK, and WORLD_SIZE as environment variables, DeepSpeed
internally computes parallelism ranks: TP / DP/ PP Rank using APIs like topo.get_tensor_model_parallel_rank( )

However, these ranks are not exposed as environment variables, which makes it difficult to access them from low level tools such as CUPTI.

Would you consider adding  the following environment variables after initialize parallelism?

like

        os.environ["TP_RANK"] = str(self._topo.get_tensor_model_parallel_rank(rank))
        os.environ["PP_RANK"] = str(self._topo.get_pipeline_model_parallel_rank(rank))
        os.environ["DP_RANK"] = str(self._topo.get_data_parallel_rank(rank))

Would you consider accepting a PR for this?

Thank you again for your work and consideration!
Looking forward to your thoughts.

Best Regards.
[Woosung Myung]

sravan500:
Hi @WoosungMyung
I am new to deepspeed and I am using deepspeed for distributed finetuning with seqtoseqtrainer.
I observed that with deepspeed(v 14.5), the training time is longer than FSDP2.

According to below issue:
https://github.com/deepspeedai/DeepSpeed/issues/5047

Can you please help me out and confirm if FP32 precision is the default for optimizer states as mentioned in this blog and that is the reason for latency in deepspeed?

Please help me.
Thanks in advance.

Best regards,
Sravan

tjruwase:
@WoosungMyung thanks for your question and PR offer. Yes, we are interested in making DeepSpeed easier to integrate with profiling tools. We would appreciate your PR. 

Unless I misunderstand, but I do think a design that involves DeepSpeed modifying env variables after launch is problematic since those env variable changes (1) could unexpectedly leak to other processes, and (2) need to be reset completion to avoid unexpected side effects. Another problem is that env vars are typically system-wide whereas rank information is process-specific. Please clarify in case I misunderstood something. 

I think a better approach would be to add these APIs to the DS engine/runtime so that the client process can query and use as appropriate. What do you think?

tjruwase:
[...]

PR #7427, opened 2025-07-12
Add getter APIs for TP/PP/DP ranks in DeepSpeedEngine

WoosungMyung:
Thanks again for giving opportunity for improving this Community!
This PR is from Issue #7423.

1) Motivation

To improve compatibility with low-level profiling tools (e.g., NVIDIA CUPTI or DCGM), it can be useful to expose parallelism-specific rank (tensor/pipeline/data) at the engine level.

2) Changes

I Added three getter methods to DeepSpeedEngine:
  - get_tensor_parallel_rank()
  - get_pipeline_parallel_rank()
  - get_data_parallel_rank()

Thank you for reviewing this contribution!


deepspeed/runtime/engine.py
```diff
@@ -730,6 +730,27 @@ def random_ltd_initialize(self):
...
+            return None
+
+    def get_pipeline_parallel_rank(self):
+        # Return pipeline parallel rank if mpu is defined, otherwise return None
+        if self.mpu:
+            return self.mpu.get_pipeline_model_parallel_rank()
+        else:
+            return None
+
+    def get_data_parallel_rank(self):
```
sfc-gh-truwase: Please take a close look at [utils/groups.py](https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/utils/groups.py). I believe all the needed APIs are available there, such as [this](https://github.com/deepspeedai/DeepSpeed/blob/affee605e47c9befd21c4c1445e11fd29d295201/deepspeed/utils/groups.py#L182). I think you just need to connect them.

WoosungMyung: @sfc-gh-truwase 
Thank you for the feedback. I've updated the implementation in the second commit to expose only data/tensor/model parallel ranks using existing group APIs, as suggested.

sfc-gh-truwase: @WoosungMyung thanks for making the changes. Can you confirm that the PR is working for your scenario?

WoosungMyung: @sfc-gh-truwase Thanks for your Response!! 

1) Having getter methods in DSEngine class works for integrating with DCGM. And I think It is essensial  in scenarios like RLHF or Knowledge Distillation, where multiple engine instances may run within a single process. So I’m considering adding the same methods to PipelineEngine as well in this PR. I would like to hear your thoughts.


2) For CUPTI integration, exporting ranks via environment variables still seems necessary , since it operates via C/C++-level hooking. I’ve been thinking about introducing an argument-based (boolean flag) approach to control whether Parallelism rank should be exported to the environment, which may help avoid issues in multi-engine cases. Adding a boolean argument like 'export_parallel_rank_env` (default: false). However, I understand this may add complexity and possibly go against the design philosophy of DeepSpeed. I’d totally understand if this direction is not preferred, and would really appreciate your feedback.

sfc-gh-truwase: @WoosungMyung apologies for late response. 
1. Can you double check that the getter methods are not available to PipelineEngine? If not, yes adding would be great. 
2. Yes, the complexity of argument for exposing env_var is not ideal. In the case, it seems that creating a client code that uses DeepSpeed as backend might be a better design. This is what frontends such as HF Transformers, Accelerate, Lightning, etc. do. It provides greater flexibility on the client side while keeping DeepSpeed simple.

WoosungMyung: @sfc-gh-truwase  Thanks for the reply.
I added a getter method for pipeline parallelism in PipelineEngine, since it was not defined in the DeepSpeedEngine and therefore could not be inherited. As for the second suggestion, I fully understand that it was not accepted in order to maintain the simplicity that is central to DeepSpeed’s design philosophy! Thanks

WoosungMyung:
@sfc-gh-truwase 
Hi, Given that my PR only introduces a getter method in `Accelerator` and doesn't affect MoE or Zero-related logic, I believe these failures are unrelated to the changes : xpu-max1100. Thanks

sfc-gh-truwase:
@delock can you please help with the xpu CI? Thanks!

loadams:
> @delock can you please help with the xpu CI? Thanks!

I've also reached out to @Liangliang-Ma on this as well.  The test is currently skipped on this PR so it should merge now.

loadams (approved):

WoosungMyung:
Thanks a lot for merging the previous PR! I really appreciate the review and guidance throughout the process!

PR #7428, opened 2025-07-14
Use past_key_value when provided

deepcharm:
The KV cache can be passed via `layer_past` or `past_key_value` arguments. Previously, `past_key_value` was ignored, causing workload incompatibilities.

This PR fixes the issue while preserving the original logic.

### DIFF
--- a/deepspeed/model_implementations/transformers/ds_transformer.py
+++ b/deepspeed/model_implementations/transformers/ds_transformer.py
@@ -133,6 +133,9 @@ class DeepSpeedTransformerInference(nn.Module):
         if "hidden_states" in kwargs:
             input = kwargs["hidden_states"]
 
+        if layer_past is not None and past_key_value is not None:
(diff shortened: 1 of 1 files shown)


deepcharm:
While reviewing this change, we also noted a possible existing bug, such that `self.layer_last` is not used at even iterations, since the `layer_last` argument is overwritten. We fixed it by using a local var name `_layer_last` and preserving the passed argument value.

Should it be a separate PR maybe?

sfc-gh-truwase (approved):

sfc-gh-truwase:
> While reviewing this change, we also noted a possible existing bug, such that `self.layer_last` is not used at even iterations, since the `layer_last` argument is overwritten. We fixed it by using a local var name `_layer_last` and preserving the passed argument value.
> 

Thanks for providing this fix. 

> Should it be a separate PR maybe?

Yes, it would be better as a separate PR.

PR #7431, opened 2025-07-16
fix issues raised by Coverity scans

NirSonnenschein:
This commit combines fixes for 37 potential code issues found in Coverity scans.
the issues include but are not limited to potential access to uninitialized variables, dead and redundant code.
We understand that reviewing such a commit can be difficult and will be happy to help with any questions or changes required.

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -534,6 +534,9 @@ class PipelineEngine(DeepSpeedEngine):
         """True if this process is in the last stage in the pipeline."""
         return self.stage_id == self.num_stages - 1
 
+    def get_pipeline_parallel_rank(self):
+        return self.stage_id
+
     def _reduce_outputs(self, outputs, reduce='avg', reduce_dp=True, micro_batches=None):
         if reduce is None:
             return outputs
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -729,6 +729,15 @@ class DeepSpeedEngine(Module):
             raise ValueError(f'not yet support')
             #self.lr_scheduler = lr_schedules.WarmupLayerTokenDecayLR(self.optimizer, self.random_ltd_scheduler)
 
+    def get_data_parallel_rank(self):
+        return groups.get_data_parallel_rank()
+
+    def get_tensor_parallel_rank(self):
+        return groups.get_tensor_model_parallel_rank()
+
+    def get_model_parallel_rank(self):
+        return groups.get_model_parallel_rank()
+
     def get_sequence_parallel_group(self):
         return self.seq_parallel_group
 


deepspeed/module_inject/replace_module.py
```diff
@@ -316,11 +322,13 @@ def replace_wo_policy(module, all_reduce_linears, prefix="", state_dict=None):
         return _autotp._replace_module(module)
 
     def replace_fn(child, _policy, layer_id=0, prefix="", state_dict=None):
-        training = False  # todo: refactor this part to go in the config
-        if training:
```
sfc-gh-truwase, line 320: Based on old L319, I don't understand why new L325-331 is introduced. Since `if training` is False, I think only L333+ should remain. Can you please help clarify?

deepspeed/runtime/pipe/engine.py
```diff
@@ -709,7 +708,6 @@ def _next_batch(self):
 
     def _exec_forward_pass(self, buffer_id):
         self.tput_timer.start()
-        self.mem_status('BEFORE FWD', reset_max=True)
```
sfc-gh-truwase, line 715: I believe these debug statements are used for memory profiling. Is there a way to preserve them even when disabled?

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -1563,7 +1563,7 @@ def _ensure_availability_of_partitioned_params(self, params):
                 assert param.ds_tensor.final_location == OffloadDeviceEnum.nvme and param.ds_status == ZeroParamStatus.NOT_AVAILABLE
                 swap_in_list.append(param)
             if param.ds_tensor.status == PartitionedParamStatus.INFLIGHT:
-                assert param.ds_tensor.final_location == OffloadDeviceEnum.nvme and param.ds_status == ZeroParamStatus.NOT_AVAILABLE
+                assert param.ds_tensor.final_location == OffloadDeviceEnum.nvme and param.ds_status == ZeroParamStatus.INFLIGHT
```
sfc-gh-truwase: I believe this change is incorrect because `param.ds_status` cannot be in `INFLIGHT` until `param.ds_tensor.status` is `AVAILABLE`.

sfc-gh-truwase: I believe this change is incorrect because `param.ds_status` cannot be in `INFLIGHT` until `param.ds_tensor.status` is `AVAILABLE`.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1656,16 +1656,11 @@ def allreduce_bucket(self, bucket, rank=None, log=None):
...
         tensor_to_allreduce.div_(dist.get_world_size(group=self.dp_process_group) / float(self.sequence_parallel_size))
 
-        if rank is None:
-            #    "All Reducing"
-            dist.all_reduce(tensor_to_allreduce, group=self.dp_process_group)
-        else:
-            global_rank = dist.get_global_rank(self.dp_process_group, rank)
-            dist.reduce(tensor_to_allreduce, global_rank, group=self.dp_process_group)
+        #    "All Reducing"
+        dist.all_reduce(tensor_to_allreduce, group=self.dp_process_group)
```
sfc-gh-truwase, line 1660: This is incorrect as it replaces a `dist.reduce()` with `dist.all_reduce()` on the else path.

sfc-gh-truwase: On a second view, I see that this is fine because of L1644 which makes the removed code dead.

sfc-gh-truwase: This code has a deeper problem since the hardcoded L1644 is quite old and is now dead code after these changes.

deepspeed/module_inject/replace_module.py
NirSonnenschein: Thanks @sfc-gh-truwase , it looks like there was mistake due to a rebase which we didn't catch. will upload a fix shortly.
the correct change is to remove the if statement and leave just the else block.

deepspeed/runtime/pipe/engine.py
NirSonnenschein: Thanks @sfc-gh-truwase , 
the self.mem_status function was removed due to not doing anything (the first line in the function was return - see engine.py line 1308) and as a result all calls to the function were removed as well since they do nothing.
we can remove this from the PR if you prefer, but this code currently will do nothing unless somehow overridden (e.g. in part of an inheritance).

deepspeed/runtime/zero/partition_parameters.py
NirSonnenschein: thanks @sfc-gh-truwase ,
you are probably right, I'll explain the logic provided:
there are 2 if and assert blocks 
```python
            if param.ds_tensor.status == PartitionedParamStatus.NOT_AVAILABLE:
                assert param.ds_tensor.final_location == OffloadDeviceEnum.nvme and param.ds_status == ZeroParamStatus.NOT_AVAILABLE
                swap_in_list.append(param)
            if param.ds_tensor.status == PartitionedParamStatus.INFLIGHT:
                assert param.ds_tensor.final_location == OffloadDeviceEnum.nvme and param.ds_status == ZeroParamStatus.INFLIGHT
```                
  the first if checks that param.ds_tensor.status is NOT_AVAILABLE and asserts that param.ds_status will be NOT_AVAILABLE , the second checks that param.ds_tensor.status is INFLIGHT and asserts that param.ds_status is NOT_AVAILABLE , Coverity assumed that if the first if condition was like the assert , then the second if condition should be like the second assert. 
  if this is not correct we will remove the chage.

NirSonnenschein: same as above comment

sfc-gh-truwase: Thanks for the explanation. The change is incorrect, please revert. 

With z3+nvme_offload, param materialization involves the following two-step process and status transitions 

1. Swap from disk to HBM with `param.ds_tensor.status`:  NOT_AVAILABLE -> IN_FLIGHT -> AVAILABLE
2. All-gather across GPUs with `param.ds_status`: NOT_AVAILABLE -> IN_FLIGHT -> AVAILABLE

NirSonnenschein: thanks @sfc-gh-truwase for the explanation, removed this change

deepspeed/runtime/pipe/engine.py
sfc-gh-truwase: I think it is fine to remove. It is better to write new memory profiling code using latest tools and APIs.

loadams:
FYI @NirSonnenschein and @sfc-gh-truwase we will get this merged shortly let me fix the CI blockers

loadams (approved):

tjruwase:
Ignoring flaky nv-ds-chat CI for now.

NirSonnenschein:
thanks @tjruwase

PR #7445, opened 2025-07-23
Remove additional unused tests (human-eval)

loadams:


stas00:
Logan, I don't know who/why those were added in the first place to know if they are ok to remove, so perhaps blame to find who to tag?

If we don't have that info I'm OK to remove those... but best to ask first?

loadams:
> Logan, I don't know who/why those were added in the first place to know if they are ok to remove, so perhaps blame to find who to tag?
> 
> If we don't have that info I'm OK to remove those... but best to ask first?

@stas00  - this was added by @lekurile originally but we never got it working all the way.  I'll follow up with him on this.

loadams:
> Sure, I will approve then and you can quickly merge if the author is in agreement.
> 
> Thank you for checking, Logan.

@stas00 - checked with Lev and this was good to be merged, the examples in DSE should stay: https://github.com/deepspeedai/DeepSpeedExamples/tree/master/evaluation/inference/human_eval

PR #7454, opened 2025-07-28
Support Muon Optimizer

qimcis:
### Adds Muon optimizer support 

Key Changes:

- New Muon optimizer integration with `MuonWithAuxAdam` support
- Hybrid parameter groups: automatic segregation based on `_use_muon` attribute  
- Unit tests for the optimizer


qimcis:
Not sure if this is the best way to be adding it to the dependencies list, Muon isn't on PyPi yet as far as I'm aware. If there's a better way for us to be doing this, let me know!! Perhaps I could move it to an extra instead?

qimcis:
@loadams If you have the bandwidth to look into this sometime it'd be greatly appreciated!

requirements/requirements.txt
```diff
@@ -1,6 +1,7 @@
 einops
 hjson
 msgpack
+muon-optimizer @ git+https://github.com/KellerJordan/Muon.git
```
loadams: I suspect we only need this in the -dev?

loadams:
> Not sure if this is the best way to be adding it to the dependencies list, Muon isn't on PyPi yet as far as I'm aware. If there's a better way for us to be doing this, let me know!! Perhaps I could move it to an extra instead?

An extra might be better.  We aren't opposed to this but it would be preferred if there was a version on pypi that was going to be maintained we could point at

qimcis:
> > Not sure if this is the best way to be adding it to the dependencies list, Muon isn't on PyPi yet as far as I'm aware. If there's a better way for us to be doing this, let me know!! Perhaps I could move it to an extra instead?
> 
> An extra might be better. We aren't opposed to this but it would be preferred if there was a version on pypi that was going to be maintained we could point at

I added Muon as an extra, let me know if there's anything else I should change! @loadams

setup.py
```diff
@@ -92,6 +92,7 @@ def get_env_if_set(key, default: typing.Any = ""):
     'sd': fetch_requirements('requirements/requirements-sd.txt'),
     'triton': fetch_requirements('requirements/requirements-triton.txt'),
     'deepcompile': fetch_requirements('requirements/requirements-deepcompile.txt'),
+    'muon': ['muon-optimizer @ git+https://github.com/KellerJordan/Muon.git'],
```
tjruwase, line 95: @qimcis, why not add this to `requirements.txt` or `requirements-dev.txt`?

qimcis: @tjruwase I originally had it in `requirements.txt`, but moved it to an extra in `setup.py` following @loadams' suggestion since Muon isn't on PyPI yet. Happy to adjust back if necessary though, let me know what the consensus is!

sfc-gh-truwase: Apologies, I missed your interaction with @loadams. This looks fine.

qimcis:
@loadams @tjruwase Thank you for taking the time to review, and let me know if there are any other changes I should make! Would love to get this merged whenever possible

PKUWZP:
@qimcis  Thanks for your contribution. However, there're several issues that I think require substantial amount of work to resolve: 

1. With stage 1, 2 or 3, the optimizer states will be partitioned within the same data parallel group. This means that each process is already handling only parts of the model parameters and there is no need to use the DP solution as in the [code](https://github.com/KellerJordan/Muon/blob/master/muon.py#L195).
2. In this current implementations, it is using get_rank() and get_world_size(), however, we should use dist.get_rank(group=self.optimizer.dp_process_group) and dist.get_world_size(self.optimizer.dp_process_group).
3. Be careful of flattening and raw parameters, also be careful of padded parameters. This is because the parameters (and the gradients) will be flattened to 1D vector before being used in the optimizer. 

We will publish a new PR and acknowledge your contribution. Thanks so much for starting this effort!

PR #7457, opened 2025-07-29
Fix invalid f-strings

cyyever:
Fix invalid f-strings detected by ruff.


loadams:
@cyyever - could you please run the formatter?  Also is it worth adding ruff to the formatter?

cyyever:
@loadams I have formatted the files. I think it is worth adding ruff. However, to reduce the impact of refactoring we can apply the changes gradually.

loadams:
cc: @sfc-gh-mwyatt

sfc-gh-truwase (approved):

cyyever:
The test failures are unrelated.

sfc-gh-mwyatt:
FYI @loadams we probably don't want to add `ruff` to the formatting unless we have very good reason to. It could result in very big changes to the code base. Right now we are using `yapf` and `flake8` to do the job that `ruff` would do. The reason this error is not caught by `flake8` is because we have an exception for `F541`: https://github.com/deepspeedai/DeepSpeed/blob/8aadf6cbe42f9249c51f691e441a37ed5bfb7aa5/.flake8#L2. I would remove this before switching to a different linter/formatter.

PR #7474, opened 2025-08-08
add --bind_cores_to_rank to zero offload tutorial

delock:
In ZeRO offload, significant time is spent on CPUAdam, which is CPU code.  Thus use `--bind_cores_to_rank` in deepspeed launch command would help improve the performance of ZeRO offload.  This PR add this command to ZeRO offload tutorial to increase user awareness.

For Qwen2.5-3B finetuning on 2 A100-40B cards, running on CPU host with 128 CPU cores, the average step time is as follow, near 1.3x performance improvement:
without `--bind_cores_to_rank`: 3084.44ms per step
with `--bind_cores_to_rank`: 2383.16ms per step

### DIFF
--- a/docs/_tutorials/zero-offload.md
+++ b/docs/_tutorials/zero-offload.md
@@ -74,4 +74,11 @@ Finally, here is a screenshot of `htop` showing host CPU and memory activity dur
 <img src="/assets/images/zero_offload_dp1_10B_cpu.png">
 </a>
 
+### CPU Adam perf tuning
+ZeRO offload already support multi-gpu training.  If the workload is using CPU optimizer, the workload can be further tuened by pass `--bind_cores_to_rank` to the deepspeed launch command. This switch will mainly do two things:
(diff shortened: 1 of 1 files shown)


docs/_tutorials/zero-offload.md
```diff
@@ -74,4 +74,11 @@ Finally, here is a screenshot of `htop` showing host CPU and memory activity dur
 <img src="/assets/images/zero_offload_dp1_10B_cpu.png">
 </a>
 
+### CPU Adam perf tuning
+ZeRO offload already support multi-gpu training.  If the workload is using CPU optimizer, the workload can be further tuened by pass `--bind_cores_to_rank` to the deepspeed launch command. This switch will mainly do two things:
```
tjruwase:
```suggestion
ZeRO offload already support multi-gpu training.  If the workload is using CPU optimizer, the workload can be further tuned by passing `--bind_cores_to_rank` to the deepspeed launch command. This switch will mainly do two things:
```

PR #7489, opened 2025-08-15
DeepCompile ZeRO-3: robust allgather for uneven shards; fix profiling…

juyterman1000:
… meta key (max_mem)


sfc-gh-truwase:
@juyterman1000 can you please address the formatting issue using https://github.com/deepspeedai/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites

csrc/compile/z3.cpp
```diff
@@ -69,14 +69,68 @@ class Z3CustomOpExecutor : public CustomOpExecutor {
...
+            for (int i = 0; i < world_size; ++i) {
+                all_counts[i] = torch::empty_like(local_count_tensor);
+            }
+            process_group_->allgather(all_counts, local_count_tensor)->wait();
+
+            int64_t max_count = 0;
+            std::vector<int64_t> host_counts(world_size);
+            for (int i = 0; i < world_size; ++i) {
+                host_counts[i] = all_counts[i].to(torch::kCPU).item<int64_t>();
+                if (host_counts[i] > max_count) { max_count = host_counts[i]; }
```
eternalNight: Could you elaborate more on when `ds_tensor.numel()` of the same paramter can differ on different ranks? I think padding is already taken into account when the parameter is partitioned among the ranks (ref: https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/runtime/zero/partition_parameters.py#L1664)

In case partition sizes do vary across ranks, can we fix that in `partition_parameters.py` to avoid synchronous communication here? `launchAllGather()` is on the critical path, so synchronous allgather can hurt performance.

juyterman1000: Thanks for the sharp catch., I’ve removed the synchronous size‑allgather from the hot path in launchAllGather() and now use a fixed‑count NCCL allgather, trimming any end padding to the true param size. To keep things better without paying the runtime cost, I added a one‑time registration‑time assertion that shard sizes match across ranks, if there’s ever a mismatch, we’ll catch it at source rather than synchronize in the critical path. Changes are in the updated PR.

juyterman1000:
> @juyterman1000 can you please address the formatting issue using https://github.com/deepspeedai/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites

Thanks for the check. I followed the formatting prerequisites in the contributing guide and ran the full pre-commit suite. I’ve pushed the updates.

csrc/compile/z3.cpp
eternalNight: I think we can further optimize the code by making `allgatherParam()` allocating a buffer **with padding** in the first place (today it allocates a buffer of `ds_shape` which is the true size of the gathered parameter). With that we don't need any additional memcpy or GPU memory allocation/deallocation. Instead we can slice the gathered `output_buf` before returning it. My understanding is that torch can correctly track the refcount to the underlying buffers even living tensors use only part of them, but correct me if I'm wrong.

sfc-gh-truwase: @eternalNight thanks for the suggestion. @juyterman1000 if you agree with this, do you want to address in a follow up PR? A benefit of a follow up PR is that it could document the perf benefit of the optimization separately from functionality.

tohtana: Hi @juyterman1000,

Thank you for the PR! As some changes are unclear to me, can you explain a bit more?
You now added an assertion to ensure the even sharding, which totally makes sense to me. Do we still need the changes `launchAllGather`? The additional memory allocation and copy might cause the significant overhead in some cases.

juyterman1000: @eternalNight Yes, we can allocate a buffer sized to world_size * shard_elems up front and slice it to the true size on return. PyTorch views hold a reference to the underlying storage; returning a sliced view does not break refcounting. We can cache the padded buffer per param to avoid repeat allocations. @sfc-gh-truwase . Agreed on the follow up.I’ll include micro-benchmarks showing the removal of 1 alloc + 1 memcpy per all-gather and any other gains. @tohtana With the even-sharding assertion in place, we don’t need the extra copy logic in launchAllGather(). We can issue a direct AllGather with a uniform shard ,elems count into the padded buffer and return a view of the first true_numel elements reshaped to the original param shape. The symmetric-memory path will stay as-is. This  is to  avoid additional copy overhead.

tohtana: @juyterman1000 The path using symmetric memory is experimental and not well optimized. So we need to keep non-symmetric memory path as the choice for the best performance.
If the allocation and copy are for uneven partitioning and the assertion block such an uneven partitioning, why can't we remove them?

csrc/compile/z3.cpp
```diff
@@ -69,9 +69,14 @@ class Z3CustomOpExecutor : public CustomOpExecutor {
         const at::Tensor& ds_tensor = param.getDSTensor();
 
         if (symm_mem == nullptr) {
-            ncclResult_t result = ncclAllGather(ds_tensor.contiguous().data_ptr(),
+            // Fast path: assume uniform shard sizes (ZeRO-3 partitions are padded to uniform size)
+            const int world_size = process_group_->getSize();
+            const int64_t shard_elems = ds_tensor.numel();
+
+            // Perform all-gather directly into the pre-allocated padded output buffer
+            ncclResult_t result = ncclAllGather(ds_tensor.flatten().data_ptr(),
```
eternalNight: Why replacing `.contiguous()` with `.flatten()`? `.contiguous()` makes sure that the underlying storage is contiguous which is required by nccl. `.flatten()` is a view-change and does not guarantee that.

Note: I believe the sharded tensors are already contiguous as they are already defragmented by DeepSpeedZeroOptimizer_Stage3.defragment(), but adding a `.contiguous()` does not hurt anyway and may help later when the layout of sharded tensors is changed.

csrc/compile/z3.cpp
```diff
@@ -104,13 +109,30 @@ class Z3CustomOpExecutor : public CustomOpExecutor {
...
+        if (param_registry_->isValid(ds_id)) {
+            // Return a view sliced to the true size with the original shape
+            auto base = param_registry_->getGatheredParam(ds_id);
+            return base.flatten()
+                .index({torch::indexing::Slice(0, true_numel)})
+                .view(param.getShape());
+        }
+
+        at::Tensor output_buf;
+        if (param_registry_->hasGatheredParam(ds_id)) {
```
eternalNight, line 129: I'm not sure when `isValid(ds_id)` is false while `hasGatheredParam(ds_id)` is true. They are both set at the end of `launchAllGather()`, and releasing a gathered param will unset the valid flag in `unregisterGatheredParam()`.

csrc/compile/z3.cpp
juyterman1000: @tohtana  Yep. The non-symmetric memory path is indeed the performance path, and we can remove the extra copy logic since the uniform shard assertion prevents uneven partitioning.

csrc/compile/z3.cpp
juyterman1000: yep its already contiguous due to DeepSpeedZeroOptimizer_Stage3.defragment().

eternalNight: The latest version looks good to me. Unfortunately resolving threads do not work ...

tohtana:
Hi @juyterman1000,

Thank you, the changes look almost good to me, but I encountered the compilation error.
```
/home/ray/default/DeepSpeed/deepspeed/ops/csrc/compile/z3.cpp:569:30: error: cannot convert ‘std::vector<at::Tensor>’ to ‘std::vector<std::vector<at::Tensor> >&’
  569 |     process_group->allgather(all_counts, local_count_tensor)->wait();
      |                              ^~~~~~~~~~
      |                              |
      |                              std::vector<at::Tensor>
```

I think just wrapping the arguements with `std::vector` should work. Can you also add a test in `TestDeepCompile` (`tests/unit/runtime/compile/test_compile_zero.py`)? You can reuse the existing test and set a different hidden size to make the shard size uneven.

tohtana:
Hi @juyterman1000 
I still see a compilation error. The first argument is a buffer to store the result of allgather. So you can't pass an rvalue.

```
/home/ray/default//DeepSpeed.pr7489/deepspeed/ops/csrc/compile/z3.cpp:431:26: error: cannot bind non-const lvalue reference of type ‘std::vector<std::vector<at::Tensor> >&’ to an rvalue of type ‘std::vector<std::vector<at::Tensor> >’
  431 |         ->allgather(std::vector<std::vector<at::Tensor>>{all_counts},
      |                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/ray/default/DeepSpeed.pr7489/deepspeed/ops/csrc/includes/deepcompile.h:21,
                 from /home/ray/default/DeepSpeed.pr7489/deepspeed/ops/csrc/compile/z3.h:6,
                 from /home/ray/default/DeepSpeed.pr7489/deepspeed/ops/csrc/compile/z3.cpp:6:
```

juyterman1000:
All changes are implemented.

tohtana:
@juyterman1000 
The existing tests in test_compile_zero.py::TestDeepCompile::test don't pass. They raise the new error that this PR adds:
```python
RuntimeError: ZeRO-3 registration error: non-uniform shard sizes detected across ranks. Please check parameter partitioning.
```

By the way, it seems #7509 broke z1/2 paths for DeepCompile, but you can still check it with Z3.

tohtana:
Hi @juyterman1000 @adalakoti90 

When I added this logging:
```cpp
    std::cout << "rank: " << process_group->getRank()
        << " input_tensors[0]: " << input_tensors[0].to(torch::kCPU).item<int64_t>()
        << " output_tensors[0][0]: " << output_tensors[0][0].to(torch::kCPU).item<int64_t>()
        << " output_tensors[1][0]: " << output_tensors[1][0].to(torch::kCPU).item<int64_t>()
        << std::endl;
```
The output was
```
rank: 1 input_tensors[0]: 50 output_tensors[0][0]: 9223372036854775807 output_tensors[1][0]: 50
rank: 0 input_tensors[0]: 50 output_tensors[0][0]: 9223372036854775807 output_tensors[1][0]: 50
```

I might have misunderstood the usage of the allgather API. Do you have any idea?

eternalNight:
> Hi @juyterman1000 @adalakoti90
> 
> When I added this logging:
> 
> ```c++
>     std::cout << "rank: " << process_group->getRank()
>         << " input_tensors[0]: " << input_tensors[0].to(torch::kCPU).item<int64_t>()
>         << " output_tensors[0][0]: " << output_tensors[0][0].to(torch::kCPU).item<int64_t>()
>         << " output_tensors[1][0]: " << output_tensors[1][0].to(torch::kCPU).item<int64_t>()
>         << std::endl;
> ```
> 
> The output was
> 
> ```
> rank: 1 input_tensors[0]: 50 output_tensors[0][0]: 9223372036854775807 output_tensors[1][0]: 50
> rank: 0 input_tensors[0]: 50 output_tensors[0][0]: 9223372036854775807 output_tensors[1][0]: 50
> ```
> 
> I might have misunderstood the usage of the allgather API. Do you have any idea?

`ProcessGroup::allgather` expects the first argument to be a single-element vector containing a vector of WORLD_SIZE tensors. See https://github.com/pytorch/pytorch/blob/main/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L4684 for more details.

So it should look like:

```
    std::vector<std::vector<at::Tensor>> output_tensors = {all_padded_counts};
    std::vector<at::Tensor> input_tensors = {local_padded_tensor};
    process_group->allgather(output_tensors, input_tensors)->wait();
```

tohtana:
Thank you, @eternalNight! This is super helpful.

@juyterman1000, Can you fix it? It would be great if you could check it with the unit test.

tohtana (approved):
Thank you for the fix, @adalakoti90!

Issue #7478, opened 2025-08-09
[REQUEST] Auto-Tuning CPU Core Binding for DeepSpeed&ZenFlow


Antlera:
## Description
Currently, DeepSpeed offers `--bind_cores_to_rank` and `--bind_core_list` flags to bind CPU cores, but these require explicit specification from the user. While core binding works, it is not fully automated and does not adapt dynamically to different NUMA node configurations or CPU layouts.

## Problem
The current core binding functionality requires users to manually specify which cores to bind to each worker. This approach does not provide flexibility in handling complex CPU layouts, particularly when dealing with non-sequential core IDs or varying NUMA node configurations.

## Proposed Discussion
We may explore the possibility of auto-tuning CPU core binding to find optimal strategies automatically. This would involve:

- Automatic Core Binding: Automatically determine the best core binding across NUMA nodes without requiring user input.

- Adaptation to Non-Sequential Core Layouts (due to logical cores): Account for irregularities in core numbering (e.g., non-sequential core IDs within a NUMA node) to ensure efficient binding.

- Maximizing Core Utilization: Ensure that the binding maximizes CPU utilization and minimizes memory contention across workers.

## Possible Context

Deepspeed CPU binding: https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/utils/numa.py#L117
ZenFlow CPU binding:
`zenflow_optimizer_process()` at #7391

Antlera:
@delock @sfc-gh-truwase @tohtana, I've moved the discussion to this issue, feel free to continue the conversation here.

delock:
I build a benchmark that test CPUAdam performance seperately.  The tensor size is to simulate Qwen2.5-3B model running on multiple cards.
```
import torch
import deepspeed
from deepspeed.ops.op_builder import CPUAdamBuilder
import deepspeed.comm as dist
import time
import argparse

parser = argparse.ArgumentParser(description="test cpu adam performance")
parser.add_argument("--local_rank", type=int, help="number of ranks")
parser.add_argument("--size", type=int, default=1542969344*2, help="number of ranks")
args = parser.parse_args()

deepspeed.init_distributed()
if args.local_rank == None:
    world_size = 1
else:
    world_size = dist.get_world_size()
adam = CPUAdamBuilder().load()
adam.create_adam(0,2e-05,0.9,0.999,1e-08,0.01,True,True)
size = args.size//world_size
pdata = torch.ones(size)
[...]

PR #7506, opened 2025-08-22
Autotune ZenFlow affinity

delock:
This PR address the following ZenFlow optimizer core binding issue.
https://github.com/deepspeedai/DeepSpeed/issues/7478

With this PR, ZenFlow optimizer worker would derive its core binding from deepspeed core binding mechanism.   The algorithm is as following:
1. Each DeepSpeed rank get its core binding by using DeepSpeed command line `--bind_cores_to_rank`, this command would assign each CPU physical cores to different workers
2. When spawing ZenFlow optimizer worker, DeepSpeed would split current CPU affinity list into two sublist: pt_affinity and zf_affinity
3. zf_affinity would be used to set affinity of ZenFlow optimizer worker.  pt_affinity would be used to set current pytorch process.
4. By default, one cores is reserved by each pytorch process, the rest is used by ZenFlow optimizer worker.  The number of cores reserved for pytorch process can be changed by ZenFlow config variable: `pt_reserved_cores`


deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
```diff
@@ -807,10 +793,16 @@ def start_optimizer_process(self):
         curr_rank = dist.get_rank()
         total_rank = dist.get_world_size()
 
+        current_process = psutil.Process()
+        current_affinity = current_process.cpu_affinity()
+        ds_num_cores = 2
```
Antlera: Hi, @delock, thanks for this great patch, I am working on validating this binding policy. Glad to see this binding policy achieve better results. However, why should we choose `ds_num_cores = 2`? I see some slow down from main process from my side.

delock: Hi @Antlera , in theory this number should be 1 since there is only one active thread in main process.   I have made this number configurable in latest patch.   How do I measure slow down from main process?   If there is a better value we can change the default value in the patch.

Antlera: Hi @delock , just to confirm — after the update, do we still need to bind cores explicitly first to get the benefit? My understanding is that the update mainly separates DS and ZF, but per-rank core binding is still handled by --bind_cores_to_rank, right?

Antlera: One concern on `--bind_cores_to_rank`: it currently relies on` numactl`, which usually requires **apt + sudo privileges** to install. Do we have an alternative way to handle core binding without depending on tools that need sudo?

Also, I noticed a potential issue: if `overlap_step` is enabled but `--bind_cores_to_rank` is not explicitly set, different ranks may end up competing for the same cores.

delock: Hi @Antlera , yes `numactl` is needed.  The reason `numactl` is chosen because `numactl` not only provide core binding but also provide NUMA binding.  Also it controls resource before DeepSpeed worker starts so its a more reliable way of setting affinity.

For the second question, I think `--bind_cores_to_rank` should be used when offloading is used.   The similiar issue is also true for zero offload.    Currently we address this by update zero offload tutorial (https://github.com/deepspeedai/DeepSpeed/blob/master/docs/_tutorials/zero-offload.md).   Maybe we can do the same with ZenFlow tutorial and also add this flag into ZenFlow examples.

Antlera: Hi, @delock Please see https://github.com/deepspeedai/DeepSpeed/issues/7478 for latest comments. We can add instruction, but to maximize usability it’s better to also include a fallback mode. For example, in Slurm environments there are many restrictions, so a soft fallback binding mechanism can improve robustness.

delock: Hi @Antlera  the latest change has a fallback path.  Which shard physical cores between ranks if workers are not already binded by `numactl`.

I also find a potential issue in both original zenflow core binding and this soft fallback solution.  When sharding cores for each rank to bind, the world_size and global rank is used, this means for multinode the sharding may not be correct.  To resolve this we need to get `local_rank` and `local_size`.   Current DeepSpeed comm has method to get `local_rank`, but no method to get `local_size`.   Core binding with numactl (the 'hard' one) does not have this issue.  So for multi-node `--bind_cores_to_rank` is the recommended method.

Anyway, the latest code should run well on single node.  Run workload without `--bind_cores_to_rank` should go to fallback path and an inspection of log will see the binding used.  Let me know if it works on your side.

deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
```diff
@@ -807,13 +796,42 @@ def start_optimizer_process(self):
         curr_rank = dist.get_rank()
         total_rank = dist.get_world_size()
 
+        current_process = psutil.Process()
+        current_affinity = current_process.cpu_affinity()
+        all_affinities = [
+            torch.zeros(len(current_affinity), dtype=torch.int32, device=get_accelerator().current_device_name())
```
sfc-gh-truwase: How about this to avoid hard-coding data types?
```suggestion
            torch.zeros(len(current_affinity), dtype=type(cpu_affinity[0]), device=get_accelerator().current_device_name())
```

deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
```diff
@@ -807,13 +796,42 @@ def start_optimizer_process(self):
...
+            for _ in range(total_rank)
+        ]
+        dist.all_gather(
+            all_affinities,
+            torch.tensor(current_affinity, dtype=torch.int32, device=get_accelerator().current_device_name()))
+        # When affinity across all ranks are the same, the workers are not binded.  Do a soft bind here
+        if self.all_tensors_equal(all_affinities):
+            print(f"Recompute worker affinities")
```
sfc-gh-truwase: Disable debug prints

deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
```diff
@@ -807,13 +796,42 @@ def start_optimizer_process(self):
...
+            num_phy_cores = psutil.cpu_count(logical=False)
+            available_phy_cores = [i for i in current_affinity if i < num_phy_cores]
+            num_available_phy_cores = len(available_phy_cores)
+            my_rank = curr_rank
+            my_size = total_rank
+            cores_per_rank = num_available_phy_cores // my_size
+            current_affinity = available_phy_cores[my_rank * cores_per_rank:(my_rank + 1) * cores_per_rank]
+        ds_num_cores = self.pt_reserved_cores
```
sfc-gh-truwase: For clarity, is `DT cores == PT cores`?
```suggestion
        pt_num_cores = self.pt_reserved_cores
```

delock: yes, `pt_num_cores` is better name.

deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
delock: Will remove.

deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
delock: Thanks for the comments, will change.

sfc-gh-truwase:
@Antlera, is this PR good to merge or are there pending concerns? Thanks.

Antlera:
Hi @sfc-gh-truwase, please see #7478 for more benchmark and discussions. There is one more issue regarding CPU core distribution that still needs to be addressed. We’d appreciate comments from you. Thanks!

sfc-gh-truwase:
@Antlera I shared some feedback. I think auto-tuning can be addressed in a followup PR. But, it seems we need alignment on a default value for `pt_reserved_cores`. I think `1` is okay since it works in all cases.

deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
```diff
@@ -652,27 +646,13 @@ def disable_accelerator():
...
-        os.sched_setaffinity(0, assigned_cores)
-        print(f"[Optimizer Thread] Rank {curr_rank} bound to CPU cores: {os.sched_getaffinity(0)}", flush=True)
-    except AttributeError:
-        print("[Optimizer Thread] sched_setaffinity not supported on this system.")
-    except Exception as e:
-        print(f"[Optimizer Thread] Failed to set affinity: {e}")
+    current_process = psutil.Process()
+    current_process.cpu_affinity(zf_affinity)
+    os.environ['OMP_NUM_THREADS'] = str(len(zf_affinity))
+    #print(f"Setting zenflow optimizer affinity to {zf_affinity}, OMP_NUM_THREADS={len(zf_affinity)}")
```
Antlera: Please fix this debug output.

deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
```diff
@@ -807,13 +795,44 @@ def start_optimizer_process(self):
...
+            pt_affinity = current_affinity
         self.process = ctx.Process(
             target=zenflow_optimizer_process,
             args=(self.child_conn, curr_rank, total_rank, param_groups_data, self.shared_overlap_grad_map,
-                  self.shared_stale_param_map),
+                  self.shared_stale_param_map, zf_affinity),
         )
         self.process.daemon = True
         self.process.start()
+        #print(f"Setting pytorch affinity to {pt_affinity}, OMP_NUM_THREADS={len(pt_affinity)}")
```
Antlera: Please fix this debug output.

Antlera:
@sfc-gh-truwase Thanks for sharing the valuable thoughts. I agree auto-tuning can be done in a follow-up PR. On the default value, from my experiments reserving about half of the cores for PT usually gives better end-to-end performance, since it avoids fwd/bwd slowdown from CPU contention. A reasonable default could be `ceil(num_cores/2)` for PT and the rest for ZenFlow, which should work better than always fixing it to 1.

deepspeed/runtime/zenflow/zenflow_config.py
```diff
@@ -39,6 +39,10 @@ class ZenFlowConfig(DeepSpeedConfigModel):
     full_warm_up_rounds: int = 0
     """Number of initial rounds during which all gradients are fully updated (no selection)."""
 
+    pt_reserved_cores: int = 1
```
sfc-gh-truwase: @Antlera, so you are suggesting that this should be something like?

```suggestion
    pt_reserved_cores: int = ceil(num_cores/2)
```

deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
delock: Thanks!  These debug output are removed.

deepspeed/runtime/zenflow/zenflow_config.py
delock: If we wish to reserve half of the cores by default, a parameter called `pt_reserved_cores_perc` would be better choice.  The default value could be `0.5`.

delock:
Hi @Antlera are there remaining issues need to be resolved for this PR?  Thanks!

loadams (approved):

Antlera:
> Hi @Antlera are there remaining issues need to be resolved for this PR? Thanks!

Hi @delock the current code LGTM. Thanks for this great PR and discussion!

Issue #7438, opened 2025-07-19
[REQUEST] Support Muon Optimizer


hijkzzz:
Support Muon Optimizer in Kimi k2

PR #7509, opened 2025-08-23
Enabling Muon Optimizer in DeepSpeed

PKUWZP:
Authorship: @pengdurice and @PKUWZP 

Related Issue: #7438

# Introduction

[Muon](https://arxiv.org/abs/2502.16982), a new optimizer that has attracted the community’s attention recently shows promising results in training large language models. Adding the Muon Optimizer to DeepSpeed, a popular OSS framework for large scale training and inference is critically important for DeepSpeed users and developers. There has been a [PR](https://github.com/deepspeedai/DeepSpeed/pull/7454) attempting the adoption. (Huge Thanks to @qimcis), which is a good starting point. It still requires more substantial effort to make it fully compatible and work within DeepSpeed. We are publishing this PR to fully enable Muon Optimizer capabilities for DeepSpeed. 

# Issues and solutions
## Issues
1. With stage 1, 2 or 3, the optimizer states will be partitioned within the same data parallel group. This means that each process is already handling only parts of the model parameters and there is no need to use the DP solution as in the [code](https://github.com/KellerJordan/Muon/blob/master/muon.py#L195).
2. The parameters (and the gradients) will be flattened to 1D vector before being used in the optimizer, thus nullifying the major hypothesis of the muon optimizer: it works by orthogonalizing the updates for each matrix (dim >=2)

## Solutions
To solve the issues, we propose this new PR in which: 
1. We simplify the Muon code by [removing](https://github.com/deepspeedai/DeepSpeed/compare/master...pengdurice:DeepSpeed:peng-add-muon-v1#diff-c9052994e41caee9ca88363749c10af08655f8019f08dc971c018663d25a3712R22) the partitioning and muon updates logics.

2. We [move](https://github.com/deepspeedai/DeepSpeed/compare/master...pengdurice:DeepSpeed:peng-add-muon-v1#diff-99dcf26ea2876ff5bbf05b5165c4133eaa0d0f36b170685643c2f7e2eb566addR1867) the muon update to the [get_flat_partition](https://github.com/deepspeedai/DeepSpeed/compare/master...pengdurice:DeepSpeed:peng-add-muon-v1#diff-99dcf26ea2876ff5bbf05b5165c4133eaa0d0f36b170685643c2f7e2eb566addR1848) function of stage 1 and 2 DeepSpeedZeroOptimizer in which per parameter gradients are collected before being flattened and used by the optimizer to update the model parameters. Since each parameter is still in its original shape, we can easily apply the muon updates.
3. We also save the momentum buffer into the optimizer’ state so that we have a smooth convergence after applying the saved checkpoints.
4. We added comprehensive unit tests to validate Muon Optimizer's correctness and functionality. 

# Future directions and roadmap
In the future, several follow up works are of interests:
- [ ] Create a CPU offload version.
- [ ] Apply Muon to Stage 3
- [ ] Use the highly optimized version of Adam for the Adam part of MuonWithAuxAdam optimizer.
- [ ] More efficient implementations e.g. a) add specialized kernels for Newton-Schulz iteration and muon updates; b) parallelize updates for the parameters (currently, each parameter is updated separately and sequentially)


deepspeed/runtime/zero/muon/original_muon.py
```diff
@@ -0,0 +1,323 @@
+# Copyright (c) 2024 Keller Jordan
```
sfc-gh-truwase, line 1: Is this a direct copy of muon? If so, it would be better to reference the master via import to avoid issues of divergence. I believe the other PR added muon as dependency.

deepspeed/runtime/zero/muon/muon_optimizer.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2025 Peng Du and Zhipeng Wang
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+import torch
```
sfc-gh-truwase, line 6: Is this implemented just for ZeRO? What about running in non-ZeRO mode (a.k.a., ZeRO stage 0)?

deepspeed/runtime/zero/muon/original_muon.py
pengdurice: Hi @sfc-gh-truwase , thank you for reviewing this PR! Yes, it is a direct copy of muon.
The reason is:
1. I first tried to import muon by adding the dependency to the setup.py (in extras_require) identical the other PR: but it still complains muon not found;

>   File "/usr/local/lib/python3.9/dist-packages/deepspeed/runtime/muon_optimizer.py", line 9, in <module>
>     class MuonWithAuxAdam(BaseMuonWithAuxAdam):
> NameError: name 'BaseMuonWithAuxAdam' is not defined

2. I then added it to install_requires, it then complains:

> Collecting muon-optimizer@ git+https://github.com/KellerJordan/Muon.git
>   Cloning https://github.com/KellerJordan/Muon.git to /tmp/pip-install-3b4nvuts/muon-optimizer_18d1ee17b99344f399bf71dd6e3c2713
>   ERROR: Error [Errno 2] No such file or directory: 'git' while executing command git version
> ERROR: Cannot find command 'git' - do you have 'git' installed and in your PATH?

Adding to extras_require won't trigger the installation of the package, however adding to install_requires will trigger, but will require git to be installed in the CI / CD process of the github repo. So, we decided to just include a copy of the file to resolve this issue. Thanks!

tests/unit/ops/muon/test_muon.py
```diff
@@ -0,0 +1,84 @@
+# Copyright (c) 2025 Peng Du and Zhipeng Wang
```
sfc-gh-truwase, line 1: Is this UT a superset of the one in the original PR? Is there something that can be pulled in from there?

deepspeed/runtime/zero/muon/original_muon.py
sfc-gh-truwase: @pengdurice thanks for the explanation. I guess this is fine for now, especially since it is unclear if the upstream will be maintained long-term or pushed into pypi.

sfc-gh-truwase: Let's keep an eye on it. Thanks!

deepspeed/runtime/zero/muon/muon_optimizer.py
PKUWZP: Yes, this PR implementation is just for ZeRO. My thinking is that we will submit another PR only tailored for enabling Muon optimizer for non-ZeRO mode.

tests/unit/ops/muon/test_muon.py
PKUWZP: Thanks for the questions Tunji! The test cases in our file contain some of the logics which are similar to the original PR, but the design and coverage are very different. Let us know your suggestions so we can see what acknowledgement or changes we need to make.

sfc-gh-truwase: Thanks for confirmation. LGTM!

sfc-gh-truwase (approved):

deepspeed/runtime/zero/muon/original_muon.py
pengdurice: @sfc-gh-truwase , sure, will keep an eye on it! Thanks!

delock:
@PKUWZP is MuonClip (Used by Kimi K2 https://arxiv.org/abs/2507.20534) also of future interest?   From the paper clipping is essential to avoid exploding attention logits if the model is very large.

harshitkd:
How can I use Muon for stage 3? As I see it, it has not yet been applied to stage 3.

pengdurice:
> How can I use Muon for stage 3? As I see it, it has not yet been applied to stage 3.

Hi, team is working on enabling on stage 3. Stay tuned. Thanks!

PR #7515, opened 2025-08-25
Enable non-ZeRO mode

sfc-gh-truwase:
Enabled via `stage=0` which corresponds to DDP. 
Remove hardwired path to b16_optimizer.
Enable`torch.autocast` for DDP training
Enable native mixed precision DDP for bfloat16
Update torch.autocast and native mixed precision UTs

image

### DIFF
--- a/deepspeed/runtime/fp16/fused_optimizer.py
+++ b/deepspeed/runtime/fp16/fused_optimizer.py
@@ -211,8 +211,8 @@ class FP16_Optimizer(DeepSpeedOptimizer):
         return self.optimizer.param_groups[0]["lr"]
 
     def override_loss_scale(self, loss_scale):
-        if not self.use_grad_scaling:
-            return
+        assert self.use_grad_scaling, f"Loss scale overriding only supported for torch.float16, rather than {self.low_precision_dtype}"
+
         if loss_scale != self.external_loss_scale:
             logger.info(f'[deepspeed] setting loss scale from {self.external_loss_scale} -> {loss_scale}')
         self.custom_loss_scaler = True
@@ -399,7 +399,6 @@ class FP16_Optimizer(DeepSpeedOptimizer):
         3. scaled_loss.backward(), which accumulates scaled gradients into the ``.grad`` attributes of the model's fp16 leaves
         """
         if self.custom_loss_scaler:
-            assert self.use_grad_scaling
             scaled_loss = self.external_loss_scale * loss
             scaled_loss.backward()
         else:


sfc-gh-truwase:
@tohtana I have extended your `torch.autocast` PR to non-ZeRO case and also added some docs. Can you please review?

sfc-gh-truwase:
@stas00 FYI.

PKUWZP:
@sfc-gh-truwase Can you check why the cpu-torch unit tests are failing? Also can we say for non-ZeRO training, we fully rely on `torch.autocast `?

Antlera:
The `cpu-torch` CI fails due to the strict check in `engine._do_optimizer_sanity_check`:

```python
elif model_dtype == grad_accum_dtype:
    if model_dtype == torch.bfloat16:
        if self.pipeline_parallelism:
            logger.warning(...)
            return BFLOAT16
        else:
            raise NotImplementedError(...)
```

It only allows BF16 accumulation when PP is enabled.
Suggest relaxing this to a warning for non-ZeRO/non-PP cases instead of raising `NotImplementedError`.

sfc-gh-truwase:
@Antlera and @PKUWZP I found the cause, but a bit delayed in pushing a fix. Thanks for reviewing.

sfc-gh-truwase:
> Also can we say for non-ZeRO training, we fully rely on `torch.autocast `?

@PKUWZP, great question. No, non-ZeRO can also use native mixed-precision training. Below is how this PR expands the mixed precision training options of DeepSpeed: 

image

deepspeed/runtime/fp16/fused_optimizer.py
```diff
@@ -388,6 +399,7 @@ def backward(self, loss, create_graph=False, retain_graph=False):
         3. scaled_loss.backward(), which accumulates scaled gradients into the ``.grad`` attributes of the model's fp16 leaves
         """
         if self.custom_loss_scaler:
+            assert self.use_grad_scaling
```
stas00: adding some message to assert?

docs/code-docs/source/training.rst
```diff
@@ -33,6 +33,46 @@ Gradient Accumulation
 .. autofunction:: deepspeed.DeepSpeedEngine.is_gradient_accumulation_boundary
 
 
+Mixed Precision Training
+-------------------------
+DeepSpeed supports mixed precision training using either native or Torch mechanisms. The desired mixed precision mode can be selected through the configuration dict.
```
stas00:
```suggestion
DeepSpeed supports mixed precision training using either native or PyTorch mechanisms. The desired mixed precision mode can be selected through the configuration dict.
```

docs/code-docs/source/training.rst
```diff
@@ -33,6 +33,46 @@ Gradient Accumulation
...
+Mixed precision training can used with ZeRO (i.e., stages > 0) and without ZeRO (i.e., stage=0).
+
+
+Native Mixed Precision
+======================================================
+DeepSpeed provides native support for
+`fp16 <https://www.deepspeed.ai/docs/config-json/#fp16-training-options>`_ and `bf16 <https://www.deepspeed.ai/docs/config-json/#bfloat16-training-options>`_ mixed precsion training.
+
+
+Torch Automatic Mixed Precision (AMP)
```
stas00:
```suggestion
PyTorch Automatic Mixed Precision (AMP)
```

docs/code-docs/source/training.rst
```diff
@@ -33,6 +33,46 @@ Gradient Accumulation
...
+                "enabled": true,
+                "dtype": "bfloat16",
+                "lower_precision_safe_modules": ["torch.nn.Linear", "torch.nn.Conv2d"]
+            },
+            ...
+        }
+
+Each configuration works as follows:
+
+* ``enabled``: Enable torch.autocast when set to True. You don't need to call torch.autocast in your code. The grad scaler is also applied in the DeepSpeed optimizer.
```
stas00:
```suggestion
* ``enabled``: Enable ``torch.autocast`` when set to ``True``. You don't need to call ``torch.autocast`` in your code. The grad scaler is also applied in the DeepSpeed optimizer.
```

docs/code-docs/source/training.rst
```diff
@@ -33,6 +33,46 @@ Gradient Accumulation
...
+* ``dtype``: Lower precision dtype passed to torch.autocast. Gradients for allreduce (reduce-scatter) and parameters for allgather (only for ZeRO3) of lower_precision_safe_modules are also downcasted to this dtype.
```
stas00:
```suggestion
* ``dtype``: Lower precision dtype passed to ``torch.autocast``. Gradients for all-reduce (reduce-scatter) and parameters for all-gather (only for ZeRO3) of ``lower_precision_safe_modules`` are also downcasted to this ``dtype``.
```

docs/code-docs/source/training.rst
```diff
@@ -33,6 +33,46 @@ Gradient Accumulation
...
+* ``lower_precision_safe_modules``: The list of modules that will be downcasted for allreduce (reduce-scatter) and allgather (ZeRO3). The precision for PyTorch operators in forward/backward follows torch.autocast's policy, not this list. If you don't set this item, DeepSpeed uses the default list: ``[torch.nn.Linear, torch.nn.Conv1d, torch.nn.Conv2d, torch.nn.Conv3d]``.
```
stas00:
```suggestion
* ``lower_precision_safe_modules``: The list of modules that will be downcasted for all-reduce (reduce-scatter) and all-gather (ZeRO3). The precision for PyTorch operators in forward/backward follows ``torch.autocast``'s policy, not this list. If you don't set this item, DeepSpeed uses the default list: ``[torch.nn.Linear, torch.nn.Conv1d, torch.nn.Conv2d, torch.nn.Conv3d]``.
```

tests/unit/runtime/test_ds_initialize.py
```diff
@@ -229,7 +229,7 @@ def test(self, optimizer_extension, model_dtype, grad_accum_dtype):
         is_supported[(None, 'fp16', None)] = True
         is_supported[(None, 'fp16', 'fp16')] = True
         # BF16 Wrapper
-        is_supported[(None, 'bf16', 'fp32')] = True
+        is_supported[(None, 'bf16', 'bf16')] = True
```
stas00, line 232: so `fp32` is no longer needed?

sfc-gh-truwase: This is was previously incorrectly routed to bf16_optimizer. But now is adjusted  for `zero stage == 0`. The bf16_optimizer is currently only available for PP.

PR #7525, opened 2025-08-29
ZeRO3: Improve mismatch detection

sfc-gh-truwase:
ZeRO3 tracks DDP (SPMD) behavior by matching values different training states across ranks. Some of these states are represented as lists, and mismatches sometimes manifests as hangs during error detection.  This PR improves error detection by first validating the list lengths across ranks before validating the list contents. 

Motivated by https://github.com/deepspeedai/DeepSpeed/issues/7461#issuecomment-3235146207

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1059,7 +1059,7 @@ class DeepSpeedEngine(Module):
     def zero_log_trace_cache_warnings(self):
         return self._config.zero_config.log_trace_cache_warnings
 
-    def enable_sanity_checks(self):
+    def is_sanity_checks_enabled(self):
         return self._config.zero_config.enable_sanity_checks
 
     def dump_state(self):
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1862,7 +1862,7 @@ class DeepSpeedEngine(Module):
                     zero_module_granularity_threshold=self.zero_module_granularity_threshold(),
                     zeropp_loco_param=self.zeropp_loco_param(),
                     log_trace_cache_warnings=self.zero_log_trace_cache_warnings(),
-                    enable_sanity_checks=self.enable_sanity_checks(),
+                    enable_sanity_checks=self.is_sanity_checks_enabled(),
                 )
 
         else:
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1059,6 +1059,9 @@ class DeepSpeedEngine(Module):
     def zero_log_trace_cache_warnings(self):
         return self._config.zero_config.log_trace_cache_warnings
 
+    def enable_sanity_checks(self):
+        return self._config.zero_config.enable_sanity_checks
+
     def dump_state(self):
         return self._config.dump_state
 
@@ -1859,6 +1862,7 @@ class DeepSpeedEngine(Module):
                     zero_module_granularity_threshold=self.zero_module_granularity_threshold(),
                     zeropp_loco_param=self.zeropp_loco_param(),
                     log_trace_cache_warnings=self.zero_log_trace_cache_warnings(),
+                    enable_sanity_checks=self.enable_sanity_checks(),
                 )
 
         else:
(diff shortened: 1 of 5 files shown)


deepspeed/runtime/zero/config.py
```diff
@@ -46,7 +46,8 @@
     "memory_efficient_linear": [true|false],
     "override_module_apply": [true|false],
     "zeropp_loco_param": {...},
-    "log_trace_cache_warnings" : [true|false],
+    ""log_trace_cache_warnings" : [true|false],
```
tohtana: You have two " though it is harmless

deepspeed/runtime/zero/utils.py
```diff
@@ -67,6 +67,22 @@ def is_zero_supported_optimizer(optimizer):
...
 
+@instrument_w_nvtx
+def assert_lst_len_same_as_other_ranks(lst: List[int]) -> None:
+    rank0_len_tensor = torch.tensor(
+        len(lst) if dist.get_rank() == 0 else -1,
+        dtype=int,
+        device=torch.device(get_accelerator().device_name(os.environ["LOCAL_RANK"])),
+        requires_grad=False,
+    )
+    local_len_tensor = len(lst)
```
tohtana: Can we simplify it a bit by using item()?

tohtana: Can we also verify ds_id (and shape, dtype)? We often have tensors with the same shapes. But it is also good to start with verifying only length.

sfc-gh-truwase: I am not sure I fully understand the ask. We already have verification calls such as [this](https://github.com/deepspeedai/DeepSpeed/blob/9bf215d21331ec92e8f761b3f1a62364c3c8fc82/deepspeed/runtime/zero/stage3.py#L1326). This PR is only augmenting existing verifications with a length verification. 

Can you please clarify further if I am missing your point?

tohtana: Oh you are right. Sorry I have missed it

deepspeed/runtime/engine.py
```diff
@@ -1059,6 +1059,9 @@ def zeropp_loco_param(self):
     def zero_log_trace_cache_warnings(self):
         return self._config.zero_config.log_trace_cache_warnings
 
+    def enable_sanity_checks(self):
+        return self._config.zero_config.enable_sanity_checks
```
stas00: this feels misnamed, since it doesn't enable anything, perhaps:

```suggestion
    def is_sanity_checks_enabled(self):
        return self._config.zero_config.enable_sanity_checks
```

deepspeed/runtime/zero/utils.py
```diff
@@ -67,6 +67,23 @@ def is_zero_supported_optimizer(optimizer):
...
+    dist.broadcast(rank0_len_tensor, src=0, async_op=False)
+    rank0_len_tensor_cpu = rank0_len_tensor.cpu().item()
+    if rank0_len_tensor_cpu != local_len_tensor:
+        raise RuntimeError(f"disagreement between rank0 and rank{dist.get_rank()}: "
```
stas00:
```suggestion
        raise RuntimeError(f"Detected a disagreement between rank0 and rank{dist.get_rank()}: "
```

deepspeed/runtime/zero/utils.py
```diff
@@ -67,6 +67,23 @@ def is_zero_supported_optimizer(optimizer):
...
+                           f"\n rank0: {rank0_len_tensor_cpu} "
+                           f"\n rank{dist.get_rank()}: {local_len_tensor}")
```
stas00: should we add `len(rank0_len_tensor_cpu)` dump as well if the first mismatch is of length?

deepspeed/runtime/zero/utils.py
stas00: and the same in the other instance of this message - line 97 in the original, 115 in this PR

deepspeed/runtime/engine.py
stas00: and the corresponding calls of course if accepted.

PR #7526, opened 2025-08-29
[logging] less startup noise

stas00:


deepspeed/launcher/launch.py
```diff
@@ -102,6 +103,8 @@ def parse_args():
                         "numbers and range. i.e. 1,3-5,7 => [1,3,4,5,7].  When not "
                         "specified, all cores on system would be used rank binding")
 
+    parser.add_argument("-q", "--quiet", action="store_true", help="Turns launcher logging off")
```
sfc-gh-truwase: How about inverting this, so that default is quiet?

sfc-gh-truwase: That will allow us to provide different degrees of verbosity over time.

stas00: I'm concerned this would break BC. e.g. what if some framework parses these logs and makes use of them? e.g. some tests are likely to do that

But we could plan this for v1? In which case we could replace `-q` with `--verbose` and set it to to `True` for now in this PR?

Alternatively we could create `--log-level` and set it to `INFO` for now and change to `WARNING` for v1?

And perhaps still keep `-q` as a shortcut for `--log-level WARNING`? or even `--log-level ERROR`? Probably `--log-level WARNING` is a sensible `-q` equivalent if warnings are important, albeit it might not be fully quiet?

sfc-gh-truwase: Good call on BC. Yes, I agree with deferring my suggestion until v1. I prefer `--log-level` option for v1, but also fine with merging this PR as is. Thanks!

stas00: ok, let me add `--log-level` first, default it for `INFO` for now

Should we start preparing for v1 default behavior changes using say `TODOV1` comments? e.g. here `TODOV1: change to WARNING`

sfc-gh-truwase: That is a great idea. Let's do it.

stas00: I can't decide whether to alias `-q` to `info.ERROR` or `info.WARNING` - what do you think Tunji?

I think `-q` can probably be used to silence other things down the road - e.g. python (non-logger) warnings as well, so it could be more than just logger silencer.

stas00: ok, added `--log_level` and used `-q` as an alias. Please kindly review this version, Tunji.

It's a bit wonky still since the `real_accelerator` path has no cmd line options - so can't change the logging level there, had to change the default to `logging.WARNING` there. We probably have to redesign the logging setup as it's quite a mess/duplication right now.

deepspeed/launcher/runner.py
```diff
@@ -553,6 +569,10 @@ def main(args=None):
             deepspeed_launch.append("--bind_cores_to_rank")
         if args.bind_core_list is not None:
             deepspeed_launch.append(f"--bind_core_list={args.bind_core_list}")
+        if args.quiet:
+            deepspeed_launch.append("--quiet")
```
sfc-gh-truwase, line 573: Won't this add  both `--quiet` and `--log_level=`  to command line?

deepspeed/launcher/launch.py
sfc-gh-truwase: I think this is good for now. I agree a some rethinking is needed. 

My initial thoughts is that there are two paths to control logging behavior
1. Command-line for launcher
2. ds_config for DS engine

However, the coverage and interaction of these two paths are still unclear to me.

deepspeed/launcher/runner.py
stas00: it's by design. The 2 are related but aren't necessarily the same. for example we might change the `-q` behavior to, say, also silence python warnings.

sfc-gh-truwase: LGTM. Please merge at your convenience.

deepspeed/launcher/launch.py
stas00: plus builder - when you import anything from deepspeed the builder gets loaded and it logs things.

proof:

```
  File "/home/yak/miniconda3/envs/dev/bin/arctic_training", line 4, in <module>
    from arctic_training_cli import main
  File "/code/users/stas/github/sf/arctictraining/arctic_training_cli.py", line 22, in <module>
    from deepspeed.launcher.runner import main as ds_runner
  File "/code/users/stas/github/DeepSpeed/deepspeed/__init__.py", line 25, in <module>
    from . import ops
  File "/code/users/stas/github/DeepSpeed/deepspeed/ops/__init__.py", line 6, in <module>
    from . import adam
  File "/code/users/stas/github/DeepSpeed/deepspeed/ops/adam/__init__.py", line 6, in <module>
    from .cpu_adam import DeepSpeedCPUAdam
  File "/code/users/stas/github/DeepSpeed/deepspeed/ops/adam/cpu_adam.py", line 8, in <module>
    from deepspeed.utils import logger
  File "/code/users/stas/github/DeepSpeed/deepspeed/utils/__init__.py", line 10, in <module>
    from .groups import *
  File "/code/users/stas/github/DeepSpeed/deepspeed/utils/groups.py", line 28, in <module>
  File "/code/users/stas/github/DeepSpeed/deepspeed/comm/__init__.py", line 7, in <module>
    from .comm import *
  File "/code/users/stas/github/DeepSpeed/deepspeed/comm/comm.py", line 32, in <module>
    from deepspeed.comm.ccl import CCLBackend
  File "/code/users/stas/github/DeepSpeed/deepspeed/comm/ccl.py", line 11, in <module>
    from deepspeed.ops.op_builder import NotImplementedBuilder
  File "/code/users/stas/github/DeepSpeed/deepspeed/ops/op_builder/__init__.py", line 53, in <module>
```

stas00: I wonder if we should clean up `deepspeed.utils.__init__` as you can see it shouldn't load all that stuff if we are just importing the logger.

stas00: probably should leave `logger` out of `deepspeed.utils.__init__` and use a direct`from deepspeed.utils.logging import logger`

stas00: @sfc-gh-truwase, should we start v1 doc somewhere and gather notes there so that we don't have to repeat these discussions?

sfc-gh-truwase: Yes, we should. Should we create as a GH ticket?

stas00: That's a good idea, I will create one

stas00: https://github.com/deepspeedai/DeepSpeed/issues/7540

jue-jue-zi:
How to change the default log level when using `accelerate` for multi-node training? It's useful to check if `accelerate` sets the right config for `deepspeed`.

PR #7534, opened 2025-09-02
Fix scaling and allgather with `torch.autocast`

tohtana:
This PR includes these two fixes:
- Use GradScaler only for FP16 (not for BF16)
- Fix dtype conversion for ZeRO3 allgather
  - The reduce hook should be called only once, even when a parameter is shared across multiple layers (tied parameters).
  - Currently, the hook is triggered at each tied layer because we temporarily set `.data` with a different dtype.
  - The fix ensures that the parameter consistently retains the same dtype.


deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -1334,8 +1333,28 @@ def all_gather_coalesced(params: Iterable[Parameter],
...
+                        # However, when the dtype changes, the hook may be triggered multiple times.
+                        # If we directly do:
+                        #   param_buffer.narrow(0, 0, param.ds_numel).view(param.ds_shape).to(param.device)
+                        # as above, the dtype may differ, causing the gradient-reduce hook
+                        # to be invoked multiple times.
+                        # We cannot simply call `.to(original_dtype)` here either, because communication runs on
+                        # a different CUDA stream asynchronously.
+                        # To avoid this, we allocate a new buffer with the original dtype, even though it is less
+                        # memory-efficient. The dtype conversion will then be handled in `wait()` of AllGatherHandle.
+                        param.data = torch.empty_like(param_buffer, dtype=original_dtype)
```
sfc-gh-truwase: Is `param.data` actually used since it seems to be overwritten in line 695?

tohtana: @sfc-gh-truwase Thank you for the comment! I thought it was necessary, but removing this line (not setting `param_buffer`) actually worked. This significantly reduces the peak memory.
I wonder if we could consolidate the two cases (original_dtype ==/!= allgather_dtype), but let's leave the conditional branch for now as the change of the other branch can be impactful.

sfc-gh-truwase: LGTM. Thanks for addressing and reducing peak memory :)

PR #7537, opened 2025-09-02
[MoE] Fix misuse of num_experts as expert parallel group size (ep_size)

### LINKED ISSUE #7535: [BUG] InferenceEngine._create_ep_parallel_group uses `num_experts` instead of `ep_size`, causing incorrect behavior
In the inference engine, MoE expert parallel groups are created in `_create_ep_parallel_group`.  
However, the parameter `moe_experts` (aliased from `num_experts`) is incorrectly used as **group size (`ep_size`)**, instead of **the number of experts**.

This leads to wrong behavior when `num_experts > world_size`.

---

## Problematic Code

From `inference_engine.py`:

```python
if isinstance(self.module, torch.nn.Module):
    moe, _ = has_moe_layers(self.module)
else:
    moe = False

if moe and dist.get_world_size() > 1:
    self._create_ep_parallel_group(config.moe.moe_experts)

```

And the function:

```python
def _create_ep_parallel_group(self, moe_experts):
    # Call the init process
    self.ep_group = {}
    self.expert_mp_group = {}
    moe_experts = moe_experts if type(moe_experts) is list else [moe_experts]
    for e in moe_experts:
        self.ep_group.update({e: None})
        self.expert_mp_group.update({e: None})
    for moe_ep_size in self.ep_group.keys():
        num_ep_groups = dist.get_world_size() // moe_ep_size
        for i in range(num_ep_groups):
            ep_cnt = i * moe_ep_size
            size = dist.get_world_size() if moe_ep_size > dist.get_world_size() else moe_ep_size
            ranks = list(range(ep_cnt, ep_cnt + size))
            _ep_group = dist.new_group(ranks)
            if dist.get_rank() in ranks:
                self.ep_group.update({moe_ep_size: _ep_group})

        if dist.get_world_size() > moe_ep_size:
            num_expert_mp_groups = dist.get_world_size() // num_ep_groups
            expert_mp_size = dist.get_world_size() // moe_ep_size
            for i in range(num_expert_mp_groups):
                expert_mp_comm_ranks = [i + nr * moe_ep_size for nr in range(expert_mp_size)]
                _expert_mp_group = dist.new_group(expert_mp_comm_ranks)
                if dist.get_rank() in expert_mp_comm_ranks:
                    self.expert_mp_group.update({moe_ep_size: _expert_mp_group})
```

**Example:** world_size = 2, num_experts = 8

```python
num_ep_groups = dist.get_world_size() // moe_ep_size
              = 2 // 8
              = 0
```

**Suggested Fix:**
```python
self._create_ep_parallel_group(config.moe.ep_size)
```

Flakes342:
Fixes #7535 

## Description
This PR fixes a bug in inference/engine.py where num_experts (moe_experts) was incorrectly passed as the expert parallel group size (ep_size) when creating expert parallel groups.

Currently:
```
if moe and dist.get_world_size() > 1:
    self._create_ep_parallel_group(config.moe.moe_experts)
```
This causes **invalid** behavior whenever `num_experts > world_size`, because `_create_ep_parallel_group` expects a group size, not the total number of experts as pointed out by @Arnoochka 

## Root Cause

num_experts = number of experts inside the MoE layer.

ep_size = how many GPUs to group together for expert parallelism.

These were mixed up in the code.

##Fix

Replaced the incorrect call with the proper ep_size argument:
```
if moe and dist.get_world_size() > 1:
    self._create_ep_parallel_group(config.moe.ep_size)
```

Additionally, added a safety check in _create_ep_parallel_group to catch invalid configurations:

```
num_ep_groups = dist.get_world_size() // moe_ep_size
if num_ep_groups == 0:
    raise ValueError(
        f"Invalid ep_size={moe_ep_size} for world_size={dist.get_world_size()}"
    )
```
## Backward compatibility
- If a user was already running with ep_size >= num_experts, the old code worked fine which would still work fine.
- Only the previously broken case (num_experts > world_size) now works correctly.

### DIFF
--- a/docs/_tutorials/ulysses-alst-sequence-parallelism.md
+++ b/docs/_tutorials/ulysses-alst-sequence-parallelism.md
@@ -116,11 +116,11 @@ for iter, batch in enumerate(dl):
     # differentiable weighted per-shard-loss aggregation across ranks
     losses_per_rank = torch.distributed.nn.functional.all_gather(loss, group=sp_group)
     # special dealing with SFT that has prompt tokens that aren't used in loss computation
-    good_tokens = sum((shift_labels != -100).view(-1))
+    good_tokens = (shift_labels != -100).view(-1).sum()
     good_tokens_per_rank = torch.distributed.nn.functional.all_gather(good_tokens, group=sp_group)
     total_loss = sum(losses_per_rank[rank] * good_tokens_per_rank[rank] for rank in range(sp_world_size))
     total_good_tokens = sum(good_tokens_per_rank)
-    loss = total_loss / total_good_tokens
+    loss = total_loss / max(total_good_tokens, 1)
 
     if dist.get_rank() == 0:
         print(f"{iter}: {loss=}")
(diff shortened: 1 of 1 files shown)


tohtana:
@Flakes342 nv-mii raised an error, but it is not related to this PR. Let's merge once you fix the formatting and DCO.

Flakes342:
Hi @tohtana, have fixed formatting and given proper signoff for DCO too. Please see if this can be merged now. Thanks!

PR #7547, opened 2025-09-04
Improve error message and reduce validation in autocast test

tohtana:
This PR improves error logging and relaxes loss value checks in the autocast test.

Previously, the test displayed error messages and mismatched loss values on all ranks, even if the mismatch only occurred on some ranks. This was confusing, since logs from other ranks could appear correct. This PR changes the behavior so that error messages are shown only on the ranks where the mismatch occurs.

Additionally, this PR skips loss value validation for `test_lower_precision_model`, where we intentionally use a different communication dtype from the baseline (standard PyTorch autocast).

### DIFF
--- a/tests/unit/runtime/zero/test_zero_autocast.py
+++ b/tests/unit/runtime/zero/test_zero_autocast.py
@@ -61,8 +61,12 @@ def step_amp(enabled, baseline_model, baseline_optimizer, target_engine, dtype, 
     # reduce-scatter in `dtype` makes a difference in the loss.
     if step <= 1 and expect_match:
         allclose_local = torch.allclose(baseline_loss.float(), target_loss.float(), rtol=rtol, atol=atol)
+        if not allclose_local:
+            print(f"Losses do not match: baseline_loss={baseline_loss}, target_loss={target_loss}")
+        # Ensure all ranks either pass or fail together.
+        # If some rank fail while others pass, subsequent tests or iterations may hang.
         if not reduce_boolean_flags(allclose_local, all):
-            assert False, f"Losses do not match on one or more ranks: baseline_loss={baseline_loss}, target_loss={target_loss}"
+            assert False, f"Losses do not match on one or more ranks."
 
     target_engine.backward(target_loss)
     target_engine.step()
--- a/tests/unit/runtime/zero/test_zero_autocast.py
+++ b/tests/unit/runtime/zero/test_zero_autocast.py
@@ -163,8 +163,8 @@ def compare_loss(model_cls,
 class TestZeroAutoCast(DistributedTest):
     world_size = 2
 
-    @pytest.mark.parametrize("zero_stage", [3])
-    @pytest.mark.parametrize("dtype", [torch.bfloat16])
+    @pytest.mark.parametrize("zero_stage", [0, 1, 2, 3])
+    @pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float16])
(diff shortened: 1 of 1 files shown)


tests/unit/runtime/zero/test_zero_autocast.py
```diff
@@ -18,6 +18,7 @@
 from deepspeed.accelerator import get_accelerator
 from deepspeed.runtime.zero import GatheredParameters
 from deepspeed.runtime.torch_autocast import PARAM_COMM_DTYPE_ATTR_NAME, get_comm_dtype
+from deepspeed.utils import logger
```
stas00: I don't think `logger` is the right tool for tests since how do you even control the log levels. 

why not just `assert`? if the local loss doesn't match that's already a reason to fail.

stas00: Also see the helpers I added 
https://github.com/deepspeedai/DeepSpeed/blob/78a74874b234ad5e9bb7d2f8f6a774eb36368cf1/tests/unit/util.py#L97-L125

which require less code in many situations. Usage example: https://github.com/deepspeedai/DeepSpeed/blob/78a74874b234ad5e9bb7d2f8f6a774eb36368cf1/tests/unit/ulysses_alst/test_tiled_compute.py#L196-L197
you can add `msg="custom message"` to the call.

tohtana: Thank you for the feedback, @stas00!

I removed the logger's output. My intention was the message by `assert` might be a bit confusing (it will say `Losses do not match` even when it shows matching values on the rank), but it won't be a big issue. We might want to clarify a recommended way to display message in tests in the future.

> why not just assert? if the local loss doesn't match that's already a reason to fail.

The following test can get stack if only some (not all) ranks fail. It's safer to make all ranks fail. For this purpose, I want to get the return value (`torch.testing.assert_close` returns None) rather than throwing an exception.

The tests now pass in my environment, but we won't see a change on CI as we fetch the tests from master branch.

stas00: Let's add a comment then to why you wait till the end to assert? so the next person won't try to "improve" it?

I have just meant to use print for your message and not the logger, since logger isn't guaranteed to print if it was shut off.

stas00 (approved):
a small suggestion left earlier. Thank you for fixing this, @tohtana

tests/unit/runtime/zero/test_zero_autocast.py
tohtana: Thank you for the comment, @stas00.
I added a comment and a print to show the error message.

tests/unit/runtime/zero/test_zero_autocast.py
```diff
@@ -61,8 +61,12 @@ def step_amp(enabled, baseline_model, baseline_optimizer, target_engine, dtype,
     # reduce-scatter in `dtype` makes a difference in the loss.
     if step <= 1 and expect_match:
         allclose_local = torch.allclose(baseline_loss.float(), target_loss.float(), rtol=rtol, atol=atol)
+        if not allclose_local:
+            print(f"Losses do not match: baseline_loss={baseline_loss}, target_loss={target_loss}")
+        # Ensure all ranks either pass or fail together.
+        # If some rank fail while others pass, subsequent tests or iterations may hang.
```
stas00:
```suggestion
        # If some ranks fail while others pass, subsequent tests or iterations may hang.
```

PR #7548, opened 2025-09-09
Fix gradient buffer access for DeepCompile Z1/2

tohtana:
The initialization of DeepCompile+Z1/2 now fails due to the change introduced in #7509.

This PR resolves the issue by:
- Adding an argument to optimizer.get_flat_partition
- Skipping the entire allreduce function in the engine

### DIFF
--- a/deepspeed/compile/init_z1.py
+++ b/deepspeed/compile/init_z1.py
@@ -31,11 +31,13 @@ def init_z1(engine, backend, compile_config, compile_kwargs, schedule=None, use_
 
     grad_buffer = {}
 
+    # Save original all_grad_tensors state as we temporarily modify it
+    original_all_grad_tensors = optimizer.all_grad_tensors.copy() if hasattr(optimizer, 'all_grad_tensors') else {}
+
     for i, group in enumerate(optimizer.bit16_groups):
         # Temporarily populate all_grad_tensors for get_flat_partition call
         # This is needed because get_flat_partition accesses all_grad_tensors[param_group_idx][i]
         # but it's empty during initialization
-        original_all_grad_tensors = optimizer.all_grad_tensors.copy() if hasattr(optimizer, 'all_grad_tensors') else {}
         if i not in optimizer.all_grad_tensors or optimizer.all_grad_tensors[i] is None:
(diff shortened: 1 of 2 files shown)


deepspeed/compile/init_z1.py
```diff
@@ -32,15 +32,26 @@ def init_z1(engine, backend, compile_config, compile_kwargs, schedule=None, use_
     grad_buffer = {}
 
     for i, group in enumerate(optimizer.bit16_groups):
+        # Temporarily populate all_grad_tensors for get_flat_partition call
+        # This is needed because get_flat_partition accesses all_grad_tensors[param_group_idx][i]
+        # but it's empty during initialization
+        original_all_grad_tensors = optimizer.all_grad_tensors.copy() if hasattr(optimizer, 'all_grad_tensors') else {}
```
sfc-gh-truwase: Could `original_all_grad_tensors` not be initialized before the loop?

deepspeed/compile/init_z1.py
```diff
@@ -32,15 +32,26 @@ def init_z1(engine, backend, compile_config, compile_kwargs, schedule=None, use_
...
                                                       optimizer.first_offset[i],
                                                       optimizer.partition_size[i],
                                                       dtype=optimizer.gradient_accumulation_dtype,
                                                       device=get_accelerator().current_device_name(),
+                                                      param_group_idx=i,
                                                       return_tensor_list=True)
         grad_buffer[i] = [p.clone().detach() for p in grad_buffer[i]]  # Maybe not necessary
 
+        # Restore original all_grad_tensors state
+        optimizer.all_grad_tensors = original_all_grad_tensors
```
sfc-gh-truwase: Similar to earlier, can you please explain why this cannot be outside the loop?

deepspeed/compile/init_z1.py
tohtana: Yes, you are right! This should be outside the loop. I moved it. Thank you for the catch!

deepspeed/compile/init_z1.py
tohtana: I moved it as well.

sfc-gh-truwase (approved):

PR #7557, opened 2025-09-12
Move modal tests to tests/v1

tohtana:
This PR moves active tests under `tests/unit/v1` to clarify which tests are run on modal.

### DIFF
--- a/ci/torch_latest.py
+++ b/ci/torch_latest.py
@@ -32,23 +32,8 @@ app = modal.App("deepspeedai-torch-latest-ci", image=image)
 )
 def pytest():
     import subprocess
-    test_files = [
-        "tests/unit/v1/zero/test_zero.py",
-        "tests/unit/v1/half_precision/test_bf16.py",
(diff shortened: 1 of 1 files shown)


ci/torch_latest.py
```diff
@@ -32,8 +32,23 @@
...
+        "tests/unit/v1/half_precision/test_bf16.py",
+        "tests/unit/v1/zero/test_zero_autocast.py",
+        "tests/unit/v1/compile/test_compile_zero.py",
+    ]
+    command = [
+        "pytest",
+        "-n",
+        "4",
+        "--verbose",
+        *test_files,
```
sfc-gh-truwase: The original proposal is to specify the `tests/unit/v1` path instead of enumerating the individual files. Then we can incrementally add files to the folder without having to change `ci/torch_latest/py`. 

```suggestion
        "tests/unit/v1",
```

tohtana: @sfc-gh-truwase Oh okay, fixed it

PR #7558, opened 2025-09-13
Add dependency for deepcompile test

tohtana:
This PR adds dependency to CI tests for DeepCompile.


tohtana:
Force-merging as the change in this PR doesn't affect remaining tests.

PR #7561, opened 2025-09-15
scripts: Check .is_cuda only in non-C++ files

eternalNight:
The check-torchcuda.py today will search for all occurrences of .is_cuda in the repository when a commit only modifies C++ headers and sources, which I believe is not intended.

Check usage of .is_cuda only when a commit modifies any non-C++ file.

### DIFF
--- a/scripts/check-torchcuda.py
+++ b/scripts/check-torchcuda.py
@@ -57,23 +57,24 @@ elif res.returncode == 2:
 
 files = []
 for file in sys.argv[1:]:
-    if not file.endswith(".cpp"):
+    if file.endswith(".py"):
         files.append(file)
 
-res = subprocess.run(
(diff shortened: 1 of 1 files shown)


scripts/check-torchcuda.py
```diff
@@ -60,20 +60,21 @@ def err(s: str) -> None:
     if not file.endswith(".cpp"):
```
sfc-gh-truwase: @eternalNight what do you think about explicitly checking for python file modifications?

```suggestion
    if file.endswith(".py"):
```

eternalNight: Looks good to me. Thanks for the suggestion! Commit updated.

loadams (approved):

PR #7562, opened 2025-09-15
deepcompile: Create a full list of no-copy ops

eternalNight:
The list of torch no-copy ops is hard coded and does not include all operations that may aliasing inputs in their outputs.

Instead of using a fixed list, iterate over all ops under torch.ops.aten and identify those with aliasing behavior by inspecting their schema.

With PyTorch 2.7.1, the default overload of ops identified by the updated logic include:

  - _nested_view_from_buffer
  - _reshape_alias
  - alias
  - as_strided
  - conj
  - detach
  - diagonal
  - expand
  - imag
  - lift_fresh
  - narrow
  - permute
  - pin_memory
  - positive
  - real
  - reshape
  - squeeze
  - t
  - unfold
  - unsqueeze
  - view
  - view_as_complex
  - view_as_real
  - most operations whose name ends with an underscore


tohtana:
Hi @eternalNight,
I really appreciate your deep understanding of the code. This fix is awesome.
I wonder if we could show a message to say the performance might not be the best when an old PyTorch version is used and ops do have the schema. Can you add it somewhere?

eternalNight:
> Hi @eternalNight, I really appreciate your deep understanding of the code. This fix is awesome. I wonder if we could show a message to say the performance might not be the best when an old PyTorch version is used and ops do have the schema. Can you add it somewhere?

Good point. I should have tested the logic on different PyTorch versions.

I just tried Python 3.8.20 with PyTorch 2.0.0, and confirmed that all ops under torch.ops.aten have schema and the logic in this PR works as expected. So older PyTorch versions (not older than 2.0.0 since torch compile is not officially introduced then) should not be a problem.

Speaking of future versions where the schema of some ops is missing, to be conservative we should assume that they **MAY** reuse tensor storage. I'll update the except case to include such ops as well as print a warning message on its potential performance impact.

tohtana:
Hi @eternalNight,
Thank you for sharing the insight. As DeepCompile intercepts an internal function of a backward pass, it depends on specific versions of PyTorch. I think we need to consider only >=v2.6.

tohtana (approved):

PR #7563, opened 2025-09-15
deepcompile: Record graph order using OrderedDict

eternalNight:
On clear, GraphOrder does not clears ordered_frames. That may confuses subsequent passes after the first iteration.

Use an OrderedDict to record the mapping from frame IDs to other graph-related information.

Also fix the type annotation of graph_order which is a list of (int , bool) tuples instead of a list of int.


tohtana:
@eternalNight Can you merge master into your branch?

eternalNight:
> @eternalNight Can you merge master into your branch?

Sure. Branch rebased.

PR #7569, opened 2025-09-19
[RFC] Deepcompile: Fix bugs when applying deepcompile to VLA-like models

eternalNight:
**Describe the bug**

When applying deepcompile to the OpenVLA model (which is composed of two vision transformers and a llama-7B), I met the following issues:

a. Not all parameters are trained, which leads to compile-time exceptions as well as incorrect invocation of `endBackward()`.
b. A use-before-define error in `fast_free_schedule()`.
c. `release_param()` can be passed a tuple, not a tensor.

This PR attempts to fix all of those issues. Patch 1~2 resolves a, 3 resolves b and 4 resolves c.

**To Reproduce the issues**
Use this script: https://gist.github.com/eternalNight/3c2cf8c703f1e9e7742d3b7f9e1edae3

1. `deepspeed --num_gpus=N openvla-like.py -c`

**Remaining issues**
This PR is currently under draft due to the following issue:

1. `release_param()` does not clone tensors in the passed tuple when using the inductor backend.

### DIFF
--- a/csrc/compile/init.cpp
+++ b/csrc/compile/init.cpp
@@ -22,6 +22,7 @@ TORCH_LIBRARY(dc, m)
     m.def("wait_reload(Tensor a, int id, int id) -> Tensor");
     m.def("offload_parameter(Tensor a, int id, int id) -> ()");
     m.def("reload_parameter(Tensor a, int id, int id) -> ()");
+    m.def("end_backward(int graph_id) -> ()");
 
     m.def("test_call(Tensor a) -> Tensor");
 }
@@ -74,6 +75,10 @@ TORCH_LIBRARY_IMPL(dc, Meta, m)
     m.impl("offload_parameter", &dc::offload_parameter_meta);
 }
 
+// The "Undefined" dispatch key is for operations whose arguments do not contain
+// a tensor.
+TORCH_LIBRARY_IMPL(dc, Undefined, m) { m.impl("end_backward", &dc::end_backward); }
+
 PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
 {
     m.def("set_persistent", &dc::set_persistent, "Set persistent flag for a parameter");
@@ -95,7 +100,6 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
     m.def("start_forward", &dc::start_forward, "Start forward pass");
     m.def("end_forward", &dc::end_forward, "End forward pass");
     m.def("start_backward", &dc::start_backward, "Start backward pass");
-    // m.def("end_backward", &dc::end_backward, "End backward pass");
     m.def("cleanup", &dc::cleanup, "Clean up DeepCompile");
     m.def("reset", &dc::reset, "Reset the state");
     m.def("invalidate_gathered_param", &dc::invalidate_gathered_param, "Invalidate gathered param");
--- a/csrc/compile/init.cpp
+++ b/csrc/compile/init.cpp
@@ -13,7 +13,7 @@ TORCH_LIBRARY(dc, m)
     m.def("allgather_param(Tensor a, int graph_id, int id) -> Tensor");
     m.def("prefetch_params_fused(int graph_id, Tensor[] params, int[] ids) -> ()");
     m.def("wait_allgather(Tensor a, int graph_id, int id) -> Tensor");
-    m.def("release_param(Tensor a, int graph_id, int id, int n_users) -> Tensor");
+    m.def("release_param(Any a, int graph_id, int id, int n_users) -> Any");
     m.def("reduce_grad(Tensor a, int graph_id, int id) -> Tensor");
     m.def("free_tensors(Tensor[] a) -> ()");
     m.def("offload_tensor(Tensor a, int id, int id) -> Tensor");
@@ -77,7 +77,11 @@ TORCH_LIBRARY_IMPL(dc, Meta, m)
 
 // The "Undefined" dispatch key is for operations whose arguments do not contain
 // a tensor.
-TORCH_LIBRARY_IMPL(dc, Undefined, m) { m.impl("end_backward", &dc::end_backward); }
+TORCH_LIBRARY_IMPL(dc, Undefined, m)
+{
+    m.impl("release_param", &dc::release_param);
+    m.impl("end_backward", &dc::end_backward);
+}
 
 PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
 {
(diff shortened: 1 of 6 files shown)


eternalNight:
@tohtana Would you please help review these changes? Also, may I know the motivation behind the tensor clone in release_param when the inductor backend is used?

eternalNight:
Update patch 3 to take an alternative approach as the previous one breaks deepcompile + inductor. Refer to the commit message for more information.

tohtana:
Hi @eternalNight, 
Thank you for the great fix! This significantly improves the coverage of DeepCompile. I have a few questions.

- Do we need the same fix for `end_backward` in z1 and z2?

- Regarding the latest fix for `release_param`, we showed this example.
```
    %native_layer_norm : [num_users=3] = call_function[target=torch.ops.aten.native_layer_norm.default](args = (%clone, [1024], %wait_allgather_ds_param__primals_8_5, %wait_allgather_ds_param__primals_9_6, 1e-06), kwargs = {})
    %getitem : [num_users=1] = call_function[target=operator.getitem](args = (%native_layer_norm, 0), kwargs = {})
    %release_ds_param_primals_9_getitem_6 : [num_users=1] = call_function[target=torch.ops.dc.release_param.default](args = (%getitem, 140032938687600, 6, 1), kwargs = {})
    %release_ds_param_primals_8_getitem_5 : [num_users=1] = call_function[target=torch.ops.dc.release_param.default](args = (%release_ds_param_primals_9_getitem_6, 140032938687600, 5, 1), kwargs = {})
```
I think there might be an operator that requires other elements of `native_layer_norm`. How do you connect `release_ds_param_primals_8_getitem_5` to it? If any of operators doesn't have the path to the end, doesn't inductor remove it?

> Also, may I know the motivation behind the tensor clone in release_param when the inductor backend is used?

Good question! I experienced that triton code generated by inductor releases the buffer soon after `release_param`, and it broke the result. I tried to set a flag to tell inductor to keep the buffer, but it didn't work after all.

eternalNight:
> Hi @eternalNight, Thank you for the great fix! This significantly improves the coverage of DeepCompile. I have a few questions.
> 
> * Do we need the same fix for `end_backward` in z1 and z2?

Good point. Will insert end_backward() call in zero1_compile.py, too.

> * Regarding the latest fix for `release_param`, we showed this example.
> 
> ```
>     %native_layer_norm : [num_users=3] = call_function[target=torch.ops.aten.native_layer_norm.default](args = (%clone, [1024], %wait_allgather_ds_param__primals_8_5, %wait_allgather_ds_param__primals_9_6, 1e-06), kwargs = {})
>     %getitem : [num_users=1] = call_function[target=operator.getitem](args = (%native_layer_norm, 0), kwargs = {})
>     %release_ds_param_primals_9_getitem_6 : [num_users=1] = call_function[target=torch.ops.dc.release_param.default](args = (%getitem, 140032938687600, 6, 1), kwargs = {})
>     %release_ds_param_primals_8_getitem_5 : [num_users=1] = call_function[target=torch.ops.dc.release_param.default](args = (%release_ds_param_primals_9_getitem_6, 140032938687600, 5, 1), kwargs = {})
> ```
> 
> I think there might be an operator that requires other elements of `native_layer_norm`. How do you connect `release_ds_param_primals_8_getitem_5` to it? If any of operators doesn't have the path to the end, doesn't inductor remove it?
> 

Yes, the other elements in the returned tuple are also used. Those elements are fetched from `%native_layer_norm` and have no relationship with any `%release_ds_param_primals_xxx`. A more complete view of the graph is:

```
%native_layer_norm : [num_users=3] = call_function[target=torch.ops.aten.native_layer_norm.default](args = (%clone, [1024], %wait_allgather_ds_param__primals_8_5, %wait_allgather_ds_param__primals_9_6, 1e-06), kwargs = {})
%getitem : [num_users=1] = call_function[target=operator.getitem](args = (%native_layer_norm, 0), kwargs = {})
%release_ds_param_primals_9_getitem_6 : [num_users=1] = call_function[target=torch.ops.dc.release_param.default](args = (%getitem, 140160312802992, 6, 1), kwargs = {})
%getitem_1 : [num_users=1] = call_function[target=operator.getitem](args = (%native_layer_norm, 1), kwargs = {})
%getitem_2 : [num_users=1] = call_function[target=operator.getitem](args = (%native_layer_norm, 2), kwargs = {})
%release_ds_param_primals_8_getitem_5 : [num_users=1] = call_function[target=torch.ops.dc.release_param.default](args = (%release_ds_param_primals_9_getitem_6, 140160312802992, 5, 1), kwargs = {})
%_to_copy_4 : [num_users=1] = call_function[target=torch.ops.aten._to_copy.default](args = (%release_ds_param_primals_8_getitem_5,), kwargs = {dtype: torch.bfloat16})
...
return (..., getitem_1, getitem_2, ...)
```

> > Also, may I know the motivation behind the tensor clone in release_param when the inductor backend is used?
> 
> Good question! I experienced that triton code generated by inductor releases the buffer soon after `release_param`, and it broke the result. I tried to set a flag to tell inductor to keep the buffer, but it didn't work after all.

That sounds weird. I may be able to take a closer look when I have time.

eternalNight:
@tohtana BTW, are we limited to call only backward-compatible APIs of FX graphs? To get the output node of a graph, `fx.Graph` already provides [`get_output_node`](https://github.com/pytorch/pytorch/blob/main/torch/fx/graph.py#L1091). I was wondering why we define our own [`get_output_node`](https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/compile/fx.py#L15).

eternalNight:
Call to `end_backward` is added to z1 and z2 as well.

tohtana:
> @tohtana BTW, are we limited to call only backward-compatible APIs of FX graphs? To get the output node of a graph, `fx.Graph` already provides [`get_output_node`](https://github.com/pytorch/pytorch/blob/main/torch/fx/graph.py#L1091). I was wondering why we define our own [`get_output_node`](https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/compile/fx.py#L15).

Probably I didn't find the official API when writing the code. Currently, I think DeepCompile works with v2.6 or later, but I'm flexible to support only newer versions if we have a good reason..

eternalNight:
> > @tohtana BTW, are we limited to call only backward-compatible APIs of FX graphs? To get the output node of a graph, `fx.Graph` already provides [`get_output_node`](https://github.com/pytorch/pytorch/blob/main/torch/fx/graph.py#L1091). I was wondering why we define our own [`get_output_node`](https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/compile/fx.py#L15).
> 
> Probably I didn't find the official API when writing the code. Currently, I think DeepCompile works with v2.6 or later, but I'm flexible to support only newer versions if we have a good reason..

This `output_node` API is introduced [last November](https://github.com/pytorch/pytorch/commit/071d48c56e6451f95a87507cddd68e836428a68e) in v2.6.0a0. Let's leave it as is for this PR, and we can later switch to those APIs after they stablize.

PR #7576, opened 2025-09-22
logging: Also set log level of logger handlers

eternalNight:
After #7526 the default logger passes logs to a StreamHandler, which has its own log level. Changing the log level of the logger alone does not take effect in such case.

Update the log level of all handlers when changing the parent logger's.

### DIFF
--- a/deepspeed/utils/logging.py
+++ b/deepspeed/utils/logging.py
@@ -155,6 +155,8 @@ def set_log_level_from_string(log_level_str, custom_logger=None):
     if custom_logger is None:
         custom_logger = logger
     custom_logger.setLevel(log_level)
+    for ch in custom_logger.handlers:
+        ch.setLevel(log_level)
 
 
 def get_current_level():


deepspeed/utils/logging.py
```diff
@@ -155,6 +155,8 @@ def set_log_level_from_string(log_level_str, custom_logger=None):
     if custom_logger is None:
         custom_logger = logger
     custom_logger.setLevel(log_level)
+    for ch in custom_logger.handlers:
+        ch.setLevel(log_level)
```
stas00:
```suggestion
    for handler in custom_logger.handlers:
        handler.setLevel(log_level)
```

stas00 (changes requested):
Thank you for this improvement.

For some reason I can't push into your PR, so please also change the function docstring to reflect this change. Suggesting:

```
   """Sets a log level in the passed `logger` and its handlers from string. e.g. "info" => `logging.INFO`
```

deepspeed/utils/logging.py
eternalNight: Updated.

eternalNight:
> Thank you for this improvement, @eternalNight
> 
> For some reason I can't push into your PR, so please also change the function docstring to reflect this change. Suggesting:
> 
> ```
>    """Sets a log level in the passed `logger` and its handlers from string. e.g. "info" => `logging.INFO`
> ```

Docstring updated.

stas00:
Thank you!

Issue #7584, opened 2025-09-23
[BUG] universal checkpoin for stage3 fail when there are multiple subgroups


zhengchenyu:
**Describe the bug**

When the model is large and there are multiple subgroups, we use ds_to_universal.py, will fail ,the error log are below:

```
*** 1. Extracting ZeRO fragments
  0%|                                                     | 0/1 [00:03<?, ?it/s]
Traceback (most recent call last):
  File "/work/zhengchenyu/ai-project/qwen3/scripts/ds_to_universal_example.py", line 21, in <module>
    main()
  File "/work/zhengchenyu/ai-project/qwen3/scripts/ds_to_universal_example.py", line 18, in main
    ds_to_universal_main(args)
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 523, in main
    _extract_zero_shard_files_stage3(args, optim_files, param_shapes, dp_degree, temp_dir)
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 375, in _extract_zero_shard_files_stage3
    _do_parallel_work(do_work, list(range(dp_degree)), args.num_extract_workers)
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 359, in _do_parallel_work
    results.append(do_work(work))
                   ^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 167, in extract_zero_shards_stage3
    dump_param_fragment(temp_dir, 0, dp_index, state_key, flat_state[state_key], name, offset,
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 194, in dump_param_fragment
    state_flat_tensor = state_flat_tensor.narrow(0, offset, numel).clone()
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: start (0) + length (155582464) exceeds dimension size (74499072).

```

**To Reproduce**
1. Use large model to run, or set sub_group_size to a lower value. Then train and save model
2. Run ds_to_universal.py

**The reason**

I found that the previous stage3 universal checkpoint implementation did not take subgroups into account. I also found the following problems during debugging.

* Unable to handle multiple sub-groups, which will result in data loss
[...]

PR #5475, opened 2024-04-29
Universal checkpoint for zero stage 3

xylian86:
This PR enables the universal checkpoint for zero stage 3. 

Notes:

- The current implementation supports Data parallelism.
- Development is ongoing for universal checkpoint Stage 3 with tensor-slicing model parallelism.
- Pipeline parallelism is not supported by ZeRO Stage 3, and hence is not included in this universal checkpoint implementation.

In this PR:

- I've updated `deepspeed/checkpoint/ds_to_universal.py ` to support converting Zero checkpoints into Universal checkpoints.
- I've updated `deepspeed/runtime/zero/stage3.py` to enable loading Universal checkpoints using the Stage 3 optimizer.

PR #7585, opened 2025-09-23
Fix the universal checkpoint issue for stage3 when there are multiple subgroups.

zhengchenyu:
**Describe the bug**

When the model is large and there are multiple subgroups, we use ds_to_universal.py, will fail ,the error log are below:

```
*** 1. Extracting ZeRO fragments
  0%|                                                     | 0/1 [00:03<?, ?it/s]
  File "/work/zhengchenyu/ai-project/qwen3/scripts/ds_to_universal_example.py", line 21, in <module>
    main()
  File "/work/zhengchenyu/ai-project/qwen3/scripts/ds_to_universal_example.py", line 18, in main
    ds_to_universal_main(args)
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 523, in main
    _extract_zero_shard_files_stage3(args, optim_files, param_shapes, dp_degree, temp_dir)
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 375, in _extract_zero_shard_files_stage3
    _do_parallel_work(do_work, list(range(dp_degree)), args.num_extract_workers)
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 359, in _do_parallel_work
    results.append(do_work(work))
                   ^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 167, in extract_zero_shards_stage3
    dump_param_fragment(temp_dir, 0, dp_index, state_key, flat_state[state_key], name, offset,
  File "/opt/conda/lib/python3.11/site-packages/deepspeed/checkpoint/ds_to_universal.py", line 194, in dump_param_fragment
    state_flat_tensor = state_flat_tensor.narrow(0, offset, numel).clone()
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: start (0) + length (155582464) exceeds dimension size (74499072).

```

**To Reproduce**
1. Use large model to run, or set sub_group_size to a lower value. Then train and save model
2. Run ds_to_universal.py

**The reason**

I found that the previous stage3 universal checkpoint implementation did not take subgroups into account. I also found the following problems during debugging.

* Unable to handle multiple sub-groups, which will result in data loss
* When load_checkpoint is True, then all process will save to same zero model checkpoint file. If multiple processes write at the same time, the file will be corrupted. Occasionally, file corruption was discovered during testing.

Relete issue:  #7584

### DIFF
--- a/deepspeed/checkpoint/ds_to_universal.py
+++ b/deepspeed/checkpoint/ds_to_universal.py
@@ -152,21 +152,21 @@ def extract_zero_shards(dir, ds_checkpoint, indices_3D):
 def extract_zero_shards_stage3(optim_files, param_shapes, dp_degree, temp_dir, dp_index):
     state_dict = torch.load(optim_files[dp_index], map_location='cpu', weights_only=False)
 
-    flat_state = dict(
-        exp_avg=state_dict[OPTIMIZER_STATE_DICT]['optimizer_state_dict']['state'][0]["exp_avg"],
-        exp_avg_sq=state_dict[OPTIMIZER_STATE_DICT]['optimizer_state_dict']['state'][0]["exp_avg_sq"],
-        fp32=state_dict[OPTIMIZER_STATE_DICT]['fp32_flat_groups'][0],
-    )
-
-    offset = 0
-    for name, shape in param_shapes.items():
-        unpartitioned_numel = shape.numel()
-        partitioned_numel, _ = _zero_partitioned_param_info(unpartitioned_numel, dp_degree)
-        padding_free_numel = min(partitioned_numel, abs(unpartitioned_numel - dp_index * partitioned_numel))
-        for state_key in flat_state.keys():
-            dump_param_fragment(temp_dir, 0, dp_index, state_key, flat_state[state_key], name, offset,
-                                padding_free_numel)
-        offset += partitioned_numel
+    for idx, sub_group_shape in enumerate(param_shapes):
+        flat_state = dict(
+            exp_avg=state_dict[OPTIMIZER_STATE_DICT]['optimizer_state_dict']['state'][idx]["exp_avg"],
+            exp_avg_sq=state_dict[OPTIMIZER_STATE_DICT]['optimizer_state_dict']['state'][idx]["exp_avg_sq"],
+            fp32=state_dict[OPTIMIZER_STATE_DICT]['fp32_flat_groups'][idx],
+        )
+        offset = 0
+        for name, shape in sub_group_shape.items():
+            unpartitioned_numel = shape.numel()
+            partitioned_numel, _ = _zero_partitioned_param_info(unpartitioned_numel, dp_degree)
+            padding_free_numel = min(partitioned_numel, abs(unpartitioned_numel - dp_index * partitioned_numel))
+            for state_key in flat_state.keys():
+                dump_param_fragment(temp_dir, 0, dp_index, state_key, flat_state[state_key], name, offset,
+                                    padding_free_numel)
+            offset += partitioned_numel
 
 
 cnt = 0
@@ -390,10 +390,10 @@ def _merge_tp_slice_files(args, ds_checkpoint, slice_shapes, temp_dir):
         print(f'Warning: Unused patterns={unmatched_patterns} while merging tp slices')
 
 
-def _merge_zero3_slice_files(args, param_shapes, dp_degree, temp_dir):
+def _merge_zero3_slice_files(args, param_keys, dp_degree, temp_dir):
     zero_output_folder = os.path.join(args.output_folder, "zero")
     do_work = partial(merge_zero3_slices, dp_degree, zero_output_folder, temp_dir)
-    _do_parallel_work(do_work, param_shapes.keys(), args.num_merge_workers)
+    _do_parallel_work(do_work, param_keys, args.num_merge_workers)
 
 
 def _zero_partitioned_param_info(unpartitioned_numel, world_size):
@@ -514,8 +514,7 @@ def main(args):
     else:
         model_files = _get_model_state_files(args.input_folder)
         param_shapes = _parse_model_states_stage3(model_files)
-        param_shapes = {k: v for d in param_shapes for k, v in d.items()}
-        dp_degree = len(model_files)
+        dp_degree = len(optim_files)
 
         temp_dir = os.path.join(args.output_folder, 'tmp')
 
@@ -523,7 +522,8 @@ def main(args):
         _extract_zero_shard_files_stage3(args, optim_files, param_shapes, dp_degree, temp_dir)
 
         print('*** 2. Merging slices .....')
-        _merge_zero3_slice_files(args, param_shapes, dp_degree, temp_dir)
+        param_keys = {key for sub_group_shapes in param_shapes for key in sub_group_shapes.keys()}
+        _merge_zero3_slice_files(args, param_keys, dp_degree, temp_dir)
 
         print('*** 3. Saving common optimizer states')
         _save_optimizer_state_stage3(args, optim_files)
--- a/deepspeed/checkpoint/ds_to_universal.py
+++ b/deepspeed/checkpoint/ds_to_universal.py
@@ -514,7 +514,7 @@ def main(args):
     else:
         model_files = _get_model_state_files(args.input_folder)
         param_shapes = _parse_model_states_stage3(model_files)
-        dp_degree = len(optim_files)
+        dp_degree = len(model_files)
 
         temp_dir = os.path.join(args.output_folder, 'tmp')
 
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1829,6 +1829,7 @@ class DeepSpeedEngine(Module):
                 optimizer = DeepSpeedZeroOptimizer_Stage3(
                     self.module,
                     optimizer,
+                    self.param_names,
                     timers=timers,
                     ds_config=self.config,
                     static_loss_scale=self.loss_scale(),
@@ -1876,6 +1877,7 @@ class DeepSpeedEngine(Module):
         model_dtype, gradient_accumulation_dtype = self.get_data_types()
         optimizer = MiCS_Optimizer(self.module,
                                    basic_optimizer,
+                                   self.param_names,
                                    timers=timers,
                                    ds_config=self.config,
                                    static_loss_scale=self.loss_scale(),
@@ -2992,7 +2994,7 @@ class DeepSpeedEngine(Module):
         bf16_mode = self.bfloat16_enabled()
         return self._get_rank_zero_ckpt_name(checkpoints_path, tag, mp_rank, pp_rank, bf16_mode)
 
-    def _get_ckpt_name(self, checkpoints_path, tag, mp_placeholder=None):
+    def _get_ckpt_name(self, checkpoints_path, tag, mp_placeholder=None, read_mode=False):
         if mp_placeholder is not None:
             mp_rank_str = mp_placeholder
         else:
@@ -3000,10 +3002,11 @@ class DeepSpeedEngine(Module):
             mp_rank_str = f"{mp_rank:02d}"
 
         if self.zero_optimization_partition_weights():
-            if self.load_universal_checkpoint():
-                filename = "zero_pp_rank_0"
-            else:
-                filename = "zero_pp_rank_{}".format(dist.get_rank(group=self.optimizer.dp_process_group))
+            # For stage 3, when loading the checkpoint, the world size may change, non-existent files may be loaded,
+            # so load the first ckpt file for every worker. When writing the checkpoint, each process still writes
+            # its own ckpt file to avoid conflicts.
+            filename = "zero_pp_rank_{}".format(0 if read_mode else dist.get_rank(
+                group=self.optimizer.dp_process_group))
             ckpt_name = os.path.join(
                 checkpoints_path,
                 str(tag),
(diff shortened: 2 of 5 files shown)


deepspeed/runtime/engine.py
```diff
@@ -2992,18 +2994,19 @@ def _get_zero_ckpt_name(self, checkpoints_path, tag):
...
         else:
             mp_rank = 0 if self.mpu is None else self.mpu.get_model_parallel_rank()
             mp_rank_str = f"{mp_rank:02d}"
 
         if self.zero_optimization_partition_weights():
-            if self.load_universal_checkpoint():
-                filename = "zero_pp_rank_0"
-            else:
-                filename = "zero_pp_rank_{}".format(dist.get_rank(group=self.optimizer.dp_process_group))
+            # For stage 3, when loading the checkpoint, the world size may change, non-existent files may be loaded,
```
sfc-gh-truwase: Please clarify this idea as it appears to contradict the design principle that we only support world size changes through universal checkpointing. In other words, if `self.load_universal_checkpoint() == False` there is no world size change.

deepspeed/runtime/engine.py
```diff
@@ -3038,7 +3041,7 @@ def _get_expert_ckpt_name(checkpoints_path, layer_id, expert_id, tag, mpu=None):
 
     def _get_all_ckpt_names(self, checkpoints_path, tag):
         # It is required that (checkpoints_path, tag) are consistent among all ranks.
-        ckpt_file_pattern = self._get_ckpt_name(checkpoints_path, tag, mp_placeholder="*")
+        ckpt_file_pattern = self._get_ckpt_name(checkpoints_path, tag, mp_placeholder="*", read_mode=True)
```
sfc-gh-truwase: Can you please clarify what is special about this call site that makes `read_mode==True`

sfc-gh-truwase:
@xylian86 can you help review?

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -514,16 +514,16 @@ def main(args):
     else:
         model_files = _get_model_state_files(args.input_folder)
         param_shapes = _parse_model_states_stage3(model_files)
-        param_shapes = {k: v for d in param_shapes for k, v in d.items()}
-        dp_degree = len(model_files)
+        dp_degree = len(optim_files)
```
xylian86: How about let's still use `dp_degree = len(model_files)`? By design, the number of `optim_files` should match the number of `model_files`, since each DP worker saves both a model file and an optimizer file. However, if a developer chooses to resume training using only the weights and deletes the optimizer state files, then `len(optim_files) = 0`, which would cause issues.

xylian86:
> `When the model is large and there are multiple subgroups, we use ds_to_universal.py, will fail `

@zhengchenyu Thanks for raising this issue and submitting the PR to address it. I just want to clarify that this is not the actual cause of the problem. The current implementation can already handle multiple subgroups, regardless of how many we shard.

The real cause is that when we initialize the ZeRO-3 Engine, the optimizer passes multiple [param_groups](https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/runtime/zero/stage3.py#L514). These param_groups may have different hyperparameters, but the current implementation does not yet account for that.

deepspeed/runtime/engine.py
zhengchenyu: After https://github.com/deepspeedai/DeepSpeed/pull/5475, all process will save to zero_pp_rank_0. This can cause multiple processes to write to a file simultaneously and lead to file corruption. So I revert this part of the logic.
We have such a scene:
* (1) we use 2 process to train, we will generate two file: `zero_pp_rank_0_mp_rank_00_model_states.pt` and `zero_pp_rank_1_mp_rank_00_model_states.pt`. 
* (2) Then we use ds_to_universal.py, converting to universal checkpoint. 
* (3) Then we use 4 process to train, the rank 2 and rank 3 will read the non-existent file `zero_pp_rank_2_mp_rank_00_model_states.pt` and `zero_pp_rank_3_mp_rank_00_model_states.pt`. Then failed.

For stage3, we can read any file, so here just read `zero_pp_rank_0_mp_rank_00_model_states`.

And it is ok for stage1 and 2, because we have one model_states file for stag 1 and 2.

zhengchenyu:
@xylian86
You can see here:  https://github.com/deepspeedai/DeepSpeed/blob/bc9ed477e9c23f167761c1a122c76616e6b0ab70/deepspeed/runtime/zero/stage3.py#L727

Not only will the incoming param_group be split, but if the param_group is too large (i.e., the group exceeds sub_group_size), it will be split twice.

The hyperparameters have been retained, see the following code:
```
     for param_group in self.optimizer.param_groups:
            # Generally, the hyperparameters of each parameter should be the same, we can obtain from any parameter.
            for key, value in optim_sd[OPTIMIZER_STATE_DICT]["param_groups"][0].items():
                if key == 'params':
                    param_group['params'] = []
                else:
                    param_group[key] = value
```

Each sub group can only belong to one incoming param group, so the hyperparameters are not lost here.

deepspeed/runtime/engine.py
zhengchenyu: Just same with the upper comment...

xylian86:
@zhengchenyu 

There are two levels of parameter grouping in the implementation:

1. **Higher level: `param_group`** - From the user's optimizer config.
2. **Lower level: `subgroup`** - Each `param_group` is further split based on `sub_group_size`

For example, the structure looks like:
```python
{
  'param_group1': [subgroup1, subgroup2, subgroup3], 
  'param_group2': [subgroup1, subgroup2]
}
```

**Already handled**: Multiple subgroups within a single `param_group`
- The existing code properly manages scenarios where one `param_group` contains multiple subgroups

**Not previously covered**: Multiple `param_groups` from user's optimizer configuration  
- This is precisely what your PR addresses, and I appreciate this contribution

> `When the model is large and there are multiple subgroups, we use ds_to_universal.py, will fail ,the error log are below:`

I disagree with the statement that model size or multiple subgroups cause the `ds_to_universal.py` failure. The current implementation can handle large models and any number of subgroups, **as long as there is only a single `param_group`**.

As I mentioned earlier, the root cause of the error is specifically the presence of **multiple `param_groups`** in the user's optimizer configuration, not the model size or subgroup count within individual parameter groups.

zhengchenyu:
@xylian86 

https://github.com/deepspeedai/DeepSpeed/blob/17d80ce4401d4453f6486c22607afb1043edaef1/deepspeed/runtime/zero/stage3.py#L726

Although `param_groups` is 2-level structure. Look the below code, `param_groups` will flatten to 1-level, just `bit16_groups`.

https://github.com/deepspeedai/DeepSpeed/blob/17d80ce4401d4453f6486c22607afb1043edaef1/deepspeed/runtime/zero/stage3.py#L735

If we only have one incoming param group, but very large, and split three sub group. the structure of `param_groups` looks like

```
{
  'param_group1': [subgroup1, subgroup2, subgroup3], 
}

```

Then the `bit16_groups` will be `[subgroup1, subgroup2, subgroup3]`. **The problem is whether fp16_groups has multiple subgroups. But not incoming has multiple subgroups.**

So my explanation is fine.

xylian86:
@zhengchenyu  Thank you for the clarification. I realize that when saving checkpoints, in `optimizer state`, each subgroup has separate tensors rather than being concatenated. I believe I’m aligned with you now.

deepspeed/runtime/engine.py
xylian86: @zhengchenyu  I can understand (1)(2)(3), but could you clarify why the previous version will fail?

https://github.com/deepspeedai/DeepSpeed/blob/17d80ce4401d4453f6486c22607afb1043edaef1/deepspeed/runtime/engine.py#L3003-L3004

When universal checkpoint is enabled, only the `zero_pp_rank_0*` will got loaded.

deepspeed/runtime/zero/stage3.py
```diff
@@ -141,6 +141,7 @@ def __init__(
         self,
         module,
         init_optimizer,
+        param_names,
```
xylian86, line 144: https://github.com/deepspeedai/DeepSpeed/blob/17d80ce4401d4453f6486c22607afb1043edaef1/deepspeed/runtime/zero/stage3.py#L2868-L2872

In earlier versions we passed `param_shapes` to supply parameter names for each subgroup. Could we reuse `param_shapes` here instead of introducing a new `param_names` argument, which seems redundant? Or is there a strong reason to replace `param_shapes` with `param_names`?

deepspeed/checkpoint/ds_to_universal.py
zhengchenyu: Good suggestion. I made this change because the number of model files was set to 1 before. However, this PR has reversed this logic. So I will switch back to using len(model_files).

deepspeed/runtime/engine.py
zhengchenyu: The previous version will not fail, but may produce corrupted files. Multiple processes may write to the same file. In my experiments, occasionally, there will be a problem that the model file cannot be loaded, and it prompts that the file is corrupted.
I just said that if I directly revert the relevant logic in #5475, it will cause the loading to fail. (1)(2)(3) just describe why loading fail if I directly revert the relevant logic in #5475.

deepspeed/runtime/zero/stage3.py
zhengchenyu: This is a critical question! 

But No, we can not use `param_shapes`! Because the `param_shapes` is from checkpoint, `param_shapes` comes from the checkpoint, that is from the last distributed environment, that is the `sub group` (or `fp16_group`)  is used in last distributed environment.

If the world size is changed, the `sub group` (or `fp16_group`) will changed. Here we set optimizer parameter for current `sub group` (or `fp16_group`). 

In fact, I also used `param_shapes` at first. But I found this problem when my new unit test (which sub_group_size is set to 100) failed to run.

zhengchenyu:
Add some experimental results, we used Qwen3, 0.6b, and 400 steps.
Experimental Group 1:
`Base` was trained without interruption, world_size=8. `Universal` was interrupted at steps 50, 100, 200, and 300, and load via the universal checkpoint, world_size=8。

截屏2025-09-25 19 06 34


Experimental Group 2:
`Base` was trained without interruption, world_size=8. `Universal_world_change` was interrupted at steps 50, 100, 200, and 300, with the world size changed, and then loaded via the universal checkpoint. At steps 50 and 200, the world_size was reduced from 8 to 4. At steps 100 and 300, the world_size was increased from 4 to 8.
> Total_batch_size was kept constant by modifying gradient_accumulation_steps.

截屏2025-09-25 19 09 49

deepspeed/runtime/zero/stage3.py
xylian86: @zhengchenyu Make sense to me. Could you remove the `param_shapes` to clean up the codebase  a bit? Since we’re using `param_names`, `param_shapes` is no longer needed..

deepspeed/runtime/engine.py
xylian86: @zhengchenyu I see your point. The loading logic itself is fine, but when we enable the universal checkpoint and then save it, the `_get_ckpt_name` function will also be called. At that point, filename gets set to `zero_pp_rank_0`, which could lead to corrupted files.

To make the codebase cleaner and avoid introducing extra complexity, we could consider refactoring:


1. Remove the `self.load_universal_checkpoint()` logic from `_get_ckpt_name`.

2. Instead, move that logic into `_get_all_ckpt_names` (which is for loading ckpt) and, if `self.load_universal_checkpoint()` is true, reset `ckpt_file_pattern` to `zero_pp_rank_0`.

This way, we avoid adding a new argument like `read_mode` and keep the code simpler.

xylian86:
@zhengchenyu Thank you for sharing the loss trace — it looks good to me. I’ve also left two additional comments on the codebase with suggestions for improvement; could you take a look and address those? The other parts LGTM.

deepspeed/runtime/engine.py
sfc-gh-truwase: @zhengchenyu and @xylian86 thanks for resolving the issues. 

@zhengchenyu please ping me when you address the last two comments so we can merge this PR. This is a great contribution. Thanks!

deepspeed/runtime/zero/stage3.py
zhengchenyu: OK, I have remove `param_shapes`.

deepspeed/runtime/engine.py
zhengchenyu: @xylian86 
The current implementation does not suffer from write corruption issues. `_get_all_ckpt_names` is only for `_load_checkpoint`. When we save, echo rank write their own `zero_pp_rank_{rank}`. It is fine.

And your suggestion to refactor is great, I have modified it and avoided introducing `read_mode`.

xylian86: @zhengchenyu Thank you!

xylian86:
@sfc-gh-truwase [zhengchenyu](https://github.com/zhengchenyu) has addressed all my comments. The current version LGTM.

sfc-gh-truwase (approved):

zhengchenyu:
@sfc-gh-truwase @xylian86 
Sorry, I found that there was a problem with the [last commit](https://github.com/deepspeedai/DeepSpeed/pull/7585/commits/ac14d7c9b1dfe0a20b059c7c3ce4de8c3e13c170), which caused the universal checkpoint to be unable to be loaded when the world size scale up.
I'm curious why the unit test didn't fail. Originally `TestZeROUniversalCheckpointDP::test_dp_world_size_2to4` could reproduce this problem
I have submit new PR https://github.com/deepspeedai/DeepSpeed/pull/7599 to solve it.

Issue #7577, opened 2025-09-22
[REQUEST] Improve GPU memory utilization with ZeRO3 + deepcompile + torch autocast + activation checkpointing


eternalNight:
# Problem Statement

Today training with deepcompile + torch autocast + activation checkpointing has sub-optimal GPU memory usage due to the following issues.

1. Parameters are gathered in FP32 and then downcasted to BF16/FP16, not vice versa (this also increases communication cost).
2. Downcasted parameters are saved for backward.
3. `free_activation` uses a hard-coded threshold.
4. Unused IPG bucket (deepcompile uses its own reduce buckets) and duplicated model parameters (`fp16` and `fp32` groups in `DeepSpeedZeroOptimizer_Stage3` are both in FP32 actually).

# Proposed solutions

To solve the issues above, my proposals are as follows.

- [x] (For issue 1) Fuse allgather and _to_copy: #7588
    + Enhance `allgather_param` to take an additional kwarg `dtype` which specify the desired dtype of the gathered parameters. If it is different from the original, downcast is done before allgather. 
    + Modify [`add_allgather_and_release`](https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/compile/passes/zero3_compile.py#L69) to create a fused downcast+allgather if the only user of a parameter is a `_to_copy` for type-casting.
    + Update `prefetch_params_fused` accordingly.
- [x] (For issue 2) Mark downcasted parameters as recomputable and use `min_cut_rematerialization_partition` to partition joint graphs. #7609 
- [x] (For issue 3) Make the threshold size of activation to free configurable. #7582
- [x] (For issue 4)
    + Release IPG buckets if deepcompile is enabled. #7610 
    + In `DeepSpeedZeroOptimizer_Stage3`, `fp32_partitioned_groups_flat` aliases `fp16_partitioned_groups_flat` when torch autocast is enabled. To be more accurate, the `fp16_xxx` members of `DeepSpeedZeroOptimizer_Stage3` should be renamed as their element size is not always 2-byte long. #7651 

# Design alternatives

For issue 2, an alternative is to create a separate pass that identifies saved parameters and drops them during forward and regathers them during backward. That is less straightforward compared to the proposed partitioner-based approach.

# Additional context

[...]

PR #7588, opened 2025-09-24
DeepCompile: Fuse allgather and downcast

eternalNight:
With autocast enabled, a majority of weights are downcasted before being used in calculations. Today zero3_compile gathers the FP32 weights before they are downcasted. That is sub-optimal because FP32 weights consumes more bandwidth to allgather and takes more time to downcast.

To reduce communication and downcast time, fuse allgather and downcast in the dc ops. The target type is now passed to allgather_param() and prefetch_params_fused() which will downcast the (partial) weights before launching allgathers.

This corresponds to issue 1 of #7577.

Tested with https://gist.github.com/eternalNight/3c2cf8c703f1e9e7742d3b7f9e1edae3 (run with `deepspeed --num_gpus=N this_file.py -c -p -m 23` to collect torch and memory profiles, and with DINOV2_DEPTH = SIGLIP_DEPTH = 3, LLAMA2_DEPTH = 4 for faster compileation) on 5090 (which has limited inter-GPU bandwidth), time per step decreases from 438ms to 337ms and peak GPU memory usage from 9.5GB to 8.5GB.

Profiles of a single step before this PR:

image

image

After this PR:

image

image

This PR also reduces peak memory usage because the `fast_free_schedule()` today always arranges param allgathers and downcasts at the beginning of the graph. While the original FP32 params can be freed early, all FP16/BF16-casted params are kept in GPU memory at the beginning of the backward graph, leading to a higher peak in memory usage.

P.S. Probably due to organization branch rule settings, I don't find anywhere to allow reviewers to modify the branch. So I'll update the branch per reviewers' comments and rebase if needed.

### DIFF
--- a/csrc/compile/init.cpp
+++ b/csrc/compile/init.cpp
@@ -10,8 +10,10 @@
 
 TORCH_LIBRARY(dc, m)
 {
-    m.def("allgather_param(Tensor a, int graph_id, int id) -> Tensor");
-    m.def("prefetch_params_fused(int graph_id, Tensor[] params, int[] ids) -> ()");
+    m.def("allgather_param(Tensor a, int graph_id, int id, ScalarType? dtype = None) -> Tensor");
+    m.def(
+        "prefetch_params_fused(int graph_id, Tensor[] params, int[] ids,"
+        "                      ScalarType[]? dtypes = None) -> ()");
     m.def("wait_allgather(Tensor(a) a, int graph_id, int id) -> Tensor(a)");
     m.def("release_param(Tensor(a) a, int graph_id, int id, int n_users) -> Tensor(a)");
     m.def("reduce_grad(Tensor a, int graph_id, int id) -> Tensor");
--- a/csrc/compile/z3.cpp
+++ b/csrc/compile/z3.cpp
@@ -68,7 +68,12 @@ public:
                          c10::intrusive_ptr<c10d::symmetric_memory::SymmetricMemory> symm_mem)
     {
         const DSParam& param = param_registry_->getParam(ds_id);
-        const at::Tensor& ds_tensor = param.getDSTensor();
+        at::Tensor ds_tensor = param.getDSTensor();
+
+        if (ds_tensor.scalar_type() != output_buf.scalar_type()) {
+            at::cuda::CUDAStreamGuard guard(ag_stream_);
+            ds_tensor = ds_tensor.to(output_buf.scalar_type(), true, true);
+        }
 
         if (symm_mem == nullptr) {
             // Fast path: assume uniform shard sizes (ZeRO-3 partitions are padded to uniform size)
@@ -110,6 +115,7 @@ public:
     }
 
     at::Tensor allgatherParam(long ds_id,
+                              std::optional<at::ScalarType> dtype,
                               c10::intrusive_ptr<c10d::symmetric_memory::SymmetricMemory> symm_mem)
     {
         const DSParam& param = param_registry_->getParam(ds_id);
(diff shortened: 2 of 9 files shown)


tohtana:
@eternalNight, 

Thank you for submitting a great PR! I believe this will bring a significant improvement across many use cases.

I’m happy to merge this change, but I’d like your feedback on one point:
Right now, the approach is to pass the downcasted dtype to allgather. As an alternative, would it be possible to identify a downcast-taking parameter and place allgather after the downcast?

The advantage of this approach is that it keeps the C++ side (the more primitive components in DeepCompile) simpler. That said, there may be trade-offs compared to your current design.

I’d really appreciate your thoughts on this.

eternalNight:
> @eternalNight,
> 
> Thank you for submitting a great PR! I believe this will bring a significant improvement across many use cases.
> 
> I’m happy to merge this change, but I’d like your feedback on one point: Right now, the approach is to pass the downcasted dtype to allgather. As an alternative, would it be possible to identify a downcast-taking parameter and place allgather after the downcast?
> 
> The advantage of this approach is that it keeps the C++ side (the more primitive components in DeepCompile) simpler. That said, there may be trade-offs compared to your current design.
> 

Adding allgather after the param downcast is possibly feasible. The major differences from this version are:

1. We don't need the additional `dtype` argument in `allgather_param()`.
2. We still need the `dtype` argument in `executor->allgatherParam` because the `output_buf` needs that dtype to determine its size.
3. Similar to `prefetch_params_fused()` and `executor->prefetchParamsFused()`. But additionally we need to move the downcast ops ahead of the prefetch one.
4. Type-casting is unnecessary in `launchAllGather()`.
5. Either the downcast op should run on the ag stream (can we move an existing op to a separate stream?), or we need an additional event to ensure that allgather does not start till the downcast completes.

On the C++ side, 1-4 reduces complexity (a bit) while 5 increases complexity. So I'm not sure if either approach is really simpler than the other.

> I’d really appreciate your thoughts on this.

eternalNight:
BTW, I found today that, with the inductor backend, loss of that openvla-like test became NaN after a few iterations, and that was so even without torch autocast. The same setup works fine with the eager backend.

I suspect that it's related to tensor lifecycle, but need some more time to find out.

tohtana:
@eternalNight Thanks for sharing your insights! I was considering whether it might be useful to implement various features as a graph modification, but I think your approach makes sense for this PR.

> BTW, I found today that, with the inductor backend, loss of that openvla-like test became NaN after a few iterations, and that was so even without torch autocast. The same setup works fine with the eager backend.

Did you encounter this issue without the changes in this PR? I’ve run into similar problems while I was writing the code of DeepCompile. In most cases, Triton code generated by Inductor releases buffer tensors earlier than expected, causing the tensor to be destroyed.
I tested this with several models, including LLaMA and Mixtral, but it’s possible I missed a pattern that shows up in a different architecture.

eternalNight:
> > BTW, I found today that, with the inductor backend, loss of that openvla-like test became NaN after a few iterations, and that was so even without torch autocast. The same setup works fine with the eager backend.
> 
> Did you encounter this issue without the changes in this PR? I’ve run into similar problems while I was writing the code of DeepCompile. In most cases, Triton code generated by Inductor releases buffer tensors earlier than expected, causing the tensor to be destroyed. I tested this with several models, including LLaMA and Mixtral, but it’s possible I missed a pattern that shows up in a different architecture.

Yes, the issue exists before this PR and without torch autocast.

eternalNight:
@tohtana I think I have found (at least one of) the cause of that loss NaN issue.

Torch op schema can specify if the output tensor aliases the storage of an input one. Take `t()` as an example: `aten::t(Tensor(a) self) -> Tensor(a)`.

Inductor-generated code explicitly drop a tensor when it's no longer needed. So for `allgather_param`, it looks like below (note the `del` statements):

```
buf0 = torch.ops.dc.allgather_param.default(primals_53, 139956546525584, 49)
del primals_53
buf1 = buf0
del buf0
```

Now for `release_param()`, it is declared as `release_param(Tensor a, int graph_id, int id, int n_users) -> Tensor`, but the returned tensor is just `a`, not a copy of `a`. With the `del` that follows, `a` can be freed early. Changing the schema to `release_param(Tensor(a) a, int graph_id, int id, int n_users) -> Tensor(a)` makes the situation better but only a bit: losses are back to normal only when I commented out `fast_free_schedule` in zero3_compile.py. There probably is yet another cause.

Anyway, this is a different issue from what this PR is aimed at. I'll open a separate issue for this.

Update: That "another cause" is the previous workaround, i.e., cloning the input tensor in `release_param`. After cleaning that up, the loss NaN issue disappears. Refer to #7597 for the full code change. This PR has been tested to be compatible with that one.

tohtana (approved):
Let's merge this after #7597 is merged. Please let me know when you are okay for merge (I think this PR should have some conflicts with #7597)

eternalNight:
> Let's merge this after #7597 is merged. Please let me know when you are okay for merge (I think this PR should have some conflicts with #7597)

@tohtana Rebased and conflicts resolved.

Issue #7596, opened 2025-09-26
[BUG] Schema of DeepCompile C++ ops does not reflect aliases


eternalNight:
**Describe the bug**

Today the schema of DeepCompile C++ ops is defined as follows. Specifically all ops that return tensors are declared as returning a fresh-new one, which is incorrect as `wait_allgather()` and `release_param` actually return their first tensor inputs without copying them.

```
    m.def("allgather_param(Tensor a, int graph_id, int id) -> Tensor");
    m.def("prefetch_params_fused(int graph_id, Tensor[] params, int[] ids) -> ()");
    m.def("wait_allgather(Tensor a, int graph_id, int id) -> Tensor");
    m.def("release_param(Tensor a, int graph_id, int id, int n_users) -> Tensor");
    m.def("reduce_grad(Tensor a, int graph_id, int id) -> Tensor");
    m.def("free_tensors(Tensor[] a) -> ()");
    m.def("offload_tensor(Tensor a, int id, int id) -> Tensor");
    m.def("reload_tensor(Tensor a, int id, int id) -> Tensor");
    m.def("wait_offload(Tensor a, int id, int id) -> Tensor");
    m.def("wait_reload(Tensor a, int id, int id) -> Tensor");
    m.def("offload_parameter(Tensor a, int id, int id) -> ()");
    m.def("reload_parameter(Tensor a, int id, int id) -> ()");
    m.def("end_backward(int graph_id) -> ()");
```

This causes the tensors to be freed, earlier than expected, by the additional tensor `del` statements in inductor-generated code, and eventually causes training losses to become NaN.

**To Reproduce**
1. Run `deepspeed --num_gpus=N openvla-like.py -c`

[openvla-like.py](https://gist.github.com/eternalNight/3c2cf8c703f1e9e7742d3b7f9e1edae3)

**Expected behavior**
With DeepCompile, loss curves of eager & inductor backends match.

eternalNight:
@tohtana This is the issue mentioned in #7588.

Here's my gut feeling of the schemas that should be used:

```
    m.def("wait_allgather(Tensor(a) a, int graph_id, int id) -> Tensor(a)");
    m.def("release_param(Tensor(a) a, int graph_id, int id, int n_users) -> Tensor(a)");
```

Would you please review if there is any other API whose schema should be updated as well (e.g. I'm not quite sure about `reduce_grad`)?

eternalNight:
[...]

PR #7597, opened 2025-09-26
DeepCompile: Specify tensor aliasing in C++ op schema

eternalNight:
PyTorch C++ op schema [1] allows specifying tensor storage aliasing by annotating `(a)` after input/output types. Torch inductor takes this information to determine where to insert explicit `del` statements for tensors that are no longer needed.

If what an op schema specifies disagrees with the op implementation, inductor-generated code is likely to release tensors earlier than expected and leads to wrong results.

`wait_allgather` and `release_param` return the first argument unchanged and that aliasing should be annotated in the schema.

Also remove the code related to `clone_custom_op_output` as it is solely a workaround of the aforementioned issue.

[1] https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/README.md

Fixes: #7596


tohtana (approved):
Wow, amazing finding! I really appreciate your deep investigation.
Thank you for the great fix!

PR #7599, opened 2025-09-28
Fixed the issue that universal checkpoint cannot be loaded for stage3 when world size expansion.

zhengchenyu:
When the world size expands from 2 to 4, then convert to universal checkpoint, and load from universal checkpoint.
The new rank, for example, rank3 will load model file `zero_pp_rank_3_mp_rank_00_model_states.pt`. But this file was not produced during the last execution.
For stage3, just load the first file, that is `zero_pp_rank_0_mp_rank_00_model_states`.
The existing unit test TestZeROUniversalCheckpointDP::test_dp_world_size_2to4 can verify this problem.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -3004,7 +3004,7 @@ class DeepSpeedEngine(Module):
         bf16_mode = self.bfloat16_enabled()
         return self._get_rank_zero_ckpt_name(checkpoints_path, tag, mp_rank, pp_rank, bf16_mode)
 
-    def _get_ckpt_name(self, checkpoints_path, tag, mp_placeholder=None):
+    def _get_ckpt_name(self, checkpoints_path, tag, mp_placeholder=None, pp_placeholder=None):
         if mp_placeholder is not None:
             mp_rank_str = mp_placeholder
         else:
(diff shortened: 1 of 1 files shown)


zhengchenyu:
convert to drat pr until all unit test pass.

deepspeed/runtime/engine.py
```diff
@@ -3047,15 +3052,15 @@ def _get_expert_ckpt_name(checkpoints_path, layer_id, expert_id, tag, mpu=None):
 
     def _get_all_ckpt_names(self, checkpoints_path, tag):
         # It is required that (checkpoints_path, tag) are consistent among all ranks.
-        ckpt_file_pattern = self._get_ckpt_name(checkpoints_path, tag, mp_placeholder="*")
+        ckpt_file_pattern = self._get_ckpt_name(checkpoints_path,
+                                                tag,
+                                                mp_placeholder="*",
```
sfc-gh-truwase, line 3064: Is this not needed to guarantee `zero_pp_rank_0_mp_rank_0*` for checkpoint loading in zero stage 3 (line 3020)?

```suggestion
                                                mp_placeholder="0" if self.load_universal_checkpoint() else "*",
```

zhengchenyu: I think this `commit suggestion` is fine for stage 3. But not for stage 1 and 2.
Maybe `mp_placeholder="0" if self.load_universal_checkpoint() and self.zero_optimization_partition_weights() else "*",` is more suitable. But it seems that it is not necessary

sfc-gh-truwase: Thanks for the explanation.

sfc-gh-truwase (approved):

PR #7601, opened 2025-09-28
Fixed the problem of loading universal checkpoint error in multi-machine mode.

zhengchenyu:
In a multi-machine environment, loading the stage3 universal checkpoint will produce incorrect results, causing the loss to increase abnormally.

### DIFF
--- a/deepspeed/runtime/zero/stage3.py
+++ b/deepspeed/runtime/zero/stage3.py
@@ -2904,7 +2904,7 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
         self.overflow = sd.get('overflow', self.overflow)
 
     def load_hp_checkpoint_state(self, folder, key):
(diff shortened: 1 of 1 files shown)


zhengchenyu:
It seems that unit testing currently only works in a single-machine environment. Therefore, I'll use my own examples in for verification. Env: 2 node, 8 gpu per node. There are three experiments:
* `base`  The base example is a non-interrupt, non-checkpoint example.
* `bug`    The bug is based on the current master version of the code, with checkpoint at step 20, then convert to universal checkpoint, and then load the universal checkpoint. Abnormal increase in loss.
* `fix`      The fix is ​​based on this PR, also with checkpoint at step 20, then convert to universal checkpoint, and then reload the universal checkpoint. It can be seen that the loss is normal.

截屏2025-09-28 19 37 40

tohtana (approved):
Great catch! Thank you for the fix!

PR #7598, opened 2025-09-27
Fix deepcompile+stage 3 fails start

nguyen599:
Start training for Deepcompile+zero s3 fails due change introduced in this [#7548](https://github.com/deepspeedai/DeepSpeed/pull/7548)

```
[rank3]:   File "/usr/local/lib/python3.11/dist-packages/transformers/trainer.py", line 4071, in training_step
[rank3]:     self.accelerator.backward(loss, **kwargs)
[rank3]:   File "/usr/local/lib/python3.11/dist-packages/accelerate/accelerator.py", line 2726, in backward
[rank3]:     self.deepspeed_engine_wrapped.backward(loss, sync_gradients=self.sync_gradients, **kwargs)
[rank3]:   File "/usr/local/lib/python3.11/dist-packages/accelerate/utils/deepspeed.py", line 281, in backward
[rank3]:     self.engine.step()
[rank3]:   File "/usr/local/lib/python3.11/dist-packages/deepspeed/runtime/engine.py", line 2495, in step
[rank3]:     self._take_model_step(lr_kwargs)
[rank3]:   File "/usr/local/lib/python3.11/dist-packages/deepspeed/runtime/engine.py", line 2390, in _take_model_step
[rank3]:     self.optimizer.step()
[rank3]:   File "/usr/local/lib/python3.11/dist-packages/deepspeed/utils/nvtx.py", line 20, in wrapped_fn
[rank3]:     ret_val = func(*args, **kwargs)
[rank3]:               ^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/usr/local/lib/python3.11/dist-packages/deepspeed/runtime/zero/stage3.py", line 2131, in step
[rank3]:     norm_groups = self._get_norm_groups()
[rank3]:                   ^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/usr/local/lib/python3.11/dist-packages/deepspeed/utils/nvtx.py", line 20, in wrapped_fn
[rank3]:     ret_val = func(*args, **kwargs)
[rank3]:               ^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/usr/local/lib/python3.11/dist-packages/deepspeed/runtime/zero/stage3.py", line 1910, in _get_norm_groups
[rank3]:     norm_groups.append(self.get_grad_norm_direct(self.averaged_gradients[i], self.fp16_groups[i]))
[rank3]:                                                  ~~~~~~~~~~~~~~~~~~~~~~~^^^
[...]

Issue #7578, opened 2025-09-22
[BUG] Exception raised at the end of training with deepcompile enabled


eternalNight:
**Describe the bug**

With a training script like the following:

```
import deepspeed
import deepspeed.comm as dist

def main(args):
    deepspeed.init_distributed()
    model = Model()
    ......
    model.destroy()
    dist.destroy_process_group()
```

The following exception is raised at the end of the training process if and only if deepcompile is enabled:

```
Exception ignored in: <function DeepSpeedEngine.__del__ at 0x7f241b4fe830>
Traceback (most recent call last):
  File "/mnt/engines/deepspeed/deepspeed/runtime/engine.py", line 519, in __del__
    self.destroy()
  File "/mnt/engines/deepspeed/deepspeed/runtime/engine.py", line 523, in destroy
    self.optimizer.destroy()
  File "/mnt/engines/deepspeed/deepspeed/runtime/zero/stage3.py", line 468, in destroy
    self.parameter_offload.destroy()
  File "/mnt/engines/deepspeed/deepspeed/runtime/zero/parameter_offload.py", line 227, in destroy
    self._remove_module_hooks()
  File "/mnt/engines/deepspeed/deepspeed/runtime/zero/parameter_offload.py", line 241, in _remove_module_hooks
    print_rank_0(f'Deleted module hooks: forward = {num_forward_hooks}, backward = {num_backward_hooks}',
  File "/mnt/engines/deepspeed/deepspeed/runtime/zero/partition_parameters.py", line 113, in print_rank_0
    rank = dist.get_rank()
  File "/mnt/engines/deepspeed/deepspeed/comm/comm.py", line 720, in get_rank
    assert cdb is not None and cdb.is_initialized(
AssertionError: DeepSpeed backend not set, please initialize it using init_process_group()
```

**To Reproduce**

1. Run https://gist.github.com/eternalNight/3c2cf8c703f1e9e7742d3b7f9e1edae3 with `deepspeed --num_gpus=N openvla-like.py -c`

**Expected behavior**
No exception is raised.

eternalNight:
Not sure if the model is referenced in any global variable that defers its deletion.

Also, I think we should consider cleaning up duplicated definitions of `print_rank_0` and replacing them with calls to the global logger (after extending it with those multi-gpu debugging facilities).

PR #7603, opened 2025-09-28
Handle the case of DeepCompile's enabled but not activated

tohtana:
This PR improves state management for DeepCompile in the engine.

Previously, the system relied only on the config flag indicating whether DeepCompile was enabled. However, DeepCompile is actually activated only when `compile()` is called. This meant that if DeepCompile was enabled in the config but `compile()` was never called, it could lead to invalid internal states (as shown in #7598).

Since `enabled == True` should be interpreted as an option that modifies the behavior of `compile()`, this PR introduces clearer state management:
- If .compile() is not called, the DeepCompile config has no effect on behavior. A one-time message is shown instead.
- A new state, DeepCompile activated, is introduced. This represents the condition where DeepCompile is both enabled in the config and .compile() has been called.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -287,12 +287,15 @@ class DeepSpeedEngine(Module):
 
         self.pipeline_parallelism = isinstance(model, PipelineModule)
 
+        self._deepcompile_active = None
+        self._deepcompile_noop_warning_emitted = False
+
         # Configure distributed model
         self._configure_distributed_model(model)
 
-        if not self.is_deepcompile_enabled():
-            self.module_forward_pre_hook = self._create_module_forward_pre_hook()
-            self.module_forward_post_hook = self._create_module_forward_post_hook()
+        # These hooks should be disabled later if DeepCompile is not active.
+        self.module_forward_pre_hook = self._create_module_forward_pre_hook()
+        self.module_forward_post_hook = self._create_module_forward_post_hook()
 
         # needed for zero_to_fp32 weights reconstruction to remap nameless data to state_dict
         self.param_names = {param: name for name, param in model.named_parameters()}
@@ -521,7 +524,7 @@ class DeepSpeedEngine(Module):
     def destroy(self):
         if self.optimizer is not None and hasattr(self.optimizer, 'destroy'):
             self.optimizer.destroy()
-        if self.is_deepcompile_enabled():
+        if self.is_deepcompile_active():
             get_deepcompile_handle().cleanup()
         debug_clear_module_and_param_names()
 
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/engine.py
```diff
@@ -2155,7 +2158,14 @@ def forward(self, *inputs, **kwargs):
         if self.autotuning_profile_model_info():
             ma = get_ma_status()
 
-        if self.is_deepcompile_enabled() and hasattr(self, "launch_compile_passes"):
+        if (self.is_deepcompile_enabled() and not self.is_deepcompile_active() and not self.is_compiled
+                and not self._deepcompile_noop_warning_emitted):
```
eternalNight: There is a `warning_once` API from the logging module which allows us to get rid of that "warning_emiited" member. But that interface does not support the "print on rank 0" semantics yet.

I have an issue (#7578) related to logging APIs as well. Do we have any plan to revisit and unify the logging APIs? If not, I can take a look and submit an RFC PR. Of course that's a separate topic and does not block this one.

deepspeed/runtime/engine.py
```diff
@@ -287,12 +287,15 @@ def __init__(self,
 
         self.pipeline_parallelism = isinstance(model, PipelineModule)
 
+        self._deepcompile_active = None
```
eternalNight: Is it possible to simplify `_deepcompile_active` to a bool (instead of an `Optional[bool]`)? The only effect of initializing it with `None` is that `_set_deepcompile_active` will try to reinstall the hooks, and that is unnecessary because the hooks are already there.

By design, not calling `model.compile` should always mean DeepCompile is not activated. So it looks reasonable to me to make it a plain bool.

deepspeed/runtime/engine.py
tohtana: The logging APIs in DeepSpeed are not organized well. We would definitely appreciate if you can propose an improvement. 
I added `log_dist_once` and removed `_deepcompile_noop_warning_emitted` for this PR, but we can clean it once you have a better proposal.

deepspeed/runtime/engine.py
tohtana: You are right, I made it a simple bool.

deepspeed/runtime/engine.py
eternalNight: Sure. Let's get this merged first.

sfc-gh-truwase (approved):

sfc-gh-truwase:
@tohtana thanks for this PR. So, is the behavior now the following? 

1. No `deepspeed.compile()` in client code equals eager-mode
2. `deepspeed.compile()` in client code with `deepcompile==False` or missing in ds_config equals `torch.compile()`
3.  `deepspeed.compile()` in client code with `deepcompile==True` equals `DeepCompile`

tohtana:
Hi @sfc-gh-truwase, Yes it is correct.

PR #7604, opened 2025-09-28
Improve leaf module interface (enable via config, relax matching criteria, add document, etc.)

tohtana:
This PR improves the usability of the leaf module feature.

Here are the changes:
- Allow enabling the leaf module via both the DeepSpeed config and APIs.
- Relax matching criteria to support class-based matching.
- Support multiple ways of specifying the target module: class, class name (with or without package name), module name, or suffix.
- Add documentation to the training guide, including config snippets and explanations of default behavior.
- Add default classes (e.g., Mixtral, Qwen2/Qwen3) that automatically enable the leaf module feature. (Welcoming requests to add more classes)


docs/code-docs/source/training.rst
```diff
@@ -73,6 +73,85 @@ Each configuration works as follows:
 .. autofunction:: deepspeed.runtime.torch_autocast.has_autocast_dtype
 
 
+Configuring ZeRO Leaf Modules
+-----------------------------
+
+ZeRO-3 relies on module execution order to prefetch partitioned parameters. Models
```
sfc-gh-truwase:
```suggestion
ZeRO-3 relies on module execution order to gather partitioned parameters. Models
```

docs/code-docs/source/training.rst
```diff
@@ -73,6 +73,85 @@ Each configuration works as follows:
...
+that pick submodules dynamically (for example, MoE routers) can cause incorrect
```
sfc-gh-truwase:
```suggestion
that pick submodules dynamically (for example, MoE routers) can cause
```

docs/code-docs/source/training.rst
```diff
@@ -73,6 +73,85 @@ Each configuration works as follows:
...
+prefetch order unless you designate those subtrees as "leaf" modules. When a
```
sfc-gh-truwase:
```suggestion
data parallel ranks to gather different parameters causing the all-gather collective to deadlock.  To avoid this problem, you can designate the parent of dynamically activated submodules (e.g., MoE exports)  as a "leaf" module. When a
```

deepspeed/runtime/zero/leaf_module_config.py
```diff
@@ -0,0 +1,52 @@
...
+    def _validate_entries(self):
+        normalized_classes = [str(cls) for cls in self.classes]
+        normalized_names = [str(name) for name in self.names]
+        normalized_suffixes = [str(suffix) for suffix in self.name_suffixes]
+
+        deduped_classes = list(dict.fromkeys(normalized_classes))
+        deduped_names = list(dict.fromkeys(normalized_names))
+        deduped_suffixes = list(dict.fromkeys(normalized_suffixes))
+
+        object.__setattr__(self, "classes", deduped_classes)
```
sfc-gh-truwase, line 49: To avoid name collisions, can we prefix the attribute names with something like `_ds_`?

sfc-gh-truwase: Actually, never mind, I see that `self==DeepSpeedZeroLeafModuleConfig`

docs/code-docs/source/training.rst
tohtana: Fixed

docs/code-docs/source/training.rst
tohtana: Fixed

docs/code-docs/source/training.rst
tohtana: Fixed, and simplified the suggested text a bit

sfc-gh-truwase (approved):

docs/code-docs/source/training.rst
```diff
@@ -73,6 +73,84 @@ Each configuration works as follows:
...
+.. code-block:: json
+
+    {
+      "train_micro_batch_size_per_gpu": 1,
+      "zero_optimization": {
+        "stage": 3,
+        "leaf_module": {
+          "classes": ["my_package.layers.CustomMoEBlock"],
+          "names": ["transformer.layers.0.experts"],
+          "name_suffixes": ["experts"]
```
sfc-gh-sbekman, line 138: This part of the doc is somewhat confusing since it may make the reader believe all 3 entries are needed if they jumped here directly not reading the API section.

Too bad json doesn't allow comments. But perhaps the first paragraph could start with - "While the example shows all 3 ``leaf_module`` keys, typically you will probably use just one of these" ?

PR #7610, opened 2025-09-30
DeepCompile: Fix IPG bucket clearing

eternalNight:
PR #6993 replaces the flat IPG buffers with a dict maintaining type-indexed buckets. The member is also renamed from `_ipg_bucket_flat_buffer` to `ipg_buckets`.

Update the bucket clearing logic in `init_z3` accordingly.


tohtana (approved):
Thank you for the fix!

PR #7617, opened 2025-10-02
[wall_clock_breakdown] always log stats when enabled

stas00:
currently when main logger is WARN level, `wall_clock_breakdown: true` never logs - which is invalid as it disables this crucial at times functionality. Plus I think we have a disconnect somewhere since the recently added `--log_level` flag doesn't seem to change this logger's level.

The future plan is to be able to have different log levels for different modules, but for now just use `print` if `wall_clock_breakdown` is `True`, so this functionality is not log-level dependent.

`print` is also less noisy than the logger, because of the long prefix generated by the latter, which is of no value to the user since we print stats and not code related logs, so the printed results are easier to digest.

### DIFF
--- a/deepspeed/utils/logging.py
+++ b/deepspeed/utils/logging.py
@@ -83,7 +83,7 @@ def print_configuration(args, name):
         logger.info("  {} {} {}".format(arg, dots, getattr(args, arg)))
 
 
-def log_dist(message, ranks=None, level=logging.INFO):
+def log_dist(message, ranks=None, level=logging.INFO, use_logger=True):
     from deepspeed import comm as dist
     """Log message when one of following condition meets
 
@@ -94,6 +94,7 @@ def log_dist(message, ranks=None, level=logging.INFO):
         message (str)
         ranks (list)
         level (int)
+        use_logger (bool): if `False` ignores the log-levels and always prints
 
     """
     should_log = not dist.is_initialized()
@@ -104,7 +105,10 @@ def log_dist(message, ranks=None, level=logging.INFO):
         should_log = should_log or (my_rank in set(ranks))
     if should_log:
         final_message = "[Rank {}] {}".format(my_rank, message)
-        logger.log(level, final_message)
+        if use_logger:
+            logger.log(level, final_message)
+        else:
+            print(final_message)
 
 
 @functools.lru_cache(None)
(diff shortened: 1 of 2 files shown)


tohtana:
I think your point is that what we show with `wall_clock_breakdown: true` is not a log. That’s why we don’t want it to be affected by the log level. If that’s the case, how about introducing something like `print_dict`?

stas00:
> I think your point is that what we show with `wall_clock_breakdown: true` is not a log. That’s why we don’t want it to be affected by the log level. 

That's correct, Masahiro. It's a data dump.

> If that’s the case, how about introducing something like `print_dict`?

I'm not sure what you mean, please clarify or exemplify what you mean - perhaps in the context of this PR if it'd help?

sfc-gh-truwase (approved):

tohtana:
> I'm not sure what you mean, please clarify or exemplify what you mean - perhaps in the context of this PR if it'd help?

Sorry for the typo, I meant making a new function `print_dist` instead of `log_dist` to clarify it is not a log output.
If `log_dist`'s default behavior is using logger, it is not very clear what `use_logger==False` means.

stas00:
Thank you for clarifying, Masahiro. I'd be happy to improve this further.

Since `log_dist` has so much logic in it, replicating it is probably not a great idea, perhaps adding a wrapper then? so keeping the code as I suggested and then adding:

```
def print_dist(*args, **kwargs):
    """ same logic as log_dist but uses print instead of logger """
    log_dist(*args, **kwargs, use_logger==False)
```
(or using `partial`)

and using it at the calling point?

unless you object to the mod to `log_dist`, in which case we should rip out all of the `log_dist` logic except the last line into a new helper util and then add 2 wrappers instead, that will use this util?


```
def print_dist(...):
    string = helper(...)
    if string is not None: print(string)

def log_dist(...):
    string = helper(...)
    if string is not None: logger.log(string)
```
Either way works for me.

tohtana:
Thank you for your response, @stas00.

I think the first approach should be sufficient though the second approach is cleaner.
Also, I don't think this is urgent.

stas00:
If we don't do it now, we won't do it, let's complete this while it's "hot" - @sfc-gh-truwase, do you have a preference for the 2 options I proposed 2 comments up as a follow up to Masahiro's suggestion?

sfc-gh-truwase:
@stas00 let's go with the second approach, which I agree is cleaner.

stas00:
please have a look https://github.com/deepspeedai/DeepSpeed/pull/7621

PR #7621, opened 2025-10-03
add print_dist util

stas00:
a refactor follow up to https://github.com/deepspeedai/DeepSpeed/pull/7617 as suggested by @tohtana to create 2 independent utils, sharing the main logic via another util.


tohtana (approved):
Looks good to me, thank you @stas00!

PR #7623, opened 2025-10-07
Clarify document of leaf module config

tohtana:
Update document of leaf module config as suggested [here](https://github.com/deepspeedai/DeepSpeed/pull/7604#discussion_r2407483616).


stas00 (approved):
Thanks you, Masahiro

PR #7630, opened 2025-10-13
[XPU][CI] recover xpu-max1100 workflow

Liangliang-Ma:
Reduce some test scope to recover CI workflow.

### DIFF
--- a/deepspeed/ops/adam/__init__.py
+++ b/deepspeed/ops/adam/__init__.py
@@ -6,4 +6,4 @@
 from .cpu_adam import DeepSpeedCPUAdam
 from .fused_adam import FusedAdam
 from .zenflow_cpu_adam import ZenFlowCPUAdam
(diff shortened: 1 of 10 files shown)


.github/workflows/xpu-max1100.yml
```diff
@@ -72,12 +72,7 @@ jobs:
         export I_MPI_SHM=off
         pytest --verbose accelerator/*
         pytest --verbose autotuning/*
-        pytest --verbose checkpoint/test_reshape_checkpoint.py
```
sfc-gh-truwase, line 75: @Liangliang-Ma is there plan to restore these UTs? Or do they need adaption to work correctly on XPU?

Liangliang-Ma: These UTs should be restored. No need adaption for ut codes.

PR #7634, opened 2025-10-17
Take **kwargs in __init__ of DeepSpeedZeroOptimizer subclasses

eternalNight:
DeepSpeedZeroOptimizer provides a rich, evolving list of keyword arguments. It is tedious and error-prone to list all of them in its subclasses. As an example, the recent introduction of zenflow_config in the middle of that list has caused unit test failures (e.g. https://github.com/deepspeedai/DeepSpeed/actions/runs/18560070656/job/52906645682?pr=7633)

Convert the keyword argument list in DeepSpeedZeroOptimizer subclasses to **kwargs for the consistency of configurable items and their default values. Passing an unknown parameter to such subclasses will now raise an error on their call to DeepSpeedZeroOptimizer.__init__() instead of their own __init__(). It still ensures that typo in such parameters fail early.

### DIFF
--- a/deepspeed/runtime/superoffload/superoffload_stage3.py
+++ b/deepspeed/runtime/superoffload/superoffload_stage3.py
@@ -3,7 +3,6 @@
 
 # DeepSpeed Team
 
-import sys
 import time
 import torch
 from typing import List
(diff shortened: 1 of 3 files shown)


eternalNight:
It looks the CI still used the old code base (see https://github.com/deepspeedai/DeepSpeed/actions/runs/18582440763/job/52979937206?pr=7634#step:5:473).

@sfc-gh-truwase  Any idea what's going wrong here?

tohtana:
Hi @eternalNight,
We have been using the test code on the master branch since we migrated to Modal. There are both technical and security reasons. (But I thought only the tests are taken from `master` and the PR branch is used for `deepspeed`). 
For this one, I confirmed it pass the test on my env. So let me approve and merge it.

Issue #7632, opened 2025-10-15
[BUG]try to train lora but the lora parameter seem not be updated, how to check the reason


523997931:
**Describe the bug**

<img width="1232" height="674" alt="Image" src="https://github.com/user-attachments/assets/63478c4b-048f-4d00-ba5c-6f87f99cc929" />
here is my train code to check, I check the parameter value, and find it not change during my train, it is always 0.0, I don't know how to debug for the problem

zhangjingxian1998:
i meet same issue, and find this. but i don't know whether right or not
change the ```return self.overflow``` position
https://github.com/deepspeedai/DeepSpeed/blob/69e03e52d0ebc567d34a163e925899751f7dbcb8/deepspeed/runtime/fp16/fused_optimizer.py#L255-L289
to
```
     def step(self, closure=None): 
         """ 
         Not supporting closure. 
         """ 
  
         if self.fused_adam_legacy: 
             return self.step_fused_adam() 
  
         # First determine if there is overflow. 
         if self.timers: 
             self.timers(OVERFLOW_CHECK_TIMER).start() 
         fp16_params = [] 
         for i, group in enumerate(self.fp16_groups): 
             fp16_params.extend([p for p in group if p.grad is not None]) 
         self.overflow = self.overflow_checker.has_overflow(fp16_params) 
         if self.timers: 
             self.timers(OVERFLOW_CHECK_TIMER).stop() 
  
         if self.use_grad_scaling: 
             prev_scale = self.cur_scale 
             self._update_scale(self.overflow) 
             if self.overflow: 
                 if self.verbose: 
                     log_dist( 
                         "Overflow detected. Skipping step. Attempted loss " 
                         f"scale: {prev_scale}, reducing to {self.cur_scale}", 
                         ranks=[0]) 
                 # Clear gradients 
                 for i, group in enumerate(self.fp16_groups): 
                     for p in group: 
                         p.grad = None 
                 return self.overflow 
  
             if self.timers: 
                 self.timers.log(OVERFLOW_TIMERS) 
             
```

523997931:
I use the register_full_backward_hook to print the module name and grad input and output, and I print the result like
[Backward Hook] 
 grad_input: [None]
 grad_output: [torch.Size([1, 16, 21, 64, 100])]
[Backward Hook] 41
 grad_input: [None]
 grad_output: [torch.Size([1, 16, 21, 64, 100])]
[Backward Hook] 41.head.head
[...]

PR #7645, opened 2025-10-23
Fix misplaced overflow handling return in fused_optimizer.py

rraminen:
This PR fixes an issue in deepspeed/runtime/fp16/fused_optimizer.py where the gradient overflow handling logic incorrectly exited the function too early, resulting in wrong forward pass and loss calculations in certain FP16 training scenarios.

The `return self.overflow` and `self.timers.log(OVERFLOW_TIMERS)` calls are now correctly moved inside the `if self.overflow:` block so that the function only returns early when an actual overflow is detected.

Origin of the error: https://github.com/deepspeedai/DeepSpeed/commit/889f0ead27435cd7755a73a9ba27e0913ab3c548

cc: @jithunnair-amd

### DIFF
--- a/deepspeed/runtime/sequence_parallel/ulysses_sp.py
+++ b/deepspeed/runtime/sequence_parallel/ulysses_sp.py
@@ -32,6 +32,7 @@ https://github.com/snowflakedb/ArcticTraining/blob/main/projects/sequence-parall
 from collections import defaultdict
 from deepspeed.runtime.utils import see_memory_usage
 from deepspeed.sequence.layer import _DimZeroAllToAll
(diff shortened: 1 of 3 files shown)


eternalNight:
Thanks! This should fix #7632.

rraminen:
Hi @tjruwase, could you please help in reviewing this PR?

tjruwase (approved):

rraminen:
Hi @tohtana, could you please help in reviewing this PR?

tohtana (approved):
@rraminen Sorry for my late response! Approved it.

PR #7656, opened 2025-10-28
ALST/UlyssesSP: more intuitive API wrt variable seqlen

stas00:
As I was integrating ALST/Ulysses SP into HF Accelerate/Trainer I noticed that the initial `UlyssesSPAttentionHF.register_with_transformers` API was a bit inflexible/confusing wrt variable seqlen. 

This PR deprecates the misleading `max_length` arg name, replaces it with `seq_length` and makes the latter optional if `seq_length_is_variable` is True.

Updated tests and docs.


tjruwase (approved):

stas00:
Thank you, Tunji!

PR #7661, opened 2025-10-31
Add sequential allgather optimization for ZeRO-3

aeeeeeep:
* Perform allgather operations on parameters sequentially instead of coalescing them into large buckets.
* Significantly reduce peak memory usage in high memory pressure scenarios.
* Improve performance by minimizing temporary buffer requirements.
* The behavior is enabled via a new boolean flag under the section 
```json 
"zero_optimization": {
  "stage3_allgather_sequential": true 
 }
```
* By default the optimization is not enabled.

### DIFF
--- a/deepspeed/runtime/zero/config.py
+++ b/deepspeed/runtime/zero/config.py
@@ -26,7 +26,7 @@ ZeRO optimization should be enabled as:
     "stage3_module_granularity_threshold": 0,
     "allgather_partitions": [true|false],
     "use_multi_rank_bucket_allreduce": [true|false],
-    "stage3_allgather_single_param": [true|false],
+    "stage3_allgather_sequential": [true|false],
     "allgather_bucket_size": 500000000,
     "reduce_scatter": [true|false],
     "contiguous_gradients" : [true|false]
@@ -269,10 +269,10 @@ class DeepSpeedZeroConfig(DeepSpeedConfigModel):
     the overhead of concatenation and slicing on the host.
     """
 
-    allgather_single_param: bool = Field(default=False, alias="stage3_allgather_single_param")
+    allgather_sequential: bool = Field(default=False, alias="stage3_allgather_sequential")
     """
-    Enables allgather on individual parameters directly, bypassing the standard parameter bucketing mechanism
-    in stage3. This significantly reduces data copy overhead (eliminating copy-to-bucket operations)
+    Performs allgather on individual parameters sequentially, bypassing the standard parameter bucketing
+    mechanism in stage3. This significantly reduces data copy overhead (eliminating copy-to-bucket operations)
     and lowers peak memory usage by avoiding the allocation of large temporary flattening buffers.
     Recommended for scenarios with high memory pressure.
     """
--- a/deepspeed/runtime/zero/partition_parameters.py
+++ b/deepspeed/runtime/zero/partition_parameters.py
@@ -1097,10 +1097,10 @@ class Init(InsertPostInitMethodToModuleSubClasses):
 
         self.use_all_reduce_for_fetch_params = get_config_default(DeepSpeedZeroConfig,
                                                                   "use_all_reduce_for_fetch_params")
-        self.allgather_single_param = get_config_default(DeepSpeedZeroConfig, "allgather_single_param")
+        self.allgather_sequential = get_config_default(DeepSpeedZeroConfig, "allgather_sequential")
         if _ds_config is not None:
             self.use_all_reduce_for_fetch_params = _ds_config.zero_config.use_all_reduce_for_fetch_params
-            self.allgather_single_param = _ds_config.zero_config.allgather_single_param
+            self.allgather_sequential = _ds_config.zero_config.allgather_sequential
 
     def _update_persist_config(self, ds_config):
         Init.apply_param_persistence = True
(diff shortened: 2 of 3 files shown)


sfc-gh-truwase:
@aeeeeeep thanks for this contribution. Are you able to share some data showing the benefits of this optimization?

aeeeeeep:
> @aeeeeeep thanks for this contribution. Are you able to share some data showing the benefits of this optimization?

Thanks for your feedback! I’ll share detailed data within the next few days.

aeeeeeep:
The optimization shows measurable benefits only on a specific accelerator (under NDA), where hardware/driver overhead for memory allocation is significantly higher. This extended allocation latency indirectly delays memory release on communication streams.

While this PR's reduction of allocation operations is theoretically sound, the practical benefit appears hardware-dependent and isn't observable on NVIDIA platforms.

sfc-gh-truwase:
> While this PR's reduction of allocation operations is theoretically sound, the practical benefit appears hardware-dependent and isn't observable on NVIDIA platforms.

Thanks for the explanation. Can you confirm there is no regression on NVIDIA platform?

aeeeeeep:
Confirmed no regression on NVIDIA platform from my tests a few weeks ago (both performance and accuracy).
I’ll re-validate with the latest code tomorrow (~20 h from now) and update code immediately if anything changes.

aeeeeeep:
Apologies for the long delay. I finally found some time to verify the changes.

I modified the relevant functions in `tests/unit/runtime/zero/test_zeropp.py` to validate the `allgather_single_param` feature as follows:

<details>
<summary>Click to see the test code</summary>

```python
    def get_config_dict(self, use_quantized_weights=False, use_hpz=False, use_allgather_single_param=False):
        """Generate the configuration dictionary for DeepSpeed."""
        config = {
            "zero_optimization": {
                "stage": 3,
                "stage3_max_reuse_distance": 0,
                "contiguous_gradients": True,
                "overlap_comm": True,
                "stage3_allgather_single_param": use_allgather_single_param,
            },
            "optimizer": {
                "type": "Adam",
                "params": {
                    "lr": 1e-5
                }
            },
            "fp16": {
                "enabled": True
            }
        }
        if use_quantized_weights:
            config["zero_optimization"]["zero_quantized_weights"] = True
        if use_hpz:
            config["zero_optimization"]["zero_hpz_partition_size"] = self.world_size // 2
        return config

    def test(self, model_name):
        torch.manual_seed(0)
        model, data_loader = self.load_and_prepare_data(model_name)
        zeropp_loss = self.get_loss(model, data_loader, self.get_config_dict(use_quantized_weights=True, use_hpz=True))
        model, data_loader = self.load_and_prepare_data(model_name)
        baseline_loss = self.get_loss(model, data_loader, self.get_config_dict())
        model, data_loader = self.load_and_prepare_data(model_name)
        allgather_single_loss = self.get_loss(model, data_loader, self.get_config_dict(use_allgather_single_param=True))
        model, data_loader = self.load_and_prepare_data(model_name)
        quantized_allgather_single_loss = self.get_loss(model, data_loader, self.get_config_dict(use_quantized_weights=True, use_allgather_single_param=True))
        model, data_loader = self.load_and_prepare_data(model_name)
        zeropp_quantized_allgather_single_loss = self.get_loss(model, data_loader, self.get_config_dict(use_quantized_weights=True, use_hpz=True, use_allgather_single_param=True))

        # # Output and assert
        print(
            f"zeropp_loss={zeropp_loss}, baseline_loss={baseline_loss}, allgather_single_loss={allgather_single_loss}, quantized_allgather_single_loss={quantized_allgather_single_loss}, zeropp_quantized_allgather_single_loss={zeropp_quantized_allgather_single_loss}"
        )
        assert zeropp_loss < baseline_loss * 1.1, f"zeropp_loss={zeropp_loss}, baseline_loss={baseline_loss}"
        assert allgather_single_loss < baseline_loss * 1.1, f"allgather_single_loss={allgather_single_loss}, baseline_loss={baseline_loss}"
        assert quantized_allgather_single_loss < baseline_loss * 1.1, f"quantized_allgather_single_loss={quantized_allgather_single_loss}, baseline_loss={baseline_loss}"
        assert zeropp_quantized_allgather_single_loss < baseline_loss * 1.1, f"zeropp_quantized_allgather_single_loss={zeropp_quantized_allgather_single_loss}, baseline_loss={baseline_loss}"
```
</details>

run

```bash
python3 -m pytest -m nightly -sv tests/unit/runtime/zero/test_zeropp.py::TestZeroPPConvergence::test
```

passed result

```text
zeropp_loss = 0.511968202702701
baseline_loss = 0.49972660798579455
allgather_single_loss = 0.5007278224267065
quantized_allgather_single_loss = 0.5084740456752479
zeropp_quantized_allgather_single_loss = 0.5130055318772793
```

Additionally, I ran a training test with the `qwen3-0.6b` model for 100 steps at 4 gpus. The loss comparison plot is shown below:

image


The validation above confirms that this feature introduces no accuracy issues. Regarding performance benefits, they are observed on specific hardware configurations, but I cannot provide concrete metrics due to NDA.

deepspeed/runtime/zero/config.py
```diff
@@ -26,6 +26,7 @@
     "stage3_module_granularity_threshold": 0,
     "allgather_partitions": [true|false],
     "use_multi_rank_bucket_allreduce": [true|false],
+    "stage3_allgather_single_param": [true|false],
```
sfc-gh-truwase: I think a more descriptive naming is something like `stage3_allgather_sequential` or `stage3_allgather_serialize`, because it is not allgathering only one param but  one-at-a-time.

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -1306,79 +1308,84 @@ def all_gather_coalesced(params: Iterable[Parameter],
...
-                        return AllGatherHandle(handles,
-                                               param,
-                                               param_buffer=param_buffer,
-                                               original_dtype=original_dtype)
-                else:
-                    if hasattr(param_ds_tensor, "ds_quant_scale"):
-                        scales = param_ds_tensor.ds_quant_scale
-                        quantized_param = param_ds_tensor.data
+                handles = []
+                for param in params:
```
sfc-gh-truwase: Can you please refactor the previous single-param logic into a separate function and loop over it here? That would greatly improve code quality and maintenance. Thanks!

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -1991,78 +1998,125 @@ def _allgather_params(self, param_list, hierarchy=0):
...
             return
 
-        partition_size = sum([param.ds_tensor.ds_numel for param in param_list])
-
-        tensor_size = partition_size * self.num_partitions
-        flat_tensor = torch.empty(tensor_size, dtype=param_list[0].ds_tensor.dtype, device=self.local_device)
-        partitions = []
-        for i in range(self.num_partitions):
-            start = partition_size * i
+        if self.allgather_single_param:
```
sfc-gh-truwase: It seems the same refactoring opportunity exists here:
1. Refactor current logic into a helper that processes list of params
2. Use helper appropriatly for both cases

```python
     if self.allgather_single_param:
         for param in param_list:
              _all_gather_params_helper([param], ...)
     else:
           _all_gather_params_helper(param_list, ...)
```

aeeeeeep:
Thanks for the review. I'm currently addressing the feedback.

deepspeed/runtime/zero/config.py
aeeeeeep: I agree that sequential is a much more accurate term to describe the "one-at-a-time" behavior.
I have updated the naming convention to match your proposal.

deepspeed/runtime/zero/partition_parameters.py
aeeeeeep: To improve readability and maintainability, I have split the logic inside `all_gather_coalesced` into two distinct internal helper functions:
1. `_all_gather_sequential`: Handles the logic when allgather_sequential is enabled or there is only a single parameter.
2. `_all_gather_coalesced`: Handles the original bucketed/coalesced all-gather logic.
The main `all_gather_coalesced` function now serves as a clean dispatcher that routes to the appropriate implementation based on the configuration and parameter count.

deepspeed/runtime/zero/partition_parameters.py
aeeeeeep: I looked into `_allgather_params` and noticed it was already serving as the handler for the single-parameter case (previously called only when `len(all_gather_list) == 1`).

To clarify its purpose and align it with the new optimization features, I have renamed `_allgather_params` to `_allgather_params_sequential`. This allows us to directly reuse this existing logic for the sequential all-gather path without code duplication, effectively achieving the pattern you suggested.

PR #7664, opened 2025-11-03
UlyssesSP: TiledMLP doc - recomputes forward twice

stas00:
Make it very clear that `TiledMLP`'s memory saving has a cost of recomputing forward.


stas00:
Thank you, Masahiro

kidlj:
I just noticed that there's a typo in this commit: `occurs trice`.

stas00:
that's not a typo, it does occur trice.

1. normal forward
2. activation checkpointing forward
3. backward's internal forward-like recomputation.

in a normal non-tiled computation module only 1+2 occur.

This is the price to save memory. So 25% more computation.

PR #7665, opened 2025-11-03
PyTorch-compatible backward API

tohtana:
Currently DeepSpeed's backward API has more constraints compared to PyTorch's normal backward API.
Here is the usage as described in the documentation:
```python
    loss = model_engine(batch)
    model_engine.backward(loss)
```

In this example, 
1. Only accepts a (scalar) loss value
1. Need to call engine's backward API

In contrast, in standard PyTorch, you can do:
```python
    output = model(batch)
    output.backward(out_grad)
```

There are several use cases that rely on this flexibility. For example, combining multiple models or using loss functions defined separately from the main model.

If you attempt the same pattern with a DeepSpeed engine, some preprocessing and postprocessing steps will be silently skipped, which can lead to incorrect results.

The [document](https://deepspeed.readthedocs.io/en/latest/training.html#jointly-training-models-with-shared-loss) explains we can call `_backward_epilogue` manually (possibly `backward_prologue` as well). However, it's easy for users to miss these calls, and passing a non-scalar gradient is still not supported.

This PR introduces the same `.backward()` behavior as PyTorch, allowing .backward() to be called directly on tensors and supporting non-scalar outputs.

To implement post-backward hooks, we had to use some torch internal APIs. See [comments](https://github.com/deepspeedai/DeepSpeed/blob/73f7ff1aab9d1387eb7dd4eca7453a25024533f4/deepspeed/runtime/engine.py#L424) for more details. When the internal APIs are not available, DeepSpeed engine only accepts the traditional way `model_engine.backward(loss)`.

### DIFF
--- a/.github/workflows/xpu-max1100.yml
+++ b/.github/workflows/xpu-max1100.yml
@@ -51,7 +51,7 @@ jobs:
         pip install --upgrade pip
         pip install py-cpuinfo
         pip install torch==2.7.0 torchvision==0.22.0 torchaudio==2.7.0 --index-url https://download.pytorch.org/whl/xpu
-        pip install intel-extension-for-pytorch==2.7.10+xpu oneccl_bind_pt==2.7.0+xpu --extra-index-url https://pytorch-extension.intel.com/release-whl/stable/xpu/us
+        pip install intel-extension-for-pytorch==2.7.10+xpu oneccl_bind_pt==2.7.0+xpu --extra-index-url https://pytorch-extension.intel.com/release-whl/stable/xpu/us --trusted-host pytorch-extension.intel.com
         pip install .[dev,autotuning]
 
     - name: Check container state


sfc-gh-truwase:
@tohtana, this is a very exciting usability improvement. Please remember to update the documentation.

tohtana:
@sfc-gh-truwase I think this PR is now ready for review, though the latest change on HF transformer causes an error with `test_zero_nesting_init.py::TestNestedParallelInit::test_nested_parallel_init`.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1230,25 +1231,36 @@ def overlapping_partition_gradients_reduce_epilogue(self):
...
+
         for i, param_group in enumerate(self.fp16_groups):
             for param in param_group:
-                if param.requires_grad:
-                    #print_rank_0(f" Before all gather {param.device}, {param.shape}")
-                    print_rank_0(f"Before all gather {param.device}, {param.shape}", force=False)
+                if param.requires_grad and not z3_leaf_parameter(param):
+                    non_leaf_params_requiring_grad.append(param)
+
+        leaf_module_count = len(self.leaf_parameters)
```
sfc-gh-truwase, line 1243: Is `self.leaf_parameters` not an empty dict at this point?

deepspeed/runtime/utils.py
```diff
@@ -1188,3 +1188,248 @@ def compare_tensors_in_structures(inputs1: Union[List, Dict], inputs2: Union[Lis
...
+    """
+    assert check_internal_apis_for_count_used_parameters(), (
+        "count_used_parameters_in_backward requires internal PyTorch APIs that are not available "
+        "in this PyTorch build.")
+
+    from torch.autograd.graph import _get_grad_fn_or_grad_acc
+    if torch._C._current_graph_task_id() == -1:
+        raise RuntimeError("count_used_parameters_in_backward must be called during backward execution")
+
+    grad_nodes = []
```
sfc-gh-truwase: I am curious if `grad_nodes` and `seen_nodes` can be combined and simplified into a dict such as: 

`grad_nodes[id(grad_fn)] = grad_fn`

deepspeed/runtime/utils.py
```diff
@@ -1188,3 +1188,248 @@ def compare_tensors_in_structures(inputs1: Union[List, Dict], inputs2: Union[Lis
...
+        if grad_fn_id in seen_nodes:
+            continue
+
+        seen_nodes.add(grad_fn_id)
+        grad_nodes.append(grad_fn)
+
+    if not grad_nodes:
+        return 0
+
+    participating = sum(map(torch._C._will_engine_execute_node, grad_nodes))
```
sfc-gh-truwase: For my education, can you please explain
1. What `torch._C._will_engine_execute_node` does?
2. Why not just return `len(grad_nodes)`

deepspeed/runtime/zero/stage3.py
```diff
@@ -2316,14 +2338,8 @@ def _has_inf_or_nan(x, j=None):
...
-    @instrument_w_nvtx
-    def backward(self, loss, retain_graph=False):
-        """
-        :attr:`backward` performs the following steps:
-
-        1. fp32_loss = loss.float()
-        2. scaled_loss = fp32_loss*loss_scale
-        3. scaled_loss.backward(), which accumulates scaled gradients into the ``.grad`` attributes of the model's fp16 leaves
+    def backward_prologue(self):
+        """Currently only scales the loss value. Defined for consistency of naming.
```
sfc-gh-truwase: Is the comment accurate, as I don't see any loss scaling logic in this function?

tohtana: Thank you, removed it.

deepspeed/runtime/utils.py
tohtana: Good catch! We are just counting effective nodes, so we can simplify it as you suggested.

deepspeed/runtime/utils.py
tohtana: 1. When a backward pass starts, the autograd engine makes a plan to visit nodes. `_will_engine_execute_node` will returns if it will visit a given grad_fn or parameter (with `requires_grad==True`). 
2. We want to trigger post-backward process after all hooks finish. Only `torch.autograd.graph.register_multi_grad_hook` can do it, but it doesn't work with ZeRO3. So we use a counter that counts calls of our post-backward hook. Howver, the hook is called when the autograd engine visits the parameter. The autograd engine may not visit all `grad_nodes`, so we can't use `len(grad_nodes)` for the counter.

deepspeed/runtime/zero/stage3.py
tohtana: Thank you, fixed it and added a test to validate the combination with leaf module.

PR #7676, opened 2025-11-06
[modal ci] fixes

stas00:
1. `modal-accelerate` needs now `uv` installed explicitly since the image change to 2025 one.
2. moved accelerate repo cloning into the job, since the original way was incorrect as it was caching some accelerate version and not updating it.
3. annotated that how to actually test the ci work when changing the workflow as `pull_request_target` will not run the updated .py+.yaml files.

### DIFF
--- a/.github/workflows/modal-accelerate.yml
+++ b/.github/workflows/modal-accelerate.yml
@@ -20,9 +20,12 @@ on:
     branches:
       - master
 
-  # you have to switch to `pull_request` if you need to change the CI job's python script, otherwise GH will use a master version of the CI files, ignoring the modifications in the PR
-  pull_request:
-  #pull_request_target:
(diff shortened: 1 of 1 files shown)


.github/workflows/modal-accelerate.yml
```diff
@@ -20,7 +20,9 @@ on:
     branches:
       - master
 
-  pull_request_target:
+  # you have to switch to `pull_request` if you need to change the CI job's python script, otherwise GH will use a master version of the CI files, ignoring the modifications in the PR
+  pull_request:
+  #pull_request_target:
```
stas00: this will be reverted before merge

stas00:
@tjruwase, please have a look - this now works.

stas00:
so I had to force merge, since _target can't pass, but the next PR should work after rebasing.

PR #7696, opened 2025-11-17
Add Qwen2.5 to AutoTP model list

delock:
The AutoTP supported models does not include Qwen2.5, which is already supported.  Update the document.
(https://www.deepspeed.ai/tutorials/automatic-tensor-parallelism/#supported-models)


xylian86:
@delock Qwen3 is the latest generation in Qwen series. Does AutoTP support Qwen3 (https://huggingface.co/Qwen/Qwen3-8B)? If yes, it would be great to include it as well.

delock:
@xylian86 Just checked AutoTP works on Qwen3 as well, will update doc, thanks!

PR #7700, opened 2025-11-20
Low-precision master params/grads/optimizer states

tohtana:
DeepSpeed optimizer always creates fp32 master params/gradients/optimizer states. 
However, we sometimes want to keep them lower precision given [torch.autocast support](https://deepspeed.readthedocs.io/en/latest/training.html#mixed-precision-training).
This PR allows lower precision master params/grads/optimizer states with bf16/fp16 enabled.

DeepSpeed currently accepts `fp16_master_weights_and_gradients` option under `fp16` section (not documented) with ZeRO1/2. This PR extends this for bf16 and also ZeRO3.

In `bf16` section, we can have new items `bf16_master_weights_and_grads` and `bf16_optimizer_states`.
Similary to `fp16_master_weights_and_grads`, `bf16_master_weights_and_grads` keeps master parameters in bf16. `bf16_optimizer_states` keeps optimizer states also in bf16. Here is an example configuration:

```json
        "bf16": {
            "enabled": true,
            "bf16_master_weights_and_grads": true,
            "bf16_optimizer_states": true
        }
```

Note that `bf16_master_weights_and_grads==True` and `bf16_optimizer_states==False` is supported only with cpu offloading. Also, we don't have `fp16_optimizer_states` as it won't be practical. More details are described in [`config-json.md`](https://github.com/tohtana/DeepSpeed/blob/88e0bbdfba89c4712d815980ddb28353d6da5b2e/docs/_pages/config-json.md)

Previously, `torch.autocast` support (`torch_autocast` section in config) was not compatible with `bf16` `fp16` enabled, but we now accept the combination.

This PR also adds some test cases for the configurations as well as the combination with `torch.autocast`.

### DIFF
--- a/deepspeed/runtime/swap_tensor/optimizer_utils.py
+++ b/deepspeed/runtime/swap_tensor/optimizer_utils.py
@@ -343,7 +343,8 @@ class OptimizerSwapper(object):
         for src, dst in zip(unswapped_srcs, unswapped_dsts):
             dst.data.copy_(src.data)
 
-        assert len(swap_tensors) == aio_handle.wait()
+        if len(swap_tensors) > 0:
+            assert len(swap_tensors) == aio_handle.wait()
 
         return swapped_fp16_tensors
 


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1890,6 +1895,12 @@ def get_flat_partition(self,
                            device,
                            param_group_idx,
                            return_tensor_list=False):
+        if len(tensor_list) == 0:
```
sfc-gh-truwase, line 1898: In what conditions will an empty list be passed in? Since this is an internal API, should we support or avoid such conditions?

docs/_pages/config-json.md
```diff
@@ -336,6 +371,55 @@ Example of <i>**scheduler**</i>
 | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
 | Any parameters outside of "enabled" will be passed to AMP's initialize call, see the API and descriptions here at the [apex.amp.initialize documentation](https://nvidia.github.io/apex/amp.html#apex.amp.initialize). | None    |
 
+### PyTorch Automatic Mixed Precision (torch.autocast) training options
```
sfc-gh-truwase, line 374: What about moving this discussion to [RTD](https://deepspeed.readthedocs.io/en/latest/training.html#mixed-precision-training) and merging with existing materials?

deepspeed/runtime/zero/stage_1_and_2.py
tohtana: This happens when parameters are small. Given numel=100 and padding=32, it will be 128 elements. If you have 8 ranks, the last one will be empty.

docs/_pages/config-json.md
tohtana: Simplified the description here and added a link to the sphinx document as they are almost duplicated.

bigyanghimire:
When doing ``` "bf16": {
            "enabled": true,
            "bf16_master_weights_and_grads": true,
            "bf16_optimizer_states": true
        }``` I get peak and allocated memory allocation lower than the baseline but I don't understand why the step time is higher with  `bf16_master_weights_and_grads` and `bf16_master_weights_and_grads` enabled. Do you have any idea? Thanks

Issue #7606, opened 2025-09-29
[BUG] The loss of the first step is too high when zero_hpz_partition_size is set.


zhengchenyu:
**Describe the bug**

When use ZERO++ and zero_hpz_partition_size is set, The loss of the first step or the first step after load checkpoint is too high.

**To Reproduce**

Qwen3 SFT train. stage3, one node, 8 gpu per node, zero_hpz_partition_size is set to 4.

**Screenshots**

Two experiment:

* `base`: not set zero_hpz_partition_size
* `bug`: set zero_hpz_partition_size to 4. And save checkpoint at step50, then load from this checkpoint

<img width="825" height="479" alt="Image" src="https://github.com/user-attachments/assets/fbcff041-d6cd-47c0-9ddd-1b4feb8d3923" />

**Docker context**
base on image: `ghcr.io/pytorch/pytorch:2.8.0-cuda12.8-cudnn9-devel`
python: 3.11.13
deepspeed: 0.16.9

zhengchenyu:
This issue is fixed by https://github.com/deepspeedai/DeepSpeed/pull/7707

PR #7707, opened 2025-11-27
Fix that ds_secondary_tensor may be dirty when loading the model or zero checkpoint for zero++.

zhengchenyu:
`ds_secondary_tensor` may be dirty during model loading or zero checkpointing for zero++.

* 1 Loading model

My task is transformers SFT. In the transformers code, initialization is done using code like the following:

```
with deepspeed.zero.Init():
    model = xxx
``` 

After this, `param` is already a ds tensor, meaning both `ds_tensor` and `ds_secondary_tensor` exist. Then `load_model` is called to reload the model.

```
with deepspeed.zero.GatheredParameters(params_to_gather, modifier_rank=0):
        module._load_from_state_dict(*args)

```
In `GatheredParameters.__exit__`, `params[0].partition` is called, and `has_been_updated` is set to `True`, indicating that data updates are needed. However, `_partition_param_sec` did not pass `has_been_updated`. This results in `ds_secondary_tensor` being dirty.

* 2 Loading zero checkpoint

The zero checkpoint is loaded into `fp16_partitioned_groups_flat`, meaning `param.ds_tensor` has been updated. However, the data in `param.ds_secondary_tensor` has not been updated. But the next `allgather` will use the dirty `param.ds_secondary_tensor`. 

A dirty `ds_secondary_tensor` can lead to abnormal loss. After calling `invalidate_secondary_tensor` in `_post_step`, the loss returns to normal. This is why loss anomaly only occurs during beginning steps.

Relate issue: #7606

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,7 @@
 
    <li>[2025/03] <a href="https://github.com/deepspeedai/DeepSpeed/blob/master/blogs/huggingface-tp/README.md">DeepSpeed AutoTP: Automatic Tensor Parallel Training of Hugging Face models</a></li>
 
-
-   <li>[2024/12] <a href="https://github.com/deepspeedai/DeepSpeed/blob/master/blogs/ulysses-offload/README.md">Ulysses-Offload: Democratizing Long Context LLM Training</a></li>
+<li>[2024/12] <a href="https://github.com/deepspeedai/DeepSpeed/blob/master/blogs/ulysses-offload/README.md">Ulysses-Offload: Democratizing Long Context LLM Training</a></li>
 
  </ul>
 </details>
--- a/tests/unit/runtime/half_precision/test_fp8.py
+++ b/tests/unit/runtime/half_precision/test_fp8.py
@@ -85,8 +85,8 @@ class TestFp8ComposabilityAcrossZero(DistributedTest):
         rtol, atol = 1e-07, 1e-05
 
         # Relax tolerance only for ROCm + FP16
-        if is_rocm_pytorch() and model_dtype == torch.float16:
-            rtol, atol = 3e-07, 3e-05
+        if is_rocm_pytorch() and base_datatype in ["fp16", "bf16"]:
+            rtol, atol = 1e-07, 1e-04
 
         # config
         zero_stage = [0, 1, 2, 3]


zhengchenyu:
This picture proves that the bug has been fixed. The experimental conditions for `fix` are exactly the same as those for `bug` in #7606. The only difference is that the executable code applies this pr.

截屏2025-11-27 11 10 55

sfc-gh-truwase:
@zhengchenyu thanks for PR. We are taking a look.

zhengchenyu:
The unit test `test_compile_zero.py::TestDeepCompile::test[True-1-dtype0]` failed. This seems to be unrelated to this PR. Despite multiple tests on my own server, I am still unable to reproduce the failure...

sfc-gh-truwase:
@zhengchenyu thanks for this PR. My opinion is that invalidating the secondary tensor is the correct solution both these cases. So I am aligned with your solution for `Loading zero checkpoints`. 

For `model loading` and other cases of `deepspeed.zero.GatheredParameters(...modifer_rank!=None)`, how about calling `invalidate_secondary_tensor()` [here](https://github.com/deepspeedai/DeepSpeed/blob/7f2f423257592725259a0950094dc9fe9d276a27/deepspeed/runtime/zero/partition_parameters.py#L2342)?

What do you think?

For context, `ds_secondary_tensor` is meant to be a cache of gathered params from a previous forward pass. Therefore for safety, it should be invalidated when model weights change in anyway.

zhengchenyu:
@sfc-gh-truwase 
I think your suggestion is correct. In fact, I initially solved the problem by using `invalidate_secondary_tensor`.

However, I found the root cause was that the `_partition` function didn't pass the `has_been_updated` parameter when calling `self._partition_param_sec`, causing `ds_secondary_tensor` to become dirty. Since `broadcast` was used to ensure each parameter was correct, it's safe not to call `invalidate_secondary_tensor` here.

In fact, we have two solutions to this problem:

* (1) Invalidate the secondary tensor when model weights change anyway.

Your solution maintains consistent logic: *if weight changes, invalidate the secondary tensor.*

If that's the case, I think [here](https://github.com/deepspeedai/DeepSpeed/blob/7f2f423257592725259a0950094dc9fe9d276a27/deepspeed/runtime/zero/partition_parameters.py#L445) might also need to invalidate the secondary tensor.

However, the drawback of this approach is that it wastes a useful secondary tensor.

* (2) pass `has_been_updated` for `_partition_param_sec` in `_partition`.

This approach avoids wasting the `secondary tensor` when the parameter has already been broadcast.

In fact, I think both are ok, but I prefer (2). However, if you think we need to maintain consistent logic, I would change it to (1).

zhengchenyu:
And do you mean the case `deepspeed.zero.GatheredParameters(...modifer_rank=None)`? If that's the case, I agree add invalidate_secondary_tensor [here](https://github.com/deepspeedai/DeepSpeed/blob/7f2f423257592725259a0950094dc9fe9d276a27/deepspeed/runtime/zero/partition_parameters.py#L2342)。Thanks for your suggestion.
But If the model is updated but `modifier_rank` is not specified, it won't be broadcast, should invalidate secondary tensor. However, this is actually incorrect usage. And ds_tensor may also be inconsistent

sfc-gh-truwase:
> But If the model is updated but `modifier_rank` is not specified, it won't be broadcast, should invalidate secondary tensor. However, this is actually incorrect usage. And ds_tensor may also be inconsistent

Yes, this would be incorrect usage but it is not the API responsibility to detect such cases. So let's not worry about it.

sfc-gh-truwase:
> And do you mean the case `deepspeed.zero.GatheredParameters(...modifer_rank=None)`? If that's the case, I agree add invalidate_secondary_tensor [here](https://github.com/deepspeedai/DeepSpeed/blob/7f2f423257592725259a0950094dc9fe9d276a27/deepspeed/runtime/zero/partition_parameters.py#L2342)。Thanks for your suggestion.

Thanks for making the change.

sfc-gh-truwase:
@zhengchenyu I apologize I realize I gave you misleading information because I didn't read existing [GathereredParameters.__exit__()](https://github.com/deepspeedai/DeepSpeed/blob/7f2f423257592725259a0950094dc9fe9d276a27/deepspeed/runtime/zero/partition_parameters.py#L2338C1-L2339C1) carefully. 

In summary, your current PR is fine as is. I will approve to unblock for merging. 

I will explain a bit more below just for the records. 

1. For `if self.src_rank is None:` [case](https://github.com/deepspeedai/DeepSpeed/blob/7f2f423257592725259a0950094dc9fe9d276a27/deepspeed/runtime/zero/partition_parameters.py#L2341): The code here is actually correct and does not need `invalidate_secondary_tensor` call. This is because the user has specified no parameter changes and thus secondary tensor is clean. It is possible for users to misuse this API like you pointed out, but let's not worry about that. 
2. For `if self.src_rank != None:` [case](https://github.com/deepspeedai/DeepSpeed/blob/7f2f423257592725259a0950094dc9fe9d276a27/deepspeed/runtime/zero/partition_parameters.py#L2345-L2348): This is when secondary tensor becomes dirty. One option is to use `invalidate_secondary_tensor` or to propagate `has_been_updated` to `_partition_param_sec` which is your current solution. I agree with your current solution. 

Apologies for the confusion and extra work.

zhengchenyu:
Thanks very much for your review!

PR #7722, opened 2025-12-11
fix typo

stas00:
I guess this code path has never been exercised - fixing a varname typo

add `flake8` to the autoformatters set, which should have caught this?


stas00:
Thank you, Masahiro.

PR #7723, opened 2025-12-11
let allgather and alltoall execute in parallel when both attention and MOE used TP

taozhiwei:
let allgather and alltoall execute in parallel when both attention and MOE used TP.


delock:
Hi @taozhiwei, do you have setup info and performance data about this optimization?

logicjiecoding:
1

PR #7724, opened 2025-12-11
[Engine] Only scale gradients if scale_wrt_gas is True

kashif:
`_backward_prologue_per_tensor` checks if `scale_wrt_gas` is True and only scales if so.

For https://github.com/huggingface/accelerate/issues/3877

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -2359,10 +2359,8 @@ class DeepSpeedEngine(Module):
         self._stop_timers(self.engine_timers.backward_timers)
 
     def _backward_prologue_per_tensor(self, grad):
-        if grad is None:
-            return grad
         # Only scale gradients if scale_wrt_gas is True, consistent with backward() parameter
-        if self._scale_wrt_gas:
+        if grad is not None and self._scale_wrt_gas:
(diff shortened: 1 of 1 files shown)


stas00:
Just to get a better understanding - are you saying DS relies on the default value `scale_wrt_gas` being `True`, but Accelerate sets it explicitly to `False` in `backward`? Would you kindly point me to the corresponding code in Accelerate?

Your proposed fix is sound, but let's make sure we have a test covering it, so that someone won't unfix it down the road.

This appears to be the most suitable test to either modify or clone?
https://github.com/deepspeedai/DeepSpeed/blob/b00b75f05852e0791f1e2b9c1cc894cd690e2da4/tests/unit/v1/zero/test_zero_user_backward.py#L286

tohtana:
Sorry, @stas00 is right. It would be great if we could add a test.
When I modified the related code, I wasn't able to find the usage of the flag. We could avoid such a regression if we have a test.

tohtana:
Hi @kashif, I opened a [PR](https://github.com/kashif/DeepSpeed/pull/1) to add a test to this PR.
Can you review and merge if you think it is good?

kashif:
thanks @tohtana all good on my side!

Issue #7680, opened 2025-11-07
[REQUEST] Switch to and unify DeviceMesh management for DP/TP/PP/EP/SP


eternalNight:
# Problem Statement

Today we have multiple sets of ProcessGroup management in the codebase for different parallel scenarios, namely:

* `PipelineModule` uses `PipelineParallelGrid` for pipeline parallelism.
* AutoTP uses one set of APIs (whose names start with letters) from `utils.groups`.
* `MoE` uses another set of APIs (whose names start with `_`) from `utils.groups`.
* Ulysses SP uses a DeviceMesh created on `deepspeed.initialize`.

All of those implementations deal with the process group creation (either using `DeviceMesh` or customized logic) under complex combinations. It significantly increases maintenance burden as well as difficulty in introducing new parallelism techniques.

Besides, that divergence also causes user confusions. `deepspeed.initialize()` accepts two parameters, i.e., `mpu` and `mesh_param`.  `mpu` is for DP/TP (without PP/EP/SP), `mesh_param` for SP, and for PP/EP parallelism topology should be configured on model initialization. That does not provide a good experience for users who want to try out different configurations for best efficiency.

# Proposal

We would like to refactor those ProcessGroup management facilities so that:

* A unified `DeviceMesh` based module serves all parallelism strategies and their combinations.
* Let `DeviceMesh` instances create and manage ProcessGroups when possible. 
* Simplify the mesh topology configuration interface of `deepseed.initialize`.
* (Advanced) Allow different models in the same world to use different parallelism strategies. This is mainly for RL post-training, but need more careful feasibility investigation due to the global map design of ProcessGroup. See https://github.com/volcengine/verl/discussions/897 MCore pain point 1 for more information.
* (Advanced) Make it easy to extend ZeRO to support HSDP-style replicate + shared parallelism for asymmetric GPU clusters. The extension itself will be tracked in a separate issue.

[...]

PR #7726, opened 2025-12-15
Introduce Megatron-style parallel state management

eternalNight:
# Summary

This PR, authored by @hahaha3210 and @DayDreamer-Li, introduces `ParallelState`, a class that manages process groups for an arbitrary combination of parallel strategies including TP, EP, PP and DP.

As is discussed in #7680, the primary approach is borrowing the process group creation logic from https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/parallel_state.py but encapsulating the states (i.e., process groups, ranks and world sizes) into a class. This design enables the coexistence of multiple, independent parallelism configurations within a single process, which is particularly valuable in scenarios involving multiple models, such as in reinforcement learning (RL) workflows. Objects of `ParallelState` can be created prior to calls to `deepspeed.initialize` so that process groups are available to custom modules, such as `UlyssesSPAttentionHF`, at an early stage.

Compatibility of `ParallelState` and current process group management facilities (including `deepspeed.runtime.sequence_parallel.parallel_state_sp` and `deepspeed.utils.groups`) is tested by [test_mpu.py](https://gist.github.com/eternalNight/b76c72216b4be84832b615b76465396f).

# Usage

Basic usage (single global instance passed to `deepspeed.initialize`):

```
    from deepspeed import DeepSpeedConfig
    ds_config = DeepSpeedConfig("config.json")

    parallel_state = ps.initialize_parallel_state_from_config(config_dict)
    model_engine, optimizer, _, _ = deepspeed.initialize(
        model=model,
        config=config_dict,
        mpu=parallel_state,
    )
```

The `ParallelState` class is compatible with Megatron's `parallel_state` module and thus can also be passed as `mpu` to existing implementations of parallelism techniques.

# Opens

- [x] ~~Support for Ulysses SP is yet to be added.~~ Dimensions for Ulysses SP can now be set by passing `sequence_parallel_size` to `initialize_parallel_state_from_config`.
- [x] ~~Support creating a `ParallelState` from a config object rather than specifying different parallel dimensions explicitly.~~ `initialize_parallel_state_from_config` now accepts a `config` argument which can be either a `dict` or a config object.
- [x] ~~Are wrappers in `parallel_state_deepspeed.py` necessary? If so, is there a better way to implement more concisely its APIs sharing similar code patterns?~~ `parallel_state_deepspeed.py` is preserved and renamed as `parallel_state_wrappers.py` to provide helper methods (create from config, wrap with context, etc.) around `ParallelState` for easier use.
- [x] ~~Are GLOO process groups necessary for DeepSpeed? If not, we can strip them from the draft.~~ GLOO process groups are not created by default as `deepspeed.comm.new_group` does not accept a `backend` argument yet. The related code is kept so that we can easily bring it back when necessary.
- [x] ~~Tweaking NCCL options require `ProcessGroupNCCL.options` from `torch.distributed`, and that is not provided by `deepspeed.comm` today. Should we introduce that to `deepspeed.comm`, or make the format-checking script allowing that specific use of `torch.distributed`?~~ Leave it as future work.

# Future work

1. Configuration of process group options (only `ProcessGroupNCCL.Options` as of pytorch 2.9).
2. Add config items for parallelism dimensions not covered in json today.
3. Lazy creation of process groups. The Metatron-style manager creates all process groups for all potential combinations of parallelism in one shot. That won't impact much on GPU memory usage as recent NCCL allocates GPU memory on the first collective operation, not on communicator creation. But still it forks a significant amount of threads (~2000 on 8 GPUs) and may hit container PID limits. Possible optimizations include (1) not creating pgs for parallelism of size 1 at all, assuming they'll never be used; (2) creating pgs lazily (i.e., on first getter), assuming all ranks will call the same getter in the same order; (3) merge pgs among the same ranks, at the cost of synchronizing ops on them. That will be looked into in a follow-up PR.
4. While unit tests in this PR works properly, there're still code referring to specific mpu/parallel_state modules explicitly and thus not friendly to explicitly-passed `mpu`.

### DIFF
--- a/.github/workflows/nv-pre-compile-ops.yml
+++ b/.github/workflows/nv-pre-compile-ops.yml
@@ -23,10 +23,19 @@ jobs:
   unit-tests:
     runs-on: ubuntu-24.04
     container:
-      image: deepspeed/gh-builder:ubuntu1804-py38-torch1131-cu116
+      image: nvidia/cuda:12.6.3-devel-ubuntu22.04
 
     steps:
+        - name: Install system dependencies
+          run: |
+            apt-get update && apt-get install -y git python3 python3-pip libaio-dev ninja-build
+            ln -sf /usr/bin/python3 /usr/bin/python
+
         - uses: actions/checkout@v4
+
+        - name: Install PyTorch
+          run: |
+            pip install torch==2.10.0 --index-url https://download.pytorch.org/whl/cu126
 
         - name: environment
           run: |
@@ -36,7 +45,7 @@ jobs:
             #python -c "import torch; print('CUDA available:', torch.cuda.is_available())"
         - name: Compile DeepSpeed Ops
           run: |
-            DS_ACCELERATOR=cuda DS_ENABLE_NINJA=1 TORCH_CUDA_ARCH_LIST="7.0;7.5;8.0" DS_BUILD_OPS=1 DS_BUILD_SPARSE_ATTN=0 DS_BUILD_FP_QUANTIZER=0 DS_BUILD_CUTLASS_OPS=0 DS_BUILD_GDS=0 DS_BUILD_RAGGED_DEVICE_OPS=0 DS_BUILD_EVOFORMER_ATTN=0 DS_BUILD_DEEP_COMPILE=0 pip3 install .
+            DS_ACCELERATOR=cuda DS_ENABLE_NINJA=1 TORCH_CUDA_ARCH_LIST="7.0;7.5;8.0;8.6;8.9;9.0" DS_BUILD_OPS=1 DS_BUILD_SPARSE_ATTN=0 DS_BUILD_FP_QUANTIZER=0 DS_BUILD_CUTLASS_OPS=0 DS_BUILD_GDS=0 DS_BUILD_RAGGED_DEVICE_OPS=0 DS_BUILD_EVOFORMER_ATTN=0 DS_BUILD_DEEP_COMPILE=0 pip3 install .
         - name: DS Report
           run: |
-             ds_report
+             DS_ACCELERATOR=cuda ds_report
--- a/accelerator/abstract_accelerator.py
+++ b/accelerator/abstract_accelerator.py
@@ -179,10 +179,6 @@ class DeepSpeedAccelerator(ABC):
 
     # Misc
     @abc.abstractmethod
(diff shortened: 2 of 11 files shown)


sfc-gh-truwase:
@stas00 @tohtana @delock FYI

delock:
I like the idea of puting parallel dimension in a single place rather than relying on user reading deeply into document to figure out how to turn on each parallelism dimension.  I also agree with the open that the class can be created from a config object.  Does it make sense to have the config in config.json file, or a seperate config file is more flexible?

eternalNight:
> I like the idea of puting parallel dimension in a single place rather than relying on user reading deeply into document to figure out how to turn on each parallelism dimension. I also agree with the open that the class can be created from a config object. Does it make sense to have the config in config.json file, or a seperate config file is more flexible?

My rough idea is to reuse the current config.json which already provides dimensions of various parallel techniques. Moving parallelism-related configs to a separate file is such a huge change that can impose a big obstable to users who try to upgrade.

tohtana:
Hi @eternalNight, this is amazing!
Do you have a plan to add a document and unit tests? They will help users learn how to use this new style of parallelization configs.

eternalNight:
> Hi @eternalNight, this is amazing! Do you have a plan to add a document and unit tests? They will help users learn how to use this new style of parallelization configs.

Good point & will do.

sfc-gh-truwase:
@eternalNight, I am curious how this is going? Thanks

eternalNight:
> @eternalNight, I am curious how this is going? Thanks

We're just back from the lunar new year holidays and are polishing the documentations. Will update another version this week probably.

delock:
@eternalNight Can you sign-off to fix DCO error?  Thanks!

eternalNight:
> @eternalNight Can you sign-off to fix DCO error? Thanks!

All commits are signed on this branch. The DCO errors are due to the mismatch between author info in the commit metadata (which are less official due to local configs) and the signed-off-by (which uses the official names and mail addresses).

eternalNight:
@sfc-gh-truwase @delock @tohtana @stas00 Here's the final refreshment and rebase of the `ParallelState` class and helpers around it, with unit tests and documentation on usage. Please kindly review.

Future works are listed in the PR description.

PR #7732, opened 2025-12-18
replace moe checkpoint dp_world_size with seq_dp_world_size 

wukong1992:
Replace moe checkpoint dp_world_size with seq_dp_world_size to sup moe module with seq parallel.

### DIFF
--- a/csrc/aio/py_lib/py_ds_aio.cpp
+++ b/csrc/aio/py_lib/py_ds_aio.cpp
@@ -124,5 +124,6 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
 
         .def("wait",
              &deepspeed_aio_handle_t::wait,
-             "Wait for (ongoing) asynchronous operations to complete");
+             "Wait for (ongoing) asynchronous operations to complete",
+             py::call_guard<py::gil_scoped_release>());
 }


stas00:
Why are you proposing to do that?

If you need `seq_dp_world_size` on resume and for some reason it's not there then probably store it additionally while leaving `dp_world_size` alone, no?

wukong1992:
> Why are you proposing to do that?
> 
> If you need `seq_dp_world_size` on resume and for some reason it's not there then probably store it additionally while leaving `dp_world_size` alone, no?

cause `no moe` save checkpoint use this https://github.com/deepspeedai/DeepSpeed/blob/37ad0c0ecdfe019b1947c8e8fb50c6b03a354841/deepspeed/runtime/engine.py#L3908 
There is no need to redefine seq_dp_world_size here.

stas00:
Thank you for explaining, then yes, it checks out.

@sfc-gh-truwase, I know you weren't involved in DS-MoE - but shouldn't `_save_checkpoint` and `_save_moe_checkpoint` follow the exact same recipe wrt saving `state_dict` (the 

https://github.com/deepspeedai/DeepSpeed/blob/37ad0c0ecdfe019b1947c8e8fb50c6b03a354841/deepspeed/runtime/engine.py#L3891
https://github.com/deepspeedai/DeepSpeed/blob/37ad0c0ecdfe019b1947c8e8fb50c6b03a354841/deepspeed/runtime/engine.py#L3822

only `model_state_dict` is what's different - perhaps abstracting this last bit into a helper util and having both functions use it, with the only difference passing the `model_state_dict` that differs? the moe saving one also saves `num_experts` but it's not using it on load. I think it's diverging in other aspects since nobody bothered to keep in in sync.

PR #7735, opened 2025-12-18
Fix #7733: Replace torch.sqrt with math.sqrt in scale_lr for sqrt method

### LINKED ISSUE #7733: [BUG] `scale_lr` fails for `lr_scaling_method="sqrt"` due to `torch.sqrt` on Python float
When using dynamic batching with `lr_scaling_method="sqrt"`, training fails with:

"TypeError: sqrt(): argument 'input' (position 1) must be Tensor, not float"

The error originates from `scale_lr()` [on line 159 in `deepspeed/runtime/data_pipeline/data_sampling/variable_batch_size_and_lr.py`](https://github.com/deepspeedai/DeepSpeed/blob/5373a88000d8017e269277662cd8e93a814c66e1/deepspeed/runtime/data_pipeline/data_sampling/variable_batch_size_and_lr.py#L159):

```python
return base_lr * torch.sqrt(batch_size / base_batch_size)
```

Here, `batch_size` and `base_batch_size` are Python integers, so `batch_size / base_batch_size` is a Python float. Passing this float to `torch.sqrt()` raises a TypeError. `torch.sqrt` should probably be replaced with `math.sqrt` here.

Rakshit-gen:
Fixes #7733 

When using lr_scaling_method='sqrt' with dynamic batching, the scale_lr function was failing with TypeError because torch.sqrt expects a Tensor but receives a Python float from batch_size/base_batch_size division.

Changed torch.sqrt to math.sqrt which correctly handles Python floats.

This fixes the issue where training would fail with: TypeError: sqrt(): argument 'input' (position 1) must be Tensor, not float


Rakshit-gen:
@tjruwase can we review this pr?

sfc-gh-truwase:
@Rakshit-gen this PR is already approved to merge. The current blockers are 
1. Formatting error: Follow this [guide](https://github.com/deepspeedai/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites) to fix. 
2.  Uncompleted DCO requirement. See image below and [here](https://github.com/apps/dco) for more details. 

image

PR #7736, opened 2025-12-18
Fix OnebitLamb NaN propagation with empty parameters

Rakshit-gen:
Fixed a critical issue where OnebitLamb optimizer would produce NaNs when optimizing models with empty parameters (numel=0).

The scaling factor calculation involved division by sqrt(numel), which resulted in 0.0/0.0 -> NaN for empty parameters. This NaN value propagated to the global scaling coefficient, corrupting the state of all other parameters.

Changed the denominator to use max(numel, 1) or conditional 1.0 to ensure safe division.


sfc-gh-truwase:
@Rakshit-gen thanks for the PR. 

By empty parameters, do you mean the model no parameters? If so, can you give more details of how this can happen as it is quite uncommon to me?

Rakshit-gen:
@sfc-gh-truwase Prevents division by zero when exp_avg has 0 elements. Empty parameter groups or filtered-out parameters can leave empty optimizer state tensors, causing crashes during momentum scale computation. The check avoids this edge case.

sfc-gh-truwase:
Thanks for the explanation. Unfortunately, that does not address my question of in what scenarios would a model have all its parameters filtered out. Or more precisely, why would a parameter group be empty? If parameters don't exist then optimizer state is not needed, and I would expect this code path not be executed at all. So I feel such situations out to be handled at the higher logical level. 

Alternatively, can you create a small example to demonstrating this problem? That would be something to add to our unit tests as well. Thanks

Rakshit-gen:
@sfc-gh-truwase 

```
import torch
import torch.nn as nn
import numpy as np

# Create a model with a normal parameter and an empty parameter
class ModelWithEmptyParam(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(10, 10)  # Normal parameter
        # Empty parameter (0 elements) - can happen in conditional architectures
        self.empty_param = nn.Parameter(torch.empty(0, 10))
    
    def forward(self, x):
        return self.linear(x)

model = ModelWithEmptyParam()

# Create parameter groups - one with normal params, one with empty param
param_groups = [
    {'params': [model.linear.weight, model.linear.bias], 'weight_decay': 0.01},
    {'params': [model.empty_param], 'weight_decay': 0.0}  # Empty parameter group
]

# Simulate the problematic calculation from OnebitLamb
# This is what happens in line 180-181 of lamb.py
for group in param_groups:
    for p in group['params']:
        exp_avg = torch.zeros_like(p.data)  # Simulated momentum
        
        # BEFORE FIX (causes NaN):
        # numel = torch.numel(exp_avg)  # = 0 for empty param
        # scale = torch.linalg.vector_norm(exp_avg) / np.sqrt(numel)
        # Result: 0.0 / sqrt(0) = 0.0 / 0.0 = NaN ❌
        
        # AFTER FIX (safe):
        numel = torch.numel(exp_avg)
        scale = torch.linalg.vector_norm(exp_avg) / np.sqrt(numel if numel > 0 else 1.0)
        # Result: 0.0 / sqrt(1.0) = 0.0 ✅
        
        print(f"Param shape: {p.shape}, numel: {numel}, scale: {scale}")
        assert not torch.isnan(scale), "Should not produce NaN!"
```

When torch.numel(exp_avg) == 0 (empty parameter), the original code does:
vector_norm(exp_avg) = 0.0 (norm of empty tensor)
np.sqrt(0) = 0.0
0.0 / 0.0 = NaN
The fix uses max(numel, 1) or numel if numel > 0 else 1.0, so:
0.0 / sqrt(1.0) = 0.0 (no NaN)
This NaN then propagates to united_scale and corrupts all other parameters' scaling_coeff values.

sfc-gh-truwase:
Thanks for sharing the example code, it is very helpful. I believe the correct solution is to avoid passing empty parameters to the optimizer since there is nothing to optimize. As a reference, this how frozen parameters (i.e., `.requires_grad==False`) are [handled](https://github.com/deepspeedai/DeepSpeed/blob/2bc16e216968198b1622a69fc88653151dec7161/deepspeed/runtime/zero/stage_1_and_2.py#L359-L364).

Rakshit-gen:
@sfc-gh-truwase Agreed. Updated the code to filter out empty parameters (numel() == 0) in OnebitLamb.__init__() before creating optimizer groups, following the same pattern as frozen parameters. Reverted the workaround and added a unit test. Changes are in the latest commit.

sfc-gh-truwase:
@Rakshit-gen thanks, I will review asap.

By the way, are you able to share how you are using OneBitLamb? I am unaware of many users. Thanks.

Rakshit-gen:
I used OneBitLamb when training a BERT model across multiple machines. Gradient synchronization between machines was slow and slowed training. OneBitLamb compresses the data sent between machines, so communication is faster.

Rakshit-gen:
@sfc-gh-truwase can we merge this now, the pr passed all the tests? are there any blockers?

sfc-gh-truwase:
@Rakshit-gen the only blocker is my review, and will do asap. Sorry for the delay,

Rakshit-gen:
No issues @sfc-gh-truwase do tell me if there is any other issue i can help with!

Rakshit-gen:
@sfc-gh-truwase any updates on this?

deepspeed/runtime/fp16/onebit/lamb.py
```diff
@@ -82,6 +82,31 @@ def __init__(self,
         if amsgrad:
             raise RuntimeError('1-bit Lamb does not support the AMSGrad variant.')
 
+        # Filter out empty parameters (numel == 0) - similar to how frozen params are handled
+        filtered_params = []
+        if isinstance(params, (list, tuple)) and len(params) > 0:
```
sfc-gh-truwase: This logic would be useful as a utility, so can you please refactor into a function?

Rakshit-gen: okay looking into it

sfc-gh-truwase: I also see a recursive structure here and an opportunity to generalize. Please see this checkpointing [utility](https://github.com/deepspeedai/DeepSpeed/blob/6d9c3dc0b8383832d181cd70784f4d7f22348d0d/deepspeed/checkpoint/utils.py#L41) for reference.

Rakshit-gen: Done. Refactored the filtering logic into filter_empty_parameters() in deepspeed/runtime/utils.py. This utility can now be reused by other optimizers that need to filter empty parameters.

Rakshit-gen:
do we need one more approval @sfc-gh-truwase

Rakshit-gen:
Thanks @sfc-gh-truwase 
Do inform me if I can work on any other issues!

sfc-gh-truwase:
@Rakshit-gen thanks for the contribution and willingness to do more. We can definitely use your help. However, things are slowing down now for the Holidays, so I will reach out again in the New Year.

Rakshit-gen:
@sfc-gh-truwase Thanks for the update — appreciate it. Totally understand the holiday slowdown. Looking forward to reconnecting in the New Year. Enjoy the holidays.

sfc-gh-truwase:
> @sfc-gh-truwase Thanks for the update — appreciate it. Totally understand the holiday slowdown. Looking forward to reconnecting in the New Year. Enjoy the holidays.

@Rakshit-gen Happy New Year! I wanted to follow up on this. If you are still interested in contribute, please consider joining our slack channel.  Thanks!

Rakshit-gen:
Hey @sfc-gh-truwase thanks, where can i join the slack channel from?

sfc-gh-truwase:
Try this link on GH home page: 
https://deepspeedworkspace.slack.com/join/shared_invite/zt-3a8pjd8dd-PCj2hMvR4Y2syPwVnjEoww#/shared-invite/email


image

Rakshit-gen:
Thanks a lot i will join the slack channel

PR #7742, opened 2025-12-20
Fix DecoupledCheckpointEngine deadlock and improve reliability

### LINKED ISSUE #7741: [BUG] DecoupledCheckpointEngine hangs indefinitely due to missing timeouts and process health checks
## Describe the bug

The `DecoupledCheckpointEngine` has multiple critical issues that can cause training jobs to hang indefinitely:

1. **No timeout on process join**: `cleanup()` calls `self.ckpt_process.join()` with no timeout - if the checkpoint process crashes or hangs, the training script hangs forever.

2. **No timeout on event wait**: `commit()` calls `self.save_event.wait()` with no timeout - if the checkpoint process dies before setting the event, training hangs forever.

3. **Assertion used for runtime validation**: Line 119 uses `assert info == self.commit_info` which is disabled with `python -O`, allowing silent data corruption in production.

4. **`__del__` calls blocking cleanup**: The destructor calls `cleanup()`, so any exception during training can cause the program to hang on exit.

5. **No process health checks**: No verification that `ckpt_process` is still alive before waiting on events or queues.

## To Reproduce

1. Enable decoupled checkpointing in DeepSpeed config
2. Start a training job that saves checkpoints
3. Kill or crash the checkpoint subprocess (e.g., OOM, signal, or disk full)
4. Observe that the main training process hangs indefinitely on `save_event.wait()` or `ckpt_process.join()`

Alternatively:

1. Run training with `python -O` (optimized mode)
2. Trigger a checkpoint with mismatched `commit_info`
3. The assert is skipped, potentially corrupting checkpoint state

## Expected behavior

- Checkpoint operations should have timeouts and fail gracefully with clear error messages
- Process health should be checked before blocking waits
- Runtime validation should use proper `if` statements, not assertions
- Training should be able to recover or exit cleanly if checkpointing fails

## Affected Code

**File**: `deepspeed/runtime/checkpoint_engine/decoupled_checkpoint_engine.py`

| Line | Issue | Severity |
|------|-------|----------|
| 119 | `assert info == self.commit_info` - disabled with `-O` | Critical |
| 123 | `self.save_event.wait()` - no timeout | Critical |
| 155 | `self.ckpt_process.join()` - no timeout | Critical |
| 99-100 | `__del__` calls `cleanup()` - can hang on destruction | High |

## System info

- **Affects**: All systems using `DecoupledCheckpointEngine`
- **OS**: Any
- **GPU**: Any
- **Python**: Any version

## Proposed Fix

```python
# Replace assert with proper validation
if info != self.commit_info:
    raise ValueError(f"Checkpoint commit info mismatch: expected {self.commit_info}, got {info}")

# Add timeout to event wait with process health check
TIMEOUT_SECONDS = 300  # 5 minutes
while not self.save_event.wait(timeout=10):
    if not self.ckpt_process.is_alive():
        raise RuntimeError("Checkpoint process died unexpectedly")

# Add timeout to process join
self.ckpt_process.join(timeout=TIMEOUT_SECONDS)
if self.ckpt_process.is_alive():
    self.ckpt_process.terminate()
    raise RuntimeError("Checkpoint process failed to exit within timeout")
```

Rakshit-gen:
## Summary
- Add timeouts to prevent indefinite hangs when checkpoint process crashes
- Replace assertion with proper runtime validation
- Add process health checks before blocking operations
- Implement graceful cleanup with escalating termination

Fixes #7741 

## Description

The `DecoupledCheckpointEngine` had several critical issues that could cause training jobs to hang indefinitely:

1. **No timeout on `save_event.wait()`**: If the checkpoint process died, training would hang forever waiting for an event that would never fire.

2. **No timeout on `ckpt_process.join()`**: If the process crashed or hung, cleanup would block indefinitely.

3. **Assertion used for runtime validation**: `assert info == self.commit_info` is disabled with `python -O`, allowing silent data corruption in production.

4. **`__del__` could hang**: The destructor called `cleanup()` which could block, causing program hang on exit.

## Changes

- Add `_wait_for_event_with_timeout()` that checks process health every 10 seconds while waiting
- Add `_check_process_alive()` helper to validate process state before blocking operations
- Replace `assert` with proper `if/raise ValueError` for commit info validation
- Add timeout to `join()` with escalating termination: `join()` → `terminate()` → `kill()`
- Wrap `__del__` in try/except and add `_cleanup_called` flag to prevent multiple cleanup calls
- Add health checks in `save()` and `commit()` before queue operations

## Test plan
- [ ] Verify normal checkpoint save/load still works
- [ ] Test behavior when checkpoint process is killed mid-save
- [ ] Verify timeout triggers after 5 minutes of no response
- [ ] Confirm graceful shutdown with Ctrl+C during checkpoint
- [ ] Test with `python -O` to verify validation still works

### DIFF
--- a/deepspeed/runtime/checkpoint_engine/decoupled_checkpoint_engine.py
+++ b/deepspeed/runtime/checkpoint_engine/decoupled_checkpoint_engine.py
@@ -188,7 +188,9 @@ class DecoupledCheckpointEngine(CheckpointEngine):
             post_size = get_checkpoint_folder_size(info.save_dir, info.tag, self.local_rank)
             self.checkpoint_size.set_post_size(post_size)
 
-        if self.global_rank == 0 and self.checkpoint_size.gb_size() is not None:
+        assert self.checkpoint_size.gb_size() is not None, "Checkpoint size should be set after commit"
+
+        if self.global_rank == 0:
             print(
                 f'{self.name} self.global_rank={self.global_rank} created checkpoint of {round(self.checkpoint_size.gb_size(), 2)} GB'
             )


sfc-gh-truwase:
@Rakshit-gen thanks for improving this feature. Are you planning to use DecoupledCheckpointEngine?

sfc-gh-truwase:
Can you also clarify if this is ready for review? I ask because of the test plan checklist.

Rakshit-gen:
@sfc-gh-truwase Yes, ready for review. The checklist is for reviewer validation - I've tested basic functionality locally but would appreciate someone with a multi-GPU decoupled checkpointing setup to verify the edge cases.

Rakshit-gen:
No, I'm not currently using DecoupledCheckpointEngine. I found these issues while reading through the codebase to understand how DeepSpeed handles async checkpointing. The missing timeouts and use of assert for validation stood out as potential production risks, so I figured I'd submit a fix while I was in there.

sfc-gh-truwase:
> understand how DeepSpeed handles async checkpointing. The missing timeouts and use of assert for validation stood out as potential production risks, so I figured I'd submit a fix while I was in there.

Awesome! Thanks for being so generous.

Rakshit-gen:
> Awesome! Thanks for being so generous.

Thanks! Happy to help improve the codebase.

Rakshit-gen:
is there any blocker? @sfc-gh-truwase

deepspeed/runtime/checkpoint_engine/decoupled_checkpoint_engine.py
```diff
@@ -112,16 +148,28 @@ def load(self, path: str, map_location=None):
     def save(self, state_dict, path: str):
         if self.ckpt_process is None:
```
sfc-gh-truwase, line 158: This condition is also checked by the new `_check_process_alive()` and so there seems to be some redundancy and inconsistency.

Rakshit-gen: preparing a fix

Rakshit-gen: pushed a fix for this and other possible inconsistencies

deepspeed/runtime/checkpoint_engine/decoupled_checkpoint_engine.py
```diff
@@ -97,7 +104,36 @@ def __init__(self, config_params, dp_writer_config, optimize_dp_state):
...
-        self.cleanup()
+        try:
+            self.cleanup()
+        except Exception:
+            # Suppress exceptions in destructor to avoid crashes during shutdown
+            pass
+
+    def _check_process_alive(self):
+        """Check if the checkpoint process is still alive."""
+        if self.ckpt_process is None:
```
sfc-gh-truwase: Actually, `self.ckpt_process == None` should not be a failing condition. The reason is that not all engines are [expected](https://github.com/deepspeedai/DeepSpeed/blob/6d9c3dc0b8383832d181cd70784f4d7f22348d0d/deepspeed/runtime/checkpoint_engine/decoupled_checkpoint_engine.py#L80-L83) to have `self.ckpt_process`.  For more details see Figure 6 of the [paper](https://arxiv.org/pdf/2406.13768).

This is why in many code paths `self.ckpt_process is None` is used to guard stateful changes. So, `_check_process_alive()` can be reduced to `self.ckpt_process.is_alive()`. 

Hope that makes sense.

Rakshit-gen: Good point, thanks for the clarification.

deepspeed/runtime/checkpoint_engine/decoupled_checkpoint_engine.py
```diff
@@ -110,27 +153,33 @@ def load(self, path: str, map_location=None):
...
+
+        self.commit_info = None
 
         if self.checkpoint_size.gb_size() is None:
             dist.barrier()
             post_size = get_checkpoint_folder_size(info.save_dir, info.tag, self.local_rank)
             self.checkpoint_size.set_post_size(post_size)
 
-        if self.global_rank == 0:
+        if self.global_rank == 0 and self.checkpoint_size.gb_size() is not None:
```
sfc-gh-truwase: I think it is better to add an assert that `self.checkpoint_size.gb_size() is not None` after line 180.

Rakshit-gen: looking into it

deepspeed/runtime/checkpoint_engine/decoupled_checkpoint_engine.py
Rakshit-gen: Fixed - _check_process_alive() now only checks self.ckpt_process.is_alive() and is only called after guarding with if self.ckpt_process is not None. The None case is handled separately as a valid state for ranks that don't have a checkpoint process.

deepspeed/runtime/checkpoint_engine/decoupled_checkpoint_engine.py
Rakshit-gen: Fixed. Added assert self.checkpoint_size.gb_size() is not None after the size is set, and reverted the print guard to the original if self.global_rank == 0:.

Issue #7747, opened 2025-12-27
[BUG]MoE router parameters are forced to bf16 under DeepSpeed bf16, causing dtype mismatch in fp32 routing logic


tangefly:
**Describe the bug**
When enabling bf16 mixed precision in DeepSpeed, all `nn.Parameter`s are forcibly cast to `torch.bfloat16` during initialization.
There is currently no supported way to keep specific parameters in `float32`.

This causes runtime dtype mismatch errors in modules that intentionally require fp32 computation (e.g., MoE routers or other numerically sensitive control-flow logic).

**To Reproduce**
1. Define a module with a parameter explicitly initialized in `float32`:

```python
class Ernie4_5_VL_MoeMoeTopKRouter(nn.Module):
    def __init__(self, config):
        super().__init__()
        self.weight = nn.Parameter(torch.zeros(config.moe_num_experts, config.hidden_size, dtype=torch.float32))
        self.moe_statics = Ernie4_5_VL_MoeMoeStatics(config)
        self.top_k = config.moe_k
        self.norm_min = config.moe_norm_min

    def forward(self, hidden_states: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
        device_type = (
            hidden_states.device.type
            if isinstance(hidden_states.device.type, str) and hidden_states.device.type != "mps"
            else "cpu"
        )

        with maybe_autocast(device_type=device_type, enabled=False):  # Force float32
            # key point #
            router_logits = F.linear(hidden_states.float(), self.weight)
            # # # # # # #
            routing_weights = F.softmax(router_logits, dim=1, dtype=torch.float)
            _, selected_experts = torch.topk(self.moe_statics(routing_weights), self.top_k, dim=-1)
            routing_weights = torch.gather(routing_weights, dim=-1, index=selected_experts)
            routing_weights = routing_weights / torch.clamp(
                routing_weights.sum(dim=-1, keepdim=True), min=self.norm_min
            )
        routing_weights = routing_weights.to(hidden_states.dtype)
        return router_logits, selected_experts, routing_weights
```

2. Enable bf16 in DeepSpeed config:

```json
{
  "bf16": {
    "enabled": true
  }
}
```

3. In the forward pass, disable autocast and perform fp32 computation:

```python
with torch.cuda.amp.autocast(enabled=False):
    out = F.linear(hidden_states.float(), self.weight)
```

4. Launch training with DeepSpeed.

5. Observe the following error:

```text
[...]

PR #7750, opened 2025-12-30
feat: add parameter-level precision control for BF16 training

nathon-lee:
fix: #7747 
Hi DeepSpeed team,

I hope this PR finds you well. I'm submitting this fix to resolve issue #7747, which addresses the problem where MoE router parameters are forcibly cast to bf16 under DeepSpeed bf16 configuration, causing dtype mismatch in fp32 routing logic.

Key changes:

Added should_preserve_dtype() helper function to check parameter preservation flags
Enhanced parameter processing in _setup_for_real_optimizer() to handle mixed precision scenarios
Updated storage management in _update_storage_to_flattened_tensor() to preserve original data types
Included comprehensive tests and documentation
Usage:
Users can now mark specific parameters with param.preserve_dtype = True to maintain their original precision while still benefiting from BF16 mixed precision training for other parameters.

This solution is backward compatible and provides an official mechanism for handling numerically sensitive modules like MoE routers. I've tested this thoroughly and believe it will be valuable for users facing similar precision-related issues.

Looking forward to your feedback and review. Thank you for your time and consideration!

Best regards,

### DIFF
--- a/deepspeed/runtime/bf16_optimizer.py
+++ b/deepspeed/runtime/bf16_optimizer.py
@@ -32,6 +32,21 @@ def print_rank_0(message, debug=False, force=False):
         print(message)
 
 
+def should_preserve_dtype(param):
+    """Check if parameter should preserve its original dtype.
+    
+    This function allows specific parameters to opt-out of bf16 conversion
+    by checking for the presence of a 'preserve_dtype' attribute.
+    
+    Args:
+        param (torch.nn.Parameter): The parameter to check
+        
+    Returns:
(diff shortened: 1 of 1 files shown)


tohtana:
Hi @leejianwoo-collab, 
This PR removes a lot of code in the BF16 optimizer, and I’m concerned it may break existing DeepSpeed features. Could you keep the current behavior?
I agree that adding parameter-level precision control is very important. If you think we should deprecate some features as part of this PR, we’re happy to discuss.

sfc-gh-truwase:
Closing due to lack of response.

Issue #7711, opened 2025-12-03
[BUG] OSError: MPI environment variables are not set.


Mulbetty:
**Describe the bug**
I am using ds-0.18.0 setting the launcher as openmpi, and get an error about MPI environment variables.

**To Reproduce**
run like following commond:
```
deepspeed \
    --hostfile=${HOSTFILE_PATH} \
    --launcher=OPENMPI \
    --launcher_args="-bind-to none -map-by slot --mca pml ob1 --oversubscribe --display-allocation --display-map" \
    --master_addr=${MASTER_ADDR} \
    --master_port=${_M_PORT} \
    --no_ssh_check \
    test.py
```
test.py could be any simple code.The error like:
```
Traceback (most recent call last):
  File "/usr/local/bin/deepspeed", line 6, in <module>
    main()
  File "/usr/local/lib/python3.12/dist-packages/deepspeed/launcher/runner.py", line 583, in main
    runner = OpenMPIRunner(args, world_info_base64, resource_pool)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/deepspeed/launcher/multinode_runner.py", line 129, in __init__
    super().__init__(args, world_info_base64)
  File "/usr/local/lib/python3.12/dist-packages/deepspeed/launcher/multinode_runner.py", line 23, in __init__
    self.validate_args()
  File "/usr/local/lib/python3.12/dist-packages/deepspeed/launcher/multinode_runner.py", line 145, in validate_args
    self._setup_mpi_environment()
  File "/usr/local/lib/python3.12/dist-packages/deepspeed/launcher/multinode_runner.py", line 160, in _setup_mpi_environment
    raise EnvironmentError("MPI environment variables are not set. "
OSError: MPI environment variables are not set. Ensure you are running the script with an MPI-compatible launcher.

```
I find the link:[[#6979-disscuss](https://github.com/deepspeedai/DeepSpeed/pull/6979/files#r1934769722)] mentioned likes my quesiton.
If I comment the self._setup_mpi_environment() in https://github.com/deepspeedai/DeepSpeed/blob/79caae1c04fca210345bdeb03bcacf87b1ac2f23/deepspeed/launcher/multinode_runner.py#L145, the error disappears. 
**Expected behavior**
no error about MPI environment variables.

**System info (please complete the following information):**
[...]

PR #7751, opened 2025-12-30
fix:  remove premature MPI environment variable check in OpenMPIRunner

nathon-lee:
fix: remove premature MPI environment variable check in OpenMPIRunner

Resolves issue #7711 where OpenMPI launcher incorrectly checked for MPI
environment variables before mpirun execution.

The problematic check in validate_args() was removed because:
- MPI environment variables (OMPI_COMM_WORLD_LOCAL_RANK, OMPI_COMM_WORLD_RANK,
  OMPI_COMM_WORLD_SIZE) are set by mpirun, not before it runs
- This timing issue caused false "MPI environment variables are not set" errors
- The _setup_mpi_environment() method is preserved for later use when appropriate

This fix allows OpenMPI launcher to work correctly without requiring users to
comment out the validation manually.

### DIFF
--- a/deepspeed/launcher/multinode_runner.py
+++ b/deepspeed/launcher/multinode_runner.py
@@ -147,21 +147,6 @@ class OpenMPIRunner(MultiNodeRunner):
         if self.args.num_nodes != -1 or self.args.num_gpus != -1:
             raise ValueError(f"{self.name} backend does not support limiting num nodes/gpus")
 
-    def _setup_mpi_environment(self):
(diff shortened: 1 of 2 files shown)


tohtana:
Hi @leejianwoo-collab,
Thank you for the fix! It was the only usage of `_setup_mpi_environment` other than tests. Thus, I think we should remove the function itself and the related tests.
I opened a PR to remove the code that will be unnecessary: https://github.com/leejianwoo-collab/DeepSpeed_woo/pull/1
Can you merge it?

PR #7767, opened 2026-01-08
Various fixes to run on mps

jeffra:
Attempting to run DeepSpeed (zero stage 0) with fp32 on MPS device. Ran into a few issues, these fixes resolve them.

* No CUDA-like timer events on MPS, should fall back to host timers
* When abstract accelerator doesn't define a comm backend we shouldn't trigger a broadcast or all-reduce


tohtana:
Hi @jeffra,
I really appreciate your contribution! This looks good to me, but shouldn’t we show a warning message saying we are skipping weight synchronization and gradient reduction just in case?

tohtana (approved):

Issue #7759, opened 2026-01-04
[BUG] ZenFlow: loss values becomes NaN after 'update_interval' number of steps.


karthik2603-theBrogrammer:
I was training a Llama model using the script released in the Deepspeed-Examples repository using ZenFlow. Interestingly, I noticed the loss values becoming NaN after the preconfigured (in the deepspeed config) update_interval number of steps.  Ex: the update interval is set to 4 in the deepspeed config, loss becomes nan from the 5th step. The code used is available here: https://github.com/deepspeedai/DeepSpeedExamples/tree/master/training/DeepSpeed-ZenFlow/finetuning

**My software configurations and versions:**

torch==2.5.0+cu118
Transformers==4.57.3
Cloned the latest version of deepspeed from github
Datasets==4.4.1

The jobs are run with DGX H200 GPUs and AMD EPYC 7742 processor.

I wish to understand why the loss turns NaN and more importantly why it happens only after the update_interval step. Can I please get some help with respect to this ?

Antlera:
Hi @karthik2603-theBrogrammer Thanks for reporting this issue. I was able to get access to DGX H200 resources to reproduce the problem today.

I noticed that there have been some API changes in the DeepSpeed engine and optimizer, and it’s likely that there is a mismatch between the current ZenFlow optimizer code and the updated engine/optimizer APIs. This may explain why the NaNs appear right after the `update_interval` boundary.

I also try to reproduce the NaN issue on other hardware, so this does not seem to be H200-specific. I’ll need a bit more time to track down exactly where the mismatch happens and fix it.

I’ll follow up once I have more concrete findings.

Antlera:
It appears that recent DeepSpeed changes unified the backward path to `loss.backward()`, bypassing `optimizer.backward()`. This means ZenFlow’s backward logic is not being triggered, which could explain the NaNs observed after the `update_interval` step.

Antlera:
Hi @karthik2603-theBrogrammer, I’ve opened a follow-up PR that fixes the ZenFlow backward mismatch.
If you get a chance, could you help test whether it works on H200? That would be very helpful.

PR #7771, opened 2026-01-11
Fix: ZenFlow Adam integration for updated PyTorch backward flow (#7759)

Antlera:
This PR refactors ZenFlow’s integration with DeepSpeed Adam/AdamW optimizers to adapt to recent changes in PyTorch’s backward execution model.
It is a follow-up to #7759 and adapts ZenFlow to recent PyTorch backward changes.
It is a fast patch to restore correct behavior under the new loss.backward() flow.

### DIFF
--- a/deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
+++ b/deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
@@ -538,15 +538,15 @@ class ZenFlowZeroOptimizer(DeepSpeedZeroOptimizer):
                 self._process_selected_fp32_groups_grad(tensor, curr_selected_reduce_size, communication_data_type)
                 self.timers(SELECTIVE_OPTIMIZER_PROCESS_TIMER).stop()
 
-    def backward(self, loss, retain_graph=False):
-        """
-        :attr:`backward` performs the following steps:
+    def backward_prologue(self):
+        """Prepare ZenFlow's per-microbatch state before the backward pass.
 
-        1. fp32_loss = loss.float()
-        2. scaled_loss = fp32_loss*loss_scale
-        3. scaled_loss.backward(), which accumulates scaled gradients into the ``.grad`` attributes of the model's fp16 leaves
+        Called by the engine at the start of each backward. Advances the
+        micro-step counter and, on an auto-update step, refreshes the
+        update-interval bookkeeping. At a selection boundary, resyncs the fp32
+        master partition from the bit16 weights and clears the selective
+        optimizer's moments so the next top-k update starts clean.
         """
-        self.backward_prologue()
         self.micro_step += 1
 
         if self.auto_update:
@@ -564,16 +564,6 @@ class ZenFlowZeroOptimizer(DeepSpeedZeroOptimizer):
             self.sync_fp32_param_from_gpu()
             self.selective_optimizer.clear_selected_mv()
             self.timers(SELECTIVE_OPTIMIZER_SYNC_TIMER).stop()
-
-        self.enter_backward()
-        if self.custom_loss_scaler:
-            scaled_loss = self.external_loss_scale * loss
-            scaled_loss.backward(retain_graph=retain_graph)
-        else:
-            self.loss_scaler.backward(loss.float(), retain_graph=retain_graph)
-
-        self.backward_epilogue()
-        self.exit_backward()
 
     def log_selective_optimizer_timers(self):
         self.timers.log(SELECTIVE_OPTIMIZER_TIMERS)


Antlera:
TODO: Ideally, ZenFlow’s backward logic should be fully compatible with the standard `loss.backward()` path in the future.

tohtana:
Thank you for the fix!
Can we throw an error if ZenFlow is enabled and backward is called in `loss.backward()`-style?

deepspeed/runtime/engine.py
```diff
@@ -2496,13 +2496,21 @@ def backward(self, loss, retain_graph=False, scale_wrt_gas=True):
...
+            # ZenFlow requires exclusive control over the backward pass to manage its
+            # selective parameter updates and synchronization boundaries.
+            if self.zenflow:
+                self.optimizer.backward(loss, **backward_kwargs)
+                self._backward_epilogue()
+                self._running_engine_backward = False
+                return gas_scaled_loss
+
+            # TODO: handle these scaling with direct calls to loss.backward()
+            if isinstance(self.optimizer, ZeROOptimizer):
```
tohtana: This block was originally outside the `compiled_autograd` context. I wonder if compiling the scaling might cause an issue when scaling factor is changed. Can we have this block outside the context?

Antlera: Hi @tohtana, Thanks for the review and reminder. Place this block outside the context should be better.

Antlera:
> Thank you for the fix! Can we throw an error if ZenFlow is enabled and backward is called in `loss.backward()`-style?

Good suggestion!

Antlera:
Hi @tohtana. I’ve added the corresponding handling in the latest commit based on your comment. Could you please help check whether this looks reasonable? Thanks!

tohtana:
Hi @Antlera,
Thank you for the fix! Some tests now fail. I wonder if those might be testing with FP16. Do you have any idea?

tohtana:
Hi @Antlera, I opened #7793 to fix the current issues of this PR. Do you think it works?

tohtana:
By the way, Stage 3 + `full_warm_up_rounds=0` still fail with:
```
AttributeError: 'Parameter' object has no attribute 'complete_column_offset'
```
Probably this is unrelated to this PR.

Antlera:
> Hi @Antlera, I opened #7793 to fix the current issues of this PR. Do you think it works?
Hi @tohtana. Thanks a lot for the fix! It works on my test bed and matches my original fix version. It seems like I misplaced the `gas_scaled_loss` in the following fix.

Antlera:
> By the way, Stage 3 + `full_warm_up_rounds=0` still fail with:
> 
> ```
> AttributeError: 'Parameter' object has no attribute 'complete_column_offset'
> ```
> 
> Probably this is unrelated to this PR.

Thanks for pointing this out! Yeah, this looks like an issue with attribute propagation or initialization.
I’ll open a new issue and handle it in a separate PR.

Antlera:
> By the way, Stage 3 + `full_warm_up_rounds=0` still fail with:
> 
> ```
> AttributeError: 'Parameter' object has no attribute 'complete_column_offset'
> ```
> 
> Probably this is unrelated to this PR.

Hi @tohtana. I’ve just created a follow-up issue for this: #7796

deepspeed/runtime/engine.py
tohtana: Yes, that's a good point. I wanted to resolve this so we can pass the full CI test suite, but it makes the core part of the optimizer more Zenflow-specific. We probably shouldn’t cut corners for this. Let me close this for now, and then consider a more general approach.

deepspeed/runtime/engine.py
```diff
@@ -2498,11 +2505,19 @@ def backward(self, loss, retain_graph=False, scale_wrt_gas=True):
 
         # TODO: handle these scaling with direct calls to loss.backward()
         if isinstance(self.optimizer, ZeROOptimizer):
-            loss = self.optimizer.scale_if_loss(loss)
```
Antlera, line 2825: Hi @tohtana, do you think this is OK?
We could keep this PR as the current fix. Since ZenFlow implements its own backward logic, we don’t need to add an extra layer of handling in the higher-level optimizer for now. This won’t affect the existing optimizer logic.

The test issue mainly comes from how the scaled loss is handled; fixing this part should be sufficient to make the tests pass.

Antlera: We may reopen this PR if you think this approach works.

tohtana:
@Antlera Sorry, I mistakenly closed this issue, being confused with #7793.
I'm still not very clear about the current change. Can you help me understand the issue?

DeepSpeed engine's forward always registers a backward hook that calls `_backward_prologue` on the original loss, which already invokes self.optimizer.backward_prologue() and enter_backward() for ZeRO optimizers. With the new ZenFlow path, `engine.backward()` also calls `self.optimizer.backward(...)`, and `ZenFlowZeroOptimizer.backward()` itself calls `backward_prologue()`, which increments micro_step. Then, `backward_epilogue()` and `exit_backward()` are called again. 

To enable this use case in a general way (not too hard-coded for Zenflow), we need to allow an advanced custom optimizer to control hooks in an earlier stage (like the engine's initialization).
What do you think?

Antlera:
> @Antlera Sorry, I mistakenly closed this issue, being confused with #7793. I'm still not very clear about the current change. Can you help me understand the issue?
> 
> DeepSpeed engine's forward always registers a backward hook that calls `_backward_prologue` on the original loss, which already invokes self.optimizer.backward_prologue() and enter_backward() for ZeRO optimizers. With the new ZenFlow path, `engine.backward()` also calls `self.optimizer.backward(...)`, and `ZenFlowZeroOptimizer.backward()` itself calls `backward_prologue()`, which increments micro_step. Then, `backward_epilogue()` and `exit_backward()` are called again.
> 
> To enable this use case in a general way (not too hard-coded for Zenflow), we need to allow an advanced custom optimizer to control hooks in an earlier stage (like the engine's initialization). What do you think?

Hi @tohtana. Sorry for missing this message. You're right about the double-invocation, that was the issue, and I've fixed it. The latest push removes ZenFlowZeroOptimizer.backward() and moves its work into a backward_prologue() override, so the prologue and epilogue run once and ZenFlow follows the PyTorch-style backward path (direct loss.backward() works).

I also verified the loss now matches the old ZenFlow step for step, where master currently goes NaN.

Antlera:
Hi @tohtana. Could you help review updated pytorch-style fix for zenflow? Thanks!

delock (approved):

PR #7776, opened 2026-01-12
Fix Muon optimizer conflict with gradient clipping in ZeRO 1/2

fy817:
### Summary

This PR fixes a critical logic error in how gradient clipping interacts with the **Muon optimizer** within the ZeRO Stage 1 and 2 implementations.

Currently, the `muon_update` (orthogonalization) is applied before gradient clipping. Since Muon's orthogonalization naturally changes the gradient magnitude (often resulting in norms significantly larger than the default `clip_grad` of 1.0), the subsequent standard clipping mechanism incorrectly identifies these valid updates as exploding gradients. This results in aggressive, erroneous clipping that severely degrades or breaks training.

This fix ensures that gradient clipping is applied to the raw gradients **before** the Muon update occurs, ensuring the orthogonalization step receives properly clipped inputs, while preventing double-clipping during the optimizer step.

### Root Cause

In the current implementation, `muon_update` is invoked inside `get_flat_partition` to leverage the shaped gradients required by the algorithm. However, `get_flat_partition` is called before the standard global gradient clipping phase.

1. **Muon Characteristics:** The Muon optimizer performs an orthogonalization step (via Newton-Schulz iteration), which typically results in gradient tensors with Frobenius norms much larger than 1.0.
2. **Clipping Conflict:** DeepSpeed defaults to a `clip_grad` value of 1.0. Because the gradients have already been processed by Muon by the time clipping occurs, their norms are naturally high.
3. **Result:** The global clipper perceives these high norms as gradient explosion and scales the gradients down by a massive factor (e.g., dividing by 100+), effectively neutralizing the Muon update and causing training failure.

### Fix

The fix rearranges the clipping logic for parameters utilizing the Muon optimizer:

1. **Pre-computation:** In `independent_gradient_partition_epilogue`, we now pre-calculate the unscaled global gradient norm (or the specific norm required for clipping) before the flattening process begins.
2. **Pre-Update Clipping:** This calculated norm is passed into `get_flat_partition`. Inside this method, we calculate the `clip_factor` and apply it (`grad = grad / clip_factor`) explicitly to the gradients **before** passing them to `muon_update`. This ensures Muon operates on clipped, stable gradients.
3. **Skip Redundant Clipping:** In the `step` method, we modified the call to `unscale_and_clip_grads` to accept a `skip_clipping=True` flag. For Muon parameter groups, we set this flag to true, ensuring that we only perform unscaling (handling loss scale) without re-applying gradient clipping, as it was already handled in the previous step. Standard AdamW parameters retain the original behavior.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1416,6 +1416,14 @@ class DeepSpeedEngine(Module):
         self.expert_data_parallel_group = groups._get_expert_data_parallel_group_dict()
         self.sequence_parallel_size = groups._get_sequence_parallel_world_size()
         if self.sequence_parallel_size > 1:
+            # Inserted Warning for PyTorch < 2.3
+            if not required_torch_version(min_version=2.3):
+                logger.warning(
+                    "DeepSpeed Sequence Parallelism (Ulysses) with PyTorch < 2.3 may encounter "
+                    "rank indexing errors during backward pass when sp_size < world_size. "
+                    "Please use the weighted all-reduce workaround shown in the regression test "
+                    "(https://github.com/deepspeedai/DeepSpeed/blob/master/tests/unit/sequence_parallelism/test_ulysses.py) "
+                    "or upgrade to PyTorch 2.3+.")
             self.communication_data_type = self._config.seq_parallel_communication_data_type
             self.seq_parallel_group = groups._get_sequence_parallel_group()
 
--- a/docs/_tutorials/ulysses-alst-sequence-parallelism.md
+++ b/docs/_tutorials/ulysses-alst-sequence-parallelism.md
@@ -222,6 +222,12 @@ In theory you could just average `losses_per_rank`, but the system supports vari
 
 ## Nuances
 
+### Note on PyTorch Versions < 2.3
+
+If you are using Sequence Parallelism with **PyTorch version < 2.3**, you may encounter an `IndexError: tuple index out of range` during the backward pass when `sequence_parallel_size < world_size`. This is due to a known issue in the `torch.distributed.all_gather` backward implementation in older versions.
+
+**Workaround:** We recommend using a **weighted `all_reduce` pattern** instead of `all_gather` for loss averaging. You can refer to our [regression test case](https://github.com/deepspeedai/DeepSpeed/blob/master/tests/unit/sequence_parallelism/test_ulysses.py) for a code example of this workaround.
+
 ### Why do labels need to be pre-shifted?
 
 When using batch sharding one can't let the upstream `loss` function do the labels shifting. Here is why:
--- a/tests/unit/sequence_parallelism/test_ulysses.py
+++ b/tests/unit/sequence_parallelism/test_ulysses.py
@@ -255,3 +255,58 @@ class TestFPDTAttention(DistributedTest):
         assert torch.allclose(
             fpdt_output, baseline_output_shuffled, rtol=0.01, atol=0.1
         ), f"rank {dist.get_rank()}, sp size: {dist.get_world_size(spg)}, input_tensor: {input_tensor.shape}, fpdt_input_tensor: {fpdt_input_tensor.shape}, fpdt_output: {fpdt_output.shape},            baseline_output_shuffled: {baseline_output_shuffled.shape},{torch.max(torch.abs(fpdt_output - baseline_output_shuffled))}"
+
+
+@pytest.mark.parametrize("sp_size", [2])
+class TestUlyssesLossBackward(DistributedTest):
+    world_size = 4
+
+    def test_sp_loss_backward_stability(self, sp_size: int) -> None:
+        """
+        Regression test for Issue #7672.
+        Verifies that using all_reduce for loss aggregation is stable
+        when sequence_parallel_size < world_size, preventing IndexError.
+        """
+        skip_on_arch(min_arch=8)
+
+        # Setup
+        dp_size = self.world_size // sp_size
+        model = SimpleModel(4)
+        ds_engine, _, _, _ = initialize(
+            model=model,
+            config_params={
+                "train_batch_size": 8,
+                "data_parallel_size": dp_size,
+                "sequence_parallel_size": sp_size
+            },
+        )
+
+        sp_group = ds_engine.seq_parallel_group
+
+        # Simulate Loss on each rank
+        rank = dist.get_rank()
+        local_loss = torch.tensor(float(rank + 1), device=ds_engine.device, requires_grad=True)
+        local_weight = torch.tensor(1.0, device=ds_engine.device)
+
+        # Numerator: Weighted Loss summation
+        weighted_loss = local_loss * local_weight
+        dist.all_reduce(weighted_loss, op=dist.ReduceOp.SUM, group=sp_group)
+
+        # B. Denominator: Sum of total weights
+        total_weight = local_weight.clone()
+        dist.all_reduce(total_weight, op=dist.ReduceOp.SUM, group=sp_group)
+
+        # C. Calculate the final loss
+        dist_loss = weighted_loss / total_weight
+
+        # Backward Pass verification
+        try:
+            dist_loss.backward()
+        except IndexError as e:
+            pytest.fail(f"Backward crashed with IndexError: {e}")
+
+        # Verify Gradients
+        # Loss = (L1*1 + L2*1) / 2 = 0.5*L1 + 0.5*L2
+        expected_grad = 0.5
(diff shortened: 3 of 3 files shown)


sfc-gh-truwase:
@PKUWZP, please help review

fy817:
Hi @PKUWZP, could you please take a look when you have a moment? Let me know if you need anything else from my side. Thanks!

fy817:
Hi @sfc-gh-truwase, it seems @PKUWZP might be busy at the moment. Would it be possible for you or another maintainer to take a look? Thanks for your help!

sfc-gh-truwase:
@fy817 can you please resolve the conflict?

fy817:
@sfc-gh-truwase Thank you for the feedback. The conflicts have been resolved. Please refer to my comment above for details on how I handled the conflicts. Looking forward to your review.

sfc-gh-truwase:
> The conflicts have been resolved. Please refer to my comment above for details on how I handled the conflicts. Looking forward to your review

Can you please clarify how the conflicts were resolved, since GH is still reporting a conflict
image

fy817:
@sfc-gh-truwase  Sorry about that! I misunderstood the context and missed the merge conflict. It's been fixed now. Thanks for the heads-up.

fy817:
@sfc-gh-truwase Hi, I noticed that this PR has been linked with #7808. Please let me know if there is anything else needed from my side. I am happy to help get this merged.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -859,6 +859,32 @@ def independent_gradient_partition_epilogue(self):
             self._clear_previous_reduced_grads()
 
         if self.cpu_offload is False:
+            # Pre-compute gradient norm for Muon clipping if needed
+            grad_norm_for_muon = None
+            if self.is_gradient_accumulation_boundary:
+                # Check if any parameter group uses Muon
+                uses_muon = False
```
sfc-gh-truwase, line 866: The main connection to #7808 is that `use_muon` should be an object field set in the constructor and defined per-param_group.

PKUWZP, line 866: @fy817 Like @sfc-gh-truwase mentioned, `use_muon` should be an object field. This should be computed once and stored as a class field.  Computing this repeatedly is inefficiency and can cause errors.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -859,6 +859,32 @@ def independent_gradient_partition_epilogue(self):
...
+                for i, _ in enumerate(self.bit16_groups):
+                    if len(self.params_in_partition[i]) > 0 and getattr(self.params_in_partition[i][0], 'use_muon', False):
```
PKUWZP, line 868: @fy817 `getattr(tensor_list[0], 'use_muon', False)` needs to be set externally by the user. Otherwise there's no documentation and validation that this attribute would exist or be set properly.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -859,6 +859,32 @@ def independent_gradient_partition_epilogue(self):
...
+                        uses_muon = True
```
PKUWZP, line 869: @fy817 One suggestion is that add validation in `__init__ ` of the `DeepSpeedZeroOptimizer` class that checks for use_muon attributes when a Muon optimizer is detected, and add documentation about this requirement.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -859,6 +859,32 @@ def independent_gradient_partition_epilogue(self):
             self._clear_previous_reduced_grads()
 
         if self.cpu_offload is False:
+            # Pre-compute gradient norm for Muon clipping if needed
+            grad_norm_for_muon = None
```
PKUWZP, line 863: Here you introduced a new instance variable `grad_norm_for_muon` that's set in `independent_gradient_partition_epilogue` function and consumed in `get_flat_partition` function. This would potentially cause data racing problem due to the resulting implicit dependency on the call orders. One solution is that we make the `grad_norm_for_muon` defined explicitly outside the function, then we can ensure that it's always valid before being passed into functions.

PKUWZP:
@fy817 Thanks so much for submitting the PR. I have reviewed the PR and left a few comments. One fundamental question I have is: do we actually need "naive gradient clipping" for Muon Optimizer? I feel that it might not even needed for Muon. Here are my thoughts: 

In Muon Optimizer Implementation (shown below), the Newton-Schulz iteration already normalizes by spectral norm internally. The output is approximately an orthogonal matrix whose norm is determined by the matrix shape, not the input gradient magnitude.

  `def zeropower_via_newtonschulz5(G, steps: int):`
          `X = G.bfloat16()`
          `# Ensure spectral norm is at most 1`
         `X = X / (X.norm(dim=(-2, -1), keepdim=True) + 1e-7) `
        ` # Newton-Schultz Iterations`
         `for _ in range(steps):`
             `A = X @ X.mT`
             `B = b * A + c * A @ A`
             `X = a * X + B @ X`
         `return X`

  This means:
  1. Exploding gradients are already handled - the spectral norm division bounds the input
  2. The output norm is shape-dependent, not gradient-dependent - no need to clipping the output
  3. Performing pre-clipping changes gradient direction unnecessarily - and the magnitude will be normalized anyway

Instead, I think MuonClilp (described in Kimi K2 technical report: https://arxiv.org/abs/2507.20534 ) is actually interesting for this case. It rescales query and key matrices within attention layers after each update to further improve training stability. I just don't feel doing naive pre-clipping is needed here for Muon. 

Also in this PR, the global norm is computed via `scaled_global_norm()` on all `averaged_gradients`, this is not suitable for mixed setup (e.g. Muon for hidden layers while AdamW for embedding layers ). If you do this, the global norm will include the orthogonalized Muon gradients, which would not be the right thing to do to clip the AdamW parameter groups. Thoughts?

PKUWZP (commented):
@fy817 Thanks so much for submitting the PR. I have reviewed the PR and left a few comments. One fundamental question I have is: do we actually need "naive gradient clipping" for Muon Optimizer? I feel that it might not even needed for Muon. Here are my thoughts: 

In Muon Optimizer Implementation (shown below), the Newton-Schulz iteration already normalizes by spectral norm internally. The output is approximately an orthogonal matrix whose norm is determined by the matrix shape, not the input gradient magnitude.

  `def zeropower_via_newtonschulz5(G, steps: int):`
          `X = G.bfloat16()`
          `# Ensure spectral norm is at most 1`
         `X = X / (X.norm(dim=(-2, -1), keepdim=True) + 1e-7) `
        ` # Newton-Schultz Iterations`
         `for _ in range(steps):`
             `A = X @ X.mT`
             `B = b * A + c * A @ A`
             `X = a * X + B @ X`
         `return X`

  This means:
  1. Exploding gradients are already handled - the spectral norm division bounds the input
  2. The output norm is shape-dependent, not gradient-dependent - no need to clipping the output
  3. Performing pre-clipping changes gradient direction unnecessarily - and the magnitude will be normalized anyway

Instead, I think MuonClilp (described in Kimi K2 technical report: https://arxiv.org/abs/2507.20534 ) is actually interesting for this case. It rescales query and key matrices within attention layers after each update to further improve training stability. I just don't feel doing naive pre-clipping is needed here for Muon. 

Also in this PR, the global norm is computed via `scaled_global_norm()` on all `averaged_gradients`, this is not suitable for mixed setup (e.g. Muon for hidden layers while AdamW for embedding layers ). If you do this, the global norm will include the orthogonalized Muon gradients, which would not be the right thing to do to clip the AdamW parameter groups. Thoughts?

fy817:
@sfc-gh-truwase, @PKUWZP, thank you for your responses. Below are some of my thoughts:
1. Indeed, the Muon optimizer may not require gradient clipping. My experiments also confirm this—setting the gradient clipping value to 0 resolves the aforementioned conflict between the Muon optimizer and gradient clipping, and normal training performance can still be achieved.
2. However, I would like to point out that the conflict between the Muon optimizer and gradient clipping still exists in the current implementation. First, users might overlook the gradient clipping setting. Since DeepSpeed defaults to a gradient clipping value of 1.0, this can lead to incorrect gradient modification due to clipping, ultimately causing training failures—a problem that still needs to be addressed. Second, even if we resolve the conflict by setting gradient clipping to 0, while Muon itself may not need clipping, users might still require gradient clipping for the Adam update portion of the optimizer.
3. Thank you very much for pointing out the shortcomings in the current PR. I look forward to further discussion to refine this solution.

PR #7777, opened 2026-01-13
Remove cron/PR triggers for outdated V100 tests

loadams:
The V100 tests are not needed anymore but this prevents the CI cron jobs from being spun up even though the jobs are disabled.

The next step will be to remove the yaml files we do not use anymore/have already ported.


loadams:
FYI @tjruwase -  I cleaned up the README separately, you can see it reflected on the main page already

tohtana (approved):

PR #7780, opened 2026-01-15
Fix gradient checkpointing with use_reentrant=True / PyTorch-style backward / ZeRO-3

tohtana:
Fixes an issue where ZeRO-3 with gradient checkpointing (`use_reentrant=True`) and non-scalar backward (`tensor.backward(gradient=...)`) would fail on subsequent training iterations with `AttributeError: 'NoneType' object has no attribute 'numel'`.

The root cause was stale parameters remaining in `ipg_buckets` between iterations due to the checkpoint's backward re-running forward. The fix clears these buckets in `_pre_step()` before each optimizer step.

Added comprehensive tests for gradient checkpointing with ZeRO-3 covering both `use_reentrant=True` and `use_reentrant=False` modes.

### DIFF
--- a/deepspeed/runtime/comm/compressed.py
+++ b/deepspeed/runtime/comm/compressed.py
@@ -8,6 +8,7 @@ import torch
 import deepspeed.comm as dist
 from deepspeed.accelerator import get_accelerator
 from deepspeed.ops.op_builder import PackbitsBuilder
+from deepspeed.runtime.comm.utils import check_and_handle_empty_buffer
 
 
 class CompressedBackend(object):
@@ -62,6 +63,9 @@ class CompressedBackend(object):
         # align size of original_buffer and error
         original_size = buffer_m.numel()
         worker_error_size = worker_error.numel()
+        result = check_and_handle_empty_buffer(buffer_m, original_shape, original_size, worker_error, server_error)
+        if result is not None:
+            return result
         if original_size != worker_error_size:
             empty_tensor = torch.zeros(worker_error_size - original_size, device=buffer_m.device)
             buffer_m = torch.cat([buffer_m, empty_tensor])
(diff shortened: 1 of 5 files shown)


deepspeed/runtime/zero/stage3.py
```diff
@@ -1963,6 +1963,14 @@ def _pre_step(self):
 
         print_rank_0("Finished Tracing at Beginning of Step")
 
+        # Clear any stale params from ipg_buckets. This is needed because with
+        # reentrant checkpointing (use_reentrant=True), the backward pass can
+        # leave params in the buckets that weren't properly processed, causing
+        # errors in the next iteration.
+        for bucket in self.ipg_buckets.values():
```
PKUWZP, line 2001: @tohtana Maybe convert to a function: 

`def clear_params(self): self.params.clear() self.element =0`, thoughts?

tests/unit/v1/zero/test_zero_user_backward.py
```diff
@@ -1148,3 +1148,221 @@ def test_scale_with_torch_autocast(self, zero_stage):
...
+        # Verify gradients were computed
+        assert len(ds_grads) > 0, \
+            f"No gradients computed with scalar loss, use_reentrant={use_reentrant}"
+
+        # Compare gradients with DDP reference
+        compare_gradients(ddp_grads, ds_grads, f"scalar loss with checkpointing use_reentrant={use_reentrant}")
+
+        model_engine.destroy()
+
+    def test_checkpointed_multiple_backward_zero3(self, use_reentrant):
```
PKUWZP: @tohtana Do we need to compare with DDP reference here as well?

loadams (approved):

deepspeed/runtime/zero/stage3.py
tohtana: Thank you for the comment, addressed it

tests/unit/v1/zero/test_zero_user_backward.py
tohtana: Added the comparison with DDP. It actually revealed a few issues. Thank you for the suggestion!

PR #7783, opened 2026-01-17
Add workflow to run full tests

tohtana:
Run all tests on the AWS infra. Not scheduled, needs to be manually launched. (Reopened from this repo, not a fork, for testing)
`aws-torch-latest` now runs full tests for debugging. Once it works, we will revert ti and add `aws-torch-latest-full`.


tohtana:
Confirmed the full tests pass after cherry-picking fixes from unmerged PRs. Closing as #7795 is open as a cleaner solution.

PR #7787, opened 2026-01-18
Fix backward for pipeline engine

tohtana:
The new backward API introduced in #7665 broke the nested backward call used in the pipeline engine. Tests in unit/checkpoint/test_pipeline.py hang because of this issue.
This PR fixes the issue by manually setting a flag to properly call backward hooks.

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -852,13 +852,22 @@ class PipelineEngine(DeepSpeedEngine):
             # manually call because we don't call optimizer.backward()
             self.optimizer.clear_lp_grads()
 
-        # This handles either a single tensor or tuple of tensors.
-        if isinstance(outputs, tuple):
-            out_tensors = [t for t in outputs if t.is_floating_point()]
-            assert len(out_tensors) == len(grad_tensors)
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/pipe/engine.py
```diff
@@ -852,13 +852,22 @@ def _exec_backward_pass(self, buffer_id):
...
-        if isinstance(outputs, tuple):
-            out_tensors = [t for t in outputs if t.is_floating_point()]
-            assert len(out_tensors) == len(grad_tensors)
-            torch.autograd.backward(tensors=out_tensors, grad_tensors=grad_tensors)
-        else:
-            torch.autograd.backward(tensors=(outputs, ), grad_tensors=(grad_tensors, ))
+        # Set _running_engine_backward to avoid RuntimeError in post-backward hook
+        # when needs_scaler=True (the hook checks this flag to skip error checking)
+        self._running_engine_backward = True
+        try:
```
sfc-gh-truwase, line 858: 1. Why is `try` needed here? 
2. What are the possible exceptions, and why can be they ignored?

tohtana: This is try-finally (not try-except), so exceptions are NOT ignored. The exceptions are propagated normally. I just wanted to make sure we reset the flag when any error happens (like OOM) in backward.

PR #7788, opened 2026-01-18
Fix issue with BF16 optimizer selection

tohtana:
**Note:** Updated based on the change 64b10739a66704afc4112d10ab2d70f2b3a2266c for #7790. With the fix, `BF16_Optimizer` now requires ZeRO stage 1 to be explicitly enabled. 

The test `test_bf16_optimizer_fragments` fails with an `AssertionError` because the `BF16_Optimizer` is not being instantiated when expected. The test checks for `_hp_mapping` attribute on parameters, which is only set by `BF16_Optimizer`.

  The test `test_bf16_optimizer_fragments` fails because:
  1. The test config (`bf16=True` without grad_accum_dtype) **correctly** uses `FP16_Optimizer`, but the test expects `BF16_Optimizer` (which sets `_hp_mapping`)
  2. `BFLOAT16` and `DDP_BFLOAT16` have the same value `"bf16"`, preventing proper optimizer selection
  3. `BF16_Optimizer` is missing attributes required by the base class API

  This PR addresses these issues.

  Optimizer selection summary:

  | ZeRO Stage | Config | Optimizer | Gradient Accumulation |
  |------------|--------|-----------|----------------------|
  | 0 | `bf16=True` (default) | `FP16_Optimizer` | bf16 |
  | 0 | `bf16=True` + `grad_accum_dtype=fp32` | `NotImplementedError` | - |
  | 1 | `bf16=True` + `grad_accum_dtype=fp32` | `BF16_Optimizer` | fp32 |

  This is confusing (e.g., `FP16_Optimizer` handles both fp16 and bf16). We would need to simplify the code paths and clarify the behaviors in the future.

### DIFF
--- a/deepspeed/runtime/base_optimizer.py
+++ b/deepspeed/runtime/base_optimizer.py
@@ -355,7 +355,9 @@ class ZeROOptimizer(DeepSpeedOptimizer):
                 return self.external_loss_scale * value
             if self.torch_autocast_gradscaler:
                 return self.torch_autocast_gradscaler.scale(value)
-            return self.loss_scaler.scale_loss(value)
+            # Only call loss_scaler if it exists (not present in BF16_Optimizer)
+            if hasattr(self, 'loss_scaler') and self.loss_scaler is not None:
+                return self.loss_scaler.scale_loss(value)
 
         return value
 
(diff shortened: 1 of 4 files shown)


sfc-gh-truwase (approved):

tohtana:
@sfc-gh-truwase Sorry, I missed to change zero stage in the test. Please review #7803.

PR #7789, opened 2026-01-18
Skip empty parameters in gradient reduction

tohtana:
#7736 fixed an issue with OnebitLamb NaN propagation. With the fix, the optimizer correctly filters out empty parameters, but DeepSpeed engine's gradient allreduce operation (which runs separately from the optimizer) still includes empty parameters' gradients.

This PR addresses the issue by skipping empty parameters (numel=0) in `_get_gradients_for_reduction()`.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -2940,6 +2940,11 @@ class DeepSpeedEngine(Module):
             if not param.requires_grad:
                 continue
 
+            # Skip empty parameters (numel=0) as they contribute nothing to gradient reduction
+            # and cause issues with flatten/unflatten operations
+            if param.numel() == 0:
+                continue
+
             if param.grad is None:
                 # In cases where there is an imbalance of empty grads across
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/engine.py
```diff
@@ -2940,6 +2940,11 @@ def _get_gradients_for_reduction(self):
             if not param.requires_grad:
                 continue
 
+            # Skip empty parameters (numel=0) as they contribute nothing to gradient reduction
+            # and cause issues with flatten/unflatten operations
+            if param.numel() == 0:
```
PKUWZP, line 2945: @tohtana Very clean fix! The only minor comment is that maybe we can add an explicit test for gradient reduction? It's optional though.

tohtana: Thank you for the feedback. Can you elaborate what test you are suggesting? Run only gradient reduction?

tohtana: Let me merge this PR for now. Please open a new issue or PR for the additional test.

PR #7790, opened 2026-01-18
Fix BF16_Optimizer being used without ZeRO

tohtana:
The test_ds_initialize.py::TestOptimizerImplementation expects the configuration (None, 'bf16', 'fp32') is unimplemented.
However, it is actually supported by DeepSpeed and uses FP16_Optimizer in bf16 mode.
This PR adds the configuration in the dict of supported combinations.

### DIFF
--- a/deepspeed/runtime/base_optimizer.py
+++ b/deepspeed/runtime/base_optimizer.py
@@ -355,7 +355,9 @@ class ZeROOptimizer(DeepSpeedOptimizer):
                 return self.external_loss_scale * value
             if self.torch_autocast_gradscaler:
                 return self.torch_autocast_gradscaler.scale(value)
-            return self.loss_scaler.scale_loss(value)
+            # Only call loss_scaler if it exists (not present in BF16_Optimizer)
+            if hasattr(self, 'loss_scaler') and self.loss_scaler is not None:
+                return self.loss_scaler.scale_loss(value)
 
         return value
 
--- a/deepspeed/runtime/bf16_optimizer.py
+++ b/deepspeed/runtime/bf16_optimizer.py
@@ -59,6 +59,11 @@ class BF16_Optimizer(ZeROOptimizer):
                                   ], f"BF16Optimizer: Unsupported gradient accumulation data type: {grad_acc_dtype}"
         self.grad_acc_dtype = grad_acc_dtype
 
+        # BF16 doesn't use loss scaling, but these attributes are needed for API compatibility
+        self.custom_loss_scaler = False
+        self.external_loss_scale = None
+        self.torch_autocast_gradscaler = None
+
         self.immediate_grad_update = bfloat16_config.immediate_grad_update
 
         self.clip_grad = clip_grad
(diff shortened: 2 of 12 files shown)


sfc-gh-truwase:
@tohtana can clarify this? 
1. I am not aware that [FP16_Optimizer](https://github.com/deepspeedai/DeepSpeed/blob/43125a753d68ab5a5fad6ddd56c1ff21d6e97d7f/deepspeed/runtime/fp16/fused_optimizer.py#L33) supports `fp32` grad accumulation. 
2. If I am wrong, and it does, then it means that `(None, 'fp16', 'fp32')`  should also be supported.

tohtana:
> 1. I am not aware that [FP16_Optimizer](https://github.com/deepspeedai/DeepSpeed/blob/43125a753d68ab5a5fad6ddd56c1ff21d6e97d7f/deepspeed/runtime/fp16/fused_optimizer.py#L33) supports `fp32` grad accumulation.

Sorry for the confusion. You are right, it was actually `BF16_Optimizer`. I was probably confused with the related topic in #7788.

sfc-gh-truwase:
>Sorry for the confusion. You are right, it was actually BF16_Optimizer

No worries. There are long existing confusions about ZeRO and BF16_Optimizer.  For context, `BF16_Optimizer` already provides `fp32` grad accum, but corresponds to `(1, 'bf16', 'fp32')`. In that case, is this PR still needed?

tohtana:
Yes, the PR is still needed. The configuration `(None, 'bf16', 'fp32')` (this is about  **without ZeRO**. `('zero1', 'bf16', 'fp32')` is already in the list) is supported but wasn't listed in the test's is_supported dictionary.

The following code in `engine.py` handles the case when ZeRO is not enabled and uses `BF16_Optimizer`.

```python
  elif model_dtype == torch.bfloat16 and grad_accum_dtype == torch.float32:
      return BFLOAT16
```

The test was incorrectly expecting (None, 'bf16', 'fp32') to raise NotImplementedError, but the code actually supports it.

sfc-gh-truwase:
I see. This [logic](https://github.com/deepspeedai/DeepSpeed/blob/3d896d7e06aa3359282fbe2300a85f9520b6b594/deepspeed/runtime/engine.py#L1484-L1485) is actually a bug and should be removed. The reason is that it will instantiate `BF16_Optimizer` which is a variant of ZeRO1, even though the user did not request ZeRO. This is one of the long-standing confusions in the ZeRO & BF16 handling that I was referring to. Sorry that I missed this line in my previous cleanup #7515.

tohtana:
I got it. I reverted the change in the test and removed the conditional branch in the engine.
I will also need to fix a test (uses `(None, 'bf16', 'fp32')`) regarding #7788.

tohtana:
Updated #7788.

sfc-gh-truwase (approved):

sfc-gh-truwase:
Thanks! Please update the PR title to reflect bug fix.

PR #7792, opened 2026-01-18
Fix bf16 dtype mismatch in ZeRO-3 with zero_quantized_weights

juyterman1000:
When using ZeRO-3 with zero_quantized_weights=True and bf16 enabled, the dequantized weights were incorrectly cast to fp16 instead of preserving the original bf16 dtype. This caused RuntimeError during training.

The fix adds original_dtype tracking to AllGatherCoalescedHandle, mirroring the existing pattern in AllGatherHandle, to ensure weights are converted back to their original dtype after dequantization.


PKUWZP:
@juyterman1000 Thanks for working on this. Can you check two things here: 1. Double-check `AllGatherHandle` (the non-coalesced version) doesn't have the same bug in its quantization path (lines 697-700 in `partition_parameters.py` file); 2. (Optional) Add a regression test for bf16 + zero_quantized_weights.  I think 1) might be an interesting verification, and if true, we can have a follow up PR.

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -735,8 +737,13 @@ def wait(self, handle_dependency=True) -> None:
 
         if self.quantization:
             instrument_w_nvtx(self.quantization.quant_handle.wait)()
-            flat_tensor = self.quantization.backend.dequantize(
-                self.quantization.quantized_param, self.quantization.scale_buffer).to(self.params[0].device)
+            # Fix for issue #7775: convert dequantized tensor back to original dtype (e.g., bf16)
+            # to prevent dtype mismatch when zero_quantized_weights is used with bf16
+            dequantized = self.quantization.backend.dequantize(
```
PKUWZP: @juyterman1000 Good fix! Can you also double-check `AllGatherHandle` (the non-coalesced version) doesn't have the same bug in its quantization path (lines 697-700)?

juyterman1000:
Thanks @PKUWZP for the suggestions! I've updated the PR with both items.

juyterman1000:
Hi @tohtana, 
please have a look when you have a moment. I’m happy to make any necessary adjustments if you have concerns.

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -1097,10 +1108,8 @@ def __init__(self,
 
         self.use_all_reduce_for_fetch_params = get_config_default(DeepSpeedZeroConfig,
                                                                   "use_all_reduce_for_fetch_params")
-        self.allgather_sequential = get_config_default(DeepSpeedZeroConfig, "allgather_sequential")
```
tjruwase, line 1110: It seems this is disabling the sequential allgather feature introduced in #7661. Is that correct?

PR #7793, opened 2026-01-18
Fix loss scaling and backward call of ZenFlow

tohtana:
@Antlera Fix loss scaling and backward call for ZeRO3 of https://github.com/deepspeedai/DeepSpeed/pull/7771 (not merged).

### DIFF
--- a/deepspeed/runtime/zero/stage3.py
+++ b/deepspeed/runtime/zero/stage3.py
@@ -2376,6 +2376,25 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
         if self.swap_optimizer:
             self.optimizer_swapper.post_backward()
 
+    def backward(self, loss, retain_graph=False):
+        """
+        Backward pass for Stage 3 optimizer.
+
+        When ZenFlow is enabled, this method handles the specialized backward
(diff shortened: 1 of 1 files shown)


deepspeed/runtime/zero/stage3.py
```diff
@@ -2376,6 +2376,25 @@ def backward_epilogue(self):
         if self.swap_optimizer:
             self.optimizer_swapper.post_backward()
 
+    def backward(self, loss, retain_graph=False):
```
Antlera, line 2379: It looks like this interface is now designed for ZenFlow-like methods, which makes the integration easier and cleaner.
I'm not sure if DeepSpeed × PyTorch prefers keeping this fully PyTorch-aligned instead of adding framework-specific logic. Shall we add a TODO here?

Antlera: Otherwise, this code LGTM. Thanks!

tohtana: That's a good point. I wanted to quickly resolve this so we can pass the full CI test suite, but it makes the core part of the optimizer more Zenflow-specific. We probably shouldn’t cut corners for this. Let me close this for now, and then consider a more general approach.

PR #7797, opened 2026-01-20
Add Muon Optimizer Support to ZeRO3

PKUWZP:
Kudos to @pengdurice for the great work and discussions! 

We aim on adding Muon Optimizer to zero stage 3 in this PR: 

- The new momentum buffers are also partitioned like `self.fp32_partitioned_groups_flat` to save memory footprint. So, before the muon update, we need to perform `all_gather` on top of each data-parallel  group rank. The Muon updates of the parameters are also divided across the data-parallel ranks, and the results are all-gathered once all updates are complete. After the `all_gather`, the momentum buffers are partitioned and flatted again.

Next steps:

### DIFF
--- a/deepspeed/runtime/zero/stage3.py
+++ b/deepspeed/runtime/zero/stage3.py
@@ -10,6 +10,7 @@ import itertools
 from typing import Deque, Dict, Set, List, Tuple, Container, Optional
 from contextlib import contextmanager
 from dataclasses import dataclass, field
+from deepspeed.runtime.zero.muon.original_muon import muon_update
 
 from deepspeed import comm as dist
 from deepspeed.utils import groups, z3_leaf_parameter
@@ -330,10 +331,15 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
 
         self.reduce_scatter = reduce_scatter
 
+        self.use_muon = 'muon' in self.optimizer.__class__.__name__.lower()
+        self.save_muon_momentum_buffer_in_memory = ds_config.get('save_muon_momentum_buffer_in_memory', False)
+        if self.use_muon and self.reduce_scatter:
+            raise ValueError("Muon Optimizer and reduce scatter are not compatible.")
+        if self.use_muon and self.all2all_process_group is not None:
+            raise ValueError("Muon and all2all process group are not compatible.")
+
         self.dp_process_group = self.parameter_offload.dp_process_group
         self.sequence_parallel_size = groups._get_sequence_parallel_world_size()
-
-        self.all2all_process_group = all2all_process_group
 
         self.zero_quantized_nontrainable_weights = zero_quantized_nontrainable_weights
 
@@ -385,6 +391,8 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
         #a single 32-bit partition of the parallel partitioned parameters
         #that this process will update
         self.fp32_partitioned_groups_flat = []
+        if self.use_muon and self.save_muon_momentum_buffer_in_memory:
+            self.muon_momentum_buffer_partitioned_groups_flat = []
         self.next_swappable_fp32_partitioned_groups = []
 
         # number of elements per partition in each group
@@ -780,6 +788,15 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
         param_groups: List[List[Parameter]] = tuple(
             self._create_fp16_sub_groups(param_group["params"]) for param_group in fp16_param_groups)
 
+        if self.use_muon:
+            self.sub_groups_using_muon = []
+            for idx, param_group in enumerate(fp16_param_groups):
+                if getattr(param_group['params'][0], 'use_muon', False):
+                    self.sub_groups_using_muon.extend([True] * len(param_groups[idx]))
+                    self.muon_beta = param_group['momentum']
+                else:
+                    self.sub_groups_using_muon.extend([False] * len(param_groups[idx]))
+
         # bookkeeping related to param groups
         for param_group_idx, param_group in enumerate(param_groups):
             for sub_group in param_group:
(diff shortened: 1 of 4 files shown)


deepspeed/runtime/zero/stage3.py
```diff
@@ -1420,6 +1456,119 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+            params_size_offset += param.grad.numel()
+        # if optimizer is swappable, swap in the momentum buffer of the parameters that need to be updated using muon and then swap them out
+        # if optimizer is not swappable, find the momentum buffer of the parameters that need to be updated using muon in memory
+        for i in params_to_subgroup_maps:
+            if self._swappable_optimizer_subgroup(i) and not self.save_muon_momentum_buffer_in_memory:
+                self.optimizer_swapper.swap_in_optimizer_state(parameter=self.fp32_partitioned_groups_flat[i])
+                for idx, dest_offset in params_to_subgroup_maps[i]:
+                    momentum_buffer[idx] = self.optimizer.state[
+                        self.fp32_partitioned_groups_flat[i]]["momentum_buffer"].narrow(
+                            0, dest_offset, param.partition_numel()).clone()
```
PKUWZP, line 1499: @pengdurice Here is a bug. The variable param refers to the last parameter from the previous loop (`for param in self.ipg_buckets[...]`), not the parameter corresponding to idx. We should change it to`use_muon_params[idx].partition_numel()`.

deepspeed/runtime/zero/stage3.py
```diff
@@ -330,11 +331,16 @@ def _enforce_optimizer_offload():
...
+        self.save_muon_momentum_buffer_in_memory = ds_config.get('save_muon_momentum_buffer_in_memory', False)
+        if self.use_muon and self.reduce_scatter:
+            raise ValueError("Muon Optimizer and reduce scatter are not compatible.")
+        if self.use_muon and self.all2all_process_group is not None:
+            raise ValueError("Muon and all2all process group are not compatible.")
+
         self.dp_process_group = self.parameter_offload.dp_process_group
         self.sequence_parallel_size = groups._get_sequence_parallel_world_size()
 
-        self.all2all_process_group = all2all_process_group
```
PKUWZP, line 336: @pengdurice Question: where did we set up the `all2all_process_group`? It seems that it's never set.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1420,6 +1456,119 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+                world_sz = dist.get_world_size(self.dp_process_group)
+                rank = dist.get_rank(self.dp_process_group)
+                params = [use_muon_params[idx] for idx, _ in params_to_subgroup_maps[i]]
+                gathered_momentums = [gathered_params_momentums[idx] for idx, _ in params_to_subgroup_maps[i]]
+                # params_pad = params + [torch.empty_like(params[-1])] * (world_sz - len(params) % world_sz)
+                grads_pad = [param.grad for param in params
+                             ] + [torch.empty_like(params[-1].grad)] * (world_sz - len(params) % world_sz)
+                gathered_momentums_pad = gathered_momentums + [torch.empty_like(gathered_momentums[-1])
+                                                               ] * (world_sz - len(gathered_momentums) % world_sz)
+                for base_i in range(len(params))[::world_sz]:
```
PKUWZP, line 1530: @pengdurice There's a padding error here. When `len(params) % world_sz == 0`, this adds `world_sz` empty tensors instead of 0. Should we change it to: `(world_sz - len(params) % world_sz) % world_sz` ?

deepspeed/runtime/zero/stage3.py
```diff
@@ -330,11 +331,16 @@ def _enforce_optimizer_offload():
 
         self.reduce_scatter = reduce_scatter
 
+        self.use_muon = 'muon' in self.optimizer.__class__.__name__.lower()
```
PKUWZP, line 334: @pengdurice This is very fragile and purely depends on the naming conventions. Can we leverage `isinstance()` instead?

deepspeed/runtime/zero/stage3.py
```diff
@@ -330,11 +331,16 @@ def _enforce_optimizer_offload():
...
+        self.save_muon_momentum_buffer_in_memory = ds_config.get('save_muon_momentum_buffer_in_memory', False)
```
PKUWZP, line 335: @pengdurice Can we add `save_muon_momentum_buffer_in_memory` to the config schema and documented?

deepspeed/runtime/zero/stage3.py
```diff
@@ -1420,6 +1456,119 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+                param.grad.data.copy_(buffer_to_reduce.narrow(0, params_size_offset,
+                                                              param.grad.numel()).view_as(param.grad),
+                                      non_blocking=False)
+                momentum_buffer.append(None)
+                if i not in params_to_subgroup_maps:
+                    params_to_subgroup_maps[i] = []
+                params_to_subgroup_maps[i].append((idx, dest_offset))
+                idx += 1
+            params_size_offset += param.grad.numel()
+        # if optimizer is swappable, swap in the momentum buffer of the parameters that need to be updated using muon and then swap them out
```
PKUWZP, line 1491: @pengdurice This doubles NVMe I/O overhead. Can we consider consolidating into a single swap in/out cycle?

deepspeed/runtime/zero/stage3.py
```diff
@@ -1420,6 +1456,119 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+                )
+        # if there are parameters that need to be updated using muon
+        if momentum_buffer:
+            # all gather the momentum buffers of the parameters to the global buffer
+            # this is done since the momentum buffers are stored in partitions just like the params themselves
+            gathered_params_momentums = self._partitioned_buffers_all_gather(use_muon_params, momentum_buffer,
+                                                                             communication_data_type)
+            for i in params_to_subgroup_maps:
+                if self._swappable_optimizer_subgroup(i) and not self.save_muon_momentum_buffer_in_memory:
+                    self.optimizer_swapper.swap_in_optimizer_state(parameter=self.fp32_partitioned_groups_flat[i])
```
PKUWZP, line 1517: @pengdurice Again same thing here, can we consolidate the two swaps into one swap?

PKUWZP (commented):
@pengdurice Also two more comments: 

- It seems that we have excessive tensor allocations: Multiple torch.empty, torch.zeros, and .clone() calls create memory footprint pressure. Consider reusing buffers where possible.

- Synchronous `all_gather`: The distributed operations could potentially be overlapped with computation.

I think we need to re-work on the PR and let's take some times to refine the code.

PR #7798, opened 2026-01-20
Extending Muon Optimizer Support for ZeRO Stage 3

PKUWZP:
Authors: @pengdurice @PKUWZP 

We aim on adding Muon Optimizer to zero stage 3 in this draft PR: 

- The new momentum buffers are also partitioned like `self.fp32_partitioned_groups_flat` to save memory footprint. So, before the muon update, we need to perform `all_gather` on top of each data-parallel  group rank. The Muon updates of the parameters are also divided across the data-parallel ranks, and the results are all-gathered once all updates are complete. After the `all_gather`, the momentum buffers are partitioned and flatted again.

Next steps:

### DIFF
--- a/deepspeed/runtime/zero/stage3.py
+++ b/deepspeed/runtime/zero/stage3.py
@@ -1497,25 +1497,22 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
                 # swap-in once, keep resident through update + writeback
                 self.optimizer_swapper.swap_in_optimizer_state(parameter=self.fp32_partitioned_groups_flat[i])
                 if "momentum_buffer" not in self.optimizer.state.get(self.fp32_partitioned_groups_flat[i], {}):
-                    self._create_momentum_buffer(self.fp16_partitioned_groups_flat_numel[i],
-                                                 i,
+                    self._create_momentum_buffer(self.fp16_partitioned_groups_flat_numel[i], i,
                                                  self.fp32_partitioned_groups_flat[i].ds_id)
                 state_buffer = self.optimizer.state[self.fp32_partitioned_groups_flat[i]]["momentum_buffer"]
                 for param, dest_offset, _ in group_items:
-                    momentum_buffer.append(state_buffer.narrow(0, dest_offset,
-                                                               param.partition_numel()).clone())
+                    momentum_buffer.append(state_buffer.narrow(0, dest_offset, param.partition_numel()).clone())
             elif self.save_muon_momentum_buffer_in_memory:
                 state_buffer = self.muon_momentum_buffer_partitioned_groups_flat[i]
                 for param, dest_offset, _ in group_items:
-                    momentum_buffer.append(state_buffer.narrow(0, dest_offset,
-                                                               param.partition_numel()).clone())
+                    momentum_buffer.append(state_buffer.narrow(0, dest_offset, param.partition_numel()).clone())
             else:
                 raise ValueError(
                     "Invalid momentum buffer save mode, momentum buffer should be saved in memory or swapped in and out to nvme"
                 )
 
-            gathered_params_momentums = self._partitioned_buffers_all_gather(
-                params, momentum_buffer, communication_data_type)
+            gathered_params_momentums = self._partitioned_buffers_all_gather(params, momentum_buffer,
+                                                                             communication_data_type)
 
             world_sz = dist.get_world_size(self.dp_process_group)
             rank = dist.get_rank(self.dp_process_group)
@@ -1550,8 +1547,8 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
                 end_offset = start_offset + chunk_sz
                 if end_offset > param.grad.numel():
                     buffer_to_update = torch.zeros(chunk_sz, device=param.grad.device, dtype=param.grad.dtype)
-                    buffer_to_update[:param.grad.numel() - start_offset] = gathered_momentum.view(
-                        -1).data[start_offset:param.grad.numel()]
+                    buffer_to_update[:param.grad.numel() -
+                                     start_offset] = gathered_momentum.view(-1).data[start_offset:param.grad.numel()]
                 else:
                     buffer_to_update = gathered_momentum.view(-1).data[start_offset:end_offset]
                 if self._swappable_optimizer_subgroup(i) and not self.save_muon_momentum_buffer_in_memory:
@@ -1572,8 +1569,8 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
             for handle in grad_handles:
                 handle.wait()
             for param, _, params_size_offset in group_items:
-                buffer_to_reduce.narrow(0, params_size_offset,
-                                        param.grad.numel()).data.copy_(param.grad.view(-1), non_blocking=False)
+                buffer_to_reduce.narrow(0, params_size_offset, param.grad.numel()).data.copy_(param.grad.view(-1),
+                                                                                              non_blocking=False)
 
     @instrument_w_nvtx
     def __avg_scatter_contiguous_grads(self, buffer_to_reduce: Tensor,
@@ -1773,9 +1770,7 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
                                                           gradient_tensors=offload_fp32_gradients[i])
         return buffers
 
-    def _partitioned_buffers_all_gather(self,
-                                        params: List[Parameter],
-                                        buffers_to_allgather: List[Tensor],
+    def _partitioned_buffers_all_gather(self, params: List[Parameter], buffers_to_allgather: List[Tensor],
                                         communication_data_type: torch.dtype):
         """
         Allgather the partitioned buffers of the parameters to the global buffer.


deepspeed/runtime/zero/stage3.py
```diff
@@ -1394,6 +1426,101 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+                idx += 1
+            params_size_offset += param.grad.numel()
+        # if optimizer is swappable, swap in the momentum buffer of the parameters that need to be updated using muon and then swap them out 
+        # if optimizer is not swappable, find the momentum buffer of the parameters that need to be updated using muon in memory
+        for i in params_to_subgroup_maps:
+            if self._swappable_optimizer_subgroup(i) and not self.save_muon_momentum_buffer_in_memory:
+                self.optimizer_swapper.swap_in_optimizer_state(parameter=self.fp32_partitioned_groups_flat[i])
+                for idx, dest_offset in params_to_subgroup_maps[i]:
+                    momentum_buffer[idx] = self.optimizer.state[self.fp32_partitioned_groups_flat[i]]["momentum_buffer"].narrow(0, dest_offset, param.partition_numel()).clone()
+                self.optimizer_swapper.swap_out_optimizer_state(parameter=self.fp32_partitioned_groups_flat[i])
```
PKUWZP: @pengdurice  Here is a bug. The variable param refers to the last parameter from the previous loop (`for param in self.ipg_buckets[...]`), not the parameter corresponding to `idx`. We should change it to`use_muon_params[idx].partition_numel()`.

deepspeed/runtime/zero/stage3.py
```diff
@@ -309,12 +311,15 @@ def __init__(
...
+        self.use_muon = 'muon' in self.optimizer.__class__.__name__.lower()
+        self.save_muon_momentum_buffer_in_memory = ds_config.get('save_muon_momentum_buffer_in_memory', False)
+        if self.use_muon and self.reduce_scatter:
+            raise ValueError("Muon and reduce scatter cannot be used together")
+        if self.use_muon and self.all2all_process_group is not None:
+            raise ValueError("Muon and all2all process group cannot be used together")
         self.dp_process_group = self.parameter_offload.dp_process_group
         self.sequence_parallel_size = groups._get_sequence_parallel_world_size()
 
-        self.all2all_process_group = all2all_process_group
```
PKUWZP, line 336: @pengdurice Question: where did we set up the `all2all_process_group`? It seems that it's never set.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1394,6 +1426,101 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+                # in the case of large numbers of parameters, we distribute the workload across the ranks
+                # because muon update is a heavy operation
+                world_sz = dist.get_world_size(self.dp_process_group)
+                rank = dist.get_rank(self.dp_process_group)
+                params = [use_muon_params[idx] for idx, _ in params_to_subgroup_maps[i]]
+                gathered_momentums = [gathered_params_momentums[idx] for idx, _ in params_to_subgroup_maps[i]]
+                # params_pad = params + [torch.empty_like(params[-1])] * (world_sz - len(params) % world_sz)
+                grads_pad = [param.grad for param in params] + [torch.empty_like(params[-1].grad)] * (world_sz - len(params) % world_sz)
+                gathered_momentums_pad = gathered_momentums + [torch.empty_like(gathered_momentums[-1])] * (world_sz - len(gathered_momentums) % world_sz)
+                for base_i in range(len(params))[::world_sz]:
```
PKUWZP: @pengdurice There's a padding error here. When `len(params) % world_sz == 0`, this adds `world_sz` empty tensors instead of 0. Should we change it to: `(world_sz - len(params) % world_sz) % world_sz` ?

deepspeed/runtime/zero/stage3.py
```diff
@@ -309,12 +311,15 @@ def __init__(
         self.all2all_process_group = all2all_process_group
 
         self.reduce_scatter = reduce_scatter
-
+        self.use_muon = 'muon' in self.optimizer.__class__.__name__.lower()
```
PKUWZP: @pengdurice This is very fragile and purely depends on the naming conventions. Can we leverage `isinstance()` instead?

deepspeed/runtime/zero/stage3.py
```diff
@@ -309,12 +311,15 @@ def __init__(
...
+        self.save_muon_momentum_buffer_in_memory = ds_config.get('save_muon_momentum_buffer_in_memory', False)
```
PKUWZP, line 335: @pengdurice Can we add `save_muon_momentum_buffer_in_memory` to the config schema and documented?

deepspeed/runtime/zero/stage3.py
```diff
@@ -1394,6 +1426,101 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+                param_grad_offsets[param] = params_size_offset
+                # copy the gradients back to the params in the ipg bucket for the muon update
+                param.grad.data.copy_(buffer_to_reduce.narrow(0, params_size_offset, param.grad.numel()).view_as(param.grad), non_blocking=False)
+                momentum_buffer.append(None)
+                if i not in params_to_subgroup_maps:
+                    params_to_subgroup_maps[i] = []
+                params_to_subgroup_maps[i].append((idx, dest_offset))
+                idx += 1
+            params_size_offset += param.grad.numel()
+        # if optimizer is swappable, swap in the momentum buffer of the parameters that need to be updated using muon and then swap them out
```
PKUWZP: @pengdurice This doubles NVMe I/O overhead. Can we consider consolidating into a single swap in/out cycle?

deepspeed/runtime/zero/stage3.py
```diff
@@ -1394,6 +1426,101 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+            else:
+                raise ValueError("Invalid momentum buffer save mode, momentum buffer should be saved in memory or swapped in and out to nvme")
+        # if there are parameters that need to be updated using muon
+        if momentum_buffer:
+            # all gather the momentum buffers of the parameters to the global buffer
+            # this is done since the momentum buffers are stored in partitions just like the params themselves
+            gathered_params_momentums = self._partitioned_buffers_all_gather(use_muon_params, momentum_buffer, communication_data_type)
+            for i in params_to_subgroup_maps:
+                if self._swappable_optimizer_subgroup(i) and not self.save_muon_momentum_buffer_in_memory:
+                    self.optimizer_swapper.swap_in_optimizer_state(parameter=self.fp32_partitioned_groups_flat[i])
```
PKUWZP: @pengdurice Again same thing here, can we consolidate the two swaps into one swap?

PKUWZP (commented):
@pengdurice Also two more comments:

- It seems that we have excessive tensor allocations: Multiple torch.empty, torch.zeros, and .clone() calls create memory footprint pressure. Consider reusing buffers where possible.

- Synchronous all_gather: The distributed operations could potentially be overlapped with computation.

I think we need to re-work on the PR and let's take some times to refine the code.

delock:
Hi @pengdurice @PKUWZP, I have a question.  I saw there is an option that save momentum buffer in memory.  Yet for adam optimizer there is no such option.   Is that because for adam optimizer such needs is covered by zero offload, while for muon optimizer zero offload is not available yet, so this is used as temporary solution?  Thanks!

deepspeed/runtime/zero/stage3.py
pengdurice: yeah, good catch, fixed it.

deepspeed/runtime/zero/stage3.py
pengdurice: it is double assigned, there is an identical line nearly, so I deleted this one.

deepspeed/runtime/zero/stage3.py
pengdurice: good catch, fixed it.

deepspeed/runtime/zero/stage3.py
pengdurice: good catch, fixed it.

deepspeed/runtime/zero/stage3.py
pengdurice: The thought is moun_update is also time consuming and it could be that splitting the moun_update for each subgroup may also cause overhead. It's worth evaluating which is better;-)

pengdurice:
> Hi @pengdurice @PKUWZP, I have a question. I saw there is an option that save momentum buffer in memory. Yet for adam optimizer there is no such option. Is that because for adam optimizer such needs is covered by zero offload, while for muon optimizer zero offload is not available yet, so this is used as temporary solution? Thanks!

Hi thank you for your question, this is because for Adam optimizer, it is handled by its own code and since Adam only does element-wise operations, so no need to specially handle it.  in DeepSpeed/deepspeed/runtime/zero/muon/muon_optimizer.py 

>                          state["exp_avg"] = torch.zeros_like(p)
>                          state["exp_avg_sq"] = torch.zeros_like(p)
>                          state["step"] = 0
However, for muon update, we need cross element operations and thus need to do it in this file. It is not about offload. The buffer can be in GPU or CPU depending on if it is offload or not. Hope that answers your question!

pengdurice:
Sanity check: quick sanity check that the loss curves are aligned between stage2 and stage3 implementations, verifying the correctness of stage3 implementation.
Screenshot 2026-02-16 at 11 31 39 AM

deepspeed/runtime/zero/stage3.py
pengdurice: done, changed the logic to apply muon update per subgroup with the tradeoff a slight more overhead of muon_update.

deepspeed/runtime/zero/stage3.py
pengdurice: done, changed the logic to apply muon update per subgroup with the tradeoff a slight more overhead of muon_update.

deepspeed/runtime/zero/stage3.py
pengdurice: yes, added a section of muon optimizer to the config json md.

pengdurice:
> @pengdurice Also two more comments:
> 
> * It seems that we have excessive tensor allocations: Multiple torch.empty, torch.zeros, and .clone() calls create memory footprint pressure. Consider reusing buffers where possible.
> * Synchronous all_gather: The distributed operations could potentially be overlapped with computation.
> 
> I think we need to re-work on the PR and let's take some times to refine the code.

Thanks for the comments, for using buffers: the concern is since during each update (when the bucket is full), due to the dynamics of training and reconstruction of computation graphs for each forward / backward pass, we don't know before hands what parameters are used and gradients are generated, so we don't know what parameters need muon update and their sizes. That makes using yet another buffer not worth it IMO. If too large, that's a waste of memory, if too small, it may not be enough. Tracking and sizing can be good but that can be an engineering overkill. WDYT?
For hiding all_gather, I added handles to make them async with the heavy muon updates.

pengdurice:
Screenshot 2026-03-12 at 10 17 07 PM
Pretrain 450M GPT2 model with 6 different LR settings using wiki text dataset for 15K steps using ZeRO stage 3.

PKUWZP:
> Screenshot 2026-03-12 at 10 17 07 PM Pretrain 450M GPT2 model with 6 different LR settings using wiki text dataset for 15K steps using ZeRO stage 3.

Interesting results. Let me also test it on bigger models (e.g. Qwen3-8B) on 8 H200.

Issue #7625, opened 2025-10-08
[REQUEST] Allow fallback to torch.distributed.all_reduce for multi-node inference.


phalani-paladugu:
**Is your feature request related to a problem? Please describe.**
At present, DeepSpeed’s inference communication backend defaults to using the SHM-based operation torch.ops.deepspeed.inference_all_reduce_ for performing all_reduce operations when the shared memory (SHM) op is available. This default behavior, while effective for intra-node communication (e.g., dual-socket CPU machines), becomes problematic in multi-node inference scenarios.

The SHM implementation is currently tailored with x86-specific intrinsics, and as such, it’s incompatible with alternative CPU architectures like Arm.  Moreover,  based on my exploration, there’s no fallback mechanism to explicitly select a different communication backend (like torch.distributed.all_reduce) even when SHM is present. This limits flexibility for multi-node or non-x86 deployments where SHM may not work or be optimal.

**Describe the solution you'd like**
We propose modifying the inference_all_reduce method in `deepspeed/comm/torch.py` to introduce a runtime-aware mechanism that conditionally chooses between the SHM and PyTorch’s default distributed backends based on node locality. 

A sample implementation might look like this:

```python
def inference_all_reduce(self, tensor, op, group=None):
    all_local_ranks = True
    if group != None:
        world_size = torch.distributed.get_world_size(group=group)
        if (world_size > int(os.getenv("LOCAL_SIZE"))):
            all_local_ranks = False
             
    if not hasattr(torch.ops, 'deepspeed') or not hasattr(torch.ops.deepspeed, 'inference_all_reduce_') or not all_local_ranks:
        op = self._reduce_op(op)
        return torch.distributed.all_reduce(tensor=tensor, op=op, group=group, async_op=False)
    else:
        return torch.ops.deepspeed.inference_all_reduce_(tensor)
```

**Describe alternatives you've considered - but not tested**
- Add a runtime flag or env var (e.g., DS_SHM_COMM_ALL_REDUCE_OFF) to disable SHM-based collectives and fallback to torch.distributed ops.
- Introduce a parameter to deepspeed.init_inference() to control SHM usage, useful for hybrid systems (e.g., dual-socket or mixed-arch setups).
- Skip SHM build in setup.py for unsupported architectures like arm to avoid x86-intrinsic issues.

**Additional Context**
[...]

PR #7519, opened 2025-08-26
Add riscv64 cpu support in deepspeed_shm_comm op

heyujiao99:
This patch adds riscv64 support for the deepspeed_shm_comm operator，enabling DeepSpeed to perform CPU training/inference on RISCV64 hosts, for research purposes.  Based on the discussion in pull #7387 , this patch refactors some original code to support multiple CPU architectures.

Related tests have passed on x86 and RISC-V CPU, and I successfully ran Qwen2.5 on a RISC-V CPU,
```bash
(myenv) [root@openeuler-riscv64 DeepSpeed ]$ pytest tests/unit/comm/test_dist.py::TestDistInferenceAllReduce -vv
====================================================================== test session starts =======================================================================
platform linux -- Python 3.11.4, pytest-7.2.0, pluggy-1.6.0 -- /root/myenv/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default'
rootdir: /root/ecosystem/DeepSpeed/tests, configfile: pytest.ini
plugins: mock-3.14.1, hypothesis-6.135.14, forked-1.6.0
collected 3 items

tests/unit/comm/test_dist.py::TestDistInferenceAllReduce::test[dtype0] PASSED                                                                              [ 33%]
tests/unit/comm/test_dist.py::TestDistInferenceAllReduce::test[dtype1] PASSED                                                                              [ 66%]
tests/unit/comm/test_dist.py::TestDistInferenceAllReduce::test[dtype2] PASSED                                                                              [100%]

(myenv) root@ubuntu-2204:~/soft-working-dir/DeepSpeed# pytest tests/unit/comm/test_dist.py::TestDistInferenceAllReduce -vv
====================================================================== test session starts =======================================================================
platform linux -- Python 3.12.3, pytest-7.2.0, pluggy-1.6.0 -- /root/soft-working-dir/myenv/bin/python3
cachedir: .pytest_cache
rootdir: /root/soft-working-dir/DeepSpeed/tests, configfile: pytest.ini
plugins: forked-1.6.0
collected 3 items

tests/unit/comm/test_dist.py::TestDistInferenceAllReduce::test[dtype0] PASSED                                                                              [ 33%]
tests/unit/comm/test_dist.py::TestDistInferenceAllReduce::test[dtype1] PASSED                                                                              [ 66%]
tests/unit/comm/test_dist.py::TestDistInferenceAllReduce::test[dtype2] PASSED                                                                              [100%]

```

PR #7800, opened 2026-01-20
Enable shm_comm support for arm

phalani-paladugu:
This PR enables shared memory communication in single node  for arm hosts - #7625 

image


csrc/cpu/comm/arm64/shm.h
```diff
@@ -0,0 +1,102 @@
...
+#include <arm_neon.h>
+#include <stdint.h>
+#include <stddef.h>
+#include <cmath>
+
+// 128 bits = 16 bytes -> fits 8 fp16/bf16 or 4 fp32 elements.
+static int vector_length_in_bytes = 16;
+// When widening fp16/bf16 -> fp32, 4 elements fit in one 128-bit register.
+// Using 8 would require two 128-bit registers, so limit to 4.
+static int full_precision_elements_in_fixed_vector = 4;
```
xylian86: suggest to change to
`static constexpr int full_precision_elements_in_fixed_vector = 4;`

csrc/cpu/comm/arm64/shm.h
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0
+
+// DeepSpeed Team
```
xylian86, line 4: It may be helpful to document the minimum supported ARM architecture (e.g., the required NEON).

tests/unit/comm/test_dist.py
```diff
@@ -120,9 +120,10 @@ class TestDistAllReduce(DistributedTest):
         world_size = [1]
 
     def test(self):
-        x = torch.ones(1, 3).to(get_accelerator().device_name()) * (dist.get_rank() + 1)
+        num_elements = 128
```
xylian86: Could we keep both 3 and 128 to cover both small and vectorized cases?

csrc/cpu/comm/shm.cpp
```diff
@@ -154,6 +157,9 @@ void reduce_bf16_buffers(int start_elements, int num_elements, char* to_buffer,
 #if TARGET_RISCV
     size_t vl = __riscv_vsetvl_e16m1(num_elements);
     vector_length_in_bytes = vl * element_size;
+#elif TARGET_ARM
+    const int vl = full_precision_elements_in_fixed_vector;
+    vector_length_in_bytes = vl * element_size;
 #else
```
delock: Suggest to add a comment after `#else` to mention it is for x86_64.

```
#else // x86_64
```

It looks not very obvious that the `#else` part is for x86_64 instruction set.

delock:
Overall looks good to me.  I'm also glad that the refractory from @heyujiao99 in #7519 really helps alot!  Around 100 lines of code to support a new HW architecture, wow!  Kudos to @phalani-paladugu and @heyujiao99 !

csrc/cpu/comm/arm64/shm.h
phalani-paladugu: this is documented for now and will be handled at build time in a future PR.

csrc/cpu/comm/arm64/shm.h
phalani-paladugu: addressed in the latest commit.

tests/unit/comm/test_dist.py
phalani-paladugu: addressed in the latest commit.

csrc/cpu/comm/shm.cpp
phalani-paladugu: addressed in the latest commit.

phalani-paladugu:
Screenshot 2026-02-02 at 3 24 33 PM

phalani-paladugu:
Hi, is there a known issue with the cpu-torch-latest CI job? 
In the last two runs, the unit tests appear to complete successfully, but the job never finishes and eventually fails. 
I don’t see any test failures in the logs—only that the workflow seems to hang after the tests have passed.

sfc-gh-truwase:
@phalani-paladugu, yes, something seems to be wrong with `cpu-torch-latest`. Right now I see that it is running for over 2hrs, which is unexpected. 

@loadams do you know what could be wrong?

sfc-gh-truwase:
@phalani-paladugu FYI, @tohtana is investigating in the linked PR.

tohtana:
Hi @phalani-paladugu,
Based on the investigation in #7851, I think this [change](https://github.com/deepspeedai/DeepSpeed/pull/7851/changes/8ce74fe0094ad117416be9e5d84274913fb4d236) will fix the hang issue in the cpu test. It's not a fully clean fix, but at least we can unblock merging your PR. 
As I cannot push a commit to this PR (not editable by maintainers), can you merge the commit?

phalani-paladugu:
Hi @tohtana ,
Thanks for the investigation and for proposing the fix. I have added the change you suggested. It looks like a few checks are currently awaiting approval. Should be good to go once they have cleared.

tohtana:
All tests passed and the PR is merged now. This is a significant update. Thank you @phalani-paladugu for your contribution!

PR #7805, opened 2026-01-21
Fix ping-pong buffer index reset and removing redundant stream sync

### LINKED ISSUE #7804: [BUG] ZeRO-2 with `overlap_comm=True` and `contiguous_gradients=True` does not work properly
**Describe the bug**
When `overlap_comm=True` and `contiguous_gradients=True` are enabled, the gradient communication (AllReduce) does not overlap with backward pass computations as expected. Profiling trace logs reveal that AllReduce operations wait for computations to complete, indicating that the ping-pong buffer or stream synchronization logic is not functioning correctly.

**To Reproduce**
1. I have prepared a minimal training script to demonstrate the bug:
[ds_overlap_comm_test.py](https://github.com/user-attachments/files/24759932/ds_overlap_comm_test.py)

2. Download the script and run it using the DeepSpeed launcher in a multi-GPU environment (reproduced on 2x H20 GPUs):
   ```bash
   deepspeed ds_overlap_comm_test.py
   ```
3. Profiling logs will be generated in the `logs/` directory.
4. Open the trace JSON file using https://ui.perfetto.dev/.
5. Observe that there is no overlap; the gradient AllReduce operations are serialized after the computations.

**Expected behavior**
The AllReduce operation on one buffer (e.g., buffer 0) should occur simultaneously with the gradient computations for the next buffer (e.g., buffer 1).

**ds_report output**
torch install path ............... ['/usr/local/lib/python3.12/dist-packages/torch']
torch version .................... 2.7.0a0+79aa17489c.nv25.04
deepspeed install path ........... ['/mnt/project/user/xxxx/DeepSpeed/deepspeed']
deepspeed info ................... 0.18.5+5aa2d17d, 5aa2d17d, master
torch cuda version ............... 12.9
nvcc version ..................... 12.9
deepspeed wheel compiled w. ...... torch 2.7, cuda 12.9
shared memory (/dev/shm) size .... 235.00 GB

**Screenshots**
The following profiler trace shows a clear gap between communication and computation, confirming the lack of overlap:
Observed Behavior

For comparison, this is the expected behavior (ideal overlap):
Expected Behavior

- Docker Image: nvcr.io/nvidia/pytorch:25.04-py3
- Hardware: 2x NVIDIA H20 GPUs

**Launcher context**
Launched via `deepspeed` CLI.

**Docker context**
nvcr.io/nvidia/pytorch:25.04-py3

**Additional context**
After investigating the code in `deepspeed/runtime/zero/stage_1_and_2.py`, I have identified the root cause. The regression regarding communication overlap was introduced in PR #7371 (https://github.com/deepspeedai/DeepSpeed/pull/7371). While the additional two-stream synchronization in that PR fixes gradient corruption, it effectively disables the overlapping behavior.

The underlying issue causing the gradient corruption (which #7371 attempted to fix) was actually introduced in PR #6993 (https://github.com/deepspeedai/DeepSpeed/pull/6993). In that PR, `bucket.clear()` incorrectly resets the ping-pong buffer index to 0 at the end of `reduce_ipg_grads`. This logic disrupts the buffer index swapping mechanism within `reduce_independent_p_g_buckets_and_remove_grads`. I will be submitting a Pull Request with a fix shortly.

undersilence:
Fix #7804  and  #7188

After investigating the code in `deepspeed/runtime/zero/stage_1_and_2.py`, I have identified the root cause. The regression regarding communication overlap was introduced in PR #7371 (https://github.com/deepspeedai/DeepSpeed/pull/7371). While the additional two-stream synchronization in that PR fixes gradient corruption, it effectively disables the overlapping behavior.

The underlying issue causing the gradient corruption (which #7371 attempted to fix) was actually introduced in PR #6993 (https://github.com/deepspeedai/DeepSpeed/pull/6993). In that PR, `bucket.clear()` incorrectly resets the ping-pong buffer index to 0 at the end of `reduce_ipg_grads`. This logic disrupts the buffer index swapping mechanism within `reduce_independent_p_g_buckets_and_remove_grads`. 

To fix this, L121 in `deepspeed/runtime/zero/stage_1_and_2.py` should be removed to prevent resetting the buffer index. Additionally, the stream synchronization logic introduced in #7371 should be removed to restore the `overlap_comm=True` functionality.

### DIFF
--- a/deepspeed/runtime/zero/utils.py
+++ b/deepspeed/runtime/zero/utils.py
@@ -47,7 +47,7 @@ ZERO_SUPPORTED_OPTIMIZERS = [
 
 # Add MuonWithAuxAdam to supported list if muon is installed
 try:
-    from deepspeed.runtime.muon_optimizer import MuonWithAuxAdam
+    from deepspeed.runtime.zero.muon.muon_optimizer import MuonWithAuxAdam
     ZERO_SUPPORTED_OPTIMIZERS.append(MuonWithAuxAdam)
 except ImportError:
     pass


undersilence:
[ds_overlap_comm_test.py](https://github.com/user-attachments/files/24760841/ds_overlap_comm_test.py)

Running the script with `deepspeed ds_overlap_comm_test.py`, the gradient norm should align with standard results (`overlap_comm=False`) in all steps.

**Result with Fix (`overlap_comm=True`):**
```
...
Step 8/10: Loss = 0.979004, Grad Norm = 0.04842831566929817
Step 9/10: Loss = 0.974609, Grad Norm = 0.046479932963848114
Step 10/10: Loss = 1.013672, Grad Norm = 0.05202537402510643
```

**Baseline (`overlap_comm=False`):**
```
...
Step 8/10: Loss = 0.979004, Grad Norm = 0.04842831566929817
Step 9/10: Loss = 0.974609, Grad Norm = 0.046479932963848114
Step 10/10: Loss = 1.013672, Grad Norm = 0.05202537402510643
```
The results match perfectly.

To demonstrate that the gradient corruption (which #7371 tried to fix) was indeed caused by the broken buffer indexing from #6993:
If we **only** remove the synchronization from #7371 *without* removing the `bucket.clear()` index reset, the gradient norms diverge from the baseline:

**Incorrect Behavior (Sync Removed, Buffer Index Bug Remains):**
```
...
Step 8/10: Loss = 0.979004, Grad Norm = 0.048596370965242386
Step 9/10: Loss = 0.974121, Grad Norm = 0.047780971974134445
Step 10/10: Loss = 1.013672, Grad Norm = 0.046479482203722
```

tohtana:
Thanks @undersilence for the fix! This is truly crucial. Looks good for the single-dtype case.

One concern for multi‑dtype: `reduce_ipg_grads()` reduces all IPG buckets (one per comm dtype), but `reduce_independent_p_g_buckets_and_remove_grads()` only flips the ping‑pong index for the overflowed bucket. With multiple dtypes, other buckets can be refilled on the default stream while their reductions are still in flight on the reduction stream. 
  
I opened another [PR](https://github.com/undersilence/DeepSpeed/pull/1) on your PR branch to address this by launching an allreduce per dtype. Could you share your thoughts?

undersilence:
Thanks for the detailed review and for opening the PR!

You are right that since `reduce_ipg_grads()` was flushing all buckets while we were only handling the buffer index/sync for the current one, it would leave the other buckets in an unsafe state.

Isolating the allreduce by dtype is the correct fix here. I really appreciate your help in making this robust!

tohtana:
@undersilence Thank you for merging my PR! This is a great improvement. We really appreciate your contribution.

Issue #7672, opened 2025-11-05
[BUG] Rank index out of range during BWD when sp_size < world_size


NiuMa-1234:
I'm using the UlyssesSPAttentionHF,  and I found when the `sequence_parallel_size` is less than the `world_size (total gpus)`, the training would encounter an error when going backwards, as shown below:

```
 self.loss_scaler.backward(loss.float(), retain_graph=retain_graph)
  File "/opt/conda/lib/python3.10/site-packages/deepspeed/runtime/fp16/loss_scaler.py", line 65, in backward
    scaled_loss.backward(retain_graph=retain_graph)
  File "/opt/conda/lib/python3.10/site-packages/torch/_tensor.py", line 522, in backward
    torch.autograd.backward(
  File "/opt/conda/lib/python3.10/site-packages/torch/autograd/__init__.py", line 266, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
  File "/opt/conda/lib/python3.10/site-packages/torch/autograd/function.py", line 289, in apply
    return user_fn(self, *args)
  File "/opt/conda/lib/python3.10/site-packages/torch/distributed/nn/functional.py", line 343, in backward
    gx = torch.empty_like(grad_outputs[rank])
IndexError: tuple index out of range
```

And the error is likely caused by that, when executing `torch._AllGather`, the `gradoutput` only has `sp_world_size` items but each GPU uses `torch.distributed.get_rank()`  as index to choose its own grad from the gradoutput, and raise this index out of range error.

So is there a must to make sure the sequence_parallel_size be equal to the world_size?

Below is my test code, which is the same as the [deepspeed sp instruction](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/), and the only change is the number of num_gpus in deepspeed running command.

**To Reproduce**
```
# train.py
from deepspeed.runtime.sequence_parallel.ulysses_sp import UlyssesSPAttentionHF, UlyssesSPDataLoaderAdapter
from deepspeed.runtime.utils import move_to_device
from deepspeed.utils import groups
from torch import tensor
from transformers import AutoModelForCausalLM
import deepspeed
import deepspeed.comm as dist
import torch

model_name_or_path = '/lpai/volumes/base-agent-ali-sh-mix/lizr/chenxingru/xtuner-muon/Qwen__Qwen2.5-0.5B/main'
seq_length = 64
sequence_parallel_size = 2
micro_batch_size = 1

[...]

PR #7809, opened 2026-01-23
[Bugfix] Resolve Rank index out of range during BWD when sp_size < world_size in Ulysses

Flink-ddd:
### Description
This PR addresses Issue #7672.

When sequence_parallel_size is smaller than world_size (e.g., sp_size=2 on 4 GPUs) with PyTorch < 2.3, using torch.distributed.nn.functional.all_gather for loss aggregation triggers an IndexError: tuple index out of range during the backward pass. This is due to a known PyTorch issue where the backward hook accesses the global rank instead of the group rank.

### Solution
1. Regression Test & Workaround: Updated the regression test TestUlyssesLossBackward to implement a Weighted All-Reduce pattern.
- Before: all_gather -> manual sum (Vulnerable to rank indexing mismatch on older PyTorch).
- After: all_reduce(weighted_loss) / all_reduce(total_weight) (Robust and supports weighted averaging).
2. Runtime Warning: Added a version check (required_torch_version) in DeepSpeedEngine. It now logs a warning if Sequence Parallelism is enabled on PyTorch < 2.3, providing a link to the workaround test case.
3. Documentation: Updated ulysses-alst-sequence-parallelism.md with a note regarding legacy PyTorch versions and the recommended workaround.

### Verification
Added and verified the regression test tests/unit/sequence_parallelism/test_ulysses.py which now validates the weighted averaging logic.

**1. Reproduction (Before Fix)**
Confirmed IndexError crash on Rank 2/3 with sp_size=2 on a 4-GPU setup.
Screenshot 2026-01-23 at 23 53 42

**2. Verification (After Fix)**
Verified the fix using the regression test logic on 4x RTX A6000. The backward pass now completes successfully on all ranks without error.
Screenshot 2026-01-23 at 23 52 54

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -565,17 +565,23 @@ class DeepSpeedEngine(Module):
                                                                             with_kwargs=True)
 
     def __del__(self):
-        self.destroy()
+        try:
+            self.destroy()
+        except Exception as exc:
+            # Avoid destructor-time exceptions for partially initialized engines.
+            logger.debug("DeepSpeedEngine.__del__ cleanup skipped: %s", exc, exc_info=True)
 
     def destroy(self):
-        if self.optimizer is not None and hasattr(self.optimizer, 'destroy'):
-            self.optimizer.destroy()
+        optimizer = getattr(self, "optimizer", None)
+        if optimizer is not None and hasattr(optimizer, 'destroy'):
+            optimizer.destroy()
         if self.is_deepcompile_active():
             get_deepcompile_handle().cleanup()
         debug_clear_module_and_param_names()
 
-        if self.checkpoint_engine is not None and self.checkpoint_engine.is_decoupled():
-            self.checkpoint_engine.cleanup()
+        checkpoint_engine = getattr(self, "checkpoint_engine", None)
+        if checkpoint_engine is not None and checkpoint_engine.is_decoupled():
+            checkpoint_engine.cleanup()
 
     def _get_model_parameters(self):
         if self.autotuning_profile_model_info():
@@ -4384,7 +4390,7 @@ class DeepSpeedEngine(Module):
         return self._config.compile_config.deepcompile
 
     def is_deepcompile_active(self) -> bool:
-        return self._deepcompile_active
+        return getattr(self, "_deepcompile_active", False)
 
     @property
     def is_compiled(self) -> bool:
--- a/.github/workflows/aws-torch-latest-full.yml
+++ b/.github/workflows/aws-torch-latest-full.yml
@@ -133,4 +133,5 @@ jobs:
             --ignore=unit/launcher/test_user_args.py \
             --ignore=unit/runtime/zenflow \
             --ignore=unit/ops/adam/test_zf_torch_adam.py \
+            --ignore=unit/ops/deepspeed4science/test_DS4Sci_EvoformerAttention.py \
             --torch_ver=${{ env.TORCH_VER }} --cuda_ver=${{ env.CUDA_VER }}
--- a/deepspeed/runtime/zero/partition_parameters.py
+++ b/deepspeed/runtime/zero/partition_parameters.py
@@ -299,7 +299,13 @@ def get_all_subclasses(cls, include_root=True):
 @instrument_w_nvtx
 def free_param(param: Parameter) -> None:
     """Free underlying storage of a parameter."""
-    assert not param.ds_active_sub_modules, param.ds_summary()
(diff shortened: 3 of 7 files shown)


tohtana:
@Flink-ddd Thank you for opening this PR! I only see changes in tests. Did you miss committing some changes?

Flink-ddd:
Hi @tohtana , Thanks for the review. There are no missing commits. The issue reported in #7672 stems from using all_gather for loss aggregation in the user's training loop, rather than a bug within DeepSpeed's internal runtime.

Since we cannot patch user scripts directly, I submitted this regression test to:
1. Verify that the correct approach (using all_reduce) works stably when sp_size < world_size.
2. Prevent future regressions or confusion regarding this usage pattern.

but, pls tell me if you have other option.

tohtana:
Thank you for your clarification, @Flink-ddd!I It looks like a bug in PyTorch. In AllGather’s backward pass, we should use the *local rank* within the given process group. It appears this was fixed in v2.3.

* [v2.3](https://github.com/pytorch/pytorch/blob/v2.3.0/torch/distributed/nn/functional.py#L331): `rank = dist.get_rank(group=ctx.group)`
* [v2.2](https://github.com/pytorch/pytorch/blob/v2.2.0/torch/distributed/nn/functional.py#L331): `rank = dist.get_rank()`

As you said, we can’t force client code to implement loss calculation in a particular way. So I’m wondering whether we should simply add an assertion to check the PyTorch version when SP is enabled. We could also note that SP requires v2.3 or later in the document, even though the DeepSpeed code itself doesn’t have an issue with older versions.

It would still be good to add a regression test. One concern is that the all-reduce approach can’t implement weighted loss averaging, which is used in the [original example](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/).

What are your thoughts?

Flink-ddd:
Hi @tohtana  Thanks for the suggestion. I agree that simulating the weighted averaging pattern is better for real-world scenarios. I will update the test case to implement the weighted all-reduce pattern (reducing both the weighted loss and total weights separately) to address this.

tohtana:
Hi @Flink-ddd
Do you think we should support SP with v2.2 or older?

Flink-ddd:
Hi @tohtana , Yes, I believe we should. Many production environments and clusters are still pinned to PyTorch v2.1 or v2.2 due to CUDA driver constraints or stability requirements. maintaining support for SP on these older versions adds significant value to DeepSpeed's compatibility. This regression test ensures that we continue to support these users stably. However, it depends on your perspective. If you think it's unnecessary, we can set `assert torch.version >= 2.3` and then turn off this PR.

tohtana:
Okay, then let's keep the older versions. But adding a regression test doesn't prevent the strange error from confusing users. How about these?
- Update your new regression test for weighted loss averaging
- Add a note in the [tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/) about the allgather issue and a link to your new regression test (or the code snippet of your solution)
- Show a warning message during initialization when SP is enabled, telling users that all-gather might cause an issue, and include a link. We can also link to the new test case (or the relevant section of the tutorial).

Flink-ddd:
Hi @tohtana , Thanks for your aadvise, That sounds like a solid plan. I agree that adding a warning and updating the documentation will greatly improve the user experience for those on older PyTorch versions. I have already pushed these update.

deepspeed/runtime/engine.py
```diff
@@ -1410,6 +1410,14 @@ def _configure_distributed_model(self, model):
         self.expert_data_parallel_group = groups._get_expert_data_parallel_group_dict()
         self.sequence_parallel_size = groups._get_sequence_parallel_world_size()
         if self.sequence_parallel_size > 1:
+            # Inserted Warning for PyTorch < 2.3
```
tohtana, line 1419: Can we use [required_torch_version](https://github.com/deepspeedai/DeepSpeed/blob/c89e0db8e2b061e9a1f872731b4d7bf92f628231/deepspeed/utils/torch.py#L11)?

Flink-ddd: sure, good suggestion, done.

Flink-ddd:
Hi @tohtana , ready for review again, Thank you for suggestion and help.

Flink-ddd:
Hi @tohtana , all CI tests are successful now. Could you please help approve it when you are free? Thanks.

tohtana:
Merged. Thank you for your contribution, @Flink-ddd!

Issue #7819, opened 2026-01-27
[BUG] DeepSpeed ZeRO Stage-3 + CPU offloaded optimizer (CPUAdam) inconsistency metadata between subgroup


st-bang97:
**Describe the bug**
When using DeepSpeed ZeRO Stage-3 + CPU offloaded optimizer (CPUAdam), if there are 2+ subgroups, the optimizer state update within a single training step appears to apply inconsistent bias-correction values across subgroups.
Concretely, for the same global step (e.g., step=3), _bias_correction2 differs between the first subgroup call and subsequent subgroup calls inside the same step, which implies that later subgroups may effectively be using a slightly different optimizer state than the earlier subgroup(s) for the same step.

This can lead to “learning progress from another subgroup” being reflected inconsistently within one step (state not being step-consistent across subgroup partitions).

**Debug instrumentation details**
```
Add a debug helper in csrc/include/cpu_adam.h inside Adam_Optimizer:

inline void print_state() {
    printf("ds_adam_state: _betta1_t=%f, _betta2_t=%f _bias_correction1=%f, _bias_correction2=%f\n",
           _betta1_t, _betta2_t, _bias_correction1, _bias_correction2);
}
```
```
Add prints in csrc/adam/cpu_adam_impl.cpp inside ds_adam_step:
opt->IncrementStep(step, beta1, beta2);
opt->update_state(lr, epsilon, weight_decay, `bias_correction);
printf("ds_adam_step, optimizer_id=%d, step=%d, lr=%f, beta1=%f, beta2=%f, epsilon=%f, weight_decay=%f, bias_correction=%d\n",
       optimizer_id, step, lr, beta1, beta2, epsilon, weight_decay, bias_correction);
opt->print_state();
```

**To Reproduce**

Build DeepSpeed with debugging prints added (details above).
`git clone https://github.com/deepspeedai/DeepSpeedExamples/ `
`cd DeepSpeedExamples`
`training/DeepSpeed-SuperOffload/finetune_llama-8b_1gpu.sh zerooffload 1 `
Ensure the run configuration results in 2+ subgroups when using ZeRO-3 + CPU offload (I observe the issue only when subgroup count ≥ 2).
Observe ds_adam_step logs showing _bias_correction2 mismatch within the same step.

**Expected behavior**
[...]

PR #7820, opened 2026-01-27
Fix subgroup optimizer metadata inconsistency

st-bang97:
## Description
This PR addresses Issue https://github.com/deepspeedai/DeepSpeed/issues/7819.

When using **ZeRO Stage-3 + CPU offloaded optimizer (CPUAdam)** with **2+ subgroups**, `ds_adam_step` can be invoked multiple times within a single global optimizer step. Before this fix, the internal bias-correction-related state (e.g., `_betta2_t` → `_bias_correction2`) could become **inconsistent across subgroup invocations within the same step**, leading to subgroup-wise optimizer state divergence.

---

## Solution
Make `IncrementStep()` **step-consistent** under repeated calls in the same global step.

### Change
Update `IncrementStep()` to only advance or recompute state when `step != _step`, preventing subgroup-to-subgroup drift inside a single step.

```cpp
inline void IncrementStep(size_t step, float beta1, float beta2)
{
    if (beta1 != _betta1 || beta2 != _betta2) {
        _step = step;
        _betta1 = beta1;
        _betta2 = beta2;
        _betta1_t = std::pow(_betta1, step);
        _betta2_t = std::pow(_betta2, step);
    } else {
        if (step != _step) {
            _step++;
            if (_step != step) {
                _betta1_t = std::pow(_betta1, step);
                _betta2_t = std::pow(_betta2, step);
                _step = step;
            } else {
                _betta1_t *= _betta1;
                _betta2_t *= _betta2;
            }
        }
    }
}
```
1. Reproduction (Before Fix)

스크린샷 2026-01-27 194552
스크린샷 2026-01-27 194716

2. Verification (After Fix)
스크린샷 2026-01-27 200422
스크린샷 2026-01-27 200410


tohtana:
Hi @st-bang97,
Thank you for the contribution!
Sorry, I just noticed you've already addressed #7819. Please disregard my comment.

Overall this PR looks good to me. Here are some suggestions:
- Apply the same fix to the header for XPU
- Add some tests to avoid addition (you can reuse some from #7859 if you find them useful)
- Fix format

st-bang97:
Thanks for the review and feedback. I’m currently finalizing a paper submission and preparing for reviews, so I won’t be able to address this immediately. I’ll take a closer look and update the PR as soon as I have time.

PR #7821, opened 2026-01-27
Skip sequence parallel operations during eval

jp1924:
hi deepspeed!

[Make it possible to evaluate when using sequence parallel in HF Trainer #43517](https://github.com/huggingface/transformers/pull/43517)

I initially opened a PR in transformers to prevent errors that occur when running eval with deepspeed sequence parallel applied, but there was feedback that this should be handled in deepspeed rather than in transformers or accelerate, so I'm opening this PR here.

If you have any questions, feel free to ask.

PR #7823, opened 2026-01-28
Add option to disable SP in eval mode

tohtana:
(Proposing as an alternative of #7821)
This PR adds `disable_in_eval` parameter to `UlyssesSPAttentionHF` that bypasses sequence parallelism operations during evaluation.
See https://github.com/huggingface/transformers/pull/43517 and https://github.com/deepspeedai/DeepSpeed/pull/7821 for the context.


tohtana:
Closing as we applied this change to #7821.

PR #7826, opened 2026-01-30
Update version.txt to 0.18.6 after latest release

loadams:


tohtana:
@loadams Can we include #7805, #7821, and #7825? They won't take long.

loadams:
> @loadams Can we include #7805, #7821, and #7825? They won't take long.

@tohtana - I already did the release, this is after.  We can easily do another release though, no problem with that.  Just let me know once those PRs are in.  I see one was merged already

Issue #7677, opened 2025-11-06
Z1/2 should flatten tensors on gpu


stas00:
When loading huge models like Qwen3-30B on multiple gpus, z1/2 perform tensor flattening on cpu, with potentially 8 gpus competing with each other, which makes `deepspeed.initialize` excruciatingly slow while putting a huge load on cpu, making the whole node very sluggish.

repro (change `train_batch_size` to the world size (i.e. 1 if using 1 gpu)
```
# cat t.py
import sys, deepspeed, torch
from transformers import AutoModel, AutoConfig;
model = "Qwen/Qwen3-30B-A3B-Instruct-2507"
c = AutoConfig.from_pretrained(model)
c.num_hidden_layers=16 # cut the number of layers for faster loading
m = AutoModel.from_pretrained(model, config=c)
dsc = dict(
    train_micro_batch_size_per_gpu=1,
    train_batch_size=8,
    zero_optimization=dict(stage=2),
    bf16=dict(enabled=True),
    optimizer=dict(type="AdamW"),
    gradient_accumulation_steps=1,
)
deepspeed.initialize(model=m, config=dsc)
torch.distributed.destroy_process_group()
```

Note that the repro script shortens the model to 16 layers from the full 48 to have mercy on the person debugging it. (you can lower it more as well)

then run:
```
$ time deepspeed --num_gpus 8 t.py
real    20m35.321s
user    2244m0.271s
sys     88m31.586s
```

now exactly the same but with 1 gpu (change to `train_batch_size=1` above),
```
$ time deepspeed --num_gpus 1 t.py
real    3m11.001s
user    49m56.073s
sys     9m16.009s
```

2min of which are used to load the model weights. So we are looking at 19m35s vs 1m11s of deepspeed.initialize time - so a huge difference.

now this model is 48 layers so to initialize the full model more than 1h will be needed, even though loading the full model is only ~5min

Currently the flattening of tensors happens on cpu, which is a legacy implementation when gpu memory was very small. There is absolutely no reason not to run this on gpu, which would be probably at least 10x faster.

@tjruwase

ksugama:
Hey @stas00, I'm currently taking a crack at this issue. It seems that flattening on the GPU would require ~2x the model size in GPU memory. Would that still help for your use case?

stas00:
Oh, thank you for working on this, Kento.

[...]

PR #7828, opened 2026-02-03
Z1/2 init: flatten params on device

ksugama:
This PR addresses #7677 by flattening parameter tensors on the accelerators instead of the CPU during zero stage 1 and 2 initialization. This should alleviate CPU contention, with the caveat that the optimization is only used when there is enough VRAM to allocate a full copy of the parameter buffers.

On 8 x H100s and a Intel Xeon Platinum 8480+, profiling the initialization of DeepSpeed on 32 layers of `Qwen3-30B` with Z2 gives the following:

Old = ~382s
New = ~130s

-------------------------

If necessary, this optimization can be extended to allowed a tiered system that trades off VRAM space with performance, which might look like the following:

```
if enough VRAM for 2x model_size:
    naive flatten
else if enough VRAM for model_size / N:
    distributed flatten across N devices
else:
    flatten on CPU
```

The distributed flatten would involve each device flattening a portion of the parameters and performing an all-gather to assemble the full flattened model. See #7677 for original discussion.

### DIFF
--- a/deepspeed/moe/sharded_moe.py
+++ b/deepspeed/moe/sharded_moe.py
@@ -18,6 +18,7 @@ The file has been adapted from two fairscale files:
 from deepspeed.utils.timer import SynchronizedWallClockTimer
 from deepspeed.utils import logger
 from deepspeed.utils.bwc import bwc_tensor_model_parallel_world_size
+from deepspeed.utils.torch import jit_script_compat
 from typing import Callable, Dict, TYPE_CHECKING, Any, Optional, Tuple, Union
 
 import torch
@@ -157,7 +158,7 @@ def einsum(rule, a, b):
 # includes stateful caching logic which is incompatible with ONNX.
 
 
-@torch.jit.script
+@jit_script_compat
 def _capacity(gates: Tensor, capacity_factor: Tensor, min_capacity: Tensor) -> Tensor:
     # gates has shape of SE
     num_tokens = gates.shape[0]
@@ -170,12 +171,12 @@ def _capacity(gates: Tensor, capacity_factor: Tensor, min_capacity: Tensor) -> T
     return capacity
 
 
-@torch.jit.script
+@jit_script_compat
 def _top_idx(source, k):
     return torch.topk(source, k=k, dim=0)[1]
 
 
-@torch.jit.script
+@jit_script_compat
 def _one_hot_to_float(x, num_classes):
     return F.one_hot(x, num_classes=num_classes).float()
 
--- a/deepspeed/runtime/zero/mics_utils.py
+++ b/deepspeed/runtime/zero/mics_utils.py
@@ -11,12 +11,12 @@ from dataclasses import dataclass
 from typing import List
 
 import numpy as np
-import torch
 from torch import Tensor
 
 from deepspeed import comm as dist
 from deepspeed.accelerator import get_accelerator
 from deepspeed.utils import logger
+from deepspeed.utils.torch import jit_script_compat
 
 
 def _log_rank0(msg):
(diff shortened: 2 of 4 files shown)


sfc-gh-truwase:
@ksugama thanks for working on this. Looking forward to when it is ready for review.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -366,18 +367,29 @@ def _enforce_cpu_offload():
...
+            alignment = self.nccl_start_alignment_factor * dist.get_world_size(group=self.real_dp_process_group[i])
+            aligned_numel = int(math.ceil(orig_group_numel / alignment)) * alignment
+            param_dtype = self.bit16_groups[i][0].dtype
+            element_size = torch.tensor([], dtype=param_dtype).element_size()
+            flat_buffer_bytes = aligned_numel * element_size
 
             empty_cache()
-            see_memory_usage(f"After moving param group {i} to CPU", force=False)
+            accelerator = get_accelerator()
+            available_vram = accelerator.available_memory() if accelerator.is_available() else 0
```
ksugama, line 380: @stas00 I believe `available_memory()` eventually calls into nvidia-smi. Is this the foot gun you were warning about?

If it is, maybe this should be fixed in a different PR since that problem touches more than is related to these changes

ksugama:
@stas00 Hoping this alleviates the issue you were running into

deepspeed/runtime/zero/stage_1_and_2.py
stas00: Please remind me what issue I flagged that you are referring to?

`available_memory` is here:

https://github.com/deepspeedai/DeepSpeed/blob/a44fb58f134afc5399b29c154a5502e14272774c/accelerator/cuda_accelerator.py#L186-L194

ksugama: You mentioned in the discussion in #7677 that in order to check device memory, nvml should not be used and PyTorch API should be used instead

stas00: thank you for the refresher, Kento. I see my comment wasn't very precise. I was trying to say that it doesn't work on all devices.

We do want to use nvml in the general case when we want to see the total stats - since any gpu could be shared by multiple processes and torch's counters can only see its own process.

BTW, `torch.cuda.memory.mem_get_info` is an alternative to pynvml stats, but I think it'll fail just the same on gb10.

So I think the solution can be this:

```
 def available_memory(self, device_index=None): 
     this_process_view_available_mem = self.total_memory(device_index) - self.memory_allocated(device_index)
     if pynvml: 
         if device_index is None: 
             device_index = self.current_device() 
         handle = pynvml.nvmlDeviceGetHandleByIndex(self._get_nvml_gpu_id(device_index)) 
         try:
             # gb10 will fail this call
             info = pynvml.nvmlDeviceGetMemoryInfo(handle) 
             return info.free
         except:
             return this_process_view_available_mem
     else: 
         return this_process_view_available_mem
```

cc: @sfc-gh-truwase

ksugama:
@sfc-gh-truwase Would you be able to approve the CI workflows whenever you get the chance? I ran them on a lambda instance and they pass on my end

stas00:
> @stas00 Hoping this alleviates the issue you were running into

@ksugama, if it solves the problem described in the OP of https://github.com/deepspeedai/DeepSpeed/issues/7677 then it's fantastic.

I will try to find time to run this check myself.

stas00:
edit: my repro isn't measuring the right thing, since I see it's taking forever for Transformers to load the shards when doing multi-gpu. your PR is definitely making things load much faster.

I think there is some other inefficiency in HF Transformers's DS handling which makes the checkpoint shards load much slower on 8 gpus as compared to 1 gpu.

stas00 (approved):
Amazing work, thank you, @ksugama. 

So many people are going to have a much better experience with DS w/ huge models.

ksugama:
Incredible! Glad I can help. Thank you for reviewing

stas00:
https://x.com/StasBekman/status/2022354880049082658

PR #7829, opened 2026-02-03
Fix gradient is ready with z2

sfc-gh-truwase:
Correctly handle `ds_grad_is_ready` in ZeRO2

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1057,7 +1057,7 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
                 bucket.index = 1 - bucket.index
             self.report_ipg_memory_usage("In ipg_remove_grads after reduce_ipg_grads", param.numel())
 
-        # Don't modify reduction metadata when not ready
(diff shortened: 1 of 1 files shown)


stas00:
Interesting, thanks for finding this, Tunji, so then we don't need the condition at:

https://github.com/deepspeedai/DeepSpeed/blob/6b9cab1dd547cfd3173a54ce22020809c14eedf2/deepspeed/runtime/zero/stage_1_and_2.py#L1084

as your added check will prevent it from ever being False at l1084

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -1081,9 +1085,8 @@ def reduce_independent_p_g_buckets_and_remove_grads(self, param, i):
         assert grad_reduc is not None, f"rank {dist.get_rank()} - Invalid to reduce Param {param_id} with None gradient"
 
         # deal with a use-case of transient grads that will be generated in a loop for the same computation involving some model params - e.g. when performing a tiled memory calculation that shards the normal single sub-module call into a loop over a shards.
```
stas00: this comment is now dangling, perhaps move it to before the code you added?

PR #7831, opened 2026-02-05
Support new python 3.14 annotation handling

### LINKED ISSUE #7673: [BUG] Can not use deepspeed.layer.moe
**Describe the bug**
Importing the `deepspeed.layer.moe` throws raises this ValueError:
`  ValueError: Target parameter "qkv_w" not found in this layer. Valid targets are []`
from: https://github.com/deepspeedai/DeepSpeed/blob/e993fea38efe654592b956d1ab52e340bfbf9714/deepspeed/inference/v2/model_implementations/layer_container_base.py#L97-L99

and this ValueError:

`
ValueError: Must have wildcard (*) in source name for ParametrizedList mapping: self_attn.q_proj.weight
`
from: 
https://github.com/deepspeedai/DeepSpeed/blob/e993fea38efe654592b956d1ab52e340bfbf9714/deepspeed/inference/v2/model_implementations/layer_container_base.py#L125-L126

**To Reproduce**

Run this script with a install of current master branch:

```py
from deepspeed.moe.layer import MoE

print("Hello DeepSpeed!")
``` 

**Expected behavior**
I can normally use the moe layer :-) 

**ds_report output**
This also causes the same error?
Continuing instead of raising the error on both cases prevents any problems so here:

```sh
(moe) [lukef@luke HTYLLM-PG]$ ds_report
async_io ............... [NO] ....... [OKAY]
cpu_lion ............... [NO] ....... [OKAY]
dc ..................... [NO] ....... [OKAY]
 [WARNING]  Please specify the CUTLASS repo directory as environment variable $CUTLASS_PATH
evoformer_attn ......... [NO] ....... [NO]
 [WARNING]  FP Quantizer is using an untested triton version (3.5.0), only 2.3.(0, 1) and 3.0.0 are known to be compatible with these kernels
fp_quantizer ........... [NO] ....... [NO]
fused_lion ............. [NO] ....... [OKAY]
gds .................... [NO] ....... [OKAY]
inference_core_ops ..... [NO] ....... [OKAY]
cutlass_ops ............ [NO] ....... [OKAY]
ragged_device_ops ...... [NO] ....... [OKAY]
ragged_ops ............. [NO] ....... [OKAY]
 [WARNING]  sparse_attn requires a torch version >= 1.5 and < 2.0 but detected 2.9
 [WARNING]  using untested triton version (3.5.0), only 1.0.0 is known to be compatible
torch install path ............... ['/home/lukef/miniconda3/envs/moe/lib/python3.14/site-packages/torch']
torch version .................... 2.9.0+cu128
deepspeed install path ........... ['/home/lukef/Dokumente/GitHub/HTYLLM-PG/DeepSpeed/deepspeed']
deepspeed info ................... 0.18.2+2f232b9d, 2f232b9d, master
torch cuda version ............... 12.8
nvcc version ..................... 13.0
deepspeed wheel compiled w. ...... torch 0.0, cuda 0.0
shared memory (/dev/shm) size .... 15.27 GB
```

 - OS: 6.17.5-arch1-1
 - GPU: RTX 4070 Ti super (x1)
 - Python version: Python 3.14.0

sdvillal:
Current metaclasses for layers and parameters access annotations in a way that is incompatible with python 3.14+

See:
  - [Python 3.14 release notes](https://docs.python.org/3/whatsnew/3.14.html)
  - [Porting annotations](https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-porting-annotations)
  - [PEP649](https://peps.python.org/pep-0649/) and [PEP749](https://peps.python.org/pep-0749/)

This PR uses annotationlib from python 3.14 onwards and keeps backwards compatibility.

closes #7673
should unblock CF builds for py3.14 https://github.com/conda-forge/deepspeed-feedstock/pull/114

A question is, does deepspeed support officially 3.14 yet? Should we test it in CIs?

### DIFF
--- a/deepspeed/compat.py
+++ b/deepspeed/compat.py
@@ -38,7 +38,7 @@ def get_annotations(obj: Any) -> Dict[str, Any]:
     """
     if annotationlib:
         has_annotations = hasattr(obj, "__annotations__") or hasattr(obj, "__annotate__")
-        if not isinstance(obj, type) and not ismodule(object) and not callable(obj) and not has_annotations:
+        if not isinstance(obj, type) and not ismodule(obj) and not callable(obj) and not has_annotations:
             obj = type(obj)
         return annotationlib.get_annotations(obj)
     try:


deepspeed/compat.py
```diff
@@ -0,0 +1,36 @@
...
+# DeepSpeed Team
+
+from typing import Mapping, Any
+try:
+    import annotationlib  # python >= 3.14
+except ImportError:
+    annotationlib = None
+
+# Deal with annotations in python versions >=3.14. See:
+#   - Python 3.4 release notes: https://docs.python.org/3/whatsnew/3.14.html
```
tohtana: 3.14?

deepspeed/compat.py
```diff
@@ -0,0 +1,36 @@
...
+#     Porting annotations: https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-porting-annotations
+#   - PEP649: https://peps.python.org/pep-0649/
+#   - PEP749: https://peps.python.org/pep-0749/
+# Backwards compatible, applies best practices (use annotationlib) from python 3.14 onwards.
+
+
+def get_annotations_from_namespace(namespace: Mapping[str, object]) -> dict[str, str]:
```
tohtana: Should it be `dict[str, Any]`?

deepspeed/compat.py
```diff
@@ -0,0 +1,36 @@
...
+    if annotationlib:
+        annotate_func = annotationlib.get_annotate_from_class_namespace(namespace)
+        if annotate_func is not None:
+            return annotationlib.call_annotate_function(annotate_func, annotationlib.Format.VALUE)
+    return namespace.get("__annotations__", {})
+
+
+def get_annotations(obj: Any) -> dict[str, str]:
```
tohtana: Same as `get_annotations_from_namespace`, should this be `dict[str, Any]`?

tohtana:
Thank you @sdvillal!
I left some minor comments but don't see any actual issue.
We should definitely support Python 3.14. Let me quickly run the full test workflow with Python 3.14 to validate it works.

deepspeed/inference/v2/model_implementations/layer_container_base.py
```diff
@@ -150,7 +151,7 @@ def __call__(cls, *args, **kwargs):
         instance = cls.__new__(cls, *args, **kwargs)
         instance.__init__(*args, **kwargs)
 
-        for name, annotation in instance.__annotations__.items():
+        for name, annotation in get_annotations(instance).items():
```
tohtana, line 154: `get_annotations` needs to take a type like `get_annotations(type(instance))`.

deepspeed/inference/v2/model_implementations/parameter_base.py
```diff
@@ -91,7 +93,7 @@ def __call__(cls, *args, **kwargs):
         setattr(new_obj, "dest_param", None)
 
         # Initialize our dependences to None/empty `ParametrizedList`s
-        for name, annotation in new_obj.__annotations__.items():
+        for name, annotation in get_annotations(new_obj).items():
```
tohtana, line 96: `get_annotations` needs to take a type like `get_annotations(type(instance))`.

tohtana: This might be cleaner:
```
  def get_annotations(obj: Any) -> dict[str, Any]:
      if annotationlib:
          if not isinstance(obj, type):
              obj = type(obj)
          return annotationlib.get_annotations(obj)
      try:
          return obj.__annotations__
      except AttributeError:
          return {}
```

tohtana:
Sorry, can you fix the annotations to ``Dict[str, Any]``? It seems Python 3.8 throws an error with ``dict``.

deepspeed/compat.py
sdvillal: Fixed in upcoming commit :-)

deepspeed/inference/v2/model_implementations/parameter_base.py
sdvillal: Thanks for catching this up. I adopted your suggestion (see below), but added more conditions to adhere to the ```annotationlib.get_annotations``` contract with regard to input handling depending on type.

Alternatively, we could actually reduce the scope of our wrapper and require the caller to provide only types, which might be a more useful and easy to maintain constraint. What do you think?

sdvillal: If you think the current solution is fine, feel free to resolve the conversation.

deepspeed/compat.py
```diff
@@ -29,8 +29,16 @@ def get_annotations_from_namespace(namespace: Mapping[str, object]) -> Dict[str,
...
+    Retrieves annotations from a Python object.
+
+    In python >=3.14 this is a thin wrapper around the `annotationlib.get_annotations` function
+    with the added convenience to automatically infer the type for non module, class, function
+    or customly annotated objects.
+    """
     if annotationlib:
-        if not isinstance(obj, type):
+        has_annotations = hasattr(obj, "__annotations__") or hasattr(obj, "__annotate__")
+        if not isinstance(obj, type) and not ismodule(object) and not callable(obj) and not has_annotations:
```
tohtana: `ismodule(object)` should be `ismodule(obj)`.

deepspeed/inference/v2/model_implementations/parameter_base.py
tohtana: The approach looks good to me. Can you just fix the typo I mentioned in a new comment? Let's run the test merge it  unless any error happens.

deepspeed/compat.py
sdvillal: Ooopsie. Fixed.

PR #7832, opened 2026-02-05
Throw error when parameter is modified in GatheredParameters

tohtana:
#7817 added a test to verify that we throw an error when parameters are modified in `GatheredParameters` and `modifier_rank` is None. However, the PR just checks devices and doesn't detect modifications on parameters.
This causes an [error](https://github.com/deepspeedai/DeepSpeed/actions/runs/21653729382/job/62424014222) in our full test run.

This PR adds the detection of parameter modifications to properly throw an error.


deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -2322,11 +2323,32 @@ def __enter__(self):
...
+            self._param_versions = [(p, p.data.data_ptr(), p._version) for p in self.params]
 
     def __exit__(self, *exc):
         if not self.enabled:
             return
         if self.src_rank is None:
+            if self._param_versions:
+                modified_params = [
+                    p for p, data_ptr, version in self._param_versions
+                    if p.data.data_ptr() != data_ptr or p._version != version
```
sfc-gh-truwase, line 2342: I have two concerns about detecting unexpected in-place modifications
1. The perf overhead, especially from all-reduce call. 
2. The use of `_version` which is a pytorch internal API not recommended for developers.

How about if we enable this feature only in debug-mode of the GatheredParameters API? For example, reuse the `sanity_checks` config: https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/runtime/zero/config.py#L52

tohtana: Great suggestion, I modified to run the verification only when `enable_sanity_checks` is True.
I also confirmed the full tests pass on my dev environment.

sfc-gh-truwase (approved):

Issue #7837, opened 2026-02-09
[BUG] Gradient norm abnormally increases when training with ZeRO-0


wxhcore:
### Problem Description

When training with **ZeRO-0**, the **gradient norm abnormally increases**; however, under **identical configurations**, switching only the ZeRO stage to **ZeRO-1, ZeRO-2, or ZeRO-3** results in normal gradient behavior.

#### Gradient Norm Comparison

- **With ZeRO-0:**

  ![ZeRO-0 Gradient](https://github.com/user-attachments/assets/046117dd-8477-46c7-a9f8-7ee592d8453a)

- **With ZeRO-1/2/3:**

  ![ZeRO-1/2/3 Gradients](https://github.com/user-attachments/assets/469578ee-bb8a-4c6d-b89b-0059326c9d95)

> The only variable is the DeepSpeed ZeRO stage configuration. All other settings—including model architecture, data, optimizer, learning rate, and batch size—are identical.  
> This suggests that ZeRO-0 may have an issue in gradient computation, synchronization, or logging.

---

### Reproduction Steps

1. **Clone the repository and set up the environment**
   ```bash
   git clone https://github.com/wxhcore/bumblecore.git
   cd bumblecore
   conda create -n bumblecore_env python=3.10 -y
   conda activate bumblecore_env
   pip install -e .
   ```

2. **Modify the pretraining script to enable ZeRO-0**
   
   Edit `./scripts/pretrain.sh` to include:
   ```bash
   --deepspeed_config_path ./configs/deepspeed/ds_z0_config.json \
   ```

3. **Start training**
   ```bash
   bash scripts/pretrain.sh
   ```

4. **Observe gradient norm logs**  
   The gradient norm rapidly increases (as shown in the first image).

5. **Switch to ZeRO-1/2/3 configurations for comparison**  
   Replace the config file with `ds_z1_config.json`, `ds_z2_config.json`, or `ds_z3_config.json`. The gradients remain stable.

> All ZeRO configuration files are located at: [configs/deepspeed](https://github.com/wxhcore/bumblecore/tree/main/configs/deepspeed)

---

### Expected Behavior

Under identical conditions, gradient behavior should be consistent across different ZeRO stages.  
In particular, **ZeRO-0 (no sharding, equivalent to standard DDP)** should exhibit the same gradient dynamics as regular distributed training.

wxhcore:
Hi @tohtana,

[...]

PR #7839, opened 2026-02-09
Fix bf16 gradient norm divergence with ZeRO stage 0

tohtana:
Fixes: #7837

ZeRO-0 + bf16 has two bugs in `engine.py`: 
1. `FP16_UnfusedOptimizer` applies `dynamic_loss_scale` with `cur_scale=65536` but `engine.backward()` never scales the loss, so `step()` divides gradients by 65536
2. `_take_model_step` skips `zero_grad` for bf16 without ZeRO, causing gradient accumulation.

Fix: disable loss scaling for bf16 and remove the `zero_optimization()` gate on `zero_grad`.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -37,6 +37,7 @@ from deepspeed.runtime.zenflow.engine import (configure_zenflow, zenflow_step, i
                                               sync_zenflow_optimizer_lr)
 
 from deepspeed.runtime.fp16.fused_optimizer import FP16_Optimizer
+from deepspeed.runtime.fp16.loss_scaler import LossScaleConfig, LossScaleProfile
 from deepspeed.runtime.fp16.unfused_optimizer import FP16_UnfusedOptimizer
 from deepspeed.runtime.bf16_optimizer import BF16_Optimizer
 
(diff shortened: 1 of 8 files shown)


deepspeed/runtime/engine.py
```diff
@@ -1776,6 +1776,16 @@ def _configure_fp16_optimizer(self, optimizer, low_precision_dtype):
...
         dynamic_loss_args = self.dynamic_loss_scale_args()
         clip_grad = self.gradient_clipping()
+        use_dynamic_loss_scale = self.dynamic_loss_scale()
+        static_loss_scale = self.loss_scale()
+
+        # bf16 has the same dynamic range as fp32 (8-bit exponent), so loss
+        # scaling is unnecessary.  Force static scale=1 to avoid the mismatch
+        # where engine.backward() does not apply the optimizer's loss scaling
+        # yet FP16_UnfusedOptimizer.step() divides gradients by cur_scale.
+        if low_precision_dtype == torch.bfloat16:
```
sfc-gh-truwase: I previously handled this for fused_optimizer in #7515 but failed to handle unfused_optimizer. I think it would be cleaner to align the approaches. 

1. In #7515, we passed the lower_precision_type and explicitly handled the logic in fused_optimizer with minimal impact on engine.py. 
2. But this PR handles most of the logic in engine.py by repurposing loss_scale with minimal impact on unfused_optimizer. 

I prefer handling in the optimizers so that the logic can be modular and reusable by other clients besides engine. On the other hand, I think the common logic can for both fused_optimizer and unfused_optimizer can be refactored into a utility. Happy to chat more.

tohtana: Thank you for the feedback! It makes sense. Following the suggestion, I pushed a new commit to
- Introduce `LossScaleConfig` to consolidate many config items for loss scaling
- Create it in the engine and pass to optimizers (both `FP16_UnfusedOptimizer` and `FP16_Optimizer`)
- Remove duplicated code blocks to set loss scaler configs

Feel free to share your thoughts.

PR #7845, opened 2026-02-11
fix: replace deprecated fractions.gcd with math.gcd

Mr-Neutr0n:
## Bug
`fractions.gcd` was deprecated in Python 3.5 and removed in Python 3.9. This causes an `AttributeError` on Python 3.9+.

## Fix
Replaced `fractions.gcd` with `math.gcd` which is the standard replacement.


tohtana:
Thank you for the fix, @Mr-Neutr0n!

tohtana (approved):

PR #7851, opened 2026-02-13
Verify fix of CI hang (#7800)

tohtana:
Verify fix of CI hang (#7800)

### DIFF
--- a/deepspeed/comm/torch.py
+++ b/deepspeed/comm/torch.py
@@ -167,7 +167,11 @@ class TorchBackend(Backend):
         return torch.distributed.all_reduce(tensor=tensor, op=op, group=group, async_op=async_op)
 
     def inference_all_reduce(self, tensor, op, group=None):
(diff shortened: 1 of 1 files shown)


tohtana:
This PR confirms that `TestDistInferenceAllReduce` hangs in the CPU test of #7800. We still need to investigate the root cause, but here are the observations:
- The PR introduces the new test argument `num_elements==128`
- The default path (not RISCV or ARM) assumes x86_64 ([code](https://github.com/phalani-paladugu/DeepSpeed/blob/1ea484bc4a5916b7b14b05b393edb3af2f02526c/csrc/cpu/comm/shm.cpp))
- Given `num_elements==128`, the default path uses SIMD implementation for x86_64
- The x86_64 implementation hangs on AMD CPU platform when `world_size == 1`

This PR also confirms that we can avoid the hang by skipping allreduce when `world_size==1`. We might still have an issue when `world_size>1`, but I didn't see an issue on my local dev env. So I suggest skipping allreduce when `world_size==1` for now. It will unblock merging #7800.

PR #7854, opened 2026-02-16
Add news entry for DeepSpeed updates

PKUWZP:
Added a new news entry about DeepSpeed ZeRO++ support for LLM distillation work at LinkedIn.


tohtana (approved):
You could also add this to `docs/index.md`

PR #7859, opened 2026-02-18
Fix CPUAdam same-step subgroup drift in ZeRO-3 (#7819)

### LINKED ISSUE #7819: [BUG] DeepSpeed ZeRO Stage-3 + CPU offloaded optimizer (CPUAdam) inconsistency metadata between subgroup

tohtana:
Fixes #7819 

The root cause was non-idempotent CPUAdam step-state handling under ZeRO-3 subgroup updates: repeated calls at the same logical step could take different internal paths and produce slightly different bias-correction metadata.  

The fix makes same-step calls a no-op while preserving fast sequential updates, and adds regression tests covering both `step_subgroup()` and `step()` subgroup-style paths.
Validated with focused CPUAdam tests.


PKUWZP (approved):
@tohtana very well-written fix, a few optional work suggested here: 

  - Adding a test with non-sequential steps (e.g., jump from step 2 to step 5) to validate the pow fallback path after the refactor
  - Adding a test with beta changes mid-training to exercise the if-branch + fallthrough path

tohtana:
Sorry @PKUWZP, 
@st-bang97 already addressed this issue in #7820. Let's close this one and focus on #7820.

PR #7860, opened 2026-02-19
Merging AutoSP into DeepSpeed

neeldani:
# AutoSP: Unlocking Long-Context LLM Training Via Compiler-Based Sequence Parallelism

## Overview

AutoSP is a compiler optimization pass that shards inputs along the sequence dimension and enables Ulysses styled sequence parallelism while preventing graph breaks during `torch.compile()`. All the passes operate at the Torch IR on the forward graph.

### User-Facing Entry Point: `prepare_autosp_inputs()`

Users must explicitly call this function to prepare inputs for AutoSP compilation:

```python
def prepare_autosp_inputs(
    input_id: torch.Tensor,
    label_id: torch.Tensor,
    position_id: torch.Tensor = None,
    attention_mask: torch.Tensor = None,
    seq_dim: int = 1
) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]
```

**Purpose**: Symbolize sequence dimension and annotate tensors for identification.

**Operations**:
1. Mark sequence dimension as dynamic using `torch._dynamo.decorators.mark_dynamic()`
2. Attach metadata tags for tensor identification for auto-sharding:
   - `input_id.tag = constants.INPUT_ID_KEY`
   - `label_id.tag = constants.LABEL_ID_KEY`
   - `position_id.tag = constants.POSITION_ID_KEY` (if provided)

**Rationale**: PyTorch's FX graph tracer requires explicit annotation of data-dependent dimensions. Marking the sequence dimension as dynamic prevents symbolic shape propagation from losing dimension information through reshape/view operations.

### Pass 1: `pass_shard_seq_dim()`

**Objective**: Propagate sharded sequence dimension to all consumers.

**Algorithm**:
1. Extract symbolic sequence dimension from `input_id` shape metadata
2. Locate the symbolic dimension node in the FX graph
3. Create a floor-divide node: `seq_dim / world_size`
4. Perform worklist-based graph traversal to find all direct and indirect consumers of input node, label node and position id node.
5. Replace symbolic dimension references with sharded dimension in consumer nodes

**Rationale**: Reshapes and views that consume the sequence dimension as an argument do not get updated during propagation of symbolic shapes. This pass explicitly rewires the computation graph to use sharded dimensions, enabling proper shape inference downstream.

### Pass 2: `pass_shard_input_ids()` / `pass_shard_label_ids()` / `pass_shard_position_ids()`

**Objective**: Insert slicing operations after input tensors.

**Implementation**: Call `shard_tensor_node()` utility which inserts slice operations. Each rank retains only the portion of the tensor corresponding to its sequence partition and drops the remaining buffer.

**Note on `attention_mask`**: Not sharded because it applies to the full sequence length, not the partitioned dimension.

### Pass 3: `pass_insert_attention_all_to_all()`

**Objective**: Insert all-to-all collectives around attention (Ulysses styled) to avoid graph breaks during compilation.

**Algorithm**:
1. Identify all SDPA (Scaled Dot-Product Attention) nodes in the graph
2. For each SDPA node with inputs Q, K, V, after each of Q, K, V: insert A2A scatter heads (dim=1), gather sequence (dim=2)
3. Insert A2A after thre attention output O: scatter sequence (dim=2), gather heads (dim=1)

**Graph Rewrite Example**:
```
Q [B, N, S/P, H] --A2A(scatter_heads,gather_seq)--> [B, N/P, S, H]
K [B, N, S/P, H] --A2A(scatter_heads,gather_seq)--> [B, N/P, S, H]
V [B, N, S/P, H] --A2A(scatter_heads,gather_seq)--> [B, N/P, S, H]
                     |
                    SDPA
                     |
O [B, N/P, S, H] --A2A(scatter_seq,gather_heads)--> [B, N, S/P, H]
```

**Current support**: Currently only supports `torch.nn.functional.scaled_dot_product_attention()`. Composite attention patterns require additional pattern matching logic.

### Pass 4: `pass_propagate_shapes()`

**Objective**: Compute static shapes for all nodes using fake tensor execution.

**Implementation**:
1. Create `ShapeEnv` for symbolic dimension tracking
2. Construct `FakeTensorMode` with the shape environment
3. Execute `FakeTensorProp.propagate()` to compute shape metadata

### Pass 5: `pass_canonicalize()`

**Objective**: Finalize graph representation.

**Operations**:
1. `eliminate_dead_code()`: Remove unused operations
2. `lint()`: Validate graph structure
3. `recompile()`: Regenerate compiled representation

## Execution Order

```
prepare_autosp_inputs()
    ↓
pass_shard_seq_dim
    ↓
pass_shard_input_ids
    ↓
pass_shard_label_ids
    ↓
pass_shard_position_ids
    ↓
pass_insert_attention_all_to_all
    ↓
pass_propagate_shapes
    ↓
pass_canonicalize
    ↓
pass_selective_activation_checkpointing

```

## Memory savings
AutoSP adds some heuristics to torch.compile's partitioniner which splits the joint graph into the forward and backward graph. Matmul and related ops are not checkpointed since recomputing them is much cheaper compared to the attention op, while reducing the peak active memory.

## Reducing gradients across ranks
AutoSP requires an all-reduce to reduce the gradients across ranks. This is automatically called by DeepSpeed's engine [here](https://github.com/deepspeedai/DeepSpeed/blob/93524c8931799a7631a2321d7ef4afaff6b6e54b/deepspeed/runtime/engine.py#L2433)

## Known Limitations

1. **Attention Pattern Matching**: Only `torch.nn.functional.scaled_dot_product_attention()` is supported. Fused attention implementations require pattern-specific handling.
2. **No Graph Break Requirement**: AutoSP will fail if there are graph breaks because use-def chains are lost and it becomes tricky to propagate auto-sharding information across graph modules.

## Example 

DeepSpeedExample PR: https://github.com/deepspeedai/DeepSpeedExamples/pull/999

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,7 +7,7 @@
 
 - All commits MUST have a `Signed-off-by` line (use `--signoff`). Get the name and email from `git config user.name` / `git config user.email`.
 - Formatting: yapf (column_limit=119, `.style.yapf`) + flake8 (`.flake8`).
-- Always verify changed files pass pre-commit checks before committing. Config: `.pre-commit-config.yaml`.
+- Always verify changed files pass pre-commit checks before committing: `pre-commit run --files <changed_files>`. Only check modified files, not the entire codebase. Config: `.pre-commit-config.yaml`.
 - `check-torchdist` hook: NEVER directly import torch's distributed module. Use `import deepspeed.comm as dist` instead.
 - New files require license header:
   ```
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -7,7 +7,7 @@
 
 - All commits MUST have a `Signed-off-by` line (use `--signoff`). Get the name and email from `git config user.name` / `git config user.email`.
 - Formatting: yapf (column_limit=119, `.style.yapf`) + flake8 (`.flake8`).
-- Always verify changed files pass pre-commit checks before committing. Config: `.pre-commit-config.yaml`.
+- Always verify changed files pass pre-commit checks before committing: `pre-commit run --files <changed_files>`. Only check modified files, not the entire codebase. Config: `.pre-commit-config.yaml`.
 - `check-torchdist` hook: NEVER directly import torch's distributed module. Use `import deepspeed.comm as dist` instead.
 - New files require license header:
   ```
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@
 
 ## Latest News
 
+* [2026/03] DeepSpeed Team gave a tutorial at ASPLOS 2026 titled ["Building Efficient Large-Scale Model Systems with DeepSpeed: From Open-Source Foundations to Emerging Research" ](https://supercomputing-system-ai-lab.github.io/events/asplos2026-llm-tutorial/index.html)
+
+* [2026/03] [Our SuperOffload work received an Honorable Mention for the ASPLOS 2026 Best Paper Award](https://dl.acm.org/doi/10.1145/3760250.3762217)
+
 * [2025/12] [DeepSpeed Core API updates: PyTorch-style backward and low-precision master states](https://github.com/deepspeedai/DeepSpeed/blob/master/blogs/core_api_update/README.md)
 
 * [2025/11] [DeepSpeed ZeRO++ powers large-scale distillation training of LLMs for Recommendation Systems at LinkedIn](https://aclanthology.org/2025.emnlp-industry.119/)
--- a/deepspeed/ops/adam/cpu_adam.py
+++ b/deepspeed/ops/adam/cpu_adam.py
@@ -236,14 +236,12 @@ class DeepSpeedCPUAdam(torch.optim.Optimizer):
                     f"CPUAdam param is on {param.device} and must be 'cpu', "
                     f"make sure you enabled 'offload_optimizer': 'cpu' in your ZeRO config.")
 
-                # Decrement step count
-                subgroup_state['step'] -= 1
-
-                # Extract hyperparameters
                 beta1, beta2 = group['betas']
 
                 self.ds_opt_adam.adam_rollback(self.opt_id, subgroup_state['step'], group['lr'], beta1, beta2,
                                                group['eps'], group['weight_decay'], group['bias_correction'],
                                                param.data, param.grad.data, subgroup_state['exp_avg'],
                                                subgroup_state['exp_avg_sq'])
+
+                subgroup_state['step'] -= 1
         return loss
--- a/deepspeed/runtime/superoffload/superoffload_stage3.py
+++ b/deepspeed/runtime/superoffload/superoffload_stage3.py
@@ -5,7 +5,6 @@
 
 import time
 import torch
-from collections import deque
 from typing import List
 
 from deepspeed.runtime.superoffload.superoffload_utils import SuperOffloadCPUOptimizer, TaskKeys, ResultKeys, EventTypes
@@ -16,6 +15,13 @@ from deepspeed.utils import logger
 from deepspeed.accelerator import get_accelerator
 
 OPTIMIZER_STEP_TIMER = 'optimizer_step'
+
+
+def _validate_superoffload_accelerator():
+    """Validate that the current accelerator is compatible with SuperOffload."""
+    accelerator = get_accelerator()
+    assert accelerator.device_name() == 'cuda', (
+        f"SuperOffload only supports NVIDIA CUDA GPUs, but found accelerator '{accelerator.device_name()}'.")
 
 
 class SuperOffloadOptimizer_Stage3(DeepSpeedZeroOptimizer_Stage3):
@@ -29,24 +35,26 @@ class SuperOffloadOptimizer_Stage3(DeepSpeedZeroOptimizer_Stage3):
         ds_config,
         **kwargs,
     ):
+        _validate_superoffload_accelerator()
 
         self.sub_group_to_param_num = {}
-        self.params_in_ipg_bucket_buffer = deque()
-        self._cur_bucket_index = -1
+        self.sub_group_grad_partition_counts = {}
         self.async_cpuadam_num = 0
         self.max_grad_numel = 0
 
         super().__init__(module, init_optimizer, param_names, timers, ds_config, **kwargs)
 
-        optimizer_config = {
-            "lr": self.optimizer.param_groups[0]["lr"],
-            "betas": self.optimizer.param_groups[0]["betas"],
-            "eps": self.optimizer.param_groups[0]["eps"],
-            "weight_decay": self.optimizer.param_groups[0]["weight_decay"],
-            "amsgrad": self.optimizer.param_groups[0]["amsgrad"]
-        }
+        optimizer_configs = []
+        for pg in self.optimizer.param_groups:
+            optimizer_configs.append({
+                "lr": pg["lr"],
+                "betas": pg["betas"],
+                "eps": pg["eps"],
+                "weight_decay": pg["weight_decay"],
+                "amsgrad": pg["amsgrad"],
+            })
         cpuadam_cores_perc = kwargs.get("cpuadam_cores_perc", 0.8)
-        self.superoffload_cpu_optimizer = SuperOffloadCPUOptimizer(optimizer_config=optimizer_config,
+        self.superoffload_cpu_optimizer = SuperOffloadCPUOptimizer(optimizer_config=optimizer_configs,
                                                                    cpuadam_cores_perc=cpuadam_cores_perc,
                                                                    max_grad_numel=self.max_grad_numel)
 
@@ -56,6 +64,9 @@ class SuperOffloadOptimizer_Stage3(DeepSpeedZeroOptimizer_Stage3):
         sub_group_size = self.sub_group_size
 
         if sub_group_size is None or sub_group_size >= params_group_numel:
+            global_idx = len(self.sub_group_to_param_num)
+            self.sub_group_to_param_num[global_idx] = len(params_group)
+            self.max_grad_numel = max(self.max_grad_numel, params_group_numel)
             return [params_group]
 
         sub_groups = []
@@ -69,7 +80,8 @@ class SuperOffloadOptimizer_Stage3(DeepSpeedZeroOptimizer_Stage3):
             if local_sub_group_size >= sub_group_size or id(param) == id(params_group[-1]):
                 self.max_grad_numel = max(self.max_grad_numel, local_sub_group_size)
                 sub_groups.append(sub_group)
-                self.sub_group_to_param_num[len(sub_groups) - 1] = len(sub_group)
+                global_idx = len(self.sub_group_to_param_num)
+                self.sub_group_to_param_num[global_idx] = len(sub_group)
 
                 sub_group = []
                 local_sub_group_size = 0
@@ -93,43 +105,16 @@ class SuperOffloadOptimizer_Stage3(DeepSpeedZeroOptimizer_Stage3):
             step_with_gradscaler(self.backup_optimizer)
             self.backup_optimizer.param_groups[param_group_id]['params'] = []
 
-    def reduce_independent_p_g_buckets_and_remove_grads(self, param):
-        comm_dtype = self.get_param_comm_dtype(param)
-        bucket = self.ipg_buckets[comm_dtype]
-        i, _, _ = self.grad_position[self.get_param_id(param)]
-
-        if len(bucket.params) == 0:
-            self._cur_bucket_index = i
-            if getattr(param, "ds_grad_is_ready", True):
-                self._DeepSpeedZeroOptimizer_Stage3__add_grad_to_ipg_bucket(param)
-
-            # If this is a single-parameter sub-group, reduce immediately
-            if self.sub_group_to_param_num[self._cur_bucket_index] == 1:
-                self._DeepSpeedZeroOptimizer_Stage3__reduce_and_partition_ipg_grads(comm_dtype)
-
-        elif i != self._cur_bucket_index:
-            # Parameter belongs to different sub-group, buffer it
-            self.params_in_ipg_bucket_buffer.append(param)
-        else:
-            # Parameter belongs to current bucket
-            if getattr(param, "ds_grad_is_ready", True):
-                self._DeepSpeedZeroOptimizer_Stage3__add_grad_to_ipg_bucket(param)
-
-            # Check if bucket is complete
-            if self.sub_group_to_param_num[self._cur_bucket_index] == len(bucket.params):
-                self._DeepSpeedZeroOptimizer_Stage3__reduce_and_partition_ipg_grads(comm_dtype)
-
-                # Process buffered parameters
-                while self.params_in_ipg_bucket_buffer:
-                    buffered_param = self.params_in_ipg_bucket_buffer.popleft()
-                    ci, _, _ = self.grad_position[self.get_param_id(buffered_param)]
-                    self._cur_bucket_index = ci
-                    if getattr(buffered_param, "ds_grad_is_ready", True):
-                        self._DeepSpeedZeroOptimizer_Stage3__add_grad_to_ipg_bucket(buffered_param)
+    @instrument_w_nvtx
+    def independent_gradient_partition_epilogue(self):
+        super().independent_gradient_partition_epilogue()
+        self.sub_group_grad_partition_counts.clear()
 
     @instrument_w_nvtx
     def _reassign_or_swap_out_partitioned_parameters(self, sub_group_id):
         if self.subgroup_to_device[sub_group_id] == 'cpu':
+            self.fp16_partitioned_groups_flat[sub_group_id].data.copy_(
+                self.fp32_partitioned_groups_flat[sub_group_id].data)
             self._unflatten_partitioned_parameters(sub_group_id)
             return
 
@@ -147,62 +132,61 @@ class SuperOffloadOptimizer_Stage3(DeepSpeedZeroOptimizer_Stage3):
 
     @instrument_w_nvtx
     def partition_grads(self, params_to_release: List[Parameter], grad_partitions: List[Tensor]) -> None:
-        # print("[DEBUG] partition_grads called")
-        buffers = []
-        device_buffers = {}
-        buffer_numel_min = {}
-        buffer_numel_max = {}
+        completed_sub_groups = []
 
         for param, grad_partition in zip(params_to_release, grad_partitions):
             i, dest_offset, _ = self.grad_position[self.get_param_id(param)]
 
-            if self.is_gradient_accumulation_boundary:
-                self.norm_for_param_grads[self.get_param_id(param)] = self._constant_buffered_norm2(grad_partition)
-
-            buffer_numel = grad_partition.numel()
-            buffers.append(grad_partition)
-
-            if i not in device_buffers:
-                device_buffers[i] = []
-            device_buffers[i].append(grad_partition)
-
-            if i not in buffer_numel_min:
-                buffer_numel_min[i] = dest_offset
-                buffer_numel_max[i] = dest_offset + buffer_numel
+            # Accumulate gradient into the grad_buffer, mirroring base class logic
+            grad_buffer = self._DeepSpeedZeroOptimizer_Stage3__param_id_to_grad_partition[param.ds_id].narrow(
+                0, 0, grad_partition.numel())
+            if self.micro_step_id == 0:
+                grad_buffer.copy_(grad_partition, non_blocking=True)
+                grad_buffer = grad_buffer.to(grad_partition.device, non_blocking=True)
+            elif get_accelerator().on_accelerator(grad_buffer):
+                grad_buffer.add_(grad_partition.to(self.gradient_accumulation_dtype).view(grad_buffer.shape))
             else:
-                buffer_numel_min[i] = min(buffer_numel_min[i], dest_offset)
-                buffer_numel_max[i] = max(buffer_numel_max[i], dest_offset + buffer_numel)
+                cuda_grad_buffer = grad_buffer.to(grad_partition.device, non_blocking=True)
+                cuda_grad_buffer.add_(grad_partition.to(self.gradient_accumulation_dtype).view(cuda_grad_buffer.shape))
+                grad_buffer.copy_(cuda_grad_buffer, non_blocking=True)
+                grad_buffer = cuda_grad_buffer
 
-        if self.is_gradient_accumulation_boundary:
-            for i in buffer_numel_min.keys():
+            if self.is_gradient_accumulation_boundary:
+                self.norm_for_param_grads[self.get_param_id(param)] = self._constant_buffered_norm2(grad_buffer)
+
                 fp32_grad_tensor = self.fp32_partitioned_groups_flat[i].grad.narrow(
-                    0, buffer_numel_min[i], buffer_numel_max[i] - buffer_numel_min[i])
-                concatenated_buffer = torch.cat(device_buffers[i], dim=0).to(dtype=self.master_weights_and_grads_dtype)
+                    0, dest_offset, grad_buffer.numel())
+                fp32_grad_tensor.copy_(grad_buffer.to(dtype=self.master_weights_and_grads_dtype), non_blocking=True)
 
-                if self.subgroup_to_device[i] == 'cpu':
-                    # Trigger asynchronous CPU optimization
+            self.sub_group_grad_partition_counts[i] = self.sub_group_grad_partition_counts.get(i, 0) + 1
+            if self.sub_group_grad_partition_counts[i] == self.sub_group_to_param_num[i]:
+                completed_sub_groups.append(i)
+
+        if self.is_gradient_accumulation_boundary and completed_sub_groups:
+            get_accelerator().current_stream().synchronize()
+            for i in completed_sub_groups:
+                if self.subgroup_to_device[i] == 'cpu' and not self.clip_grad:
                     param_group_id = self.sub_group_to_group_id[i]
                     fp32_param = self.fp32_partitioned_groups_flat[i]
+                    current_lr = self.optimizer.param_groups[param_group_id]['lr']
 
-                    self.superoffload_cpu_optimizer.async_step(param_group_id, i, fp32_param.data,
-                                                               concatenated_buffer.data)
+                    self.superoffload_cpu_optimizer.async_step(param_group_id,
+                                                               i,
+                                                               fp32_param.data,
+                                                               fp32_param.grad.data,
+                                                               lr=current_lr)
                     self.async_cpuadam_num += 1
 
-                    # Check for completed async operations
                     result = self.superoffload_cpu_optimizer.get_result()
                     if result is not None:
                         self._reassign_or_swap_out_partitioned_parameters_async(result[TaskKeys.SUB_GROUP_ID],
                                                                                 result[ResultKeys.UPDATED_PARAM])
                         self.async_cpuadam_num -= 1
 
-                    fp32_grad_tensor.copy_(concatenated_buffer, non_blocking=True)
-                else:
-                    fp32_grad_tensor.copy_(concatenated_buffer, non_blocking=True)
-
-        # Clean up parameter gradients
         for param in params_to_release:
             if not get_accelerator().is_synchronized_device():
-                param.grad.record_stream(get_accelerator().current_stream())
+                if param.grad is not None:
+                    param.grad.record_stream(get_accelerator().current_stream())
             param.grad = None
 
     @instrument_w_nvtx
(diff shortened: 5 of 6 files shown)


tohtana:
Hi @neeldani,
Thank you for opening this PR! This is truly exciting.

Since this is a large PR, let’s proceed step by step. Here are my suggestions:
- Code Location: This PR contains a significant amount of client code in `bench_dc_ulysses`. Could we move that to DeepSpeedExamples instead? Feel free to open a separate PR there for it.
- Documentation: The README in `bench_dc_ulysses` appears to be outdated. Could you update it with instructions so we can reproduce the results?
- API Design: Could you share the current API design? As you mentioned, we should discuss this further. You can either add the details to this PR or start a new Discussion in this repo.

neeldani:
@tohtana thank you for the feedback:
1. Moved the scripts to DeepSpeedExample and put up a new PR: https://github.com/deepspeedai/DeepSpeedExamples/pull/999
2. Updated the [README.md ](https://github.com/neeldani/DeepSpeedExamples/blob/2459c001b54a223b636cee639736c938a9138eed/benchmarks/autosp/README.md) with the setup instructions
3. Updated the description of this PR with the API design

Please let me know if there are any hiccups to run AutoSP or have any questions related to the design - happy to discuss them on this PR.

deepspeed/compile/custom_ops/all_to_all.py
```diff
@@ -0,0 +1,64 @@
+import torch
```
tohtana: (Not about this file) Don't we need `__init__.py` in `custom_ops`?

deepspeed/runtime/constants.py
```diff
@@ -501,3 +501,10 @@ class ValidationMode:
 #########################################
 USE_DATA_BEFORE_EXPERT_PARALLEL = "use_data_before_expert_parallelism"
 USE_DATA_BEFORE_EXPERT_PARALLEL_DEFAULT = False
+
+#########################################
+# AUTOSP
+#########################################
+INPUT_ID_KEY = "input_id"
```
tohtana: Can we make these AutoSP specific? like AUTOSP_*

deepspeed/runtime/engine.py
```diff
@@ -4361,7 +4362,8 @@ def compile(self,
         enable_deepcompile = self.is_deepcompile_enabled()
         if enable_deepcompile and self.zero_optimization_stage() != ZeroStageEnum.optimizer_states \
                 and self.zero_optimization_stage() != ZeroStageEnum.weights \
-                and self.zero_optimization_stage() != ZeroStageEnum.gradients:
+                and self.zero_optimization_stage() != ZeroStageEnum.gradients \
```
tohtana: Can you elaborate the intension of this change? (the preexisting condition  also seems weird, though)
Assuming `zero_optimization_stage` returns 0, 1, 2, 3, we never enter this block?
Maybe we need the fallback only when sp is disabled and zero stage is 0?

deepspeed/runtime/engine.py
```diff
@@ -4396,6 +4398,8 @@ def passes_name_to_fn(passes):
                         "DeepCompile with ZeRO stage 3 is not currently supported on PyTorch >= 2.9. "
                         "Please use ZeRO stage 1 or 2 with DeepCompile, or disable DeepCompile for ZeRO stage 3.")
                 backend = init_z3(self, backend, compile_config, compile_kwargs, schedule)
+            elif self.zero_optimization_stage() == ZeroStageEnum.disabled:
```
tohtana: Currently, do we enable AutoSP by these? 
- set zero stage to 0
- Enable deepcompile
If so, I think we should make it more explicit.
For AutoTP (see [example](https://github.com/tohtana/DeepSpeedExamples/tree/tohtana/custom_auto_tp/training/tensor_parallel/custom_patterns)), we do 
```
"tensor_parallel": {
   "autotp_size": args.tp_size,
...
}
```
For AutoEP proposal, 
```
"expert_parallel": {
   "autoep_size": args.ep_size,
...
}
```
AutoEP is currently just a proposal, but how about making the config
```
"sequence_parallel": {
   "autosp_size": args.sp_size,
...
}
```
You may want to require DeepCompile to be enabled too. As we don't have *eager* AutoSP now, it might be good to enable DeepCompile automatically when `sequence_parallel` is enabled.

tohtana:
Thank you @neeldani for the update! As we don't have a lot of changes in existing code, I don't think we have much risk.
The key discussion is the interface to enable AutoSP. See this [comment](https://github.com/deepspeedai/DeepSpeed/pull/7860#discussion_r2855806656). I also want to get thoughs from @sfc-gh-truwase @minjiazhang 

We should also have clear assertions to terminate early when we hit these limitations: Attention Pattern Matching and No Graph Break Requirement.

spikerheado1234:
> Thank you @neeldani for the update! As we don't have a lot of changes in existing code, I don't think we have much risk. The key discussion is the interface to enable AutoSP. See this [comment](https://github.com/deepspeedai/DeepSpeed/pull/7860#discussion_r2855806656). I also want to get thoughs from @sfc-gh-truwase @minjiazhang

Hi @tohtana, just merged in the code that correctly enables zero-1 and AutoSP interoperability.

deepspeed/runtime/engine.py
spikerheado1234: Here is the flow for enabling autosp (and its interoperability with zero-1 DP):
1. User specifies in a config, the pertinent compiler passes and their parameters. This looks like the following
```
"compile": {
        "deepcompile": true,
	"passes": ["autosp"],
	"sp_size": 2,
	"dp_size": 1
```
Specifying what the SP and DP size should be. 

Next, if the DP size is larger than one, the user can opt to turn on zero-1. The entire configuration would look like this:
```
   "zero_optimization":{
        "stage": 1
    },
"compile": {
        "deepcompile": true,
	"passes": ["autosp"],
	"sp_size": 2,
	"dp_size": 1
```

Using the legacy config-style to specify the zero_optimization stage. This would then accordingly compose SP with zero-1 DP. However, note that this is not Zero-1 DP from deepcompile, rather the Zero-1 DP implemented originally in DeepSpeed. 

Here, I have currently opted to make both the sp_size and dp_size explicitly controllable by the user. Another option is to automatically infer the DP-size from the sp-size by computing dp-size = num-devices/sp-size. 

For a sample config, here is the deepspeed config in DeepSPeedExamples (link [here](https://github.com/deepspeedai/DeepSpeedExamples/pull/999/changes#diff-b84d74633edc085e4c52db409205084a9ca963a63e3a17ebf592c1d44e206444R1)).

tohtana: Thank you, @spikerheado1234!
Can you clarify a few points?
- Can we automatically determine dp_size based on world size and sp size?
- I feel a bit weird to have "sp_size" as a part of compile config. Shouldn't we make a new item to set arguments to compiler passes? (I also want to get thoughts from @sfc-gh-truwase)
- In current SP (Ulysses), we have the size for zero sharding is dp_size * sp_size. Is this same for AutoSP?
- Can you add tests to run the matrix of sp_size * dp_size * zero_stage? It will clarify that what we support and guarantee it works.

tjruwase: I also find it odd to include both `sp_size` and `dp_size` in ds_config. To my knowledge, `dp_size` is implicitly derived from world_size, and `sp_size` is subsequently inferred from `dp_size`. 

I am aware that `dp_size` and `sp_size` is a common confusion source in SP, so this probably requires offline discussion for AutoSP.

deepspeed/runtime/constants.py
tjruwase: Additionally, how about moving these to `deepsped.compile.constants`?

deepspeed/compile/util.py
```diff
@@ -521,3 +525,100 @@ def pad_tensors(specs: List[Tuple[torch.Tensor, int, int]]) -> List[torch.Tensor
         padded.append(out)
 
     return padded
+
+INPUT_ID_KEY = "input_id"
```
tjruwase: Are these duplicates of entries in `deepspeed.runtime.constants`?

deepspeed/compile/passes/sp_compile.py
```diff
@@ -0,0 +1,250 @@
...
+from torch.fx.passes.fake_tensor_prop import FakeTensorProp
+from torch.fx.experimental.symbolic_shapes import ShapeEnv
+
+from deepspeed.runtime import constants
+
+from ..custom_ops import all_to_all, sp_dp_registry
+from ..fx import find_node_by_name, get_node_shape_meta
+from ..util import get_input_id_node, get_label_id_node, get_position_id_node, shard_tensor_node, get_sdpa_nodes 
+
+def prepare_autosp_inputs(input_id: torch.Tensor, label_id: torch.Tensor, position_id: torch.Tensor = None, attention_mask: torch.Tensor = None, seq_dim: int = 1):
```
tjruwase: I am curious if it is feasible to call this [here](https://github.com/deepspeedai/DeepSpeed/blob/04d69cc6562ee3e4ff3f21256c544f2cd7439938/deepspeed/runtime/engine.py#L2313) rather than requiring users to call it.

tjruwase: @tohtana FYI

deepspeed/compile/passes/sp_compile.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    Apply AutoSP (Ulysses) transformation passes to the graph and setup either DP/SP (2D) or SP (1D) mesh.
+    
+    Args:
+        gm: GraphModule to transform
+        real_inputs: Example inputs for shape propagation
+        debug: If True, print graph before/after each pass
+        passes: Optional custom list of passes (default: DEFAULT_PASSES)
+    """
+    assert sp_size * dp_size <= torch.cuda.device_count(), 'Insufficient device count for mesh size'
```
tjruwase: Does this work for >1 nodes?

deepspeed/runtime/constants.py
neeldani: That's a good suggestion, addressed.

deepspeed/runtime/engine.py
neeldani: Yeah I agree, I refactored this a bit into different functions to improve readability. We still fallback to torch compiler if validation fails to keep the behavior same as before

deepspeed/compile/util.py
neeldani: Yes I think that was a merge conflict, cleaned it up!

deepspeed/compile/custom_ops/all_to_all.py
neeldani: Done

neeldani:
> We should also have clear assertions to terminate early when we hit these limitations: Attention Pattern Matching and No Graph Break Requirement.

Thanks @tohtana, I added checks for the invariants.

Attention pattern matching is handled [here](https://github.com/neeldani/DeepSpeed/blob/4fdc54b84a3003517f16512c7b327070f2069b93/deepspeed/compile/passes/sp_compile.py#L160)

For no graph break requirement, I am forcing a full graph capture via torch.compile's `fullgraph=True` flag [here](https://github.com/neeldani/DeepSpeed/blob/4fdc54b84a3003517f16512c7b327070f2069b93/deepspeed/runtime/engine.py#L4356). The program would crash. There's another option where we run `dynamo._explain` and have a more graceful exit but the tradeoff is that the graph needs to be traced twice (first time to check for graph breaks, second time for actual compilation)

deepspeed/compile/passes/sp_compile.py
neeldani: That would indeed be good for user experience. A requirement would be to have a strict order of inputs to identify each of the individual inputs and so maintaining backward-compatibility could be an issue. But curious to know if there's a better way to achieve this.

sfc-gh-truwase: Yes, the strict requirement of ordering the inputs makes my proposal challenging. Let's revisit this issue after this release.

deepspeed/compile/passes/sp_compile.py
sfc-gh-truwase: @neeldani @spikerheado1234, ping on this

deepspeed/runtime/engine.py
```diff
@@ -4358,53 +4422,22 @@ def compile(self,
...
-            elif self.zero_optimization_stage() == ZeroStageEnum.weights:
-                if required_torch_version(min_version=2.9):
-                    raise RuntimeError(
-                        "DeepCompile with ZeRO stage 3 is not currently supported on PyTorch >= 2.9. "
-                        "Please use ZeRO stage 1 or 2 with DeepCompile, or disable DeepCompile for ZeRO stage 3.")
-                backend = init_z3(self, backend, compile_config, compile_kwargs, schedule)
+        if self.is_deepcompile_enabled():
+            resolved_backend, schedule = self.get_deepspeed_compile_backend(backend, compile_kwargs, schedule)
+
+        is_deepspeed_compile_backend = resolved_backend is not None
```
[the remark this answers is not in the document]
sfc-gh-truwase: @neeldani @spikerheado1234 FYI

deepspeed/runtime/engine.py
```diff
@@ -1004,6 +1005,14 @@ def zero_sub_group_size(self):
     def zero_optimization_stage(self):
         return self._config.zero_optimization_stage
 
+    def compile_zero_optimization_stage(self):
+        """Determines if zero-pass is set in deepcompile's passes attributes."""
+        return "z1" in self._config.compile_config.passes or "z3" in self._config.compile_config.passes
+
+    def compile_autosp(self):
+        """Determines if AutoSP is set in deepcompile's passes attributes."""
+        return "autosp" in self._config.compile_config.passes
```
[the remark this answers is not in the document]
sfc-gh-truwase: @neeldani @spikerheado1234 FYI

deepspeed/compile/custom_ops/all_to_all.py
```diff
@@ -0,0 +1,74 @@
...
+    _, scatter_idx, gather_idx, name = inputs
+    ctx.scatter_idx = gather_idx
+    ctx.gather_idx = scatter_idx
+    ctx.name = name + "_grad"
+
+
+def _all_to_all_backward(ctx, grad):
+    return (
+        all_to_all(grad, ctx.scatter_idx, ctx.gather_idx, ctx.name),
+        None, None, None, None
```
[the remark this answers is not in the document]
sfc-gh-truwase: @neeldani @spikerheado1234 FYI

deepspeed/compile/passes/sp_compile.py
```diff
@@ -0,0 +1,246 @@
...
+def pass_shard_seq_dim(gm: GraphModule, example_inputs):
+    """
+    Finds all direct and indirect consumers of the input sequence, label and position ids.
+    Shard the sequence dimension used by all such consumers.
+    """
+    sp_size = sp_dp_registry.sp_size()
+    
+    input_ids_node = get_input_id_node(gm)
+    val = get_node_shape_meta(input_ids_node)
+    seq_symint = val.shape[1]
```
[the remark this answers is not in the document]
sfc-gh-truwase: @neeldani @spikerheado1234 FYI

deepspeed/compile/passes/sp_compile.py
spikerheado1234: Addressed.

deepspeed/runtime/engine.py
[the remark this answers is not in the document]
spikerheado1234: Addressed.

deepspeed/runtime/engine.py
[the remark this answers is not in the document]
spikerheado1234: Addressed.

deepspeed/compile/custom_ops/all_to_all.py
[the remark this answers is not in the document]
spikerheado1234: Addressed.

deepspeed/compile/passes/sp_compile.py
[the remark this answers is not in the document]
spikerheado1234: For this, while valid, the very next line is an assertion that double-checks that the extracted value is of type SymInt. Our `prepare_autosp` function ensures that the sequence dimension is always of this type, and so the next line will conclude whether or not the extracted dimension is of sequence length (in the pathological event that `val.shape[1]` is not the sequence dimension). 

In light of this, do you suggest that we still modify the code?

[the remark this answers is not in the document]
sfc-gh-truwase: Got it. No modification needed.

tohtana:
> For no graph break requirement, I am forcing a full graph capture via torch.compile's fullgraph=True flag [here](https://github.com/neeldani/DeepSpeed/blob/4fdc54b84a3003517f16512c7b327070f2069b93/deepspeed/runtime/engine.py#L4356).

I think using `fullgraph` is the right approach. We just need to clarify it in the document. This PR currently doesn't have a document, but we would need to add a section in `training.rst` and `config-json.md`.
It would also be good if we could handle the error due to the `fullgraph` violation.

deepspeed/runtime/engine.py
spikerheado1234: Hi @tohtana @tjruwase, following up on this thread with the necessary answers. 

Prior to doing that. Wanted to update that, based on this conversation, I have slightly modified the interface of using autosp. Now, instead of passing in the sp-size/dp-size explicitly, there is a `pass_args` field in the config that the user can dump in compiler arguments. The new config for an sp-size of 2 will look as follows:

```
    "zero_optimization": {
        "stage": 0
    },
    "compile": {
        "deepcompile": true,
	"passes": ["autosp"],
	"pass_args" : {
		"sp_size" : 2
	}
```
Moreover, the user no longer needs to specify dp-size (it is extracted from world-size and sp-size).

Now the answer to @tohtana are as follows:
1. Can we automatically determine dp_size based on world size and sp size? Yes, and now we do with the latest change. 
2. I feel a bit weird to have "sp_size" as a part of compile config. The latest change now refactors the interface to instead specify the sp-size as a pass_arg. To my knowledge, I don't think we can get around the user specifying the sp-size at least (given this determines how many ranks participate in the Ulysses all-to-all).
3. In current SP (Ulysses), we have the size for zero sharding is dp_size * sp_size. Is this same for AutoSP? Yes, it is the same for AutoSP.
4. Can you add tests to run the matrix of sp_size * dp_size * zero_stage? I have added a separate correctness check that does this. The relevant commit is in DeepSpeedExamples (link [here](https://github.com/deepspeedai/DeepSpeedExamples/commit/af8dff895638407fad3139b2802462dd45fbb779))

sfc-gh-truwase: @spikerheado1234 thanks for the updates. Is it possible to reuse the existing "sequence_parallel_size" field in ds_config instead of introducing a new `pass_args:sp_size`. That could make it easy to switch between autosp and non-compiled ulyssess. You can see the usage in the ALST tutorial: https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/

image

spikerheado1234: Addressed @sfc-gh-truwase . The config looks like this now:

```
    "zero_optimization": {
        "stage": 0
    },
    "compile": {
        "deepcompile": true,
        "passes": ["autosp"]
    },
    "sequence_parallel_size": 2
``` 

Leveraging the `sequence_parallel_size` config in the ALST tutorial. I have modified the correctness checks in DeepSpeedExamples accordingly as well.

sfc-gh-truwase: LGTM!

docs/_pages/config-json.md
```diff
@@ -1896,6 +1896,76 @@ Different pruning sets, this is used for different pruning parameters. In this e
...
+| List of compiler passes to apply. Currently supported: `["autosp"]`.     | `[]`    |
+
+
+<i>**sp_size**</i>: [integer]
+
+| Description                                                                         | Default |
+| ----------------------------------------------------------------------------------- | ------- |
+| Sequence parallel degree (number of devices to shard the sequence dimension across).   | `1`     |
+
+**Requirements:**
```
sfc-gh-truwase: The content under **Requirements** are great but should go into a tutorial. So, please remove.

docs/_pages/config-json.md
```diff
@@ -1896,6 +1896,76 @@ Different pruning sets, this is used for different pruning parameters. In this e
...
+        input_id=batch["input_ids"],
+        label_id=batch["labels"],
+        position_id=batch.get("position_ids"),
+        seq_dim=1
+    )
+    loss = engine(input_ids)
+    engine.backward(loss)
+    engine.step()
+```
+
```
sfc-gh-truwase:
```suggestion
```

sfc-gh-truwase:
Please fix formatting using https://github.com/deepspeedai/DeepSpeed/blob/master/CONTRIBUTING.md#prerequisites

docs/_pages/config-json.md
neeldani: Addressed!

docs/_pages/config-json.md
```diff
@@ -1896,6 +1896,34 @@ Different pruning sets, this is used for different pruning parameters. In this e
...
+DeepSpeed provides compiler-based optimization passes through the `compile` configuration. This includes enabling Ulysses-styled sequence paralllelism and a custom heuristic selective activation checkpointing pass. To enable Automatic Sequence Parallelism (AutoSP), configure the `compile` section:
+
+```json
+{
+    "zero_optimization": {"stage": 0},
+    "compile": {
+        "deepcompile": true,
+        "passes": ["autosp"],
+        "pass_args": {"sp_size": 2}
```
sfc-gh-truwase: I believe `{"sp_size"}` is no longer needed as discussed [earlier](https://github.com/deepspeedai/DeepSpeed/pull/7860#discussion_r2903354106).
```suggestion
```

docs/_pages/config-json.md
```diff
@@ -1896,6 +1896,34 @@ Different pruning sets, this is used for different pruning parameters. In this e
...
+| Description                                                              | Default |
+| ------------------------------------------------------------------------ | ------- |
+| List of compiler passes to apply. Currently supported: `["autosp"]`.     | `[]`    |
+
+
+<i>**sp_size**</i>: [integer]
+
+| Description                                                                         | Default |
+| ----------------------------------------------------------------------------------- | ------- |
+| Sequence parallel degree (number of devices to shard the sequence dimension across).   | `1`     |
```
sfc-gh-truwase: I believe also not needed. 
```suggestion
```

tohtana:
Thank you @spikerheado1234 for adding tests. Some tests raise an error on my environment. The versions of PyTorch and Transformers are:
```
Torch: 2.9.1+cu128
Transformers: 5.1.0
```

I ran this command:
```
export CUDA_VISIBLE_DEVICES=0,1,2,3 HF_HOME=... && python -m pytest -q tests/unit/v1/compile/test_compile_autosp.py -k "TestAutoSPCompile and test[2-0-dtype1]"'
```

[autosp_test.log](https://github.com/user-attachments/files/25991025/autosp_test.log)

Do we have any restriction on the versions?

spikerheado1234:
> Thank you @spikerheado1234 for adding tests. Some tests raise an error on my environment. The versions of PyTorch and Transformers are:
> 
> ```
> Torch: 2.9.1+cu128
> Transformers: 5.1.0
> ```
> 
> I ran this command:
> 
> ```
> export CUDA_VISIBLE_DEVICES=0,1,2,3 HF_HOME=... && python -m pytest -q tests/unit/v1/compile/test_compile_autosp.py -k "TestAutoSPCompile and test[2-0-dtype1]"'
> ```
> 
> [autosp_test.log](https://github.com/user-attachments/files/25991025/autosp_test.log)
> 
> Do we have any restriction on the versions?

Hi @tohtana. Yes there is a restriction on the version of Transformers. You must use a version of at most 4.50.3. This is because later versions of transformers' `modeling_llama` introduce graph-breaks upon compilation, which will throw an error in our compilation pipeline (as seen in the error log you have posted).

tohtana:
I tried with `transformers==4.50.3` but the same error was raised.
Do you also have any restriction on PyTorch? It would be helpful if you could share versions of major dependencies.

spikerheado1234:
> I tried with `transformers==4.50.3` but the same error was raised. Do you also have any restriction on PyTorch? It would be helpful if you could share versions of major dependencies.

We don't have any restriction on PyTorch that we are aware of. 

In my setup, I am using torch `2.7.0+cu128`. I have frozen my pip enviornment as well - [deps.txt](https://github.com/user-attachments/files/25991871/deps.txt) - in case that helps at all.

tohtana:
I tried this combination:
  - torch 2.7.0+cu126
  - transformers 4.50.3

and tests passed! Thank you for sharing the compatible version.

It is now clear that PyTorch version makes a difference/ Can we investigate it?
I know DeepCompile+ZeRO3 is broken for PyTorch v2.9 or newer. But ZeRO1 and 2 work.
It would be great if we could support newer versions. Al lease, we need to clarify the supported versions and raise a proper error (the same goes for transformers).

spikerheado1234:
> I tried this combination:
> 
> * torch 2.7.0+cu126
> * transformers 4.50.3
> 
> and tests passed! Thank you for sharing the compatible version.
> 
> It is now clear that PyTorch version makes a difference/ Can we investigate it? I know DeepCompile+ZeRO3 is broken for PyTorch v2.9 or newer. But ZeRO1 and 2 work. It would be great if we could support newer versions. Al lease, we need to clarify the supported versions and raise a proper error (the same goes for transformers).

Hi @tohtana. I have added an additional check that checks for the relevant pytorch & transformers major/minor version. The check occurs in `init_autosp` and throws an error if the user's PyTorch and/or Transformers version is incompatible. 

Moreover, I have investigated what the issue with PyTorch 2.9 is. The lowering of the embedding layer in PyTorch 2.9 is slightly different compared to prior PyTorch versions. In the IR in PyTorch 2.9, the embedding layer represents float-constants as tensors, with corresponding `.item()` methods to recover their true values. However, the last pass of AutoSP: `pass_propagate_shapes` re-runs the forward-pass to propagate the values of sharded sequences. This pass, when it encounters the `.item()` function, throws an error as the pass reruns the forward pass with `FakeTensors`, which do not have concrete values. When this occurs, Dynamo inserts a graph-break, breaking AutoSP's compilation.  

I have created a minimal-example to reproduce the exact issue (when using PyTorch 2.7 it works, but fails for PyTorch 2.9). The file is here: [sample_test.py](https://github.com/user-attachments/files/26000899/sample_test.py).

tohtana:
I tried to address the issues with PyTorch v2.9. I opened a PR to update this PR branch in https://github.com/neeldani/DeepSpeed/pull/2.
I would appreciate if you could verify it on your end.

deepspeed/compile/passes/long_context_checkpointing.py
```diff
@@ -93,14 +93,8 @@ def register_long_context_checkpointing():
...
     # Locate the original should_ban_recomputation and the function after it.
-    start = next(
-        i for i, l in enumerate(lines)
-        if l.startswith('    def should_ban_recomputation(')
-    )
-    end = next(
-        i for i, l in enumerate(lines)
-        if i > start and l.startswith('    def ')
-    )
+    start = next(i for i, l in enumerate(lines) if l.startswith('    def should_ban_recomputation('))
```
neeldani, line 96: I am wondering if we should find an alternative since this approach seems to be hack @spikerheado1234 @tohtana

spikerheado1234: I tried a bunch of things, but I came up with nothing that matches the exact schedule induced by the exec-type method unfortunately.

spikerheado1234:
> I tried to address the issues with PyTorch v2.9. I opened a PR to update this PR branch in [neeldani#2](https://github.com/neeldani/DeepSpeed/pull/2). I would appreciate if you could verify it on your end.

I have verified this and merged this in. I have updated the PyTorch version requirements to be 2.9+, rather than <2.9 as well.

docs/_pages/config-json.md
neeldani: Addressed

docs/_pages/config-json.md
neeldani: Done

deepspeed/runtime/engine.py
```diff
@@ -4394,53 +4459,23 @@ def compile(self,
...
-                 and "offload_param" in self.config["zero_optimization"])
-                    and self._config.zero_config.offload_param.device == "cpu"
-                    and self._config.zero_config.offload_optimizer.device == "cpu"):
-                compile_config.offload_parameters = True
-            if self.zero_optimization_stage() == ZeroStageEnum.optimizer_states:
-                backend = init_z1(self, backend, compile_config, compile_kwargs, schedule)
-            elif self.zero_optimization_stage() == ZeroStageEnum.gradients:
-                backend = init_z1(self, backend, compile_config, compile_kwargs, schedule, use_z2=True)
-            elif self.zero_optimization_stage() == ZeroStageEnum.weights:
-                if required_torch_version(min_version=2.9):
```
tohtana, line 4430: @neeldani @spikerheado1234 the gate to block Z3+DeepCompile with >=v2.10. Can you revert the check?

tohtana (changes requested):

deepspeed/runtime/engine.py
neeldani: I reverted this change, thanks for catching this. Although the previous code requires version >= 2.9, not 2.1 so I reverted accordingly.

tohtana (approved):
Thank you, @neeldani, @spikerheado1234! Looks good to me.

Issue #949, opened 2021-04-13
ImportError: dynamic module does not define module export function (PyInit_transformer_op)


gongjingcs:
hi，I install deepspeed  using the following command：
DS_BUILD_OPS=1 python setup.py build_ext -j8 bdist_wheel 
pip install dist/deepspeed-0.3.13+69bca4a-cp38-cp38-linux_x86_64.whl

I ran a demo with transformer kernel， I met with the following error:
![image](https://user-images.githubusercontent.com/15030497/114506597-9c6d3f00-9c64-11eb-982a-1d674c7fefe1.png)

gongjingcs:
![image](https://user-images.githubusercontent.com/15030497/114514141-8e6fec00-9c6d-11eb-9510-04a761ff0295.png)

tjruwase:
@gongjingcs. can you please share the results of running `ds_report `in both working and failing installations?

gongjingcs:
@tjruwase 
working installations：
![image](https://user-images.githubusercontent.com/15030497/114644290-cdec1600-9d09-11eb-977a-2b0c56480190.png)


failing installations：
![image](https://user-images.githubusercontent.com/15030497/114644789-c24d1f00-9d0a-11eb-87c2-52b1ab541935.png)

both ds_report results of working and failing installations are the same

gongjingcs:
@tjruwase would you please provide some tips  for solving this problem ?

tjruwase:
@gongjingcs, apologies for the delay on this. I will take a closer look today.

RezaYazdaniAminabadi:
Hi @gongjingcs 

I think you might have some incompatibility issue between the torch you installed and the one you are using. Also, the CUDA version you are using with Torch1.8 is lower than what torch is supporting based on their website. Could you please try resolving these and rerun the experiment?
![image](https://user-images.githubusercontent.com/44502768/115435998-4c8bfc00-a1bf-11eb-8b86-f9674deb1d7d.png)
![image](https://user-images.githubusercontent.com/44502768/115436026-56adfa80-a1bf-11eb-883b-83a3d983a504.png)

tjruwase:
@gongjingcs, did you try resolving the torch incompatibility?

gongjingcs:
> @gongjingcs, did you try resolving the torch incompatibility?

hi, I tried resolving the torch incompatibility. I downgraded my torch version to 1.7.1, however it reports the same error

[...]

PR #7862, opened 2026-02-20
Fix compilation of Evoformer

Flamefire:
`EvoformerAttnBuilder` returns instances of `Path` from `include_paths` which then cause failures in `OpBuilder.builder` when passing them to `strip_empty_entries` that calls `len` on them which isn't defined for `Path` instances:
>   TypeError: object of type 'PosixPath' has no len()

Fixes regression introduced in #7760

cc @sdvillal

### DIFF
--- a/op_builder/evoformer_attn.py
+++ b/op_builder/evoformer_attn.py
@@ -115,7 +115,7 @@ class EvoformerAttnBuilder(CUDAOpBuilder):
         if not cutlass_path.is_dir():
             raise RuntimeError(f"CUTLASS_PATH {cutlass_path} does not exist")
         include_dirs = cutlass_path / "include", cutlass_path / "tools" / "util" / "include"
-        include_dirs = [include_dir for include_dir in include_dirs if include_dir.is_dir()]
+        include_dirs = [str(include_dir) for include_dir in include_dirs if include_dir.is_dir()]
         if not include_dirs:
             raise RuntimeError(f"CUTLASS_PATH {cutlass_path} does not contain any include directories")
         return include_dirs


sdvillal:
Thanks a lot for the fix, I think it should be merged asap.

Should we then add tests for the different ways of configuring CUTLASS location?

Flamefire:
> Should we then add tests for the different ways of configuring CUTLASS location?

Having a test on CI that builds all operations and loads them, even without execution would catch things like this and e.g. https://github.com/deepspeedai/DeepSpeed/issues/949
Not sure why this isn't done already, so I only assume there are reasons.

sdvillal:
> > Should we then add tests for the different ways of configuring CUTLASS location?
> 
> Having a test on CI that builds all operations and loads them, even without execution would catch things like this and e.g. #949 Not sure why this isn't done already, so I only assume there are reasons.

Not necessarily, as there are ways to compile the extension without making CUTLASS_PATH point to a directory - this is the codepath we use in our side. I was testing all use cases locally before I introduced the simple mistake and clearly I became too confident and failed to consistently continue testing.

To make sure all works well, we should test all 3 ways of configuring CUTLASS_PATH: DS_IGNORE_CUTLASS_DETECTION, DS_USE_CUTLASS_PYTHON_BINDINGS and hardcoded paths. Right now compilation works for the first case. I personally would actually remove the other two and always ask the user to have a well setup compiler configuration, without deepspeed specific config. But that would be an intended breaking change. 

Alternatively, we could vendor cutlass to provide a sane default that always works and allow knowledgeable users to redefine it, but I guess that is not something we want to do.

sdvillal:
Mentioned in the other conversation (let's move here): As a workaround, would pointing to the relevant include dirs (e.g., using CPATH) and removing deepspeed specific configuration (i.e., setting CUTLASS_PATH=DS_IGNORE_CUTLASS_DETECTION) work for you?

Flamefire:
Yes `CUTLASS_PATH=DS_IGNORE_CUTLASS_DETECTION` avoids the problematic code path and as CUTLASS is header-only CPATH works.

tohtana:
Hi @Flamefire, thank you for the fix, and thanks @sdvillal for the discussion.

Because this issue causes a real compilation break, I’m going to merge this PR now.

For follow-up, we should add coverage for Evoformer/CUTLASS build paths. Our per-PR CI currently runs a reduced unit-test set and does not compile Evoformer yet. We’re also working on restoring nightly full tests, and I think this coverage should be added there first.

sdvillal:
Thanks @tohtana.

I think it would be best to remove all the custom configuration from the evoformer and report a changing break. Do we have other examples in deepspeed on how cutlass is handled?

PR #7882, opened 2026-03-02
Fix performance regression in grad_handling_hook

rraminen:
This PR fixes a performance drop introduced by calling count_used_parameters_in_backward() inside every gradient hook.

In the previous implementation, the expected hook count was computed once per backward phase. After a recent PR changes (https://github.com/deepspeedai/DeepSpeed/commit/311674ffadc7e456819b3e66950c8a5fe1320749#diff-99dcf26ea2876ff5bbf05b5165c4133eaa0d0f36b170685643c2f7e2eb566addL1002-L1010), it is being recomputed on every hook invocation, resulting in a drop in samples/sec values.

With the fix in this PR performance returns to pre-regression samples/sec values.

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1046,12 +1046,14 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
                     def wrapper(param, i):
 
                         def grad_handling_hook(*notneeded):
-                            self.reenter_backward_if_needed()
-                            self.process_gradients(param, i)
                             if self._remaining_grad_acc_hooks == 0:
-                                self.current_expected_hooks = count_used_parameters_in_backward(all_params_requiring_grad)
-                            self.update_hook_state_and_maybe_run_epilogue(self.current_expected_hooks)
-                            self._remaining_grad_acc_hooks -= 1
+                                self.reenter_backward_if_needed()
+                            self.process_gradients(param, i)
+                            if self._hooks_fired_this_backward == 0:
+                                current_expected = count_used_parameters_in_backward(all_params_requiring_grad)
+                            else:
+                                current_expected = self._max_expected_hooks_seen
+                            self.update_hook_state_and_maybe_run_epilogue(current_expected)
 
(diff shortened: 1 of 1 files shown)


rraminen:
@codex review

tohtana:
Hi @rraminen, 
Thank you for the PR! The change introduced in #7780 definitely adds some overhead, but we didn’t expect it to be this significant.
Could you share the performance numbers from before and after the change? If you can also share a repro, we’ll look into ways to improve the performance and mitigate the impact.

rraminen:
> Hi @rraminen, Thank you for the PR! The change introduced in #7780 definitely adds some overhead, but we didn’t expect it to be this significant. Could you share the performance numbers from before and after the change? If you can also share a repro, we’ll look into ways to improve the performance and mitigate the impact.

Thank you for looking into this, I will gather the info and share it here.

rraminen:
@codex review

tohtana:
Hi @rraminen,
As I was able to reproduce the issue, I opened a PR #7886  to address the issue. I thought the condition to block calling `count_used_parameters_in_backward` needs to be a bit more complicated.

Can you help us review it? Also, can I add you as a co-author of the PR because you identified the root cause?

rraminen:
> Hi @rraminen, As I was able to reproduce the issue, I opened a PR #7886 to address the issue. I thought the condition to block calling `count_used_parameters_in_backward` needs to be a bit more complicated.
> 
> Can you help us review it? Also, can I add you as a co-author of the PR because you identified the root cause?

Hi @tohtana, thank you for the fix. The hook count regression introduced in v0.18.5 caused noticeable overhead in large models where the number of modules is high.

I'll take a look at PR #7886 and leave comments there shortly. I'm happy to be added as a co-author. Thanks for the offer and for working on the fix.

rraminen:
https://github.com/deepspeedai/DeepSpeed/pull/7886 fixed the issue addressed in this PR.

tohtana:
@rraminen I really appreciate your contribution. You identified the cause of the performance issue, and that's why we were able to fix it quickly. I included you as a co-author of #7886.

Issue #7885, opened 2026-03-04
[BUG] Training with Zero-2 slow down a lot from v0.18.4 to v0.18.5


jue-jue-zi:
**Describe the bug**
Training with Zero-2 slow down a lot from v0.18.4 to v0.18.5 / v0.18.6.

**ds_report output**
```
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
 [WARNING]  async_io: please install the libaio-devel package with yum
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
async_io ............... [NO] ....... [NO]
fused_adam ............. [NO] ....... [OKAY]
cpu_adam ............... [NO] ....... [OKAY]
cpu_adagrad ............ [NO] ....... [OKAY]
cpu_lion ............... [NO] ....... [OKAY]
dc ..................... [NO] ....... [OKAY]
 [WARNING]  Please specify CUTLASS location directory as environment variable CUTLASS_PATH
 [WARNING]  Possible values are: a path, DS_IGNORE_CUTLASS_DETECTION and DS_USE_CUTLASS_PYTHON_BINDINGS
evoformer_attn ......... [NO] ....... [NO]
 [WARNING]  FP Quantizer is using an untested triton version (3.6.0), only 2.3.(0, 1) and 3.0.0 are known to be compatible with these kernels
fp_quantizer ........... [NO] ....... [NO]
fused_lamb ............. [NO] ....... [OKAY]
fused_lion ............. [NO] ....... [OKAY]
 [WARNING]  gds requires the dev libaio .so object and headers but these were not found.
 [WARNING]  gds: please install the libaio-devel package with yum
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
gds .................... [NO] ....... [NO]
transformer_inference .. [NO] ....... [OKAY]
inference_core_ops ..... [NO] ....... [OKAY]
[...]

PR #7886, opened 2026-03-04
Fix hook count performance regression from v0.18.5

tohtana:
Fixes performance regressions reported in #7882 and #7885.

PR #7780 added dynamic hook count computation for reentrant checkpointing correctness, but placed the call inside every gradient hook closure. For a model with n parameter tensors, this creates significant overhead per backward pass.

Summary:

1. Added `should_refresh_expected_hook_count()` predicate that returns true only at backward phase boundaries (first hook, or new reentrant phase), so `count_used_parameters_in_backward()` is called once per phase instead of once per hook.
2. Applied this predicate in ZeRO-1/2 (stage_1_and_2.py) and both ZeRO-3 hook sites (stage3.py), reusing the `cached_max_expected_hooks_seen` value when refresh isn't needed.
3. Changed enter_backward() to reset hook counters on first real backward entry, preventing pollution from pre-user-backward autograd calls (e.g., TiledFusedLogitsLoss).

With 24-layer transformer, ~267M params (147 parameter tensors), ZeRO-2, 8×H100 80GB, bf16, batch size 8, 20 warmup + 20 measured iterations:
  - Before fix: 0.1265s/iter
  - After fix: 0.0505s/iter

### DIFF
--- a/deepspeed/runtime/zero/stage3.py
+++ b/deepspeed/runtime/zero/stage3.py
@@ -1279,6 +1279,7 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
 
                         @instrument_w_nvtx
                         def reduce_partition_and_remove_grads(*notneeded):
+                            # Evaluate refresh condition before reenter_backward_if_needed()
                             refresh_expected = self.should_refresh_expected_hook_count()
                             # Re-enter backward for subsequent phases in reentrant checkpointing
                             self.reenter_backward_if_needed()
@@ -1307,6 +1308,7 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
             def make_hook(params):
 
                 def reduce_leaf_module_grads(module, grad_input, grad_output):
+                    # Evaluate refresh condition before reenter_backward_if_needed()
                     refresh_expected = self.should_refresh_expected_hook_count()
                     self.reenter_backward_if_needed()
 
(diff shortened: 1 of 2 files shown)


tohtana:
@codex

deepspeed/runtime/zero/stage3.py
```diff
@@ -1279,14 +1279,18 @@ def wrapper(param):
 
                         @instrument_w_nvtx
                         def reduce_partition_and_remove_grads(*notneeded):
+                            refresh_expected = self.should_refresh_expected_hook_count()
```
delock, line 1283: Hi @tohtana the `refresh_expected = self.should_refresh_expected_hook_count()` line is used on L1289, why it is placed before L1284 (`reenter_backward_if_needed()`) and L1286 (`reduce_ready_partitions_and_remove_grads()`).  Is there implicit dependency here?

tohtana: Thank you for checking, @delock!

Yes, there's actually a dependency. `should_refresh_expected_hook_count()` detects reentrant phase boundaries by checking `backward_active_depth == 0`. `reenter_backward_if_needed()` increments `backward_active_depth` from 0 → 1 when it detects a new phase. If we called `reenter_backward` first, it would set `backward_active_depth = 1` before the predicate runs, making the condition always false.

I added a comment to clarify it.

mjkvaak-amd:
Confirming that this fix improves AMD 8× MI300X training performance of Qwen-Image by ~10× compared to DeepSpeed v0.18.6.

rraminen:
This fix resolves the ~20% performance regression introduced in DeepSpeed v0.18.5 ([311674f](https://github.com/deepspeedai/DeepSpeed/commit/311674ffadc7e456819b3e66950c8a5fe1320749#diff-99dcf26ea2876ff5bbf05b5165c4133eaa0d0f36b170685643c2f7e2eb566addL1002-L1010)) for the DeepSpeed-Megatron GPT-3 13B workload.

PR #7887, opened 2026-03-05
[SP] add SP deny list instead of allow

kashif:
this way one can register kernels based flash-attn as well with SP

### DIFF
--- a/deepspeed/module_inject/layers.py
+++ b/deepspeed/module_inject/layers.py
@@ -369,6 +369,12 @@ class TensorParallel_Layer(nn.Module, ABC):
     def _mark_uc_metadata(self):
         return
 
+    def _should_materialize_tp_partition(self):
+        # AutoTP partitioning should only materialize parameters when an actual
+        # TP process group is present. Metadata-only construction with
+        # mp_group=None should not touch device placement.
+        return self.mp_group is not None
+
     def is_training_mode(self):
         global DEEPSPEED_AUTOTP_MODE
         return DEEPSPEED_AUTOTP_MODE == AUTOTP_MODE.TRAINING
@@ -579,7 +585,8 @@ class LinearAllreduce(TensorParallel_Layer):
         self.weight = module.weight
         self.bias = module.bias
 
-        self._tp_partition([self.weight, self.bias])
+        if self._should_materialize_tp_partition():
+            self._tp_partition([self.weight, self.bias])
         self.support_training = True
         self.config_tp_params(self.weight)
         if self.bias is not None:
@@ -674,7 +681,7 @@ class LinearLayer(TensorParallel_Layer):
         super(LinearLayer, self).__init__(mp_group, **kwargs)
         self.weight = module.weight
         self.bias = module.bias
-        if not skip_partition:
+        if not skip_partition and self._should_materialize_tp_partition():
             self._tp_partition([self.weight, self.bias])
         self.support_training = True
         self.config_tp_params(self.weight)
@@ -1234,7 +1241,8 @@ class SubParamLinearLayer(TensorParallel_Layer):
             raise ValueError(f"AutoTP layer '{self.name}' bias size {self.bias.numel()} does not match output shape "
                              f"{self._output_shape}.")
 
-        self._tp_partition([self.weight, self.bias])
+        if self._should_materialize_tp_partition():
+            self._tp_partition([self.weight, self.bias])
         self.support_training = True
         self.config_tp_params(self.weight)
         if self.bias is not None:
@@ -1352,7 +1360,8 @@ class SubParamLinearAllreduce(TensorParallel_Layer):
          self._bias_partition_dim) = _infer_subparam_logical_shapes(self._orig_weight_shape, self.shape,
                                                                     self.partition_dim, self.name)
 
-        self._tp_partition([self.weight, self.bias])
+        if self._should_materialize_tp_partition():
+            self._tp_partition([self.weight, self.bias])
         self.support_training = True
         self.config_tp_params(self.weight)
         if self.bias is not None:
--- a/deepspeed/runtime/zero/stage3.py
+++ b/deepspeed/runtime/zero/stage3.py
@@ -7,7 +7,7 @@ import sys
 import gc
 import collections
 import itertools
-from typing import Deque, Dict, Set, List, Tuple, Container, Optional
+from typing import Deque, Dict, Set, List, Container, Optional
 from contextlib import contextmanager
 from dataclasses import dataclass, field
 
@@ -21,13 +21,13 @@ from deepspeed.utils.torch import register_grad_hook, required_torch_version
 from deepspeed.runtime.fp16.loss_scaler import CreateLossScaler
 from deepspeed.runtime.torch_autocast import get_autocast_dtype, get_all_comm_dtypes, is_autocast_initialized, sort_dtypes
 from deepspeed.runtime.comm.coalesced_collectives import reduce_scatter_coalesced, all_to_all_quant_reduce, all_to_all_loco_quant_reduce
-from deepspeed.runtime.utils import inf, is_model_parallel_parameter, get_only_unique_item, mask_nan_or_inf_with_val_inplace, count_used_parameters_in_backward
+from deepspeed.runtime.utils import inf, is_model_parallel_parameter, mask_nan_or_inf_with_val_inplace, count_used_parameters_in_backward
 from deepspeed.runtime.zero.partition_parameters import *
 from deepspeed.runtime.zero.config import ZeroStageEnum
 from deepspeed.runtime.zero.offload_config import OffloadDeviceEnum, OffloadStateTypeEnum
 from deepspeed.runtime.zero.parameter_offload import DeepSpeedZeRoOffload
 import deepspeed.runtime.zenflow.engine_stage3 as zf_engine_stage3
-from deepspeed.runtime.zero.utils import get_mapping_to_flat_buffer
+from deepspeed.runtime.zero.utils import get_mapping_to_flat_buffer, defragment
 from deepspeed.runtime.zero.offload_states import offload_adam_states, reload_adam_states
 from deepspeed.ops.adam import DeepSpeedCPUAdam
 from deepspeed.runtime.swap_tensor.partitioned_param_swapper import PartitionedParamStatus
@@ -655,38 +655,6 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
         """Return the current learning rate."""
         return self.optimizer.param_groups[0]["lr"]
 
-    # TODO. factor out to a utility outside of stage3
-    @staticmethod
-    def defragment(tensors: List[Tensor]) -> Tensor:
-        """move provided tensors into a contiguous flat buffer, with some additional
-        measures taken to reduce memory fragmentation"""
-        assert len(set(t.dtype for t in tensors)) == 1
-        assert len(set(t.device for t in tensors)) == 1
-
-        cpu_buffer = torch.empty(sum(p.numel() for p in tensors),
-                                 dtype=get_only_unique_item(t.dtype for t in tensors),
-                                 device="cpu")
-        tensor_infos: List[Tuple[Tensor, int, int]] = get_mapping_to_flat_buffer(tensors)
-        orig_device = get_only_unique_item(t.device for t in tensors)
-
-        offset = 0
-        for tensor, offset, tensor_numel in tensor_infos:
-            # move the tensor from device memory to host memory
-            cpu_buffer.narrow(0, offset, tensor_numel).copy_(tensor)
-            tensor.data = torch.empty(0, dtype=tensor.dtype, device=tensor.device)
-
-        gc.collect()
-        get_accelerator().empty_cache()
-
-        # copy tensors (now flattened and contiguous) back to GPU
-        device_buffer = cpu_buffer.to(orig_device)
-
-        # restore device tensors
-        for tensor, offset, tensor_numel in tensor_infos:
-            tensor.data = device_buffer.narrow(0, offset, tensor_numel)
-
-        return device_buffer
-
     def _get_param_coordinator(self):
         return self.parameter_offload.get_param_coordinator()
 
@@ -834,7 +802,7 @@ class DeepSpeedZeroOptimizer_Stage3(ZeROOptimizer):
             parameter_partitions = self._get_parameter_partitions()
 
             # We need to keep the reference to this buffer to make sure you can free it in `offload_states`
-            self.lp_param_buffer = __class__.defragment(parameter_partitions)
+            self.lp_param_buffer = defragment(parameter_partitions)
             self._set_fp16_partitioned_groups_flat()
 
         else:  # partitioned params offloaded to CPU when not in use
--- a/deepspeed/runtime/zero/utils.py
+++ b/deepspeed/runtime/zero/utils.py
@@ -4,6 +4,7 @@
 # DeepSpeed Team
 
 import os
+import gc
 from typing import List, Tuple
 
 import torch
@@ -15,6 +16,7 @@ from deepspeed.ops.adam import FusedAdam
 from deepspeed.ops.lion import DeepSpeedCPULion, FusedLion
 from deepspeed.utils.nvtx import instrument_w_nvtx
 from deepspeed.accelerator import get_accelerator
+from deepspeed.runtime.utils import get_only_unique_item
 
 # ensure we only warn once, otherwise every iteration will trigger a warning
 warned = False
@@ -200,3 +202,34 @@ def get_mapping_to_flat_buffer(tensors: List[torch.Tensor]) -> List[Tuple[torch.
         offset += tensor_numel
 
     return tensor_infos
+
+
+def defragment(tensors: List[torch.Tensor]) -> torch.Tensor:
+    """move provided tensors into a contiguous flat buffer, with some additional
+    measures taken to reduce memory fragmentation"""
+    assert len(set(t.dtype for t in tensors)) == 1
+    assert len(set(t.device for t in tensors)) == 1
+
+    cpu_buffer = torch.empty(sum(p.numel() for p in tensors),
+                             dtype=get_only_unique_item(t.dtype for t in tensors),
+                             device="cpu")
+    tensor_infos: List[Tuple[torch.Tensor, int, int]] = get_mapping_to_flat_buffer(tensors)
+    orig_device = get_only_unique_item(t.device for t in tensors)
+
+    offset = 0
+    for tensor, offset, tensor_numel in tensor_infos:
+        # move the tensor from device memory to host memory
+        cpu_buffer.narrow(0, offset, tensor_numel).copy_(tensor)
+        tensor.data = torch.empty(0, dtype=tensor.dtype, device=tensor.device)
+
+    gc.collect()
+    get_accelerator().empty_cache()
+
+    # copy tensors (now flattened and contiguous) back to GPU
+    device_buffer = cpu_buffer.to(orig_device)
+
+    # restore device tensors
+    for tensor, offset, tensor_numel in tensor_infos:
+        tensor.data = device_buffer.narrow(0, offset, tensor_numel)
+
+    return device_buffer
--- a/tests/unit/common.py
+++ b/tests/unit/common.py
@@ -12,6 +12,7 @@ import subprocess
 from abc import ABC, abstractmethod
 from pathlib import Path
 import random
+import tempfile
 import numpy as np
 from typing import Callable, Any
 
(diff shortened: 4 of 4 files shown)


tohtana (commented):
Hi @kashif,

Thank you for opening this PR! I think supporting HF hub kernels is is a significant update.

Regarding the approach, we [check](https://github.com/deepspeedai/DeepSpeed/blob/6c59d5446afd8900218121ecfd00003a6cde2b9e/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L424) if `core_attn_implementation` is in `ALL_ATTENTION_FUNCTIONS` but HF hub kernels like `kernels-community/flash-attn2` is not in the list. So HF hub kernels won’t still be available with this fix.

We probably need to do the proper registration steps:

1. Reject known-bad impls explicitly: eager, flex_attention, and probably paged|eager.
2. If `core_attn_implementation` is an HF hub kernel string, call the HF registration path first. (Using `lazy_import_flash_attention(…)`)
3. Then read `core_attn_function = ALL_ATTENTION_FUNCTIONS[core_attn_implementation]`.
4. Build `uattn` from that original function.
5. Replace that key with `uattn_wrapper`.

Does it make sense to you?

kashif:
thanks @tohtana I have tried to fix all the issues raised, if you can kindly check again?

stas00:
> Reject known-bad impls explicitly: eager, flex_attention, and probably paged|eager.

We actually don't know if flex_attention is bad, we just haven't tried it out. Do you have resources to try it out, Kashif? Same for the others on the list.

That's why we started with approve list, rather than deny.

The only reason eager is denied is that it requires 4D attention_mask which is a bad idea for long sequence.

BTW, SDPA is silently broken with packed samples - when there is no attn mask, it ignores pos ids and attends to the whole sequence instead. Expect bad results. Not sure how to flag that to users - probably need to inspect pos ids and see if they reset at least once and disallow sdpa then.

tohtana:
Hi @kashif,
Thank you for addressing my comments! It looks good to me.

I also think Stas's comment makes sense. Can you try implementing such a validation?
You can refer to transformers' [find_packed_sequence_indices]( https://github.com/huggingface/transformers/blob/aad13b87ed59f2afcfaebc985f403301887a35fc/src/transformers/masking_utils.py#L709).

kashif:
sure @tohtana i can check

stas00:
to make things more exact - it's packed samples + pos ids + 4D `attention_mask=None` where sdpa silently does the wrong thing. I haven't validated but it most likely will do the right thing with 4D attention mask being `not None`- but it can't be used with SP because it becomes too large too quickly.

stas00:
oh, Kashif, I'm being told `eager` has the exact same problem as `sdpa` - so both need to be fixed on the transformers side. Thank you very much!

kashif:
I ran some experiments comparing flash_attention_2, sdpa, and flex_attention with SP=4 on Qwen3-4B (GQA: 32 Q
  heads, 8 KV heads), 8K seq length, 10 steps.

  Without SP (1 GPU baseline): flash_attention_2 and sdpa produce identical losses — confirming the backends are
  equivalent in the standard path.
```
  ┌──────┬───────┬───────┐
  │ Step │  fa2  │ sdpa  │
  ├──────┼───────┼───────┤
  │ 1    │ 0.736 │ 0.737 │
  ├──────┼───────┼───────┤
  │ 2    │ 1.841 │ 1.843 │
  ├──────┼───────┼───────┤
  │ 3    │ 0.806 │ 0.807 │
  └──────┴───────┴───────┘
```

With SP=4 (4 GPUs): sdpa and flex_attention match each other, but both diverge significantly from
  flash_attention_2:
```
  ┌──────┬──────┬──────┬──────┐
  │ Step │ fa2  │ sdpa │ flex │
  ├──────┼──────┼──────┼──────┤
  │ 1    │ 2.37 │ 4.55 │ 4.55 │
  ├──────┼──────┼──────┼──────┤
  │ 5    │ 2.28 │ 3.52 │ 3.52 │
  ├──────┼──────┼──────┼──────┤
  │ 10   │ 2.29 │ 3.02 │ 3.02 │
  └──────┴──────┴──────┴──────┘
```

  @stas00 any ideas on what flash_attention_2 might be doing differently after the all-to-all that
  sdpa/flex_attention aren't? The Q/K/V shapes and attention_mask=None + is_causal=True path should be equivalent,
  but something in the SP gather/scatter is exposing a difference.

kashif:
ok @stas00 I now enerate position_ids if missing from batch,  build causal BlockMask for flex_attention and do a one-time packed sample validation for packed samples + sdpa/eager

Now the outputs are matching:
```
  ┌──────┬───────────────────┬───────┬────────────────┐
  │ Step │ flash_attention_2 │ sdpa  │ flex_attention │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 1    │ 2.152             │ 2.152 │ 2.150          │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 2    │ 2.469             │ 2.468 │ 2.468          │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 3    │ 2.045             │ 2.044 │ 2.045          │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 4    │ 2.197             │ 2.197 │ 2.197          │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 5    │ 2.113             │ 2.112 │ 2.112          │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 6    │ 2.173             │ 2.173 │ 2.172          │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 7    │ 2.351             │ 2.350 │ 2.351          │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 8    │ 2.380             │ 2.380 │ 2.379          │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 9    │ 1.847             │ 1.847 │ 1.847          │
  ├──────┼───────────────────┼───────┼────────────────┤
  │ 10   │ 2.151             │ 2.151 │ 2.151          │
  └──────┴───────────────────┴───────┴────────────────┘
```

stas00:
Thank you for running those quality comparison experiments, Kashif

I'm a bit unclear about your last "success" comment - what was missing to make FA2 match? are you saying the mismatch was from missing position_ids? but we said that already that SDPA (and now most likely FlexAttenion) have a trouble with no-attn-mask / yes-pos-id and will ignore packed samples. SDPA on the other hand does the right thing here.

And it's great to hear Flex Attention works as well with Ulysses, so we could add it to the allow list.

deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -272,20 +278,29 @@ def forward(
...
+            # One-time check: detect packed samples (multiple documents concatenated in one
+            # sequence) with attn backends that silently break. We check LOCAL (pre-gather)
+            # position_ids for resets, which indicate document boundaries from packing.
+            if not self._packed_sample_validated:
+                self._packed_sample_validated = True
+                local_pos_ids = kwargs["position_ids"]
+                pos_diff = torch.diff(local_pos_ids, dim=-1)
+                has_packed_samples = (pos_diff < 0).any().item()
+                if has_packed_samples and self.core_attn_implementation in ("sdpa", "eager"):
+                    raise ValueError(
```
stas00: heh, I thought we were discussing that it's HF Transformers that has to do that, not Ulysses SP. It affects all users regardless of whether they use Ulysses or not. Unless HF Transformers disallows not providing attn-mask with sdpa/eager, which I don't think is the case.

deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -311,6 +326,28 @@ def forward(
...
+        # XXX: currently hardcodes a causal mask_mod — models with sliding window or other
+        # non-standard patterns would need the mask_mod extracted from the original BlockMask.
+        if self._flex_block_mask_cls is not None and isinstance(attention_mask, self._flex_block_mask_cls):
+            seq_len = query_layer.shape[2]
+            batch_size = query_layer.shape[0]
+
+            def causal_mask(batch_idx, head_idx, q_idx, kv_idx):
+                return q_idx >= kv_idx
+
+            attention_mask = self._flex_create_block_mask(
```
stas00: what are the dimensions of this attn-mask? - if it's like 4D attn mask I don't think we want to support Flex Attn since 4D attn-mask defeats the purpose since it's too big.

deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -574,6 +653,16 @@ def refill(self):
         micro_batches = defaultdict(dict)
         # XXX: replace with more efficient all-to-all?
 
+        # Ensure position_ids exist before all_gather and sharding.
+        # Without this, the Trainer generates local position_ids [0,...,chunk_len-1]
+        # on each rank AFTER sharding, which after all_gather in UlyssesSPAttentionHF
+        # looks like packed sequences [0,...,N, 0,...,N, ...]. flash_attention_2 handles
+        # this via flash_varlen_fn, but sdpa/flex_attention apply full causal masking
+        # across the resets, producing incorrect attention.
+        if "position_ids" not in batch:
```
stas00, line 654: I'm not sure about this. This might lead to a user getting the wrong behavior if they packed samples but forgot to supply pos ids. Should we simply assert if pos ids aren't there and not potentially create invalid pos ids?

I agree there needs to be a check and it's not there.

kashif:
> Thank you for running those quality comparison experiments, Kashif
> 
> I'm a bit unclear about your last "success" comment - what was missing to make FA2 match? are you saying the mismatch was from missing position_ids? but we said that already that SDPA (and now most likely FlexAttenion) have a trouble with no-attn-mask / yes-pos-id and will ignore packed samples. SDPA on the other hand does the right thing here.
> 
> And it's great to hear Flex Attention works as well with Ulysses, so we could add it to the allow list.

So, FA2 was the one producing correct results, while SDPA/flex were wrong. Here's what was happening:

When `position_ids` are not in the dataloader batch (common with SFTTrainer + `packing=False`), `UlyssesSPDataLoaderAdapter` doesn't generate them. The Trainer then generates `position_ids = [0,...,chunk_len-1]` on each rank AFTER the adapter has already sharded the sequence. After `all_gather` in `UlyssesSPAttentionHF.forward()`, the concatenated `position_ids` become:
```
 [0,...,2047, 0,...,2047, 0,...,2047, 0,...,2047]  # looks like 4 packed documents!
```
  FA2 "accidentally" handles this correctly — `_is_packed_sequence()` detects the resets and switches to `flash_varlen_fn`, treating each shard as a separate document. This gives correct attention within each shard.

SDPA with `is_causal=True` applies a simple lower-triangular causal mask over the entire gathered sequence, allowing tokens to attend across the `position_id` resets. This produced `loss=4.55 `vs FA2's correct `loss=2.37`.

The fix: generate `position_ids` in `UlyssesSPDataLoaderAdapter.refill()` BEFORE `all_gather` and sharding, so each rank gets correct global positions (rank 1 gets `[2048,...,4095]`, not `[0,...,2047]`). After gather they reconstruct to monotonic `[0,...,8191]` — no  resets, all backends produce identical results.

With this fix, all three backends match within numerical precision:
```
  ┌──────┬───────┬───────┬───────┐
  │ Step │  FA2  │ SDPA  │ Flex  │
  ├──────┼───────┼───────┼───────┤
  │ 1    │ 2.152 │ 2.152 │ 2.150 │
  ├──────┼───────┼───────┼───────┤
  │ 5    │ 2.113 │ 2.113 │ 2.112 │
  ├──────┼───────┼───────┼───────┤
  │ 10   │ 2.151 │ 2.152 │ 2.151 │
  └──────┴───────┴───────┴───────┘
```
For `flex_attention`, we also needed to rebuild the `BlockMask` for the full gathered sequence length after the all-to-all (the wrapper discards the original one since it was built for the local shard).

deepspeed/runtime/sequence_parallel/ulysses_sp.py
kashif: agree, removed from DeepSpeed side

deepspeed/runtime/sequence_parallel/ulysses_sp.py
kashif: It's a compressed sparse representation. `BlockMask` stores a bitmask over blocks and does not materialize the full square matrix

deepspeed/runtime/sequence_parallel/ulysses_sp.py
kashif: yes, It would need to be in the TRL trainer, for the collator to always provide position_ids when SP is enabled, so the adapter never needs to generate them. I Can try to fix it there.

deepspeed/runtime/sequence_parallel/ulysses_sp.py
stas00: ok, then it should be compact, thank you.

deepspeed/runtime/sequence_parallel/ulysses_sp.py
stas00: Thank you, Kashif.

And probably then add an assert on SP side if pos id isn't there?

stas00:
great explanations, Kashif - thank you!

1. let's assert if pos ids isn't there, trusting that the user will set it up correctly. Generating a warning doesn't guarantee the user will see. But an assert and telling them to do it correctly is probably the safest/resilient way forward.
2. and as you said a special treatment needs to be added for BlockMask for flex attn - I'm not familiar with this one, so will see your implementation when you get a chance to add it.

Thank you, Kashif

kashif:
@stas00, regarding point 2,  we added `BlockMask` handling for `flex_attention` in these places:
1. `uattn_wrapper`: keeps the `BlockMask` instead of discarding it (other mask types are set to `None`)
2. `UlyssesSPAttentionHF.forward()`: rebuilds the `BlockMask` for the full gathered sequence length after the all-to-all (the original was built for the local shard)
3. `register_with_transformers()`: imports `BlockMask` and `create_block_mask` once and stores them on the instance  (only when `core_attn_implementation == "flex_attention"`)

stas00:
Thank you very much, Kashif. 

Do you think all this amazing tooling you added should live here and not in HF Transformers?

kashif:
checking

kashif:
So some SP-specific things tied to the all-to-all make sense to be here...

  - `position_ids` assert in the adapter: the adapter owns the sharding contract and needs `position_ids` before it shards
  - `BlockMask` rebuild in `forward()`: after the all-to-all, the gathered sequence is longer than the original mask was built for. This is inherently SP logic I believe
  - Smart mask handling in `uattn_wrapper`: distinguishing `BlockMask` (keep) from 4D tensor masks (discard) is specific to how SP nulls out masks
  - `flex_attention` allow/deny list: SP-specific compatibility
  
 Agree that in Transformers:
   - The packed samples + sdpa/eager silent failure: any user with packed sequences + `attention_mask=None` gets wrong results, regardless of SP. sdpa/eager should respect `position_ids` for document boundaries like FA2 does (or at minimum warn)
   
On the TRL side:
  * Ensuring `position_ids` are always in the batch when SP is enabled: currently requires `padding_free=True`. The collator could detect SP and always emit `position_ids`

stas00:
Thank you for the detailed summary, Kashif.

I agree with everything, except:

> sdpa/eager should respect position_ids for document boundaries like FA2 does (or at minimum warn)

I think it should assert. Warnings don't work and allowing invalid training can be so so costly to the user who missed the warning in the sea of warnings. I wonder how many people will discover their model has been mistrained and they had no clue that was the case, other than getting bad outcomes.

stas00:
Please let us know when things are ready for the final review, Kashif.

kashif:
thanks @stas00 yes we are asserting the `position_ids` in the adapter see: https://github.com/deepspeedai/DeepSpeed/pull/7887/changes#diff-93102c31b16242dc2a15b2cee19920eb9c228d298863c4c4218ac93c2c62ab6bR652

kashif:
yes ready for review, thanks!

stas00:
> thanks @stas00 yes we are asserting the `position_ids` in the adapter see: https://github.com/deepspeedai/DeepSpeed/pull/7887/changes#diff-93102c31b16242dc2a15b2cee19920eb9c228d298863c4c4218ac93c2c62ab6bR652

I meant inside transformers. Currently transformers may provide a disservice to users if they use packed samples w/o attention w/ sdpa/eager - or is it the case that transformers enforces 4D attention mask?

deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -411,15 +440,34 @@ def register_with_transformers(
...
+
+        # eager always materializes a 4D attention_mask (O(n²) memory) and cannot fall back
+        # to is_causal=True like sdpa — so it's incompatible with SP which discards masks.
+        unsupported_attn_implementation = ["eager", "paged|eager"]
+        if core_attn_implementation in unsupported_attn_implementation:
             raise ValueError(
                 f"{core_attn_implementation} attn_implementation isn't currently supported by Ulysses sequence"
-                f" parallelism. Set core_attn_implementation arg to one of {supported_attn_implementation}.")
+                f" parallelism because it requires a 4D attention_mask (O(n²) memory)."
+                f" Use 'flash_attention_2', 'flash_attention_3', 'flex_attention', 'sdpa',"
```
stas00: Should we future proof this for fa and say any official flash attention version?

tests/unit/ulysses_alst/test_ulysses_sp_hf.py
```diff
@@ -301,3 +301,87 @@ def test_disable_in_eval(self):
         # Verify: with disable_in_eval=True, full sequence input should produce
         # the same output as baseline (SP is bypassed)
         torch_assert_equal(logits_baseline, logits_sp)
+
+
+class TestUlyssesSPHFHubKernel(DistributedTest):
```
stas00, line 309: since your PR adds support for flex attention would it be difficult to add a test exercising this path?

stas00 (changes requested):
Kashif, 

overall looks great - added a few suggestions

- do you think we should discuss the different supported attn types in the tutorial as well?

https://github.com/deepspeedai/DeepSpeed/blob/5f7b687018bd1e0340c661859820fd97aa80a616/docs/_tutorials/ulysses-alst-sequence-parallelism.md

if so let's add a brief section there?

- also we can now probably test fa4 and add it to the list - fa4 support has been merged in transformers a few days ago.

kashif:
the test is failing... i am investigating

kashif:
ok tests passing now

deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -288,6 +286,11 @@ def forward(
             position_ids_list = [torch.empty_like(kwargs["position_ids"]) for _ in range(self.world_size)]
             dist.all_gather(position_ids_list, kwargs["position_ids"], group=self.process_group)
             kwargs["position_ids"] = torch.cat(position_ids_list, dim=1)
+        else:
+            logger.warning_once(
```
stas00: I say this should be an assert, warnings don't work. 

I feel we should not allow not having pos ids at all when doing Ulysses. Please correct me if I'm wrong, Kashif.

stas00: If the samples aren't packed, it shouldn't be too hard to still provide pos ids, no? 

If you can remind me would the HF ecosphere always send pos ids, regardless of whether the samples are packed or not?

deepspeed/runtime/sequence_parallel/ulysses_sp.py
```diff
@@ -315,30 +318,35 @@ def forward(
...
-                    mask_mod=causal_mask,
-                    B=batch_size,
-                    H=None,
-                    Q_LEN=seq_len,
-                    KV_LEN=seq_len,
-                    device=query_layer.device,
-                    _compile=True,
-                )
+        if self.core_attn_implementation == "flex_attention":
+            from torch.nn.attention.flex_attention import BlockMask, create_block_mask
```
stas00, line 323: Should we move all these imports out to the top - not sure the overhead they create on the subsequent invocations but this is a tight forward loop, so surely they add up to some overhead, no?

stas00: there are others below, I think mainly flex attn code paths.

deepspeed/runtime/sequence_parallel/ulysses_sp.py
kashif: - Transformers Trainer: `_prepare_context_parallel_inputs` auto-generates `position_ids` if missing (simple `arange` for non-packed)
- TRL `SFTTrainer`: Forces `padding_free=True` on the collator when Ulysses SP is active, which always emits position_ids (boundary-aware   for packed, `arange` for non-packed)

stas00: so do we need to require pos ids with Ulysses? assert if it's not there?

kashif: When called from HF, we do not need it, but when going through other trainers that call `UlyssesSPAttentionHF` manually, it might help catch silent errors without the assert.

Issue #7833, opened 2026-02-06
[BUG] Ulysses crashes in Stage 0


avicooper1:
**Describe the bug**
Training with Ulysses sequence parallel in Stages 1/2/3 works out of the box. Using the same training pipeline with Stage 0 crashes with the following error stack trace:
```
File "/home/acooper/KosupaLLM/.venv/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 2470, in step
  self._take_model_step(lr_kwargs)
File "/home/acooper/KosupaLLM/.venv/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 2365, in _take_model_step
  self.optimizer.step()
File "/home/acooper/KosupaLLM/.venv/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
  return func(*args, **kwargs)
File "/home/acooper/KosupaLLM/.venv/lib/python3.10/site-packages/deepspeed/runtime/bf16_optimizer.py", line 278, in step
  non_expert_grads_for_norm, expert_grads_for_norm = self.get_grads_for_norm()
File "/home/acooper/KosupaLLM/.venv/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
  return func(*args, **kwargs)
File "/home/acooper/KosupaLLM/.venv/lib/python3.10/site-packages/deepspeed/runtime/bf16_optimizer.py", line 394, in get_grads_for_norm
  tensor_mp_rank = bwc_tensor_model_parallel_rank(mpu=self.mpu)
File "/home/acooper/KosupaLLM/.venv/lib/python3.10/site-packages/deepspeed/utils/bwc.py", line 42, in bwc_tensor_model_parallel_rank
  return mpu.get_model_parallel_rank()
AttributeError: module 'deepspeed.runtime.sequence_parallel.parallel_state_sp' has no attribute 'get_model_parallel_rank'
```

As noted, the Ulysses code does not contain a method for `get_model_parallel_rank`.

Looking through the stack trace, the optimizer class for Stage 0 differs from that of Stage 1 (which makes sense, Stage 1 parallelizes the optimizer,) ie. `deepspeed.runtime.bf16_optimizer.BF16_Optimizer` and `deepspeed.runtime.zero.stage_1_and_2.DeepSpeedZeroOptimizer` respectively. This code divergence means that only the Stage 0 optimizer calls this problematic line.

**To Reproduce**
[...]

PR #7649, opened 2025-10-25
ulysses mpu: additional api

stas00:
It looks like `save_checkpoint` expects `get_model_parallel_*` API in the `mpu` object. So adding it to the Ulysses slim mpu version.

This solves this problem in HF Trainer:
```
[rank1]:   File "/code/users/stas/github/transformers-alst-integration/src/transformers/trainer.py", line 3248, in _save_optimizer_and_scheduler
[rank1]:     self.model_wrapped.save_checkpoint(output_dir)
[rank1]:   File "/code/users/stas/github/DeepSpeed/deepspeed/runtime/engine.py", line 3497, in save_checkpoint
[rank1]:     self._save_checkpoint(save_dir,
[rank1]:   File "/code/users/stas/github/DeepSpeed/deepspeed/runtime/engine.py", line 3709, in _save_checkpoint
[rank1]:     save_path = self._get_ckpt_name(save_dir, tag)
[rank1]:                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/code/users/stas/github/DeepSpeed/deepspeed/runtime/engine.py", line 3039, in _get_ckpt_name
[rank1]:     mp_rank = 0 if self.mpu is None else self.mpu.get_model_parallel_rank()
[rank1]:                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]: AttributeError: module 'deepspeed.runtime.sequence_parallel.parallel_state_sp' has no attribute 'get_model_parallel_rank'. Did you mean: 'get_sequence_parallel_rank'?
```

PR #7888, opened 2026-03-06
Fix Stage 0 + Ulysses crash: make bwc_tensor_model_parallel_rank() resilient to MP API absence

nathon-lee:
## Title
Fix Stage 0 + Ulysses crash: make `bwc_tensor_model_parallel_rank()` resilient to MP API absence

## Summary
This PR fixes a hard crash when using Ulysses sequence parallelism with ZeRO Stage 0 (BF16_Optimizer).  
In this configuration, DeepSpeed calls `deepspeed.utils.bwc.bwc_tensor_model_parallel_rank(mpu=...)`, and the passed `mpu` object can be `deepspeed.runtime.sequence_parallel.parallel_state_sp`, which does not implement the deprecated `get_model_parallel_rank()` API. The current fallback path unconditionally calls `mpu.get_model_parallel_rank()`, raising `AttributeError`.

The fix adds a defensive capability check before calling the deprecated API. If the provided `mpu` does not expose any known tensor/model-parallel rank API, we treat it as “no tensor model parallelism” and return rank `0`.

## Motivation / Context
- Affected scenario: Ulysses sequence parallel + ZeRO Stage 0
- Failure mode: `AttributeError: ... parallel_state_sp has no attribute get_model_parallel_rank`
- Root cause: `bwc_tensor_model_parallel_rank()` falls back to a deprecated API without an `hasattr()` check.

This change keeps the original priority order intact:
1) `get_tensor_model_parallel_rank()`
2) `get_slice_parallel_rank()`
3) `get_model_parallel_rank()` (deprecated)
4) fallback to `0` if none exist

## Changes
- `deepspeed/utils/bwc.py`
  - Update `bwc_tensor_model_parallel_rank()` to check `hasattr(mpu, "get_model_parallel_rank")` before calling it.
  - If `mpu` provides none of the expected tensor/model-parallel rank APIs, return `0` (no TP).

## Why this is safe
- For Megatron / DeepSpeed Topology / any existing MPU that already implements `get_tensor_model_parallel_rank()` or `get_slice_parallel_rank()` or `get_model_parallel_rank()`, behavior is unchanged.
- The new code path only affects the previously-crashing case where the `mpu` object does not provide any of these methods.

## Reproduction
Using the Ulysses ALST tutorial flow, switching ZeRO stage from 3 to 0 triggers the crash during optimizer step when grad norm is computed.

## Testing
- Existing unit tests should continue to pass.
- Minimal repro: calling `bwc_tensor_model_parallel_rank(mpu=deepspeed.runtime.sequence_parallel.parallel_state_sp)` should no longer raise.

## References
- DeepSpeed Issue: #7833 “Ulysses crashes in Stage 0”

### DIFF
--- a/deepspeed/utils/bwc.py
+++ b/deepspeed/utils/bwc.py
@@ -37,9 +37,13 @@ def bwc_tensor_model_parallel_rank(mpu=None):
     elif hasattr(mpu, 'get_slice_parallel_rank'):
         # Some DeepSpeed + pipeline parallelism versions
         return mpu.get_slice_parallel_rank()
-    else:
+    elif hasattr(mpu, 'get_model_parallel_rank'):
         # Deprecated Megatron and DeepSpeed convention
         return mpu.get_model_parallel_rank()
+    else:
+        # mpu does not provide any known tensor/model-parallel rank API.
+        # Treat as "no tensor model parallelism".
+        return 0
 
 
 def bwc_tensor_model_parallel_world_size(mpu=None):


tohtana:
Hi @nathon-lee,
Thank you for reporting! 

I found that we already have a [fallback](https://github.com/deepspeedai/DeepSpeed/blob/6c59d5446afd8900218121ecfd00003a6cde2b9e/deepspeed/runtime/sequence_parallel/parallel_state_sp.py#L95) from `get_model_parallel_world_size` to `get_sequence_parallel_world_size`.
This was introduced in #7649. Can you make sure that the latest version still raises the error?

nathon-lee:
Hi @tohtana, thanks for looking into this!

I double-checked `master` and I think #7649 addresses a *different* fallback path than the one crashing in Stage 0.

- The crash in #7833 happens in **`deepspeed/utils/bwc.py:bwc_tensor_model_parallel_rank()`** (called from `BF16_Optimizer.get_grads_for_norm()` in ZeRO **Stage 0**).
- In current `master`, `bwc_tensor_model_parallel_rank()` falls back to the deprecated API **unconditionally**:

  `else: return mpu.get_model_parallel_rank()`

  When `mpu` is `deepspeed.runtime.sequence_parallel.parallel_state_sp` (Ulysses), that module does **not** implement `get_model_parallel_rank()`, so we still hit:

  `AttributeError: ... parallel_state_sp has no attribute 'get_model_parallel_rank'`

My understanding is that #7649 added/adjusted APIs and fallbacks around *world size* (e.g., `get_model_parallel_world_size` -> `get_sequence_parallel_world_size`), but it doesn't protect the *rank* fallback above.

That’s why I opened #7888: it makes `bwc_tensor_model_parallel_rank()` defensive (checks for `get_model_parallel_rank` before calling it, and otherwise treats it as “no tensor MP” and returns rank 0).

If you have a preferred behavior (e.g., should we fallback to a sequence-parallel rank instead of 0?), I’m happy to adjust the PR.

tohtana:
Hi @nathon-lee,
Thank you for the reply. I still don't see why this issue happens with the latest master.
Can you share a small repro?

Issue #7852, opened 2026-02-13
[Bug] Missing input validation for fp16 loss_scale accepts float('inf')


amadhan882:
### **Describe the bug**
The `fp16` configuration in DeepSpeed accepts `float('inf')` for the `loss_scale` parameter without any validation. While other parameters like `gradient_accumulation_steps` and `stage3_max_live_parameters` correctly trigger `AssertionError` or `ValidationError` for invalid/negative inputs, `loss_scale` set to infinity allows the engine to initialize successfully. This leads to a logic leak where gradients will silently become `NaN` during training.

### **To Reproduce**
The following script was used on Google Colab to test logic edge cases. The "Infinite Scale Test" successfully bypasses validation while others are caught.

**Reproduction Code:**
```python
import deepspeed
import torch
import torch.nn as nn

class TinyModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(1, 1)
    def forward(self, x): return self.linear(x)

def run_test(name, config):
    print(f"\n[#] Testing: {name}")
    print(f"[-] Config: {config}")
    try:
        model = TinyModel()
        engine, _, _, _ = deepspeed.initialize(
            model=model,
            config=config,
            model_parameters=model.parameters()
        )
        print("[+] STATUS: EXECUTED (Possible Logic Leak)")
    except Exception as e:
        print(f"[X] CRASH: {type(e).__name__} -> {e}")

test_cases = {
    "Negative Accumulation": {
        "train_batch_size": 16,
        "gradient_accumulation_steps": -1
    },
    "Zero Discovery Stage": {
        "train_batch_size": 16,
        "zero_optimization": {
            "stage": 3,
            "stage3_max_live_parameters": -100
        }
    },
    "Infinite Scale Test": {
        "train_batch_size": 16,
        "fp16": {
            "enabled": True,
            "loss_scale": float('inf')
        }
    }
}

if __name__ == "__main__":
    for name, cfg in test_cases.items():
        run_test(name, cfg)
```

**Observed Output:**
```text
[#] Testing: Negative Accumulation
[-] Config: {'train_batch_size': 16, 'gradient_accumulation_steps': -1}
[X] CRASH: AssertionError -> Micro batch size per gpu: -16 has to be greater than 0

[#] Testing: Zero Discovery Stage
[-] Config: {'train_batch_size': 16, 'zero_optimization': {'stage': 3, 'stage3_max_live_parameters': -100}}
[...]

PR #7889, opened 2026-03-06
fix: Validate fp16.loss_scale is finite and non-negative

nathon-lee:
Validate fp16.loss_scale is finite and non-negative

Add a Pydantic field validator to DeepSpeedFP16Config to reject NaN/inf/-inf and negative values for fp16.loss_scale (while keeping 0 as dynamic loss scaling). This prevents invalid configs from silently initializing and causing NaNs during training.

Test:
Run pytest -q tests/unit/runtime/test_precision_config_loss_scale.py

Result: 
```
root@72170d0458e9:/home/DeepSpeed_woo# pytest -q tests/unit/runtime/test_precision_config_loss_scale.py
=================================================================== test session starts ===================================================================
platform linux -- Python 3.11.10, pytest-8.3.5, pluggy-1.6.0 -- /usr/bin/python
cachedir: .pytest_cache
Using --randomly-seed=1526199052
rootdir: /home/DeepSpeed_woo/tests
configfile: pytest.ini
plugins: xdist-3.8.0, randomly-4.0.1, forked-1.6.0, anyio-4.6.0
collected 10 items

tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_accepts_valid_values[3] PASSED                                         [ 10%]
tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_accepts_valid_values[0] PASSED                                         [ 20%]
tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_rejects_invalid_values[inf] PASSED                                     [ 30%]
tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_accepts_valid_values[1] PASSED                                         [ 40%]
tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_rejects_invalid_values[nan] PASSED                                     [ 50%]
tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_accepts_valid_values[2.0] PASSED                                       [ 60%]
tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_rejects_invalid_values[True] PASSED                                    [ 70%]
tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_invalid_type_has_clear_error[loss_scale0] PASSED                       [ 80%]
tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_rejects_invalid_values[-1] PASSED                                      [ 90%]
tests/unit/runtime/test_precision_config_loss_scale.py::test_fp16_loss_scale_invalid_type_has_clear_error[loss_scale1] PASSED                       [100%]

(30 durations < 1s hidden.  Use -vv to show these durations.)
============================================================= 10 passed, 16 warnings in 4.18s =============================================================
```
Fix issue #7852

### DIFF
--- a/csrc/deepspeed4science/evoformer_attn/gemm_kernel_utils.h
+++ b/csrc/deepspeed4science/evoformer_attn/gemm_kernel_utils.h
@@ -42,45 +42,28 @@
 template <typename arch, typename scalar_t>
 struct CheckArch {
     static constexpr bool isPreVolta = arch::kMinComputeCapability < 70;
-    static constexpr bool isPreAmpere =
-        arch::kMinComputeCapability < 80 && arch::kMinComputeCapability >= 70;
+    // DISPATCH_ARCHTAG only binds Sm70/Sm75/Sm80+, so overlap with isPreVolta is unreachable.
+    static constexpr bool isPreAmpere = arch::kMinComputeCapability < 80;
     static constexpr bool isAmpere = arch::kMinComputeCapability >= 80;
-#if defined(__CUDA_ARCH__)
-    static constexpr bool compiler_cc = arch::kMinComputeCapability * 10 <= __CUDA_ARCH__;
-#else
-    static constexpr bool compiler_cc = true;
-#endif
     static constexpr bool value = (isPreVolta && std::is_same_v<scalar_t, float>) ||
                                   (isPreAmpere && !std::is_same_v<scalar_t, cutlass::bfloat16_t>) ||
-                                  isAmpere && compiler_cc;
+                                  isAmpere;
 };
 
-#define DISPATCH_ARCHTAG(CC, func)                                                      \
-    {                                                                                   \
-        if constexpr (GPU_ARCH >= 80) {                                                 \
-            if (CC >= 80) {                                                             \
-                using ArchTag = cutlass::arch::Sm80;                                    \
-                func;                                                                   \
-            } else {                                                                    \
-                EVOFORMER_CHECK(false, "Compile flag error. Unexpected GPU");           \
-            }                                                                           \
-        } else if constexpr (GPU_ARCH >= 75) {                                          \
-            if (CC >= 75) {                                                             \
-                using ArchTag = cutlass::arch::Sm75;                                    \
-                func;                                                                   \
-            } else {                                                                    \
-                EVOFORMER_CHECK(false, "Compile flag error. Unexpected GPU");           \
-            }                                                                           \
-        } else if constexpr (GPU_ARCH >= 70) {                                          \
-            if (CC >= 70) {                                                             \
-                using ArchTag = cutlass::arch::Sm70;                                    \
-                func;                                                                   \
-            } else {                                                                    \
-                EVOFORMER_CHECK(false, "Compile flag error. Unexpected GPU");           \
-            }                                                                           \
-        } else {                                                                        \
-            EVOFORMER_CHECK(false, "Only GPUs with Tensor Core are supported for now"); \
-        }                                                                               \
+#define DISPATCH_ARCHTAG(CC, func)                                                         \
+    {                                                                                      \
+        if ((CC) >= 80) {                                                                  \
+            using ArchTag = cutlass::arch::Sm80;                                           \
+            func;                                                                          \
+        } else if ((CC) >= 75) {                                                           \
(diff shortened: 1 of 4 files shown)


deepspeed/runtime/precision_config.py
```diff
@@ -107,12 +109,31 @@ class DeepSpeedFP16Config(DeepSpeedConfigModel):
...
     Automatically cast inputs to fp16
     """
-
+    
     loss_scale: float = 0
     """
     Loss scaling value. Default value of 0 means dynamic loss scaling instead of static loss scale.
     """
 
+    @field_validator("loss_scale")
```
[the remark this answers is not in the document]
tohtana: I think this comment makes sense. Can you address it? @nathon-lee

deepspeed/runtime/precision_config.py
```diff
@@ -107,12 +109,31 @@ class DeepSpeedFP16Config(DeepSpeedConfigModel):
...
+    @classmethod
```
PKUWZP, line 119: - Consider using `mode="before"` for the entire validator rather than splitting into two validators. A single
  `mode="before" `validator can handle both the bool check and the finite/negative checks:

  ```@field_validator("loss_scale", mode="before")
  @classmethod
  def _validate_loss_scale(cls, v):
      if isinstance(v, bool):
          raise ValueError("fp16.loss_scale must be a number, not bool")
      v = float(v)
      if not math.isfinite(v):
          raise ValueError("fp16.loss_scale must be a finite number (not inf/-inf/nan)")
      if v < 0:
          raise ValueError("fp16.loss_scale must be >= 0 (0 enables dynamic loss scaling)")
      return v ```

 
 - Test coverage: There are no tests included. A few unit tests in `tests/unit/runtime/` asserting that invalid loss_scale values `(-1, float('inf'), float('nan'), True)` raise `ValidationError` would strengthen this PR and prevent regressions.

The existing pattern in the repo uses `DeepSpeedFP16Config(loss_scale=...)` directly, which makes such tests straightforward.

deepspeed/runtime/precision_config.py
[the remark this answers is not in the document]
nathon-lee: Thanks — I agree this comment makes sense. I’ll address it and push an update shortly. @tohtana

deepspeed/runtime/precision_config.py
nathon-lee: Thanks — good suggestion. I’ll consolidate into a single `mode="before"` validator and add unit tests (e.g. `-1`, `inf`, `nan`, `True` -> `ValidationError`) using `DeepSpeedFP16Config(loss_scale=...)`. I’ll push an update shortly. @PKUWZP

nathon-lee: Thanks for the review — addressed: the `loss_scale` validator now runs with `mode="before"` (so bools are rejected prior to coercion) and I added unit tests for (-1, inf, nan, True).

deepspeed/runtime/precision_config.py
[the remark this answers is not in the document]
nathon-lee: Thanks for the review — addressed: the `loss_scale` validator now runs with `mode="before"` (so bools are rejected prior to coercion) and I added unit tests for (-1, inf, nan, True).

deepspeed/runtime/precision_config.py
tohtana: @nathon-lee If we pass an invalid value like `[]` and `{}`, won't `float()` raise `TypeError` now?
The current master raises Pydantic's `ValidationError` for these, which is clearer than a raw `TypeError`.

nathon-lee: @tohtana  Thanks for catching this.
﻿
I added a `try/except (TypeError, ValueError)` around `float(v)` in the `mode="before"` validator so invalid types (e.g. `[]`, `{}`) are converted into a clear `ValueError`, which Pydantic wraps as `ValidationError` (instead of surfacing a raw `TypeError`). I also added unit tests covering `[]` and `{}` to prevent regressions.

tohtana:
@PKUWZP Can you confirm that your request has been met? We need your confirmation to merge this PR.

PR #7890, opened 2026-03-06
[Bloom] Fix hangs of bloom test

k-artem:
`test_checkpoint_sharding.py::TestCheckpointShard::test[bigscience/bloom-560m-fp16]` silently hangs if incorrect version of transformers(>4.43.4) is installed, due to no proper handling of SystemExit in the pool worker. 

Fix is throw RuntimeError instead of sys.exit.

* Test:
Run `pytest tests/unit/inference/test_checkpoint_sharding.py -k 'bloom-560m-fp16'`

* Result: no hangs, throw exception:
```
E           RuntimeError: Transformers version 4.57.3 exceeds version 4.43.4! After transformers version 4.43.4, BLOOM inference with DeepSpeed is no longer supported.

/usr/lib/python3.12/multiprocessing/pool.py:774: RuntimeError
```

### DIFF
--- a/deepspeed/runtime/zero/parameter_offload.py
+++ b/deepspeed/runtime/zero/parameter_offload.py
@@ -110,7 +110,7 @@ class DeepSpeedZeRoOffload(object):
         log_trace_cache_warnings=False,
     ):
 
-        see_memory_usage("DeepSpeedZeRoOffload initialize [begin]", force=True)
+        see_memory_usage("DeepSpeedZeRoOffload initialize [begin]", force=False)
 
         print_rank_0(f"initialized {__class__.__name__} with args: {locals()}", force=False)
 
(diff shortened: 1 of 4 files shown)


PKUWZP (approved):
Thanks for sending the fix. This PR fixes a test hang in `test_checkpoint_sharding.py` for the BLOOM 560m model. The root cause is that `sys.exit()` is called inside a multiprocessing pool worker when an incompatible transformers version is detected. `sys.exit()` raises `SystemExit`, which `multiprocessing.Pool` does not propagate back to the parent process cleanly — it silently hangs instead. The fix replaces `sys.exit()` with `raise RuntimeError(...)`, which is properly caught and re-raised by the pool. 

If @loadams and @hwchen2017 do not have questions I approve this PR fix.

PR #7891, opened 2026-03-08
Suppress see_memory_usage logs

sfc-gh-truwase:
Make zero initialization less noisy


sfc-gh-sbekman:
Thank you for fixing, Tunji!

PR #7898, opened 2026-03-12
fix(zero3): use current_stream() instead of default_stream() for grad…

michaelroyzen:
fix(zero3): use current_stream() instead of default_stream() for gradient reduce-scatter sync

## Problem

DeepSpeed ZeRO Stage 3 produces NaN values in weight partitions after the optimizer step when used with PyTorch 2.10+ and bf16 training. The NaN appears stochastically across ranks and affects different layers on each run (typically attention projections and MLP weights). Training works correctly with identical configuration on older PyTorch versions.

## Root Cause

In `__add_grad_to_ipg_bucket` (stage3.py), the reduce-and-partition stream synchronizes with the **default** CUDA stream before copying gradients:

```python
self.reduce_and_partition_stream.wait_stream(get_accelerator().default_stream())
```

However, PyTorch's autograd engine executes each backward op on the **same stream its forward op used**. In ZeRO-3, forward ops involve all-gather operations that may execute on non-default streams. When the corresponding backward ops produce gradients on those non-default streams, `wait_stream(default_stream)` creates a dependency on the wrong stream — the reduce-scatter proceeds before the gradient CUDA kernel has finished writing, reading uninitialized memory from the gradient buffer.

This was not observable on older PyTorch versions because the autograd engine's stream assignment behavior was more conservative. PyTorch 2.10 introduced changes to autograd stream handling that make this race condition reliably trigger when gradient magnitudes are large enough for the resulting NaN to be distinguishable from valid values.

## Minimal Reproduction

1. Configure ZeRO Stage 3 with bf16, `overlap_comm: true`, `contiguous_gradients: true`
2. Train any model where backward produces gradients on non-default CUDA streams (e.g., multi-GPU training with PyTorch 2.10+)
3. After the first optimizer step, some weight partitions contain NaN on a subset of ranks
4. The affected layers and ranks vary between runs (stochastic, depends on CUDA kernel scheduling)

## Fix

Replace `default_stream()` with `current_stream()` in the synchronization call. Since `__add_grad_to_ipg_bucket` is called from DeepSpeed's gradient hook (which executes on the same stream as the backward op), `current_stream()` correctly captures whatever stream actually computed the gradient:

```python
self.reduce_and_partition_stream.wait_stream(get_accelerator().current_stream())
```

This is a zero-overhead change — `wait_stream` uses CUDA events internally, which cost ~1µs per parameter regardless of which stream is waited on.

## Verification

Tested with Qwen3-4B on 7×H200 GPUs, DeepSpeed 0.18.7, PyTorch 2.10.0, CUDA 12.8, NCCL 2.27.5:
- **Before fix**: 150K+ NaN values across 55 weight layers after step 1, grad_norm clipped to 1.0 (corrupted)
- **After fix**: 0 NaN across all weight layers for 3+ steps, grad_norm healthy at 0.08–0.27

@tjruwase @stas00

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,33 @@
+<!-- This file is duplicated as CLAUDE.md and AGENTS.md. Keep them in sync. -->
+# AGENTS.md — Workspace-level instructions for AI coding agents
+
+## DeepSpeed Project Rules
+
+### Commit & CI requirements
+
+- All commits MUST have a `Signed-off-by` line (use `--signoff`). Get the name and email from `git config user.name` / `git config user.email`.
+- Formatting: yapf (column_limit=119, `.style.yapf`) + flake8 (`.flake8`).
+- Always verify changed files pass pre-commit checks before committing. Config: `.pre-commit-config.yaml`.
+- `check-torchdist` hook: NEVER directly import torch's distributed module. Use `import deepspeed.comm as dist` instead.
+- New files require license header:
+  ```
+  # Copyright (c) Microsoft Corporation.
+  # SPDX-License-Identifier: Apache-2.0
+  # DeepSpeed Team
+  ```
+
+### Code change discipline
+
+- NEVER make cosmetic/formatting-only changes to existing code. Only add/modify lines that are functionally necessary. Minimizing diff noise is critical for code review.
+- Delete dead code decisively — if code is unused at runtime (only referenced in tests), remove it along with its tests.
+- Prefer consolidating tests over proliferating test files.
+- Blend in: when modifying code, read the surrounding context and match the style of neighboring code (naming, spacing, patterns, idioms).
+- Write beginner-friendly code: avoid deeply nested expressions or chained logic. Break complex expressions into clear, named intermediate steps.
+- Comments should explain **why**, not **what**. Describe the purpose and reasoning, not the mechanics that the code already shows.
+- New features must include corresponding tests and documentation updates.
+
+## Tool Caveats
+
+### Edit tool auto-formatter
+
+The Edit tool has a hidden auto-formatter that silently changes quotes, whitespace, blank lines, and line wrapping. For format-sensitive modifications (e.g., when exact formatting matters for pre-commit), use `bash` with `sed`, `python`, or `cat` instead.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,33 @@
+<!-- This file is duplicated as CLAUDE.md and AGENTS.md. Keep them in sync. -->
+# AGENTS.md — Workspace-level instructions for AI coding agents
+
+## DeepSpeed Project Rules
+
+### Commit & CI requirements
+
+- All commits MUST have a `Signed-off-by` line (use `--signoff`). Get the name and email from `git config user.name` / `git config user.email`.
+- Formatting: yapf (column_limit=119, `.style.yapf`) + flake8 (`.flake8`).
+- Always verify changed files pass pre-commit checks before committing. Config: `.pre-commit-config.yaml`.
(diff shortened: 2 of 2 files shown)


stas00:
Thank you for the detailed diagnosis, Michael.

Question: could somehow `current_stream()` be not we expect it to be? e.g. someone switching it to be a different stream in a user application doing something else? It feels like there should be a specific stream specified rather than relying on a current stream, no?

michaelroyzen:
Thanks for the quick response @stas00!

I think that `current_stream()` is correct here because of how PyTorch's autograd engine dispatches backward ops:

PyTorch records the CUDA stream that was active when each forward op executed, and replays the corresponding backward op on that same stream (see [PyTorch CUDA semantics - Stream semantics of backward passes](https://pytorch.org/docs/stable/notes/cuda.html#stream-semantics-of-backward-passes)). DeepSpeed's gradient hook (`register_post_accumulate_grad_hook`) fires synchronously from the autograd engine on that same stream. So `current_stream()` inside the hook is guaranteed to be the stream that produced the gradient — it's not ambient state that a user could have switched, it's the stream the autograd engine explicitly set for this backward op.

The previous `default_stream()` was actually the "rely on convention" approach — it assumed all backward ops run on stream 0, which happened to be true in older PyTorch but is not guaranteed by the API. `current_stream()` is the principled choice because it asks "what stream am I running on right now?" rather than "what stream do I hope the gradient was computed on?"

That said, if there's a preference for being even more explicit, we could capture the stream via a CUDA event recorded inside a `register_hook` (which fires during the backward op itself) and pass it to the reduce-scatter. But that would require more invasive changes and additional state management, and `current_stream()` already gives us the same stream since the post-accumulate hook runs in the same autograd dispatch context.

stas00:
If what you explained is so, your solution works me. But I'd have Tunji or Masahiro 2nd opinion before merging.

Also what's the impact for older pytorch versions? Should it be doing this for pt>=2.10 or is it safe to go back in versions? If you are not sure perhaps it's the safest to make it version specific?

stas00:
and you need to hit the failing DCO test and sign your commit - in the future using `git commit -s ...` will do the same.

michaelroyzen:
Thanks, just signed the commit. I manually verified it also works on pt==2.9.0 and did some digging on the docs:

The backward stream guarantee — "Each backward CUDA op runs on the same stream that was used for its corresponding forward op" — has been documented since [PyTorch 1.9](https://pytorch.org/docs/1.9.0/notes/cuda.html#stream-semantics-of-backward-passes) and is unchanged through [2.10](https://pytorch.org/docs/stable/notes/cuda.html#stream-semantics-of-backward-passes). Since `__add_grad_to_ipg_bucket` is called from DeepSpeed's `register_post_accumulate_grad_hook`, which fires in the autograd engine's backward dispatch context, `current_stream()` is the stream the autograd engine set for this specific backward op — not user-controlled ambient state.

The previous `default_stream()` worked in practice on older setups because the ZeRO-3 forward all-gather ops happened to run on the default stream, so backward ops also landed on the default stream, making `wait_stream(default_stream)` correct by coincidence. With newer PyTorch/NCCL (e.g., shipped with vLLM 0.17), the all-gather ops can run on non-default streams, breaking that assumption.

`current_stream()` is a no-op change on setups where backward already runs on the default stream, and is the correct fix on setups where it doesn't. It's backward compatible all the way to PyTorch 1.9.

stas00:
Michael, this is good, Tunji promised to review today and then he will merge and make a new release since your fix is very critical.

BC-wise we discussed that deepspeed practically should be supporting about 2 years back of pytorch versions, so your research "since pt 1.9" fits the bill.

Thank you again!

sfc-gh-truwase:
Michael, thanks for this critical contribution. Your detailed analysis and explanation are greatly appreciated.

Issue #7861, opened 2026-02-19
(Draft) [Roadmap] DeepSpeed Roadmap Q2 2026


tohtana:
This is a draft roadmap for DeepSpeed Q2 2026. Feedback is welcome — please leave comments on this issue or join the `#2026q2-roadmap` channel on the [DeepSpeed Slack](https://join.slack.com/t/deepspeedworkspace/shared_invite/zt-3a8pjd8dd-PCj2hMvR4Y2syPwVnjEoww).

# New feature and enhancement

## AutoEP support

AutoEP enables Expert Parallelism (EP) for major Mixture-of-Experts (MoE) models out of the box, eliminating the need for users to write model-specific parallelization code. By automatically distributing expert layers across devices, AutoEP allows users to scale MoE training with minimal configuration changes.

A [prototype implementation](https://github.com/tohtana/DeepSpeedExamples/tree/tohtana/add_auto_ep/training/expert_parallel) has been validated on 8xH100, achieving ~5x throughput improvement over ZeRO-3 baselines. We will build on this work to extend AutoEP support to production readiness in Q2.

- [ ] **Convergence validation**: Verify training convergence matches non-EP baselines across latest MoE model architectures
- [ ] **Model coverage**: Add support for additional MoE architectures (e.g., Qwen-MoE)
- [ ] **ZeRO-3 support**: Extend AutoEP to work with ZeRO Stage 3
- [ ] **AutoTP integration**: Combine AutoEP with AutoTP for hybrid expert/tensor parallelism
- [ ] **Benchmarking**: Publish throughput, memory, and scaling efficiency numbers across model sizes and GPU counts
- [ ] **Universal Checkpoint support**: Enable saving and resuming from Universal Checkpoints with AutoEP

## AutoTP extension

AutoTP was significantly revamped in Q1 ([PR #7806](https://github.com/deepspeedai/DeepSpeed/pull/7806)), introducing a flexible, configuration-driven API for custom layer partitioning patterns. In Q2, we will extend this foundation to support a broader range of models and scales.

- [ ] **HuggingFace `tp_plan` support**: Leverage the `base_model_tp_plan` metadata provided by HuggingFace Transformers models to automatically derive partitioning configurations, enabling out-of-the-box TP for any model that ships with a tp_plan
- [ ] **Combination with AutoEP**: Support parallel folding for hybrid expert/tensor parallelism
[...]

PR #7901, opened 2026-03-13
Add HuggingFace tp_plan support for AutoTP

delock:
## Summary
                                                                                                                                       
Adds automatic detection and use of HuggingFace's built-in `base_model_tp_plan` for AutoTP, addressing the HuggingFace tp_plan support item from #7861.                                                                                                                       
                                                                                                                                       
Models that ship with a `tp_plan` (e.g. Llama, Qwen, Gemma2) now work with AutoTP out of the box — no `preset_model` or `partition_config` needed, just set `autotp_size`.                                                                                     
                                                                                                                                       
## Changes                                                                                                                             
                                                                                                                                       
**Runtime**                                                                                                                            
- `engine.py`: Added tp_plan fallback in `_apply_autotp_partitioning`. Priority order: `partition_config` > HF `tp_plan` > AutoTP heuristics.
- `config.py`: Added `_get_hf_tp_plan(model)` to extract tp_plan from `model._tp_plan` or `model.config.base_model_tp_plan`.
- `tp_plan_converter.py`: New file. `TPPlanConverter` converts HF tp_plan entries (`colwise`/`rowwise`) to DeepSpeed `TPLayerSpec`.
Other HF partition types (`colwise_rep`, `local_colwise`, etc.) are not yet supported (documented with TODO).

**Tests** (11 files, 17 CPU + 5 GPU tests)
- `test_tp_plan_converter.py`: Unit tests for the converter (alternate prefixes, projection names, unsupported types, etc.)
- `test_tp_plan_extraction.py`: Unit tests for `_get_hf_tp_plan` with mock models.
- `test_tp_plan_e2e.py`: GPU e2e tests with ZeRO 0/1/2 (requires 2 GPUs).
- `test_tp_plan_real_models.py`: GPU tests with Qwen2 and custom models (requires 2 GPUs).

**Documentation**
- Tutorial: New "HuggingFace tp_plan Support" section in `autotp-training.md`.
- Config reference: Added tp_plan paragraph in `config-json.md`.
- API docs: Added tp_plan subsection in `training.rst`.
- Blog: Updated ongoing work in `blogs/huggingface-tp/README.md`.

## Limitations

- Only `colwise` and `rowwise` partition types are supported. Extended types (`colwise_rep`, `local_colwise`, `local_rowwise`,
`local_packed_rowwise`, `gather`, `sequence_parallel`) are deferred.

### DIFF
--- a/deepspeed/checkpoint/constants.py
+++ b/deepspeed/checkpoint/constants.py
@@ -56,6 +56,8 @@ UNIVERSAL_CHECKPOINT_INFO = 'universal_checkpoint_info'
 UNIVERSAL_CHECKPOINT_VERSION_KEY = 'universal_checkpoint_version'
 # Reserve version 0.1  for the hardcoded logic used in BLOOM-176B training
 UNIVERSAL_CHECKPOINT_VERSION_VALUE = 0.2
+# Attribute name used to store AutoTP universal-checkpoint metadata on torch Parameters.
+DS_AUTOTP_UC_META = "ds_autotp_universal_checkpoint_meta"
 
 # Vocabulary padding
 VOCAB_TENSOR = 'vocab_tensor'
--- a/deepspeed/checkpoint/universal_checkpoint.py
+++ b/deepspeed/checkpoint/universal_checkpoint.py
@@ -9,7 +9,8 @@ import torch
 import types
 from typing import List, Tuple, Union
 from dataclasses import dataclass
-from .constants import (FP32_WEIGHT_KEY, PARAM, VOCAB_TENSOR, CAT_DIM, PARAM_N_SUB_PARAMS, SUB_PARAM_SHAPE)
+from .constants import (FP32_WEIGHT_KEY, PARAM, VOCAB_TENSOR, CAT_DIM, PARAM_N_SUB_PARAMS, SUB_PARAM_SHAPE,
+                        DS_AUTOTP_UC_META)
 
 
 @dataclass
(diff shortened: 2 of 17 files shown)


delock:
Hi @inkcherry @tohtana  @PKUWZP @tjruwase , this is the PR provide HuggingFace tp_plan support for AutoTP.   Hoping to see your comments, thanks!

tohtana:
Hi @delock,
This is an amazing enhancement!

One gap to handle is the mismatch between HuggingFace TP-plan styles and what AutoTP can consume directly. The new converter currently supports only `colwise` and `rowwise`, while HuggingFace TP plans use [additional styles](https://github.com/huggingface/transformers/blob/d64a6d67d8c004a25570db4df5689e06caea6af7/src/transformers/integrations/tensor_parallel.py#L1215) such as `colwise_gather_output` and `rowwise_split_input`.

AutoTP does not handle those HF styles directly, but some models are already supported through existing AutoTP specs/presets.  Phi-3 is a good example: its HF TP plan uses `colwise_gather_output` / `rowwise_split_input`, but DeepSpeed already [supports Phi-3](https://github.com/deepspeedai/DeepSpeed/blob/5f7b687018bd1e0340c661859820fd97aa80a616/deepspeed/module_inject/autotp_config.py#L511) through the existing AutoTP configuration with sub-parameter partitioning.

So I think the safer behavior would be:

1. Inspect the HF TP plan styles.
2. If all styles are supported by the converter, use the HF TP plan.
3. Otherwise, skip the HF TP plan path and fall back to the existing AutoTP path.

This would avoid regressing models like Phi-3 while still enabling the new HF-plan path for models whose TP plans map cleanly to AutoTP.

delock:
@tohtana Thanks for the comments, this is a good suggestion.  Let me add this bypass in the PR and test Phi-3 accordingly.

In the long run, I think it would help if AutoTP can consume these additional styles, which can be done in a seperate PR.

> Hi @delock, This is an amazing enhancement!
> 
> One gap to handle is the mismatch between HuggingFace TP-plan styles and what AutoTP can consume directly. The new converter currently supports only `colwise` and `rowwise`, while HuggingFace TP plans use [additional styles](https://github.com/huggingface/transformers/blob/d64a6d67d8c004a25570db4df5689e06caea6af7/src/transformers/integrations/tensor_parallel.py#L1215) such as `colwise_gather_output` and `rowwise_split_input`.
> 
> AutoTP does not handle those HF styles directly, but some models are already supported through existing AutoTP specs/presets. Phi-3 is a good example: its HF TP plan uses `colwise_gather_output` / `rowwise_split_input`, but DeepSpeed already [supports Phi-3](https://github.com/deepspeedai/DeepSpeed/blob/5f7b687018bd1e0340c661859820fd97aa80a616/deepspeed/module_inject/autotp_config.py#L511) through the existing AutoTP configuration with sub-parameter partitioning.
> 
> So I think the safer behavior would be:
> 
> 1. Inspect the HF TP plan styles.
> 2. If all styles are supported by the converter, use the HF TP plan.
> 3. Otherwise, skip the HF TP plan path and fall back to the existing AutoTP path.
> 
> This would avoid regressing models like Phi-3 while still enabling the new HF-plan path for models whose TP plans map cleanly to AutoTP.

delock:
Hi @tohtana fallback path for unsupported tp_plan layer specs are added.  Before the change, phi3 will report unsupported layer specs, after the change phi3 does not have such complain.

Note my test on Phi3-mini raise a seperate failure with shape mismatch, which is also raised on master.  I havn't looked into it yet.  It can be addressed in a seperated investigation and probably a fix.

> @tohtana Thanks for the comments, this is a good suggestion. Let me add this bypass in the PR and test Phi-3 accordingly.
> 
> In the long run, I think it would help if AutoTP can consume these additional styles, which can be done in a seperate PR.
> 
> > Hi @delock, This is an amazing enhancement!
> > One gap to handle is the mismatch between HuggingFace TP-plan styles and what AutoTP can consume directly. The new converter currently supports only `colwise` and `rowwise`, while HuggingFace TP plans use [additional styles](https://github.com/huggingface/transformers/blob/d64a6d67d8c004a25570db4df5689e06caea6af7/src/transformers/integrations/tensor_parallel.py#L1215) such as `colwise_gather_output` and `rowwise_split_input`.
> > AutoTP does not handle those HF styles directly, but some models are already supported through existing AutoTP specs/presets. Phi-3 is a good example: its HF TP plan uses `colwise_gather_output` / `rowwise_split_input`, but DeepSpeed already [supports Phi-3](https://github.com/deepspeedai/DeepSpeed/blob/5f7b687018bd1e0340c661859820fd97aa80a616/deepspeed/module_inject/autotp_config.py#L511) through the existing AutoTP configuration with sub-parameter partitioning.
> > So I think the safer behavior would be:
> > 
> > 1. Inspect the HF TP plan styles.
> > 2. If all styles are supported by the converter, use the HF TP plan.
> > 3. Otherwise, skip the HF TP plan path and fall back to the existing AutoTP path.
> > 
> > This would avoid regressing models like Phi-3 while still enabling the new HF-plan path for models whose TP plans map cleanly to AutoTP.

delock:
Hi @sfc-gh-truwase @tohtana the comments had been addressed, can you resume the merge process?  Thanks!

PR #7908, opened 2026-03-16
Add Feature Universal Checkpoint for AutoTP

nathon-lee:
Hi DeepSpeed team — thanks for your time reviewing this PR.

## Summary
Add Universal Checkpoint (UC) metadata support for DeepSpeed AutoTP to enable saving and resuming from Universal Checkpoints.

## Motivation
AutoTP partitions parameters across TP ranks. To make checkpoints portable and restorable, we need a stable UC metadata representation that can be collected at save time and consumed at restore time.

## What’s in this PR
- Collect AutoTP-specific Universal Checkpoint metadata for TP-partitioned parameters.
- Provide restore/merge helpers that normalize shapes and correctly interpret the saved conversion/partition view.
- Keep existing (non-AutoTP / non-UC) checkpoint paths unchanged (no behavior change expected for other users).

## Testing
- `pytest -q tests/unit/runtime/tensor_parallel/test_autotp_universal_checkpoint.py`
- `pytest -q tests/unit/checkpoint/test_autotp_universal_checkpoint.py`

## Request for feedback
Could you please take a look at the UC metadata schema and let me know if you’d prefer any changes to naming, field placement, or compatibility expectations? I’m happy to iterate quickly based on your guidance.

## References
- Refs: #7861 (Q2 2026 roadmap — AutoTP Universal Checkpoint support)

### DIFF
--- a/deepspeed/ops/transformer/inference/triton/matmul_ext.py
+++ b/deepspeed/ops/transformer/inference/triton/matmul_ext.py
@@ -53,7 +53,8 @@ class TritonCacheDir:
 
     @staticmethod
     def default_cache_dir():
-        tmp_path = os.path.join(Path.home(), ".triton", "autotune")
+        tt_home = os.environ.get('TRITON_HOME') or os.path.join(Path.home(), ".triton")
+        tmp_path = os.path.join(tt_home, "autotune")
         return tmp_path
 
 


deepspeed/checkpoint/universal_checkpoint.py
```diff
@@ -12,13 +12,73 @@
 from .constants import (FP32_WEIGHT_KEY, PARAM, VOCAB_TENSOR, CAT_DIM, PARAM_N_SUB_PARAMS, SUB_PARAM_SHAPE)
 
 
+AUTOTP_UC_META_KEY = 'ds_autotp_universal_checkpoint_meta'
```
delock: This is not the only place the reference the string `'ds_autotp_universal_checkpoint_meta'`.  The other place is in layers.py with a different name `DS_AUTOTP_UC_META`.  The name is not consistent.  (`DS_AUTOTP_UC_META` looks better).  If the string needs to be the same all over the place, should have a single definition (i.e. in this file) then import from other places (i.e. `from deepspeed.checkpoint.universal_checkpoint.py import DS_AUTOTP_UC_META`).  There shouldn't be a second place with the content of the string, including tests.

deepspeed/checkpoint/universal_checkpoint.py
```diff
@@ -12,13 +12,73 @@
...
+
+    AutoTP parameter metadata intentionally serves two separate consumers:
+    - restore-time fields at the top level, consumed here by UC loading
+    - conversion-time fields under `conversion`, consumed by
+      `collect_autotp_universal_checkpoint_info()` in `layers.py`
+    """
+    return getattr(param, AUTOTP_UC_META_KEY, None)
+
+
+def _resolve_autotp_partition(self, ckpt_dict, full_hp_param, tp_rank, tp_world_size):
```
delock: The parameter name `self` is not a proper name for a none class member function (This is also true for `load_hp_checkpoint_state` which is already there).  If `self` mean current parameter, can we changes its name to `current_param` or a proper name?  Thanks!

deepspeed/checkpoint/universal_checkpoint.py
```diff
@@ -12,13 +12,73 @@
...
+    partition_dim = meta.get('partition_dim')
+    logical_shape = meta.get('logical_shape')
+    output_shape = meta.get('output_shape')
+    sub_param_shape = meta.get('sub_param_shape')
+    sub_param_sizes = meta.get('sub_param_sizes')
+    target_partition_shape = meta.get('target_partition_shape')
+    is_bias = meta.get('is_bias', False)
+    replicated = meta.get('replicated', False)
+
+    if replicated or partition_dim is None:
```
delock: Is there a scenario that one of them is `True` and another one is `False`?  Probably better to have the following instead:
```
if replicated:
    assert(partition_dim is None)
    ...
```

deepspeed/checkpoint/universal_checkpoint.py
```diff
@@ -12,13 +12,73 @@
...
+        slice_tensor = full_hp_param
+    else:
+        target_shape = output_shape if is_bias else logical_shape
```
delock: This line implies that `output_shape` would be different from `logical_shape` when is_bias is `True`.  Will this really happen?

deepspeed/checkpoint/universal_checkpoint.py
```diff
@@ -12,13 +12,73 @@
...
+        if target_shape is None:
+            return None
+
+        full_view = full_hp_param.view(target_shape)
```
delock: `full_view` is not used in this `if` statement, better move it down below before the next `if` statement.

deepspeed/checkpoint/universal_checkpoint.py
```diff
@@ -12,13 +12,73 @@
...
+        if sub_param_sizes is None and sub_param_shape is not None:
```
delock: Is there a case that `sub_param_sizes` cannot be deduced from `sub_param_shape`, and must save `sub_param_size` but not `sub_param_shape` when saving the meta data?  I feel one of them is redudant.

deepspeed/module_inject/layers.py
```diff
@@ -324,6 +419,86 @@ def configure_tensor_parallel_runtime(config):
...
+
+    row_parallel_patterns = []
+    replicated_patterns = []
+    vocabulary_patterns = []
+    parameter_with_sub_params = []
+    original_vocab_size = None
+
+    for module_name, module in model.named_modules():
+        marker = getattr(module, '_mark_uc_metadata', None)
+        if callable(marker):
```
delock: Is it necessary to check whether `marker` is callable?  I didn't see this possibility unless its an implementation error.

tests/unit/checkpoint/test_autotp_uc_checkpoint.py
```diff
@@ -0,0 +1,241 @@
+import types
```
inkcherry, line 6: Thanks for the PR! The implementation looks good overall. Could you please add some documentation and usage descriptions?

nathon-lee:
Hi Delock, thanks for the suggestion. I’ve centralized the AutoTP universal-checkpoint metadata attribute name into deepspeed/checkpoint/constants.py as DS_AUTOTP_UC_META, and updated both universal_checkpoint.py and module_inject/layers.py to import and use it (so we no longer hardcode the string in multiple places).

Please let me know if you’d prefer I split the other _resolve_autotp_partition changes into a separate PR to keep this one focused.
@delock

tests/unit/checkpoint/test_autotp_uc_checkpoint.py
nathon-lee: Thanks @inkcherry! Sure — I’ll add documentation and usage descriptions and update the PR shortly.

deepspeed/module_inject/layers.py
nathon-lee: Good point. _mark_uc_metadata is intended to be a callable hook; using callable() could hide an implementation error. I’ll remove the callable() guard and simply check for None so incorrect types fail fast.

tests/unit/checkpoint/test_autotp_universal_checkpoint.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def _write_tp_slice(base_dir, param_name, tp_idx, state_name, tensor):
+    shard_dir = base_dir / param_name / str(tp_idx)
+    shard_dir.mkdir(parents=True, exist_ok=True)
+    torch.save(tensor.reshape(-1), shard_dir / f"{state_name}.00")
+
+
+def _write_tp_states(base_dir, param_name, tp_idx, fp32_tensor):
+    # merge_tp_slices 会尝试合并这三个 state，所以测试必须把它们都写出来
```
delock: This might not matter but I would suggest to keep comments in plain English for consistency.

tests/unit/runtime/tensor_parallel/test_autotp_universal_checkpoint.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def test_subparam_layer_marks_standardized_param_metadata():
+    layer = SubParamLinearLayer(torch.nn.Linear(12, 12, bias=True),
+                                mp_group=None,
+                                shape=(3, -1),
+                                partition_dim=0,
+                                name="packed")
+
+    weight_meta = getattr(layer.weight, "ds_autotp_universal_checkpoint_meta")
```
delock: I would suggest to keep these string imported rather than hard coded.

tests/unit/checkpoint/test_autotp_uc_checkpoint.py
```diff
@@ -0,0 +1,241 @@
...
+        'replicated': False,
+    })
+    full_hp_param = torch.arange(24, dtype=torch.float32).view(6, 4)
+
+    slice_flat = _resolve_autotp_partition(param, {PARAM: full_hp_param}, full_hp_param, tp_rank=0, tp_world_size=2)
+
+    chunks = [sub.chunk(2, dim=0)[0] for sub in full_hp_param.view(3, 2, 4)]
+    expected = torch.cat(chunks, dim=0).flatten()
+    assert torch.equal(slice_flat, expected)
+
```
delock, line 90: Is it possible to add a test when size of each subparam are not equal to each other?   This could cover the case of GQA
```
  sub_param_sizes 描述的是多个子参数在拼接维度上各自的大小不等的情况。

  典型场景是 QKV fused linear，比如一个模型的 Q/K/V head 数量不同：
  Q: 32 heads × 128 dim = 4096
  K: 8 heads × 128 dim = 1024
  V: 8 heads × 128 dim = 1024
  拼在一起后 sub_param_shape = (4096, 1024, 1024)，总共 6144。

  做 TP 切分时，如果简单地把 6144 均分成 N 份，会切到 Q/K/V 的边界中间。正确做法是按每个子参数各自均分：
  TP=2 时:
    Q: 4096/2 = 2048
    K: 1024/2 = 512
    V: 1024/2 = 512
    每个 rank 拿到 3072

  sub_param_sizes 就是告诉 restore 逻辑每个子参数有多大，这样才能按子参数分别切而不是整体均分。

  在 _resolve_autotp_partition 中，当 sub_param_sizes 不为 None 时，代码会对每个子参数单独做 chunk/split，然后按 tp_rank 取对应的片段，再 concat 回来。而当 sub_param_sizes 为 None
  时（所有子参数等大，或者根本没有子参数），直接整体 chunk 就行。

  所以测试中缺的 case 就是：子参数大小不等时，restore 是否正确按各自大小切分。现有测试 2 的 sub_param_shape = ((2,2,2), 4) 三个子参数等大（都是 2），没有覆盖不等大的情况。
```

delock:
Hi @nathon-lee , thanks for the timely response!  No need to split.

I have given my comments.  Overall it LGTM.   I agree with @inkcherry that documentation better be updated, specifically, `universal-checkpointing.md`, `autotp-training.md`, and `model-checkpointing.rst`.
> Hi Delock, thanks for the suggestion. I’ve centralized the AutoTP universal-checkpoint metadata attribute name into deepspeed/checkpoint/constants.py as DS_AUTOTP_UC_META, and updated both universal_checkpoint.py and module_inject/layers.py to import and use it (so we no longer hardcode the string in multiple places).
> 
> Please let me know if you’d prefer I split the other _resolve_autotp_partition changes into a separate PR to keep this one focused. @delock

nathon-lee:
Hi @delock, thank you for the review and LGTM!
﻿
Totally agree — I’ll update universal-checkpointing.md, autotp-training.md, and model-checkpointing.rst accordingly and push the doc changes shortly.
@delock

nathon-lee:
> LGTM

Thank you, Delock, but there was a unit test that failed. I just fixed it and need to run it again

nathon-lee:
It's a bit awkward. Pre commit run -- all files can detect formatting errors, but they can't help me modify them. I used the pre commit run -- all files command to correct the formatting issue again. It would be great if pre commit run -- all files could automatically correct it afterwards

delock:
> It's a bit awkward. Pre commit run -- all files can detect formatting errors, but they can't help me modify them. I used the pre commit run -- all files command to correct the formatting issue again. It would be great if pre commit run -- all files could automatically correct it afterwards

My way of fixing formatting is add a tailing space by the end of a random line in the files I touched in the PR, then turn on pre-commit (pre-commit install) to commit it.  Pre-commit hook will fix this tailing space along with other formatting errors.  Not elegant but it works.

PR #7916, opened 2026-03-21
fix: add setup_context for torch.func compatibility

### LINKED ISSUE #7913: [BUG] LinearFunctionForZeroStage3 crashes with torch.func transforms (missing setup_context)
**Describe the bug**

`LinearFunctionForZeroStage3` in `deepspeed/runtime/zero/linear.py` uses the legacy `autograd.Function` pattern where `forward(ctx, ...)` directly calls `ctx.save_for_backward()`. This crashes when any code path uses `torch.func` transforms (`torch.func.grad`, `torch.func.grad_and_value`, `vmap`, etc.) on a model running with ZeRO Stage 3.
```
RuntimeError: In order to use an autograd.Function with functorch transforms
(vmap, grad, jvp, jacrev, ...), it must override the setup_context staticmethod.
```

This affects any library that uses `torch.func` internally, including Liger-Kernel (fused cross-entropy) and Axolotl's KD (Knowledge Distillation) kernel.

**To Reproduce**

1. Enable ZeRO Stage 3
2. Use any operation that internally calls `torch.func.grad_and_value` during forward/loss computation (e.g. Axolotl offline KD with Liger kernel, or any custom loss using functorch)
3. Training crashes at the first forward pass

Minimal example:
```python
import torch
import deepspeed

model = ...  # any model
model, _, _, _ = deepspeed.initialize(model=model, config={"zero_optimization": {"stage": 3}, ...})

# Any torch.func usage on a ZeRO-3 wrapped model triggers the crash
# because F.linear is replaced by LinearFunctionForZeroStage3
# which lacks setup_context
torch.func.grad_and_value(some_loss_fn)(params, inputs)
```

**Expected behavior**

`LinearFunctionForZeroStage3` should be compatible with `torch.func` transforms. The `ctx` usage in `forward` is minimal (only `save_for_backward`), so splitting into the `setup_context` pattern is straightforward:
```python
# Current (crashes with torch.func)
class LinearFunctionForZeroStage3(torch.autograd.Function):
    @staticmethod
    def forward(ctx, input, weight, bias=None):
        ctx.save_for_backward(input, weight, bias)
        ...

# Fix: separate setup_context (compatible with torch.func)
class LinearFunctionForZeroStage3(torch.autograd.Function):
    @staticmethod
    def forward(input, weight, bias=None):
        ...
        return ret

    @staticmethod
    def setup_context(ctx, inputs, output):
        input, weight, bias = inputs
        ctx.save_for_backward(input, weight, bias)
```

`backward` needs no changes. Note: `setup_context` requires PyTorch >= 2.0, so a version-conditional definition may be needed if PyTorch < 2.0 support is still required.

**ds_report output**
```
NOTE: Alarm op not installed
NOTE: AsyncIO op not installed
...
torch install path: /usr/local/lib/python3.11/dist-packages/torch
torch version: 2.8.0+cu128
deepspeed install path: /usr/local/lib/python3.11/dist-packages/deepspeed
deepspeed info: 0.16.4, unknown, unknown
torch cuda version: 12.8
torch hip version: None
nvcc version: Not Available
deepspeed wheel compiled w.: torch 2.8, cuda 12.8
```

**System info:**
 - OS: Ubuntu 22.04
 - GPU: H100 80GB PCIe
 - Python: 3.11
 - PyTorch: 2.8.0+cu128

**Launcher context**

deepspeed CLI launcher

**Additional context**

The root cause is that PyTorch's `torch.func` transforms require `autograd.Function` subclasses to use the separate `forward()` + `setup_context()` pattern instead of the combined `forward(ctx, ...)` pattern. This is documented in [Extending torch.func with autograd.Function](https://docs.pytorch.org/docs/stable/notes/extending.func.html).

The current `forward` only uses `ctx` for a single `save_for_backward` call, making the migration minimal. The `backward` method uses `ctx.saved_tensors` and `ctx.needs_input_grad`, both of which work identically with the new pattern.

One consideration: `setup_context` was introduced in PyTorch 2.0. DeepSpeed's documented minimum is PyTorch 1.9+. if PyTorch < 2.0 is no longer actively supported, the migration can be done unconditionally.

Happy to submit a PR if this direction is acceptable.

roycho96:
`LinearFunctionForZeroStage3` uses the legacy `forward(ctx, ...)` pattern which is incompatible with `torch.func` transforms (`torch.func.grad`, `torch.func.grad_and_value`, `vmap`, etc.):
```
RuntimeError: In order to use an autograd.Function with functorch transforms
(vmap, grad, jvp, jacrev, ...), it must override the setup_context staticmethod.
```

This affects any library that uses `torch.func` internally on a ZeRO-3 model.

## Fix

Fixes #7913 

## Note

As pointed out by @zhangj1an in #7913, `PostBackwardFunctionModule` and `PreBackwardFunctionForModule` in `parameter_offload.py` have the same issue. Those will be addressed in a follow-up commit within this PR.


deepspeed/runtime/zero/linear.py
```diff
@@ -35,69 +35,122 @@ def print_rank_0(message, debug=False, force=False):
...
+                if bias is not None:
+                    output += bias
+                ret = output
+
+            return ret
+
+        @staticmethod
+        def setup_context(ctx, inputs, output):
+            input, weight, bias = inputs
+            ctx.save_for_backward(input, weight, bias)
```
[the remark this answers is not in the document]
zhangj1an: Following up on codex's commet, I suggest for Pytorch >=2.0, for `forward`, this line should be removed, `@autocast_custom_fwd`. So `forward` will look like this,
```
@staticmethod
# bias is an optional argument
def forward(input, weight, bias=None):
    if input.dim() == 2 and bias is not None:
        # fused op is marginally faster
        ret = torch.addmm(bias, input, weight.t())
    else:
        output = input.matmul(weight.t())
        if bias is not None:
            output += bias
        ret = output
    return ret
```

Reason being, according to line 35, `@autocast_custom_fwd` is alias of `torch.amp.autocast_mode.py:custom_fwd`. Its nested `decorate_fwd` method looks like this:

```
@functools.wraps(fwd)
def decorate_fwd(*args, **kwargs):
        args[0]._dtype = torch.get_autocast_dtype(device_type)
        if cast_inputs is None:
            args[0]._fwd_used_autocast = torch.is_autocast_enabled(device_type)
            return fwd(*args, **kwargs)
        else:
            ...
    return decorate_fwd
```

the `args[0]` is actually treated as `cxt` (e.g. `args[0]._dtype`, `args[0]._fwd_used_autocast`), which is no longer available after we remove the deprecated `cxt` from `forward` method.  We can re-implement these few lines in `setup_context` like below,

```
@staticmethod
def setup_context(ctx, inputs, output):
    device_type = get_accelerator().device_name()
    ctx._dtype = torch.get_autocast_dtype(device_type)
    ctx._fwd_used_autocast = torch.is_autocast_enabled(device_type)
    input, weight, bias = inputs[0], inputs[1], inputs[2] if len(inputs) > 2 else None
    ctx.save_for_backward(input, weight, bias)
```

deepspeed/runtime/zero/linear.py
```diff
@@ -35,76 +35,126 @@ def print_rank_0(message, debug=False, force=False):
 autocast_custom_fwd = functools.partial(torch.amp.custom_fwd, device_type=get_accelerator().device_name())
 autocast_custom_bwd = functools.partial(torch.amp.custom_bwd, device_type=get_accelerator().device_name())
 
+# PyTorch >= 2.0 supports setup_context, which is required for
+# torch.func transforms (vmap, grad, jvp, jacrev, etc.)
+_SUPPORTS_SETUP_CONTEXT = hasattr(torch.autograd.Function, 'setup_context')
```
zhangj1an: I tried to implement a version check too, but cannot find a way to avoid duplicated code. What do you think about adding a error log like below,

```
def _require_pt2_zero3_linear():
    if int(torch.__version__.split(".")[0]) < 2:
        raise RuntimeError("LinearFunctionForZeroStage3 requires PyTorch >= 2.0")
```

and use it in `LinearModuleForZeroStage3`'s `forward` method? 

For pytorch 1.9, we will just deprecate it by adding a new line in `README.md` in the requirements section,
```
For ZeRO Stage 3,  `PyTorch >= 2.0` is required, as parameter-offload hooks rely on `torch.autograd.Function.setup_context`.
```

zhangj1an (commented):
Thanks for the work! I implemented the same fix, so it looks good to me. To reduce reviewer's effort, I had 2 minor comments. This should lead to `linear.py` to only have 17 insertions (+) and 7 deletions (-).

zhangj1an:
Hey @roycho96, I used this script to reproduce your issue #7913. 
[repro_zero3_functorch.py](https://github.com/user-attachments/files/26160935/repro_zero3_functorch.py)

After I implement the fixes in `parameter_offload.py`, I am still not able to run this file normally, it returns segmentation fault. I will need more time to investigate what is happening. 

I am not authorised to push directly into your branch. In case I blocked your PR, here is my fix for `parameter_offload.py`. 
[parameter_offload.py](https://github.com/user-attachments/files/26160960/parameter_offload.py) Feel free to update it on your end!

roycho96:
> Thanks for the work! I implemented the same fix, so it looks good to me. To reduce reviewer's effort, I had 2 minor comments. This should lead to `linear.py` to only have 17 insertions (+) and 7 deletions (-).

Hi @zhangj1an, I've sent you a collaborator invite to my fork. Feel free to push your fix directly to the branch. Thanks for the suggestion!

deepspeed/runtime/zero/linear.py
roycho96: Good catch! You're right.

Also, I agree the version branch duplication isn't ideal. However, I think dropping PyTorch < 2.0 support is a policy decision that should come from the maintainers

zhangj1an:
Hey @roycho96 , thanks for the branch invite! I finished the fix required from my side.
<details> <summary> summary of fix </summary>

- https://github.com/deepspeedai/DeepSpeed/pull/7916/commits/5e83d056fc72b4909d44976cd7ad50da80a491fb: implemented the fix in `parameter_offload.py` as discussed.  
- [minor] https://github.com/deepspeedai/DeepSpeed/pull/7916/commits/6df37afb96961aeaf92a0870e674183d8e6aaf61: signed your commit https://github.com/deepspeedai/DeepSpeed/commit/444122cfaf871731f86c9f869f684f4918f1c712 to pass DCO check. 
- [minor] https://github.com/deepspeedai/DeepSpeed/pull/7916/commits/c0b9694cce1b1eaee9d2119ebcc57ab854b2bef0: in `LinearFunctionForZeroStage3`, removed `@autocast_custom_bwd` to be consistent with the removal of `@autocast_custom_fwd` in https://github.com/deepspeedai/DeepSpeed/commit/444122cfaf871731f86c9f869f684f4918f1c712

</details>

~~However, now I experience a different error. Assume when Zero-3 is enabled, a model that contains `F.linear` goes through ZeRO-3’s `forward` method in `LinearFunctionForZeroStage3`. When calling `LinearFunctionForZeroStage3.apply`, the lines including `weight.t()`, `torch.addmm`, and `matmul` will trigger segmentation fault in functorch. (GDB shows the exact place is at `at::functorch::TensorWrapper::is_alive()`)~~

~~Due to this error, I cannot yet find a verification script that runs on this branch and fails on main branch. I'm not sure how to go about to solve it. Do you have any idea? 🤔 Thanks in advance!~~

Update: found a verification script. My previous script did not go through `torch.run` properly and used dummy arguments, which caused segmentation fault error.

zhangj1an:
Hey @roycho96, I think this PR is ready for review! Below is my attempt for the updated PR description. Feel free to make any changes. 

## Summary

Fixes https://github.com/deepspeedai/DeepSpeed/issues/7913

ZeRO-3’s custom linear `autograd.Function` used the legacy `forward(ctx, …)` + `ctx.save_for_backward` pattern. On recent PyTorch (e.g. 2.8+), any `torch.func` transform (`grad`, `grad_and_value`, `vmap`, …) over that path raises:

```
RuntimeError: In order to use an autograd.Function with functorch transforms … it must override the setup_context staticmethod.
```

This breaks real stacks that combine ZeRO-3 with functorch-based code (e.g. some Liger-Kernel / Axolotl KD paths). More details in the issue above.

## Fix

- **`deepspeed/runtime/zero/linear.py`** : For `LinearFunctionForZeroStage3`, adopt the **`autograd.Function` pattern PyTorch expects under functorch: non-context `forward` + `setup_context`** (saved tensors moved into `setup_context`). **`backward` is unchanged.**
- **`deepspeed/runtime/zero/parameter_offload.py`** : Same change for ZeRO-3 offload `autograd.Function` definitions that used `ctx` inside the old-style `forward`.
- **Regression test:** `tests/unit/v1/zero/test_zero_functorch_linear.py` (skips if `torch.autograd.Function.setup_context` is missing, e.g. PyTorch < 2.0).

When `setup_context` is not available, the **legacy `forward(ctx, …)`** path is retained so older PyTorch builds are unaffected.

## How to verify

The bug can be reproduced on `main` branch (DeepSpeed `0.18.8+…`) with **PyTorch 2.8**.

Two **temporary** scripts are added to help reviewers compare this branch vs `master`.

| File | Role |
|------|------|
| `scripts/repro_pr7916.py` | Minimal repro: ZeRO-3 via `deepspeed.initialize`, then `torch.func.grad_and_value` is called via **`zero3_linear_wrap`** . |
| `scripts/setup_pr7916.sh` | Creates a `.venvs/pr7916` with **PyTorch 2.8 + cu128**, runs the repro on **this branch** then on **`master`**. |

### How to run

```bash
./scripts/setup_pr7916.sh
# Optional: wipe venv and reinstall from scratch
./scripts/setup_pr7916.sh --force-install
```

Above script will print result from both branches. Alternatively, if the venv is already set up, you can run

```bash
source .venvs/pr7916/bin/activate
torchrun --standalone --nproc_per_node=1 scripts/repro_pr7916.py
```

### Expected output

- **This branch:** completes; rank 0 prints  
  `repro: grad_and_value over zero3_linear_wrap (LinearFunctionForZeroStage3) OK.`
- **`master` (unfixed):** fails with the functorch error, `RuntimeError: In order to use an autograd.Function with functorch transforms … it must override the setup_context staticmethod.`

## Test plan

- [x] `pytest tests/unit/v1/zero/test_zero_functorch_linear.py`

## Cleanup

- [ ] Remove `scripts/repro_pr7916.py`, `scripts/setup_pr7916.sh`, after PR approval.

---

## Example environment where the bug was reproduced

| Item | Value |
|------|--------|
| OS | Ubuntu 22.04 |
| GPU | NVIDIA H100 80GB PCIe |
| Python | 3.11 |
| PyTorch | 2.8.0+cu128 |
| PyTorch CUDA (wheel) | 12.8 |
| DeepSpeed (original report) | 0.16.4 (wheel) |

Flink-ddd:
Hi @tohtana , would you mind reviewing this PR when you're free? It addresses a useful compatibility fix for torch.func. Much appreciated!

deepspeed/runtime/zero/linear.py
sfc-gh-truwase: @roycho96, @zhangj1an please feel free to drop support for pytorch < 2.0

roycho96:
I additionally fix autocast backward: always wrap with autocast(enabled=ctx._fwd_used_autocast) to match `@custom_bwd` semantics and prevent outer autocast leaking into backward.

deepspeed/runtime/zero/linear.py
roycho96: @sfc-gh-truwase Thanks! I dropped support for pytorch <2.0.

scripts/repro_pr7916.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
```
tohtana: Thank you for creating the repro and setup scripts! I think this PR is almost ready. Can you now remove the repro scripts (repro_pr7916.py and setup_pr7916.sh)?

deepspeed/runtime/zero/linear.py
```diff
@@ -33,38 +31,11 @@ def print_rank_0(message, debug=False, force=False):
         print(message)
```
tohtana, line 35: I merged my PR #7920 earlier and it caused conflicts.
I submitted a PR to resolve the conflict: https://github.com/roycho96/DeepSpeed/pull/1
Can you merge it?

tohtana (commented):
Thank you @roycho96, @zhangj1an!

This PR overall looks good to me. The issue is I merged my PR #7920 and it caused a conflict as I hadn't checked the changes by this PR. Sorry for that.
I opened a [new PR](https://github.com/roycho96/DeepSpeed/pull/1) on your fork to resolve the conflict. Please check and merge it.

deepspeed/runtime/zero/linear.py
zhangj1an: Thanks for the help! Done in https://github.com/deepspeedai/DeepSpeed/pull/7916/commits/8468149b1f507e998985c966dd8d262181c40925

scripts/repro_pr7916.py
zhangj1an: Thanks! Done in https://github.com/deepspeedai/DeepSpeed/pull/7916/commits/e309a6f00f6e8120cd8348c01689e8d2a4508a80. I tested the latest status of this PR in https://github.com/deepspeedai/DeepSpeed/pull/7916/commits/39f7e3ce1c2bbaaef142b526fb1d3cd1ccb43626, the fix still works.

deepspeed/runtime/zero/linear.py
zhangj1an: Hey @roycho96, do you mind mark this as resolved? I cannot do it from my side. Thanks!

deepspeed/runtime/zero/linear.py
```diff
@@ -35,76 +35,141 @@ def print_rank_0(message, debug=False, force=False):
...
+        # This function has only a single output, so it gets only one gradient
+        @staticmethod
+        def backward(ctx, grad_output):
+            # Do not use @autocast_custom_bwd here: it pairs with @autocast_custom_fwd on
+            # legacy forward(ctx, ...). With forward + setup_context, use AMP state from setup_context.
+            device_type = get_accelerator().device_name()
+            if getattr(ctx, "_fwd_used_autocast", False):
+                with torch.amp.autocast(device_type=device_type, enabled=True, dtype=ctx._dtype):
+                    return LinearFunctionForZeroStage3._backward_core(ctx, grad_output)
+            return LinearFunctionForZeroStage3._backward_core(ctx, grad_output)
```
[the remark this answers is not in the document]
zhangj1an: Done in https://github.com/deepspeedai/DeepSpeed/pull/7916/commits/bb245b2ecc089d1ccdb2373b10bedd6de3b5d9b2

deepspeed/runtime/zero/linear.py
roycho96: @zhangj1an Already marked as resolved.

roycho96:
@tohtana
Thank you for the review! Your PR https://github.com/deepspeedai/DeepSpeed/pull/7920 is merged!
Seems like all the conflicts are solved.

roycho96:
@tohtana
Gentle ping, happy to address anything if needed, thanks!

tohtana:
Hi @roycho96,
Sorry I have missed the update for a long time, and thank you for pinging me.
I'm happy to merge this PR now, but it seems the current update doesn't cover `vmap`. Do you want to update this PR before we merge or merge it now leaving it as follow-up work.

roycho96:
> Hi @roycho96, Sorry I have missed the update for a long time, and thank you for pinging me. I'm happy to merge this PR now, but it seems the current update doesn't cover `vmap`. Do you want to update this PR before we merge or merge it now leaving it as follow-up work.

Thanks @tohtana! Sounds good to merge this one and I'll send a follow-up for vmap. It'll need generate_vmap_rule on LinearFunctionForZeroStage3 and the offload PreBackwardFunctionForModule / PostBackwardFunctionModule.

tohtana (approved):
Okay let's leave it for now and merge this PR now.

PR #7917, opened 2026-03-21
Add torch_xla TPU support for ZeRO-1/2

PKUWZP:
## Summary
- add an XLA accelerator for TPU execution through torch_xla
- initialize DeepSpeed distributed with the torch.distributed xla backend and xla:// init flow
- make ZeRO-1/2 gradient bucketing backend-agnostic and add TPU/XLA-focused tests and docs

## Testing
- git diff --check
- python3 -m py_compile accelerator/xla_accelerator.py accelerator/real_accelerator.py deepspeed/comm/constants.py deepspeed/comm/comm.py deepspeed/comm/torch.py deepspeed/runtime/zero/stage_1_and_2.py tests/accelerator/test_ds_init.py tests/unit/common.py tests/unit/accelerator/test_accelerator.py tests/unit/comm/test_xla_backend.py

## Notes
- local python environment in this session did not have pytest or pre-commit installed, so those checks could not be executed here
- scope is intentionally limited to TPU support for ZeRO-1/2 via torch_xla

### DIFF
--- a/accelerator/real_accelerator.py
+++ b/accelerator/real_accelerator.py
@@ -137,7 +137,7 @@ def get_accelerator():
 
                 if len(xm.get_xla_supported_devices(devkind='TPU')) > 0:
                     accelerator_name = "xla"
-            except ImportError as e:
+            except (ImportError, RuntimeError):
                 pass
         if accelerator_name is None:
             try:
--- a/accelerator/xla_accelerator.py
+++ b/accelerator/xla_accelerator.py
@@ -48,6 +48,10 @@ class XLA_Accelerator(DeepSpeedAccelerator):
             raise RuntimeError("No addressable XLA devices are available in the current process.")
         if device_index is None:
             return 0
+        if isinstance(device_index, torch.device):
+            device_index = device_index.index if device_index.index is not None else 0
+        elif isinstance(device_index, str):
+            device_index = int(device_index) if device_index.isdigit() else int(device_index.split(':')[-1])
         return min(device_index, len(devices) - 1)
 
     def is_synchronized_device(self):


accelerator/xla_accelerator.py
```diff
@@ -0,0 +1,265 @@
...
+        return True
+
+    def handles_memory_backpressure(self):
+        return True
+
+    # Device APIs
+    def device_name(self, device_index=None):
+        if device_index is None:
+            return 'xla'
+        return f'xla:{device_index}'
```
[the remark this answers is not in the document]
PKUWZP: Addressed in 7f82c20d. `device_name(device_index)` now resolves through the process addressable XLA device list instead of treating `LOCAL_RANK` as a raw XLA ordinal, so single-device-per-process workers map nonzero local ranks back to their local `xla:0` device correctly.

accelerator/xla_accelerator.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def device(self, device_index=None):
+        xm_module = self._require_xm()
+        return xm_module.xla_device(n=device_index, devkind='TPU')
+
+    def set_device(self, device_index):
+        os.environ['LOCAL_RANK'] = str(device_index)
```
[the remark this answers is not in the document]
PKUWZP: Addressed in 7f82c20d. `set_device()` now calls into `xm.xla_device()` to select the XLA default device for the current process before DeepSpeed moves the model, and it preserves the launcher-provided rank information in the environment.

accelerator/xla_accelerator.py
```diff
@@ -0,0 +1,290 @@
...
+        xm_module = self._require_xm()
+        return [torch.device(device) for device in xm_module.get_xla_supported_devices(devkind='TPU')]
+
+    def _normalize_device_index(self, device_index=None):
+        devices = self._addressable_devices()
+        if not devices:
+            raise RuntimeError("No addressable XLA devices are available in the current process.")
+        if device_index is None:
+            return 0
+        return min(device_index, len(devices) - 1)
```
tohtana, line 55: This assumes `device_index` is int, but can actually be
- string (`device_name`: [code](https://github.com/deepspeedai/DeepSpeed/blob/38bd11a758875cc48b9def31c74464140658ce43/deepspeed/runtime/zero/utils.py#L95-L96))
- `torch.device` (`set_device`: [code](https://github.com/deepspeedai/DeepSpeed/blob/38bd11a758875cc48b9def31c74464140658ce43/deepspeed/runtime/zero/partition_parameters.py#L1060))

For these, the function should throw `TypeError`. I suggest handling these types in this function.

accelerator/real_accelerator.py
```diff
@@ -125,6 +131,14 @@ def get_accelerator():
                     accelerator_name = "xpu"
         except ImportError as e:
             pass
+        if accelerator_name is None:
+            try:
+                import torch_xla.core.xla_model as xm
+
+                if len(xm.get_xla_supported_devices(devkind='TPU')) > 0:
+                    accelerator_name = "xla"
+            except ImportError as e:
```
tohtana: `xm.get_xla_supported_devices()` can raise `RuntimeError` when `torch_xla` is installed but no TPU/PJRT runtime is available.
Can we catch it to fall back to the default path?

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -59,14 +59,11 @@ def input(msg):
...
-        "torch.{}.HalfTensor".format(device_type), "torch.{}.FloatTensor".format(device_type),
-        "torch.{}.DoubleTensor".format(device_type), "torch.{}.BFloat16Tensor".format(device_type)
-    ]
     buckets = []
-    for i, dtype in enumerate(dtypes):
-        bucket = [t for t in tensors if t.type() == dtype]
+    accelerator = get_accelerator()
+    dtype_order = (torch.float16, torch.float32, torch.float64, torch.bfloat16)
+    for dtype in dtype_order:
+        bucket = [tensor for tensor in tensors if tensor.dtype == dtype and accelerator.on_accelerator(tensor)]
```
sfc-gh-truwase: Why does `xla` require special handling given the prior code worked for other accelerators?

sfc-gh-truwase: Another issue is that it seems `on_accelerator` is only defined for `xla` and so other accelerators will break here.

deepspeed/comm/comm.py
```diff
@@ -821,6 +823,8 @@ def init_distributed(dist_backend=None,
         utils.logger.info(f'cdb={cdb}')
     if cdb is None and torch.distributed.is_initialized():
         # The user initialized torch.dist themselves, create cdb and short-circuit
+        if dist_backend is None:
```
sfc-gh-truwase, line 827: Why do we need this behavior? Is it specific to `xla`?

tests/unit/accelerator/test_accelerator.py
```diff
@@ -57,3 +59,104 @@ def test_abstract_methods_defined(module_name, accel_class_name):
...
+
+        def __str__(self):
+            return f"xla:{self.index}"
+
+    def xla_device(n=None, devkind=None):
+        if n is not None:
+            selected_device["index"] = n
+        return FakeDevice(selected_device["index"])
+
+    torch_xla.devices = lambda: [FakeDevice(idx) for idx in range(device_count)]
```
sfc-gh-truwase: I don't think using a fake TPU device is useful since computation cannot be tested. Rather, I think we should condition these tests to only run when TPU is available. Another possibility is to setup CI tests specifically for TPU on the cloud credits.

deepspeed/runtime/zero/stage_1_and_2.py
PKUWZP: @sfc-gh-truwase Thanks for the comments:) I removed the `accelerator.on_accelerator(tensor)` filter that would have changed behavior for all backends. I also kept the dtype-based comparison (replacing the old string-based type names like `torch.cuda.HalfTensor`) since that's the actual fix needed for XLA compatibility without breaking other accelerators.

deepspeed/comm/comm.py
PKUWZP: @sfc-gh-truwase Thanks for the call out! I added a comment clarifying that it's a general fix (not XLA-specific) — it prevents passing `None` to `TorchBackend` when the user pre-initialized `torch.distributed` without specifying `dist_backend`.

tests/unit/accelerator/test_accelerator.py
PKUWZP: @sfc-gh-truwase Great suggestions. I initially thought adding the fake TPU device is useful for testing, now I removed all three fake TPU device tests and the `_install_fake_torch_xla` helper. I agree with you that these should be conditioned on real TPU availability instead of using faked devices.

accelerator/xla_accelerator.py
PKUWZP: @tohtana good catch! I made some changes, it now handles `str` (e.g., "0" or "xla:0" from `os.environ["LOCAL_RANK"]`) and `torch.device` (`from torch.device(get_accelerator().device_name(...))`) in addition to int.

accelerator/real_accelerator.py
PKUWZP: @tohtana Thanks for the comments. I go ahead and changed `except ImportError` to `except (ImportError, RuntimeError)` so that when `torch_xla` is installed but no TPU/PJRT runtime is available, detection smoothly falls through to the next accelerator.

accelerator/xla_accelerator.py
tohtana: Thank you, I see you added handling in `_normalize_device_index`. But I wonder if `set_device` still has an issue with the path from `partition_parameters.py`.

In `set_device`, we write device_index back to the env vars ([code](https://github.com/deepspeedai/DeepSpeed/blob/65cf60fdae87dcac7fe78bdeef0da91eb1db4860/accelerator/xla_accelerator.py#L80])). If the `device_index` is `torch.device`, `LOCAL_RANK` and `PJRT_LOCAL_PROCESS_RANK` will include device type (e.g. `xla:0`), not only a number.

PR #7919, opened 2026-03-23
Extending Muon Optimizer Support for ZeRO Stage 3

PKUWZP:
Authors: @pengdurice and @PKUWZP

Create a separate PR based on #7798 with the same functional diff on a clean signed-off branch to resolve DCO issues.

We aim on adding Muon Optimizer to zero stage 3 in this draft PR:

- The new momentum buffers are also partitioned like `self.fp32_partitioned_groups_flat` to save memory footprint. So, before the muon update, we need to perform `all_gather` on top of each data-parallel group rank. The Muon updates of the parameters are also divided across the data-parallel ranks, and the results are all-gathered once all updates are complete. After the `all_gather`, the momentum buffers are partitioned and flattened again.

Next steps:

### DIFF
--- a/docs/_pages/config-json.md
+++ b/docs/_pages/config-json.md
@@ -752,6 +752,8 @@ Configuring the asynchronous I/O module for offloading parameter and optimizer s
 
 ### Tensor Parallel (AutoTP)
 Configure AutoTP tensor parallelism for training via the DeepSpeed config and hybrid TP + ZeRO. AutoTP supports ZeRO stages 0, 1, and 2 (stage 3 is not supported). `deepspeed.tp_model_init()` remains supported for backward compatibility but is not required when `tensor_parallel` is set in the config.
+
+When a HuggingFace model provides a built-in `tp_plan` (via `model.config.base_model_tp_plan`), DeepSpeed automatically detects and uses it. In this case, neither `preset_model` nor `partition_config` is required -- just set `autotp_size`. If `partition_config` is also provided, it takes precedence over the model's `tp_plan`.
 ```json
   "tensor_parallel": {
     "autotp_size": 4,
--- a/blogs/huggingface-tp/README.md
+++ b/blogs/huggingface-tp/README.md
@@ -232,6 +232,7 @@ Furthermore, if users are not using transformers library, you can replace the ``
 - **Optimization**: Communication/Activation optimization.
 - **Usability**: Support the [Transformers TP plan](https://github.com/huggingface/transformers/blob/336dc69d63d56f232a183a3e7f52790429b871ef/src/transformers/models/llama/configuration_llama.py#L145), decouple the AutoTP parser, and expand model testing.
   - [UPDATE] We now support [custom partitioning](https://deepspeed.readthedocs.io/en/latest/training.html#custom-layer-specs) in the same spirit as HF's partitioning plan, and will build Transformers TP plan support on top of that ([PR](http://github.com/deepspeedai/DeepSpeed/pull/7806)).
+  - [UPDATE] DeepSpeed now automatically detects and uses HuggingFace's built-in `base_model_tp_plan` (e.g. Llama, Qwen, Gemma2). When a model provides a `tp_plan`, AutoTP uses it directly without requiring `preset_model` or `partition_config`. Currently `colwise` and `rowwise` partition types are supported. See the [AutoTP training tutorial](https://deepspeed.readthedocs.io/en/latest/training.html#huggingface-tp-plan) for details ([PR](https://github.com/deepspeedai/DeepSpeed/pull/7901)).
 
 Theoretically, features supported by ZeRO should also be supported, though extensive testing is pending.
 Welcome bug reports, enhancement, and additional model training examples.
--- a/deepspeed/checkpoint/constants.py
+++ b/deepspeed/checkpoint/constants.py
@@ -56,6 +56,8 @@ UNIVERSAL_CHECKPOINT_INFO = 'universal_checkpoint_info'
 UNIVERSAL_CHECKPOINT_VERSION_KEY = 'universal_checkpoint_version'
 # Reserve version 0.1  for the hardcoded logic used in BLOOM-176B training
 UNIVERSAL_CHECKPOINT_VERSION_VALUE = 0.2
+# Attribute name used to store AutoTP universal-checkpoint metadata on torch Parameters.
+DS_AUTOTP_UC_META = "ds_autotp_universal_checkpoint_meta"
 
 # Vocabulary padding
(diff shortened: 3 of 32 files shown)


tests/unit/ops/muon/test_muon.py
```diff
@@ -42,8 +43,16 @@ def test(self, optimizer_type, zero_stage, lr, hidden_dim, nlayer):
             },
             "zero_optimization": {
                 "stage": zero_stage,
-            }
+                "reduce_scatter": False,
+            },
+            "save_muon_momentum_buffer_in_memory": True,
```
tohtana: We should test both True and False for save_muon_momentum_buffer_in_memory

deepspeed/runtime/zero/stage3.py
```diff
@@ -329,12 +331,15 @@ def _enforce_optimizer_offload():
         self.all2all_process_group = all2all_process_group
 
         self.reduce_scatter = reduce_scatter
-
+        self.use_muon = isinstance(self.optimizer, MuonWithAuxAdam)
+        self.save_muon_momentum_buffer_in_memory = ds_config.get('save_muon_momentum_buffer_in_memory', False)
```
tohtana: I think it is cleaner to have have the default setting in (one of) `config.py`.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1430,6 +1470,118 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+                    self._create_momentum_buffer(self.fp16_partitioned_groups_flat_numel[i], i,
+                                                 self.fp32_partitioned_groups_flat[i].ds_id)
+                state_buffer = self.optimizer.state[self.fp32_partitioned_groups_flat[i]]["momentum_buffer"]
+                for param, dest_offset, _ in group_items:
+                    momentum_buffer.append(state_buffer.narrow(0, dest_offset, param.partition_numel()).clone())
+            elif self.save_muon_momentum_buffer_in_memory:
+                state_buffer = self.muon_momentum_buffer_partitioned_groups_flat[i]
+                for param, dest_offset, _ in group_items:
+                    momentum_buffer.append(state_buffer.narrow(0, dest_offset, param.partition_numel()).clone())
+            else:
```
tohtana, line 1525: No valid path for `self._swappable_optimizer_subgroup(i) == False` and `self.save_muon_momentum_buffer_in_memory == False`.
If we disable both Muon and NVMe offloading, it will fail.

deepspeed/runtime/zero/stage3.py
```diff
@@ -907,6 +922,20 @@ def _get_sub_group_partitions(self, sub_group_id):
...
+                                                        device=self.device,
+                                                        dtype=self.communication_data_type)
+            unpinned_fp32_buffer_momentum.requires_grad = False
+            if self.fp32_partitioned_groups_flat[i] not in self.optimizer.state:
+                self.optimizer.state[self.fp32_partitioned_groups_flat[i]] = {}
+            self.optimizer.state[
+                self.fp32_partitioned_groups_flat[i]]["momentum_buffer"] = unpinned_fp32_buffer_momentum
+            if self.save_muon_momentum_buffer_in_memory:
+                self.muon_momentum_buffer_partitioned_groups_flat.append(unpinned_fp32_buffer_momentum)
+                self.muon_momentum_buffer_partitioned_groups_flat[i].ds_id = ds_id
```
tohtana, line 943: We append an element to `muon_momentum_buffer_partitioned_groups_flat` only for `muon param group (self.sub_groups_using_muon[i] is True)`, but we access `self.muon_momentum_buffer_partitioned_groups_flat[i]` where `i` is the global index.
If we have non-muon param group earlier, `i` will be beyond the boundary.

deepspeed/runtime/zero/stage3.py
```diff
@@ -780,6 +787,14 @@ def _create_fp16_partitions_with_defragmentation(self, fp16_param_groups):
         param_groups: List[List[Parameter]] = tuple(
             self._create_fp16_sub_groups(param_group["params"]) for param_group in fp16_param_groups)
 
+        if self.use_muon:
+            self.sub_groups_using_muon = []
+            for idx, param_group in enumerate(fp16_param_groups):
+                if getattr(param_group['params'][0], 'use_muon', False):
+                    self.sub_groups_using_muon.extend([True] * len(param_groups[idx]))
+                    self.muon_beta = param_group['momentum']
```
tohtana: If you ave multiple muon param groups. `self.muon_beta` will be overwritten.

PKUWZP: @tohtana Good call out, I have fixed the issue.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1430,6 +1470,118 @@ def __reduce_and_partition_ipg_grads(self, communication_data_type: torch.dtype)
...
+            params_size_offset += param.grad.numel()
+
+        # process muon updates per subgroup to avoid holding all parameters and states at once
+        for i, group_items in params_by_group.items():
+            params = [param for param, _, _ in group_items]
+            if not params:
+                continue
+
+            momentum_buffer = []
+            if self._swappable_optimizer_subgroup(i) and not self.save_muon_momentum_buffer_in_memory:
```
tohtana, line 1512: The condition looks correct, but consider simplifying like the following (should be equivalent)

```
  if self.save_muon_momentum_buffer_in_memory:
      ...
  elif self._swappable_optimizer_subgroup(i):
      ...
  else:
      ...
```

delock:
Suggest to turn on stage 3 for `test_muon_partial_training.py` as well, this would check the case when almost all parameters are freezed so all trainable parameters use muon optimizer.

delock:
I remember deepspeed allow seperate learning rate for muon and adam (muon_lr and adam_lr), can we have a config in UT to cover this usage?

delock (approved):
LGTM, some test case coverage suggestion are added to comments.

docs/_pages/config-json.md
```diff
@@ -62,6 +64,24 @@ The Adam optimizer also supports the following two params keys/values in additio
...
+      "lr": 0.001,
+      "momentum": 0.9,
+      "weight_decay": 0.0,
+      "muon_lr": 0.001
+    }
+  },
+  "zero_optimization": {
+    "stage": 3,
+    "save_muon_momentum_buffer_in_memory": true
+  }
```
delock, line 82: Did Muon optimizer for stage 3 mandates `reduce_scatter = false`?   Does `"reduce_scatter": false` need to be added to the example?

PR #7939, opened 2026-03-31
feat(zero2): add CPU offload support for Muon optimizer

delock:
Add Muon optimizer support in ZeRO Stage 1&2 CPU offload path:
Momentum is stored on CPU memory and Newton-Schultz algorithm happens on GPU.

This PR complete the piece in ZeRO2 and make CPU offload has same numerical behavior with non-CPU offload of ZeRO2.

Note this PR also contains code from PR #7953, this PR is intend to be merged after #7953.


delock:
Close this PR since CPU offload in Z2 slows down performance, makes me think whether Z3 is better choice at this point.  Revisit this if we see needs to combine Muon CPU offload with Z2.

PR #7948, opened 2026-04-02
fix(zero): detach flat buffer to prevent autograd inplace error on CP…

delock:
…U accelerator

The on-device flatten path (introduced in #7828) passes nn.Parameter objects with requires_grad=True to torch.cat(), creating a flat buffer with CatBackward0 grad_fn. Later, _unflatten_dense_tensors produces SplitBackward0 views that are assigned to model params. Inplace copy_() on these views during optimizer step raises:
  RuntimeError: Output 0 of SplitBackward0 is a view and is being modified inplace.

This especially affects CPU training where CPU_Accelerator.is_available() returns True and available_memory() returns system RAM, so the on-device path is always taken.

Fix: add .detach() to the flattened buffer, matching the implicit detach behavior of the CPU-offload path (param.data.cpu() + .to(device)).

Also rename flatten_on_gpu -> flatten_on_accelerator and replace GPU-specific terminology in comments/logs with accelerator-generic equivalents.


tests/unit/v1/zero/test_stage2_flatten_on_gpu.py
```diff
@@ -107,3 +107,40 @@ def test_flat_buffers_on_accelerator(self, zero_stage, dtype):
...
+        engine, _, _, _ = deepspeed.initialize(
+            config=config_dict,
+            model=model,
+            model_parameters=model.parameters(),
+        )
+        for flat in engine.optimizer.bit16_groups_flat:
+            assert flat.grad_fn is None, ("Flat buffer must be detached from autograd graph"
+                                          " to prevent inplace-modification errors during optimizer step")
+
+        data_loader = random_dataloader(model=engine, total_samples=8, hidden_dim=hidden_dim, device=engine.device)
```
tohtana: Shouldn't `random_dataloader` take `dtype`? The default is `preferred_dtype()`, which could mismatch `dtype`.

tohtana (approved):
Thank you, @delock! I left a comment in the new test.

tests/unit/v1/zero/test_stage2_flatten_on_gpu.py
delock: Thanks for the catch!

delock:
@tohtana Thanks for the comments!  I also verified that the newly added test will fail before applying this PR.

PR #7949, opened 2026-04-02
Fix/warnings stacklevel mvapich runner

nathon-lee:
`MVAPICHRunner.backend_exists` calls `warnings.warn` without `stacklevel`,
so Python reports the warning as coming from inside DeepSpeed rather than
the user's code. Adding `stacklevel=2` fixes the reported call site.

No functional change.


tohtana:
Hi @nathon-lee,
Thank you for the PR~ I wonder if `stacklevel=2` still doesn't point the user code. Before this PR, it is`deepspeed/launcher/multinode_runner.py` but it will be `deepspeed/launcher/runner.py`.

nathon-lee:
> Hi @nathon-lee, Thank you for the PR~ I wonder if `stacklevel=2` still doesn't point the user code. Before this PR, it is`deepspeed/launcher/multinode_runner.py` but it will be `deepspeed/launcher/runner.py`.

Hi @tohtana, you're right — stacklevel=2 still points inside DeepSpeed, and since a [RuntimeError](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is raised right after, users won't even see the warning. This change has no real effect. Converting to draft while I reconsider. Thanks!

sfc-gh-truwase:
@nathon-lee do you intend to keep this as draft?

nathon-lee:
Hi, @sfc-gh-truwase 
Thanks for checking! I don’t plan to keep it as draft — I’ll mark it as Ready for review.
This PR only adds stacklevel=2 to warnings.warn (no functional change), so the warning points to user call sites and is easier to debug.
If this looks good to you, could you help merge it? Thank you!

sfc-gh-truwase:
@nathon-lee can you clarify that your [reconsideration](https://github.com/deepspeedai/DeepSpeed/pull/7949#issuecomment-4211006696) based on @tohtana [comment](https://github.com/deepspeedai/DeepSpeed/pull/7949#issuecomment-4209524935) is to merge even if there is no functional impact?

nathon-lee:
Hi @tohtana and @sfc-gh-truwase,

Thanks for the review. After revisiting the call path, I agree with your point: stacklevel=2 does not actually point to user code in this case, so this change does not provide meaningful user-facing value.

Since the launcher also raises a RuntimeError immediately afterward, I do not think this PR is worth keeping open as-is. I will close it for now rather than keep a low-impact change under review.

Thanks again for the careful feedback. If useful, I can come back later with a more meaningful improvement to the MVAPICH error/reporting path.

PR #7951, opened 2026-04-03
Fix DeepCompile+Z3 on PyTorch v2.9/2.10

tohtana:
DeepCompile+Z3 didn't work with PyTorch v2.9/2.10 because:
- PyTorch v2.9+ started enforcing stricter TorchDynamo parameter tensor-match guards. During DeepCompile tracing, some ZeRO-3 parameters were temporarily all-gathered, so Dynamo recorded full sizes such as 4096
- By the time guard evaluation ran, DeepSpeed had already released those params back to the normal ZeRO-3 partitioned representation, where `param.data` is `empty(0)`. That produced guard failures like `expected 4096, actual 0`.

This PR resolves the issue by:
- Leep full-shape dummy tensors for symbolic tracing
- Override guard size/stride metadata for ZeRO-3 params to the stable released representation instead of transient gathered sizes

This PR also includes fixes of these bugs:
- For v2.7 and v2.8, the compiled backward graph could hoist `end_backward` ahead of the real `reduce_grad` calls. - Selective unsharding pass can overcount the persistence memory budget.

Note: DeepCompile is still incompatible with v2.11. It will be addressed by another PR.

### DIFF
--- a/deepspeed/compile/passes/selective_gather.py
+++ b/deepspeed/compile/passes/selective_gather.py
@@ -11,6 +11,7 @@ from torch.fx import GraphModule
 
 import deepspeed.comm as dist
 from deepspeed.accelerator import get_accelerator
+from deepspeed.utils import log_dist
 
 from ..util import get_deepcompile_handle
 from ..graph_param import DSGraphParamManager
@@ -23,8 +24,7 @@ MEM_MARGIN = 0.1
 
 
 def print_rank_0(message):
-    if dist.get_rank() == 0:
-        print(message)
+    log_dist(message, ranks=[0])
 
 
 def _compute_persistence_budget(all_graph_mem_records: List[List[Tuple[str, int, int, int]]], total_mem: int,
--- a/deepspeed/compile/passes/selective_gather.py
+++ b/deepspeed/compile/passes/selective_gather.py
@@ -4,7 +4,7 @@
 # DeepSpeed Team
 
 from collections import defaultdict
-from typing import List, Tuple
+from typing import Dict, List, Tuple
 
 import torch
 from torch.fx import GraphModule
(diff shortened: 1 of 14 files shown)


deepspeed/compile/patch_fake_tensor.py
```diff
@@ -37,8 +48,20 @@ def patch_fake_tensor():
     def wrap_to_fake_tensor_and_record_wrapper(t, *args, **kwargs):
         dummy_tensor = wrap_if_ds_param(t)
         ret = original_wrap_to_fake_tensor_and_record(dummy_tensor, *args, **kwargs)
+        tx = kwargs.get("tx") if "tx" in kwargs else args[0]
+        source = kwargs.get("source")
```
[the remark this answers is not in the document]
tohtana: I don’t think this applies to the torch versions targeted by this PR.

In local v2.9.1 and v2.10.0, `wrap_to_fake_tensor_and_record` has the signature `(e, tx, *, source, is_tensor, parent_context=None)`, so source is keyword-only rather than positional.
At the relevant Dynamo calls in those versions, they pass source=... explicitly rather than positionally. So for the supported matrix here, reading source from kwargs is expected and the guard override path should still run.
The positional fallback is only needed for tx, which can be passed as args[0].

deepspeed/compile/passes/selective_gather.py
```diff
@@ -19,11 +19,45 @@
 
 max_alloc_mem = 0
 last_optimize_step = 0
+MEM_MARGIN = 0.1
+
+
+def print_rank_0(message):
```
eternalNight, line 26: I would recommend using `deepspeed.utils.logging.log_dist` here so that logs can be managed by the log level setting. We can define a simple wrapper (such as`print_rank_0 = lambda message: log_dist(message, ranks=[0])` to avoid duplicating the `ranks=[0]` every time.

tohtana: Hi @eternalNight, thank you for reviewing! 
I replaced `print` with `log_dist`.

eternalNight (approved):

PR #7968, opened 2026-04-11
Enable PyTorch version selection for full test

tohtana:
This PR enables the full test workflow to choose PyTorch version. It resolves each the version into matching `torch` / `torchvision` / `torchaudio` install versions.
It also keeps nightly change detection schedule-only, so manual runs do not affect the daily baseline.


delock (approved):
LGTM

PR #7974, opened 2026-04-14
fix(op_builder): avoid duplicate/wrong -gencode flags

### LINKED ISSUE #7972: [BUG] Duplicate/Wrong(?) Compute capability flags added
https://github.com/deepspeedai/DeepSpeed/blob/0ba235294fcbe35f5c42681bd85666dff5c48a93/op_builder/builder.py#L667-L671

adds `-gencode=` flags for the current devices.

As 
https://github.com/deepspeedai/DeepSpeed/blob/0ba235294fcbe35f5c42681bd85666dff5c48a93/op_builder/builder.py#L571-L573

clears `$TORCH_CUDA_ARCH_LIST` PyTorch will compile for the current device(s?) and add appropriate flags, again.

This results in an overly long command line:
`nvcc [...] -D__CUDA_NO_HALF_OPERATORS__ -D__CUDA_NO_HALF_CONVERSIONS__ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr -gencode=arch=compute_80,code=compute_80 -gencode=arch=compute_80,code=sm_80 [...] -gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_80,code=compute_80 [...]/deepspeed/ops/csrc/fp_quantizer/fp_quantize_impl.cu -o fp_quantize_impl.cuda.o`

Depending on how nvcc handles the repetition this may lead to redundant compilation.

A better approach would be to NOT pass CUDA CC flags and instead set TORCH_CUDA_ARCH_LIST appropriately and let PyTorch handle this which also avoids 
> UserWarning: TORCH_CUDA_ARCH_LIST is not set, all archs for visible cards are included for compilation. 
>  If this is not desired, please set os.environ['TORCH_CUDA_ARCH_LIST'].

Cursx:
## Summary

- Fix duplicate/wrong `-gencode=` flags in both JIT and non-JIT compilation paths (`op_builder/builder.py`)
- Fix `TORCH_CUDA_ARCH_LIST` env-var restore logic in `OpBuilder.jit_load()`

DeepSpeed's `compute_capability_args()` generates its own `-gencode` flags, but PyTorch (`load()` in JIT mode, `BuildExtension` in non-JIT mode) *also* reads `TORCH_CUDA_ARCH_LIST` and generates `-gencode` flags. This causes two problems:

1. **JIT mode**: `jit_load()` set `TORCH_CUDA_ARCH_LIST=""`, which PyTorch treats as *unset* and falls back to auto-detection — resulting in every flag appearing **twice**.
2. **Non-JIT mode**: subclasses that override `filter_ccs()` (e.g. `FPQuantizerBuilder`, `EvoformerAttnBuilder`) remove certain archs, but `BuildExtension` re-reads the **unfiltered** `TORCH_CUDA_ARCH_LIST` and adds them back — **undermining the filter**.

The fix synchronises `TORCH_CUDA_ARCH_LIST` with the filtered arch list in `compute_capability_args()`, for both JIT and non-JIT paths.

Fixes #7972

## Before / After

<details>
<summary>Before (buggy behavior)</summary>

**JIT mode** — `TORCH_CUDA_ARCH_LIST` cleared to `""`, PyTorch auto-detects and adds flags, DeepSpeed also adds the same flags:

```
nvcc ... -gencode=arch=compute_80,code=compute_80 -gencode=arch=compute_80,code=sm_80
     ... -gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_80,code=compute_80
```

Plus a spurious warning:
```
UserWarning: TORCH_CUDA_ARCH_LIST is not set, all archs for visible cards are included for compilation.
```

**Non-JIT mode** — `FPQuantizerBuilder.filter_ccs()` removes `< 8.0`, but `BuildExtension` re-adds them from the unfiltered env var:

```
# FPQuantizer compiled for sm_70 even though filter_ccs() removed it
nvcc ... -gencode=arch=compute_80,code=sm_80   # from DeepSpeed (correct)
     ... -gencode=arch=compute_70,code=sm_70   # from BuildExtension (wrong!)
```

</details>

<details>
<summary>After (fixed behavior)</summary>

**JIT mode** — `TORCH_CUDA_ARCH_LIST` is set to the detected architectures, PyTorch generates flags once, no duplicates:

```
nvcc ... -gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_80,code=compute_80
```

No spurious warning. Env var is properly restored/removed after build.

**Non-JIT mode** — `TORCH_CUDA_ARCH_LIST` is updated to the filtered list. Each extension keeps its own `-gencode` flags, and `BuildExtension` reads the filtered env var:

```
# FPQuantizer: only sm_80+ as intended
nvcc ... -gencode=arch=compute_80,code=sm_80   # from DeepSpeed
     ... -gencode=arch=compute_80,code=sm_80   # from BuildExtension (harmless dup)
```

> **Note:** in multi-builder `setup.py` builds, the last builder's filtered arch list wins for `TORCH_CUDA_ARCH_LIST`. This may cause harmless duplicates for some extensions, but will never reintroduce archs that any builder's `filter_ccs()` removed — a strict improvement over the current behavior where the unfiltered original is always used.

</details>

## Changes

- `op_builder/builder.py`
  - `CUDAOpBuilder.compute_capability_args()`:
    - Always sync `TORCH_CUDA_ARCH_LIST` with the filtered arch list
    - JIT mode: return `[]` (PyTorch generates flags via `load()`)
    - Non-JIT mode: return `-gencode` args as before (per-builder flags in `extra_compile_args`)
  - `OpBuilder.jit_load()`: simplified stash/restore — properly `del` the env var if it was not originally set

### DIFF
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -563,14 +563,12 @@ class OpBuilder(ABC):
         sources = [os.path.abspath(self.deepspeed_src_path(path)) for path in self.sources()]
         extra_include_paths = [os.path.abspath(self.deepspeed_src_path(path)) for path in self.include_paths()]
 
-        # Torch will try and apply whatever CCs are in the arch list at compile time,
-        # we have already set the intended targets ourselves we know that will be
-        # needed at runtime. This prevents CC collisions such as multiple __half
-        # implementations. Stash arch list to reset after build.
-        torch_arch_list = None
-        if "TORCH_CUDA_ARCH_LIST" in os.environ:
-            torch_arch_list = os.environ.get("TORCH_CUDA_ARCH_LIST")
-            os.environ["TORCH_CUDA_ARCH_LIST"] = ""
+        # Stash the original TORCH_CUDA_ARCH_LIST so we can restore it after build.
+        # In JIT mode, compute_capability_args() will set TORCH_CUDA_ARCH_LIST to
+        # the detected GPU architectures, letting PyTorch generate the -gencode
+        # flags. This avoids duplicate flags and the "TORCH_CUDA_ARCH_LIST is not
+        # set" warning.  See https://github.com/deepspeedai/DeepSpeed/issues/7972
+        torch_arch_list = os.environ.get("TORCH_CUDA_ARCH_LIST")
 
         nvcc_args = self.strip_empty_entries(self.nvcc_args())
         cxx_args = self.strip_empty_entries(self.cxx_args())
@@ -603,9 +601,12 @@ class OpBuilder(ABC):
         if verbose:
             print(f"Time to load {self.name} op: {build_duration} seconds")
 
-        # Reset arch list so we are not silently removing it for other possible use cases
-        if torch_arch_list:
+        # Restore the original TORCH_CUDA_ARCH_LIST so we are not silently
+        # modifying it for other possible use cases.
+        if torch_arch_list is not None:
             os.environ["TORCH_CUDA_ARCH_LIST"] = torch_arch_list
+        elif "TORCH_CUDA_ARCH_LIST" in os.environ:
+            del os.environ["TORCH_CUDA_ARCH_LIST"]
 
         __class__._loaded_ops[self.name] = op_module
 
@@ -618,18 +619,22 @@ class CUDAOpBuilder(OpBuilder):
         """
         Returns nvcc compute capability compile flags.
 
-        1. `TORCH_CUDA_ARCH_LIST` takes priority over `cross_compile_archs`.
-        2. If neither is set default compute capabilities will be used
-        3. Under `jit_mode` compute capabilities of all visible cards will be used plus PTX
+        1. Under ``jit_mode`` the visible-card architectures are detected,
+           ``TORCH_CUDA_ARCH_LIST`` is set accordingly, and an **empty list**
+           is returned so that PyTorch generates the ``-gencode`` flags
+           itself (avoiding duplicates).  See
+           https://github.com/deepspeedai/DeepSpeed/issues/7972
+        2. ``TORCH_CUDA_ARCH_LIST`` takes priority over ``cross_compile_archs``.
+        3. If neither is set default compute capabilities will be used.
 
         Format:
 
-        - `TORCH_CUDA_ARCH_LIST` may use ; or whitespace separators. Examples:
+        - ``TORCH_CUDA_ARCH_LIST`` may use ; or whitespace separators. Examples:
 
         TORCH_CUDA_ARCH_LIST="6.1;7.5;8.6;9.0;10.0" pip install ...
         TORCH_CUDA_ARCH_LIST="6.0 6.1 7.0 7.5 8.0 8.6 9.0 10.0+PTX" pip install ...
 
-        - `cross_compile_archs` uses ; separator.
+        - ``cross_compile_archs`` uses ; separator.
 
         """
         ccs = []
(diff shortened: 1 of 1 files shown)


Flamefire:
Shouldn't this always be done, not only for JIT mode? Otherwise you set TORCH_CUDA_ARCH_LIST, `builder` will filter them but leave them set AND add flags so you build for wrong ones in addition to the desired ones. Or am I missing anything?

Cursx:
@Flamefire 
You're right, the non-JIT path should also be modified. I feel the multi-builder scenario might be quite complex. Trying to solve this issue might expand the scope of changes.

I'm looking into a proper fix for this as a follow-up.

delock (approved):
LGTM

op_builder/builder.py
```diff
@@ -618,18 +614,22 @@ def compute_capability_args(self, cross_compile_archs=None):
...
+           is returned so that PyTorch generates the ``-gencode`` flags
+           itself (avoiding duplicates).  See
+           https://github.com/deepspeedai/DeepSpeed/issues/7972
+        2. ``TORCH_CUDA_ARCH_LIST`` takes priority over ``cross_compile_archs``.
+        3. If neither is set default compute capabilities will be used.
 
         Format:
 
-        - `TORCH_CUDA_ARCH_LIST` may use ; or whitespace separators. Examples:
+        - ``TORCH_CUDA_ARCH_LIST`` may use ; or whitespace separators. Examples:
```
Flamefire, line 627:
```suggestion
        - `TORCH_CUDA_ARCH_LIST` may use ; or whitespace separators. Examples:
```

op_builder/builder.py
```diff
@@ -662,17 +662,28 @@ def compute_capability_args(self, cross_compile_archs=None):
             raise RuntimeError(
                 f"Unable to load {self.name} op due to no compute capabilities remaining after filtering")
 
-        args = []
         self.enable_bf16 = True
+        for cc in ccs:
+            if int(cc[0]) <= 7:
+                self.enable_bf16 = False
```
Flamefire, line 668: How about using `any`?

```suggestion
self.enable_bf16 = not any(int(cc[0]) <= 7 for cc in ccs)
```

op_builder/builder.py
```diff
@@ -618,18 +614,22 @@ def compute_capability_args(self, cross_compile_archs=None):
...
         Format:
 
-        - `TORCH_CUDA_ARCH_LIST` may use ; or whitespace separators. Examples:
+        - ``TORCH_CUDA_ARCH_LIST`` may use ; or whitespace separators. Examples:
 
         TORCH_CUDA_ARCH_LIST="6.1;7.5;8.6;9.0;10.0" pip install ...
         TORCH_CUDA_ARCH_LIST="6.0 6.1 7.0 7.5 8.0 8.6 9.0 10.0+PTX" pip install ...
 
-        - `cross_compile_archs` uses ; separator.
+        - ``cross_compile_archs`` uses ; separator.
```
Flamefire, line 632:
```suggestion
        - `cross_compile_archs` uses ; separator.
```

op_builder/builder.py
```diff
@@ -662,17 +662,28 @@ def compute_capability_args(self, cross_compile_archs=None):
...
+        # Keep TORCH_CUDA_ARCH_LIST in sync with the filtered arch list so
+        # PyTorch does not re-add archs that filter_ccs() removed.
+        arch_list = ";".join(f"{cc[0]}.{cc[1]}" for cc in ccs)
+        os.environ["TORCH_CUDA_ARCH_LIST"] = arch_list
+
+        if self.jit_mode:
+            # Let PyTorch generate -gencode flags from the env var.
+            return []
+
+        # Non-JIT: return explicit flags per builder for extra_compile_args.
```
Flamefire, line 679: Won't this cause duplicate, possibly wrong flags again? If `TORCH_CUDA_ARCH_LIST` is used for all extensions and one allows a CC another doesn't you'll still get it for both

I guess for non-JIT mode you'd want to set TORCH_CUDA_ARCH_LIST to a single CC out of the intersection of the CCs allowed by all extensions if that is the case to at least not add the wrong one if you can't avoid the duplication.

Or if more high level is possible:
- Determine CCs to add  (from GPU arch and/or TORCH_CUDA_ARCH_LIST)
- Filter them per extension
- if the intersection is equal to the full list set TORCH_CUDA_ARCH_LIST to all of them, else to one of them or error if empty
- add extension specific flags for each non-filtered arch of that extension that is not in TORCH_CUDA_ARCH_LIST, this might be none

Cursx: I totally agree with your intersection-based approach, but implementing it would likely require changes to setup.py (splitting the builder loop into two passes), and I'm concerned the scope of that refactor could be too large for this PR.

op_builder/builder.py
Cursx: Yes, it is better this way.

PR #7978, opened 2026-04-16
Add Qwen 3.5 preset to AutoTP

### LINKED ISSUE #7947: Has DeepSpeed supported Qwen 3.5 yet?
I’m fine-tuning Qwen 3.5 9B with Axolotl. Using FSDP works fine, but when I use DeepSpeed, it throws an error.

tohtana:
Fixes #7947: This PR adds Qwen 3.5 preset to AutoTP.

### DIFF
--- a/deepspeed/module_inject/auto_tp.py
+++ b/deepspeed/module_inject/auto_tp.py
@@ -477,6 +477,23 @@ class AutoTP():
 
         raise ValueError(f"Unknown AutoTP shape_resolver '{resolver}' for layer '{name}'.")
 
+    def _can_replace_qwen35_linear_attn(self, module, name: str) -> bool:
+        if self.partition_config is None:
+            return False
+        required_attrs = [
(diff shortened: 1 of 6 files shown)


deepspeed/module_inject/auto_tp.py
```diff
@@ -437,37 +437,96 @@ def _replace_with_config(self, child, name):
...
+            v_size = int(config.linear_num_value_heads) * int(config.linear_value_head_dim)
+            if q_size + k_size + v_size != module.weight.shape[0]:
+                raise ValueError(f"AutoTP layer '{name}' resolved fused QKV sizes "
+                                 f"({q_size}, {k_size}, {v_size}) do not match weight output dim "
+                                 f"{module.weight.shape[0]}.")
+            return ((q_size, k_size, v_size), -1)
+
+        raise ValueError(f"Unknown AutoTP shape_resolver '{resolver}' for layer '{name}'.")
+
+    def _can_replace_qwen35_linear_attn(self, module, name: str) -> bool:
```
delock, line 480: It feels like this module is looking at the 'fingerprint' of qwen35.  Eventurally, do we need to extend fingerprint probing to other new models as well?

PR #7993, opened 2026-05-06
docs: add test directory convention to AGENTS.md

delock:
This PR is to tell agents that new tests needs to go to tests/unit/v1 instead of tests/unit/ to enter CI test.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,6 +24,7 @@
 - Write beginner-friendly code: avoid deeply nested expressions or chained logic. Break complex expressions into clear, named intermediate steps.
 - Comments should explain **why**, not **what**. Describe the purpose and reasoning, not the mechanics that the code already shows.
 - New features must include corresponding tests and documentation updates.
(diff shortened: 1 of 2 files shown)


tohtana:
Hi @delock,
This update looks good, but isn't "New tests must go under `tests/unit/v1/`" too strict?
We have the [full nightly test](https://github.com/deepspeedai/DeepSpeed/actions/workflows/aws-torch-latest-full.yml). We sometimes miss issues that only the full test capure when we merge a PR, but it is still good to add some non-critical tests there rather than `tests/unit/v1/`.

The other direction is we can actually add everything to `tests/unit/v1/` and occasionally move them to the nightly test.

delock:
> Hi @delock, This update looks good, but isn't "New tests must go under `tests/unit/v1/`" too strict? We have the [full nightly test](https://github.com/deepspeedai/DeepSpeed/actions/workflows/aws-torch-latest-full.yml). We sometimes miss issues that only the full test capure when we merge a PR, but it is still good to add some non-critical tests there rather than `tests/unit/v1/`.
> 
> The other direction is we can actually add everything to `tests/unit/v1/` and occasionally move them to the nightly test.

Hi @tohtana thanks for the comments.  I agree that we'd expect an agent to figure this out by making a more rational decison based on the situation, rather than following a simple rule.

Let me close this PR and go back if we have better hints.

PR #7999, opened 2026-05-07
zero3: SDMA allgather via mori (sdma_allgather) 

inkcherry:
## Summary
RFC: https://github.com/deepspeedai/DeepSpeed/issues/7884

Wire `sdma_allgather` into ZeRO-3's parameter prefetch path
(`_dist_allgather_fn`).  When enabled, ZeRO-3 allgather routes through
`mori_cpp.AllGatherIntoTensor` (intra-node SDMA copy on AMD MI300), with a
transparent fallback to `dist.allgather_fn` (RCCL/NCCL) on init failure.

End-to-end demo + repro steps + verified numbers live in
[`examples/sdma_allgather/README.md`](examples/sdma_allgather/README.md).

Headline (8x MI300X, DeepSpeed default ZeRO-3 buckets, 100 steps):

| | GPT-7B-ish | Qwen3-32B |
|---|---|---|
| SDMA off | 697.7 ms / step | 1402.5 ms / step |
| SDMA on  | 622.0 ms / step | 1263.2 ms / step |
| **gain** | **+10.85 %**    | **+9.93 %**      |

Loss curves match off ↔ on, peak memory unchanged.

Speedup is workload-dependent — gains shrink (or invert) when allgather can't be overlapped with compute

Co-authored-by: wuyl1 <yangwu@amd.com>


delock:
Hi @inkcherry  @wuyl1, thanks for your contributed code.  I have read this PR and your discussion with @tjruwase in RFC 7884.  And I have some high level comments before going into detailed review.
1. I saw SDMA allgather is exposed in deepspeed config file.  If I understand correctly, this is HW specific option, which should be turn on by default (or huristic) with AMD GPU backend, thus a flag in config file may not be needed.   Config file may be convienient for debugging, but make it transparent to user looks like a better choice.
2.  I agree with @tjruwase 's comments in RFC 7884 that Mori should be integrated into deepspeed.comm as an alternative path to all_gather, which should be transparent (zero modification) to ZeRO3 code.  Is this possible?

The following are current diagram and the ideal diagram
```
                                                                                      
 ┌─────────────────────────────────────────────────────────────────────┐              
 │                        ZeRO-3 (partitioned_param_coordinator)       │              
 │                                                                     │              
 │   with allgather_stream:                                            │              
 │       handle = _dist_allgather_fn(input, output)                    │              
 │       handle.wait()                                                 │              
 └──────────────────────────────┬──────────────────────────────────────┘              
                                │                                                     
                                │  当前 PR 的做法（hardcoded hook）                   
                                │                                                     
                     ┌──────────▼──────────┐                                          
                     │ _dist_allgather_fn  │  ← partition_parameters.py:109           
                     │                     │                                          
                     │  if mori:           │                                          
                     │    sdma allgather   │──── mori.py ──► SDMA 引擎 (硬件)         
                     │  else:              │                                          
                     │    dist.allgather   │──── RCCL ─────► CU (GPU)                 
                     └─────────────────────┘                                          
                     
```

```
 ┌─────────────────────────────────────────────────────────────────────┐           
 │                     理想架构（透明路由）                              │         
 │                                                                     │           
 │   ZeRO-3 只调用:  dist.allgather_into_tensor(output, input)         │           
 └──────────────────────────────┬──────────────────────────────────────┘           
                                │                                                  
                                │                                                  
                     ┌──────────▼──────────┐                                       
                     │   deepspeed.comm    │  ← Backend 选择层                     
                     │                     │                                       
                     │  NVIDIA ──► NCCL    │                                       
                     │  AMD ────► MoriBackend (自动检测)                           
                     └──────────┬──────────┘                                       
                                │                                                  
                     ┌──────────▼──────────┐                                       
                     │    MoriBackend      │  extends TorchBackend                 
                     │                     │                                       
                     │  all_gather_into_tensor()  ──► SDMA 引擎 (硬件)             
                     │  all_reduce()              ──► RCCL ──► CU                  
                     │  reduce_scatter()          ──► RCCL ──► CU                  
                     │  broadcast()               ──► RCCL ──► CU                  
                     │  ...                       ──► RCCL ──► CU                  
                     └─────────────────────┘                                       
```

deepspeed/comm/ccl.py
```diff
@@ -8,7 +8,13 @@
 
 import torch
 from deepspeed.accelerator import get_accelerator
-from deepspeed.ops.op_builder import NotImplementedBuilder
+try:
```
delock: What is the condition of this fallback?   Is this fallback related to Mori library?

deepspeed/runtime/comm/mori.py
```diff
@@ -0,0 +1,229 @@
+# Copyright (c) Microsoft Corporation.
```
delock, line 1: Should replace Microsoft Corporation in licence header with DeepSpeed team.

deepspeed/comm/mori.py
```diff
@@ -0,0 +1,229 @@
...
+When the user opts in, ``deepspeed.comm`` routes ``all_gather_into_tensor``
+on the WORLD process group through ``mori_cpp.AllGatherIntoTensor``
+(intra-node SDMA copy on AMD MI300).  Any failure (mori missing,
+non-AMD/ROCm runtime, shmem init error, oversized call, non-WORLD group)
+yields ``None`` and the caller falls back to the underlying RCCL/NCCL
+allgather.
+
+User-visible controls (env vars, no ``ds_config`` field):
+
+* ``DS_SDMA_ALLGATHER=1``              opt in to the SDMA path.  Required:
```
delock, line 16: Can we use `DS_DMA_ALLGATHER` and `DS_DMA_ALLGATHER_MAX_NUMEL` as behavior control environment variable?  I'm thinking about in the future other accelerators wants to use similiar techniques, and can share the same environment variable name that is not brand binded.

delock (changes requested):

deepspeed/comm/torch.py
```diff
@@ -233,6 +250,15 @@ def all_gather(self, tensor_list, tensor, group=None, async_op=False):
 
     @disable_compiler_collective
     def all_gather_into_tensor(self, output_tensor, input_tensor, group=None, async_op=False):
+        # Transparent SDMA fast-path on AMD/ROCm: when the mori backend is
+        # available and the call is on the WORLD process group, route
+        # through mori_cpp.AllGatherIntoTensor.  Any condition that makes
+        # SDMA unsafe (non-WORLD group, oversized shard, unsupported dtype,
+        # mori unavailable) yields None and we fall through to RCCL/NCCL.
+        from deepspeed.runtime.comm import mori as _mori
+        sdma_work = _mori.allgather_into_tensor(input_tensor, output_tensor, group=group)
```
delock, line 259: What is the overhead here if the hardware does not support mori?  I'm a little bit concerned in potential overhead into this common path.   How long does it take on a CUDA system without Mori library, for example?

deepspeed/comm/mori.py
inkcherry: I initially wanted to use a config file, but considering that currently only the AMD path is available, I set it as an environment variable.
I suggest adding it to the config file in the future if more accelerators support this approach， because environment variables tend to be more specific.

delock: @inkcherry I feel that SDMA implementation should be a drop in replacement of RCCL implementation.  So maybe it should be turn on by default.  In comments, I saw descriptions that Mori allgather may not be bit by bit identical to RCCL implementation.  However allgather has no computation, how could it be not bit by bit identical?   

I agree for *MAX_NUMEL option, if in the future more accelerator use it, we need to move it as a config controled variable and give it a generic name.

deepspeed/comm/ccl.py
inkcherry: removed try/except here.

deepspeed/runtime/comm/mori.py
inkcherry: thanks, fixed.

deepspeed/comm/torch.py
inkcherry: Refactored to a lazy check at TorchBackend.__init__

deepspeed/comm/mori.py
inkcherry: For now, set it as an environment variable. we don't want DeepSpeed's collective backend to silently change behaviour based on which extra packages happen to be installed — mori may end up on the system as a transitive dep of some other framework, not because the user explicitly chose to enable SDMA.

Allgather itself will not have changes. Due to timing variations (data dependencies are correct), end-to-end results might have bit-level differences at the tail due to the order of additions (which can be considered noise, and this is normal in training). However, the loss remains unchanged. Thanks, I have revised the wording to prevent ambiguity.

delock (approved):

delock:
Hi @inkcherry can you resolve merge conflicts?  Thanks!

inkcherry:
> Hi @inkcherry can you resolve merge conflicts? Thanks!

@delock  resolved.

PR #8009, opened 2026-05-17
Fix ZeRO-3 forward crash on modules with plain dict _parameters

### LINKED ISSUE #6961: [BUG] deepspeed fails with torch 2.5 due to module._parameters is a dict, no longer a OrderedDict
**Describe the bug**
File "/opt/conda/lib/python3.10/site-packages/deepspeed/runtime/engine.py", line 1904, in forward
   module._parameters._in_forward = False
 AttributeError: \'dict\' object has no attribute 

**To Reproduce**
use torch 2.5 with ds.

torch 2.4 uses OrderedDict, which can add _in_forward  attribute. 
torch 2.5 uses dict for _parameters, and attribute adding is not supported.

**Expected behavior**
No failure.

**ds_report output**

**Screenshots**

 - OS: [e.g. Ubuntu 18.04]
 - GPU count and types [e.g. two machines with x8 A100s each]
 - Python version

**Launcher context**

**Docker context**

**Additional context**

roycho96:
## Summary

Fixes #6961

ZeRO-3 forward crashes with `AttributeError: 'dict' object has no attribute '_in_forward'` since torch 2.5. PyTorch changed `nn.Module._parameters` from `OrderedDict` to plain `dict` (pytorch/pytorch#129164), and a plain `dict` does not allow attribute assignment.

DeepSpeed wraps every module into `ZeROOrderedDict` at engine init via `_inject_parameters`. Any module not present at that point keeps the plain dict and crashes the next forward. This includes a submodule attached after `deepspeed.initialize()` (PEFT/LoRA adapters), or a module restored by `deepspeed/compile/init_z3.py:35`.

The fix adds `ensure_zero_ordered_dict()` and calls it from the forward prologue. It wraps lazily, is idempotent, and keeps the original container so the deepcompile un-injection path still works. The epilogue gets an `isinstance` guard for modules that show up between the two hooks.

This only fixes the crash. Late-attached parameters are still not in the optimizer and not partitioned by ZeRO-3. For full ZeRO-3 semantics on a late adapter, build it inside `deepspeed.zero.Init()`.

## Tests

`tests/unit/runtime/zero/test_zero_late_module_attach.py`

- forward after attaching a Linear post-init, with `_parameters` forced to plain dict so the bug reproduces on any torch version
- repeated forwards do not re-wrap an already-wrapped module

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -31,7 +31,7 @@ from deepspeed.runtime.zero.stage_1_and_2 import DeepSpeedZeroOptimizer
 from deepspeed.runtime.zenflow.zenflow_stage_1_and_2 import ZenFlowZeroOptimizer
 from deepspeed.runtime.zero.partition_parameters import ZeroParamStatus
 from deepspeed.runtime.zero.utils import is_zero_supported_optimizer, ZeRORuntimeException
-from deepspeed.runtime.zero.parameter_offload import DeepSpeedZeRoOffload
+from deepspeed.runtime.zero.parameter_offload import DeepSpeedZeRoOffload, ZeROOrderedDict, ensure_zero_ordered_dict
 from deepspeed.runtime.zero.config import ZERO_OPTIMIZATION
 from deepspeed.runtime.zenflow.engine import (configure_zenflow, zenflow_step, is_zenflow_update_boundary,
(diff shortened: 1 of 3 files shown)


tohtana (approved):
Thank you for the contribution, @roycho96! This is an important fix.

PR #8010, opened 2026-05-17
Support bf16 optimizer states with CPU offload

lucaspirola:
## What

DeepSpeed's bf16 `bf16_optimizer_states` option and `offload_optimizer` are
currently mutually exclusive, per the support matrix in `docs/_pages/config-json.md`:

| ZeRO 1/2/3 | `bf16_optimizer_states=false` | `bf16_optimizer_states=true` |
|---|---|---|
| | requires ZeRO-Offload + `DeepSpeedCPUAdam`; states **fp32, on CPU** | supported **without offload**; states **bf16, on GPU** |

This PR fills the missing cell: `bf16_optimizer_states=true` *together with*
`offload_optimizer: {device: cpu}` for ZeRO 1/2/3 — Adam moments held in **bf16**
*and* offloaded to **CPU host RAM**. That reduces the offloaded optimizer state
from ~10 to ~6 bytes/param (bf16 master + two bf16 moments) with no added GPU
memory.

## Why

CPU offload currently forces fp32 optimizer states; for large models the
offloaded optimizer state dominates host RAM. Keeping the moments in bf16
(matching the already-bf16 master weights) cuts that footprint substantially
while keeping the state off the GPU.

## How

`DeepSpeedCPUAdam` already supports bf16 momentum/variance through its
`fp32_optimizer_states` constructor flag — the feature was simply not wired up.
No C++/CUDA kernel changes.

- **`engine.py`** — `_configure_basic_optimizer` builds `DeepSpeedCPUAdam` /
  `ZenFlowCPUAdam` with `fp32_optimizer_states=False` when `bf16_optimizer_states`
  is set (a user-supplied value is popped to avoid a keyword clash, and overridden
  with a warning).
- **`base_optimizer.py`** — `_configure_master_weights` runs the offload +
  `DeepSpeedCPUAdam` validator whenever offload is configured (not only for the
  fp32-states case), and asserts a user-provided optimizer actually stores bf16
  moments.
- **`stage3.py` / `stage_1_and_2.py`** — pass `offload_enabled` through.
- **`config-json.md`** — updated bf16 support matrix.

## Backward compatibility

`false`+offload and `true`+no-offload configs are unaffected: the default
resolves to `fp32_optimizer_states=True` (prior behavior), and the no-offload
bf16-states path (FusedAdam on GPU) is untouched.

## Numerics

`bf16_optimizer_states` continues to require `bf16_master_weights_and_grads`, so
master weights are bf16 — identical precision to the existing on-GPU bf16-states
path. CPU Adam computes updates in fp32 internally and rounds moments to bf16
(round-to-nearest-even), matching that path.

## Testing

- `tests/unit/ops/adam/test_cpu_adam.py` — `DeepSpeedCPUAdam` bf16 moment
  allocation + fp32 parity.
- `tests/unit/v1/half_precision/test_bf16.py` — `bf16_optimizer_states` + CPU
  offload across ZeRO 1/2/3 (extends `TestBF16MasterWeightsGradients`), plus a
  guard test that a user-provided `DeepSpeedCPUAdam` must opt into bf16 moments.

All new and affected existing tests pass; `TestBF16MasterWeightsGradients`
(9 cases) was verified on a 2-GPU host.

### DIFF
--- a/deepspeed/ops/adam/cpu_adam.py
+++ b/deepspeed/ops/adam/cpu_adam.py
@@ -64,7 +64,9 @@ class DeepSpeedCPUAdam(torch.optim.Optimizer):
                 (default: False) NOT SUPPORTED in DeepSpeed CPUAdam!
             adamw_mode: select between Adam and AdamW implementations (default: AdamW)
             fp32_optimizer_states: creates momentum and variance in full precision regardless of
-                        the precision of the parameters (default: True)
+                        the precision of the parameters. Set to False to keep optimizer states
+                        in the parameter dtype (e.g. bf16), which reduces the optimizer-state
+                        memory footprint at the cost of lower state precision. (default: True)
         """
 
         default_args = dict(lr=lr,
--- a/deepspeed/runtime/base_optimizer.py
+++ b/deepspeed/runtime/base_optimizer.py
@@ -448,11 +448,14 @@ class ZeROOptimizer(DeepSpeedOptimizer):
                                   fp16_master_weights_and_gradients=False,
                                   bf16_master_weights_and_gradients=False,
                                   bf16_optimizer_states=False,
+                                  offload_enabled=False,
                                   fp16_offload_validator=None,
-                                  bf16_fp32_offload_validator=None):
+                                  bf16_offload_validator=None):
         """
         Common validation and dtype selection for ZeRO optimizer master-weight settings.
         Optionally accepts callables that enforce backend-specific offload requirements.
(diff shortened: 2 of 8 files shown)


tohtana (approved):
This looks good to me, thank you @lucaspirola!
When I developed `bf16_optimizer_states` feature, I wasn't able to include CPU offload due to the lack of the bandwidth. I really appreciate you completing it.

tohtana:
It seems that our CI has an issue with the PyTorch and transformers versions combination. Let me take a look on it.

PR #8011, opened 2026-05-17
Remove stale step() docstring from DeepSpeedCPUAdam

lucaspirola:
## What

Removes an obsolete paragraph from the `DeepSpeedCPUAdam` constructor docstring.

## Why

The docstring described a `step()` option that "updates optimizer states and
copies the parameters back to GPU at the same time" — the old `adam_update_copy`
kernel, invoked via `step(fp16_param_groups=...)`. That fused-copy path no longer
exists in the codebase:

- `csrc/adam/cpu_adam.cpp` binds only `adam_update` (no `adam_update_copy`).
- `DeepSpeedCPUAdam.step()` / `step_subgroup()` take no `fp16_param_groups`
  argument and only call `adam_update`.

So the "two options" text is stale and misleading to anyone reading the API.
Docstring-only change; no functional impact.


tohtana (approved):
Thank you for updating the old doc, @lucaspirola!

PR #8012, opened 2026-05-18
[fix] fix test_zf.py hang bug

xbcReal:
fix:When running following test command, it may hang.
```pytest -v runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[epoch-1-4-False-0-3]```
The reason is that when param.selected_indices got an empty result, its dtype would be torch.float32 instead of torch.int64. However, if the float32 empty tensor is used as an index just like grad_2d[param.selected_indices, :], it would cause a hang. So in order to solve this bug, I add a dtype cast to int64 when judge the param.selected_indices is empty, which means its original dtype is torch.float32.

### DIFF
--- a/deepspeed/runtime/zenflow/engine_stage3.py
+++ b/deepspeed/runtime/zenflow/engine_stage3.py
@@ -278,6 +278,8 @@ def _process_selected_fp32_groups_grad(optimizer_z3, params_to_update, grad_part
         else:
             grad_2d = grad_partition.narrow(0, param.complete_column_offset,
                                             param.complete_numel).view(param.complete_numel // num_row, num_row)
+            if param.selected_indices.numel() == 0:
+                param.selected_indices = param.selected_indices.long()
             param.selected_grad = grad_2d[param.selected_indices, :].clone().detach()
 
             if optimizer_z3.auto_update:

### CODE REVIEW

**Comment by xbcReal** [comment]
You can reproduce what I said by adding some print logs just before the line I added. 

```python
print("param.selected_indices:", param.selected_indices)
print("param.selected_indices.dtype:", param.selected_indices.dtype)
```

And run the following command several times:

```
pytest -v -s unit/runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[epoch-1-4-True-3-3]
```

you would find that when `param.selected_indices` is empty, the dtype would be `torch.float32` and the case would hang.The reason that you should run the same command several times is that not every time you can get an empty `param.selected_indices` tensor.
And I came across this kind of hang bug in the following cases:
```
runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[auto-auto-auto-True-0-3]
runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[auto-auto-auto-True-3-3]
runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[auto-auto-auto-False-0-3]
runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[auto-auto-auto-False-3-3]
runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[step-10-3-True-3-3]
runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[epoch-1-4-True-3-3]
runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[epoch-1-4-False-0-3]
runtime/zenflow/test_zf.py::TestZenFlowDistributed::test_zenflow_distributed[epoch-1-4-False-3-3]
```

tohtana:
Hi @xbcReal, thank you for the fix!
@JoshWoo2003, @Antlera, do you have any comments?

Antlera:
@xbcReal Thanks for the fix! @JoshWoo2003 Could you help check this out？It looks good to me though.

JoshWoo2003:
@xbcReal Thanks for the fix! It looks good to me.

xbcReal:
@tohtana Hi, how could I know is the ci failure caused by my pr? I saw similiar error on other pr too.

tohtana (approved):

tohtana:
Thank you for the fix, @xbcReal! The CI issue has been addressed by another PR.
@Antlera, @JoshWoo2003, I appreciate your help!

PR #8016, opened 2026-05-20
Add configurable torch-latest dependency versions

tohtana:
## Summary

The modal test now shows the error because of the combination of PyTorch v2.7 and Transformer `main` branch, and it is blocking PRs. To address it, we improve our test workflows as follows.

- Add manual dependency version inputs for the torch-latest CI workflows and default the torch-latest family to PyTorch 2.10 plus Transformers git `main`.
- Let CPU and AWS full torch-latest runs select either released Transformers package versions or an explicit Transformers git ref for manual validation.
- Let Modal torch-latest runs select supported PyTorch/CUDA image presets and an optional Transformers git ref, defaulting to `2.10.0-cuda12.8` and Transformers git `main`.

## Known follow-up

- The AWS full real CI lane for PyTorch 2.10 plus Transformers main reached `Unit tests (parallel)` but failed with 33 failures. Some of these may overlap with fixes in #8015; I am opening this PR now so the workflow/input changes can be reviewed while those failures are handled separately.
- CPU and Modal real CI validation for the requested tuple passed.

### DIFF
--- a/.github/workflows/aws-torch-latest-full.yml
+++ b/.github/workflows/aws-torch-latest-full.yml
@@ -326,6 +326,23 @@ jobs:
           pip install --no-build-isolation .[dev,1bit,autotuning,deepcompile]
           ds_report
 
+      - name: Reinstall selected Transformers
+        run: |
+          case "$TRANSFORMERS_SOURCE" in
+            'pypi')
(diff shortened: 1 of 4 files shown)


tohtana:
Addressed the comments @codex

PKUWZP (approved):

PR #8021, opened 2026-05-21
Sort and dedupe -gencode flags emitted by op_builder.builder

### LINKED ISSUE #7871: Let PyTorch set `-gencode` flags
The following code parses `TORCH_CUDA_ARCH_LIST`: https://github.com/deepspeedai/DeepSpeed/blob/93524c8931799a7631a2321d7ef4afaff6b6e54b/op_builder/builder.py#L616-L675

This mostly duplicates the logic already present in PyTorch including falling back to device architectures of visible devices.

Is there any specific reason for that?

I just noticed this while investigating https://github.com/deepspeedai/DeepSpeed/issues/7863 and found that in PyTorch the -gencode flags are sorted while the deepspeed flags are not.
Although it would have resolved the issue in my environment, it just changes the issue but makes it at least independent of the order used in `TORCH_CUDA_ARCH_LIST`

adityasingh2400:
## Summary
- Sort and dedupe `ccs` in `CUDAOpBuilder.compute_capability_args` so the emitted `-gencode` flags are deterministic regardless of the order in which architectures appear in `TORCH_CUDA_ARCH_LIST` or `cross_compile_archs`.
- Matches PyTorch's own canonicalisation, which already sorts the gencode sequence (noted in #7871 while investigating #7863).
- Also dedupes so repeated arches do not produce duplicate `-gencode` entries.

## Why
Issue #7871 observed that PyTorch sorts `-gencode` flags but DeepSpeed emits them in the order entries appear in `TORCH_CUDA_ARCH_LIST`. That order dependence contributed to the regression discussed in #7863. The non-JIT branch in `op_builder/builder.py` did not sort or dedupe before iterating over `self.ccs()`, so calls like `TORCH_CUDA_ARCH_LIST="8.0;7.5;8.0;7.0"` produced an out-of-order, duplicated flag sequence. The JIT branch already sorts (line 669), so this brings the non-JIT branch in line.

## Changes
- `op_builder/builder.py`: after `filter_ccs`, sort and dedupe `ccs` by numeric `(major, minor)` (stripping any `+PTX` suffix for comparison). The downstream `+PTX` handling at the emission site is preserved.
- `tests/unit/ops/test_op_builder.py`: new `test_non_jit_branch_sorts_and_dedupes_gencode_flags` covering the unsorted + duplicated input case. The existing `test_non_jit_branch_unchanged` continues to pass.

## Test plan
- [x] `pytest tests/unit/ops/test_op_builder.py -x -v` (7 passed, including the new test and the prior non-JIT regression test)
- [x] `yapf` (no diff)
- [x] `codespell` (clean)

Fixes #7871

### DIFF
--- a/op_builder/builder.py
+++ b/op_builder/builder.py
@@ -689,41 +689,47 @@ class CUDAOpBuilder(OpBuilder):
             raise RuntimeError(
                 f"Unable to load {self.name} op due to no compute capabilities remaining after filtering")
 
-        # Sort and dedupe so the emitted -gencode sequence is deterministic and
-        # matches PyTorch's own canonicalisation (see #7871). Use a tuple key
-        # of (major, minor, has_PTX) so X.Y stays distinct from X.Y+PTX and the
-        # +PTX variant always sorts after the bare arch, with no ties left for
-        # hash-ordered iteration to resolve non-deterministically.
-        def _cc_sort_key(cc):
+        # Canonicalize by numeric (major, minor) so the emitted -gencode
+        # sequence matches PyTorch's own dedupe (see #7871). For mixed inputs
+        # such as "8.0;8.0+PTX" or "8.0+PTX;8.0", PyTorch collapses to one
+        # sm_80 entry plus one compute_80 PTX entry. Track has_PTX per arch
+        # as the OR across all variants of that arch so any +PTX appearance
+        # carries through after dedupe.
+        canonical = {}
+        for cc in ccs:
             major = int(cc[0])
             minor_part = cc[1]
             has_ptx = minor_part.endswith('+PTX')
             minor = int(minor_part.split('+')[0])
-            return (major, minor, has_ptx)
-
-        unique_ccs = {tuple(cc): None for cc in ccs}
-        ccs = [list(cc) for cc in sorted(unique_ccs, key=_cc_sort_key)]
+            key = (major, minor)
+            canonical[key] = canonical.get(key, False) or has_ptx
+        canonical_archs = sorted(canonical.items())
 
         self.enable_bf16 = True
-        for cc in ccs:
-            if int(cc[0]) <= 7:
+        for (major, _minor), _has_ptx in canonical_archs:
+            if major <= 7:
                 self.enable_bf16 = False
 
         # Keep TORCH_CUDA_ARCH_LIST in sync with the filtered arch list so
-        # PyTorch does not re-add archs that filter_ccs() removed.
-        arch_list = ";".join(f"{cc[0]}.{cc[1]}" for cc in ccs)
-        os.environ["TORCH_CUDA_ARCH_LIST"] = arch_list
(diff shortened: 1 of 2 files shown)


adityasingh2400:
Updated the dedupe to preserve X.Y vs X.Y+PTX as distinct entries while producing deterministic sorted output. The new sort key is a (major, minor, has_PTX) tuple, so 8.0 and 8.0+PTX are never tied and the +PTX variant always sorts after the bare arch. Added a regression test that exercises mixed forms across multiple input orders.

tohtana:
Thank you for submitting this PR, @adityasingh2400! 
Regarding the update, sorting by `(major, minor, has_PTX)` does make the mixed plain/PTX case deterministic.

However, it still seems to keeps `8.0` and `8.0+PTX` as separate entries. That means `TORCH_CUDA_ARCH_LIST="8.0;8.0+PTX"` still emits two `sm_80` flags before the `compute_80` PTX flag.

For mixed duplicate inputs such as `TORCH_CUDA_ARCH_LIST="8.0;8.0+PTX"` or `"8.0+PTX;8.0"`, PyTorch canonicalizes the architecture to one `sm_80` entry plus one `compute_80` PTX entry. So shouldn't we dedupe by the canonical numeric architecture? In that case, the expected outputs will be one sm_80 and one compute_80, not duplicate sm_80 entries.

adityasingh2400:
Good catch, @tohtana. I pushed af7a607 which canonicalizes dedupe by the numeric (major, minor) architecture instead of treating X.Y and X.Y+PTX as separate keys.

For each arch the helper now tracks has_PTX as the OR across all input variants, then emits exactly one `-gencode arch=compute_X,code=sm_X` line plus, when has_PTX is true, one `-gencode arch=compute_X,code=compute_X` PTX line. TORCH_CUDA_ARCH_LIST is rewritten with one token per arch in the X.Y or X.Y+PTX form, matching PyTorch's own canonical parsing.

So TORCH_CUDA_ARCH_LIST="8.0;8.0+PTX" now produces `-gencode arch=compute_80,code=sm_80 -gencode arch=compute_80,code=compute_80` rather than two sm_80 entries, and the reverse-order "8.0+PTX;8.0" gives the same output.

I added test_non_jit_branch_canonical_dedupe_mixed_ptx_combinations covering the four mixed-PTX cases for a single arch (8.0;8.0+PTX, 8.0+PTX;8.0, 8.0;8.0, 8.0+PTX;8.0+PTX) and updated the existing multi-arch test to expect the canonical one-sm-plus-one-PTX layout. All nine unit tests in tests/unit/ops/test_op_builder.py pass locally.

tohtana:
@adityasingh2400 Thank you for the very quick fix! The remaining issue is just formatting in CI. Can you run 
`pre-commit run --all-files` (for full check) and push the updated file?

adityasingh2400:
Done. Ran pre-commit and pushed 7d9eb82.

tohtana (approved):
Great, thank you @adityasingh2400!

PR #8023, opened 2026-05-22
fix(zero): enable vmap on LinearFunctionForZeroStage3

roycho96:
Follow-up to #7916.

Adds `generate_vmap_rule = True` to `LinearFunctionForZeroStage3` so `torch.func.vmap` works on the Function directly. The previous PR covered `grad` / `jacrev` via `setup_context` but not `vmap`.

Test:
`pytest tests/unit/v1/zero/test_zero_functorch_linear.py::TestLinearFunctionVmap`


tohtana (approved):
Thank you @roycho96!

PR #8024, opened 2026-05-24
Fix DeepCompile AOT kwargs patching for PyTorch >= v2.11

tohtana:
DeepCompiles breaks for PyTorch >= v2.11 because these versions can construct the AOT Autograd backend without a bw_compiler kwarg, while DeepSpeed's Inductor patch assumes that key is always present.

This PR fixes DeepCompile's AOT Autograd patch so unrelated AOT backend registrations can pass through unchanged. `TestDeepCompile` passes with this fix.


PKUWZP (approved):
LGTM, thanks for submitting the PR.

PR #8032, opened 2026-05-28
Fix DeepCompile ZeRO-3 release parameter lifetime

tohtana:
PR #7489 made ZeRO-3 all-gather allocate a padded base buffer for uneven shards and return a true-shape view into that buffer. That means the registry tensor and the tensor returned to the compiled graph no longer necessarily share the same `TensorImpl`, although they still share the same underlying storage.

The existing release path only did `set_data(empty)` on the registry tensor before unregistering it. With the new base/view relationship, that clears the registry-side tensor metadata but does not resize the shared `StorageImpl` still referenced by returned views. As a result, the padded gathered allocation can remain live after the final `release_param`.

This patch keeps the release graph ordering unchanged and makes final non-persistent release resize the registered gathered storage to 0 bytes before unregistering it.

### DIFF
--- a/csrc/compile/z3.cpp
+++ b/csrc/compile/z3.cpp
@@ -6,6 +6,8 @@
 #include "z3.h"
 #include "deepcompile.h"
 
+#include <ATen/native/cuda/Resize.h>
+
 namespace dc {
 
 const size_t TIMEOUT_SYMMETRIC_MEMORY_BARRIER = 60000;
@@ -222,6 +224,14 @@ public:
             at::Tensor gathered_param = param_registry_->getGatheredParam(ds_id);
 
             if (gathered_param.defined()) {  // gathered param is undefined while profiling
+                auto storage = gathered_param.storage();
+                if (storage.nbytes() > 0) {
+                    // Required so the caching allocator defers reuse for consumer-stream kernels
+                    // queued behind wait_allgather.
+                    gathered_param.record_stream(at::cuda::getCurrentCUDAStream());
+                    at::native::resize_bytes_cuda(storage.unsafeGetStorageImpl(), 0);
+                }
+
                 const auto options = gathered_param.options();
                 at::Tensor empty_buffer = torch::empty({0}, options);
                 gathered_param.set_data(empty_buffer);
--- a/tests/torch_compile/test_deepcompile_z3_release.py
+++ b/tests/torch_compile/test_deepcompile_z3_release.py
@@ -1,0 +1,130 @@
+# Copyright (c) DeepSpeed Team.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+import math
+
+import pytest
+import torch
+
+import deepspeed.comm as dist
+from deepspeed.accelerator import get_accelerator
+from deepspeed.compile.config import CompileConfig
+from deepspeed.compile.util import get_deepcompile_handle, is_deepcompile_supported
+from unit.common import DistributedTest
+
+pytestmark = pytest.mark.skipif(not is_deepcompile_supported(),
+                                reason="DeepCompile requires CUDA and supported PyTorch")
+
+
+class TestDeepCompileZ3ReleaseStorage(DistributedTest):
+    world_size = 2
+    non_daemonic_procs = True
+
+    def _device(self):
+        return torch.device(get_accelerator().current_device_name())
+
+    def _init_dc(self):
+        dc = get_deepcompile_handle()
+        dc.init(dist.get_world_group(), CompileConfig(deepcompile=True), 1024)
+        return dc
+
+    def _register_param(self, dc, graph_id, ds_id, shape, persistent=False):
+        device = self._device()
+        world_size = dist.get_world_size()
+        true_numel = math.prod(shape)
+        shard_numel = math.ceil(true_numel / world_size)
+        rank = dist.get_rank()
+        values = torch.arange(rank * shard_numel, (rank + 1) * shard_numel, device=device, dtype=torch.float32)
+        grad_buffer = torch.zeros_like(values)
+        dc.register_z3_param(ds_id, list(shape), values, grad_buffer, persistent)
+        dc.register_graph_z3(graph_id, [ds_id])
+        return values
+
+    def _gather_view_and_storage(self, shard, graph_id, ds_id):
+        gathered = torch.ops.dc.allgather_param.default(shard, graph_id, ds_id)
+        gathered = torch.ops.dc.wait_allgather.default(gathered, graph_id, ds_id)
+        view = gathered.reshape(-1).narrow(0, 0, gathered.numel() - 1)
+        assert view.untyped_storage().data_ptr() == gathered.untyped_storage().data_ptr()
+        storage = view.untyped_storage()
+        assert storage.nbytes() >= gathered.numel() * gathered.element_size()
+        return view, storage
+
+    def _release(self, view, graph_id, ds_id, n_users, synchronize=True):
+        torch.ops.dc.release_param.default(view, graph_id, ds_id, n_users)
+        if synchronize:
+            get_accelerator().synchronize()
+
+    def _expected_view_sum(self, shape):
+        world_size = dist.get_world_size()
+        shard_numel = math.ceil(math.prod(shape) / world_size)
+        values = torch.arange(0, world_size * shard_numel, dtype=torch.float32, device=self._device())
+        values = values[:math.prod(shape)].reshape(-1)
+        return values.narrow(0, 0, values.numel() - 1).sum()
+
+    def test_storage_resized_to_zero_after_release_single_use(self):
+        graph_id, ds_id = 9010, 9011
+        dc = self._init_dc()
+        try:
+            shard = self._register_param(dc, graph_id, ds_id, [4097])
+            view, storage = self._gather_view_and_storage(shard, graph_id, ds_id)
+            self._release(view, graph_id, ds_id, 1)
+            assert storage.nbytes() == 0
+        finally:
+            dc.cleanup()
+
+    def test_storage_nonzero_until_final_release_when_multi_use(self):
+        graph_id, ds_id = 9020, 9021
+        dc = self._init_dc()
+        try:
+            shard = self._register_param(dc, graph_id, ds_id, [3])
+            view, storage = self._gather_view_and_storage(shard, graph_id, ds_id)
+            before_release_nbytes = storage.nbytes()
+            self._release(view, graph_id, ds_id, 2)
+            assert storage.nbytes() == before_release_nbytes
+            self._release(view, graph_id, ds_id, 2)
+            assert storage.nbytes() == 0
+        finally:
+            dc.cleanup()
+
+    def test_persistent_param_storage_unchanged_across_release(self):
+        graph_id, ds_id = 9030, 9031
+        dc = self._init_dc()
+        try:
+            shard = self._register_param(dc, graph_id, ds_id, [4], persistent=True)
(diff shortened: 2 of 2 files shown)


eternalNight:
The gathered buffer is registered in `DSParamRegistry` which holds a reference to the tensor until `release_param` removes it from the registry. So I think torch will not release the buffer storage even inductor drops the gathered view early. I was wondering why that ref-count-based mechanism fails in your case?

tohtana:
Thank you for your review, @eternalNight!

The actual issue is that, after #7489, the release op does not release the underlying gathered buffer storage. It is not an early-release issue. I updated the PR description to clarify this.
I also reverted the graph modification after verifying that the existing ordering is still safe.

This PR is much simpler now.

eternalNight:
> The actual issue is that, after #7489, the release op does not release the underlying gathered buffer storage. It is not an early-release issue. I updated the PR description to clarify this. I also reverted the graph modification after verifying that the existing ordering is still safe.

Thanks for the clarification! That makes the picture much clearer. I still have one doubt, though.

The `DSParamRegistry` holds a reference to the raw, possibly-padded buffer. The buffer is expected to be released once that reference is dropped in `unregisterGatheredParam`. If it still persists after being unregistered, there must be a reference alive somewhere else.

You may capture a torch memory history which records exactly when (by recording the call stack) each storage is finally released. There may be more hints in the call stack about the residual reference.

tohtana:
Hi @eternalNight,

Thanks for the suggestion. Yes, that makes sense, and I agree there must still be a live reference after `unregisterGatheredParam`. My current understanding is that the reference is the true-shape view returned from `allgather_param`, rather than an unknown extra reference.

After #7489:

- The registry stores the padded base tensor
- `allgather_param` returns a true-shape view

Those tensors no longer necessarily share the same `TensorImpl`, but they still share the same `StorageImpl`. So `set_data(empty)` on the registry tensor only updates the registry-side tensor metadata. It does not resize the storage that is still visible through the returned view.

That is why this patch resizes the old `StorageImpl` directly with `resize_bytes_cuda(storage.unsafeGetStorageImpl(), 0)`. The intent is not to change the registry tensor metadata again, but to release the underlying padded storage that may still be referenced by the returned view.

eternalNight:
> Hi @eternalNight,
> 
> Thanks for the suggestion. Yes, that makes sense, and I agree there must still be a live reference after `unregisterGatheredParam`. My current understanding is that the reference is the true-shape view returned from `allgather_param`, rather than an unknown extra reference.
> 
> After #7489:
> 
> * The registry stores the padded base tensor
> * `allgather_param` returns a true-shape view
> 
> Those tensors no longer necessarily share the same `TensorImpl`, but they still share the same `StorageImpl`. So `set_data(empty)` on the registry tensor only updates the registry-side tensor metadata. It does not resize the storage that is still visible through the returned view.
> 
> That is why this patch resizes the old `StorageImpl` directly with `resize_bytes_cuda(storage.unsafeGetStorageImpl(), 0)`. The intent is not to change the registry tensor metadata again, but to release the underlying padded storage that may still be referenced by the returned view.

That is like manually freeing a buffer while we know there are some other pointers dangling around, which looks risky to me. Do you have a minimal reproduction of the issue? I can help investigate where the residual reference is and whether it is safe to assume that reference, while being held, will never be touched post `release_param`.

tohtana:
Hi @eternalNight,

I think the residual reference is already identified here: it is the true-shape view returned from `allgather_param`, which can outlive the registry entry while sharing the same `StorageImpl`. That view is a normal graph value passed to downstream consumers, so it is not something the registry can simply delete.
I added `tests/torch_compile/test_deepcompile_z3_release.py` as the minimal repro for this case.

For the safety part of your question, the guarantee I am relying on is that the final `release_param` is placed after the last consumer of that view. The view object may still exist after release, but no later op should read from or write to it.

This follows the existing ZeRO-3 (no-DeepCompile) explicit-release model, which clears parameter data with `set_data(empty)`. After #7489, that is no longer enough in DeepCompile because it only updates the registry tensor, while the returned view can still share the underlying `StorageImpl`. If you see a path where the returned view can be accessed after `release_param`, that is the unsafe case I should fix.

eternalNight:
> Hi @eternalNight,
> 
> I think the residual reference is already identified here: it is the true-shape view returned from `allgather_param`, which can outlive the registry entry while sharing the same `StorageImpl`. That view is a normal graph value passed to downstream consumers, so it is not something the registry can simply delete. I added `tests/torch_compile/test_deepcompile_z3_release.py` as the minimal repro for this case.
> 
> For the safety part of your question, the guarantee I am relying on is that the final `release_param` is placed after the last consumer of that view. The view object may still exist after release, but no later op should read from or write to it.
> 

That's what confused me. In torchinductor-generated graphs, references to tensors are droped after their last use. If the true-shape view is not consumed after `release_param`, its reference should be dropped, which allows torch to release the storage.

I have met a similar issue before with amp enabled, because amp caches casted weights by default. The amp weight cache thus holds another reference to the gathered buffer till the end of the forward phase, preventing `release_param` from releasing the weight buffers early. It can be worked around by passing `cache_enabled=False` to `torch.autocast`.

That's why I would like to have a repro and check its memory profile: storage references may be hold not because how torch inductor tracks data dependency, but because of other mechanisms under the hood. The unit tests do not serve the purpose because they already hold a reference to `storage` till the end for assertion checking.

> This follows the existing ZeRO-3 (no-DeepCompile) explicit-release model, which clears parameter data with `set_data(empty)`. After #7489, that is no longer enough in DeepCompile because it only updates the registry tensor, while the returned view can still share the underlying `StorageImpl`. If you see a path where the returned view can be accessed after `release_param`, that is the unsafe case I should fix.

tohtana:
@eternalNight I got your point, let me do small experiments to clarify it.

tohtana:
Hi @eternalNight,

I ran a separate small DeepCompile ZeRO-3 experiment to check where we keep the reference. At the Python level, the relevant part is roughly just a linear layer:

```python
out = linear(x)
loss = loss_fn(out, target)
```

With DeepCompile / ZeRO-3, the generated code is conceptually closer to this:

```python
bias_buf = wait_allgather(allgather_param(linear.bias))
weight_buf = wait_allgather(allgather_param(linear.weight))

out = torch.addmm(bias_buf, x, weight_buf.t())

out = release_param(out, linear.bias)
out = release_param(out, linear.weight)

loss = loss_fn(out, target)
```

So after the `addmm`, the gathered parameter views should no longer be consumed. However, the generated wrapper I observed keeps local aliases (`buf1`/`buf3`/`buf5`/`buf7` below) of the gathered views alive across the later `release_param` calls:

```python
buf0 = torch.ops.dc.allgather_param.default(...)
buf1 = buf0
del buf0

buf2 = torch.ops.dc.allgather_param.default(...)
buf3 = buf2
del buf2

buf4 = torch.ops.dc.wait_allgather.default(buf3, ...)
buf5 = buf4

buf6 = torch.ops.dc.wait_allgather.default(buf1, ...)
buf7 = buf6

extern_kernels.addmm(buf5, x, reinterpret_tensor(buf7, ...), out=buf8)
del buf4
del buf6

buf9 = torch.ops.dc.release_param.default(buf8, ...)
buf10 = buf9
buf11 = torch.ops.dc.release_param.default(buf10, ...)
buf12 = buf11

triton_kernel.run(..., buf12, ...)
```

This behavior is related to how these DeepCompile custom ops are registered as fallback kernels. Today DeepCompile intentionally marks the outputs as `never_reuse_output=True`:

```python
register_fallback_no_reuse(
    torch.ops.dc.allgather_param.default,
    never_reuse_input=False,
    never_reuse_output=True,
)
register_fallback_no_reuse(
    torch.ops.dc.wait_allgather.default,
    never_reuse_input=True,
    never_reuse_output=True,
)
```

If I change `wait_allgather` and `allgather_param` to `never_reuse_output=False`, Inductor emits deletes for those aliases. However, this version hit a CUDA illegal memory access in my experiment. This matches the reason those outputs are marked no-reuse.

But we cannot make this purely GC-driven either. ZeRO-3 needs the `Parameter` object itself to remain stable across the module lifetime. My understanding is that this is why the existing non-DeepCompile ZeRO-3 path used `set_data(empty)`.

A more complete long-term direction might be to implement a custom lowering / aliasing model for these ops, so Inductor can understand the external lifetime more precisely. That would be a larger change, though.

eternalNight:
> Hi @eternalNight,
> 
> I ran a separate small DeepCompile ZeRO-3 experiment to check where we keep the reference. At the Python level, the relevant part is roughly just a linear layer:
> 
> ```python
> out = linear(x)
> loss = loss_fn(out, target)
> ```
> 
> With DeepCompile / ZeRO-3, the generated code is conceptually closer to this:
> 
> ```python
> bias_buf = wait_allgather(allgather_param(linear.bias))
> weight_buf = wait_allgather(allgather_param(linear.weight))
> 
> out = torch.addmm(bias_buf, x, weight_buf.t())
> 
> out = release_param(out, linear.bias)
> out = release_param(out, linear.weight)
> 
> loss = loss_fn(out, target)
> ```
> 
> So after the `addmm`, the gathered parameter views should no longer be consumed. However, the generated wrapper I observed keeps local aliases (`buf1`/`buf3`/`buf5`/`buf7` below) of the gathered views alive across the later `release_param` calls:
> 
> ```python
> buf0 = torch.ops.dc.allgather_param.default(...)
> buf1 = buf0
> del buf0
> 
> buf2 = torch.ops.dc.allgather_param.default(...)
> buf3 = buf2
> del buf2
> 
> buf4 = torch.ops.dc.wait_allgather.default(buf3, ...)
> buf5 = buf4
> 
> buf6 = torch.ops.dc.wait_allgather.default(buf1, ...)
> buf7 = buf6
> 
> extern_kernels.addmm(buf5, x, reinterpret_tensor(buf7, ...), out=buf8)
> del buf4
> del buf6
> 
> buf9 = torch.ops.dc.release_param.default(buf8, ...)
> buf10 = buf9
> buf11 = torch.ops.dc.release_param.default(buf10, ...)
> buf12 = buf11
> 
> triton_kernel.run(..., buf12, ...)
> ```
> 
> This behavior is related to how these DeepCompile custom ops are registered as fallback kernels. Today DeepCompile intentionally marks the outputs as `never_reuse_output=True`:
> 
> ```python
> register_fallback_no_reuse(
>     torch.ops.dc.allgather_param.default,
>     never_reuse_input=False,
>     never_reuse_output=True,
> )
> register_fallback_no_reuse(
>     torch.ops.dc.wait_allgather.default,
>     never_reuse_input=True,
>     never_reuse_output=True,
> )
> ```
> 
> If I change `wait_allgather` and `allgather_param` to `never_reuse_output=False`, Inductor emits deletes for those aliases. However, this version hit a CUDA illegal memory access in my experiment. This matches the reason those outputs are marked no-reuse.

Thanks for the detailed investigation, @tohtana ! That explains why the residual reference remains till the end of the graph. Torch inductor marks a tensor to be unreusable by keeping its reference alive, which is simple but makes such tensors live longer and eventually increases the peak GPU memory usage.

> 
> But we cannot make this purely GC-driven either. ZeRO-3 needs the `Parameter` object itself to remain stable across the module lifetime. My understanding is that this is why the existing non-DeepCompile ZeRO-3 path used `set_data(empty)`.
> 
> A more complete long-term direction might be to implement a custom lowering / aliasing model for these ops, so Inductor can understand the external lifetime more precisely. That would be a larger change, though.

I agree that a thorough solution would require more systematic analysis. Let's manually free the buffer to unblock the issue for now.

eternalNight (approved):

tohtana:
Thank you, @eternalNight, for sharing your thoughtful insights!

PR #8036, opened 2026-05-29
Fix DeepCompile ZeRO-1 grad target lifetime

tohtana:
DeepCompile ZeRO-1 kept compile-time reduce target buffers alive into the optimizer step, causing backward gradient storage to overlap with optimizer temporaries. This PR fixes the issue by making DeepCompile ZeRO-1 reduce targets follow the normal step-local ZeRO partition gradient-buffer lifetime, instead of preserving cloned target storage from compile setup.

The actual code changes are:
- During compile initialization, register empty DeepCompile ZeRO-1 gradient targets, then bind them to the step-local flat ZeRO partition gradient buffer and per-parameter views when gradients are ready to synchronize.
- After ZeRO-1 builds the optimizer-facing fp32 gradient partition, release the DeepCompile registry references and clear reduce bucket storage after backward synchronization.

|  | Step 10-30 avg sec | Peak alloc GiB |
| --- | --- | ---: |
| Without this PR | 0.858 | 43.594 |
| Without this PR | 0.859 | 39.366 |

Fine-tuning style training (8xH100, Qwen3-8B random weights, bs/GPU=1, seq=4096, GAS=1) showed only finite value losses for 1000 steps.

### DIFF
--- a/csrc/includes/deepcompile.h
+++ b/csrc/includes/deepcompile.h
@@ -291,6 +291,11 @@ public:
         return ds_tensor_;
     }
     at::Tensor getGradBuffer() const { return grad_buffer_; }
+    void setGradBuffer(at::Tensor grad_buffer, int64_t offset)
+    {
+        grad_buffer_ = grad_buffer;
+        offset_ = offset;
+    }
     bool isPartitioned() const { return partitioned_; }
     int64_t getOffset() const { return offset_; }
     void setPersistent(bool persistent) { persistent_ = persistent; }
@@ -395,6 +400,12 @@ public:
         valid_[ds_id] = false;
     }
 
+    void updateGradBuffer(long ds_id, at::Tensor grad_buffer, int64_t offset)
+    {
+        if (grad_buffer.numel() > 0) { grad_buffer.zero_(); }
+        params_.at(ds_id).setGradBuffer(grad_buffer, offset);
+    }
+
     void registerGatheredParam(long ds_id, at::Tensor ds_tensor)
     {
         gathered_params_.emplace(ds_id, ds_tensor);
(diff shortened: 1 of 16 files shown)


csrc/includes/deepcompile.h
```diff
@@ -477,6 +488,10 @@ class CustomOpExecutor {
 
         // This synchronization ensures all of reduce calls are done before optimizer's step.
         at::cuda::stream_synchronize(rs_stream_);
+
+        // Match ZeRO's IPG buffer lifecycle: reduction buckets are backward work buffers and
+        // should not stay allocated while Adam creates optimizer temporaries.
+        reduce_buckets_->clear();
```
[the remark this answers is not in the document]
tohtana: Addressed

deepspeed/compile/init_z1.py
```diff
@@ -26,53 +98,80 @@ def init_z1(engine, backend, compile_config, compile_kwargs, schedule=None, use_
...
+
+        def release_grad_buffer(group_idx=None):
+            group_indices = grad_buffer_metadata.keys() if group_idx is None else [group_idx]
+            for idx in group_indices:
+                for param_id, param, _ in grad_buffer_metadata[idx]:
+                    dc.update_param_grad_buffer(param_id, _empty_grad_buffer(param), 0)
+                if idx in optimizer._deepcompile_z1_current_grad_buffers:
+                    optimizer._deepcompile_z1_current_grad_buffers[idx] = None
+
+        optimizer._deepcompile_z1_release_grad_buffers = release_grad_buffer
```
eternalNight: I may have missed something, but I don't find any reference to `optimizer._deepcompile_z1_release_grad_buffers` in the latest changes. Is it a dangling change after the refinement since the first draft?

Similarly, `optimizer._deepcompile_z1_grad_buffer_metadata` seems out-dated as well. Meanwhile, `optimizer._deepcompile_z1_current_grad_buffers` is still used by the unit tests.

tohtana: @eternalNight Thank you for the catch! After the later refinement, I forgot to remove these. It should be clean now.

eternalNight (approved):

PR #8048, opened 2026-06-04
fix: log block eigenvalue summary events

### LINKED ISSUE #7983: Incorrect variable name: `self.ev_values`
`engine.py:2838-2846`:

```python
    ...
```

he-yufeng:
## Summary
- use the eigenvalues from `self.block_eigenvalue.values()` when building monitor summary events
- avoid indexing the Python 3 `dict_values` view
- add focused coverage for the eigenvalue summary-event path without constructing a full engine

Fixes #7983.

## To verify
- `python -m py_compile deepspeed\runtime\engine.py tests\unit\runtime\test_engine.py` (passed)
- `git diff --check` (passed)
- `python -m pytest tests\unit\runtime\test_engine.py -q` was attempted on Windows, but collection failed before the test ran because this clone has `core.symlinks=false` and `deepspeed/accelerator` was checked out as a plain text symlink file, so `deepspeed.accelerator` cannot be imported locally. The new test should run in the normal Linux checkout used by CI.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -2892,6 +2892,19 @@ class DeepSpeedEngine(Module):
         for param_name, param in self.module.named_parameters():
             param.grad = None
 
+    def _eigenvalue_summary_events(self):
+        if not (self.eigenvalue_enabled() and not self.gas_boundary_ctr % self.eigenvalue_gas_boundary_resolution()):
+            return []
+
+        events = []
+        for i, ev_value in enumerate(self.block_eigenvalue.values()):
+            events.append((
+                f"Train/Eigenvalues/ModelBlockParam_{i}",
+                ev_value[0],
+                self.global_samples,
+            ))
+        return events
+
     def clip_fp32_gradients(self):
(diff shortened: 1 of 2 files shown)


he-yufeng:
Rebased this branch onto the latest upstream `master` and force-pushed the cleaned head. I also amended the commit with DCO signoff to clear the `Signed-off-by` requirement.

Validated locally:
- `python -m py_compile deepspeed/runtime/engine.py tests/unit/runtime/test_engine.py`
- `git diff --check origin/master..HEAD`
- `git log --format=%B -1` includes `Signed-off-by: Yufeng He <40085740+he-yufeng@users.noreply.github.com>`

I also retried the focused pytest target:
- `python -m pytest -q tests/unit/runtime/test_engine.py`

That still fails during collection in this Windows checkout before running the test because `deepspeed/accelerator` is checked out locally as a small link file containing `../accelerator/`, so importing `deepspeed.accelerator` fails. The previous Linux CI did run the DeepSpeed job; the new push should let CI rerun after the rebase/signoff update.

tohtana:
Hi @he-yufeng,
Looks like this PR largely overlaps with your another PR #8049. I’ve approved #8049 and will merge it, so I’ll close this one. Thanks for your contribution!

PR #8049, opened 2026-06-05
fix: log eigenvalue monitor values

### LINKED ISSUE #7983: Incorrect variable name: `self.ev_values`

he-yufeng:
## Summary

- replace the nonexistent `self.ev_values` reference in eigenvalue monitor logging
- avoid indexing the `dict_values` view returned by `self.block_eigenvalue.values()`
- add a small regression helper/test for the event payloads

Fixes #7983.

## To verify

- `python -m py_compile deepspeed\runtime\engine.py tests\unit\runtime\test_engine_eigenvalue.py`
- `git diff --check`
- AST-extracted helper smoke test: passed

I also tried `python -m pytest tests\unit\runtime\test_engine_eigenvalue.py -q` on this Windows checkout. Collection failed before the test body because the local package import resolves `deepspeed.accelerator` differently in this clone/environment. The touched code and the helper itself both compile, and the helper smoke test exercises the exact event construction without importing the full package.


he-yufeng:
Rebased this branch onto the latest upstream `master` and force-pushed the cleaned head.

Validated locally:
- `python -m py_compile deepspeed/runtime/engine.py tests/unit/runtime/test_engine_eigenvalue.py`
- `git diff --check origin/master..HEAD`

I also tried the focused pytest target locally:
- `python -m pytest -q tests/unit/runtime/test_engine_eigenvalue.py`

That collection failed in this Windows checkout before running the test because `deepspeed/accelerator` is checked out as a small link file containing `../accelerator/`, so importing `deepspeed.accelerator` fails locally. The prior upstream CI for this PR had the DeepSpeed test job green; this rebase should let CI rerun on a normal Linux checkout.

tohtana (approved):
Hi @he-yufeng,
Thank you for the contribution! Sorry for reviewing so late. The fix looks good to me.
I made changes about formatting and license.

PR #8050, opened 2026-06-06
Validate fp16 dynamic loss scaling parameters are positive

aryanputta:
## What

`fp16.loss_scale_window` and `fp16.min_loss_scale` drive dynamic loss scaling but are not validated, so invalid values initialize silently and fail later during training:

- **`loss_scale_window`** is used as `stable_interval % self.scale_window` in `DynamicLossScaler.update_scale` (`deepspeed/runtime/fp16/loss_scaler.py`), so a value of `0` raises `ZeroDivisionError` mid-training.
- **`min_loss_scale`** is the loss-scale floor (`max(cur_scale / scale_factor, min_scale)`); a value `<= 0` collapses dynamic loss scaling.

This is the same class of silent-misconfiguration bug as `fp16.loss_scale` accepting `inf`, fixed in #7889.

## Change

Add a single Pydantic `mode="before"` field validator on `DeepSpeedFP16Config` covering both fields. It rejects `bool`, non-numeric, non-finite (`inf`/`-inf`/`nan`), and non-positive values, raising a clear `ValidationError` (e.g. `fp16.loss_scale_window must be > 0`). Following the #7889 review, `mode="before"` runs prior to type coercion (so `True` is rejected), and `float()` is wrapped in `try/except` so `[]`/`{}` surface a clear `ValidationError` rather than a raw `TypeError`.

## Tests

Adds `tests/unit/runtime/test_precision_config_dynamic_scale.py`, parametrized over both fields:
- invalid: `0, -1, inf, nan, True, [], {}` -> `ValidationError`
- valid: `1, 1000, "2"` -> accepted

```bash
pytest -q tests/unit/runtime/test_precision_config_dynamic_scale.py
```

The validator logic was verified against the full matrix locally; the import-level test runs under CI.


aryanputta:
Addressed in 27af879: replaced the per-field validator with a `model_validator(mode="after")` that only enforces `loss_scale_window > 0` and `min_loss_scale > 0` when `self.enabled and self.loss_scale == 0`. Static configs (loss_scale > 0) and fp16-disabled configs now construct fine even if these fields carry 0. Added tests for both the static and fp16-disabled cases.

Thoughts... @tohtana

tohtana:
Hi @aryanputta, thank you for your contribution!

As the fix runs the validator in `mode="after"`, when `loss_scale_window` or `min_loss_scale` are True, they are coerced to 1. Then the boolean values pass the validator.
A minimal fix would be to keep the current "after" validation to check positivity and add a "before" validator that rejects bool/non-numeric/non-finite values.

aryanputta:
Thanks @tohtana, good catch. Fixed in 427695f: added a `field_validator(..., mode="before")` on `loss_scale_window` and `min_loss_scale` that rejects bool, non-finite (inf/nan), and non-numeric values before pydantic coerces them. This mirrors the existing `loss_scale` before-validator. The `mode="after"` model validator still enforces positivity only when dynamic loss scaling is active (fp16 enabled and loss_scale=0), so static-scale configs that carry unused values are unaffected.

Added tests covering bool (True/False), inf/nan, strings, and None for both fields.

tohtana (approved):
Looks good to me, thank you for the fix!

PR #8051, opened 2026-06-07
Fix incorrect variable name 

### LINKED ISSUE #7983: Incorrect variable name: `self.ev_values`

Muneerali199:
Fixes #7983

Two bugs in the eigenvalue monitoring logging code:

1. **Non-subscriptable `dict_values`**: `self.block_eigenvalue.values()` returns a `dict_values` view object which does not support indexing. Wrapped with `list()` to make it subscriptable.

2. **Wrong variable name**: `self.ev_values` was used instead of the local variable `ev_values`. `self.ev_values` is never defined anywhere in the class, causing `AttributeError` when this code path is hit.


tohtana (approved):
@Muneerali199 Thank you for the fix, and sorry for reviewing this so late!

tohtana:
Hi @Muneerali199,
I just realized that #8049 implements the same fix. By convention, we’d like to proceed with the earlier PR. Sorry for the confusion, and thank you for your contribution!

PR #8057, opened 2026-06-10
Fix eigenvalue parsing for compression-only quantize configs

sowndappan5:
## Description
This PR fixes eigenvalue configuration parsing in `deepspeed/runtime/config.py` for quantization-enabled configs.

Previously, `get_eigenvalue_config()` assumed the legacy top-level `quantize_training` section always existed after `get_quantize_enabled()` returned true. That caused a `KeyError` for supported compression configs that enable weight quantization through `compression_training` without also defining `quantize_training`.

This change makes the parser fall back to the default eigenvalue settings when the legacy section is absent, while preserving the existing behavior when `quantize_training` is present.

## Testing
Added a regression test covering a `compression_training` weight-quantization config without a legacy `quantize_training` section.

Validated locally with the repository's focused pre-commit checks.


tohtana:
Hi @sowndappan5,
Thank you for submitting this PR, and sorry for reviewing so late.

The parser change looks correct to me. But it seems to have an issue in the new test.

The new `test_quantized_eigenvalue_config_parses` test currently uses `get_test_path('model/BingBertSquad/...')`. `get_test_path()` resolves relative to `tests/unit`, so this points at `tests/unit/model/...`. The fixture is actually under `tests/model/BingBertSquad`.
When I ran the two focused tests, that test failed before reaching the parser with `binascii.Error: Incorrect
padding` because `DeepSpeedConfig` received a non-existent path and tried to decode it as inline base64.

Changing the path to `get_test_path('../model/BingBertSquad/deepspeed_bsz24_fp16_eigenvalue_quantize_config.json')`
will fix the issue.

sowndappan5:
Thanks for the review and for pointing this out. I'll check the test path on my end and verify the behavior you described. If that's the issue, I'll update the test accordingly and push a fix.

sowndappan5:
@tohtana I have updated the test path to ../model/BingBertSquad/... and pushed the fix. All local pre-commit checks are passing now. Please take another look!

tohtana (approved):
Thank you, @sowndappan5! Let's merge it now.

PR #8058, opened 2026-06-10
feat(zenflow): run the overlapped CPU optimizer in a native process

Antlera:
## What changes this PR introduce

ZenFlow's overlapped CPU optimizer step previously ran in a Python `multiprocessing`
subprocess coordinated by a pickling `Pipe`. This PR moves that optimizer into a
**native CPU optimizer process** packaged inside the `cpu_adam` op, coordinated through a
**shared-memory POSIX-semaphore control block** instead of pickling. The Adam state is
allocated in that process, NUMA-local to the optimizer's pinned thread pool.

Highlights:
- **Fused multi-tensor CPU Adam** (`adam_update_multi`): drives a whole flattened partition
  in C++ and writes the stale snapshot natively, removing the per-parameter Python↔C++ loop
  and the Python-side `clone()`.
- **`ZenFlowAdam`** native class: a pinned `std::thread` pool (pinned to ZenFlow's dedicated
  cores) running the serial Adam kernel per slice, driven from the main process via the
  shared-memory control block (`run_worker` / `submit` / `wait`).
- Covers ZeRO **stages 1, 2, and 3**; removes the old pickling subprocess entirely.
- **Chunked copyback**: streaming the updated fp32 master partition back to the GPU bit16
  partition in chunks drops a transient GPU spike from ~2944 MiB to ~256 MiB for a
  0.75B-param partition (the old `fp32.to(device)` materialized the whole fp32 partition on
  the GPU first).
- `ZenFlowCPUAdam` is now a **recognized ZeRO optimizer**, so `zero_allow_untested_optimizer`
  is no longer required in ZenFlow configs.

## Correctness & performance

- Bit-identical to the previous subprocess path: cross-process and fused-op unit tests, plus
  seeded end-to-end loss across ZeRO stages 1/2/3.
- Real Qwen2.5-1.5B end-to-end (ZeRO-2, CPU offload, 2 GPUs): per-step throughput unchanged
  (no regression). Small / IPC-bound configurations are faster (the per-step pickling/IPC
  overhead is removed).

## Dependency / merge order

This branch is based on top of #7771 ("Fix ZenFlow NaN under PyTorch-style backward"), so its
`backward_prologue` commit rides along here. **Please merge #7771 first, then this PR** — after
#7771 lands, that commit is already in `master` and only the native-optimizer changes remain.

## Testing

- `tests/unit/ops/adam/test_cpu_adam.py`: `test_zenflow_adam_cross_process` (production path,
  bit-identical to the fused reference) and `TestCPUAdamFusedMultiTensor`.
- End-to-end ZenFlow finetuning on ZeRO stages 1/2/3 (single- and multi-GPU).

Note: the native optimizer process uses POSIX semaphores and is Linux-only.


Antlera:
Hi @tohtana @delock, could you help review this PR. Thanks! I am trying to build better cpu_op for zenflow. So that we have better control over the cpu optimizer and memory usage part.

csrc/adam/cpu_adam_impl.cpp
```diff
@@ -338,3 +411,360 @@ int destroy_adam_optimizer(int optimizer_id)
...
+
+// A persistent pool of threads pinned to a fixed core set. parallel_for() splits
+// [0, total) into one contiguous chunk per thread and blocks until all finish.
+class PinnedThreadPool {
+public:
+    explicit PinnedThreadPool(const std::vector<int>& affinity)
+    {
+        n_ = std::max<size_t>(1, affinity.size());
+        for (size_t i = 0; i < n_; ++i) {
+            int core = affinity.empty() ? -1 : affinity[i % affinity.size()];
```
delock: Why the subscription has to mod affinity.size()?

csrc/adam/cpu_adam_impl.cpp
```diff
@@ -338,3 +411,360 @@ int destroy_adam_optimizer(int optimizer_id)
...
+enum { ZEN_CMD_STEP = 0, ZEN_CMD_EXIT = 1 };
+
+struct ZenControl {
+    sem_t cmd_ready;
+    sem_t done;
+    int cmd;
+    int now_state;
+    int64_t step;
+    int num_groups;
+    float hp[ZEN_MAX_GROUPS * 5];  // lr, beta1, beta2, eps, weight_decay per group
```
delock: The number 5 appeared many times in this file, could it be a macro?

deepspeed/runtime/zenflow/engine_stage3.py
```diff
@@ -542,40 +542,41 @@ def zenflow_cpu_optimizer_overlap_step(optimizer_z3, now_state, scaled_global_gr
     if not optimizer_z3.process_optimizer_established:
         optimizer_z3.start_optimizer_process()
 
-    group_infos = []
+    lr, beta1, beta2, eps, weight_decay, bias_correction = [], [], [], [], [], []
```
delock, line 545: stage3 does not chunked copyback as in stage 1/2, is this intended?

tests/unit/ops/adam/test_cpu_adam.py
```diff
@@ -175,6 +177,217 @@ def test_bf16_optimizer_states_match_fp32(self, model_size):
         check_equal(param_fp32_states.float().norm(), param_bf16_states.float().norm(), atol=tolerance)
 
 
+class TestCPUAdamFusedMultiTensor(DistributedTest):
```
delock: The following UT should also be included.
1. correctness check for _copyback_fp32_partition_to_bit16
2. correctness check for _compute_zf_pt_affinity

csrc/adam/cpu_adam_impl.cpp
```diff
@@ -236,6 +260,55 @@ int ds_adam_step(int optimizer_id,
...
+                "ds_adam_step_multi: stale_params length mismatch");
+
+    std::shared_ptr<Adam_Optimizer> opt =
+        std::static_pointer_cast<Adam_Optimizer>(s_optimizers[optimizer_id]);
+    // All tensors share one optimizer step, so advance bias-correction state once.
+    opt->IncrementStep(step, beta1, beta2);
+    opt->update_state(lr, epsilon, weight_decay, bias_correction);
+
+    for (size_t i = 0; i < num_tensors; ++i) {
+        auto params_c = params[i].contiguous();
```
delock: The `contiguous()` operation would make a copy if the original is not contiguous.   This would make changes to `params_c` in `invoke()` call ineffective.

A proper way might be assert for contiguous, or copy the value back if source is not contiguous.   Do you plan to use this function for normal CPU offload?   I noticed that this function is orphan but could be used by normal CPU offload.

Antlera:
Hi @delock. Thanks for the comments and reviews. I will get back to you soon over the reviewed code.

delock:
Hi @Antlera a gentle ping if you have any update on this PR, thanks!

Antlera:
> Hi @Antlera a gentle ping if you have any update on this PR, thanks!

Hi @delock, thanks for the ping and for the detailed review! Sorry for the delay. I'm working through your comments now and will push the updates shortly.

csrc/adam/cpu_adam_impl.cpp
Antlera: Yeah, 5 is the per-group hyperparam count. We should give it a macro.

csrc/adam/cpu_adam_impl.cpp
Antlera: Thanks for catching this. This should be redundant. Will remove.

deepspeed/runtime/zenflow/engine_stage3.py
Antlera: I measured stage 3's direct copy adds 0 GPU memory, so no peak here.

Antlera: I found the memory peak is introduced by redundant `.to(device)`. I will remove this.

tests/unit/ops/adam/test_cpu_adam.py
Antlera: I removed the copyback helper. So I'll add a test for the affinity split logic.

csrc/adam/cpu_adam_impl.cpp
Antlera: Thanks for catching this. This is a temporary function I introduced during the development. The further migration commits make this a orphan.

Antlera:
Hi, thanks again for the detailed reviews @delock, addressed all the comments. All replies inline. Ready for another look when you have time.

delock (approved):

PR #8059, opened 2026-06-11
[DeepCompile] fix gather params in dynamo skipped frames for ZeRO3

### LINKED ISSUE #7942: [BUG][DeepCompile] DeepCompile fails on Qwen1.5-MoE-A2.7B-Chat with `RuntimeError: 'weight' must be 2-D` (LLaMA works in the same environment)
**Describe the bug**
When using **DeepCompile** to train **Qwen1.5-MoE-A2.7B-Chat** (a Qwen-series MoE model), the run fails during the first forward pass with:

The error happens at the embedding lookup:
inputs_embeds = self.embed_tokens(input_ids)

and eventually fails inside:

torch.embedding(weight, input, ...)

In the same environment, DeepCompile can run LLaMA models successfully, but Qwen MoE models fail with the error above.

At least from the traceback, the failure happens before entering the MoE block itself, in the embedding layer. This suggests that under the DeepCompile path, embed_tokens.weight may no longer be a 2-D tensor when F.embedding(...) is called.

Model
	•	Qwen/Qwen1.5-MoE-A2.7B-Chat

What works
	•	LLaMA models run successfully with DeepCompile in the same environment.

What fails
	•	Qwen-series MoE models
	•	Reproduced with Qwen1.5-MoE-A2.7B-Chat

Full traceback
[rank1]: Traceback (most recent call last):
[rank1]:   File "<deepspeed_root>/examples/deepcompile/run_bench_lm.py", line 365, in <module>
[rank1]:     main()
[rank1]:   File "<deepspeed_root>/examples/deepcompile/run_bench_lm.py", line 288, in main
[rank1]:     outputs = model(input_ids=input_ids, attention_mask=attention_mask, labels=input_ids, use_cache=False)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1755, in _wrapped_call_impl
[rank1]:     return self._call_impl(*args, **kwargs)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1766, in _call_impl
[rank1]:     return forward_call(*args, **kwargs)
[rank1]:   File "<deepspeed_root>/deepspeed/utils/nvtx.py", line 20, in wrapped_fn
[rank1]:     ret_val = func(*args, **kwargs)
[rank1]:   File "<deepspeed_root>/deepspeed/runtime/engine.py", line 2237, in forward
[rank1]:     loss = self.module(*inputs, **kwargs)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1753, in _wrapped_call_impl
[rank1]:     return self._compiled_call_impl(*args, **kwargs)  # type: ignore[misc]
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/_dynamo/eval_frame.py", line 662, in _fn
[rank1]:     return fn(*args, **kwargs)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1766, in _call_impl
[rank1]:     return forward_call(*args, **kwargs)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/transformers/utils/deprecation.py", line 172, in wrapped_func
[rank1]:     return func(*args, **kwargs)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/transformers/models/qwen2_moe/modeling_qwen2_moe.py", line 1317, in forward
[rank1]:     outputs = self.model(
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1755, in _wrapped_call_impl
[rank1]:     return self._call_impl(*args, **kwargs)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1766, in _call_impl
[rank1]:     return forward_call(*args, **kwargs)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/transformers/models/qwen2_moe/modeling_qwen2_moe.py", line 974, in forward
[rank1]:     inputs_embeds = self.embed_tokens(input_ids)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1755, in _wrapped_call_impl
[rank1]:     return self._call_impl(*args, **kwargs)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1766, in _call_impl
[rank1]:     return forward_call(*args, **kwargs)
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/modules/sparse.py", line 190, in forward
[rank1]:     return F.embedding(
[rank1]:   File "<conda_env>/lib/python3.10/site-packages/torch/nn/functional.py", line 2551, in embedding
[rank1]:     return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
[rank1]: RuntimeError: 'weight' must be 2-D

Environment
	•	Python: 3.10.12
	•	torch: 2.7.0+cu124
	•	transformers: 4.49.0
	•	deepspeed: 0.18.4+28009dff
	•	accelerate: 1.13.0
	•	triton: 3.0.0
	•	tokenizers: 0.21.4
	•	safetensors: 0.7.0
	•	CUDA: 12.4
accelerate         1.13.0
datasets           3.1.0
deepspeed          0.18.4+28009dff
einops             0.8.1
huggingface_hub    0.36.2
numpy              1.26.4
safetensors        0.7.0
tokenizers         0.21.4
torch              2.7.0+cu124
torchaudio         2.7.0+cu124
torchvision        0.22.0+cu124
transformers       4.49.0
triton             3.0.0

Additional context

A few observations:
	1.	The same setup works for LLaMA models, so this does not look like a generic DeepCompile failure.
	2.	The failure happens in the embedding layer, not inside the expert computation itself.
	3.	The failing model is a Qwen MoE model, so this may be related to how DeepCompile handles this architecture under compilation / sharding.
	4.	The traceback goes through self._compiled_call_impl(...), so the issue appears on the compiled path.

XAheli:
Fixes #7942 

**Root cause:** When `init_z3()` initializes DeepCompile it removes all three parameter-gathering mechanisms (ZeROOrderedDict, module hooks, engine forward hooks) and relies entirely on compiled FX graph ops for allgather/release. but `torch._dynamo` may skip entire frames when it detects graph breaks in for/while loops. Skipped frames execute eagerly with no gathering mechanism, so parameters stay partitioned at shape `[0]`.

## Testing

Validated on 2× H200 with ZeRO3 + DeepCompile:

| Test | Result |
|------|--------|
| **Qwen2 MoE** (actual failing model from #7942) | PASS — 5 training steps, no crash |
| **LLaMA** (regression — already worked) | PASS |
| **Tied embeddings** (shared param across frame types) | PASS |
| **Gradient correctness** (loss decreases on fixed input) | PASS — 12.09 → 10.28 |
| **Guard stability** (no recompilation loops) | PASS — 22 compilations with fix = 22 without |
| **Existing test_compile.py** (100 steps) | PASS |

## Test plan

- [x] `pre-commit run` passes on all changed files
- [x] Existing `tests/torch_compile/test_compile.py` passes (2 GPU, ZeRO-3)
- [x] New regression test `test_deepcompile_skipped_frame.py` passes
- [x] Real Qwen2 MoE model trains without crash
- [x] Real LLaMA model trains without regression
- [x] Tied embedding model trains without crash
- [x] Compilation count unchanged vs upstream (no guard instability)

cc @tohtana @eternalNight

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -137,11 +137,13 @@ from deepspeed.accelerator import get_accelerator
 
 from deepspeed.runtime.config import DtypeEnum
 
-from deepspeed.compile.util import is_deepcompile_supported, get_deepcompile_handle, deepcompile_backward_prologue
+from deepspeed.compile.util import (is_deepcompile_supported, get_deepcompile_handle, deepcompile_backward_prologue,
+                                    deepcompile_backward_epilogue)
 from deepspeed.compile.backend import register_compile_pass, opt_passes
 from deepspeed.compile.passes import zero3_compile, prefetch, selective_gather, offload_adam_states
 from deepspeed.compile.init_z1 import init_z1
 from deepspeed.compile.init_z3 import init_z3
+from deepspeed.compile.z3_eager_fallback import deepcompile_z3_forward_context
 from deepspeed.compile.init_sp import init_autosp
 
 MEMORY_OPT_ALLREDUCE_SIZE = 500000000
@@ -2619,7 +2621,7 @@ class DeepSpeedEngine(Module):
             # We can't have this in forward prologue as the compiler compiles hooks including the forward prologue.
             self.launch_compile_passes(self.global_steps)
 
-        with autocast_if_enabled(self):
+        with deepcompile_z3_forward_context(self), autocast_if_enabled(self):
             loss = self.module(*inputs, **kwargs)
 
         # Register output backward hooks
@@ -2752,6 +2754,9 @@ class DeepSpeedEngine(Module):
             if not bf16_optimizer:
                 self.optimizer.backward_epilogue()
             self.optimizer.exit_backward()
+
+        if self.is_deepcompile_active():
+            deepcompile_backward_epilogue()
 
         see_memory_usage("Engine after backward", force=self.memory_breakdown())
         self._stop_timers(self.engine_timers.backward_reduce_timers)
--- a/deepspeed/compile/init_z3.py
+++ b/deepspeed/compile/init_z3.py
@@ -13,7 +13,8 @@ from deepspeed.runtime.zero.parameter_offload import DeepSpeedZeRoOffload
 from .passes import zero3_compile, prefetch, selective_gather, offload_parameters
 from .backend import make_backend, launch_compile_passes, init_schedule
 from .patch_fake_tensor import patch_fake_tensor
-from .util import get_deepcompile_handle, add_pre_backward_hook
+from .util import get_deepcompile_handle, add_pre_backward_hook, add_post_backward_hook
+from .z3_eager_fallback import DeepCompileZ3EagerFallback
 
 WARMUP = 5
 
@@ -44,9 +45,8 @@ def init_z3(engine, backend, compile_config, compile_kwargs, schedule=None):
     dc = get_deepcompile_handle()
     dc.init(engine.data_parallel_group, compile_config, engine.zero_reduce_bucket_size())
 
-    # Unset hooks
-    for m in engine.module.modules():
-        m._parameters = m._original_parameters
+    engine._deepcompile_z3_eager_fallback = DeepCompileZ3EagerFallback(engine)
+    add_post_backward_hook(engine._deepcompile_z3_eager_fallback.release_gathered_params)
 
     if use_opt:
         optimizer.parameter_offload._remove_module_hooks()
--- a/deepspeed/compile/util.py
+++ b/deepspeed/compile/util.py
@@ -60,10 +60,15 @@ def is_backend_inductor(backend):
 
 backward_started = False
 pre_backward_hooks = []
+post_backward_hooks = []
 
 
 def add_pre_backward_hook(hook):
     pre_backward_hooks.append(hook)
+
+
+def add_post_backward_hook(hook):
+    post_backward_hooks.append(hook)
 
 
 def deepcompile_backward_prologue(is_gradient_accumulation_boundary):
@@ -73,6 +78,11 @@ def deepcompile_backward_prologue(is_gradient_accumulation_boundary):
 
     dc = get_deepcompile_handle()
     dc.start_backward(is_gradient_accumulation_boundary)
+
+
+def deepcompile_backward_epilogue():
+    for hook in post_backward_hooks:
+        hook()
 
 
 def log_rank0(msg: str, enable: bool = False):
--- a/deepspeed/compile/z3_eager_fallback.py
+++ b/deepspeed/compile/z3_eager_fallback.py
@@ -1,0 +1,99 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+from contextlib import contextmanager
+
+import torch
+
+from deepspeed.runtime.zero.parameter_offload import ZeROOrderedDict, ensure_zero_ordered_dict
+from deepspeed.runtime.zero.partition_parameters import ZeroParamStatus
+
+_ACTIVE_FALLBACK = None
+
+
+def get_active_z3_eager_fallback():
+    return _ACTIVE_FALLBACK
+
+
+def record_z3_eager_fallback_param(param):
+    fallback = get_active_z3_eager_fallback()
+    if fallback is None:
+        return False
+    fallback.record_gathered_param(param)
+    return True
+
+
+@contextmanager
+def deepcompile_z3_forward_context(engine):
+    fallback = getattr(engine, "_deepcompile_z3_eager_fallback", None)
+    if fallback is None or not engine.is_deepcompile_active() or not engine.zero_optimization_partition_weights():
+        yield
+        return
+
+    with fallback.forward_context():
+        yield
+
+
+class DeepCompileZ3EagerFallback:
+
+    def __init__(self, engine):
+        self.engine = engine
+        self._depth = 0
+        self._tracked_params = {}
+        self._last_gathered_param_ids = []
+        self._last_released_param_ids = []
+        self.total_gathered_params = 0
+
+    @contextmanager
+    def forward_context(self):
+        global _ACTIVE_FALLBACK
+        previous = _ACTIVE_FALLBACK
+        self._depth += 1
+        if self._depth == 1:
+            self._last_gathered_param_ids = []
+            self._enable_forward_fallback()
+        _ACTIVE_FALLBACK = self
+        try:
+            yield
+        finally:
+            _ACTIVE_FALLBACK = previous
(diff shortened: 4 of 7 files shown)


deepspeed/runtime/engine.py
```diff
@@ -2449,9 +2449,29 @@ def forward(self, *inputs, **kwargs):
...
+                if isinstance(module._parameters, ZeROOrderedDict):
+                    module._parameters._in_forward = True
+
         with autocast_if_enabled(self):
             loss = self.module(*inputs, **kwargs)
 
+        if _dc_z3_eager_fallback:
+            for p in self.module.parameters():
+                if hasattr(p, "ds_status") and p.ds_status == ZeroParamStatus.AVAILABLE and not p.ds_persist:
+                    p.partition()
```
eternalNight: Thanks for fixing this!

One question: I was wondering why we need to walk through all parameters and free those that are still gathered at this point? Does it mean the parameters gathered outside the compiled graphs are all alive till this point? If so, it can increase the peak GPU memory usage, which can hurt training efficiency in some cases.

tohtana:
Thanks @XAheli for digging into this! The root-cause analysis is correct, and keeping `ZeROOrderedDict` as an eager fallback guarded by `torch.compiler.is_compiling()` is the right direction. But I think we still need some rework.

One issue is the post-forward release loop just moves the failure into backward (this confirms the earlier P1 review comment). With DeepCompile actually enabled and the parameters not persistent (`stage3_param_persistence_threshold: 0`), training still crashes in `engine.backward()`.

```
RuntimeError: The size of tensor a (0) must match the size of tensor b (384) at non-singleton dimension 1
```

Autograd saves leaf parameters by reference and reads `param.data` at backward time; ZeRO-3's `free_param` swaps `.data` to an empty tensor. In eager ZeRO-3 the pre-backward hooks swap the data back, but DeepCompile removed those hooks, so partitioning every `AVAILABLE` non-persistent param right after forward frees exactly the weights the skipped frame's eager autograd nodes still need.

Another issue is that this PR releases gathered parameters based on global `ds_status` state that other components also manage. For example, `selective_gather` keeps chosen parameters gathered via the C++-side registry without updating Python's `ds_persist`, so the sweep's `not p.ds_persist` guard cannot exclude them. This kind of interference is reproducible: with DeepCompile actually enabled, the regression test fails on this branch with `KeyError: wait_allgather_ds_param__arg0_1_0` (master fails with the expected `'weight' must be 2-D`), and the failure disappears once the sweep is replaced by tracked-set release.

Your tests didn't catch these issues because `test_deepcompile_skipped_frame.py` never activates DeepCompile (`ds_config_z3.json` doesn't set `"compile": {"deepcompile": true}`). Also, all its parameters are below the default persistence threshold (100000 elements). So the release loop excludes them even when DeepCompile is on. The test passes on master without your fix.

To make this concrete, I've opened a PR against your branch implementing the rework: https://github.com/XAheli/DeepSpeed/pull/1. In summary it:
1. records the params the fallback actually gathers and releases only that set after their eager backward use completes (the all-parameter post-forward sweep is removed);
2. skips `register_external_parameter` in the fallback when DeepCompile is active (its consumers are removed with the module hooks, so the registration is dead state — eager-mode behavior is unchanged);
3. moves the pre/post logic out of `DeepSpeedEngine.forward` into DeepCompile-owned preprocess/postprocess in `deepspeed/compile/`;
4. makes the regression test real: DeepCompile enabled, persistence defeated, asserts the frame was actually skipped, and fails on master without the fix.

**On @eternalNight's question:** This still increases the peak memory, though the cost is now bounded to the fallback-gathered set. But I think it would avoid errors and keep correctness.

Can you take a look at the rework, adjust it as needed, and merge it into your branch if it looks reasonable? (if you'd prefer to address these issues in your own way, that works just as well) Either way, once the backward-safe release Thanks again for working on this!

XAheli:
@tohtana @eternalNight thanks a lot for the detailed review :) I'll take a deeper look and push the changes soon!

tohtana:
Hi @XAheli, just following up here. I’m thinking of using the repair PR I opened against your branch to unblock this:
https://github.com/XAheli/DeepSpeed/pull/1

If you’d like to make the changes yourself or prefer a different approach, please let me know. Otherwise I’ll proceed with this repair path so we can complete this PR.

XAheli:
Hello @tohtana ! Apologies for the delay, I'll be pushing the changes by eod.

XAheli:
@tohtana Thanks again for the review and the repair PR :)

I've reworked the fix to incorporate your approach. I confirmed that post-forward partition corrupts backward. `param.data = shard` between forward and backward breaks autograd because it saves leaf parameters by reference and reads `param.data` at backward time. Simple repro:

```python
p = torch.nn.Parameter(torch.randn(4, 4))
x = torch.randn(4, 4, requires_grad=True)
y = F.linear(x, p)
p.data = torch.zeros(2)  # simulate partition
y.sum().backward()  # RuntimeError: size mismatch
```

The fix now releases gathered params in a post-backward hook (`deepcompile_backward_epilogue`) instead of after forward. Only the params the fallback actually gathered are tracked and released. There's no global `ds_status` sweep.

`register_external_parameter` is skipped in DeepCompile mode, since the module hooks that consume `_external_params` are removed by `init_z3()`.

Tests now activates DeepCompile with the new config (`ds_config_z3_deepcompile_no_persist.json`) sets `"deepcompile": true` with `stage3_param_persistence_threshold: 0`. The model uses `vocab_size=384, hidden=384` so all params exceed the 100k threshold. The test also asserts `fallback_stats.total_gathered_params > 0` to confirm the fallback path was exercised. Uses `@torch.compiler.disable` instead of `graph_break()` in a loop for deterministic behavior. Validated all on 2× H200

- Regression test (DeepCompile ON, persistence=0, skipped frame): PASS 
- Existing test_compile.py (100 steps): PASS
- LLaMA + DeepCompile: PASS
- LLaMA with tied embeddings + DeepCompile: PASS

One thing I want to flag is that Qwen2 MoE + DeepCompile + `persistence_threshold=0` hits a `KeyError: wait_allgather_ds_param__arg0_1_0` in `list_schedule.py:get_last_uses()`. I tested this with current approach (no `_original_parameters` restoration) and it's the same error, so it's not from this PR. On upstream without the fix it's masked by the earlier `'weight' must be 2-D` crash at the embedding lookup. To me looks like a separate graph scheduling issue with MoE model structure when all persistence is disabled. Happy to dig into that if useful!

tohtana (approved):
Hi @XAheli,
Thank you for the update! I also confirmed this fix works on my environment. This is a significant improvement of DeepCompile. I appreciate your contribution.

PR #8060, opened 2026-06-11
Support AutoEP with ZeRO-3 zero.Init source modules

tohtana:
This PR enables ZeRO3 support for AutoEP-managed MoE layers by partitioning expert parameters over expert replica groups while router and replicated parameters use the global data-parallel group.

With ZeRO3 enable, AutoEP preserves global data-parallel gradient averaging for AutoEP expert parameters while reducing them over expert replica groups. ZeRO parameters are gathered before AutoEP reads router or expert tensors when replacing MoE modules created under `deepspeed.zero.Init()`.


deepspeed/runtime/zero/stage3.py
```diff
@@ -1634,12 +1728,18 @@ def __avg_scatter_grads(self, params_to_reduce: List[Parameter],
...
         if self.postscale_gradients and self.gradient_predivide_factor != 1.0 and self.gradient_predivide_factor != dist.get_world_size(
-                self.dp_process_group):
+                process_group):
             grad_partitions_for_rank = [g.mul(self.gradient_predivide_factor) for g in grad_partitions_for_rank]
 
+        partition_world_size = dist.get_world_size(group=process_group)
+        averaging_world_size = self._gradient_averaging_world_size(params_to_reduce, partition_world_size)
+        if averaging_world_size != partition_world_size:
+            scale = partition_world_size / float(averaging_world_size)
+            grad_partitions_for_rank = [g.mul(scale) for g in grad_partitions_for_rank]
```
delock, line 1741: zero3 expert gradient averaging is more complicated than zero2, suggest to add a UT to ensure zero3 averaged gradients are close to zero2 averaged gradients.

deepspeed/runtime/engine.py
```diff
@@ -3766,6 +3896,149 @@ def load_checkpoint(self,
...
+            _AutoEPMoELayer = None
+
+        live_autoep_modules = {}
+        if _AutoEPMoELayer is not None and model is not None:
+            live_autoep_modules = {
+                module_name: module
+                for module_name, module in model.named_modules() if isinstance(module, _AutoEPMoELayer)
+            }
+
+        required_fields = {
```
delock: What is the relationship between `required_fields` and `partitioned_fields` here and the two fields in `ds_to_universal.py`?  They look similiar but not exactly identical.   If they need to be identical, is there a way to ensure the coupling?

from `ds_to_universal.py`
```
required_fields = {
        'moe_layer_id',
        'module_path',
        'num_experts',
        'num_local_experts',
        'ep_size',
        'expert_key_prefix',
}
partitioned_fields = {
        AUTOEP_ZERO3_EXPERT_STATE_FORMAT_VERSION_KEY,
        'ep_rank',
        'expert_data_parallel_rank',
        'expert_data_parallel_world_size',
        'global_expert_start',
        'global_expert_end',
}

File: deepspeed/runtime/engine.py [hunk 3]
```diff
@@ -4291,36 +4635,76 @@ def _save_moe_checkpoint(self, save_dir, tag, client_state={}, exclude_frozen_pa
...
+            if autoep_layer_info:
+                universal_checkpoint_info.setdefault(UNIVERSAL_CHECKPOINT_VERSION_KEY,
+                                                     UNIVERSAL_CHECKPOINT_VERSION_VALUE)
+                universal_checkpoint_info[EXPERT_PARAMETER_PATTERNS] = [r'.*\.experts\.w[123]$']
+                universal_checkpoint_info['ds_autoep_layers'] = autoep_layer_info
+
             state = {
                 'module':
                 model_state_dict,
+                'buffer_names':
```
delock: most of the fields is the same as in `_save_checkpoint()`, is there a way to keep them synced?  i.e. through a builder function:
```
state = self._common_checkpoint_state(model_state_dict) 
state['num_experts'] = self.num_experts                 
state['ds_autoep_layers'] = autoep_layer_info or None
```

deepspeed/utils/zero_to_fp32.py
```diff
@@ -99,10 +101,37 @@ def get_model_state_files(checkpoint_dir):
     return get_checkpoint_files(checkpoint_dir, "*_model_states.pt")
 
 
+def _has_autoep_zero3_partitioned_metadata(state_dict):
```
delock, line 104: it occurs to me currently ZeRO3+AutoEP does not have a path to convert checkpoint to HF compatible format.  Is it on the plan?

delock:
Hi @tohtana , thanks for the great pull request!  Reviewing it is also a great learning opportunity for me.  I have left my review comments in the code, thanks!

deepspeed/runtime/zero/stage3.py
tohtana: Added a 4-rank distributed parity test in `tests/unit/v1/moe/test_autoep_grad_parity.py` that runs ZeRO-2 AutoEP and ZeRO-3 AutoEP from the same initial state and deterministic logical-DP microbatches, then compares the reconstructed expert gradients.

deepspeed/runtime/engine.py
tohtana: Good point. I moved the shared AutoEP ZeRO-3 metadata schema and validation into `deepspeed/checkpoint/autoep_zero3_metadata.py`, and both `DeepSpeedEngine` and `ds_to_universal.py` now use that helper. The required metadata fields and partitioned-entry invariants are now coupled in one place.
The engine still layers runtime model-aware validation on top: when a live model is available, it derives the expected AutoEP module paths / expert prefixes from the `AutoEPMoELayer` instances and rejects checkpoint metadata that does not match the runtime model.

deepspeed/runtime/engine.py
tohtana: Thank you for the suggestion! I added `_common_checkpoint_state(...)` and now both `_save_checkpoint()` and `_save_moe_checkpoint()` use it for the common checkpoint fields. `_save_moe_checkpoint()` only appends the MoE-specific `num_experts` / `ds_autoep_layers` / universal-checkpoint metadata after that.

deepspeed/utils/zero_to_fp32.py
tohtana: Yes, that path is not implemented in this PR. For ZeRO-3 + AutoEP, `zero_to_fp32.py`/`save_16bit_model` intentionally reject the partition-native expert checkpoint because global-DP consolidation would produce incomplete expert tensors. This PR supports the expert-aware Universal Checkpoint conversion path via `ds_to_universal.py`.
HF-compatible export would need a separate expert-aware consolidation path and should be a follow-up.

delock (approved):

PR #8064, opened 2026-06-13
Add AutoEP + AutoTP parallel folding

tohtana:
This PR adds **parallel folding** for AutoEP: tensor parallelism (AutoTP) for the dense/attention path can now coexist with expert parallelism (AutoEP) for the routed-expert path on the same set of ranks, **without forcing EP to be a subset of DP** — an EP group may span TP lanes and dense-DP ranks (cross-lane EP).
(This PR should be adjusted for ZeRO3 support after #8060 is merged)

## Design

Attention/dense and MoE are treated as two independent partitionings of the same rank set, parameterized per parameter family:

- Dense / attention / shared-expert params: `stage_size = tp * dp`
- Routed-expert params: `stage_size = ep * etp * edp`

`dp` and `edp` are always derived, never user-configured, so the invariant `tp * dp == ep * etp * edp == stage_size` cannot be broken from config. The only structural requirement is that the expert width tiles the stage (`stage_size % (ep * etp) == 0`); EP groups are then laid across a TP-lane-major rank ordering, so they may span TP lanes and dense-DP ranks.

### Configuration

No new config section. Folding is expressed by the coexistence of the existing `tensor_parallel` and `expert_parallel` sections:

```json
{
  "tensor_parallel":  { "autotp_size": 4 },
  "expert_parallel":  { "enabled": true, "autoep_size": 4,
                        "expert_tensor_parallel_size": 1 }
}
```

`expert_tensor_parallel_size` is carried as a config field but currently must be `1` (expert-internal TP is reserved as follow-up and rejected fail-fast). Validation enforces stage divisibility, TP/sequence-parallel exclusivity, and `preset_model` consistency between the two sections.

### Cross-lane expert parallelism

Expert parallelism no longer has to be a subset of data parallelism. Shapes where the expert width exceeds (or does not divide) the dense data-parallel size are supported, for example:

- `world=4, TP=4, EP=4` (`dp=1`): the EP group is the whole TP group — one expert per rank.
- `world=4, TP=2, EP=4` (`dp=2`): the EP group spans both TP lanes and both DP ranks.
- `world=8, TP=4, EP=4` (`dp=2, edp=2`): EP groups span TP lanes with expert replication.

The per-family gradient convention is keyed to each parameter's replication structure, not to the EP layout, so it holds across the whole `tp*dp` pool:

- **Router/gate and dense/LayerNorm** are AVERAGE over the TP (token-replication) group. The folded router runs redundantly on every TP peer; its partitioned work is reconstructed into a replicated full view by `restore_combined`, whose all-gather backward injects a `tp_size` factor that AVERAGE divides out.
- **Routed experts** cancel that same `restore_combined` `tp_size` factor (divide by `tp_size`, no TP all-reduce) and reduce data-parallel over the expert-data-parallel (EDP) group. Without the cancellation, folded expert gradients are over-scaled by `tp_size` — invisible to scale-invariant Adam, but real for non-adaptive optimizers and for gradient clipping (it inflates the expert contribution to the global grad norm). This is now fixed for all folded shapes (the MVP TP2×EP4 shape included).

## What's included

- Folded process-group derivation using the generalized expert/data-parallel group creation (`mp_mode` TP-strided vs SP-consecutive ordering), including cross-lane EP group tables.
- Route-full / partition-dispatch path for folded MoE (`deepspeed/moe/ep_tp_dispatch.py`), with AutoTP skipping AutoEP subtrees.
- **Per-family folded gradient reduction**: AVERAGE for replicated router/gate and dense/LayerNorm; a dedicated `tp_size` cancellation for routed experts; SKIP for genuinely TP-sharded params; SUM contracts reserved for a future true sequence-parallel path.
- Per-parameter-family ZeRO checkpoint metadata (routed-expert vs dense/router/shared placement) and folded ZeRO-1/2 optimizer-state handling.

## Correctness & validation

- Router/gate and LayerNorm gradient parity against a non-folded ZeRO baseline (atol=1e-1, rtol=5e-3, fp32), on TP2×EP4 (8-rank) and the cross-lane shapes TP2×EP4 (4-rank, `ep>dp`) and TP4×EP4 (4-rank, `dp=1`); scale 1.0.
- Routed-expert weight parity against a non-folded baseline, verified with SGD (Adam is scale-invariant and would mask a uniform gradient-scale error), for the MVP TP2×EP4 shape and cross-lane TP4×EP4 with `edp=1` and `edp=2`.
- New folding unit tests for config, cross-lane group layout, dispatch, runtime, gradient parity, and checkpoint save/load (multi-rank GPU cases gated for GPU runners; CPU/Gloo parity runs on CI).
- Real-H100 confirmation (8×H100): router/gate, LayerNorm, and routed-expert gradient parity to the non-folded baseline hold (scale 1.0) for MVP TP2×EP4 and cross-lane TP4×EP4 (`edp=1` and `edp=2`); cross-lane folded training runs with finite loss and finite, non-zero expert/router gradients.
- Passes the full unit test suite (`aws-torch-latest-full`) on H100 GPUs.

## Scope / follow-ups

- This PR covers AutoEP + AutoTP folding, including cross-lane EP with `etp=1`. The replicated-grad reduction is mode-aware so the sequence-parallel (Ulysses) folding case fits the same contract; AutoTP + AutoEP is the validated path here.
- Expert-internal tensor parallelism (`expert_tensor_parallel_size > 1`) is reserved for a follow-up.
- ZeRO-3 composition with folding is planned as separate follow-up work (after #8060 is merged).


deepspeed/module_inject/auto_ep_folding.py
```diff
@@ -0,0 +1,261 @@
...
+        raise ValueError("tensor_parallel.autotp_size and Ulysses sequence parallelism are mutually exclusive "
+                         f"for AutoEP folding (autotp_size={spec.tp_size}, sp_size={sp_size}).")
+
+    if spec.etp_size != 1:
+        raise ValueError(f"expert_parallel.expert_tensor_parallel_size={spec.etp_size} is reserved for "
+                         "expert-internal tensor parallelism and is not supported yet. Use 1; ETP support "
+                         "is planned as follow-up work.")
+
+    expert_width = spec.ep_size * spec.etp_size
+    if spec.tp_size > 1 and expert_width > spec.dp_size:
```
delock: What will happen if `expert_width % spec.dp_size != 0`?

deepspeed/runtime/engine.py
```diff
@@ -2530,6 +2592,38 @@ def allreduce_gradients(self, bucket_size=MEMORY_OPT_ALLREDUCE_SIZE):
...
+
+        # TP and SP folding modes produce disjoint per-lane router/shared partials.
+        # Duplicated-token modes already hold a full replicated gradient per lane.
+        partitioned_grad_mode = getattr(folding_spec, "mp_mode", "tp") in ("tp", "sp")
+        tp_world_size = dist.get_world_size(group=tp_group)
+        for _, param in self.module.named_parameters():
+            if not param.requires_grad or param.grad is None:
+                continue
+            if is_moe_param(param) or is_model_parallel_parameter(param):
+                continue
```
delock: This filter cannot distingush router grads from laynorm grads.   The router grads needs SUM because of dispatch of token to experts, but laynorm does not need SUM, so they need different reduce strategy.

delock: Here is the test, I verified on a CPU system that this test case cannot pass.
```
# DeepSpeed Team
"""Engine-path (zero_stage=0) parity: does folded layernorm/gate match the non-folded
DP baseline in the FULL flow? (The author only tested ZeRO-2 parity; the engine path
that runs at zero_stage=0/1 is untested.) CPU/Gloo, world=8.
"""
import deepspeed
from unit.v1.moe.autoep_test_utils import make_autoep_config, run_cpu_gloo_test, seed_everything
from unit.v1.moe.test_autoep_autotp_grad_parity import (
    _router_grad_model,
    _run_router_grad_boundary,
    _full_grad_by_suffix,
)

GATE_BASELINE = "model.layers.0.mlp.gate.weight"
GATE_FOLDED = "model.layers.0.mlp.router.gate.weight"
LN = "model.layers.0.input_layernorm.weight"


def _baseline_cfg():
    c = {k: v for k, v in make_autoep_config(zero_stage=0, ep_size=1, mixed_precision=False).items()
         if k != "expert_parallel"}
    c["gradient_accumulation_steps"] = 2
    c["gradient_clipping"] = 0.0
    c["communication_data_type"] = "fp32"
    c["optimizer"]["params"]["torch_adam"] = True
    return c


def _folded_cfg():
    c = make_autoep_config(zero_stage=0, ep_size=4, mixed_precision=False)
    c["gradient_accumulation_steps"] = 2
    c["gradient_clipping"] = 0.0
    c["communication_data_type"] = "fp32"
    c["optimizer"]["params"]["torch_adam"] = True
    c["expert_parallel"]["autoep_size"] = 4
    c["tensor_parallel"] = {"autotp_size": 2, "partition_config": {
        "use_default_specs": False, "layer_specs": [{"patterns": [r".*\.weight$"], "partition_type": "skip"}]}}
    return c


def _worker(rank, world_size, tmpdir):
    seed = 1234
    tp_size = 2
    logical_dp_world_size = world_size // tp_size
    logical_dp_rank = rank // tp_size

    seed_everything(seed)
    reference_state = _router_grad_model().state_dict()

    baseline_model = _router_grad_model()
    baseline_model.load_state_dict(reference_state)
    baseline_engine, *_ = deepspeed.initialize(model=baseline_model, config=_baseline_cfg())
    _run_router_grad_boundary(baseline_engine,
                              logical_dp_world_size=logical_dp_world_size,
                              logical_dp_rank=logical_dp_rank,
                              seed=seed)
    base_gate = _full_grad_by_suffix(baseline_engine, GATE_BASELINE)
    base_ln = _full_grad_by_suffix(baseline_engine, LN)

    folded_model = _router_grad_model()
    folded_model.load_state_dict(reference_state)
    folded_engine, *_ = deepspeed.initialize(model=folded_model, config=_folded_cfg())
    _run_router_grad_boundary(folded_engine,
                              logical_dp_world_size=logical_dp_world_size,
                              logical_dp_rank=logical_dp_rank,
                              seed=seed)
    folded_gate = _full_grad_by_suffix(folded_engine, GATE_FOLDED)
    folded_ln = _full_grad_by_suffix(folded_engine, LN)

    gate_ratio = (folded_gate.norm() / base_gate.norm()).item()
    ln_ratio = (folded_ln.norm() / base_ln.norm()).item()
    print(f"[rank {rank}] ENGINE(zero0) gate_ratio={gate_ratio:.4f} ln_ratio={ln_ratio:.4f}")

    if rank == 0:
        assert abs(gate_ratio - 1.0) <= 5e-3, f"gate parity: {gate_ratio}"
        assert abs(ln_ratio - 1.0) <= 5e-3, f"ln parity: {ln_ratio}"


def test_b1_engine_path_parity(tmpdir):
    run_cpu_gloo_test(_worker, tmpdir, world_size=8)
```

deepspeed/module_inject/auto_ep_folding.py
tohtana: Good question! Your comment made me realize this was not only a remainder case, but also that the original version was still enforcing a stricter layout than Parallel Folding actually requires.

I generalized the layout after that: EP no longer has to be nested inside dense DP, so the previous checks around `expert_width > dp_size` and `dp_size % expert_width != 0` are removed. In particular, `expert_width` does not need to divide `dp_size`; EP/EDP groups are derived over the per-stage rank set instead.

The remaining expert-side divisibility requirement is that the expert width tiles the stage:

`stage_size % (ep * etp) == 0`

If that does not hold, we still reject it fail-fast because `edp` cannot be derived cleanly. Otherwise `dp` and `edp` are derived under:

`tp * dp == ep * etp * edp == stage_size`

Examples that were previously rejected and now run include `world=4 TP4/EP4` (`dp=1, edp=1`), `world=4 TP2/EP4` (`dp=2, edp=1`), and `world=8 TP4/EP4` (`dp=2, edp=2`).

deepspeed/runtime/engine.py
tohtana: Thanks for the precise diagnosis! You are right that router/gate and LayerNorm gradients cannot be safely distinguished by that old filter. I changed the reduction logic so it is keyed to each parameter family’s actual replication structure instead.

For the current folded AutoEP+AutoTP path, both router/gate and LayerNorm end up using AVERAGE over the TP token-replication group, but for different reasons. LayerNorm is simply TP-replicated. Router/gate is different: we drop the duplicated TP activations before MoE dispatch, then reconstruct the full TP-replicated token view after MoE with `restore_combined`; the backward of that reconstruction introduces a `tp_size` factor, and the TP AVERAGE reduction cancels it. The SUM path is kept for the token-partial / true-SP cases, but it is not the live router/gate path here.

While fixing this, I also found a separate scaling issue for routed-expert weights. Their gradients go through the same `restore_combined` backward factor, but they had been classified as SKIP, so the `tp_size` factor reached the optimizer. That made folded expert gradients `tp_size` too large. The fix now cancels that factor by dividing routed-expert grads by `tp_size` without adding a TP all-reduce; it composes with the existing EDP all-reduce.

Your `zero_stage=0` reproducer now passes on the head of this PR branch. I also added parity coverage against the non-folded baseline for router/gate, LayerNorm, and routed-expert weights across TP2/EP4 and TP4/EP4 folded shapes, including ZeRO 0 and ZeRO 2.

PKUWZP:
@tohtana Thanks for the extensive amount of work, super appreciated. A couple of questions (I will also run some tests on my end): 

- Unbucketed per-param TP `all_reduce. _reduce_autoep_folding_tp_replicated_gradients` walks
  `module.named_parameters()` and issues an individual `all_reduce` per replicated param at each grad-accum boundary(ZeRO-0/1 path). It might be fine for the current PR stage with small replicated set, but worth a note. 
- We might want to run tests on multi-node settings (e.g. 2 node = 16 H100/200) to benchmark the performance.

deepspeed/runtime/engine.py
delock: @tohtana Thank you for fixing this.  I have no further review comments.

delock (approved):

PKUWZP (changes requested):
Thanks for this comprehensive PR @tohtana. What I understand about this PR is that it enables AutoTP (tensor parallelism for the dense/attention path) and AutoEP (expert parallelism for routed experts) coexist on the same ranks without forcing EP to be a subset of DP. It adds a folding topology module (`auto_ep_folding.py`), a route-full/partition-dispatch MoE path (`ep_tp_dispatch.py`), per-parameter-family gradient reduction (AVERAGE for replicated router/dense, a `tp_size`-cancel for routed experts), and per-family ZeRO checkpoint metadata. the parity tests are fp32 + non-partitioned only, which leaves a class of gradient-pipeline configs unexercised.

One risk from my perspective is that the folded TP gradient reduction is bolted onto `allreduce_gradients()` reading `param.grad`, but several permitted configs consume, snapshot, or clear `param.grad` before that point — so the correction silently doesn't apply.

Several thoughts on this PR: 

1. `deepspeed/runtime/engine.py:2825` — BF16 folded gradients skip the TP reduction entirely.   In `_backward_epilogue`, `self.optimizer.backward_epilogue()` (→ `update_hp_grads`, which does `hp_grad.data.add_(lp.grad...)`) runs before `allreduce_gradients()`, which is where `_reduce_autoep_folding_tp_replicated_gradients()` mutates `lp.grad` in place. The optimizer steps on the fp32 HP buffer built from the uncorrected LP grad. With bf16 + tp_size>1 (permitted — `validate_folding_global` doesn't reject it, and the parity tests are fp32-only), routed-expert grads stay `tp_size` multiply, which is too large and router/dense grads are never averaged → wrong global grad norm/clipping (and wrong updates under SGD/Lion/Muon).
2. `deepspeed/runtime/zero/stage_1_and_2.py:1033` — ZeRO-1 + `overlap_comm` loses the folding reduction. `_maybe_reduce_autoep_folding_tp_gradient` early-returns on `not self.partition_gradients`, delegating stage-1 to the engine boundary sweep. But `overlap_comm=True` installs reduce hooks (line 625) that clear `param.grad = None` during backward (lines 1045/1066), so by the time `_reduce_autoep_folding_tp_replicated_gradients` runs it finds grad is None and skips every param. The potential effect: no TP folding correction for ZeRO-1 + `overlap_comm`, with no fail-fast guard flag.
3. `deepspeed/runtime/engine.py:2745` — DeepCompile silently bypasses the folding `reduction. allreduce_gradients()` returns early when `is_deepcompile_active()` and `not self.compile_autosp()`, before reaching `_reduce_autoep_folding_tp_replicated_gradients()` (line 2751). Nothing in `validate_folding_global` rejects DeepCompile + folding, so a folded DeepCompile run produces uncorrected gradients with no error.

1-3 are the ones we should act on before merge — ideally by adding an explicit fail-fast in `validate_folding_global` for `bf16/overlap_comm/DeepCompile` until the reduction is integrated into the real gradient pipeline (or moving the correction to where the `fp32/partitioned` grads are actually assembled).

I think the PR is almost ready to be merged. Let me know your thoughts:) Great work! @tohtana

tohtana:
Hi @PKUWZP,
Thanks for the detailed review. I agree these were real correctness gap. I pushed updates to address the three blocking items:

1. BF16 folded gradients: The folded TP correction now runs before BF16 high-precision grad accumulation, so the fp32 HP grad buffer, grad norm, clipping, and optimizer step see the corrected gradients. I also added a correction marker for this early-correction path to avoid double correction.

2. ZeRO-1 + overlap_comm: The ZeRO-1 overlap path now applies the folded TP correction before the overlap hook can consume or clear the gradient. This uses the same marker to avoid a second engine-boundary correction. The non-overlap ZeRO-1 path remains owned by the engine-boundary sweep. ZeRO-2 partitioned-gradient hooks continue to apply the correction per hook without using that marker, so gradient accumulation hooks are not accidentally skipped.

3. DeepCompile + folding: Folded AutoEP+AutoTP with DeepCompile is now explicitly rejected during config validation. Non-folded AutoEP + DeepCompile remains accepted.

tohtana:
Hi @PKUWZP,

> Unbucketed per-param TP all_reduce. _reduce_autoep_folding_tp_replicated_gradients walks
module.named_parameters() and issues an individual all_reduce per replicated param at each grad-accum boundary(ZeRO-0/1 path). It might be fine for the current PR stage with small replicated set, but worth a note.

Good point. I am leaving this as the simple per-parameter correction in this PR because this PR is already pretty large. Let's address this in a following PR.

> We might want to run tests on multi-node settings (e.g. 2 node = 16 H100/200) to benchmark the performance.

I ran a 2x8 H100 benchmark on `Qwen/Qwen3.6-35B-A3B` with the model reduced to 20 layers. Sequence length is 1k, and all rows below use the same global batch size. Sequence length: 1k, effective batch size: 64, micro batch size: 1 for ZeRO3 and TP1 / 2 for TP2 (to match the effective batch size in an iteration).

Config | tokens / optimizer step | Mean throughput | Peak allocated
-- | -- | -- | --
ZeRO-3 leaf, TP1/EP1 | 65,536 | 27,941 tok/s (1,746/GPU) | 52.23 GiB
ZeRO-1, TP1DP16/EP8 | 65,536 | 57,414 tok/s (3,588/GPU) | 31.11 GiB
ZeRO-1, TP2DP8/EP8 | 65,536 | 33,477 tok/s (2,092/GPU) | 37.55 GiB

TP2DP8/EP8 is slower than TP1DP16/EP8 here because of the additional TP communication. TP in attention could still be useful for memory at a higher TP degree, but this model only has 2 KV heads. It would be better to benchmark across more models, but I think that should be another PR.

PKUWZP:
@tohtana Thanks for all the additional work, I think we should merge this PR first and leave all the remaining work to another PR.

PKUWZP (approved):
LGTM.

PR #8066, opened 2026-06-15
Mixed-precision: per-policy param/buffer dtype cast (preserve fp32 buffers)

sfc-gh-truwase:
## Summary
- Add `data_types.param_dtype` and `data_types.buffer_dtype` (both default `None`), mirroring FSDP `MixedPrecisionPolicy`.
- Replace the blanket `module.half()` / `module.bfloat16()` in `_configure_distributed_model` with a targeted cast: parameters go to `param_dtype`; floating buffers keep their loaded dtype unless `buffer_dtype` is explicitly set.

## Motivation
The blanket cast downcasts every floating buffer, including the rotary `inv_freq` buffer that HF/FSDP2 keep in fp32. On long contexts the bf16 `inv_freq` loses precision, RoPE angles drift, and logits/grads diverge from the FSDP2 reference. Preserving fp32 buffers by default fixes this; `buffer_dtype` is the escape hatch to reproduce the legacy behavior.

## Behavior
- `param_dtype` unset -> derived from the fp16/bf16 enabled flag (legacy param behavior).
- `buffer_dtype` unset -> buffers keep their loaded dtype (e.g. fp32 `inv_freq`).
- `buffer_dtype` set -> buffers force-cast (legacy blanket-cast parity).

## Test plan
- [ ] `param_dtype=bf16`, `buffer_dtype` unset -> params bf16, `inv_freq` stays fp32.
- [ ] `buffer_dtype=bf16` -> buffers downcast (legacy parity).
- [ ] bf16/fp16 run with neither key set behaves as before except fp32 buffers preserved.
- [ ] 8B / 32B ZeRO-3 long-context run -> grad_norm tracks the FSDP2 reference.

### DIFF
--- a/deepspeed/runtime/config.py
+++ b/deepspeed/runtime/config.py
@@ -860,6 +860,10 @@ class DeepSpeedConfig(object):
 
         data_types_params = get_data_types_params(param_dict)
         self.grad_accum_dtype = data_types_params.get(GRAD_ACCUM_DTYPE, GRAD_ACCUM_DTYPE_DEFAULT)
+        # Raw strings ("bf16"/"fp16"/"fp32") or None; resolved via DtypeEnum at
+        # use-time. buffer_dtype=None keeps buffers at their loaded dtype.
+        self.param_dtype = data_types_params.get(PARAM_DTYPE, PARAM_DTYPE_DEFAULT)
+        self.buffer_dtype = data_types_params.get(BUFFER_DTYPE, BUFFER_DTYPE_DEFAULT)
 
         par_write_pipe = get_checkpoint_parallel_write_pipeline(checkpoint_params)
         self.checkpoint_parallel_write_pipeline = par_write_pipe
--- a/deepspeed/runtime/config.py
+++ b/deepspeed/runtime/config.py
@@ -861,8 +861,9 @@ class DeepSpeedConfig(object):
         data_types_params = get_data_types_params(param_dict)
         self.grad_accum_dtype = data_types_params.get(GRAD_ACCUM_DTYPE, GRAD_ACCUM_DTYPE_DEFAULT)
         # Raw strings ("bf16"/"fp16"/"fp32") or None; resolved via DtypeEnum at
-        # use-time. buffer_dtype=None keeps buffers at their loaded dtype.
+        # use-time.
         self.param_dtype = data_types_params.get(PARAM_DTYPE, PARAM_DTYPE_DEFAULT)
+        # buffer_dtype=None keeps buffers at their loaded dtype.
         self.buffer_dtype = data_types_params.get(BUFFER_DTYPE, BUFFER_DTYPE_DEFAULT)
 
         par_write_pipe = get_checkpoint_parallel_write_pipeline(checkpoint_params)
--- a/deepspeed/runtime/constants.py
+++ b/deepspeed/runtime/constants.py
@@ -470,11 +470,9 @@ CHECKPOINT_PARALLEL_WRITE_PIPELINE_STAGE_DEFAULT = False
 #   buffer_dtype=["bf16"|"fp16"|"fp32"]
 #   }
 # }
-# param_dtype / buffer_dtype mirror FSDP MixedPrecisionPolicy. When unset (None):
-#   - param_dtype follows the enabled bf16/fp16 mode; if set it must match that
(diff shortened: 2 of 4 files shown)


stas00 (approved):

deepspeed/runtime/engine.py
```diff
@@ -1322,6 +1322,42 @@ def get_data_types(self):
...
         return (model_dtype, grad_accum_dtype)
 
+    def _mixed_precision_dtypes(self):
+        """Resolve (param_dtype, buffer_dtype) for the module cast.
+
+        param_dtype: config override, else the fp16/bf16 enabled flag, else None.
+        buffer_dtype: config override, else None (buffers keep their loaded dtype).
+        """
+        if self._config.param_dtype is not None:
+            param_dtype = DtypeEnum(self._config.param_dtype).value
```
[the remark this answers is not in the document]
sfc-gh-truwase: Good catch. Fixed in 5d4caad1 by rejecting a `data_types.param_dtype` that disagrees with the enabled fp16/bf16 mode, rather than letting it diverge.

A new `_assert_valid_mixed_precision_config()` is called from `_do_sanity_check()` (config-only, so it fails before any module cast or optimizer construction). `param_dtype` now follows the enabled mode and, when set, must match it; `_mixed_precision_dtypes()` is back to a pure resolver. Added unit tests covering accept (matching/unset) and reject (`fp16`/`fp32` vs bf16, `bf16` vs fp16/neither).

deepspeed/runtime/config.py
```diff
@@ -860,6 +860,10 @@ def _initialize_params(self, param_dict):
 
         data_types_params = get_data_types_params(param_dict)
         self.grad_accum_dtype = data_types_params.get(GRAD_ACCUM_DTYPE, GRAD_ACCUM_DTYPE_DEFAULT)
+        # Raw strings ("bf16"/"fp16"/"fp32") or None; resolved via DtypeEnum at
+        # use-time. buffer_dtype=None keeps buffers at their loaded dtype.
+        self.param_dtype = data_types_params.get(PARAM_DTYPE, PARAM_DTYPE_DEFAULT)
+        self.buffer_dtype = data_types_params.get(BUFFER_DTYPE, BUFFER_DTYPE_DEFAULT)
```
stas00, line 867:
```suggestion
        # Raw strings ("bf16"/"fp16"/"fp32") or None; resolved via DtypeEnum at
        # use-time.
        self.param_dtype = data_types_params.get(PARAM_DTYPE, PARAM_DTYPE_DEFAULT)
        # buffer_dtype=None keeps buffers at their loaded dtype.
        self.buffer_dtype = data_types_params.get(BUFFER_DTYPE, BUFFER_DTYPE_DEFAULT)
```

stas00 (commented):
probably should add these to the config md in the docs, no?

PKUWZP:
@sfc-gh-truwase Can you check the formatting by running "pre-commit" and fix any formatting issues? Thanks for the PR!

deepspeed/runtime/constants.py
```diff
@@ -466,12 +466,23 @@ class ValidationMode:
...
 #   grad_accum_dtype=["bf16"|"fp16"|"fp32"]
+#   param_dtype=["bf16"|"fp16"|"fp32"]
+#   buffer_dtype=["bf16"|"fp16"|"fp32"]
 #   }
 # }
+# param_dtype / buffer_dtype mirror FSDP MixedPrecisionPolicy. When unset (None):
+#   - param_dtype follows the enabled bf16/fp16 mode; if set it must match that
+#     mode (it is validated, not an independent override).
+#   - buffer_dtype keeps each buffer's loaded dtype (e.g. fp32 rotary inv_freq).
+#     Set it to a concrete dtype to force-cast buffers (legacy blanket cast).
```
stas00:
```suggestion
# param_dtype and buffer_dtype mirror FSDP's MixedPrecisionPolicy.
#   - param_dtype: if None uses the specified mixed precision dtype, otherwise casts the params into the provided dtype
#   - buffer_dtype: if None uses the buffers' dtype found when the model was loaded (e.g. fp32 rotary inv_freq), otherwise casts the buffers into the provided dtype (which is likely to lead to unintended consequences)
```

PKUWZP (approved):

PR #8068, opened 2026-06-15
Default gradient_clipping to 1.0

sfc-gh-truwase:
## Summary
- Change `GRADIENT_CLIPPING_DEFAULT` from `0.` (disabled) to `1.0`.

## Motivation
With the old default, configs that omit `gradient_clipping` run unclipped. Most RL/LLM training (and the FSDP2 reference) clip at `1.0`; this avoids silently-unclipped runs. Isolated into its own PR since it is a default behavior change.

## Test plan
- [ ] Init without `gradient_clipping` -> effective clip norm is `1.0`.
- [ ] Explicit `gradient_clipping: 0.0` still disables clipping (override respected).

### DIFF
--- a/deepspeed/runtime/constants.py
+++ b/deepspeed/runtime/constants.py
@@ -251,7 +251,7 @@ Gradient clipping should be enabled as:
 "gradient_clipping": 1.0
 '''
 GRADIENT_CLIPPING = 'gradient_clipping'
-GRADIENT_CLIPPING_DEFAULT = 0.
+GRADIENT_CLIPPING_DEFAULT = 1.0
 
 #########################################
 # Capture graph for short kernels sequences
--- a/deepspeed/runtime/constants.py
+++ b/deepspeed/runtime/constants.py
@@ -244,8 +244,8 @@ TORCH_AUTOCAST_LOWER_PRECISION_SAFE_MODULES = "lower_precision_safe_modules"
 #########################################
 # Gradient clipping
 #########################################
-# Gradient clipping. By default, this feature is not enabled.
-# Users can configure in ds_config.json as below example:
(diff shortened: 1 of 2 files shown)


deepspeed/runtime/constants.py
tohtana: Can we match this comment with the change?

deepspeed/runtime/constants.py
```diff
@@ -251,7 +251,7 @@
 "gradient_clipping": 1.0
 '''
 GRADIENT_CLIPPING = 'gradient_clipping'
-GRADIENT_CLIPPING_DEFAULT = 0.
+GRADIENT_CLIPPING_DEFAULT = 1.0
```
tohtana, line 254: We should add a config test for this behavior: when `gradient_clipping` is omitted, the effective value should be `1.0`, and when `gradient_clipping: 0.0` is specified explicitly, clipping should remain disabled.

tohtana (changes requested):
Thank you Tunji for the update. I agree that we should enable gradient clipping by default. I left a few comments.

deepspeed/runtime/constants.py
sfc-gh-truwase: Done.

PKUWZP:
@sfc-gh-truwase Thanks for the fix Tunji! Can I know why we want to hard-code gradient clipping to be 1.0?

sfc-gh-truwase:
> @sfc-gh-truwase Thanks for the fix Tunji! Can I know why we want to hard-code gradient clipping to be 1.0?

This seems to be standard expectation from modeling folks that I have interacted with. It is also parity with FSDP and Megatron-LM

tohtana (approved):
Thank you for the update!

PR #8080, opened 2026-06-19
ZeRO 1/2: wait on all IPG-bucket producer streams in average_tensor (#8061)

### LINKED ISSUE #8061: [BUG] ZeRO stage 1/2 overlap_comm only waits current stream, but contiguous gradient bucket copies may come from multiple streams
## Describe the bug

On current `master` (`b2145896daeddefcb70d9301c04c4d0a03b71bb6` as of 2026-06-12), the ZeRO stage 1/2 overlap communication path appears to assume that the stream current at `average_tensor()` time is the same stream that produced all contiguous gradient bucket writes.

This assumption can be false when the gradient hooks are triggered from different autograd/compiled-backward streams. In a profiler trace, the `copy_` operations that write parameter gradients into the IPG buffer are distributed across multiple device streams, while `average_tensor()` only makes the reduction stream wait for `get_accelerator().current_stream()` at the time `average_tensor()` is called.

As a result, `reduction_stream` may start reading the IPG buffer (`div_`, `narrow`, `cat/concat`, all-reduce/scatter path) before all streams that wrote slices of that same bucket have completed.

## Relevant code on master

The bucket copy is done in `reduce_independent_p_g_buckets_and_remove_grads()`:

https://github.com/deepspeedai/DeepSpeed/blob/b2145896daeddefcb70d9301c04c4d0a03b71bb6/deepspeed/runtime/zero/stage_1_and_2.py#L1119-L1122

```python
new_grad_tensor = bucket.buffer[bucket.index].narrow(0, bucket.elements, param.numel())
new_grad_tensor.copy_(
    grad_reduc.view(-1), non_blocking=self.device == get_accelerator().device_name())
```

The bucket is later reduced here:

https://github.com/deepspeedai/DeepSpeed/blob/b2145896daeddefcb70d9301c04c4d0a03b71bb6/deepspeed/runtime/zero/stage_1_and_2.py#L1575

```python
self.average_tensor(bucket.buffer[bucket.index].narrow(0, 0, bucket.elements), comm_dtype)
```

But `average_tensor()` only waits for the current stream at the moment it is called:

https://github.com/deepspeedai/DeepSpeed/blob/b2145896daeddefcb70d9301c04c4d0a03b71bb6/deepspeed/runtime/zero/stage_1_and_2.py#L1230-L1238

```python
def average_tensor(self, tensor: torch.Tensor, communication_data_type: torch.dtype):
    if self.overlap_comm:
        stream = self.reduction_stream
        if not get_accelerator().resolves_data_dependency():
            stream.wait_stream(get_accelerator().current_stream())
            get_accelerator().current_stream().wait_stream(stream)
    else:
        stream = get_accelerator().current_stream()
```

This is not sufficient if the same `bucket.buffer[bucket.index]` was filled by `copy_` calls issued on multiple earlier streams.

## Observed behavior

Configuration characteristics:

- ZeRO stage 1/2 code path
- `overlap_comm: true`
- `contiguous_gradients: true`
- `reduce_scatter: true`
- bf16 training
- `torch.compile` enabled

Symptoms:

- loss becomes `NaN` very early, starting from step 1 in our run
- disabling `torch.compile` avoids the issue
- disabling `overlap_comm` avoids the issue
- adding a device-wide synchronize in the DeepSpeed reduction path avoids the issue, but is too heavy
- adding stream synchronization from the actual gradient-copy streams to `reduction_stream` also avoids the issue

Profiler trace observation:

- the gradient bucket `copy_` operations corresponding to the DeepSpeed contiguous-gradient path are emitted on more than one device stream
- later, `average_tensor()` / concat / reduction reads the same IPG bucket on the reduction stream
- the existing wait only covers the stream returned by `get_accelerator().current_stream()` when `average_tensor()` is entered, not all streams that previously wrote slices into the bucket

Conceptually, this can happen as:

```text
stream A: copy grad slice A -> ipg bucket
stream B: copy grad slice B -> ipg bucket
stream C: calls average_tensor(); current_stream == stream C

DeepSpeed today:
  reduction_stream waits stream C

Missing dependency:
  reduction_stream should also wait stream A and stream B before reading the bucket
```

## Why this is different from existing overlap_comm fixes

This is related to, but not fully covered by, previous fixes:

- #5606 fixed the bidirectional wait between current/default stream and `reduction_stream` in `average_tensor()`.
- #7371 / #7188 addressed a synchronization issue around double IPG-buffer swapping.

The remaining issue here is narrower: even on current master, `average_tensor()` still assumes that one current stream represents all producers of the current contiguous gradient bucket. With compiled autograd / torch.compile, gradient hooks can run under different current streams, so the bucket can have multiple producer streams.

## Possible fix direction

One conservative fix is to record the unique streams used for the IPG bucket copy operations, then make `reduction_stream` wait for those streams before `average_tensor()` reads the bucket.

Sketch:

```python
# after new_grad_tensor.copy_(...)
self._record_ipg_copy_stream(bucket)

# inside average_tensor(), after switching to reduction_stream
self._wait_for_ipg_copy_streams(bucket)
```

Where `_record_ipg_copy_stream()` stores unique `get_accelerator().current_stream()` values for the current IPG bucket, and `_wait_for_ipg_copy_streams()` calls `current_stream.wait_stream(copy_stream)` for each recorded stream not already covered by the existing `stream.wait_stream(current_stream)`.

This avoids a device-wide synchronize and only adds stream dependencies for actual producer streams of the bucket.

## Expected behavior

`overlap_comm: true` with `contiguous_gradients: true` should not allow the reduction stream to read the IPG bucket before all streams that wrote gradient slices into that bucket have completed.

arunshar:
## What

Fixes #8061. In ZeRO stage 1/2 with `overlap_comm`, `average_tensor` waits the reduction stream on only the **current** stream before reducing the contiguous IPG gradient bucket:

```python
if self.overlap_comm:
    stream = self.reduction_stream
    if not get_accelerator().resolves_data_dependency():
        stream.wait_stream(get_accelerator().current_stream())   # only one stream
```

But the per-parameter gradient copies that fill the bucket (`reduce_independent_p_g_buckets_and_remove_grads`, the `new_grad_tensor.copy_(...)` into `bucket.buffer[bucket.index]`) can be issued on **multiple** streams. That is exactly the scenario #8061 reports under `torch.compile`, where gradient hooks run on different autograd streams and several device streams write slices into the same IPG buffer. Waiting on only the current stream lets the all-reduce read the bucket before the other producers finish, corrupting gradients (NaN loss from step 1).

## Change

Implements the fix direction proposed in #8061 (record the streams used for the IPG bucket copies, then make the reduction stream wait on all of them):

- `IPGBucket` gains a `copy_streams` set, cleared in `clear()`.
- After each copy into `bucket.buffer[bucket.index]`, record `current_stream()` (overlap path only).
- `average_tensor` waits the reduction stream on **every** recorded producer stream, falling back to `current_stream()` when the set is empty (e.g. the extra-large-param path that reduces without a bucket copy).

The single-stream case is unchanged: when all copies are on one stream, `copy_streams == {current_stream}`, so the wait is identical to before, and there is no behavior change for the common path.

## Tests / validation

- New CPU unit tests in `tests/unit/v1/zero/test_overlap_comm_record_stream.py` (fakes + monkeypatch, no GPU): the reduction stream waits on all producer streams, the empty-set fallback to `current_stream`, and `IPGBucket.copy_streams` reset. The two pre-existing tests in that file still pass (5 passed).
- `pre-commit run --files` is green on both changed files (yapf, flake8, check-torchdist, check-license, codespell).
- No-regression A/B on 2x A40 (seeded, identical data, ZeRO-2 + `contiguous_gradients`, overlap_comm on vs off, eager and `torch.compile`): before and after this change, suspect == baseline on every repeat with byte-identical final-param hashes, and the baseline-vs-baseline determinism gate passes. So the fix does not change results when grad-bucket copies stay on a single stream.

### Honest scope on reproduction

I could not deterministically trigger the multi-stream NaN on the available hardware (A40, small MLP): neither the `torch.compile` A/B nor a synthetic two-stream microbenchmark surfaced it (PyTorch's caching allocator inserts implicit cross-stream syncs that mask the race in a microbenchmark, and a plain `torch.compile(model)` on this model kept the grad-bucket copies on one stream). This PR is therefore offered as the minimal correct synchronization for the clearly-missing producer-stream wait identified in `average_tensor`, validated for no-regression and with unit coverage; a reviewer with the original `torch.compile` multi-stream repro from #8061 can confirm the NaN is resolved.

Opened as a draft.


tohtana (approved):
This looks good to me. Thank you @arunshar for this important fix!

PKUWZP (changes requested):
Thanks for the fix! There are several suggestions:  

  1. The extra-large-param path keeps the original bug (minor, scope). When a param exceeds `reduce_bucket_size`, it goes through `extra_large_param_to_reduce` and is reduced without a bucket copy, so `copy_streams` is empty and the code falls back to waiting only on `current_stream()` (line 1572's `average_tensor`). If that path can also be hit with multi-stream producers under `torch.compile`, the same race remains. This matches old behavior, I think it's worth a one-line comment or an issue note that the fix is properly scoped to the bucketed path.

  2. Efficacy is unverified — and there's a `torch.compile`-specific caveat. The `NaN` couldn't be reproduced. One concrete risk worth flagging for the #8061 reporter who validates: `get_accelerator().current_stream()` is queried in Python at hook-execution time. If the gradient hook + copy are captured into a CUDA graph / compiled region, the Python-observed current stream may not correspond to the stream the copy actually executes on at replay, in which case the recorded set could be incomplete. This isn't a reason to block, but the validation should specifically confirm the producer streams seen here match the real execution streams.

  3. Tests verify the wiring, not the race (acknowledged). The CPU fakes confirm "wait on all producers," the empty-set fallback, and `clear()` reset — good unit coverage for the host-side logic. But by construction they can't exercise the actual cross-stream ordering. That's an inherent limitation of a no-GPU test; fine to land, just don't read the green tests as confirmation the NaN is fixed.

PKUWZP:
Really appreciate the fix @arunshar , just left some comments. If you can provide some clarification, we can merge the PR shortly. Thanks again for your help!

arunshar:
Thanks @PKUWZP, good points. Taking them in order:

1. **Extra-large-param path:** You're right that it keeps the old behavior. I pushed 60f0193b, adding a scope comment at the extra_large_param_to_reduce branch: those params are reduced directly and never copied into the contiguous IPG bucket, so no producer stream is recorded for them, and average_tensor falls back to waiting on the current stream. The producer-stream tracking only covers the bucketed path. If that path turns out to hit multi-stream producers under torch.compile it needs the same treatment, but that is a separate change, and I kept this PR scoped to the bucket race in #8061.

2. **torch.compile / CUDA-graph caveat:** Agreed, and worth flagging for whoever validates against the original repro. current_stream() is read in Python at hook-execution time, so if the hook and the copy get captured into a CUDA graph or a compiled region, the Python-observed stream may not be the stream the copy actually runs on at replay, and the recorded set could be incomplete. The validation should confirm that the producer streams recorded here match the real execution streams. If they don't under capture, the fix would need to record the stream at the device level rather than from Python.

3. **Tests:** Right, the CPU fakes only check the host-side wiring (wait on all producers, the empty-set fallback, and the clear() reset). They cannot exercise the real cross-stream ordering, so a green run here does not evidence that the NaN is gone. I left that explicit in the PR description; confirmation needs the original torch.compile multi-stream repro on GPU.

arunshar:
Quick update: CI is green now, including DCO (I'd missed the sign-off on the scope-comment commit; fixed with `--signoff`). All three points from your review are addressed: (1) the extra-large-param path is documented as out of scope in the latest commit, (2) the torch.compile / CUDA-graph stream-capture caveat is flagged for whoever runs the original repro, and (3) agreed that the CPU tests cover the host-side wiring rather than the cross-stream race. Happy to make any further changes you'd like. Thanks again for the careful review, @PKUWZP.

tohtana:
@PKUWZP Can we merge this PR now?

arunshar:
Heads-up on the branch history: I reverted the Copilot Autofix I'd applied earlier. It changed the wait-set to always union in `current_stream`, which contradicted the existing unit test (it asserts the reduction waits on exactly the recorded producer streams) and added a redundant wait, since `copy_streams` already tracks the actual bucket writers. The branch is back to the reviewed logic; CI is re-running now and the rest of the suite is passing. Thanks @tohtana @PKUWZP.

PKUWZP (approved):
Thanks for addressing all my comments, I approve this PR.

PR #8094, opened 2026-06-26
Fix DeepCompile profile metadata backfill

tohtana:
DeepCompile ZeRO-3 scheduling can crash with KeyError or AssertionError on missing device_time when graph profiling fails partway and leaves some nodes without timing metadata.

This PR backfills default timing and memory metadata after profiling cleanup so the profiler guarantees downstream passes see complete metadata on every node.

### DIFF
--- a/deepspeed/compile/passes/selective_gather.py
+++ b/deepspeed/compile/passes/selective_gather.py
@@ -15,6 +15,7 @@ from deepspeed.utils import log_dist
 
 from ..util import get_deepcompile_handle
 from ..graph_param import DSGraphParamManager
+from ..profilers.graph_profile import is_profile_incomplete
 
 NAME = "selective_gather"
 
@@ -66,6 +67,10 @@ def _compute_persistence_budget(all_graph_mem_records: List[List[Tuple[str, int,
     }
 
 
+def _profile_result_incomplete(prof) -> bool:
+    return is_profile_incomplete(prof.fwd_graph) or is_profile_incomplete(prof.bwd_graph)
+
+
 def selective_gather(gm: GraphModule, graph_id: int, graph_order: List[Tuple[int, bool]], profiling_results,
                      create_inputs_fn, mem_budget: float, param_manager: DSGraphParamManager,
                      bwd: bool) -> GraphModule:
@@ -82,6 +87,14 @@ def selective_gather(gm: GraphModule, graph_id: int, graph_order: List[Tuple[int
 
     # Run only on the last backward graph
     if last_backward_graph_id is None or graph_id != last_backward_graph_id:
+        return gm
+
+    incomplete_profile_ids = [
+        profile_graph_id for profile_graph_id, prof in profiling_results.items() if _profile_result_incomplete(prof)
+    ]
+    if incomplete_profile_ids:
+        print_rank_0(f"selective_gather incomplete profiling data for graph_ids={incomplete_profile_ids}; "
+                     "skipping persistence update")
         return gm
 
     all_graph_mem_records = []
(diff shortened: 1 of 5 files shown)


deepspeed/compile/profilers/graph_profile.py
```diff
@@ -54,6 +54,21 @@ def _node_size(out):
     return sum([v.element_size() * v.numel() for v in tree_leaves(out) if torch.is_tensor(v)])
 
 
+_PROFILE_META_DEFAULTS = {
+    "device_time": 0.0,
+    "wall_time": 0.0,
+    "tensor_size": 0,
```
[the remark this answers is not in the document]
tohtana: Addressed

deepspeed/compile/passes/selective_gather.py
```diff
@@ -117,12 +129,12 @@ def selective_gather(gm: GraphModule, graph_id: int, graph_order: List[Tuple[int
             for n in profile.bwd_graph.nodes:
                 if n.target == torch.ops.dc.allgather_param.default:
                     assert "tensor_size" in n.meta
-                    ds_id_to_size[n.args[2]] = n.meta["tensor_size"]
+                    _maybe_update_size_from_profile(ds_id_to_size, n.args[2], n.meta["tensor_size"])
                     assert "device_time" in n.meta
                     ds_id_to_time[n.args[2]] += n.meta["device_time"]
```
sfc-gh-truwase, line 147: Don't you need similar guard for `n.meta["device_time"]`? Or is it okay because default value is `0.0`?

deepspeed/compile/profilers/graph_profile.py
```diff
@@ -110,10 +125,16 @@ def run(self, *args) -> Any:
                     return_val = super().run(*args)
         except Exception as e:
             msg = e.msg if "msg" in dir(e) else str(e)
-            print(f"Profiling error {msg}")
+            if not self.distributed or dist.get_rank() == 0:
+                print(f"DeepCompile profiling failed; using default profile metadata for incomplete nodes: {msg}")
```
sfc-gh-truwase, line 153: Can we use logger instead of raw `print`?

deepspeed/compile/profilers/graph_profile.py
```diff
@@ -110,10 +125,16 @@ def run(self, *args) -> Any:
...
         finally:
-            self.nz3.clear_all_gathered_params()
-            self.nz3.enable_profiling(False)
+            try:
+                self.nz3.clear_all_gathered_params()
+            finally:
+                try:
+                    self.nz3.enable_profiling(False)
+                finally:
+                    _backfill_missing_profile_metadata(self.graph)
```
sfc-gh-truwase: I am struggling to understand what is happening here. Can you add a comment?

sfc-gh-truwase (approved):

deepspeed/compile/passes/selective_gather.py
tohtana: Yes. The backfill uses `0.0` only when profiling did not record a `device_time`.
This is different from `tensor_size` because size has an independent fallback from `param_manager`, so a backfilled `0` must not replace that known parameter size. For `device_time`, there is no equivalent fallback value,.

deepspeed/compile/profilers/graph_profile.py
tohtana: Good catch, replaced it with `log_dist(..., ranks=[0], level=logging.WARNING)`.

deepspeed/compile/profilers/graph_profile.py
tohtana: Good point. The nested cleanup was intended to make sure profiling is disabled even if clearing gathered params fails. But I simplified the block as `enable_profiling` should be raised to a caller rather than caught here. I also added a comment to clarify it.

PR #8106, opened 2026-07-01
Fix DeepCompile profiling memory cleanup

tohtana:
This PR reduces DeepCompile ZeRO-3 memory pressure in profiling and compiled backward by fixing three related issues:

- Let the partitioner keep using its min-cut rematerialization policy for non-parameter activations, while still forcing ZeRO-3 parameter aliases/casts to be recomputed.
- Compute selective-gather persistence capacity from profiled transient headroom instead of current allocator availability.
- Avoid retaining temporary profiling outputs across warmup/measured profiling iterations, and clear gathered ZeRO-3 parameters when profiling exits through an exception.
- Mark incomplete memory profiles explicitly, synchronize the completion status across ranks, and skip profile-dependent prefetch/selective-gather decisions when profiling data is incomplete.

### DIFF
--- a/deepspeed/compile/backend.py
+++ b/deepspeed/compile/backend.py
@@ -21,6 +21,7 @@ try:
 except ImportError:
     pass
 
+import deepspeed.comm as dist
 from deepspeed.accelerator import get_accelerator
 
 from .fx import add_free_activations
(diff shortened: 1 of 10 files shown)


PKUWZP (approved):
LGTM, approved.

PR #8110, opened 2026-07-01
Make DCO workflow Probot compatible

tohtana:
The current merge-queue-required DCO check does not match Probot DCO behavior. As a result, the required GitHub Actions DCO check can remain failing even after maintainers resolve the DCO status through the Probot DCO UI.

This PR makes the merge-queue-compatible required DCO workflow consistent with Probot DCO behavior: it skips merge commits and verified bot-authored commits, and it accepts a successful official Probot DCO check on the same PR head SHA.

It also renames the GitHub Actions check to `DCO / required`, making it easier to distinguish the required merge-queue check from the Probot `DCO` check.

### DIFF
--- a/.github/workflows/dco.yml
+++ b/.github/workflows/dco.yml
@@ -3,7 +3,7 @@
 
 # DeepSpeed Team
 
-name: DCO
+name: DCO / required
 
 on:
   pull_request:
@@ -14,12 +14,13 @@ on:
       - master
 
 permissions:
+  checks: read
   contents: read
   pull-requests: read
 
 jobs:
-  DCO:
-    name: DCO
+  dco_required:
+    name: DCO / required
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/dco.yml
```diff
@@ -181,6 +193,42 @@ jobs:
...
+                  f"/commits/{ref}/check-runs?check_name=DCO&filter=latest",
+                  token,
+                  fatal=False,
+              )
+              if not payload:
+                  return False
+
+              for check_run in payload.get("check_runs", []):
+                  app = check_run.get("app") or {}
+                  is_probot_dco = app.get("slug") == "dco" or app.get("id") == 1861
```
delock: what is 1861?  May need a comment

delock (approved):

.github/workflows/dco.yml
tohtana: Thank you @delock! `1861` is the GitHub App ID for the official [DCO app](https://github.com/apps/dco). I named it `PROBOT_DCO_APP_ID` and added the source comment in [`c5dea3ed`](https://github.com/deepspeedai/DeepSpeed/commit/c5dea3ed3cb8fa2ca2ed14db3e3f5af94a577f1a).

PR #8125, opened 2026-07-07
fix(autotp): when using autotp, ignore the consistency of certain data within the tp_group.

1787648106:
When using `autotp`, the consistency check for the KV cache within the TP group is bypassed. Recent popular methods, such as on-policy distillation, require the model to generate output during training, which would otherwise cause the data consistency check to fail.


delock:
@1787648106  thanks for your fix.   I agree with codex that the import from transformers should be catched.  If there is import error, then the tensor won't be HF model KV cache.  This would avoid break existing use cases.

Also can you sign-off your commits to pass DCO check?  Thanks!

1787648106:
> @1787648106 thanks for your fix. I agree with codex that the import from transformers should be catched. If there is import error, then the tensor won't be HF model KV cache. This would avoid break existing use cases.
> 
> Also can you sign-off your commits to pass DCO check? Thanks!

Done, thank you!

delock (approved):

delock:
Hi @1787648106 can you fix the formatting errors?  You may want to refer to this link for essential pre CI checks.   Thanks!
https://github.com/deepspeedai/DeepSpeed/blob/master/AGENTS.md#commit--ci-requirements

1787648106:
> Hi @1787648106 can you fix the formatting errors? You may want to refer to this link for essential pre CI checks. Thanks! https://github.com/deepspeedai/DeepSpeed/blob/master/AGENTS.md#commit--ci-requirements

done

deepspeed/runtime/utils.py
```diff
@@ -1185,6 +1185,16 @@ def move_back_key(state, key):
             move_back_key(state, "exp_avg_sq")
 
 
+def is_transformers_cache(obj):
+    """Skip checks for the `transformers` cache class when performing AutoTP tensor comparisons."""
+    try:
+        from transformers.cache_utils import (DynamicCache, StaticCache, QuantizedCache)
```
tohtana: The `transformers` versions supporting these cache classes are different. So `is_transformers_cache(dynamic_cache)` can return False. Also, we have cache classes not listed here, e.g. `SinkCache`. 
Do you think we can check this by the parent class like:

```python
def is_transformers_cache(obj):
    try:
        from transformers.cache_utils import Cache
    except ImportError:
        return False

    return isinstance(obj, Cache)
```

tohtana (changes requested):
Hi @1787648106,
Thank you for submitting this PR, great catch! I left a comment below. Happy to hear you thought.

deepspeed/runtime/utils.py
1787648106: I agree with you. Thank you!

deepspeed/runtime/utils.py
```diff
@@ -1203,6 +1213,8 @@ def compare_tensors_in_structures(inputs1: Union[List, Dict], inputs2: Union[Lis
         if len(inputs1) != len(inputs2):
             return False
         for val1, val2 in zip(inputs1, inputs2):
+            if is_transformers_cache(val1) and is_transformers_cache(val2):
```
tohtana, line 1216: Can we make sure the both classe are the same?
For example,
```
            if is_transformers_cache(val1) and is_transformers_cache(val2):
                if type(val1) is not type(val2):
                    return False
                continue
```

tohtana (changes requested):
Thank you for updating this PR! Can you check the following comment?

PKUWZP (approved):
LGTM.

tohtana (approved):

PR #8073, opened 2026-06-17
Fix ZeRO-3: Use per-param dtype for output buffers in _allgather_params_coalesced

albertvillanova:
This PR fixes the `_allgather_params_coalesced` method in `partition_parameters.py`. The change ensures that each `flat_tensor` is created with the correct data type by referencing the corresponding parameter in `param_list`, rather than always using the first parameter's data type.

Fix #8072.

### Problem

`_allgather_params_coalesced` allocates all output buffers using the dtype of the first parameter in param_list:

```python
# before
for psize in partition_sizes:
    flat_tensor = torch.empty(tensor_size, dtype=param_list[0].ds_tensor.dtype, ...)
```

This assumed every persistent parameter shares the same dtype. The assumption was incidentally maintained before 0.19.2 because `_configure_distributed_model` called `module.bfloat16()` unconditionally, normalising all persistent parameters (including PEFT LoRA adapters) to a uniform dtype.

PR https://github.com/deepspeedai/DeepSpeed/pull/8066 "Mixed-precision: per-policy param/buffer dtype cast (preserve fp32 buffers)" (commit b919284a) correctly stopped casting ZeRO-Init model params, but exposed the latent bug: PEFT's default `autocast_adapter_dtype=True` keeps LoRA adapters in fp32 even when the base model is bf16. `persistent_parameters` therefore ends up with mixed dtypes (bf16 base-model params + fp32 LoRA params), and the mismatch between a bf16 output buffer and a fp32 input tensor raises:
> TypeError: output tensor must have the same type as input tensor

- Reported in: https://github.com/deepspeedai/DeepSpeed/issues/8072
- Downstream tracking: https://github.com/huggingface/trl/issues/6089

### Solution

Allocate each output buffer with the dtype of its own parameter:

```python
# after
for i, psize in enumerate(partition_sizes):
    flat_tensor = torch.empty(tensor_size, dtype=param_list[i].ds_tensor.dtype, ...)
```

This removes the shared-dtype assumption at the source rather than relying on upstream callers to normalise dtypes before calling `_allgather_params_coalesced`.

### Changes

- Corrected tensor data type selection in `_allgather_params_coalesced` to use the data type of each parameter in `param_list`, ensuring proper handling of mixed data types.

Issue #8072, opened 2026-06-17
[BUG] ZeRO-3 + PEFT (LoRA) regression in 0.19.2: TypeError: output tensor must have the same type as input tensor in _allgather_params_coalesced


albertvillanova:
**Describe the bug**
DeepSpeed 0.19.2 introduced a regression that breaks training with ZeRO Stage 3 + PEFT (LoRA) when the base model uses bfloat16. A dtype mismatch error is raised in `_allgather_params_coalesced` during `_post_step` when all-gathering persistent parameters after the optimizer step.

Tracking issue on TRL: 
- https://github.com/huggingface/trl/issues/6089

**To Reproduce**
1. Use a base model with native `torch_dtype = bfloat16` (e.g. Qwen2, Llama 3)
2. Wrap it with a PEFT LoRA adapter (standard, non-quantized, default `autocast_adapter_dtype=True`)
3. Train with ZeRO Stage 3 and bf16 enabled
4. Error is raised on the first optimizer step

Minimal reproduction using the TRL SFT script:
```shell
accelerate launch --config_file zero3.yaml trl/scripts/sft.py \
    --model_name_or_path Qwen/Qwen2.5-0.5B \
    --dataset_name trl-internal-testing/zen \
    --dataset_config standard_language_modeling \
    --use_peft \
    --output_dir /tmp/sft-zero3-peft
```
zero3.yaml:
```yaml
distributed_type: DEEPSPEED
deepspeed_config:
  zero_stage: 3
  bf16:
    enabled: true
num_processes: 2
```

**Expected behavior**
Training should complete successfully, as it did with DeepSpeed 0.19.1.

**Error traceback**
```python
[rank0]: Traceback (most recent call last):
...
[rank0]:   File "/__w/trl/trl/.venv/lib/python3.12/site-packages/accelerate/utils/deepspeed.py", line 281, in backward
[rank0]:     self.engine.step()
[rank0]:   File "/__w/trl/trl/.venv/lib/python3.12/site-packages/deepspeed/runtime/engine.py", line 3075, in step
[rank0]:     self._take_model_step(lr_kwargs)
[rank0]:   File "/__w/trl/trl/.venv/lib/python3.12/site-packages/deepspeed/runtime/engine.py", line 2970, in _take_model_step
[rank0]:     self.optimizer.step()
[...]

PR #8128, opened 2026-07-09
Fix ZeRO-3: use per-param dtype for quantize-scale buffer in _allgather_params_coalesced

roycho96:
## Problem

#8073 fixed `_allgather_params_coalesced` to allocate each output buffer with its own parameter's dtype, resolving the mixed-dtype crash under ZeRO-3 (bf16 base + fp32 LoRA adapters, #8072).

The quantize-scale buffer in the same method was left on the old pattern. It still uses the first parameter's scale dtype:

```python
for psize in quantize_scale_sizes:
    param = param_list[len(allgather_quantize_scale)]
    tensor_size = psize * self._partition_world_size(param)
    flat_tensor = torch.empty(tensor_size,
                              dtype=param_list[0].ds_tensor.ds_quant_scale.dtype,
                              device=self.local_device).view(-1)
```

Each scale buffer is gathered by its own `all_gather_into_tensor` call using that param's own scale tensor, so a persistent set with mixed scale dtypes hits the same "output tensor must have the same type as input tensor" error that #8073 removed for the main buffers.

## Fix

The loop already binds `param` and uses it for `tensor_size`, so use it for the scale buffer dtype as well:

```python
    dtype=param.ds_tensor.ds_quant_scale.dtype,
```

This matches the main output buffer (fixed in #8073) and the sequential path `_allgather_params_sequential`, which already uses `param.ds_tensor.ds_quant_scale.dtype`.

No cast is added, so numerics are unchanged, and it is a no-op when the scale dtypes are uniform. The trigger is narrow (quantized weights with mixed scale dtypes), so this is a consistency fix that removes the last shared-dtype assumption in the method rather than a widely hit crash.

### DIFF
--- a/deepspeed/runtime/zero/partition_parameters.py
+++ b/deepspeed/runtime/zero/partition_parameters.py
@@ -1959,7 +1959,7 @@ class Init(InsertPostInitMethodToModuleSubClasses):
                 param = param_list[len(allgather_quantize_scale)]
                 tensor_size = psize * self._partition_world_size(param)
                 flat_tensor = torch.empty(tensor_size,
-                                          dtype=param_list[0].ds_tensor.ds_quant_scale.dtype,
+                                          dtype=param.ds_tensor.ds_quant_scale.dtype,
                                           device=self.local_device).view(-1)
                 flat_tensor.requires_grad = False
                 allgather_quantize_scale.append(flat_tensor)


tohtana (approved):
Hi @roycho96,
Thank you for the great fix! It looks good to me.
I actually ran into this issue myself, so I really appreciate your continued contributions.

roycho96:
> Hi @roycho96, Thank you for the great fix! It looks good to me. I actually ran into this issue myself, so I really appreciate your continued contributions.

Thanks @tohtana, means a lot. Been a bit swamped lately but I'll keep them coming.

Issue #7693, opened 2025-11-14
Tracking excessive cpu memory usage in z2 cpu offload


stas00:
I have observed about `numel * 28 bytes` cpu memory requirements for doing z2+optim cpu offload.

Here is the mapping out of current cpu memory allocations when offload_optimizer.device: cpu is used with zero: stage 2

Ideally it should be `numel * 16 bytes` (4 bytes per master weights, grad, 2x optim states).

The total of cpu memory allocation happen in 2 parts `deepspeed.initialize` + first `step` call - I'm going to break it down into each per param allocation.

## part 1: deepspeed.initialize
1. master weights 4 bytes - kosher
2. intermediary H2D copy buffer (for speed) - 2 bytes (half precision) - seems we can't avoid this one
3. related to above - 2 byte pinned memory overhead - fixed it here https://github.com/deepspeedai/DeepSpeed/pull/7689 
4. initialize_optimizer_states - creates 4 bytes for grads - kosher

total 10 bytes per param

we confirmed empirically that's the case.

## part 2: first step
1. 8 bytes for optim states - kosher
2. unscale_and_clip_grads - adds 4 bytes - but we think it's peak cpu memory - because grad.data.mul_(1. / combined_scale) shouldn't allocate more memory other than for a temp buffer in pytorch - it doesn't appear to be a temp buffer not released by linux cpu memory - because the next allocation doesn't re-use it and allocates its own full sized tensor
3. self.ds_opt_adam.adam_update for some reason allocates 4 bytes more

total: 16 bytes per param

grand total: 26 bytes

plus additional ~2 bytes I still don't have an account for

Details:

- this leaks 4 bytes per param (inside the first `step` call)
```
        for grad in grad_groups_flat:
            # checking cpu memory before and after the next call shows an additional fp32 
            # allocation on the first call of unscale_and_clip_grads, on subsequent calls it's stable
            grad.data.mul_(1. / combined_scale) #
```
I even tried to move it to gpu do the mul there and move back to cpu I still get the leak.

```
               rank = torch.distributed.get_rank()
                device = torch.device(f"cuda:{rank}")

                t = 1. / combined_scale
                see_memory_usage(f"before", force=True)

                grad = grad.to(device)
                t = t.to(device)
                see_memory_usage(f"after to cuda", force=True)

                grad.mul_(t)
                see_memory_usage(f"after mul_", force=True)

                grad = grad.cpu()
                see_memory_usage(f"after to cpu", force=True)
```
[...]

PR #8132, opened 2026-07-11
fix(zero-offload): mitigate CPU RSS waste on ZeRO-2 optimizer offload (#7693)

liuyun7345:
## Summary

Best-effort mitigations for CPU RSS waste in ZeRO-2 + CPU optimizer offload tracked by #7693.

This PR does **not** claim a complete resolution of all overhead in that issue (H2D cast/copy buffer tiling and other unaccounted bytes remain open). Another commenter also reported being unable to reproduce the two ~4 bytes/param leaks on a newer PyTorch build.

### 1. `unscale_and_clip_grads` — float conversion only on CPU offload (`stage_1_and_2.py`)

Convert `combined_scale` to a Python `float` **only when `cpu_offload` is enabled**, so `mul_` receives a scalar on the CPU path.

Per Codex review: an unconditional `float()` would read a CUDA scalar back to Python when ZeRO-1/2 runs without CPU offload and `clip_grad > 0`, forcing a device sync every step. The GPU clipping path stays on-device.

### 2. `DeepSpeedCPUAdam` — per-group `malloc_trim(0)` (`cpu_adam.py`)

Call `malloc_trim(0)` once per param group / subgroup on that group's **first** step.

Per Codex review: ZeRO-2 CPU offload invokes `step()` once per `bit16_groups` entry. An optimizer-wide “first step” flag would skip trim for later groups after group 0. Per-group tracking fixes that.

This is a best-effort glibc RSS mitigation (Linux-only no-op elsewhere), not a correctness fix.

### 3. Tests (`test_cpu_offload_memory.py`)

- per-group trim call count
- CPU-offload float unscale path keeps storage
- non-offload path still scales correctly
- LF line endings

## Removed after review

- Dropped the `ds_adam_step` `is_contiguous()` guard in `cpu_adam_impl.cpp`. As @delock pointed out, `Tensor.contiguous()` already returns `self` when the tensor is contiguous, so the explicit check was a no-op.

## Test plan

- [x] `test_malloc_trim_helper_exists`
- [x] `test_malloc_trim_called_per_param_group`
- [x] `test_cpu_adam_first_step_initializes_optimizer_states`
- [x] `test_unscale_and_clip_grads_cpu_offload_uses_float_scale`
- [x] `test_unscale_and_clip_grads_no_clip`
- [x] `test_unscale_and_clip_grads_gpu_path_skips_float_conversion`

## Notes

- Addressed Codex P2 comments on #8132 (GPU sync + per-group trim).
- Addressed @delock review: removed unnecessary `.contiguous()` change.
- DCO: author/committer email matches `Signed-off-by: liuyun7345 <liuyun7345@sina.com>`.

Addresses #7693

### DIFF
--- a/deepspeed/ops/adam/cpu_adam.py
+++ b/deepspeed/ops/adam/cpu_adam.py
@@ -10,6 +10,26 @@ from deepspeed.utils.logging import should_log_le
 from deepspeed.ops.op_builder import CPUAdamBuilder
 
 
+def _malloc_trim():
+    """Release freed CPU memory back to the OS on Linux.
+
+    PyTorch's CPU allocator caches freed memory blocks by size rather than
+    returning them to the OS.  After the first optimizer step several large
+    temporary tensors (contiguous copies, gradient scale buffers, etc.) have
+    been created and freed, but their memory remains pinned in the allocator
+    cache and is not reused by subsequent differently-sized allocations.
+
+    malloc_trim(0) asks glibc to release freed arena memory back to the OS.
+    This is a best-effort RSS mitigation for CPU-offload workloads, not a
+    correctness fix.  This is a no-op on non-Linux platforms.
+    """
+    try:
+        import ctypes
+        ctypes.CDLL("libc.so.6").malloc_trim(0)
+    except (OSError, AttributeError):
+        pass
+
+
 class DeepSpeedCPUAdam(torch.optim.Optimizer):
     optimizer_id = 0
 
@@ -160,6 +180,18 @@ class DeepSpeedCPUAdam(torch.optim.Optimizer):
                 self.ds_opt_adam.adam_update(self.opt_id, state['step'], group['lr'], beta1, beta2, group['eps'],
                                              group['weight_decay'], group['bias_correction'], p.data, p.grad.data,
                                              state['exp_avg'], state['exp_avg_sq'])
+
+            # ZeRO-2 CPU offload calls step() once per param group.  Trim after
+            # each group's first step so later groups' lazy state init is also
+            # released. See #7693 / Codex review on #8132.
+            trimmed = getattr(self, '_trimmed_groups', None)
+            if trimmed is None:
+                trimmed = set()
+                self._trimmed_groups = trimmed
+            if group_id not in trimmed:
+                trimmed.add(group_id)
+                _malloc_trim()
+
         return loss
 
     @torch.no_grad()
@@ -197,6 +229,16 @@ class DeepSpeedCPUAdam(torch.optim.Optimizer):
                 self.ds_opt_adam.adam_update(self.opt_id, state['step'], group['lr'], beta1, beta2, group['eps'],
                                              group['weight_decay'], group['bias_correction'], p.data, p.grad.data,
                                              state['exp_avg'], state['exp_avg_sq'])
+
+        # Trim after this subgroup's first step (ZeRO-3 subgroup path).
+        trimmed = getattr(self, '_trimmed_groups', None)
+        if trimmed is None:
+            trimmed = set()
+            self._trimmed_groups = trimmed
+        if subgroup_id not in trimmed:
+            trimmed.add(subgroup_id)
+            _malloc_trim()
+
         return loss
 
     @torch.no_grad()
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -2306,6 +2306,13 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
             clip = torch.clamp(clip, min=1.0)
             combined_scale = clip * self.loss_scale
 
+        # Only convert to a Python float on the CPU-offload path.  On GPU ZeRO,
+        # total_norm / combined_scale may live on the accelerator; an unconditional
+        # float() would force a device sync every step.  The allocator concern in
+        # #7693 is CPU-specific, so keep the GPU path on-device. See #7693 / #8132.
+        if self.cpu_offload:
+            combined_scale = float(combined_scale)
+
         for grad in grad_groups_flat:
             if isinstance(grad, list):
                 sub_partitions = grad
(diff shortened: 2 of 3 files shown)


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2306,6 +2306,13 @@ def unscale_and_clip_grads(self, grad_groups_flat, total_norm):
             clip = torch.clamp(clip, min=1.0)
             combined_scale = clip * self.loss_scale
 
+        # Convert to Python float so that mul_ receives a scalar, not a 0-dim
+        # tensor. On CPU, mul_(tensor) may internally allocate a full-sized
+        # result buffer even for 0-dim tensors, pinning ~4 bytes/param in the
+        # CPU allocator cache that is never reused by later allocations.
+        # mul_(float) is guaranteed in-place. See issue #7693.
+        combined_scale = float(combined_scale)
```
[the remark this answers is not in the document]
liuyun7345: Thanks — agreed. Updated so `float(combined_scale)` only runs when `self.cpu_offload` is true, keeping the GPU ZeRO clipping path on-device and avoiding the sync regression.

deepspeed/ops/adam/cpu_adam.py
```diff
@@ -160,6 +180,16 @@ def step(self, closure=None):
...
                                              state['exp_avg'], state['exp_avg_sq'])
+
+        # Release CPU allocator cached memory after the first step.
+        # The first step lazily initialises optimizer states and creates/frees
+        # several large temporary tensors whose memory is retained by the CPU
+        # allocator cache.  malloc_trim(0) releases this back to the OS.
+        # See issue #7693.
+        if not getattr(self, '_first_step_done', False):
+            self._first_step_done = True
+            _malloc_trim()
```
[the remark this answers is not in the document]
liuyun7345: Thanks — agreed. Replaced the optimizer-wide `_first_step_done` flag with per-group / per-subgroup `_trimmed_groups` tracking so each ZeRO param group still gets `malloc_trim` after its first lazy state init.

csrc/adam/cpu_adam_impl.cpp
```diff
@@ -245,10 +245,14 @@ int ds_adam_step(int optimizer_id,
...
-    auto exp_avg_c = exp_avg.contiguous();
-    auto exp_avg_sq_c = exp_avg_sq.contiguous();
+    // Only create contiguous copies when needed.  Unconditional .contiguous()
+    // can allocate full-sized CPU buffers (~4 bytes/param for fp32) that
+    // PyTorch's CPU allocator caches after the temporary is freed, pinning
+    // memory that is not reused by later allocations. See issue #7693.
+    auto params_c = params.is_contiguous() ? params : params.contiguous();
+    auto grads_c = grads.is_contiguous() ? grads : grads.contiguous();
+    auto exp_avg_c = exp_avg.is_contiguous() ? exp_avg : exp_avg.contiguous();
+    auto exp_avg_sq_c = exp_avg_sq.is_contiguous() ? exp_avg_sq : exp_avg_sq.contiguous();
```
delock: From this document, this change should not be necessary, PyTorch ensures return the same tensor if the tensor is already contiguous.
https://docs.pytorch.org/docs/2.13/generated/torch.Tensor.contiguous.html

liuyun7345: Thanks @delock — you're right. `Tensor.contiguous()` already returns `self` when the tensor is contiguous, so the explicit `is_contiguous()` check is a no-op. I'll remove this change from the PR.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2306,6 +2306,13 @@ def unscale_and_clip_grads(self, grad_groups_flat, total_norm):
             clip = torch.clamp(clip, min=1.0)
             combined_scale = clip * self.loss_scale
 
+        # Only convert to a Python float on the CPU-offload path.  On GPU ZeRO,
+        # total_norm / combined_scale may live on the accelerator; an unconditional
+        # float() would force a device sync every step.  The allocator concern in
+        # #7693 is CPU-specific, so keep the GPU path on-device. See #7693 / #8132.
+        if self.cpu_offload:
+            combined_scale = float(combined_scale)
```
delock, line 2314: I didn't observed the 4byte/element memory usage from combined_scale.  The pytorch version I use is torch 2.11.  Did you observe this memory bloat on your environment?  Thanks!

delock:
Hi @liuyun7345 thanks for tracking this issue and do the best effort fix.   Yet I found some claims in the original issue may not be the real reason.  More over, on a single 5090 system with CPU offload, we didn't observe significant RSS change before and after the PR.   Did you observe RSS change after applying the patch?

I would suggest to digg deeper to find the real reason for RSS usage.   If the cause claimed in #7693 is not true, then we should avoid adding these complexity for no return.

liuyun7345:
Thanks @delock for the careful verification. Honestly, I did not independently observe the RSS change either — this PR was built on the hypothesis from the issue report, not from a reproduction on my side. Given that multiple people (including you) cannot reproduce the claimed leaks, I agree the complexity is not justified without evidence. Closing this PR.

deepspeed/runtime/zero/stage_1_and_2.py
liuyun7345: Thanks @delock — no, I did not observe the 4 bytes/element memory bloat from `combined_scale` on my environment either. This change was based on the hypothesis in the issue report, not a reproduction. Given that it cannot be reproduced, I'm closing the PR to avoid adding complexity for no return.

liuyun7345:
Closing this PR. The fixes were based on hypotheses from the issue report, but the claimed leaks cannot be reproduced by multiple reviewers (@delock, @xylian86) including on the latest PyTorch. Adding complexity without measurable benefit is not warranted. The underlying issue #7693 should be investigated further with actual RSS profiling before attempting another fix.

PR #8133, opened 2026-07-12
fix: add zero-guards for divisor/scale edge cases (#7838)

### LINKED ISSUE #7838: [BUG] Multiple missing zero-guards cause ZeroDivisionError / non-finite values across DeepSpeed (4 locations)
**Describe the bug**
I found 4 related zero-handling issues where divisors/scales are not validated before use.  
Three can raise `ZeroDivisionError`; one can silently propagate `inf`/`nan`.

**Affected locations**
1. `deepspeed/utils/groups.py`  
   `_ensure_divisibility(numerator, denominator)` uses `numerator % denominator` without guarding `denominator == 0`.

2. `deepspeed/utils/timer.py`  
   `ThroughputTimer._is_report_boundary()` checks `None` but not `0` before:  
   `self.global_step_count % self.steps_per_output`.

3. `deepspeed/inference/v2/inference_utils.py`  
   `ceil_div(a, b)` returns `-(-a // b)` without guarding `b == 0`.

4. `op_builder/hpu/fp_quantizer.py`  
   `FPQuantizer.dequantize()` computes `(1.0 / scale)` without guarding zero elements in `scale`, which can produce non-finite values and corrupt outputs silently.

**To Reproduce**
1. `_ensure_divisibility(8, 0)` -> `ZeroDivisionError`
2. `_is_report_boundary()` with `steps_per_output=0` -> `ZeroDivisionError`
3. `ceil_div(10, 0)` -> `ZeroDivisionError`
4. `1.0 / torch.tensor([0.0, 1.0])` -> `tensor([inf, 1.])` (same pattern used in HPU dequantize path)

**Expected behavior**
- Explicit validation for invalid zero values.
- Clear user-facing error messages (or safe clamping where appropriate).
- No raw modulo/division-by-zero exceptions.
- No silent non-finite propagation in dequantization.

**Suggested fixes**
1. `_ensure_divisibility`: add guard before modulo (`denominator != 0`).
2. `_is_report_boundary`: treat `0` as invalid/disabled (`if not self.steps_per_output:` or explicit `<= 0` validation).
3. `ceil_div`: reject `b == 0` with clear error.
4. `FPQuantizer.dequantize`: clamp scale to `torch.finfo(scale.dtype).tiny` (or validate and fail clearly) before inversion.

**System info**
- Can provide full `ds_report` output if needed.
- Repros above are minimal and mostly pure-Python, except item 4 which is on the HPU backend path.

liuyun7345:
## Summary

Fixes #7838 by adding explicit zero/non-finite guards at the four reported sites so DeepSpeed fails early with a clear error instead of raising raw `ZeroDivisionError` or silently propagating `inf`/`nan`.

This picks up the inactive draft #7855 (no updates since 2026-02-17).

## Changes

1. **`deepspeed/utils/groups.py`** — `_ensure_divisibility` asserts `denominator != 0` before modulo.
2. **`deepspeed/utils/timer.py`** — `ThroughputTimer` rejects non-positive `steps_per_output` in `__init__`, and `_is_report_boundary` re-checks if the attribute is later mutated.
3. **`deepspeed/inference/v2/inference_utils.py`** — `ceil_div` raises `ValueError` when `b == 0`.
4. **`op_builder/hpu/fp_quantizer.py`** — `dequantize` rejects zero / non-finite scales before computing `1.0 / scale`.
5. **`tests/unit/utils/test_zero_guards.py`** — regression tests for each failure mode (HPU builder tests skip if that module is unavailable).

## Test plan

- [x] `_ensure_divisibility(8, 0)` raises clear AssertionError
- [x] valid `_ensure_divisibility` inputs still pass
- [x] `ceil_div(10, 0)` raises ValueError
- [x] `ceil_div` matches `math.ceil` for valid inputs
- [x] `ThroughputTimer(..., steps_per_output=0/-1)` raises ValueError
- [x] mutated `steps_per_output=0` is caught in `_is_report_boundary`
- [x] HPU `dequantize` rejects zero / NaN / zero-scalar scales before HPU op

Fixes #7838

### DIFF
--- a/deepspeed/inference/v2/inference_utils.py
+++ b/deepspeed/inference/v2/inference_utils.py
@@ -102,4 +102,6 @@ def ceil_div(a: int, b: int) -> int:
     """
     Return ceil(a / b).
     """
+    if b == 0:
+        raise ValueError(f"ceil_div divisor must be non-zero (got a={a}, b={b})")
     return -(-a // b)
--- a/tests/unit/utils/test_zero_guards.py
+++ b/tests/unit/utils/test_zero_guards.py
@@ -1,0 +1,97 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+"""Regression tests for zero/division safety gaps reported in #7838."""
+
+import math
+
+import pytest
+import torch
+
+from deepspeed.utils.groups import _ensure_divisibility
+from deepspeed.utils.timer import ThroughputTimer
+from deepspeed.inference.v2.inference_utils import ceil_div
+
+
+class _DummyTimerConfig:
+    enabled = False
+    synchronized = False
+
+
+def test_ensure_divisibility_rejects_zero_denominator():
+    with pytest.raises(AssertionError, match="non-zero"):
+        _ensure_divisibility(8, 0)
+
+
+def test_ensure_divisibility_accepts_valid_inputs():
+    _ensure_divisibility(8, 2)
+    _ensure_divisibility(0, 4)
+
+
+def test_ceil_div_rejects_zero_divisor():
+    with pytest.raises(ValueError, match="non-zero"):
+        ceil_div(10, 0)
+
+
+def test_ceil_div_matches_math_ceil():
+    assert ceil_div(10, 3) == math.ceil(10 / 3)
+    assert ceil_div(9, 3) == 3
+    assert ceil_div(1, 1) == 1
+
+
+def test_throughput_timer_rejects_zero_steps_per_output():
+    with pytest.raises(ValueError, match="positive"):
+        ThroughputTimer(_DummyTimerConfig(), batch_size=1, steps_per_output=0)
+
+
+def test_throughput_timer_rejects_negative_steps_per_output():
+    with pytest.raises(ValueError, match="positive"):
+        ThroughputTimer(_DummyTimerConfig(), batch_size=1, steps_per_output=-1)
+
+
+def test_throughput_timer_report_boundary_guards_mutated_zero():
+    timer = ThroughputTimer(_DummyTimerConfig(), batch_size=1, steps_per_output=2)
+    timer.steps_per_output = 0
+    with pytest.raises(ValueError, match="positive"):
+        timer._is_report_boundary()
+
+
+def test_throughput_timer_report_boundary_none_is_safe():
+    timer = ThroughputTimer(_DummyTimerConfig(), batch_size=1, steps_per_output=None)
+    assert timer._is_report_boundary() is False
+
+
+def _import_hpu_fp_quantizer_builder():
+    try:
+        from op_builder.hpu.fp_quantizer import FPQuantizerBuilder
+        return FPQuantizerBuilder
+    except ImportError:
+        pytest.skip("HPU FPQuantizer builder is not available")
+
+
+def test_hpu_fp_quantizer_dequantize_rejects_zero_scale():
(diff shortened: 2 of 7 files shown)


tests/unit/utils/test_zero_guards.py
```diff
@@ -0,0 +1,97 @@
...
+
+def _import_hpu_fp_quantizer_builder():
+    try:
+        from op_builder.hpu.fp_quantizer import FPQuantizerBuilder
+        return FPQuantizerBuilder
+    except ImportError:
+        pytest.skip("HPU FPQuantizer builder is not available")
+
+
+def test_hpu_fp_quantizer_dequantize_rejects_zero_scale():
```
sfc-gh-truwase: These tests seems to be hpu-specific. Can they not be generalized?

deepspeed/inference/v2/inference_utils.py
```diff
@@ -102,4 +102,6 @@ def ceil_div(a: int, b: int) -> int:
     """
     Return ceil(a / b).
     """
+    if b == 0:
```
sfc-gh-truwase, line 105: Can you please share the usage case that found this issue?

tests/unit/utils/test_zero_guards.py
liuyun7345: Good point — generalizing is possible, and I'll do it. The `1.0 / scale` inversion currently only exists in the HPU `FPQuantizer.dequantize`, but the same failure mode applies to all backends: `quantize` computes `scale = q_range / max_vals`, so inf/nan in the input drives `max_vals` to inf and silently produces `scale = 0`.

I'll move the zero/non-finite scale validation up into the shared front-end `FP_Quantize.dequantize` (`deepspeed/ops/fp_quantizer/quantize.py`) so it covers CUDA and HPU alike, and relocate the test into `tests/unit/ops/fp_quantizer/test_fp_quant.py` using the existing accelerator-agnostic pattern (`deepspeed.ops.op_builder.FPQuantizerBuilder` + module-level skip via `__compatible_ops__`). The direct `op_builder.hpu` import goes away.

deepspeed/inference/v2/inference_utils.py
liuyun7345: Honest answer: there is no production usage case on my side. All four guards in this PR come from issue #7838, which was reported from code inspection with synthetic repros (`ceil_div(10, 0)` raising a raw `ZeroDivisionError`), not from a real workload failure.

The case for keeping this guard is narrow: `ceil_div` is a small shared utility, and the guard turns a cryptic `ZeroDivisionError` into a clear `ValueError` at the API boundary, with zero behavior change for valid inputs. That said, if you'd rather not add guards without a demonstrated trigger, I'm happy to drop this hunk.

tests/unit/utils/test_zero_guards.py
liuyun7345: Done — pushed as 1098c0a9. The zero/non-finite scale validation now lives in the shared front-end `FP_Quantize.dequantize` / `selective_dequantize` (`deepspeed/ops/fp_quantizer/quantize.py`), and the regression tests moved to `tests/unit/ops/fp_quantizer/test_fp_quant.py` using the accelerator-agnostic pattern (`deepspeed.ops.op_builder.FPQuantizerBuilder` + module-level skip via `__compatible_ops__`), exercised through the public `FP_Quantize.dequantize` API. The HPU-only guard in `op_builder/hpu/fp_quantizer.py` and the direct `op_builder.hpu` import in the tests are both gone.

Issue #8076, opened 2026-06-18
[BUG]ds_z3_config error


whyseu:
I use Python 3.11 + PyTorch 2.6.0 + CUDA 12.4. I install the deepspeed using "pip install deepspeed --no-deps".
when I use LLama factory to SFT(LoRA) the Qwen3.5 9B model with 4*H100, there is error:
TypeError: output tensor must have the same type as input tensor

How can I fix it?

pengdurice:
thank you for raising this issue. but this is a very general error msg, would you mind share in more details and also the error trace for debugging? thank you!

PR #8134, opened 2026-07-12
Tidy docs style in DeepSpeed (#8076)

bglglzd:
This PR is addressing an open issue. It is intentionally a minimal change.

Related to #8076.


tohtana:
Hi @bglglzd 
This change doesn't look related to #8076. Can you clarify your intention?

bglglzd:
You’re absolutely right, thank you. You’re also right that this change is not aligned with #8076. I’m going to stop on this PR for now, fix the exact #8076 concern in a new follow-up, and post the scope there. I’ll only push a targeted change once it maps to that issue.

PR #8137, opened 2026-07-12
Fix CommsLogger.stop_profiling_comms not disabling global profiling

vineethsaivs:
## What

`CommsLogger.stop_profiling_comms()` set `self.prof_all = True`, the same value `start_profiling_comms()` sets, so calling stop was a no-op:

```python
def start_profiling_comms(self):
    self.prof_all = True

def stop_profiling_comms(self):
    self.prof_all = True   # should be False
```

`prof_all` is the flag that makes the comms logger profile every collective (`deepspeed/comm/comm.py` gates on `... or comms_logger.prof_all or ...`). Because stop set it back to `True`, once global comm profiling was started it could never be turned off. The sibling pair in the same class, `start_profiling_op` / `stop_profiling_op`, correctly adds then removes ops, which confirms the intended start=on / stop=off asymmetry.

## Fix

Set `prof_all = False` in `stop_profiling_comms()`.

## Test

Added `tests/unit/comm/test_comms_logger.py`, a CPU-only unit test that starts then stops comm profiling and asserts `prof_all` ends up `False`. It fails before the change (stays `True`) and passes after.

### DIFF
--- a/deepspeed/utils/comms_logging.py
+++ b/deepspeed/utils/comms_logging.py
@@ -91,7 +91,7 @@ class CommsLogger:
         self.prof_all = True
 
     def stop_profiling_comms(self):
-        self.prof_all = True
+        self.prof_all = False
 
     # E.g. start_profiling_op('all_reduce')
     def start_profiling_op(self, op_name_list):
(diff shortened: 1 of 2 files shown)


tohtana (changes requested):
Great catch! This looks good to me.
Can you just change the license header to "# Copyright (c) DeepSpeed Team."? DeepSpeed is not under Microsoft anymore.

vineethsaivs:
Thanks for the review! Updated the copyright header to `# Copyright (c) DeepSpeed Team.`.

tohtana (approved):
Thank you for the fix!

Issue #8104, opened 2026-06-30
DeepSpeed - Q3 Roadmap


PKUWZP:
This is the roadmap for DeepSpeed Q3 2026. Feedback is welcome — please leave comments on this issue or join the `#2026q3-roadmap` channel on the [DeepSpeed Slack](https://join.slack.com/t/deepspeedworkspace/shared_invite/zt-3a8pjd8dd-PCj2hMvR4Y2syPwVnjEoww).

Some of the items are the extension of the Q2 roadmap: https://github.com/deepspeedai/DeepSpeed/issues/7861

# New feature and enhancement

## Enhance Expert-Parallelism (EP) support

- [x] **AutoEP Support**: AutoEP enables Expert Parallelism (EP) for major Mixture-of-Experts (MoE) models out of the box, eliminating the need for users to write model-specific parallelization code. By automatically distributing expert layers across devices, AutoEP allows users to scale MoE training with minimal configuration changes.
- [x] Combining AutoEP with AutoTP
- [ ] Extend AutoTP capabilities by integrating Liger Kernel (sharding LM heads and adopting Online Softmax)
- [ ] EP performance optimization and benchmarking

## New Accelerators Support (Q3)
- [ ] DeepSpeed TPU accelerators support
- [ ] Performance Optimization on Emerging AI Accelerators (e.g. Biren AI Accelerators) 

## On-Policy Distillation Trainer Support (Q3) 
- [x] Systems Design, prototyping and benchmarking (reference: https://arxiv.org/abs/2604.14084) 

## DeepCompile efficiency and robustness improvements

- [ ] **Formal pass contracts and validation of optimization passes**: Add lightweight optimization pass contracts for automatic compatibility validation and ordering.
- [ ] **Composable AutoTP / AutoEP / SP optimization passes**: Implement AutoTP and AutoEP at a compiler level and integrate them with AutoSP.
- [ ] **Enhance optimization passes**: Add some optimization passes including activation offloading.

## Pipeline parallelism with Ray

- [ ] **Ray-backed pipeline stage placement and execution**: Prototype pipeline stage placement and execution using Ray actor groups for a representative transformer workload, enabling *heterogeneous* resource allocation across pipeline stages. (Will be a generalized implementation of [this](https://www.anyscale.com/blog/30-faster-multimodal-ai-training-with-ray-and-disaggregated-hybrid)).
[...]

PR #8139, opened 2026-07-14
[DeepCompile] Add lightweight pass contracts for optimization passes

jahnavi-yelamanchi:
### Summary
Based on the DeepCompile efficiency and robustness track in the Q3 2026 roadmap #8104 , this PR implements: _Formal pass contracts and validation of optimization passes: Add lightweight optimization pass contracts for automatic compatibility validation and ordering_

### Motivation
DeepCompile passes execute in the order given by the schedule (`init_schedule`, `backend.py`), with no validation of that order. Inter-pass dependencies are implicit: `prefetch` and `selective_gather` rewrite the all-gather/release ops inserted by `zero3_compile` and require it to run first (cf. the default schedule in `init_z3.py`). An invalid order currently surfaces as a downstream failure rather than a diagnostic at schedule time.

### Changes
- `deepspeed/compile/passes/contract.py`: `PassContract` dataclass (`provides` / `requires` / `conflicts_with` / `phase`), a contract registry, and `validate_schedule()`, which raises `PassContractError` on the first unmet requirement or conflict. No torch dependency.
- Contracts declared for the four built-in passes: `zero3_compile` provides `z3_gather_release`; `prefetch` and `selective_gather` require it.
- `register_compile_pass(name, fn, contract=None)`: optional `contract` argument, backward compatible.
- `validate_schedule()` invoked on user-supplied schedules in the compile path (`engine.py`).

### Design
- Passes without a contract are unconstrained; existing and custom passes are unaffected. Internal default schedules are not validated, only user-supplied ones.
- Dependencies are expressed as capability tags (`z3_gather_release`) rather than pass names, so passes can be added or swapped without editing dependents.
- Conflicts are symmetric: either pass may declare the incompatibility, and both orderings are rejected.
- Scope is limited to metadata and validation, not dependency resolution or automatic reordering.

### Testing
- `tests/unit/compile/test_pass_contract.py`: unit tests for valid ordering, missing requirements, same-step providers, symmetric conflicts, uncontracted-pass pass-through, and callable resolution.

### DIFF
--- a/deepspeed/compile/backend.py
+++ b/deepspeed/compile/backend.py
@@ -77,8 +77,11 @@ opt_passes = {}
 fwd_real_inputs = []
 
 
-def register_compile_pass(name: str, opt_pass_fn):
+def register_compile_pass(name: str, opt_pass_fn, contract=None):
     opt_passes[name] = opt_pass_fn
+    if contract is not None:
+        from .passes.contract import register_pass_contract
+        register_pass_contract(contract)
 
 
 def init_schedule(schedule):
--- a/deepspeed/compile/backend.py
+++ b/deepspeed/compile/backend.py
@@ -78,10 +78,11 @@ fwd_real_inputs = []
 
 
 def register_compile_pass(name: str, opt_pass_fn, contract=None):
+    from .passes.contract import register_pass_contract
     opt_passes[name] = opt_pass_fn
-    if contract is not None:
-        from .passes.contract import register_pass_contract
-        register_pass_contract(contract)
+    # Replace the callable and its contract together under the same name; contract=None clears any
+    # contract previously registered for the pass.
+    register_pass_contract(name, contract)
 
 
 def init_schedule(schedule):
--- a/deepspeed/compile/passes/contract.py
+++ b/deepspeed/compile/passes/contract.py
@@ -1,4 +1,4 @@
-# Copyright (c) Microsoft Corporation.
+# Copyright (c) DeepSpeed Team.
 # SPDX-License-Identifier: Apache-2.0
 
 # DeepSpeed Team
--- a/deepspeed/compile/passes/contract.py
+++ b/deepspeed/compile/passes/contract.py
@@ -20,9 +20,11 @@ class PassContract:
     passes that name each other in ``conflicts_with`` may not share a schedule. ``phase`` is
     informational for now and records whether a pass rewrites the forward graph, the backward
     graph, or both.
+
+    A contract does not carry the pass name: the registry key set by ``register_compile_pass`` is
+    the single source of truth for a pass's identity.
     """
 
-    name: str
     provides: FrozenSet[str] = frozenset()
     requires: FrozenSet[str] = frozenset()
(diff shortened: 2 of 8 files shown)


jahnavi-yelamanchi:
Thank you @tohtana, I hadn't thought about making the registration atomic and I completely missed it. Really appreciate your help while I'm still getting my footing. Happy to pull your fix into my branch, or if you'd prefer to push it directly, that works for me too; whichever's easier on your end. Anything else I should be doing?

tohtana:
Hi @jahnavi-yelamanchi,

Thank you for your interest in DeepCompile! We definitely need a contract mechanism for composing optimization passes. There are many possible approaches, but I would prefer to start with a simple design and evolve it as we learn more, rather than waiting until we have a comprehensive solution. From that perspective, I think this PR is a good first step.

My concerns are mostly about a few technical details in the registration and resolution model:

- A stale contract remains registered when a pass is replaced through the existing two-argument `register_compile_pass(name, fn)` API.
- If the same callable is registered under multiple names, the reverse `fn -> name` dictionary is lossy, so a named schedule may be validated against the wrong contract.
- The implementation does not verify that the name passed to `register_compile_pass()` matches `PassContract.name`, so a mismatch can silently disable contract enforcement.

The last issue in particular seems to come from having two sources of truth for the pass identity. If `PassContract.name` must always match the first argument to `register_compile_pass()`, could we remove `name` from `PassContract` and let the registry key be the sole identity?

For example:

```python
register_compile_pass(
    "foo",
    foo_fn,
    PassContract(
        requires=frozenset({"missing_cap"}),
    ),
)
```

Ideally, `register_compile_pass()` would atomically replace both the callable and its contract under "foo", with contract=None clearing any previous contract. Named schedules could then be validated before names are converted to callables, preserving the identity selected by the user. A callable used under multiple registered names could require an explicit name in the schedule.

Regarding the [supplemental PR](https://github.com/tohtana/DeepSpeed/pull/28), I apologize for the confusion. I asked Codex to explore a possible solution, but it opened the PR without my intending to submit that design. Since it is not aligned with the approach described above, I will close it now.

jahnavi-yelamanchi:
Thank you, this direction makes a lot of sense; a single source of truth for the pass identity is much cleaner. I'll try to implement it and also look for any further edge cases along the way. No worries at all about the supplemental PR. I'll keep working on it and push a fix as soon as it's ready.

jahnavi-yelamanchi:
Hi @tohtana, thank you for the guidance, I've pushed the changes based on my understanding of your review. Please let me know if anything needs adjusting.

tohtana:
Hi @jahnavi-yelamanchi,
Thank you for addressing my comments! The change looks good to me. Can you fix the copyright in the header? We now use: 

```python

# DeepSpeed Team
```

jahnavi-yelamanchi:
Thanks @tohtana ! Updated the copyright header to Copyright (c) DeepSpeed Team. on the new files and pushed.

tohtana (approved):
Thank you, @jahnavi-yelamanchi!
Let's merge it. I appreciate your contribution to DeepCompile.

Issue #7807, opened 2026-01-23
[BUG] Cross-partition parameters incorrectly updated when using ZeRO-1/ZeRO-2 with reduce_scatter=true and Muon optimizer


wflying000:
**Describe the bug**

When using the Muon optimizer with ZeRO-1 or ZeRO-2 and `reduce_scatter=true`, parameters that span partition boundaries receive incorrect gradient updates due to incomplete gradient reduction before orthogonalization. This causes the Newton-Schulz orthogonalization in Muon to operate on partially reduced gradients, leading to incorrect parameter updates for cross-partition parameters.

**To Reproduce**

1. Set up a training configuration with:
   - ZeRO stage 1 or 2 with `reduce_scatter: true` in the DeepSpeed config
   - Muon optimizer (or MuonWithAuxAdam) for 2D weight matrices
   - Multi-GPU training (e.g., 4 GPUs)

2. Run training and observe that some parameters (those crossing partition boundaries) receive inconsistent gradient updates across ranks

**Expected behavior**

All ranks should receive fully reduced gradients before Muon's orthogonalization step, ensuring that:
- The Newton-Schulz iteration operates on complete, correctly reduced gradients
- All ranks compute identical orthogonalized updates for the same parameter
- Cross-partition parameters are updated consistently

**ds_report output**
```
--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
 [WARNING]  async_io: please install the libaio-dev package with apt
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
async_io ............... [NO] ....... [NO]
fused_adam ............. [NO] ....... [OKAY]
cpu_adam ............... [NO] ....... [OKAY]
cpu_adagrad ............ [NO] ....... [OKAY]
cpu_lion ............... [NO] ....... [OKAY]
dc ..................... [NO] ....... [OKAY]
 [WARNING]  Please specify CUTLASS location directory as environment variable CUTLASS_PATH
[...]

PR #8141, opened 2026-07-14
fix(muon): support ZeRO-1/2 reduce scatter

Micuks:
## Problem

#8090 made `Muon + ZeRO-1/2 + reduce_scatter` fail fast because a Muon matrix that crosses a ZeRO partition boundary does not have a fully reduced gradient on any owner rank.

The reduce-scatter path copies each averaged slice only to that slice's owner. Muon, however, runs the nonlinear Newton-Schulz update on the full matrix before `get_flat_partition()` selects the local ZeRO slice. Each owner therefore sees a different matrix: its own reduced slice plus local, unreduced slices owned by other ranks.

## Change

For a Muon parameter that spans more than one ZeRO partition, each reduced slice is copied to every rank that owns part of that parameter. Those ranks then run `muon_update` on the same full reduced gradient, and the existing code keeps only the local ZeRO partition.

Parameters that stay within one partition keep the current owner-only copy path. With the default multi-rank bucket all-reduce, this does not add a collective. When `use_multi_rank_bucket_allreduce=false`, only ranges belonging to split Muon parameters switch to all-reduce. The copy helper still accepts the integer rank form used by ZenFlow.

Optimizer offload remains unsupported because that path keeps only partition-local gradients.

## Tests

The numerical test compares the applied update with `muon_update` on the full data-parallel gradient. The 2-GPU cases cover:

- ZeRO stages 1 and 2
- `gram` and `standard` Newton-Schulz methods on the existing all-reduce path
- reduce-scatter with gradient accumulation and overlap communication
- `use_multi_rank_bucket_allreduce=false`
- extra-large parameters and non-contiguous gradients
- optimizer offload rejection

Focused helper tests cover both the existing integer copy target and the new multi-owner target.

Refs #7807

Alternative to #7878; replaces the fail-fast behavior added by #8090 with a targeted reduction path.


delock (approved):

delock:
@Micuks  thanks for the fix!

PR #8146, opened 2026-07-15
[AutoTP] Enable HF `colwise_gather_output` to support `lm_head` replace

jinyouzhi:
### Changes

Enable HF `colwise_gather_output` in `tp_plan` through AutoTP, allowing vocab parallelism for an untied `lm_head`.

### Known Limitations

- Only supports an untied `lm_head`. Tied `lm_head` falls back to the legacy implementation.
- Uneven tensor parallelism is not supported in this PR.


deepspeed/module_inject/auto_tp.py
```diff
@@ -461,21 +475,88 @@ def _create_row_parallel_layer(self, module, spec: TPLayerSpec, name: str):
 
     def _create_column_parallel_layer(self, module, spec: TPLayerSpec, name: str):
         """Create column-parallel layer (AllReduce in backward)."""
+        if spec.gather_output and self.mp_size is not None and self.mp_size > 1:
+            output_dim = module.weight.shape[0]
+            if output_dim % self.mp_size != 0:
```
delock, line 480: Does this condition mean TP on lm_head is only possible when vocab size divisible by number of TP ranks?

PKUWZP (changes requested):
Thanks for submitting the PR, a couple of suggestions: 

1. I don't see this PR leverages online softmax and in-place computation, this is the common optimization that saves GPU memory significantly. 

2. We need to run some profiling to measure the GPU footprint.

Happy to work closely together to have a separate PR for this.

deepspeed/module_inject/autotp_config.py
```diff
@@ -49,6 +49,13 @@ class TPLayerSpec:
             partition_type=PartitionType.COLUMN,
         )
 
+        # Column-parallel layer with replicated output (e.g., an untied LM head)
+        TPLayerSpec(
+            patterns=[".*\\.lm_head$"],
```
delock: pattern should be changed to `*lm_head\.weight$` in order to be matched.

deepspeed/module_inject/auto_tp.py
delock, line 480: can you also add a test to test checkpoint are correct in uneven case?  i.e. even number vocab size, with 3 ranks.

jinyouzhi:
> Thanks for submitting the PR, a couple of suggestions:
> 
> 1. I don't see this PR leverages online softmax and in-place computation, this is the common optimization that saves GPU memory significantly.
> 2. We need to run some profiling to measure the GPU footprint.
> 
> Happy to work closely together to have a separate PR for this.

Thank you for your suggestions! You‘re absolutely right, this PR is far from the target. I've refined the PR description, and I'll submit a separate PR for those optimizations.

deepspeed/module_inject/auto_tp.py
jinyouzhi: added.

deepspeed/module_inject/autotp_config.py
jinyouzhi: fix

deepspeed/module_inject/auto_tp.py
jinyouzhi: Yes, let uneven fallback to legacy.

delock:
> Thanks for submitting the PR, a couple of suggestions:
> 
> 1. I don't see this PR leverages online softmax and in-place computation, this is the common optimization that saves GPU memory significantly.
> 2. We need to run some profiling to measure the GPU footprint.
> 
> Happy to work closely together to have a separate PR for this.

Suggest profile GPU foot print with the separate PR.   Online softmax and in-place computate could save memory for OPSD teacher TP sharding.

jinyouzhi:
> 1. saves GPU memory significantly.

Great! list the keypoints into an issue to track https://github.com/deepspeedai/DeepSpeed/issues/8173

jinyouzhi:
Hi @PKUWZP, Could you have a look?

PKUWZP (approved):
LGTM.

PR #8154, opened 2026-07-18
Fix repeated gradient scaling across pipeline stages

### LINKED ISSUE #8152: [BUG] Gradient accumulation scaling is applied repeatedly across pipeline stages
**Describe the bug**
When using PipelineEngine with pipeline parallelism and gradient_accumulation_steps > 1, gradient accumulation scaling appears to be applied once at every pipeline stage boundary instead of once for the entire pipeline.

For a pipeline with two stages and gradient_accumulation_steps = 8, the last stage receives the expected 1/8 scaling, but the first stage receives an additional 1/8 scaling and therefore gets 1/64 of the accumulated gradient.

More generally, for P pipeline stages and gradient accumulation factor G, the gradient at an earlier stage appears to be scaled as:

last stage:       1 / GAS
previous stage:   1 / GAS^2
...

**To Reproduce**
Under GAS>1 and PP>1 situations. 

**ds_report output**
The issue was observed with DeepSpeed version: 0.19.2

**Additional context**
1. `PipelineEngine._exec_forward_pass()` calls `super().forward(inputs)` on every pipeline stage.
2. `DeepSpeedEngine.forward()` registers an output backward hook on each stage output.
3. `_backward_prologue_per_tensor()` divides the incoming output gradient by gradient_accumulation_steps.
4. The last stage performs backward from the loss.
5. Each non-last stage receives the downstream gradient and calls `out.backward(gradient=received_grad)`.
6. The output hook on that stage divides the already-scaled gradient by `gradient_accumulation_steps` again.

Functionhx:
## Summary

Fix repeated gradient accumulation scaling when pipeline parallelism is combined with gradient accumulation.

Every pipeline stage registers an output backward hook that scales gradients by `gradient_accumulation_steps`. The last stage therefore scales the gradient correctly before sending it upstream, but each preceding stage scales the already-scaled gradient again.

Fixes #8152

## Changes

- Apply gradient accumulation scaling only at the last pipeline stage.
- Preserve the existing behavior for single-stage pipelines and `gradient_accumulation_steps=1`.
- Add a deterministic two-stage regression test that verifies both stages receive correctly averaged gradients.
- Disable gradient clipping in the regression test so it cannot mask the scaling difference.

## Testing

- `DS_ACCELERATOR=cpu LOCAL_SIZE=2 OMP_NUM_THREADS=1 pytest -q tests/unit/runtime/pipe/test_pipe.py::TestPipeGradientAccumulationScaling::test_gradients_are_scaled_once`
- `pre-commit run --files deepspeed/runtime/pipe/engine.py tests/unit/runtime/pipe/test_pipe.py`
- Two-process CPU/Gloo numerical comparison for `gradient_accumulation_steps=1,2,8`

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -534,6 +534,12 @@ class PipelineEngine(DeepSpeedEngine):
         """True if this process is in the last stage in the pipeline."""
         return self.stage_id == self.num_stages - 1
 
+    def _backward_prologue_per_tensor(self, grad):
+        # The last stage applies GAS scaling before sending gradients upstream.
+        if self.is_last_stage():
+            return super()._backward_prologue_per_tensor(grad)
+        return grad
+
     def get_pipeline_parallel_rank(self):
         return self.stage_id
 
--- a/tests/unit/runtime/pipe/test_pipe.py
+++ b/tests/unit/runtime/pipe/test_pipe.py
@@ -13,6 +13,7 @@ import deepspeed
 import deepspeed.comm as dist
 from deepspeed.runtime.pipe.topology import PipeDataParallelTopology
 from deepspeed.runtime.pipe.module import PipelineModule
+from deepspeed.utils import RepeatingLoader
 from unit.alexnet_model import AlexNetPipe, train_cifar
 from unit.common import DistributedTest
 from unit.util import skip_on_arch, no_child_process_in_deepspeed_io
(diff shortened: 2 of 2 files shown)


tohtana (approved):
Hi @Functionhx,
Thank you for submitting this PR! This looks good to me.
I know this type of bug is very hard to find and fix. I appreciate your continuous contribution.

PKUWZP (approved):
LGTM, thanks for the PR!

PR #8157, opened 2026-07-19
Fix DeepCompile ZeRO-3 gathered parameter ownership

tohtana:
## Problem

DeepCompile inserts ZeRO-3 parameter all-gather and release operations into compiled graphs. When Dynamo skips a frame because of a graph break, however, that frame executes eagerly and does not run those graph operations. The eager fallback introduced in #8059 handles this case by all-gathering a partitioned parameter when the skipped frame accesses it through `ZeROOrderedDict`.

The fallback is enabled around `DeepSpeedEngine.forward()`. Dynamo guard evaluation occurs inside that outer forward context and also resolves parameters through `ZeROOrderedDict`, while `torch.compiler.is_compiling()` is false. The fallback could therefore mistake a guard lookup for actual eager execution and unnecessarily all-gather the parameter.

Parameters gathered by the fallback are normally partitioned after backward, but that cleanup does not run when backward is skipped. A fallback-gathered parameter may also be passed to an explicit `GatheredParameters` context, which must keep the full tensor available until the context exits.

## Why it matters

These cases require different behavior:

- Dynamo guard evaluation should not trigger an all-gather.
- A parameter gathered for an eagerly executed frame must remain available through backward and then be partitioned.
- If backward does not run, a leftover full parameter must be partitioned before the next outermost forward.
- A parameter covered by `GatheredParameters` must remain fully gathered until that context exits.

Without distinguishing these cases, a full parameter can remain allocated into a later forward, or fallback cleanup can partition it while a `GatheredParameters` block is still using it.

## Solution

This PR:

- detects parameter access during Dynamo guard evaluation and skips the eager fallback all-gather;
- partitions leftover nonpersistent full parameters before the next outermost forward when the normal post-backward cleanup did not run;
- removes a parameter from fallback cleanup when it is passed to `GatheredParameters`, so that context alone partitions it on exit;
- restores the `GatheredParameters` state even when context exit raises; and
- rejects nested `GatheredParameters` contexts that overlap on the same parameter, while continuing to allow nesting over disjoint parameter sets.


deepspeed/compile/z3_eager_fallback.py
```diff
@@ -37,22 +51,31 @@ def deepcompile_z3_forward_context(engine):
...
     def forward_context(self):
+        """Enable fallback lookup for the outermost forward and restore nested state on exit."""
         global _ACTIVE_FALLBACK
         previous = _ACTIVE_FALLBACK
         self._depth += 1
         if self._depth == 1:
             self._last_gathered_param_ids = []
+            self._last_guard_suppressed_param_ids = []
+            self._last_pre_forward_released_param_ids = []
+            self.release_available_params_for_next_forward()
```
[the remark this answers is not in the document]
tohtana: Addressed @codex

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -2364,13 +2365,37 @@ def load(module: nn.Module, prefix=""):
...
+            int(param.ds_id) for param in self.params if getattr(param, _GATHERED_PARAM_CONTEXT_DEPTH_ATTR, 0) > 0
+        ]
+        if overlapping_param_ids:
+            raise RuntimeError("Nested GatheredParameters contexts cannot overlap parameters; "
+                               f"parameter ds_ids already gathered by an outer context: {overlapping_param_ids}")
         self.params[0].all_gather(param_list=self.params)
+        for param in self.params:
+            depth = getattr(param, _GATHERED_PARAM_CONTEXT_DEPTH_ATTR, 0)
+            setattr(param, _GATHERED_PARAM_CONTEXT_DEPTH_ATTR, depth + 1)
+            fallback_owner = getattr(param, "_deepcompile_z3_eager_fallback_owner", None)
```
pengdurice: let's import and use _FALLBACK_OWNER_ATTR?

deepspeed/compile/z3_eager_fallback.py
```diff
@@ -25,34 +27,59 @@ def record_z3_eager_fallback_param(param):
     return True
 
 
+def is_dynamo_guard_evaluation():
+    """Return whether the current parameter access originates from Dynamo guard evaluation."""
+    # This intentionally depends on private Dynamo stack layout. Real GuardBuilder source resolution was verified with
+    # official CPU builds 2.8.0+cpu, 2.9.1+cpu, 2.10.0+cpu, 2.11.0+cpu, 2.12.0+cpu, and 2.13.0+cpu.
+    frame = sys._getframe()
+    while frame is not None:
+        if frame.f_globals.get("__name__") == "torch._dynamo.guards":
```
pengdurice, line 39: is "torch._dynamo.guards" prone to change e.g. when torch version changes?

deepspeed/runtime/zero/partition_parameters.py
tohtana: Good catch! Fixed.

deepspeed/compile/z3_eager_fallback.py
tohtana: Yes, it can change. I could not find a public API for detecting guard evaluation specifically. `torch.compiler.is_compiling()` returns false in this path. However, it still works even we fail to find `torch._dynamo.guards`.

If the module path changes, this check returns false and the guard access is treated as a normal eager fallback access. That causes an unnecessary all-gather and may increase communication and memory usage, but it does not skip an all-gather required by actual eager execution.

pengdurice (commented):
General question, if some one does this:

```
loss = engine(input)
with GatheredParameters([W]):
    log(W.norm())
 engine.backward(loss)
```
will that cause issue since the GatheredParameters context manager already partitions the parameters when it exits?

tohtana:
@pengdurice Great catch! Yes, in this pattern, the parameter can be partitioned too early.

I updated the PR branch to address it. Instead of transferring the ownership of the fallback and the gathered parameter, we now records claims of the ownership as a set (it is like a reference count). The parameter is partitioned only after the last claim is released.

deepspeed/compile/z3_eager_fallback.py
```diff
@@ -4,11 +4,16 @@
...
 from contextlib import contextmanager
+import sys
 
 import torch
 
 from deepspeed.runtime.zero.parameter_offload import ZeROOrderedDict, ensure_zero_ordered_dict
-from deepspeed.runtime.zero.partition_parameters import ZeroParamStatus
+from deepspeed.runtime.zero.partition_parameters import (
+    _FALLBACK_OWNER_ATTR,
+    _GATHERED_PARAM_CONTEXT_DEPTH_ATTR,
```
sfc-gh-truwase: Consider prefixing attributes with `DS_` to avoid collision and make DS-ownership explicit.

deepspeed/compile/z3_eager_fallback.py
```diff
@@ -75,25 +110,169 @@ def _disable_forward_fallback(self):
...
         self._tracked_params[ds_id] = param
+        setattr(param, _FALLBACK_OWNER_ATTR, self)
+        self._current_forward_param_ids.add(ds_id)
         self._last_gathered_param_ids.append(ds_id)
         self.total_gathered_params += 1
 
+    def record_param_access(self, param):
+        """Associate an already-gathered fallback parameter with the current forward."""
+        ds_id = int(param.ds_id)
+        if self._tracked_params.get(ds_id) is param:
```
sfc-gh-truwase: 1. Why is int cast needed since `param.ds_id` is already an integer?
2. L116 that updates the dict does not use an int cast
3. In what cases could L125 return `False`?

deepspeed/compile/z3_eager_fallback.py
```diff
@@ -25,34 +30,64 @@ def record_z3_eager_fallback_param(param):
...
     def forward_context(self):
+        """Enable fallback lookup for the outermost forward and restore nested state on exit."""
         global _ACTIVE_FALLBACK
         previous = _ACTIVE_FALLBACK
         self._depth += 1
         if self._depth == 1:
             self._last_gathered_param_ids = []
+            self._last_guard_suppressed_param_ids = []
+            self._last_pre_forward_released_param_ids = []
+            self._current_forward_param_ids = set()
```
sfc-gh-truwase: Why not `.clear()` instead of allocations?

deepspeed/runtime/zero/parameter_offload.py
```diff
@@ -61,10 +61,22 @@ def __getitem__(self, key):
         if param is None:
             return param
 
+        is_eager_forward = self._parent_module._parameters._in_forward and not torch.compiler.is_compiling()
+        fallback = None
+        if hasattr(param, "ds_status") and is_eager_forward:
+            from deepspeed.compile.z3_eager_fallback import get_active_z3_eager_fallback, is_dynamo_guard_evaluation
```
sfc-gh-truwase, line 67: Does fallback tracking incur any runtime costs in eager mode?

tests/unit/v1/compile/test_z3_eager_fallback.py
```diff
@@ -0,0 +1,568 @@
+# Copyright (c) DeepSpeed Team.
```
sfc-gh-truwase, line 1: tests/unit/v1/compile

deepspeed/compile/z3_eager_fallback.py
tohtana: Added the prefix to these values.

deepspeed/compile/z3_eager_fallback.py
tohtana: Replaced them with `.clear()`.

tests/unit/v1/compile/test_z3_eager_fallback.py
tohtana: Moved the tests.

deepspeed/runtime/zero/parameter_offload.py
tohtana: I think it is negligible. The overhead monitored in my environment was 0.50 µs for each parameter access.

deepspeed/compile/z3_eager_fallback.py
tohtana: Thank you for pointing it out, removed those cast.
For 3, it can be false when a parameter is `AVAILABLE` but is not owned by this fallback instance. For example, it would be False when the parameter may already have been gathered by `GatheredParameters` context before this fallback starts.

pengdurice (approved):
thank you for so much for the effort!

sfc-gh-truwase (approved):

PR #8158, opened 2026-07-19
Fix ZeRO-3 all_reduce param fetch stride for padded parameters

ebarkhordar:
## Problem

In `Init._all_gather_coalesced`, the branch taken when
`stage3_use_all_reduce_for_fetch_params` is enabled sizes the flat buffer with the
*aligned* element count but walks it with the *unaligned* one:

```python
flat_buffer_size = sum(p.ds_numel_aligned for p in params)   # aligned
...
    start_param += param.ds_numel                            # unaligned
```

`ds_numel_aligned` is `ds_numel` rounded up to a multiple of the partition world size, so
the two are equal only when `ds_numel % world_size == 0`. When a parameter is padded, every
following parameter's base offset lands `padding_i` slots inside its predecessor's aligned
span.

Those overlapping slots are the predecessor's partition padding. They are not harmless:
`_partition_param` allocates the partition with `torch.empty` and, on the one rank whose
partition extends past the end of the parameter, copies only `elems_to_copy` elements, so
the tail is never written and holds whatever the allocator returned. Because the fetch is
an `all_reduce` with SUM, that uninitialized tail is added into the next parameter's
leading elements. The sibling `all_gather` path writes the same padding but each parameter
narrows only its own `ds_numel` out of the result, so it is never read; the corruption is
specific to the SUM reconstruction.

Total stride is smaller than the allocated buffer, so there is no out-of-bounds access and
nothing crashes. The parameter simply comes back with wrong leading values.

## Invariant

Each parameter owns a disjoint contiguous span of `ds_numel_aligned` slots in `flat_tensor`,
which is exactly what `flat_buffer_size` already reserves for it.

`ds_numel_aligned` has a single writer in the package, `param.ds_numel_aligned = tensor_size`
in `_partition_param`, set in the same block that creates `ds_tensor`. It has two readers:
the buffer sizing above and the line changed here. Since the buffer sizing already reads the
attribute for every param in this same loop, the change introduces no new requirement on
when the attribute must exist.

## Fix

Advance `start_param` by `ds_numel_aligned`, matching the buffer sizing.

## Verification

Added `tests/unit/runtime/zero/test_zero_allreduce_fetch_params.py`, parametrized over a
padded case (numels 5 and 7 under `world_size=2`) and an aligned case (numels 4 and 6):

| case | master | this branch |
|---|---|---|
| padded | fails | passes |
| aligned | passes | passes |

On master the padded case reports:

```
param p1 was not reconstructed exactly:
  expected [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0]
  got      [7778.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0]
```

7778 is 7777 (the sentinel filling `p0`'s uninitialized padding) plus 1.0 (`p1`'s real
first element), which is the SUM overlap rather than a generically wrong value.

About that sentinel: the padding tail is uninitialized by construction, and a fresh
allocation usually reads back as zero, in which case the SUM adds zero and the bug is
invisible. A first attempt at this reproduction showed no corruption for exactly that
reason. The test therefore fills new allocations with a sentinel so the tail's contents are
deterministic rather than dependent on whether the allocator hands back a recycled block.
It does not inject anything into the buffer under test.

Run in a CPU-only container, world_size 2 over gloo, `LOCAL_SIZE=2`:

```
python -m pytest unit/runtime/zero/test_zero_allreduce_fetch_params.py
```

`pre-commit run --files` passes on both changed files.

## What I did not verify

I have no multi-GPU machine here, so this was exercised only on CPU with gloo, not with
NCCL on real devices. I also did not measure the effect on a full training run, so I cannot
say how the corrupted leading elements affect convergence in practice; the claim here is
limited to the reconstructed parameter values being wrong.


ebarkhordar:
Good catch, this is real and I have pushed a fix in 409ccd71.

`_remove_tensor_creation_wrappers` (partition_parameters.py:659) reads the module global at exit time, so `Init.__exit__` rebound the public `torch.empty` to the test's `poisoned_empty` before the `finally` block restored `_orig_torch_empty`. Measured it directly against this branch:

```
torch.empty is real_empty : False
torch.empty is poisoned   : True
post-block torch.empty(3) : [7777.0, 7777.0, 7777.0]
```

Every later allocation in that worker process would have come back sentinel-filled. The fix saves the public binding before the block and restores it alongside the module global; the same probe after the change reports `torch.empty restored: True` and a normal uninitialized tensor.

The poisoning inside the block is unchanged, so the padded/aligned differential the PR rests on still holds.

tohtana (approved):
Hi @ebarkhordar,
Great catch! It's surprising that we still have such a bug. I really appreciate your contribution.

PKUWZP (changes requested):
@ebarkhordar Thanks so much for this PR! In order to merge the PR, it would be ideal to run some tests on multi-GPU settings. I can help run some tests on my end, can you give me a 1-2 days so we can fully validate the fix and merge? Thanks for your contribution to DeepSpeed.

ebarkhordar:
Thanks @tohtana, and thanks @PKUWZP for offering to run the multi-GPU validation. That plan works for me, please take the time you need. I don't have a multi-GPU box, so I verified the fix single-process only (the unit test simulates world_size=2 with padded numels); a real multi-rank run on the all_reduce fetch path is exactly the check it should have before merging, so I would rather you confirm it than merge on my word. Happy to follow up on anything the runs turn up.

On the red checks, so they do not muddy the picture: the two failing modal-torch-latest legs are the base workflow's pull_request_target checkout refusing fork code on the re-run ("Refusing to check out fork pull request code from a 'pull_request_target' workflow"), not the change. The same commit (409ccd71) passed the full modal CI on the first run.

ebarkhordar:
Thanks @tohtana and @PKUWZP. No rush at all, take the days you need. I verified the stride fix with a 2-rank test locally (test_zero_allreduce_fetch_params.py), but I do not have real multi-GPU hardware here, so your validation is the right gate. Happy to add cases or adjust if your runs surface anything.

ebarkhordar:
@PKUWZP whenever the multi-GPU validation fits into your schedule, this is still ready on my side. No rush, and happy to add test cases or rebase onto current master if either would help.

sfc-gh-truwase (approved):
Ran the outstanding multi-GPU validation on real hardware (2× GPU, **NCCL**, `world_size=2`) — the check that was still open since the author only had CPU/gloo. I checked out this PR head (`d1f72e59`) in an isolated worktree and ran its own test, `tests/unit/runtime/zero/test_zero_allreduce_fetch_params.py`, both **with** the fix and **without** it (reverting `ds_numel_aligned → ds_numel` on the same NCCL setup) to confirm the bug actually reproduces on-device and that the fix resolves it.

| case | with fix (`ds_numel_aligned`) | without fix (`ds_numel`) |
| --- | --- | --- |
| padded (numels 5,7) | **PASS** | **FAIL** |
| aligned (numels 4,6) | PASS | PASS |

Without the fix, the padded case fails over NCCL exactly as described:

```
AssertionError: param p1 was not reconstructed exactly:
  expected [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0]
  got      [7778.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0]
```

`7778 = 7777` (sentinel in `p0`'s uninitialized partition tail) `+ 1.0` (`p1`'s real first element), i.e. the SUM `all_reduce` folds the predecessor's padding into the next parameter's leading slot. Advancing `start_param` by `ds_numel_aligned` eliminates the overlap.

Conclusion: the corruption reproduces on real multi-GPU/NCCL (not just CPU/gloo), the fix resolves it, and the aligned case is unaffected in both directions (no regression). This satisfies the multi-GPU validation gate. Thanks for the clean root-cause writeup, @ebarkhordar.

sfc-gh-truwase:
@PKUWZP the multi-GPU validation you asked for is now done. I ran the fix on real hardware (2× GPU, NCCL, `world_size=2`) using this PR's own test, both with and without the fix:

| case | with fix | without fix |
| --- | --- | --- |
| padded (numels 5,7) | PASS | FAIL |
| aligned (numels 4,6) | PASS | PASS |

Without the fix the padded case fails over NCCL exactly as described (`p1` reconstructs as `[7778.0, 2.0, ...]` vs expected `[1.0, 2.0, ...]`, the SUM overlap of `p0`'s uninitialized padding); with the fix both cases pass and the aligned case is unaffected in both directions. Details are in my approving review above. This should clear the multi-GPU gate — could you re-review so we can dismiss the requested-changes block and merge?

ebarkhordar:
Thank you, this is exactly the check I could not run. Reverting `ds_numel_aligned` back to `ds_numel` on the same NCCL setup is what makes it conclusive: it shows the corruption is real on-device rather than an artifact of the gloo simulation, and the aligned case staying green in both directions rules out a regression.

One update on my earlier caveat about the red `modal-torch-latest` legs: they are green now. All 13 checks on this head read SUCCESS as of a few minutes ago.

@PKUWZP the multi-GPU validation you asked for is done, so the changes-requested review from 2026-07-20 is the last thing holding merge. Happy to rebase onto current master if that would help.

PKUWZP (approved):
LGTM, thanks for the multi-GPU tests by @sfc-gh-truwase

ebarkhordar:
Thanks @tohtana, @PKUWZP and @sfc-gh-truwase. The multi-GPU NCCL run was the one check I could not do here, and having the padded case fail identically on device and under gloo is what made the offset arithmetic the clear culprit rather than anything backend specific.

PR #8164, opened 2026-07-21
NVMe write warning

sfc-gh-truwase:
- Add write endurance warning
- Move UTs


stas00 (approved):
Looking good, Tunji

I'd also add a note to the online docs

PR #8168, opened 2026-07-23
[AutoTP] Fix ZeRO-3 checkpoint consolidation to gather across TP and DP

delock:
AutoTP + ZeRO-3 silently produced incomplete checkpoints: both export paths handled only the ZeRO data-parallel dimension and dropped the tensor-parallel shards.

- ds_to_universal.py: stage3 conversion recovers the (tp,dp) grid from checkpoint file names, extracts shards under the real tp_index, and reuses the stage<=2 TP-aware merge when tp_degree>1 (DP-only path preserved for tp_degree==1 -> no regression for plain ZeRO-3).
- engine.py: _zero3_consolidated_16bit_state_dict nests GatherReplacedLayerParams inside GatheredParameters so save_16bit_model gathers both DP and TP; remove the blanket autotp+zero3 training block now that checkpoint consolidation is implemented.
- stage3.py: load_hp_checkpoint_state resolves the TP shard before the ZeRO-DP partition, so universal checkpoint restore round-trips.

Add end-to-end universal conversion tests and update existing tests for the refactored merge_tp_slices / extract_zero_shards_stage3 signatures.


deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -284,12 +303,12 @@ def get_matched_sub_params_pattern(name_):
...
     matched_sub_params_shape = get_matched_sub_params_pattern(name)
 
-    step_merged = _merge_zero_shards(slice_base_path, "step", tp_degree, shape)
+    step_merged = _merge_zero_shards(slice_base_path, "step", tp_degree, per_tp_shapes)
     if step_merged:
         _save_checkpoint(os.path.join(param_base_path, "step.pt"), step_merged[0])
 
     for state in ("fp32", "exp_avg", "exp_avg_sq"):
-        slices = _merge_zero_shards(slice_base_path, state, tp_degree, shape)
+        slices = _merge_zero_shards(slice_base_path, state, tp_degree, per_tp_shapes)
```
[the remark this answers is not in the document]
delock: DeepSpeed's own AutoTP gather/partition infrastructure already blocks this case: `_gather_logical_tensor` (layers.py L1207) and `_partition_logical_tensor` (layers.py L1161) both raise `ValueError` when a sub-parameter size is not divisible by `tp_size`, so a model with non-divisible kV heads cannot be created under AutoTP in the first place. The per-tp-shape fix in the second commit addresses the real uneven scenario (a single parameter whose partition dim isn't divisible by TP, e.g. vocab size), which is the case that can actually occur. This sub-param path is a valid observation for a follow-up but doesn't block this PR.

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -779,20 +843,21 @@ def main(args):
...
+        # Keep one shape list per param so uneven TP shards merge to the correct per-rank shape.
+        slice_shapes_by_tp = []
         for mp_rank_file in ds_checkpoint.mp_rank_files:
             mp_sd = torch.load(mp_rank_file, map_location=torch.device('cpu'), weights_only=False)
-            slice_shapes += mp_sd[PARAM_SHAPES]
-
-        # fix back to normal flat dict, merge duplicates for tp>1
-        slice_shapes = dict((k, v) for d in slice_shapes for k, v in d.items())
+            slice_shapes_by_tp.append(dict((k, v) for d in mp_sd[PARAM_SHAPES] for k, v in d.items()))
+        slice_shapes = {name: [d[name] for d in slice_shapes_by_tp] for name in slice_shapes_by_tp[0]}
```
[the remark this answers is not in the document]
delock: Thanks -- confirmed this was a regression from the per-TP refactor. Fixed in the latest commit: extracted a `_group_per_tp_shapes` helper that groups mp_rank_files by TP rank (they are tp-major) and unions PP stages within each TP rank, so PP-local params are preserved. Added `test_group_per_tp_shapes_handles_pp_local_params` covering the exact failure scenario (tp=2, pp=2, uneven TP shards, params only in pp0 / only in pp1).

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -659,6 +715,31 @@ def _parse_model_states_stage3(files):
...
+def _group_per_tp_shapes(slice_shapes_by_tp, pp_degree, tp_degree):
+    # mp_rank_files are tp-major (tp0_pp0, tp0_pp1, ..., tp1_pp0, ...).
+    # Group slice_shapes_by_tp entries by TP rank, union PP stages within each
+    # TP rank so that PP-local parameters are not lost, then build one per-TP
+    # shape list per param name.
+    per_tp = []
+    for tp in range(tp_degree):
+        tp_dict = {}
+        for pp in range(pp_degree):
+            tp_dict.update(slice_shapes_by_tp[tp * pp_degree + pp])
```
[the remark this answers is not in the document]
delock: Done, fixed and add a new test that actually test the contract, instead of test on fake data that originate from same wrong understanding.

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -856,22 +945,48 @@ def main(args):
...
+        if tp_degree > 1:
+            # Reassemble both the ZeRO data-parallel and tensor-parallel dimensions by
+            # reusing the stage<=2 TP-aware merge machinery. Use the per-TP-rank shapes from
+            # param_shapes_grid (not a single rank's shape) so uneven TP shards -- e.g. a
+            # partition dim not divisible by tp_degree, or uneven GQA sub-params -- merge
+            # under their correct per-rank shape instead of the first rank's shape.
+            param_shapes_by_tp = [
+                dict((k, v) for sub in grid_tp for k, v in sub.items()) for grid_tp in param_shapes_grid
+            ]
+            slice_shapes = {name: [param_shapes_by_tp[tp][name] for tp in range(tp_degree)] for name in param_keys}
```
tohtana, line 984: I think parameters that AutoTP leaves unchanged (and therefore lack AutoTP conversion metadata) should be classified as TP-replicated rather than omitted from the conversion schema. Otherwise, LayerNorm/RMSNorm weights fall through to the default dim-0 concatenation and are incorrectly expanded from `[H]` to `[H * tp_degree]`.

deepspeed/runtime/engine.py
```diff
@@ -628,11 +628,10 @@ def _configure_tensor_parallel_states(self, model):
...
         self._set_client_model(model)
-        if self.zero_optimization_stage() > 2 and (self.client_optimizer or self.optimizer_name()):
-            raise NotImplementedError("AutoTP together with ZeRO stage 3 is currently supported only for inference "
-                                      "(no optimizer). Running it with an optimizer is disabled because TP-aware "
-                                      "checkpoint consolidation is not yet implemented and would silently produce "
-                                      "incomplete checkpoints.")
+        # AutoTP + ZeRO stage 3 is supported for both inference and training. The
+        # checkpoint paths (16-bit consolidated export and universal checkpoint
+        # conversion) gather across both the ZeRO data-parallel and the tensor-parallel
+        # dimensions, so saved/converted weights are complete.
```
tohtana, line 634: We can now remove `TestAutoTpZeroStage3.test_autotp_zero3_training_blocked`.

deepspeed/runtime/engine.py
```diff
@@ -5352,34 +5351,40 @@ def _zero3_consolidated_16bit_state_dict(self, exclude_frozen_parameters=False):
 
         state_dict = OrderedDict() if dist.get_rank() == 0 else None
         shared_params = {}
+        # GatheredParameters (below) merges only the ZeRO data-parallel partition. When
```
tohtana, line 5354: The `GatheredParameters` context materializes a TP shard, and `GatherReplacedLayerParams.gather_params()` retains that tensor through `param.data_partition`. Its exit path changes `param.data` back to a TP shard but does not clear `data_partition`.
 So, the layer-wise export can retain gathered storage after the context exits.

docs/_tutorials/autotp-training.md
```diff
@@ -224,9 +224,9 @@ For Grouped Query Attention with different Q/K/V sizes:
 
 ## Limitations
 
-1. **ZeRO Stage 3 training not supported**: AutoTP with ZeRO stage 3 is supported only for inference (no optimizer). Training with an optimizer is blocked until TP-aware checkpoint consolidation is implemented.
+1. **TP size must divide model dimensions**: The tensor parallel size must evenly divide the attention head count and hidden dimensions for the runtime TP math to be correct. (Checkpoint conversion additionally handles uneven parameter shards -- e.g. a non-divisible partition dimension or uneven fused/GQA sub-parameters -- so save/convert no longer require every dimension to be divisible.)
```
tohtana: The `matched_sub_params_shape` path uses floor division for both the shared partition shape and each per-rank sub-parameter size. So genuinely uneven per-rank sub-parameter partitions are not handled.

tohtana (changes requested):
Hi @delock,
This is truly a great enhancement of DeepSpeed core feature. I left some comments about the detailed.

deepspeed/runtime/engine.py
delock: `data_partition` looks like dead code and needs to be removed.  Thanks for the catch!  Also @inkcherry

docs/_tutorials/autotp-training.md
delock: Good catch!  I'll remove the claim for fused/GQA sub-parameters and wait for #8185 to support uneven subparam sharding.

delock:
@tohtana thank you for your comments!  All comments had bene addressed.  Thanks!

deepspeed/runtime/engine.py
inkcherry: Agree, safe to remove.

tohtana (approved):
Thank you @delock for the update!
This looks good to me. This is an amazing update.

PR #8169, opened 2026-07-23
Stabilize DeepCompile ZeRO-3 memory scheduling

tohtana:
Depends on #8159.

## Overview

DeepCompile's ZeRO-3 scheduler does not consistently account for memory pressure across gather/release ordering, prefetch, selective gathering, and native gather-buffer reuse. This can retain excess gathered storage or trigger a full-parameter gather while Dynamo evaluates guards.

This PR adds rank-consistent scheduler budgeting and diagnostics, coordinates prefetch and selective gathering with graph profiling, reuses native gather storage within a bounded pressure-aware lifecycle, and avoids guard-time full-parameter gathers while the DeepCompile eager fallback is active.

## Results

The end-to-end measurements were collected.

- Model: Qwen3-14B
- Hardware: one node with 8 H100 GPUs
- Activation checkpointing: enabled
- Micro-batch size: 4
- Sequence length: 4096
- Measurement: 10 warmup iterations followed by 20 measured iterations

| Mode | Allocator retries per rank | Average step time | Throughput |
| --- | --- | --- | --- |
| ZeRO-3-only pass | 18–19 → 0 | 8.4% faster | 9.2% higher |
| Full DeepCompile pass | 20 → 2–4 | 9.2% faster | 10.1% higher |

In a longer observation window covering steps 31–300, no allocator retries occurred.


deepspeed/utils/allocator_telemetry.py
```diff
@@ -0,0 +1,183 @@
+# Copyright (c) DeepSpeed Team.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+"""Small opt-in allocator diagnostics for the DeepCompile memory experiment."""
```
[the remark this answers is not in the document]
tohtana: Addressed

csrc/compile/init.cpp
```diff
@@ -105,6 +105,15 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m)
     m.def("register_graph_z3",
           &dc::register_graph_z3,
           "Register graph with a list of ds parameter ids");
+    m.def("set_z3_gather_buffer_pool_budget_for_test",
```
pengdurice, line 108: wondering if these new items are just test purpose only?

deepspeed/compile/inductor.py
```diff
@@ -21,6 +22,37 @@
...
 from .partitioner import get_wrapped_partitioner
 
+_DEEP_COMPILE_Z3_INDUCTOR_REDUCTION_CONFIG = {
+    "triton.mix_order_reduction": False,
+    "triton.persistent_reductions": False,
+}
+
+
+def deepcompile_z3_inductor_config_patch(enabled: bool):
+    """Disable reduction heuristics that create oversized kernels for DeepCompile ZeRO-3 graphs."""
```
eternalNight, line 38: How does inductor-generated mix-order or persistent reductions bother ZeRO-3? Is that related to GPU memory use?

deepspeed/compile/util.py
```diff
@@ -90,6 +90,24 @@ def log_rank0(msg: str, enable: bool = False):
         print(msg)
 
 
+def all_reduce(tensor, op, process_group=None):
+    if process_group is None:
+        return dist.all_reduce(tensor, op)
+    return dist.all_reduce(tensor, op, group=process_group)
```
eternalNight: That wrapper is unnecessary. `dist.all_reduce` already takes `group` as optional and default to `None` (which means the default process group will be used). Same to `get_rank` and `barrier`.

csrc/compile/z3.cpp
```diff
@@ -7,16 +7,640 @@
...
+                       at::ScalarType dtype,
+                       const c10::Device& device,
+                       at::cuda::CUDAStream stream)
+    {
+        flushCompletedPressureRecovery();
+        observeAllocatorPressure(device.index());
+        if (!enabled_ || pressure_recovery_in_progress_) { return at::Tensor(); }
+
+        Entry* best = nullptr;
+        for (auto& entry : entries_) {
```
pengdurice, line 53: is entries_  ordered? if not, is the full loop expensive vs an max heap like solution?

pengdurice:
general question, do we have experiments results showcasing the changes' benefits?

csrc/compile/init.cpp
tohtana: Yes, these three new APIs are for testing purpose only. We need them to set or inspect the internal state of the memory pool. It doesn't look great to have these as public APIs, but I didn't have an idea about other approaches.

deepspeed/compile/util.py
tohtana: Thank you for pointing this out. I removed them.

csrc/compile/z3.cpp
```diff
@@ -7,16 +7,640 @@
...
+std::optional<int64_t> gather_buffer_pool_test_budget;
+
+std::shared_ptr<GatherBufferPool> get_gather_buffer_pool()
+{
+    auto pool = weak_gather_buffer_pool.lock();
+    if (!pool) {
+        pool = std::make_shared<GatherBufferPool>();
+        weak_gather_buffer_pool = pool;
+        if (gather_buffer_pool_test_budget) {
+            pool->setBudgetForTest(gather_buffer_pool_test_budget.value());
```
pengdurice, line 605: setBudgetForTest is only called when it is test purpose?

csrc/compile/init.cpp
pengdurice: just curious if it is possible to do integration test rather than unit test on this level? For example, instead of writing the scaffold to test GatherBufferPool, can do a test of functions that call the natural creation of this class, e.g. calling from Z3CustomOpExecutor?

deepspeed/compile/inductor.py
tohtana: Great question, this is a workaround for an issue we encountered.

For a ZeRO-3 DeepCompile backward graph, we observed that Inductor generated a fused persistent reduction that exceeded the H100 per-kernel resource limit (295000 bytes required versus 232448 available).
Disabling `persistent_reductions` alone was insufficient because the mix-order path explicitly forces persistent reduction, so we had to disable both heuristics must be disabled.

I added a comment to describe this.

csrc/compile/z3.cpp
tohtana: `entries_` is currently an unsorted `std::vector` as we don't simply choose the maximum one (best fit to capacity, LRU for eviction, search by storage identity, etc.)
However, I don't think the scan is significant overhead. I expect the pool to contain tens of entries rather than thousands because it is byte-bounded and the dominant gather buffers are relatively large. Given an 80GB GPU, the hard cap is `80 GiB / 32 = 2.5 GiB`. If we the average buffer size is 64MB, we will have 80 entries. This should be negligible in the C++ code.

tohtana:
> general question, do we have experiments results showcasing the changes' benefits?

Yes. We have some measurements although I have not rerun the exact current head yet.

The experiment used
- Qwen3-14B
- One node with 8 H100s
- Activation checkpointing enabled
- Micro-batch 4 
- Sequence length 4096
- 10 warmup and 20 measured iterations

The results are:
- With the ZeRO-3-only pass, the measured PyTorch caching-allocator retry count fell from 18–19 retries per rank to zero on every rank. Average step time improved by 8.4%, and throughput improved by 9.2%.
- With the full DeepCompile pass, the retry count fell from 20 per rank to 2–4 per rank. Average step time improved by 9.2%, and throughput improved by 10.1%.
For a longer run, from 31-300 steps, no allocator retry happened.

csrc/compile/z3.cpp
tohtana: Yes. `gather_buffer_pool_test_budget` is unset by default and is populated only by `set_z3_gather_buffer_pool_budget_for_test()`.

csrc/compile/init.cpp
tohtana: Yes, that is actually how the current tests are structured. They do not construct or call `GatherBufferPool` directly. `register_graph_z3()` creates `Z3CustomOpExecutor` with the shared pool, and the CUDA test runs the real `allgather_param`, `wait_allgather`, and `release_param` ops.
The `_for_test` APIs are limited to making the budget and allocator-pressure inputs deterministic and exposing accounting of internal values.

pengdurice: OK, your last sentence explains my confusion. OK, maybe we can call it deterministic or pre determined size etc?

pengdurice:
> > general question, do we have experiments results showcasing the changes' benefits?
> 
> Yes. We have some measurements although I have not rerun the exact current head yet.
> 
> The experiment used
> 
> * Qwen3-14B
> * One node with 8 H100s
> * Activation checkpointing enabled
> * Micro-batch 4
> * Sequence length 4096
> * 10 warmup and 20 measured iterations
> 
> The results are:
> 
> * With the ZeRO-3-only pass, the measured PyTorch caching-allocator retry count fell from 18–19 retries per rank to zero on every rank. Average step time improved by 8.4%, and throughput improved by 9.2%.
> * With the full DeepCompile pass, the retry count fell from 20 per rank to 2–4 per rank. Average step time improved by 9.2%, and throughput improved by 10.1%.
>   For a longer run, from 31-300 steps, no allocator retry happened.

That's great, maybe add that to the PR's description?

pengdurice:
the gather buffer pool and the memory budget seem to be two separate optimization, worth considering splitting to to PRs?

deepspeed/compile/inductor.py
eternalNight: I see. It looks odd that inductor does not test if the fused persistent reduction is within hardware limits. But we can keep this workaround till we know what's wrong with inductor itself.

csrc/compile/z3.cpp
tohtana: Small correction to my previous example: a 2.5 GiB cap with an average 64 MiB buffer gives about 40 entries, not 80.

csrc/compile/init.cpp
tohtana: Yes. I think “fixed budget” is more precise because the test fixes the pool’s byte budget rather than individual buffer sizes. If we keep `for_test`, how about these 
- `set_z3_gather_buffer_pool_budget_for_test` -> `set_z3_gather_buffer_pool_fixed_budget_for_test`
- `setBudgetForTest` -> `setFixedBudgetForTest`
- `update_z3_gather_buffer_pool_allocator_pressure_for_test` -> `update_z3_gather_buffer_pool_allocator_pressure_for_test`

tohtana:
> the gather buffer pool and the memory budget seem to be two separate optimization, worth considering splitting to to PRs?

I split this PR and opened #8233 and #8234. Can you review these? @pengdurice
I'm closing this now.

PR #7360, opened 2025-06-16
[Bugfix][scheduler]: WarmupLR inherits optimizer lr when not specified

Flink-ddd:
This PR fixes issue #7303.

### 1. Description of the Bug

Currently, when using the WarmupLR scheduler, if warmup_max_lr is not explicitly set in the scheduler's parameters, it incorrectly falls back to its internal default value (0.001), ignoring the learning rate set in the optimizer's parameters. This can lead to unexpected training behavior and diverges from user expectations.

### 2. Description of the Fix

This fix modifies the __init__ method of the `WarmupLR` scheduler in `deepspeed/runtime/lr_schedules.py`.

- The default value for the warmup_max_lr argument in the function signature is changed from 0.001 to None.
- Logic is added to check if `warmup_max_lr` is `None` upon initialization. If it is, the scheduler now correctly inherits the learning rate from the optimizer's parameter groups.

This change ensures that the optimizer's learning rate is respected as the default warmup_max_lr, aligning the scheduler's behavior with the user's configuration intent.

### 3. Verification

The fix has been verified using a minimal reproduction script that clearly demonstrates the behavioral change.

**Before Fix:**
Without `warmup_max_lr` in the scheduler config, the learning rate incorrectly defaults to `0.001`.
Screenshot 2025-06-16 at 18 34 31

**After Workaround (Demonstrating the Mechanism):**
By explicitly adding `warmup_max_lr` to the scheduler config, the learning rate behaves as expected. My code change makes this the default behavior.
Screenshot 2025-06-16 at 20 17 11

PR #8171, opened 2026-07-24
Fix WarmupLR collapsing multi-group base LRs to group 0's

ebarkhordar:
## What

When `warmup_max_lr` is left unspecified, `WarmupLR` inherits the optimizer's learning rate (added in #7360). The fallback computed:

```python
warmup_max_lr = [group['lr'] for group in self.optimizer.param_groups][0]
```

The trailing `[0]` reduces the per-group list to group 0's scalar. `_format_param` then broadcasts that scalar back to every group (`[value] * len(param_groups)`). So on an optimizer with multiple parameter groups that have distinct base LRs, every group warms up to group 0's lr and the other groups' configured LRs are silently discarded.

## Fix

Drop the trailing `[0]` so `_format_param` receives the full per-group list and each group warms up to its own base lr. This mirrors #7969, which fixed the same multi-group collapse in the sibling `WarmupCosineLR`.

## Verification

Reproduced and verified on a CPU-only container against this branch (real `import deepspeed`, module resolved from the checkout). With two param groups at lr 0.1 and 0.2 and `warmup_max_lr` omitted:

- before: `max_lrs == [0.1, 0.1]` (group 1 collapsed to group 0)
- after: `max_lrs == [0.1, 0.2]`

Added `test_warmup_lr_inherits_per_group_lr_when_max_unspecified` in `tests/unit/runtime/test_lr_schedulers.py`, mirroring the existing `test_warmup_cosine_lr_initializes_all_param_groups`. It fails on master (`assert [0.1, 0.1] == [0.1, 0.2]`) and passes with this change. `WarmupDecayLR` defaults `warmup_max_lr=0.001`, so this path only changes behavior when the value is left unspecified.

Ran the repo's formatting hooks (yapf, flake8, codespell, license, end-of-file) on the changed files; all pass.

Note: this is a small follow-on in the same file as my open #8166 (a different scheduler class), kept to a one-line change plus one test.

### DIFF
--- a/deepspeed/runtime/lr_schedules.py
+++ b/deepspeed/runtime/lr_schedules.py
@@ -670,7 +670,7 @@ class WarmupLR(object):
         self.optimizer = get_torch_optimizer(optimizer)
 
         if warmup_max_lr is None:
-            warmup_max_lr = [group['lr'] for group in self.optimizer.param_groups][0]
+            warmup_max_lr = [group['lr'] for group in self.optimizer.param_groups]
 
         self.min_lrs = self._format_param(self.optimizer, warmup_min_lr, "min_lr")
         self.max_lrs = self._format_param(self.optimizer, warmup_max_lr, "max_lr")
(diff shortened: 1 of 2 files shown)


tohtana (approved):
Hi @ebarkhordar,
Great catch! This is a very important fix. Let's merge it.

PR #8172, opened 2026-07-24
Stabilize fork-sensitive nightly tests and AutoSP coverage

tohtana:
This PR fixes issues with the nightly full test:

- Fork-sensitive tests: Run these tests in the non-forked sequential phase
  - Heavy two-rank NVMe cases contended in the forked parallel phase
  - The CPU-only FlopsProfiler formatting test could inherit initialized CUDA state and fail in a forked subprocess
- AutoSP coverage: `TestSDPANodesCompile` inspected only the first FX graph produced from a Transformers model, but a compiled model can be split into multiple graphs and the first graph may not contain attention. The test now constructs an FX GraphModule with Q/K/V inputs and one `torch.nn.functional.scaled_dot_product_attention` call. The separate four-GPU tests still continue to use a tiny Transformers Llama model.
- Transformers version output: Nightly tests install Transformers from Git, but the workflow incorrectly displayed the unused PyPI fallback version (`4.50.0`) as if it were installed. It now reports the actually installed version and Git commit, shows unused fields as `N/A`.


delock (approved):
Looks good to me

PR #6948, opened 2025-01-14
Update sharded_moe.py to support top2 gate with Tutel

xenshinu:
Given the fact that multiple experts per token is very common, and the gather and scatter operation without Tutel is so inefficient, I added support of tutel to top2 gate and tested on pipeline engine. This can be done for any k actually, I'll push that later when I have time to test,

PR #8174, opened 2026-07-24
Enable support for Tutel when k != 1 for shared moe

jinyouzhi:
- Continue #6948 

Remove the limited on top2 and topk >2 for Tutel.


tohtana (approved):
Looks good to me, thank you @jinyouzhi!

Performance note from the validation on my side: on one 2×H100 BF16, the Tutel/native step-time ratio was 0.948 for k=2, but 1.119 for k=3. Correctness looks good so it is fine to merge, but Tutel might not have performance advantage for k=3 in this configuration.

jinyouzhi:
> Looks good to me, thank you @jinyouzhi!
> 
> Performance note from the validation on my side: on one 2×H100 BF16, the Tutel/native step-time ratio was 0.948 for k=2, but 1.119 for k=3. Correctness looks good so it is fine to merge, but Tutel might not have performance advantage for k=3 in this configuration.

Thank you for your review and perf metrics. With your insights, I further tried to optimize the k=3 case to reduce overhead.

PR #8178, opened 2026-07-25
Make PipelineModule.set_checkpoint_interval actually change the interval

vineethsaivs:
## Problem

`PipelineModule.set_checkpoint_interval()` does not change the activation checkpoint interval.

```python
def set_checkpoint_interval(self, interval):
    assert interval >= 0
    self.checkpoint_interval = interval
```

Every reader uses `self.activation_checkpoint_interval`: `forward()` branches on it and steps the layer loop by it (`module.py` L370-379), `_precompute_checkpointable_values()` keys its cache on it (L223-230), and `PipelineEngine` assigns it directly (`pipe/engine.py` L212). `self.checkpoint_interval` is read nowhere in the repo, so the call assigns a dead attribute and the schedule silently stays as it was.

Assigning the right attribute is not sufficient on its own, because the recompute path it feeds is also broken:

```python
def _precompute_checkpointable_values(self):
    if self.activation_checkpoint_interval > 0 and self.is_checkpointable_results_interval != self.activation_checkpoint_interval:
        num_layers = len(self.forward_funcs)
        self.interval_was_zero = False
        for start_idx in range(0, num_layers, self.activation_checkpoint_interval):
            ...
            self.is_checkpointable_results.append(self._is_checkpointable(funcs))
        self.is_checkpointable_results_interval = self.activation_checkpoint_interval
```

The `is_checkpointable_results_interval != activation_checkpoint_interval` guard exists precisely so the values are recomputed when the interval changes, but the loop appends to `self.is_checkpointable_results` without clearing it, so results computed for the previous interval stay at the front of the list.

`forward()` then pairs the layer blocks with that list positionally:

```python
for start_idx, is_checkpointable_result in \
    zip(range(0, num_layers, self.activation_checkpoint_interval), self.is_checkpointable_results):
```

so each block is checkpointed according to a decision made for a different partitioning of the layers.

## Reproduction

8 layers, the first 4 without parameters and the last 4 with, so `_is_checkpointable` genuinely differs per block. Going from interval 4 to interval 1:

```
interval=4  results: [False, True]

set_checkpoint_interval(1) -> activation_checkpoint_interval = 4     # unchanged
  results: [False, True]                                            # never recomputed

# assigning activation_checkpoint_interval directly, the way PipelineEngine does:
  results: [False, True, False, False, False, False, True, True, True, True]   # 10 entries for 8 blocks
  expected:               [False, False, False, False, True, True, True, True]

forward blocks : [(0, False), (1, True), (2, False), (3, False), (4, False), (5, False), (6, True), (7, True)]
expected       : [(0, False), (1, False), (2, False), (3, False), (4, True), (5, True), (6, True), (7, True)]
```

Blocks 1, 4 and 5 are checkpointed against the wrong decision: block 1 holds no parameters and is checkpointed anyway, blocks 4 and 5 hold parameters and are not.

## Fix

Clear the cached results before recomputing them, and have the setter assign `activation_checkpoint_interval` and rebuild the cache:

```python
self.is_checkpointable_results = []
```

```python
def set_checkpoint_interval(self, interval):
    assert interval >= 0
    self.activation_checkpoint_interval = interval
    self._precompute_checkpointable_values()
```

The setter has to do both. Assigning the interval alone would leave `forward()` zipping the new, longer block range against a list still sized for the old interval, and `zip` stops at the shorter one, so trailing layer blocks would be dropped from the forward pass entirely.

Nothing changes for the normal path: `PipelineEngine` assigns the interval once and calls `_precompute_checkpointable_values()` while the cache is still empty, so clearing an empty list is a no-op and the guard still skips the recompute when the interval is unchanged.

## Testing

`TestPipeModuleCheckpointInterval` in `tests/unit/pipe/test_pipe_module.py` builds a `PipelineModule` at interval 4, calls `set_checkpoint_interval(1)`, and asserts the interval is updated and the results match a module constructed at interval 1 directly. It fails on master on both assertions (the interval stays 4, and the results stay `[False, True]`) and passes with the fix. The mixed `ReLU`/`Linear` model is deliberate: with a uniformly parameterised model only the length of the list is wrong, and the misalignment would not show.

`yapf` and `flake8` are clean on both changed files.

### DIFF
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -208,16 +208,20 @@ class PipelineEngine(DeepSpeedEngine):
         self.agg_train_loss = None
         self.agg_additional_losses = None
 
+        # use_reentrant picks the module's checkpoint function, and that choice also feeds
+        # _is_checkpointable(), so resolve it whatever the configured interval is: the module's
+        # set_checkpoint_interval() can enable checkpointing later, and it would otherwise run
+        # with the reentrant default even though the config asked for non-reentrant.
+        # set use_reentrant default to True.
+        if self._config.pipeline.get('use_reentrant') is None:
+            self._config.pipeline['use_reentrant'] = True
+        if self._config.pipeline['use_reentrant'] is False:
+            # set activation_checkpoint_func to non_reentrant_checkpoint func.
+            self.module.activation_checkpoint_func = ds_checkpointing.non_reentrant_checkpoint
+            if self.grid.get_global_rank() == 0:
+                logger.info('CONFIG: activation_checkpoint_func=non_reentrant_checkpoint')
         if self._config.pipeline['activation_checkpoint_interval'] > 0:
             self.module.activation_checkpoint_interval = self._config.pipeline['activation_checkpoint_interval']
-            # set use_reentrant default to True.
-            if self._config.pipeline.get('use_reentrant') is None:
-                self._config.pipeline['use_reentrant'] = True
-            if self._config.pipeline['use_reentrant'] is False:
-                # set activation_checkpoint_func to non_reentrant_checkpoint func.
-                self.module.activation_checkpoint_func = ds_checkpointing.non_reentrant_checkpoint
-                if self.grid.get_global_rank() == 0:
-                    logger.info('CONFIG: activation_checkpoint_func=non_reentrant_checkpoint')
         if self.module.activation_checkpoint_interval > 0:
             self.module._precompute_checkpointable_values()
 
--- a/deepspeed/runtime/pipe/engine.py
+++ b/deepspeed/runtime/pipe/engine.py
@@ -891,6 +891,20 @@ class PipelineEngine(DeepSpeedEngine):
             self.timers(BACKWARD_MICRO_TIMER).stop()
             self.timers(BACKWARD_GLOBAL_TIMER).stop()
 
+    def _reentrant_activation_checkpointing(self):
+        """True when the module checkpoints activations with the reentrant function.
+
+        Reentrant checkpointing needs the first stage's inputs to require grad, or the
+        first checkpointed segment is detached from autograd. Key that off the module,
+        not off ``self._config.pipeline``: the config only seeds the module at __init__,
+        so a module built with its own ``activation_checkpoint_interval``, or one changed
+        later via ``set_checkpoint_interval()``, would leave the config stale. The forward
+        pass already branches on the module attribute, so this keeps the two in agreement.
+        """
+        if self.module.activation_checkpoint_interval <= 0:
+            return False
+        return self.module.activation_checkpoint_func is not ds_checkpointing.non_reentrant_checkpoint
+
     def _exec_load_micro_batch(self, buffer_id):
         if self.wall_clock_breakdown():
             self.timers(BATCH_INPUT_TIMER).start()
@@ -901,8 +915,7 @@ class PipelineEngine(DeepSpeedEngine):
             loaded = None
             if torch.is_tensor(batch[0]):
                 loaded = batch[0].clone().to(self.device).detach()
-                if self._config.pipeline['activation_checkpoint_interval'] > 0 and self._config.pipeline[
-                        'use_reentrant']:
+                if self._reentrant_activation_checkpointing():
                     loaded.requires_grad = loaded.is_floating_point()
             else:
                 assert isinstance(batch[0], (tuple, list))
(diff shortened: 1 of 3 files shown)


tohtana (changes requested):
Hi @vineethsaivs,
Thank you for submitting this PR! I think the comment from Codex bot is reasonable. Can you address it?

vineethsaivs:
Thanks @tohtana. Agreed, the Codex comment is a real bug and I have addressed it in cb8dae6.

I checked it against the source before fixing, and it holds:

- `_exec_load_micro_batch` gates the `requires_grad` marking on `self._config.pipeline['activation_checkpoint_interval'] > 0` (engine.py:904 and :914)
- but the pipeline config only ever seeds the module, once, in `__init__` (engine.py:211-212)
- while `PipelineModule.forward()` branches on `self.activation_checkpoint_interval`, the module attribute (module.py:370)

So once the setter really changes the interval, the module starts checkpointing while the config still reads 0, and under reentrant checkpointing the first stage's inputs stay detached. The first checkpointed segment is then cut off from autograd.

Codex offered two options: keep the config in sync, or key the decision off the module. I went with the second, for two reasons.

The first is that the module is already the source of truth everywhere else. `engine.py:221` decides whether to call `_precompute_checkpointable_values()` off `self.module.activation_checkpoint_interval`, and `forward()` branches on the same attribute. Syncing the config would add a third copy of the same state rather than remove one.

The second is that this hole is not actually new. `PipelineModule(layers=..., activation_checkpoint_interval=N)` under a pipeline config of 0 already checkpoints, via the module attribute, while the engine already declines to mark inputs. My setter made an existing inconsistency reachable a second way. Keying off the module closes both.

I also had to fold `use_reentrant` into it rather than drop it: the config only defaults that flag inside the `interval > 0` branch, so reading it directly would be stale for exactly the same reason. The module's `activation_checkpoint_func` is what `use_reentrant` ultimately selects, so the helper tests that instead, which is equivalent on the config-driven path and correct on the setter path.

```python
def _reentrant_activation_checkpointing(self):
    if self.module.activation_checkpoint_interval <= 0:
        return False
    return self.module.activation_checkpoint_func is not ds_checkpointing.non_reentrant_checkpoint
```

Two tests added to `TestPipeModuleCheckpointInterval`: one starts from a config with checkpointing disabled and asserts the flip happens only after `set_checkpoint_interval()`, the other pins that `use_reentrant: False` still does not mark inputs, so the decision tracks the function and not just the interval. yapf clean, and flake8 under the repo's own `.flake8` selection is clean.

tohtana:
Thank you for the update! The original issue Codex mentioned is now resolved.
That said, I found a new issue with this combination:

- activation_checkpoint_interval=0
- use_reentrant=false
- set_checkpoint_interval(1)

In this case, `PipelineEngine.__init__()` does not install the non-reentrant function because that assignment is inside the positive-interval branch. A later `module.set_checkpoint_interval(1)` therefore enables checkpointing with the default reentrant function.
Can you double check and address it if it is true?

vineethsaivs:
You are right, and I checked it against the source rather than taking it.

`engine.py` had the whole `use_reentrant` block nested inside `if self._config.pipeline["activation_checkpoint_interval"] > 0:`, so with a configured interval of 0 the module keeps `PipelineModule`'s default `activation_checkpoint_func`, which is the reentrant `checkpointing.checkpoint`. `set_checkpoint_interval(1)` then turns checkpointing on and `forward()` calls that function, so `use_reentrant: false` is silently ignored.

One consequence beyond the one you named: `_is_checkpointable()` (module.py:672) branches on `activation_checkpoint_func is not non_reentrant_checkpoint`, and the setter now calls `_precompute_checkpointable_values()`. So the cached results were being decided under the wrong assumption as well, not just the checkpoint call itself.

Fixed in d19e901 by hoisting the resolution above the interval assignment, so it runs whatever the configured interval is:

```python
# set use_reentrant default to True.
if self._config.pipeline.get("use_reentrant") is None:
    self._config.pipeline["use_reentrant"] = True
if self._config.pipeline["use_reentrant"] is False:
    self.module.activation_checkpoint_func = ds_checkpointing.non_reentrant_checkpoint
    ...
if self._config.pipeline["activation_checkpoint_interval"] > 0:
    self.module.activation_checkpoint_interval = self._config.pipeline["activation_checkpoint_interval"]
```

I kept it ordered so the function is still chosen before any `_precompute_checkpointable_values()` call, which the old code also relied on.

On blast radius: with an interval of 0 the function is never invoked, since `PipelineModule.forward()` returns early at module.py:373, and engine init does not precompute either. So nothing observable changes for an existing configuration until checkpointing is actually enabled, which is exactly the path being fixed.

Test added to `TestPipeModuleCheckpointInterval`: interval 0 plus `use_reentrant: False`, assert the module already holds `non_reentrant_checkpoint` after `initialize()`, then `set_checkpoint_interval(1)` and assert it still does and that inputs are not marked. It fails on the previous commit and passes now. yapf and flake8 clean under the repo's own configs.

tohtana (approved):
Thank you for the update! This looks good to me.

PR #8180, opened 2026-07-26
Add Triton grouped-GEMM for MoE experts on Ampere/Ada

hwchen2017:
**Motivation**

AutoEP expert computation uses  torch._grouped_mm , which only has a fused grouped-GEMM kernel on Hopper (sm90) and newer. On Ampere/Ada GPUs (sm80/sm86 — A100/A6000/A40/RTX 30xx) it silently falls back to a per-group Python loop ( _grouped_mm_fallback ), issuing one  at::mm  per expert plus a device→host sync on  offs . This is slow for MoE, which invokes grouped GEMM 3× per layer in the forward and again in the backward.

**What this PR does**

Adds  deepspeed/moe/group_gemm_triton.py : an autograd-aware Triton grouped GEMM that is a drop-in for  torch._grouped_mm  (2D×3D +  offs ), and wires it into  GroupedExperts  so it is auto-selected on sm < 9.0 (native path is kept on sm90+).

**Key points:**

• Single fused kernel per grouped GEMM, no per-group launches and no device→host sync (per-tile → group mapping is computed on device from  offs ).
• Support forward + backward for both operands, supporting fp16/bf16/fp32.
•  trans_b  option: expert weights are passed in their native  [E, N, K]  layout and transposed via strides inside the autograd Function, keeping  .transpose  off the tape. The weight gradient is produced directly in the native layout, avoiding a contiguous-materialization copy in backward. 
• Device-side group metadata via a small Triton kernel ( _group_meta_kernel ).
• Auto-selection is overridable via  DS_DISABLE_TRITON_GROUPED_MM=1 .
• Auto-tunning won't be called twice for same experts shape. 

**Performance**
Configuration: `dim=2048`, `hidden=768`, `avg_tokens/expert=64`,
`dtype=bf16`, distribution: **unbalanced**

| Experts | Tokens | Method | Forward (ms) | Fwd + Bwd (ms) | Backward (ms) |
|---:|---:|:---|---:|---:|---:|
| 16 | 760 | Triton | 0.107 | 0.589 | 0.482 |
| 16 | 760 | Torch | 0.256 | 1.093 | 0.837 |
| 16 | 760 | For-loop | 0.776 | 6.337 | 5.561 |
| 32 | 1,848 | Triton | 0.213 | 0.622 | 0.410 |
| 32 | 1,848 | Torch | 0.755 | 2.222 | 1.466 |
| 32 | 1,848 | For-loop | 1.603 | 22.732 | 21.129 |


deepspeed/moe/ep_experts.py
```diff
@@ -128,6 +131,77 @@ def _run_experts_grouped_mm(
...
+    (sm8x, where native would fall back to a slow per-group loop) as long as
+    Triton is available.
+
+    Set ``DS_DISABLE_TRITON_GROUPED_MM=1`` to disable the Triton path.
+    """
+    import os
+    if os.getenv("DS_DISABLE_TRITON_GROUPED_MM", "0") == "1":
+        return False
+
+    if not torch.cuda.is_available() or torch.version.hip is not None: # Not verified on AMD GPU
```
delock: need to supress check here (#ignore-cuda），this check seems transient, consider add an abstract accelerator interface to let accelerator backend decide whether goes to pytorch path or triton path independently.  Something like `if get_accelerator().prefer_triton_grouped_mm(): ...`

deepspeed/moe/ep_experts.py
```diff
@@ -128,6 +131,77 @@ def _run_experts_grouped_mm(
...
+
+    from deepspeed.moe.group_gemm_triton import is_available
+    if not is_available():
+        return False
+
+    if not hasattr(torch, "_grouped_mm"):
+        return True
+    
+    try:
+        major, _ = torch.cuda.get_device_capability()
```
delock: Same here, should leave the decision to abstract accelerator interface.

deepspeed/moe/ep_experts.py
```diff
@@ -128,6 +131,77 @@ def _run_experts_grouped_mm(
...
+
+    offsets = torch.cumsum(num_tokens_per_expert, dim=0, dtype=torch.int32)
+
+    # trans_b=True: pass expert weights in their native [E, hidden, dim] layout
+    # (no .transpose on the autograd tape). The kernel applies the transpose via
+    # strides, and backward writes the weight gradient directly in that layout,
+    # avoiding a contiguous-materialization copy of the transposed grad.
+
+    dtype = x.dtype
+    h = F.silu(group_gemm_triton(x, w1.to(dtype), offsets, trans_b=True))
```
delock, line 165: why silu cannot be fused with group_gemm_triton with epilogue?

deepspeed/moe/group_gemm_triton.py
```diff
@@ -0,0 +1,467 @@
...
+
+if _TRITON_AVAILABLE:
+
+    def _gmm_configs():
+        # Forward / grad_a kernel: output [M, N] tiled by BLOCK_M x BLOCK_N,
+        # reduction over K in BLOCK_K steps.
+        #   BLOCK_M (M/token tile): 32 / 64 / 128
+        #   BLOCK_N (N output tile): 128 / 256
+        #   BLOCK_K (K reduction):   32 / 64
+        return [
```
delock, line 69: Is it possible to abstract _gmm_configs into abstract accelerator interfaces to let different accelerator return their preferred configs?

delock:
Hi @hwchen2017 thanks for your PR.  I have left my comments in the code.  One comment is, this PR is intended for Ampere/Ada, but triton can be used for other accelertors as well.   Current code contains many CUDA specific checks, these code should be generized.  The suggested solution is inlined.  @tjruwase @sfc-gh-truwase for accelerator abstraction topic.

sfc-gh-truwase:
@hwchen2017 thanks for the PR. Please address @delock's feedback. Also, can you please include some perf numbers?

deepspeed/moe/ep_experts.py
```diff
@@ -128,6 +131,77 @@ def _run_experts_grouped_mm(
...
+
+    if not hasattr(torch, "_grouped_mm"):
+        return True
+    
+    try:
+        major, _ = torch.cuda.get_device_capability()
+    except Exception:
+        return False
+
+    return major < 9
```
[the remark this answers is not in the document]
hwchen2017: https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/runtime/engine.py#L1620
The sanity check here won't make it happen.

deepspeed/moe/ep_experts.py
hwchen2017: It can be. But we will only see less than 1% performance gain.  I will have a fused swiglu kernel later.

deepspeed/moe/ep_experts.py
hwchen2017: Fixed.

deepspeed/moe/ep_experts.py
hwchen2017: Fixed.

hwchen2017:
Hi @sfc-gh-truwase , I add some perf numbers and benchmark code.

deepspeed/moe/group_gemm_triton.py
hwchen2017: I'd prefer to keep  _gmm_configs  local to  group_gemm_triton.py  for now, for a few reasons:

• These configs are the autotune search space for this specific Triton kernel, not a final device-tuned choice —  triton.autotune  already picks the best one per problem shape at runtime. So they're an implementation detail of the kernel rather than a device capability.
•  The candidates for different kinds of device are likely same. 
•  Keeping it here also keeps all the Triton-specific logic (kernels + their autotune configs) in one place.

accelerator/cuda_accelerator.py
```diff
@@ -270,6 +270,25 @@ def is_triton_supported(self):
         else:
             return False
 
+    def prefer_triton_grouped_mm(self):
+        # torch._grouped_mm only has a fused grouped-GEMM kernel on Hopper (sm90)
+        # and newer; on sm8x it falls back to a slow per-group loop, so a Triton
+        # grouped-GEMM kernel is preferred there when Triton is available.
+        # Set DS_DISABLE_TRITON_GROUPED_MM=1 to force the native path.
+        import os
+        if os.getenv("DS_DISABLE_TRITON_GROUPED_MM", "0") == "1":
```
sfc-gh-truwase: Runtime env vars can be brittle in multi-node settings, and so generally avoided. If this is unavoidable then see https://www.deepspeed.ai/getting-started/#multi-node-environment-variables for proper usage.

tests/unit/v1/moe/test_group_gemm_triton.py
```diff
@@ -0,0 +1,396 @@
+# Copyright (c) DeepSpeed Team.
```
sfc-gh-truwase, line 1: Move UT to tests/unit/v1/moe

hwchen2017: Done.

delock (approved):

accelerator/cuda_accelerator.py
sfc-gh-truwase: @hwchen2017 any thoughts on this?

hwchen2017: I will replace it with a variable in config.

sfc-gh-truwase: Feel free to create an `accelerator` section in the config. 

Please also add some documentation. You can create an accelerator section here
https://deepspeed.readthedocs.io/en/latest/index.html

hwchen2017: Code has been updated to replace env var with a variable in config.

deepspeed/moe/group_gemm_triton.py
```diff
@@ -0,0 +1,473 @@
...
+                a_ptr + offs_m[:, None] * stride_am + offs_k[None, :] * stride_ak,
+                mask=mask_m[:, None] & mask_k[None, :],
+                other=0.0,
+            )
+            b_tile = tl.load(
+                b_base + offs_k[:, None] * stride_bk + offs_n[None, :] * stride_bn,
+                mask=mask_k[:, None] & mask_n[None, :],
+                other=0.0,
+            )
+            acc += tl.dot(a_tile, b_tile)
```
tohtana: Lines 32-34 say "Only ``float16``, ``bfloat16`` and ``float32`` inputs are supported, and both operands must share the same dtype. Matmuls accumulate in fp32 and the result
is cast back to the input dtype, matching ``torch._grouped_mm`` numerics." but the default precision of `tl.dot` is TF32.
I think this GEMM is not precision sensitive. So we could use TF32 but we also should clarify it. If we pass `ieee` to `input_precision`, FP32 inputs will be computed in FP32. It would also be good to add a config item to control it.

accelerator/abstract_accelerator.py
```diff
@@ -203,6 +203,14 @@ def communication_backend_name(self):
     def is_triton_supported(self):
         ...
 
+    # Whether the fused MoE expert path should prefer a Triton grouped-GEMM
+    # kernel over ``torch._grouped_mm`` on this accelerator. Backends return
+    # True when the Triton path is faster than the native op (e.g. CUDA sm8x,
+    # where ``torch._grouped_mm`` falls back to a slow per-group loop).
+    @abc.abstractmethod
```
tohtana: If it is an abstract method, all accelerator impl classes need to implement it including a new one. How about implementing a default method like

```python
def prefer_triton_grouped_mm(self):
    return False
```

and overrides it in the CUDA accelerator class.

accelerator/cuda_accelerator.py
```diff
@@ -270,6 +270,19 @@ def is_triton_supported(self):
...
 
+    def prefer_triton_grouped_mm(self):
+        # torch._grouped_mm only has a fused grouped-GEMM kernel on Hopper (sm90)
+        # and newer; on sm8x it falls back to a slow per-group loop, so a Triton
+        # grouped-GEMM kernel is preferred there when Triton is available.
+        from deepspeed.moe.group_gemm_triton import is_available as triton_grouped_mm_is_available
+        # not verified on AMD GPU
+        if torch.version.hip is not None or not triton_grouped_mm_is_available():
+            return False
+        if not hasattr(torch, "_grouped_mm"):
```
tohtana, line 281: As you [mentioned](https://github.com/deepspeedai/DeepSpeed/pull/8180#discussion_r3678700511), the engine-level BF16 sanity check covers the normal bf16 initialization path, but I think it is still good to filter out `major < 8` here for clarity.

tohtana (changes requested):
Hi @hwchen2017,

Thank you for submitting this amazing PR! This fills an important gap in AutoEP.
I left a few comments. I don't think any of them are blockers, but please consider addressing them.

accelerator/abstract_accelerator.py
hwchen2017: Done.

accelerator/cuda_accelerator.py
hwchen2017: Triton also support cuda capability 7.x. I update the code to filter out major < 7.

deepspeed/moe/group_gemm_triton.py
hwchen2017: • The FP32 path is an edge case since half precision is the default precision in training now. 
• TF32 is the de-facto standard for FP32 matmuls on Ampere+;  torch.matmul /cuBLAS and  torch._grouped_mm  also use TF32 by default. Forcing  ieee  would actually diverge from stock PyTorch numerics and give up most of the tensor-core throughput. Fully disabling tensor cores for FP32 is rare these days.
• Adding an  input_precision  config knob is extra surface area for a path almost nobody exercises; we can add it later if a real use case shows up.

tohtana: Okay, how about noting it with "Only float16, bfloat16 and float32 inputs are supported"?

deepspeed/moe/group_gemm_triton.py
```diff
@@ -0,0 +1,475 @@
...
+    out = grouped_mm(mat_a, mat_b, offs)
+
+    mat_a : ``[M, K]``      (2D, row-major contiguous)
+    mat_b : ``[E, K, N]``   (3D; arbitrary strides, so transposed views work)
+    offs  : ``[E]`` int32   cumulative row boundaries into ``mat_a`` along M,
+                            i.e. group ``g`` owns rows ``offs[g-1] : offs[g]``
+                            (``offs[-1] == M``). Empty groups are allowed.
+    out   : ``[M, N]``      same floating dtype as inputs.
+
+Only ``float16``, ``bfloat16`` and ``float32`` inputs are supported, and both
```
hwchen2017, line 32: @tohtana It's mentioned here.

tohtana (approved):
Thank you for the update, @hwchen2017! Looks good to me.

PR #8181, opened 2026-07-27
Non-reentrant activation checkpoint CPU offload

winglian:
upstreaming feature to support non-reentrant checkpointing from axolotl: https://github.com/axolotl-ai-cloud/axolotl/pull/3776

@sfc-gh-sbekman 

saved_tensors_hooks context manager that offloads checkpoint hidden-state inputs to a pinned CPU buffer pool on a side stream; works with use_reentrant=False.

### DIFF
--- a/deepspeed/runtime/activation_checkpointing/offload_activations.py
+++ b/deepspeed/runtime/activation_checkpointing/offload_activations.py
@@ -1,0 +1,333 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+"""
+Offload activation checkpoint inputs (hidden_states) to CPU for non-reentrant
+checkpointing (use_reentrant=False, the HF Transformers default)
+
+A saved_tensors_hooks context manager offloads only marked checkpoint inputs to
+a pinned CPU buffer pool on a side stream, overlapping copies with compute
+Independent of deepspeed.checkpointing.configure() state; eager-mode
+counterpart of the DeepCompile offload_activation pass
+transformers is imported lazily so importing this module works without it
+
+Adapted from axolotl checkpoint_activation_offload.py
+https://github.com/axolotl-ai-cloud/axolotl/pull/3776
+"""
+
+from __future__ import annotations
+
+import contextlib
+import importlib.util
+from dataclasses import dataclass
+from typing import Tuple
+
+import torch
+from torch.autograd.graph import saved_tensors_hooks
+
+from deepspeed.accelerator import get_accelerator
+
+
+@dataclass
+class CheckpointActivationOffloadStats:
+    saved_tensors_seen: int = 0
+    marked_tensors: int = 0
+    offloaded_tensors: int = 0
+    restored_tensors: int = 0
+    skipped_marked_tensors: int = 0
+    offloaded_bytes: int = 0
+    restored_bytes: int = 0
+
+
+@dataclass(frozen=True)
+class _OffloadedTensorRef:
+    tensor_id: int
+
+
+_BufferKey = Tuple[Tuple[int, ...], Tuple[int, ...], torch.dtype, torch.layout, bool]
+_MANAGER_STACK: list["CheckpointHiddenStatesOffload"] = []
+_ORIG_GRADIENT_CHECKPOINTING_LAYER_CALL = None
+
+
+def _current_manager() -> "CheckpointHiddenStatesOffload | None":
+    return _MANAGER_STACK[-1] if _MANAGER_STACK else None
+
+
+def patch_gradient_checkpointing_layer_marker() -> None:
+    global _ORIG_GRADIENT_CHECKPOINTING_LAYER_CALL
+
+    if _ORIG_GRADIENT_CHECKPOINTING_LAYER_CALL is not None:
+        return
+
+    try:
+        from transformers import GradientCheckpointingLayer
+    except ImportError as e:
+        raise RuntimeError("patch_gradient_checkpointing_layer_marker requires the `transformers` "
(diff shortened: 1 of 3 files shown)


stas00:
Thank you very much, @winglian, for porting it to Deepspeed

@tohtana, @tjruwase - for a better context - when Wing and I talked I said my implementation of activation offload was a very inefficient blocking prototype implementation. When Wing said he created an efficient 0-cost (full overlap) I have asking him if he could contribute his code to Deepspeed.

stas00:
@winglian, I think the last transformer block's activation should not be offloaded as it's almost immediately used on the first backward call and it could be exposed and blocking operation if the head's bwd compute is very fast - do you have that already and I missed it in the code?

Do you think there should be an arg controlling how many last activations to keep, probably with 1 being the default - is there a use-case where more than 1 would be beneficial?

deepspeed/runtime/activation_checkpointing/offload_activations.py
```diff
@@ -0,0 +1,333 @@
...
+    def _checkpoint_offload_call(self, *args, **kwargs):
+        manager = _current_manager()
+        if (manager is not None and self.training and torch.is_grad_enabled()
+                and getattr(self, "gradient_checkpointing", False)):
+            hidden_states = args[0] if args else kwargs.get("hidden_states")
+            if torch.is_tensor(hidden_states):
+                manager.mark(hidden_states)
+        return _ORIG_GRADIENT_CHECKPOINTING_LAYER_CALL(self, *args, **kwargs)
+
+    GradientCheckpointingLayer.__call__ = _checkpoint_offload_call
```
delock, line 81: `__call__` will stay changed after exit from the context manager.  What will happen when the model is forwarded again without this context manager?  I'm thinking about the scenario that same model is used for both training and rollout with HybridEngineRollout.

deepspeed/runtime/activation_checkpointing/offload_activations.py
```diff
@@ -0,0 +1,333 @@
...
+                           "package (>= 4.52, which introduced GradientCheckpointingLayer). "
+                           "Install it with `pip install transformers`.") from e
+
+    _ORIG_GRADIENT_CHECKPOINTING_LAYER_CALL = GradientCheckpointingLayer.__call__
+
+    def _checkpoint_offload_call(self, *args, **kwargs):
+        manager = _current_manager()
+        if (manager is not None and self.training and torch.is_grad_enabled()
+                and getattr(self, "gradient_checkpointing", False)):
+            hidden_states = args[0] if args else kwargs.get("hidden_states")
```
delock, line 76: Here `args[0]` depends on signature of `GradientCheckpointingLayer`.  Is it possible to add a comment about this assumption, and also add a test to check whether signature changed?
```
 def test_gradient_checkpointing_layer_signature_contract():
     """Pin the HF signature _checkpoint_offload_call depends on.
     If HF moves/renames hidden_states, fail loudly so the marker patch gets updated."""
     transformers = pytest.importorskip("transformers")
     if not hasattr(transformers, "GradientCheckpointingLayer"):
         pytest.skip("...")
     import inspect
     from transformers import GradientCheckpointingLayer

     sig = inspect.signature(GradientCheckpointingLayer.forward)
     params = [p for p in sig.parameters.values() if p.name != "self"]
     assert params, "GradientCheckpointingLayer.forward has no params"
     assert params[0].name == "hidden_states", (
         f"_checkpoint_offload_call assumes args[0] is hidden_states, but "
         f"GradientCheckpointingLayer.forward now starts with '{params[0].name}'. "
         f"Update the marker patch in patch_gradient_checkpointing_layer_marker."
     )
```

docs/_tutorials/ulysses-alst-sequence-parallelism.md
```diff
@@ -352,16 +352,27 @@ You can of course come up with a different way of computing the number of shards
...
+    get_checkpoint_hidden_states_offloading_ctx_manager,
+)
+
+ctx = get_checkpoint_hidden_states_offloading_ctx_manager()
+with ctx:
+    loss = model(**batch).loss
+    loss.backward()
+```
+
+Re-use one manager and wrap each training step (forward+backward) in it. Requires `transformers`: the marker identifying checkpoint inputs is installed on `GradientCheckpointingLayer`, and all other saved tensors pass through untouched. Tune with `use_pin_memory`, `use_streams`, `min_offload_size`, `max_fwd_stash_size` and `max_cpu_buffer_pool_size`.
```
delock, line 375: The `_size` here does not have same meaning, suggest to use `_bytes` and `_count` instead for less ambiguity.

deepspeed/runtime/activation_checkpointing/offload_activations.py
```diff
@@ -0,0 +1,333 @@
...
+        self.min_tensor_size_bytes = min_offload_size
+        self.max_fwd_stash_size = max_fwd_stash_size
+        self.max_cpu_buffer_pool_size = max_cpu_buffer_pool_size
+        self.stats = CheckpointActivationOffloadStats()
+        self._allowed: dict[int, int] = {}
+        self._next_id = 0
+        self._tracker: dict[int, tuple[torch.Tensor, torch.device, torch.Size, tuple[int, ...], _BufferKey]] = {}
+        self._fwd_stash: dict[int, tuple[torch.Tensor, torch.Event]] = {}
+        self._cpu_buffer_pool: dict[_BufferKey, list[torch.Tensor]] = {}
+        self._cpu_buffer_pool_size = 0
```
delock, line 112: same here, is it bytes or count?  Better clarify by name.

delock (changes requested):

delock:
Hi @winglian, can you followup with review comments?  Thanks!

delock:
Close as PR #8282 merged.  Thanks @winglian for contributing this feature!

PR #8184, opened 2026-07-27
Add managed_gradient_accumulation for ZeRO stage 0/1

### LINKED ISSUE #8183: [REQUEST] Unmanaged gradient accumulation: let callers own the optimizer-step boundary
## Motivation

DeepSpeed owns the gradient-accumulation boundary via an internal micro-step counter:
`backward()` bumps `micro_steps`, and `step()` only reduces gradients + runs the optimizer when

```python
(self.micro_steps + 1) % self.gradient_accumulation_steps() == 0
```

This enables symmetric `forward`/`backward`/`step` on every micro-batch, which is ideal for the
standard training loop.

It breaks down for callers that want to own the boundary themselves — notably HTTP/RPC-driven RL
backends (e.g. Tinker-protocol servers) where one logical optimizer step arrives as **N `backward()`
calls followed by exactly one `step()`**, with N client-controlled and not known at config time.
FSDP-based backends handle this natively (they call `optimizer.step()` when they choose); DeepSpeed's
only lever today is `set_gradient_accumulation_boundary(bool)`, which must be toggled around every
backward, is easy to get wrong, and couples the caller to engine internals.

## Proposal

Add a boolean ds_config flag `managed_gradient_accumulation` (default `True`, i.e. current behavior).
When `False` ("unmanaged" mode):

- micro-step tracking is disabled; there is no counter-driven boundary,
- `is_gradient_accumulation_boundary()` is `True` only while `step()` executes,
- `backward()` accumulates; `step()` finalizes accumulated grads and applies exactly one optimizer
  update — unconditionally, every call,
- the caller is responsible for calling `step()` at the true boundary.

This gives a first-class, config-level API for "backward and step are separate primitives, the client
owns the boundary" without per-backward toggling.

## Per-stage semantics

- **ZeRO 0/1 / DDP:** `backward()` accumulates gradients locally (no reduction); `step()` performs the
  all-reduce then the optimizer update.
- **ZeRO 2/3:** keep reducing/partitioning on every `backward()` (their natural behavior), but defer
  the `averaged_gradients` finalization + optimizer update to `step()` via a
  `finalize_gradient_accumulation_boundary()` hook.
- **ZeRO offload (CPU/NVMe optimizer):** supported; the boundary-time grad-norm computation and
  fp32/optimizer-buffer copies move into that finalize hook invoked from `step()`.

## Gradient scaling (open question)

DeepSpeed scales loss/grads by the configured `gradient_accumulation_steps`:

```python
gas_scaled_loss = loss / self.gradient_accumulation_steps() if scale_wrt_gas else loss
```

In unmanaged mode the number of `backward()` calls per step may differ from the configured GAS (and
may vary per step). Callers who accumulate a variable count should either set
`gradient_accumulation_steps` accordingly or use `backward(loss, scale_wrt_gas=False)` and average
client-side. Open question: should unmanaged mode default to `scale_wrt_gas=False`, or expose a
per-step "effective accumulation count"?

## Non-goals / open questions

- **Pipeline parallelism** (`PipelineModule.train_batch()` owns the whole schedule) is out of scope
  for v1; `managed_gradient_accumulation=False` would be rejected there.
- **Naming:** `managed_gradient_accumulation` vs `external_grad_accum_boundary` / `manual_optimizer_step`.
- **Scaling knob:** should there be a first-class control for variable accumulation counts (see above)?
- **Interactions** with ZenFlow, eigenvalue computation, and gradient clipping edge cases.

## Reference implementation

A working prototype exists covering ZeRO 0-3 including CPU optimizer offload, with unit tests asserting
numerical parity against managed mode (N backwards + 1 step == managed GAS=N). Happy to contribute as
stacked PRs (ZeRO 0/1 -> 2 -> 3 -> offload) if there's interest.

sfc-gh-truwase:
When managed_gradient_accumulation=false, disable micro-step tracking and treat each engine.step() as the accumulation boundary: reduce locally accumulated grads then apply the optimizer update. Stage 2/3 and pipeline remain unsupported in this change.

Part fix for #8183


stas00 (commented):
Looks good, but missing docs/examples of how one would do managed gas

sfc-gh-truwase:
@stas00 thanks. I will work on docs: https://deepspeed.readthedocs.io/en/latest/training.html#gradient-accumulation
It seems some other improvements are needed.

sfc-gh-truwase:
@stas00, see the docs in the staging: https://deepspeed.readthedocs.io/en/rtd-staging/training.html#gradient-accumulation

sfc-gh-truwase:
@codex review

stas00:
@sfc-gh-truwase, looking clear and it's easy to understand, thank you. 

I just have a question about:

> DeepSpeed scales the loss and gradients by the configured gradient_accumulation_steps. In unmanaged mode the number of backward() calls per step may differ from that value (and may vary per step); in that case call engine.backward(loss, scale_wrt_gas=False) and average the loss yourself.

If this is the case, shouldn't the example use `engine.backward(loss, scale_wrt_gas=False)` and show the averaging loss code? 

W/o it how would the example even work correctly?

docs/code-docs/source/training.rst
```diff
@@ -89,11 +89,14 @@ are responsible for calling ``step()`` when accumulation is complete:
...
-    for step_batches in batched_micro_batches:   # caller decides the boundary
+    for step_batches in batched_micro_batches:            # caller decides the boundary
+        num_micro_batches = len(step_batches)             # may differ from gradient_accumulation_steps
         for micro_batch in step_batches:
             loss = model_engine(micro_batch)
-            model_engine.backward(loss)          # accumulate only
-        model_engine.step()                      # reduce + optimizer update
+            # scale_wrt_gas=False disables DeepSpeed's 1/gradient_accumulation_steps scaling;
+            # average over the actual micro-batch count instead.
+            model_engine.backward(loss / num_micro_batches, scale_wrt_gas=False)   # accumulate only
```
stas00: I first missed the division - might it be more prominent / obvious / teachable if you were to add an explicit code, something along the lines of:
```
scaled_loss = loss / num_micro_batches
model_engine.backward(scaled_loss, ...
```

I'm not sure if `scaled_loss` is the best temp var name

The reader can then "optimize" it away in their own code.

deepspeed/runtime/engine.py
```diff
@@ -3120,11 +3138,18 @@ def is_gradient_accumulation_boundary(self):
...
 
+        In managed gradient accumulation (default), the result is driven by the internal
+        micro-step counter. With ``managed_gradient_accumulation=false``, it is ``True``
+        only while ``step()`` runs, since that call is the caller-owned boundary.
+
         Returns:
             bool: if the current step is a gradient accumulation boundary.
 
         """
         if self._is_gradient_accumulation_boundary is None:
```
tohtana, line 3210: For unmanaged mode to work as intended, `_is_gradient_accumulation_boundary` must remain `None` so that `is_gradient_accumulation_boundary()` uses the step-owned `_running_engine_step` marker. However, `set_gradient_accumulation_boundary()` can still set it to `True` or `False`, and either value breaks this contract. With `False`, the step-time `allreduce_gradients()` call performs no reduction and the optimizer update is skipped. With `True`, the backward epilogue reduces the gradients, and `step()` then enters the reduction path again.

Can we reject `set_gradient_accumulation_boundary()` when `managed_gradient_accumulation=False`?

tests/unit/v1/zero/test_zero_user_backward.py
```diff
@@ -1829,3 +1829,169 @@ def test_no_grad_input_accumulation(self, use_reentrant, dtype, param_persistenc
                 seed += 1
 
         model_engine.destroy()
+
+
+def build_managed_gas_config(zero_stage, gradient_accumulation_steps, managed_gradient_accumulation):
+    """fp32 config toggling managed_gradient_accumulation for exact managed-vs-unmanaged comparison."""
+    return {
+        "train_micro_batch_size_per_gpu": 2,
```
tohtana: `total_samples` in `random_dataloader()` is the number of individual samples, not the number of micro-batches. Here it is 8, and `train_micro_batch_size_per_gpu` is 2, so the loader yields only 4 micro-batches. However, the assertion expects 8 step results, corresponding to 8 micro-batches. 
The same accounting issue occurs below: 12 samples produce only 6 micro-batches, but the test indexes 12.

docs/code-docs/source/training.rst
```diff
@@ -57,8 +57,73 @@ Optimizer Step
...
+   owned by the caller and may differ from that value (and may vary per step), so the default scaling
+   would be incorrect. Pass ``scale_wrt_gas=False`` to disable DeepSpeed's scaling and average the
+   loss yourself over the actual micro-batch count, as shown above. (If you intentionally call
+   ``backward()`` exactly ``gradient_accumulation_steps`` times per step, the default scaling still
+   applies and no manual averaging is needed.)
+
+.. note::
+   Unmanaged mode is being added incrementally. Only ZeRO stage 0/1 (and DDP) is supported today;
+   ZeRO stage 2/3 and ZeRO optimizer offload are planned but **not yet available** -- enabling them
+   with ``managed_gradient_accumulation=false`` raises an ``AssertionError`` at initialization.
```
tohtana, line 122: The new check rejects ZeRO stages 2/3, but ZeRO-1 optimizer offload still initializes with unmanaged accumulation.

tohtana (changes requested):
Hi Tunji,
This is a very useful setting to make DeepSpeed more flexible for various use cases.
I left a few inline comments. We would need to fix the tests at least.

docs/code-docs/source/training.rst
sfc-gh-truwase: @tohtana good catch. I have follow up PRs for stages 2/3.

deepspeed/runtime/engine.py
sfc-gh-truwase: @tohtana, yes that makes sense. 

Seperately. do you think it user experience will be simplified by deprecating `set_gradient_accumulation_boundary()` in the long run?

sfc-gh-truwase:
Thanks @tohtana and @stas00 for the careful review. All the inline comments are addressed in the latest commits:

**@tohtana**
- **`set_gradient_accumulation_boundary()` breaks the unmanaged contract** — it is now rejected when `managed_gradient_accumulation=False` (the caller owns the boundary via `step()`). Commit `bee8ed56`.
- **ZeRO-1 optimizer offload still initialized** — unmanaged mode now rejects ZeRO optimizer/param offload at init, so behavior matches the docs. Commit `bee8ed56`.
- **Test micro-batch accounting** — `build_managed_gas_config` now uses micro-batch size 1 so `random_dataloader`'s `total_samples` equals the micro-batch count, fixing the off-by-2× step-count assertions. Commit `bee8ed56`.

**@stas00**
- The unmanaged example now uses `scale_wrt_gas=False` with an explicit `averaged_loss = loss / num_micro_batches` step, consistent with the note. Commits `dec2b145`, `ab6e8ac7`.

**Codex P1 (overlap_comm)**
- `overlap_comm=True` is now rejected in unmanaged mode, since `reduce_gradients()` skips its param walk under `overlap_comm` and the backward hooks skip reduction when the boundary is False. It defaults to False for the supported ZeRO stage 0/1. Commit `6b3d3564`.

**Docs**
- ZeRO stage 2/3 and optimizer offload are now clearly marked as planned / not-yet-available (rejected at init) rather than described as available. Commit `c6ba3415`.

New validation tests cover the `set_gradient_accumulation_boundary`, offload, and `overlap_comm` rejections. Re-requesting review — thanks again.

deepspeed/runtime/engine.py
tohtana: Yes, we can deprecate `set_gradient_accumulation_boundary` in the future. We would need to make this managed gradieent accumulation work with Z2/3 before we remove it.

deepspeed/runtime/engine.py
tohtana: `global_samples` can be incorrect in the unmanaged mode as we always use fixed `train_batch_size()`.

tests/unit/v1/zero/test_zero_user_backward.py
```diff
@@ -1829,3 +1829,216 @@ def test_no_grad_input_accumulation(self, use_reentrant, dtype, param_persistenc
...
+            engine.backward(loss)
+            engine.step()
+            applied.append(engine.was_step_applied())
+
+        # Only every GAS-th micro-step (the boundary) applies an update.
+        assert applied == [False, False, False, True, False, False, False, True]
+
+        engine.destroy()
+
+    def test_unmanaged_matches_managed(self, zero_stage):
```
tohtana, line 1915: Don't we test scenarios where GAS changes?

tohtana (changes requested):
Thank you @sfc-gh-truwase for addressing all my comments.
As I found a few more small issues, can you consider fixing them? I don't think they are critical, though.

stas00:
> * `overlap_comm=True` is now rejected in unmanaged mode, since `reduce_gradients()` skips its param walk under `overlap_comm` and the backward hooks skip reduction when the boundary is False. It defaults to False for the supported ZeRO stage 0/1. Commit `6b3d3564`.

Are you going to offer some workaround for this? Since otherwise the unmanaged mode may come at a big expense of exposed comms.

sfc-gh-truwase:
> Are you going to offer some workaround for this? Since otherwise the unmanaged mode may come at a big expense of exposed comms.

Good point. Since this is regarding reduction, I expect minimal impact for stage 0/1 where it is more performant to delay reduction until the last backward. For stage 2/3 which require reduction every backward pass, we will support overlap_comm.

tohtana (approved):
Thank you for the update, @sfc-gh-truwase! This looks good to me.

PR #8185, opened 2026-07-28
[AutoTP] Complete uneven sharding and universal checkpoint support

jinyouzhi:
Follow up #8146.

## Summary
This pull request introduces support for uneven sub-parameter sharding in DeepSpeed's universal checkpoint conversion, updating the universal checkpoint format to version 0.4. The changes ensure that partitioned parameters with sub-parameters of varying sizes are correctly handled during checkpoint conversion, merging, and restoration. Additionally, the PR adds validation to prevent conversion of unsupported checkpoint layouts and improves error handling and metadata validation.

Key updates by theme:

**Universal Checkpoint Format and Metadata:**
- Bumped the universal checkpoint version to 0.4 and introduced the `SUB_PARAM_SHARD_WIDTHS` field to record per-rank widths for each sub-parameter, enabling correct handling of uneven sub-parameter layouts. (`deepspeed/checkpoint/constants.py`, `deepspeed/checkpoint/ds_to_universal.py`) [[1]](diffhunk://#diff-7dfbb96f4f4bdab1e2be9ef97bda5a23e25e32d7e218991c268aea8065aec05eL61-R64) [[2]](diffhunk://#diff-7dfbb96f4f4bdab1e2be9ef97bda5a23e25e32d7e218991c268aea8065aec05eR93-R97) [[3]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R35) [[4]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R302-R303) [[5]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76L335-R441)
- Added `AUTOTP_UNSUPPORTED_PARAMETER_PATTERNS` to checkpoint metadata and implemented validation to prevent conversion if unsupported parameter patterns are present. (`deepspeed/checkpoint/constants.py`, `deepspeed/checkpoint/ds_to_universal.py`) [[1]](diffhunk://#diff-7dfbb96f4f4bdab1e2be9ef97bda5a23e25e32d7e218991c268aea8065aec05eL61-R64) [[2]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R47) [[3]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R950-R957) [[4]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R1134-R1139)

**Parameter Merging and Sharding Logic:**
- Enhanced the merging logic to correctly handle missing fragments for ranks with no data in uneven parameter sharding, ensuring proper alignment of slices and placeholder insertion. (`deepspeed/checkpoint/ds_to_universal.py`) [[1]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R239-R248) [[2]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R273-R285)
- Refactored the sub-parameter merging code to use the new shard widths metadata, supporting both legacy (even) and new (uneven) layouts. Added logic to reconstruct logical shapes with placeholder dimensions. (`deepspeed/checkpoint/ds_to_universal.py`) [[1]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76L302-R327) [[2]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76L335-R441) [[3]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R463-R482)

**Validation and Error Handling:**
- Added early validation for unsupported AutoTP conversions to fail fast before expensive extraction steps. (`deepspeed/checkpoint/ds_to_universal.py`) [[1]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R950-R957) [[2]](diffhunk://#diff-ef90f4743f09a5fcb81bcf3487a1d0f4b638e1191ca2c08c1d1d1f2b2c6b6f76R1134-R1139)
- Improved shape consistency checks for pipeline-parallel parameters to ensure all replicas agree on shape. (`deepspeed/checkpoint/ds_to_universal.py`)

**Restoration Logic:**
- Updated the restoration logic to use the new shard widths metadata, ensuring correct reconstruction of sub-parameters during model loading. (`deepspeed/checkpoint/universal_checkpoint.py`) [[1]](diffhunk://#diff-22d7b9e3b6eac1dc6e989cd8582946321c36f4fc80b527648eeb2d77a8fd3ee2L13-R13) [[2]](diffhunk://#diff-22d7b9e3b6eac1dc6e989cd8582946321c36f4fc80b527648eeb2d77a8fd3ee2R34-R73)

These changes collectively improve the robustness and flexibility of DeepSpeed's checkpoint conversion, especially for advanced tensor parallelism scenarios.

## Testing

* Added coverage for uneven vocabulary, GQA projections, checkpoint conversion/restore, and PP + TP tied parameters.
* loss curve https://github.com/deepspeedai/DeepSpeedExamples/pull/1008

## Limitations

tp_size > num_kv need to further be optimized


deepspeed/runtime/engine.py
```diff
@@ -726,19 +763,22 @@ def _apply_autotp_partitioning(self, model, tp_config):
...
             if layer_specs is not None:
-                logger.info(f"Using HuggingFace tp_plan with {len(layer_specs)} layer specifications")
+                gathered_output_patterns = [
+                    pattern for pattern, style in hf_tp_plan.items()
+                    if style.lower() in ("colwise_rep", "colwise_gather_output")
+                ]
+                print_dist(
+                    f"Using HuggingFace tp_plan with {len(layer_specs)} layer specifications; "
+                    f"gathered column output patterns={gathered_output_patterns}",
+                    ranks=[0],
```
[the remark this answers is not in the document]
jinyouzhi: Fix it separately in #8187, because it seemly was not introduced by these changes.

deepspeed/module_inject/layers.py
```diff
@@ -228,6 +231,48 @@ def backward(ctx: Any, grad_output: torch.Tensor) -> Tuple[None, torch.Tensor]:
...
+
+        tp_world_size = dist.get_world_size(group=group)
+        ctx.tp_index = dist.get_rank(group=group)
+        if tp_world_size == 1:
+            ctx.partition_sizes = (input.shape[-1], )
+            return input
+
+        local_size = torch.tensor([input.shape[-1]], dtype=torch.long, device=input.device)
+        gathered_sizes = [torch.empty_like(local_size) for _ in range(tp_world_size)]
+        dist.all_gather(gathered_sizes, local_size, group=group)
```
delock: should call  `get_shard_size_list` to compute size list, rather than use one additional `all_gather`.  Pass `total_size` and `name` from parameter list.

deepspeed/module_inject/layers.py
```diff
@@ -228,6 +231,48 @@ def backward(ctx: Any, grad_output: torch.Tensor) -> Tuple[None, torch.Tensor]:
...
+        max_partition_size = max(ctx.partition_sizes)
+        if input.shape[-1] == max_partition_size:
+            input_padded = input.contiguous()
+        else:
+            padded_shape = (*input.shape[:-1], max_partition_size)
+            input_padded = input.new_zeros(padded_shape)
+            input_padded[..., :input.shape[-1]].copy_(input)
+
+        gathered = [torch.empty_like(input_padded) for _ in range(tp_world_size)]
+        dist.all_gather(gathered, input_padded, group=group)
```
delock: should use `all_gather_into_tensor` for better collective efficiency.

deepspeed/module_inject/layers.py
jinyouzhi: Very good catch! Have pushed fix.

deepspeed/module_inject/layers.py
jinyouzhi: Very good catch! Have pushed fix.

deepspeed/checkpoint/ds_to_universal.py
delock: would `d` always divisible by `tp_degree` here?

delock: same question, does `sub_dim_size` always divisible by `tp_degree`.  If we always intentionally not handle this case (i.e. the case is too complicated), then we should guard it where the case originated.

deepspeed/checkpoint/universal_checkpoint.py
delock: here also have even assumption.

deepspeed/module_inject/tp_shard.py
```diff
@@ -76,4 +76,11 @@ def get_shard_size_list(total_size, mp_size, name=None):
     shard_sizes = []
     for i in range(mp_size):
         shard_sizes.append(get_shard_size(total_size, mp_size, name, i))
+    # Shards must tile the dimension exactly, otherwise the partitioned weights no longer
+    # reconstruct the original tensor. tp_grain_size quantization can violate this when the
+    # dimension is not a multiple of the grain size.
+    assert sum(shard_sizes) == total_size, (
```
delock, line 109: I have concern that grain is usually set to meet kernel performance constraints and its global.  And vocab size may not be divisible by grain, so the final result will be either grain = 1, which is not performance optimial, or vocab cannot divide by grain and get an assertion.   A proper solution should be ither handle the remainder gracefully, or use a different grain for vocab.  Preferrable the first one if solution is not too complicated.

deepspeed/runtime/engine.py
```diff
@@ -707,6 +707,43 @@ def _apply_autotp_partitioning(self, model, tp_config):
...
+                if any(part in ("lm_head", "embed_out") for part in pattern.split('.'))
+            }
+
+        lm_head_modules = [
+            name for name, _ in model.named_modules()
+            if name and any(part in ("lm_head", "embed_out") for part in name.split('.'))
+        ]
+        selected_route = "custom partition_config" if partition_config is not None else "HuggingFace tp_plan or AutoTP"
+        model_class = f"{type(model).__module__}.{type(model).__qualname__}"
+        print_dist(
```
delock: should remove diagnostics message before merge.

deepspeed/runtime/engine.py
```diff
@@ -727,19 +764,22 @@ def _apply_autotp_partitioning(self, model, tp_config):
...
             from deepspeed.module_inject.autotp_config import AutoTPConfig
 
             layer_specs = TPPlanConverter.convert(hf_tp_plan)
             if layer_specs is not None:
-                logger.info(f"Using HuggingFace tp_plan with {len(layer_specs)} layer specifications")
+                gathered_output_patterns = [
+                    pattern for pattern, style in hf_tp_plan.items()
+                    if style.lower() in ("colwise_rep", "colwise_gather_output")
+                ]
+                print_dist(
```
delock: same here, print_dist will generate output regardless of log level, suggest change to logger.info.

delock:
Hi  @jinyouzhi , thanks for your PR!   I have left my comments.  One thing is print_dist will add output to default run, should use logging instead, if the message is temporary for debugging before merge, they should be removed.

Have you done any end to end run for this feature?   This is a big change and I want to know how it works for real model training.

jinyouzhi:
> Hi @jinyouzhi , thanks for your PR! I have left my comments. One thing is print_dist will add output to default run, should use logging instead, if the message is temporary for debugging before merge, they should be removed.
> 
> Have you done any end to end run for this feature? This is a big change and I want to know how it works for real model training.

Thank you for the careful review. Your suggestions are very helpful and important.
I have not run a full end-to-end training test with this uneven-sharding implementation yet. For the previous PR, I tested the `lm_head` replacement using the DeepSpeedExamples fine-tuning scripts, but I agree that this change requires validation with a real model training workload. I will extend those tests to cover uneven sharding.

deepspeed/runtime/engine.py
jinyouzhi: done.

deepspeed/runtime/engine.py
jinyouzhi: done

delock:
Hi @jinyouzhi , can you also resolve conflicts with master branch?   Thanks!

jinyouzhi:
> Hi @jinyouzhi , can you also resolve conflicts with master branch? Thanks!

Sure, have pushed the merge with latest code.

deepspeed/module_inject/tp_shard.py
jinyouzhi: https://github.com/deepspeedai/DeepSpeed/pull/8185/changes#diff-214e32993d5440123080193836e988f024771aa4f6931c614ef9ad42a493f398R64

Insightful. I prefer to follow your suggestion and keep the existing logic: for `total_size >= grain_size`, keep the grain-aligned allocation first, even if some ranks end up with zero width. Align as many ranks as possible to `grain_size`, and let the remainder go to the last rank.

However, for the more fine-grained case—specifically when total_size falls in `[grain_size, grain_size * mp_size)`—I think it's worth exploring further. For example, given [64,64,64,64,64,64,0,0], should we instead distribute evenly across all ranks so none sit idle: [48,48,48,48,48,48,48,48]. Taking it a step further, we could search for a more balanced allocation via binary search, e.g. [64,64,64,64,32,32,32,32].
This could be left as a open question.

deepspeed/checkpoint/ds_to_universal.py
delock: Hi @jinyouzhi do you have answer for this question?  Thanks!

jinyouzhi:
@codex review

deepspeed/checkpoint/ds_to_universal.py
jinyouzhi: (L347 -> now ~L394) `sub_dim_size` is divisible by `tp_degree` now only in legacy path, which is from _partition_logical_tensor https://github.com/deepspeedai/DeepSpeed/pull/8185/changes#diff-0df5862062e8a6030b3b4c689914f06701a7023014f94d353cbd756e8a898776L1210, torch.chunk-based partitoning must perserve dim divisible. This even assertion break the GQA kv-head alignment-wise partitioning now is removed. The corresponding check move to L390.

deepspeed/checkpoint/universal_checkpoint.py
jinyouzhi: same to previous one.

deepspeed/checkpoint/ds_to_universal.py
jinyouzhi: This line (now refactored to ~L415) inherited its even-split assumption from [the original code](https://github.com/deepspeedai/DeepSpeed/pull/8185/changes#diff-0df5862062e8a6030b3b4c689914f06701a7023014f94d353cbd756e8a898776L1214). The new implementation accumulates `shard_widths` instead, so it no longer relies on that assumption. @delock Thanks!

deepspeed/module_inject/layers.py
```diff
@@ -829,6 +909,131 @@ def from_weights(cls, weight_shape=None, dtype=torch.half, weight=None, bias=Non
...
+    def gather_params(self, params_list):
+        if self._subparam_sizes is None:
+            # The inherited gather concatenates the shards in rank order, which is not how this
+            # layout was split, so it would hand back a silently wrong weight to consolidate.
+            raise RuntimeError(self._unsupported_uc_reason)
+
+        for idx, param in enumerate(params_list):
+            if param is None:
+                continue
+            logical_shape = self._orig_weight_shape if idx == 0 else self._orig_bias_shape
```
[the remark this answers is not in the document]
jinyouzhi: add fix

delock:
Hi @jinyouzhi, one coverage gap I think is worth closing before merge.

**The uneven round-trip tests check weights, not forward output.** `TestUnevenColumnUniversalCheckpoint` / `TestUnevenRowUniversalCheckpoint` assert the restored weights match, but weight equality ≠ forward correctness. A gather/concat **order** bug (or a row-parallel reduce bug) leaves every per-rank shard correct, so the round-trip still passes, while the gathered/reduced forward output is silently wrong — and a `loss = out.sum()` check can't catch it either, since sum is order-invariant.

Confirmed by mutation: reversing the per-rank cat order in `GatherFromTensorParallelRegion` (`module_inject/layers.py`) or in `merge_tp_slices` (`checkpoint/ds_to_universal.py`) keeps the existing weight round-trip tests green, but flips an element-wise forward comparison against an unsharded reference.

**Patch (two strict `tp=3` tests):** https://gist.github.com/delock/d5bb0f9276638978831babc592fee572
- Column `lm_head` (`vocab=10 / 3 → [4,3,3]`) and row-parallel GQA attention (`num_heads=4 / 3 → [2,1,1] → dim [6,3,3]`, the 847c4e4 scenario).
- Forward output is compared element-wise against an independent unsharded `nn.Linear` / `F.linear` reference; the column path also checks per-rank gradient shards.
- Then the full ucp save → `ds_to_universal` convert → load round-trip, with the "expected" weights gathered by a plain `dist.all_gather` (not `gather_params`) so the reference stays independent of the code under test.
- Mutation-tested all three paths: gather-order, convert-order, and row all-reduce bugs each flip the relevant assertion.
- `tp=3` (non-power-of-two) is deliberate — tp=2/4 lets common dims (4096, 32000, …) divide evenly and hide the remainder / grain-quantization path.

Two asks:
1. OK to add these as-is? They're additive; the genuinely new part is the forward-correctness check + `tp=3`, since the round-trip mechanism itself overlaps your existing tp=2/4 tests.
2. Would you also want the element-wise forward check bolted onto `TestUnevenColumn/RowUniversalCheckpoint` themselves, so the gap closes at tp=2/4 too, not just tp=3?

(CPU note: these need `LOCAL_SIZE=3` so `CPU_Accelerator.device_count()` reports 3 and the distributed harness doesn't skip them.)

jinyouzhi:
> Hi @jinyouzhi, one coverage gap I think is worth closing before merge.
> 
> **The uneven round-trip tests check weights, not forward output.** `TestUnevenColumnUniversalCheckpoint` / `TestUnevenRowUniversalCheckpoint` assert the restored weights match, but weight equality ≠ forward correctness. A gather/concat **order** bug (or a row-parallel reduce bug) leaves every per-rank shard correct, so the round-trip still passes, while the gathered/reduced forward output is silently wrong — and a `loss = out.sum()` check can't catch it either, since sum is order-invariant.
> 
> Confirmed by mutation: reversing the per-rank cat order in `GatherFromTensorParallelRegion` (`module_inject/layers.py`) or in `merge_tp_slices` (`checkpoint/ds_to_universal.py`) keeps the existing weight round-trip tests green, but flips an element-wise forward comparison against an unsharded reference.
> 
> **Patch (two strict `tp=3` tests):** https://gist.github.com/delock/d5bb0f9276638978831babc592fee572
> 
>     * Column `lm_head` (`vocab=10 / 3 → [4,3,3]`) and row-parallel GQA attention (`num_heads=4 / 3 → [2,1,1] → dim [6,3,3]`, the [847c4e4](https://github.com/deepspeedai/DeepSpeed/commit/847c4e4460aa3ef82e1a19f1a0c00150611cc58b) scenario).
> 
>     * Forward output is compared element-wise against an independent unsharded `nn.Linear` / `F.linear` reference; the column path also checks per-rank gradient shards.
> 
>     * Then the full ucp save → `ds_to_universal` convert → load round-trip, with the "expected" weights gathered by a plain `dist.all_gather` (not `gather_params`) so the reference stays independent of the code under test.
> 
>     * Mutation-tested all three paths: gather-order, convert-order, and row all-reduce bugs each flip the relevant assertion.
> 
>     * `tp=3` (non-power-of-two) is deliberate — tp=2/4 lets common dims (4096, 32000, …) divide evenly and hide the remainder / grain-quantization path.
> 
> 
> Two asks:
> 
>     1. OK to add these as-is? They're additive; the genuinely new part is the forward-correctness check + `tp=3`, since the round-trip mechanism itself overlaps your existing tp=2/4 tests.
> 
>     2. Would you also want the element-wise forward check bolted onto `TestUnevenColumn/RowUniversalCheckpoint` themselves, so the gap closes at tp=2/4 too, not just tp=3?
> 
> 
> (CPU note: these need `LOCAL_SIZE=3` so `CPU_Accelerator.device_count()` reports 3 and the distributed harness doesn't skip them.)


Thank you for the review—this is a very valuable suggestion.

I’ve cherry-picked your proposed test cases into this PR and followed your approach to strengthen the existing TP=2/4 coverage as well.

1. Added TP=3 round-trip cases, and expanded checkpoints across the suite for broader coverage.
2. Upgraded TP=2/4 tests to the same element-wise forward-correctness standard used in TP=3.

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -331,25 +352,85 @@ def get_matched_sub_params_pattern(name_):
...
-
-            partition_shape = [sum(d) if isinstance(d, tuple) else d for d in matched_sub_params_shape.shape]
-            partition_shape = [d // tp_degree if i == partition_dim else d for i, d in enumerate(partition_shape)]
-            slices = [s.view(partition_shape) for s in slices]
+            sub_dim_spec = matched_sub_params_shape.shape[partition_dim]
+            if isinstance(sub_dim_spec, tuple):
+                sub_dim_sizes = sub_dim_spec
+                normalized_subparam_shape = matched_sub_params_shape
+            else:
+                # Legacy metadata used a scalar at partition_dim as the number of equal
```
delock: I think this handling of legacy metadata looks fragile.   Two possible ways:
1. Utilize UCP version for backward compatibility, look at checkpoint version rather than look at checkpoint fields to make the decision.
2. Provide tools to convert from old version to new version.  This could be optional.

@sfc-gh-truwase for other opinion.

deepspeed/checkpoint/ds_to_universal.py
```diff
@@ -958,6 +1074,12 @@ def main(args):
             _parse_model_states_stage3([model_files_grid[tp_index][0]]) for tp_index in range(tp_degree)
         ]
 
+        # Fail before the extraction pass so an unconvertible checkpoint does not pay the
+        # cost of loading every optimizer shard and writing fragments into temp_dir.
+        uc_info = _load_universal_checkpoint_info_stage3(model_files)
```
delock: `uc_info` only needs to be computed when tp_degree>1.  Only L1107 use uc_info.

jinyouzhi: done.

deepspeed/checkpoint/universal_checkpoint.py
```diff
@@ -31,6 +31,46 @@ def _get_param_uc_restore_meta(param):
...
+
+    ``shard_widths`` gives the per-rank width of each sub-parameter, so a fused attention
+    weight is cut on key/value head boundaries. Checkpoints written before uneven
+    sub-parameters were supported carry no widths and can only have been split evenly.
+    """
+    assert sum(sub_dim_sizes) == full_view.shape[partition_dim], (
+        f"Sub-parameter sizes {list(sub_dim_sizes)} sum to {sum(sub_dim_sizes)}, but dimension {partition_dim} "
+        f"of the parameter is {full_view.shape[partition_dim]}. The metadata describes a different parameter.")
+
+    if shard_widths is None:
```
delock: During which condition that shard_widths will be `None`?   Since this is already the new version of AutoTP, `shard_width is None:` should be asserted right away.

deepspeed/checkpoint/universal_checkpoint.py
```diff
@@ -40,8 +80,17 @@ def _resolve_autotp_partition(current_param, ckpt_dict, full_hp_param, tp_rank,
...
     sub_param_shape = meta.get('sub_param_shape')
     sub_param_sizes = meta.get('sub_param_sizes')
+    sub_param_shard_widths = meta.get('sub_param_shard_widths')
+    partition_sizes = meta.get('partition_sizes')
     replicated = meta.get('replicated', False)
 
+    # The layer could not describe how it split this parameter, so conversion refuses it. The
+    # generic paths below would reassemble it as contiguous rank-ordered slices, which is not
+    # how a fused layout was cut, so restoring here would silently corrupt the weight.
+    unsupported_reason = meta.get('conversion', {}).get('unsupported_reason')
```
delock, line 91: In the long run, I think a non-semantic schema generated from tp sharding would simplify UCP conversion and restore.   Making UCP conversion and loading no longer semantic dependent.  @sfc-gh-truwase I'm considering write an RFC for discussion.

deepspeed/module_inject/auto_tp_model_utils.py
```diff
@@ -3,13 +3,84 @@
...
+            object.__setattr__(self, name, value)
+        else:
+            setattr(self._module, name, value)
+
+
+def get_head_shard_sizes(num_heads, mp_group=None, num_kv_heads=None):
+    tp_world_size = dist.get_world_size(group=mp_group)
+    if num_kv_heads is None:
+        num_kv_heads = get_num_kv_heads()
+    if num_kv_heads is not None and num_heads % num_kv_heads == 0:
```
delock: `get_shard_size_list` down below should be able to handle divisible case, should add `num_kv_heads` to param list of `get_shard_size_list`  (default value could be `None` so don't need to change each call site), thus you can avoid this `if` here doing the same thing as get_shard_size_list

deepspeed/module_inject/auto_tp_model_utils.py
```diff
@@ -52,20 +124,25 @@ def build_bloom_alibi_tensor(attention_mask: torch.Tensor, num_heads: int, dtype
...
-    if not self.training and dist.is_initialized():
-        num_heads_per_rank = get_shard_size(self.n_head, dist.get_world_size())
-        offset = sum(get_shard_size_list(self.n_head, dist.get_world_size())[0:dist.get_rank()])
+def get_alibi_mask(self, tensor, seq_length_with_past, mp_group=None, head_shard_sizes=None, total_num_heads=None):
+    original = self.get_alibi_mask_orig
+    if total_num_heads is not None and self.n_head != total_num_heads:
+        original_function = getattr(original, "__func__", None)
+        if original_function is None:
+            raise TypeError("Cannot invoke get_alibi_mask with the original total head count.")
+        mask = original_function(_HeadCountProxy(self, total_num_heads), tensor, seq_length_with_past)
```
delock, line 137: This piece of code looke patchy, can you explain why it has to be written this way?  @jinyouzhi Thanks!

delock:
@jinyouzhi I left some comments from manual walk through, I'll finish the rest walk through next week.   Thanks!

jinyouzhi:
> @jinyouzhi I left some comments from manual walk through, I'll finish the rest walk through next week.   Thanks!

Thank you very much for patient detailed review. I will proceed one by one. 
Very appreciate the insight about UC and multi-model within one process. Looking forward to further details.

deepspeed/checkpoint/ds_to_universal.py
sfc-gh-truwase: @delock, I agree with both points. 
1. Yes, we added versioning to help with BC issues.
2. Yes, such conversion tools would be helpful.

@jinyouzhi, since this PR is changing schema, please bump up the version.

deepspeed/checkpoint/universal_checkpoint.py
sfc-gh-truwase: @delock an RFC will be very useful. It is great that increased usage is exposing limitations of the initial UCP design.

deepspeed/checkpoint/ds_to_universal.py
jinyouzhi: Thanks for the guidance. I'll bump the UCP version accordingly. @sfc-gh-truwase

Quick note on the current BC design: `SUB_PARAM_SHARD_WIDTHS` is kept as a separate top-level key intentionally, so that old readers never see an unexpected field when constructing SubparamShape . This legacy path fallback here and the `shard_widths is None` branch in `universal_checkpoint.py` together cover the legacy path.

@delock Got it, just to confirm (since I haven't found the UCP version field actually gated on anywhere right now):
Bump the version AND replace the legacy fallback path with a hard assertion. Correct?

deepspeed/module_inject/auto_tp_model_utils.py
jinyouzhi: Nice catch, updated.

delock (approved):

deepspeed/checkpoint/ds_to_universal.py
jinyouzhi: Modified in https://github.com/deepspeedai/DeepSpeed/pull/8185/changes/2f397b1407ae25bbf324c888b234cd0010dafcde . Notes for record (Thx for delock's suggetsions):
1. Bumped up UCP version.
2. Use UCP version to gate legacy path (and preseve BC).

deepspeed/module_inject/auto_tp_model_utils.py
jinyouzhi: > This piece of code looke patchy, can you explain why it has to be written this way? @jinyouzhi Thanks!

Yes, this is quite tricky. It is necessary to maintain thread safety. The function needs `self.n_head` to represent the original total head count when calling HuggingFace functions. To preserve thread safety without modifying `self.n_head` temporarily, we use a Proxy pattern to intercept access to this attribute. Thanks for your review! @delock

jinyouzhi:
Fixed UT failure and resolved conflicts (about autotp compile)

deepspeed/checkpoint/universal_checkpoint.py
jinyouzhi: same to https://github.com/deepspeedai/DeepSpeed/pull/8185#discussion_r3765094768

PR #8186, opened 2026-07-28
Wiring up offload_opt_states

pengdurice:
#### Summary of Changes

- **Registered ops instead of closures** — four ops in the existing `dc` namespace. The graph
  carries a tensor anchor and an integer index; live tensors stay in module state.
- **ORDERED effect registration** so stock inductor keeps the ops *and* preserves their order.
  Reload-before-sync is correctness-critical. A test fails without it.
- **Config wiring, capacity-first schedule** `[(0,[z3]), (1,[for_init, z3, move_opt_states])]`:
  states are emptied to host before profiling, so the plan is made against the floor and a job
  that only fits with offloading never runs a step with everything resident. (The pass author's
  own ordering from their test harness; the budget formula is unchanged.)
- **Three stream races fixed**, all silent if wrong: copies wait for the optimizer's writes;
  `record_stream` protects reload buffers from early reuse; and the copy stream waits for the
  compute stream before writing a reload buffer — without it a mid-backward reload overwrites a
  live activation, which showed up as NaN losses. All are stream dependencies, no host waits.
- **Reload buffers allocate from the pool with room** (compute stream, just freed by backward).
  The wrong pool costs an allocator retry plus a device-wide sync per step.
- **Multi-graph gating** so graph breaks do not duplicate copies; `empty_cache` once per compile
  phase rather than per step (per-step measured +28%); mutual exclusion with `offload_parameters`.

#### Results

Qwen3-14B, 8×H200 (141 GB), ZeRO-3, micro-batch 4, fp32 states (22.2 GB/rank),
`expandable_segments:True`. Medians over each run's final phase, single campaign.

| configuration | seq 2048 | seq 2400 |
|---|---|---|
| plain eager ZeRO-3, no compile | OOM | OOM |
| ZeRO-Offload (CPU Adam) | 10.11 | OOM |
| DeepCompile, no offload (z3 only) | 1.89 | dies at the recompile |
| DeepCompile, no offload (default schedule) | 1.82 | dies at the recompile |
| this pass, blocking variant (eager-only) | 4.06 | 4.42 |
| **this pass (async), eager** | 2.43 | 4.32 |
| **this pass (async), inductor** | **2.01** | **3.70** |
| this pass at micro-batch 1 | 0.76 — planner offloads nothing | 0.85 |

#### Limitations

- Runs **in place of** prefetch and selective gather (worth 3–27%); combining them is future work.
- Designed for `gradient_accumulation_steps=1`: the graph runs per micro-batch, so accumulation
  repeats the whole cycle. Documented in the config docstring.
- Mutually exclusive with `offload_parameters`.

#### Tests

`tests/unit/v1/compile/test_offload_opt_states.py` — op and ORDERED-effect registration; a
mechanism test compiling side-effect ops through stock inductor and asserting program order;
budget planning and node placement; re-run and multi-graph gating; once-per-phase `empty_cache`;
and a 2-GPU end-to-end loss-parity test whose op counters prove the ops ran in the compiled graph
(reloads are skipped while profiling, so a nonzero reload count is the proof).

### DIFF
--- a/deepspeed/compile/init_z3.py
+++ b/deepspeed/compile/init_z3.py
@@ -37,12 +37,14 @@ def init_z3(engine, backend, compile_config, compile_kwargs, schedule=None):
 
     # Validate before touching the engine: everything below removes hooks and unpatches modules,
     # so raising later would leave a half-converted engine behind.
-    if compile_config.offload_opt_states and engine.zero_offload_optimizer() is not None:
+    # zero_use_cpu_optimizer(), not zero_offload_optimizer(): the latter returns the config
+    # object, which is present but inert for `offload_optimizer: {}` or `device: none`.
+    if compile_config.offload_opt_states and engine.zero_use_cpu_optimizer():
         raise ValueError("compile.offload_opt_states cannot be combined with ZeRO's "
-                         "zero_optimization.offload_optimizer: both manage the same optimizer state. "
-                         "ZeRO keeps it off the accelerator for the whole step and runs the optimizer "
-                         "there, while this pass keeps it resident when memory allows and moves it "
-                         "around the compiled graph. Enable one of them.")
+                         "zero_optimization.offload_optimizer set to cpu or nvme: both manage the "
+                         "same optimizer state. ZeRO keeps it off the accelerator for the whole step "
+                         "and runs the optimizer there, while this pass keeps it resident when memory "
+                         "allows and moves it around the compiled graph. Enable one of them.")
 
     optimizer = engine.optimizer
     use_opt = not isinstance(optimizer, DeepSpeedZeRoOffload)
--- a/deepspeed/compile/init_z3.py
+++ b/deepspeed/compile/init_z3.py
@@ -35,6 +35,15 @@ def _resolve_expected_grad_dtype(param):
 
 def init_z3(engine, backend, compile_config, compile_kwargs, schedule=None):
 
+    # Validate before touching the engine: everything below removes hooks and unpatches modules,
+    # so raising later would leave a half-converted engine behind.
+    if compile_config.offload_opt_states and engine.zero_offload_optimizer() is not None:
+        raise ValueError("compile.offload_opt_states cannot be combined with ZeRO's "
+                         "zero_optimization.offload_optimizer: both manage the same optimizer state. "
+                         "ZeRO keeps it off the accelerator for the whole step and runs the optimizer "
+                         "there, while this pass keeps it resident when memory allows and moves it "
+                         "around the compiled graph. Enable one of them.")
+
     optimizer = engine.optimizer
     use_opt = not isinstance(optimizer, DeepSpeedZeRoOffload)
 
--- a/deepspeed/compile/init_z3.py
+++ b/deepspeed/compile/init_z3.py
@@ -21,11 +21,6 @@ WARMUP = 5
 _MISSING = object()
 
 
-def _print_rank0(msg):
-    if dist.get_rank() == 0:
-        print(f"[DeepCompile] {msg}", flush=True)
-
-
 def _resolve_expected_grad_dtype(param):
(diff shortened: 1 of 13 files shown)


deepspeed/compile/init_z3.py
```diff
@@ -74,9 +74,21 @@ def init_z3(engine, backend, compile_config, compile_kwargs, schedule=None):
...
     if schedule is None:
+        if compile_config.offload_parameters and compile_config.offload_opt_states:
+            raise ValueError("offload_parameters and offload_opt_states cannot be enabled together; "
+                             "choose one offloading target per run. Note that offload_parameters may have "
+                             "been enabled implicitly: the engine turns it on when the ZeRO config "
+                             "offloads both optimizer and parameters to CPU.")
         schedule = []
         if (compile_config.offload_parameters):
             schedule.append((0, [zero3_compile.add_z3_gather_release, offload_parameters.offload_parameter_fwd]))
+        elif compile_config.offload_opt_states:
```
tohtana, line 96: This pass can be enabled with ZeRO optimizer's offload, but they won't work together. Can we reject the combination?

tests/unit/v1/compile/test_offload_opt_states.py
```diff
@@ -0,0 +1,320 @@
...
+import deepspeed.compile.passes.offload_adam_states as offload_pass
+from deepspeed.accelerator import get_accelerator
+from deepspeed.runtime.zero.offload_states import _make_offload_state_key
+from deepspeed.utils.torch import required_torch_version
+
+from unit.common import DistributedTest
+from unit.util import bf16_required_version_check, skip_on_arch
+from unit.v1.compile.util import compare_loss
+
+pytestmark = pytest.mark.skipif(not required_torch_version(min_version=2.1),
```
tohtana: DeepCompile already limits the version to 2.6+. I think we should make this consistent.

tohtana (changes requested):
Hi @pengdurice,
Thank you for submitting this PR! The adaptive offloading feature was described in the paper, but the code had remained unorganized for a long time. I’m very glad that you cleaned it up and enabled the feature.

I left a few comments about some minor issues. Can you please address them? I don’t see any issues with the core implementation of this PR.

deepspeed/compile/init_z3.py
pengdurice: sure, just added before this line: 
```
    optimizer = engine.optimizer
    use_opt = not isinstance(optimizer, DeepSpeedZeRoOffload)
```
Thank you!

tests/unit/v1/compile/test_offload_opt_states.py
pengdurice: sure, updated to 2.6. thank you!

tohtana:
Hi @pengdurice,
Thank you for the update!

I found the condition `engine.zero_offload_optimizer() is not None` also rejects an explicitly disabled ZeRO offload config (`offload_optimizer: {}` or `device: none`). Can you fix it? `engine.zero_use_cpu_optimizer()` might be useful. The expected behavior would be:

- `device: cpu`: Reject with `ValueError`
- `device: nvme`: Reject with `ValueError`
- `device: none` or `{}`: Accept

pengdurice:
> Hi @pengdurice, Thank you for the update!
> 
> I found the condition `engine.zero_offload_optimizer() is not None` also rejects an explicitly disabled ZeRO offload config (`offload_optimizer: {}` or `device: none`). Can you fix it? `engine.zero_use_cpu_optimizer()` might be useful. The expected behavior would be:
> 
> * `device: cpu`: Reject with `ValueError`
> * `device: nvme`: Reject with `ValueError`
> * `device: none` or `{}`: Accept

thank you for the comment, just fixed;-)

tohtana (approved):
Thank you for the update! It looks good to me now.

PR #8189, opened 2026-07-28
Enable ZeRO-3 linear wrapper for existing models

tohtana:
Passing an already-constructed model to `deepspeed.initialize()` with ZeRO-3 and `memory_efficient_linear=true` does not install the ZeRO-3 Linear wrapper. The wrapper is currently installed only when the model is constructed inside a `deepspeed.zero.Init()` context.

Without the wrapper, the standard Linear implementation can retain the gathered weight storage until backward completes, significantly increasing memory usage.

This PR activates the existing ZeRO-3 Linear wrapper for the `deepspeed.initialize(model=...)` path when `memory_efficient_linear=true`, without requiring a `deepspeed.zero.Init()` context.

### DIFF
--- a/deepspeed/runtime/zero/linear.py
+++ b/deepspeed/runtime/zero/linear.py
@@ -114,22 +114,34 @@ class LinearFunctionForZeroStage3(torch.autograd.Function):
             input, weight, bias = ctx.saved_tensors
 
             grad_input = grad_weight = grad_bias = None
+            weight_was_partitioned = (hasattr(weight, "ds_status")
+                                      and getattr(weight.ds_status, "name", None) == "NOT_AVAILABLE")
+            if weight_was_partitioned:
+                weight.all_gather()
 
-            dim = grad_output.dim()
-            if ctx.needs_input_grad[0]:
-                grad_input = grad_output.matmul(weight)
-            if ctx.needs_input_grad[1]:
-                if dim > 2:
-                    grad_weight = grad_output.reshape(-1, grad_output.shape[-1]).t().matmul(
-                        input.reshape(-1, input.shape[-1]))
-                else:
-                    grad_weight = grad_output.t().matmul(input)
-            if bias is not None and ctx.needs_input_grad[2]:
-                if dim > 2:
-                    grad_bias = grad_output.sum([i for i in range(dim - 1)])
-                else:
-                    grad_bias = grad_output.sum(0)
-            return grad_input, grad_weight, grad_bias
+            try:
+                dim = grad_output.dim()
+                if ctx.needs_input_grad[0]:
+                    grad_input = grad_output.matmul(weight)
+                if ctx.needs_input_grad[1]:
+                    if dim > 2:
+                        grad_weight = grad_output.reshape(-1, grad_output.shape[-1]).t().matmul(
+                            input.reshape(-1, input.shape[-1]))
+                    elif dim == 1:
+                        grad_weight = grad_output.unsqueeze(-1) * input.unsqueeze(0)
+                    else:
+                        grad_weight = grad_output.t().matmul(input)
+                if bias is not None and ctx.needs_input_grad[2]:
+                    if dim > 2:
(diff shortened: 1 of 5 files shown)


tests/unit/runtime/zero/test_zero_linear_direct_init.py
```diff
@@ -0,0 +1,131 @@
+# Copyright (c) DeepSpeed Team.
```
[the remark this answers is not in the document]
tohtana: False report

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -1018,6 +1021,9 @@ def __init__(self,
             init_distributed()
             assert dist.is_initialized(), "Parameters cannot be scattered without initializing deepspeed.comm"
 
+        if module is not None and self.enabled and self.mem_efficient_linear:
+            self._enable_mem_efficient_linear()
```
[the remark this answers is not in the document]
tohtana: This is a bug fix and doesn't change existing "expected" behaviors of `deepspeed.initialize(model=...)`.

deepspeed/runtime/zero/partition_parameters.py
```diff
@@ -395,6 +395,14 @@ def _set_dtype(self, ds_config, dtype):
             self.dtype = dtype or torch.float16 if get_accelerator().is_fp16_supported(
             ) else torch.bfloat16 if get_accelerator().is_bf16_supported else torch.float32
 
+    def _enable_mem_efficient_linear(self):
+        print_rank_0(
+            "nn.functional.linear has been overridden with a more memory efficient version. This will persist unless manually reset.",
+            force=False)
+        if not hasattr(InsertPostInitMethodToModuleSubClasses, "linear_bk"):
+            InsertPostInitMethodToModuleSubClasses.linear_bk = torch.nn.functional.linear
+            torch.nn.functional.linear = zero3_linear_wrap
```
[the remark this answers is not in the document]
tohtana: Addressed

deepspeed/runtime/zero/linear.py
```diff
@@ -114,22 +114,34 @@ def backward(ctx, grad_output):
...
-                        input.reshape(-1, input.shape[-1]))
-                else:
-                    grad_weight = grad_output.t().matmul(input)
-            if bias is not None and ctx.needs_input_grad[2]:
-                if dim > 2:
-                    grad_bias = grad_output.sum([i for i in range(dim - 1)])
-                else:
-                    grad_bias = grad_output.sum(0)
-            return grad_input, grad_weight, grad_bias
+            weight_was_partitioned = (hasattr(weight, "ds_status")
```
delock: It occurs to me that this better be a function of weight as well i.e. `weight.is_partitioned()`, maybe worth a seperate PR.

delock:
Hi @tohtana I have left my comments.  How many memory we may save from this PR?  Is a documentation change needed?

delock (approved):

deepspeed/runtime/zero/linear.py
sfc-gh-truwase: How about using i[s_zero_param()](https://github.com/deepspeedai/DeepSpeed/blob/d6fbd4768ae4d3e6559b323e1abe943263f56ba5/deepspeed/runtime/zero/utils.py#L142) for detection?

sfc-gh-truwase: @tohtana any thoughts on this?

tohtana: Yes, updated to `is_zero_param`.

tohtana:
Hi @delock, @sfc-gh-truwase,
Thank you for your review!

> How many memory we may save from this PR? Is a documentation change needed?

Without the fix, the memory for the materialized parameter of Linear layers will remain between forward and backward. If it is a Linear with 16k x 16x weight, it will be 512MB.
I don't think we need need to update the documentation because the resulting behavior is what we normally expect.

PR #8195, opened 2026-07-29
MoE dispatch: fix silent Tutel + TP corruption, speed up native and Tutel paths

jinyouzhi:
Follow-up to #8174. Three independent changes to `deepspeed/moe/sharded_moe.py`.

### 1. Speed up Tutel route extraction for k > 2

`topkgating` rebuilt a dense `[s, e]` one-hot mask for every route, so gating cost
grew linearly with `k`. On launch-bound configurations Tutel therefore lost to the
native path once `k > 2`. Since the top-k columns already name the selected experts,
the per-route index/location/gate values are now a `gather`. Routes are handed to
Tutel as contiguous int32 so its own casts become no-ops, and the capacity tensor is
resolved once instead of forcing a device-to-host sync per use.

Routing output is bit-identical.

### 2. Shard the Tutel dispatch buffer along capacity under tensor parallelism

Tutel's `encode()` returns a flat `[e * c, m]` buffer, whereas the dense path
produces `[e, c, m]`. The `drop_tokens(dispatched_input, dim=1)` that follows
therefore sliced the **model** dim instead of the **capacity** dim. The result still
reshaped cleanly into the expert input (`e * c * m / tp` divided by `e * m` happens to
yield `c / tp`), so there was no error — experts silently ran on a hidden-state slice
reinterpreted as capacity, producing wrong results.

Reshaping the buffer to `[e, c, m]` before sharding fixes it. This is a pre-existing
bug on `master`, not introduced by #8174, but #8174 widens its reach by lifting the
`k == 1` restriction on `use_tutel`.

### 3. Dispatch MoE tokens by index instead of a dense one-hot einsum

The native path materialised a dense `[s, e, c]` one-hot and ran
`einsum("sec,sm->ecm")`, costing `O(s * e * c * m)` to move only `s * k * m` elements.
Tokens are now routed through their capacity slots directly, so dispatch and combine
each become a single gather.

The sparse routing metadata already existed for Tutel, so the gate always returns it
and `MOELayer` selects a backend. The dense return path is kept because
`deepspeed/ops/transformer/inference/moe_inference.py` still consumes it.

Combine accumulates in fp32, matching the einsum it replaces — the einsum accumulated
in fp32 inside the tensor-core matmul, so a naive low-precision gather would have been
*less* accurate than the code it replaced.

## Results

Measured on 2 GPUs, bf16, forward+backward.

Tutel/native step-time ratio (change 1):

| k | before | after |
|---|--------|-------|
| 2 | 0.947  | 0.937 |
| 3 | 1.078  | 1.01  |
| 4 | 1.158  | 1.04  |

Native path (change 3), `s=8192 m=2048 e=16 k=3`:

| | step time | peak memory |
|---|-----------|-------------|
| before | 58.33 ms | 4353 MB |
| after  | 37.25 ms | 2918 MB |

The native path now matches Tutel (37.25 vs 37.49 ms).

## Testing

`tests/unit/moe/`: 50 passed with Tutel installed, 48 passed / 2 skipped without.

- `test_sparse_dispatch_matches_dense` (k=1/2/3) checks the index path against the
  dense einsum for both dispatch and combine.
- `TestMOETensorParallelTutel` checks Tutel against the dense path under tensor
  parallelism; it fails without change 2 and passes with it.

Cross-checked end-to-end against `master` (fp32 max diff 1.2e-07) and verified that
bf16/fp16 accuracy versus an fp32 golden reference does not regress.

> `TestMOETensorParallelTutel` requires Tutel and is skipped in CI, which does not
> install it. Change 2 was validated locally on 2 GPUs: max diff `7.27e-01` before the
> fix, `0.00e+00` after.


deepspeed/moe/sharded_moe.py
```diff
@@ -250,28 +301,23 @@ def top1gating(logits: Tensor,
...
+    if sparse_routes:
+        # The dispatch flags routes evicted by the capacity limit with a negative
+        # expert index, so replace the masked indices with -1.
         indices_mask = mask1.sum(dim=1) * num_experts - 1
         indices1_s = torch.min(indices1_s, indices_mask)
 
     # Compute locations in capacity buffer
-    if use_tutel:
+    if sparse_routes and TUTEL_INSTALLED:
         locations1 = tutel_moe.fast_cumsum_sub_one(mask1)
```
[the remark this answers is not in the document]
jinyouzhi: fixed

tests/unit/v1/moe/test_moe.py
```diff
@@ -14,7 +14,8 @@
 import deepspeed.moe.sharded_moe as sharded_moe
 from deepspeed import get_accelerator
 from deepspeed.moe.layer import MoE
-from deepspeed.moe.sharded_moe import top1gating, top2gating, topkgating
+from deepspeed.moe.sharded_moe import (top1gating, top2gating, topkgating, _route_slots, _sparse_encode,
```
sfc-gh-truwase, line 17: Move to tests/unit/v1/moe

tests/unit/v1/moe/test_moe_tp.py
```diff
@@ -9,6 +9,8 @@
 from unit.common import DistributedTest
 from deepspeed.utils.torch import required_torch_version
 from deepspeed.moe.layer import MoE
+from deepspeed import get_accelerator
```
sfc-gh-truwase, line 12: Move to tests/unit/v1/moe

tests/unit/v1/moe/test_moe.py
jinyouzhi: Moved to `tests/unit/v1/moe/` in  36d580377e482ef674ea4ddae44f34cda1e03627.

tests/unit/v1/moe/test_moe_tp.py
jinyouzhi: Moved to `tests/unit/v1/moe/` in  36d580377e482ef674ea4ddae44f34cda1e03627.

tohtana (approved):
Hi @jinyouzhi,
Thank you for your continuous contribution to the project! This looks good to me.

PR #8198, opened 2026-07-31
Fix AutoEP ZeRO-1/2 universal conversion

### LINKED ISSUE #8147: [BUG] AutoEP + ZeRO-1/2 universal checkpoint conversion silently leaves expert optimizer momentum stale/EP-unaware
**Describe the bug**

For AutoEP combined with ZeRO stage 1 or 2, universal checkpoint conversion (`ds_to_universal.py`) leaves expert parameters' Adam momentum (`exp_avg.pt`/`exp_avg_sq.pt`) silently stale/incorrect after conversion, while the corresponding weight file (`fp32.pt`) is correctly reshaped. This produces an internally inconsistent universal checkpoint for MoE experts: on resume, the weight is EP-sliced correctly but the optimizer state is not, so training resumes with mismatched/wrong Adam state for expert parameters.

AutoEP's introducing PR (#7938) explicitly lists "ZeRO stages 0, 1, and 2 support" plus "checkpoint save/load support" and "universal checkpoint conversion support" as in-scope ("ZeRO-3 extensions are intentionally left as follow-up work"), so this is inside the feature's claimed, already-merged scope rather than an acknowledged gap.

**Root cause (traced through source, see exact locations below)**

1. `save_checkpoint` in `deepspeed/runtime/engine.py` (~line 4931) writes each `expp_rank_{r}_mp_rank_{t}_optim_states.pt` with:
   ```python
   optimizer_state = {
       'optimizer': self.optimizer.state_dict() if self.optimizer and not self.zero_optimization() else None
   }
   ```
   `zero_optimization()` is `True` for any ZeRO stage &gt;= 1 (`deepspeed/runtime/config.py`), and the guard on this whole block is only `not self.zero_optimization_partition_weights()` (a ZeRO-3-only check). So under ZeRO 1/2 this file is written, but with `'optimizer': None`. The adjacent code comment only explains the ZeRO-3-vs-non-ZeRO distinction ("Non-ZeRO AutoEP keeps per-expert optimizer files. ZeRO-3 AutoEP restores experts from ZeRO optimizer shards") and doesn't address that ZeRO 1/2 falls into the same `None` branch.

2. In `deepspeed/checkpoint/ds_to_universal.py`, for `zero_stage &lt;= 2` (~line 770), the generic `_extract_zero_shard_files`/`_merge_tp_slice_files` (~lines 792, 795) run over every named parameter from `PARAM_SHAPES` with no exclusion mechanism. Unlike the ZeRO-3 path, which takes an `exclude_param_names` parameter (~lines 161-182, 390-397, 869) specifically to skip expert params, the stage&lt;=2 path has no equivalent, so it produces generic (EP-unaware) `fp32.pt`/`exp_avg.pt`/`exp_avg_sq.pt` for expert param names too.

3. Step "2.5" then runs `consolidate_autoep_expert_files` (~line 814), which unconditionally overwrites `fp32.pt` for expert params with correctly EP-tagged, full `[num_experts, ...]` tensors (setting `EP_IS_EXPERT_PARAM`/`EP_NUM_EXPERTS`).

4. `consolidate_autoep_optimizer_states` in `deepspeed/checkpoint/autoep_universal.py` (~line 816 call site, function body ~288-311) is supposed to do the equivalent for `exp_avg.pt`/`exp_avg_sq.pt`, but:
   ```python
   optim_sd = optim_states[0].get('optimizer')
   if optim_sd is None:
       return
   ```
   Per step 1, this is `None` under ZeRO 1/2, so the function returns having written nothing — silently, no warning/log.

**Result:** the universal checkpoint is left internally inconsistent for expert params: `fp32.pt` correctly reshaped/EP-tagged, `exp_avg.pt`/`exp_avg_sq.pt` silently left as the stale, EP-unaware generic-merge output. On load, `load_hp_checkpoint_state` sees `EP_IS_EXPERT_PARAM` on the weight key but not on the optimizer keys, so EP-slicing is applied only to the weight, not the momentum.

Separately, I noticed `EXPERT_PARAMETER_PATTERNS` (`deepspeed/checkpoint/constants.py`, ~line 106) is written into checkpoint metadata in three places (`ds_to_universal.py` ~lines 837, 895; `engine.py` ~line 4982) but doesn't appear to be read anywhere in the codebase — possibly an orphaned exclusion mechanism intended for exactly this gap.

**To Reproduce**

I was not able to build a full multi-rank reproduction — this requires a real distributed AutoEP + ZeRO-1/2 MoE training run plus a universal checkpoint conversion, which needs multi-GPU infrastructure I don't have access to. The trace above is from direct source reading of the exact code paths listed (file/line references given for each step), confirmed self-consistent across `engine.py`, `ds_to_universal.py`, and `autoep_universal.py`.

To reproduce directly, the general shape would be:
1. Train an AutoEP-enabled MoE model (e.g. Mixtral or Qwen3-MoE preset) with `zero_optimization.stage` set to 1 or 2.
2. Save a checkpoint, run universal checkpoint conversion (`ds_to_universal.py`) on it.
3. Inspect the resulting universal checkpoint's `exp_avg.pt`/`exp_avg_sq.pt` for an expert parameter and check whether it carries `EP_IS_EXPERT_PARAM`/`EP_NUM_EXPERTS` metadata and the expected `[num_experts, ...]` shape the way `fp32.pt` does for the same parameter — my trace says it won't.

**Expected behavior**

Under AutoEP + ZeRO stage 1/2, universal checkpoint conversion should either (a) correctly reshape/EP-tag the expert optimizer momentum the same way it already does for expert weights, or (b) if that's not yet implemented, fail loudly / warn rather than silently producing an internally inconsistent checkpoint.

**System info**

Traced against current `master` (commit referenced: AutoEP added in `7c3fbdc0`, "Add AutoEP" / #7938). I don't have a GPU/distributed setup to report `ds_report` output for an actual run against this bug.

**Additional context**

This was found via static code review, not observed in a live training run — flagging that explicitly since I wasn't able to empirically confirm the failure the way I would prefer to for a checkpoint-correctness issue. Happy to help narrow this down further or test a proposed fix if someone can point me at (or share access to) a small multi-GPU AutoEP + ZeRO-1/2 repro setup.

tohtana:
Fixes #8147

## Problem

AutoEP checkpoints saved with ZeRO stages 1 and 2 store expert optimizer state in ZeRO-sharded fragments, but Universal Checkpoint conversion currently merges those expert fragments through the generic path before AutoEP consolidation.
With source EP size greater than one, conversion can combine different EP ranks and fail while reshaping the result. The fallback path can also produce expert tensors without repartition metadata and use mixed-precision model weights where FP32 optimizer masters are required.

## Approach

- Identify fused AutoEP expert parameters from checkpoint metadata before the generic merge, then reconstruct their FP32  masters and Adam states from the ZeRO fragments in EP-rank order.
- Use the saved topology setting to handle both expert-before-data and data-before-expert rank layouts.
- Save the existing expert metadata with every expert state so Universal Checkpoint load can repartition to a different EP size.
- Restrict pipeline layer discovery to complete pipeline checkpoint filenames so per-expert files are not misclassified.


ErenAta16 (commented):
Thanks for turning this around so quickly, @tohtana. I went through the diff against the three code paths I traced in #8147, and the fix closes all of them. Worth noting it does so in a better way than what I suggested in the issue.

My suggestion was to either reshape the expert momentum correctly or fail loudly. What this PR does instead is stop depending on the per-expert optimizer file at all: `consolidate_autoep_zero12_expert_states` reconstructs the FP32 masters and both Adam moments directly from the ZeRO fragments in `temp_dir`. That sidesteps root cause 1 entirely (the `'optimizer': None` written by `save_checkpoint` under ZeRO 1/2, since `zero_optimization()` is true for any stage >= 1), rather than needing a change on the save side. The old `consolidate_autoep_optimizer_states` early return on `optim_sd is None` goes away with it.

Root cause 2 is addressed the way the ZeRO-3 path already did it: `_merge_tp_slice_files` now takes `exclude_param_names`, and the AutoEP metadata load moves ahead of the generic merge so `get_autoep_zero12_expert_param_info` can populate that exclusion set. That removes the "generic merge writes EP-unaware files, then step 2.5 overwrites only `fp32.pt`" split that left the moments stale.

The part I most wanted to check was the DP-to-EP rank mapping, since that is the load-bearing detail and getting it backwards would silently produce plausible-looking but wrong momentum. It matches the runtime grouping:

```
consolidate_autoep_zero12_expert_states:
  D+E: dp_ranks = range(ep_rank * edp_size, (ep_rank + 1) * edp_size)   # edp_size = dp_degree // ep_size
  E+D: dp_ranks = range(ep_rank, dp_degree, ep_size)

groups.py, expert data parallel group construction (mp_size <= 1 path):
  D+E: ranks = range(pp_stage_start + i * ep_stride, pp_stage_start + (i + 1) * ep_stride)
  E+D: ranks = range(pp_stage_start + i, pp_stage_start + pp_stride, expert_parallel_size_)
```

Same contiguous-block vs strided shapes, with `ep_rank` in the position `i` occupies, and `edp_size` equal to `ep_stride`. Reading `use_data_before_expert_parallelism` out of the saved `ds_config` rather than assuming a default is the right call, since picking the wrong topology here is exactly the kind of thing that would not throw and would just corrupt the state.

On the tests: parametrizing over both `E+D` and `D+E` covers the mapping above, and the `nonzero_moments == [True, True]` assertion is the one that would actually have caught the original bug. A shape-and-metadata-only check would pass on a stale or zeroed `exp_avg`, so asserting the moments are non-zero is what makes this a real regression test. The exact comparison (`rtol=0, atol=0`) against the source snapshot plus the `coverage == 1` check are good, and exercising a reshard (save at `ep_size=2`, load at `ep_size=4`) covers the case the metadata is there for in the first place.

The `NotImplementedError` for `tp_degree != 1` is the loud failure I was asking for in the issue, much better than emitting a checkpoint that looks fine.

One question, not a blocker: `consolidate_autoep_zero12_expert_states` takes `dp_degree` and `tp_degree` but not `pp_degree`, while the EDP groups in `groups.py` are built per pipeline stage with a `pp_stage_start` offset. If the ZeRO fragment `dp_rank` indexing is already per-stage then the relative ordering holds and there is nothing to do, but if AutoEP with `pp_degree > 1` is meant to be in scope it might be worth either confirming that or adding a guard next to the TP one.

Small thing I checked while reading: the new `if not unmatched_patterns_lists: return` in `_merge_tp_slice_files` skips the unused-pattern verification. That only triggers when the exclusion set covers everything in `slice_shapes` (the list has one entry per work item, so it is non-empty whenever anything was merged), so it does not weaken the check for normal runs.

I do not have a multi-GPU setup, so this is source-level review plus the cross-check against `groups.py` rather than a run of the new test. Flagging that for the same reason I flagged it in the issue. Happy to look again if anything changes.

deepspeed/checkpoint/autoep_universal.py
```diff
@@ -204,6 +204,142 @@ def resolve_expert_ckpt_path(checkpoint_dir, moe_layer_id, global_expert_id):
...
+    """Validate AutoEP metadata and map fused expert parameter names to layer metadata."""
+    if not isinstance(autoep_layers_metadata, list) or not autoep_layers_metadata:
+        raise RuntimeError("AutoEP metadata must be a non-empty list for ZeRO-1/2 universal conversion.")
+
+    param_info = {}
+    for layer_info in autoep_layers_metadata:
+        if not isinstance(layer_info, dict):
+            raise RuntimeError("AutoEP layer metadata must contain dictionaries.")
+
+        required_fields = ('expert_key_prefix', 'num_experts', 'num_local_experts', 'ep_size')
```
sfc-gh-truwase: Can you replace hard coded constants with symbolic constants?

ErenAta16 (commented):
Thanks for picking this up so fast. I read #8198 against the four root causes in the report; three are addressed and one is routed around rather than fixed, which I think is a reasonable call but worth stating explicitly. Two smaller things below.

**Mapping to the report**

| report | #8198 |
|---|---|
| 2 — `stage <= 2` merge path has no `exclude_param_names`, unlike the ZeRO-3 path | fixed directly: `_merge_tp_slice_files` takes `exclude_param_names` and filters `slice_shapes` before `_do_parallel_work` |
| 3 — step 2.5 overwrites generic `fp32.pt` after the fact | restructured away: AutoEP metadata is loaded *before* the merge and experts are excluded from it, so there's no generic artifact to overwrite |
| 4 — `consolidate_autoep_optimizer_states` returns silently when `optimizer` is `None` | routed around: `consolidate_autoep_zero12_expert_states` replaces it in the conversion path and reconstructs from the ZeRO fragments |
| 1 — `save_checkpoint` writes `'optimizer': None` into `expp_rank_*_optim_states.pt` under ZeRO 1/2 | **not addressed** — `deepspeed/runtime/engine.py` isn't in the diff |

Reconstructing from ZeRO fragments instead of fixing the writer is defensible — the fragments are the authoritative copy under ZeRO 1/2, and it keeps the fix inside `deepspeed/checkpoint/`. But the consequence is that those per-expert optimizer files keep being written with a `None` payload, and the comment above that block still explains only the ZeRO-3-vs-non-ZeRO distinction. Anything that reads them directly gets `None` with no signal. Either skipping the write under ZeRO 1/2 or extending that comment would stop this being rediscovered.

**The silent function survives.** `consolidate_autoep_optimizer_states` is still defined in `autoep_universal.py` on this branch, still opening with

```python
optim_sd = optim_states[0].get('optimizer')
if optim_sd is None:
    return
```

and `ds_to_universal.py` no longer calls it — I checked both names against the branch, only `consolidate_autoep_zero12_expert_states` appears there now. So the exact code path that made this fail quietly is now unreachable from conversion but still exported. If it's genuinely dead, removing it removes the trap; if it's kept for the non-ZeRO path, that early return could log instead of returning silently, which is what made the original bug invisible.

**Validation is much stricter now**, and I'd rather flag it than have it surprise someone: `get_autoep_zero12_expert_param_info` raises `RuntimeError` on missing fields, non-dict entries, and non-positive `num_experts`/`num_local_experts`/`ep_size`, and `main` now raises if `ds_config` isn't a dict or `use_data_before_expert_parallelism` isn't a bool. That's the right direction, but it turns previously-tolerated older checkpoints into hard conversion failures. If old AutoEP checkpoints can lack any of those fields, the error text pointing at what to do would help more than the bare field list.

**What I couldn't check.** I don't have an AutoEP MoE checkpoint or a working DeepSpeed environment here, so none of the above is from running the conversion — it's read against the branch. The behavioural claim in the report was traced the same way, so `test_autoep_universal_checkpoint.py` passing on your side is the part that actually settles it.

ErenAta16 (commented):
You asked me to try this and I said I had no multi-GPU box, which is still true. But the part of it I was least willing to take on trust, the DP-to-EP rank mapping, turns out to be checkable without one: both sides are integer arithmetic, so I transcribed them verbatim and compared them exhaustively.

I was careful not to make it circular. Comparing the two formulas against each other proves nothing, since either could be wrong in the same direction. Instead I derived each global rank's `ep_rank` from the **expert parallel** group construction (`groups.py:383-397`), which the conversion code never looks at, then checked two things against it:

1. every expert-**data**-parallel group built at `groups.py:366-379` is exactly one `ep_rank` equivalence class, and
2. `consolidate_autoep_zero12_expert_states`'s `dp_ranks` for a given `ep_rank` is exactly that class.

Over every `(world_size, ep_size)` pair with `world_size` in 2..64 and both topologies:

```
checked 492 (world_size, ep_size, topology, ep_rank) combinations
mismatches: 0
```

The `world_size=8, ep_size=4` case, so the shapes are visible:

```
E+D   ep_rank per global rank: [0, 1, 2, 3, 0, 1, 2, 3]
        ep_rank=0 -> dp_ranks [0, 4]     ep_rank=2 -> dp_ranks [2, 6]
        ep_rank=1 -> dp_ranks [1, 5]     ep_rank=3 -> dp_ranks [3, 7]

D+E   ep_rank per global rank: [0, 0, 1, 1, 2, 2, 3, 3]
        ep_rank=0 -> dp_ranks [0, 1]     ep_rank=2 -> dp_ranks [4, 5]
        ep_rank=1 -> dp_ranks [2, 3]     ep_rank=3 -> dp_ranks [6, 7]
```

Strided under E+D, contiguous under D+E, matching the runtime in both.

**Withdrawing my pipeline-parallel question.** I asked whether the missing `pp_degree` parameter mattered, since `groups.py` builds these groups per stage with a `pp_stage_start` offset while the conversion function has no notion of stages. I extended the same check to `pp_degree` in (1, 2, 4):

```
checked 784 EDP groups across pp_degree in (1, 2, 4)
mismatches after reducing global ranks to per-stage dp_rank: 0
```

The reason is that the offset cancels. Each stage occupies a contiguous block of `pp_stride` global ranks, and reducing a group's members to `global_rank - pp_stage_start` gives the same local pattern in every stage:

```
pp_degree=2, dp_degree=4, ep_size=2, E+D
  stage@0  global [0, 2] -> local [0, 2]  (ep_rank 0)   conversion picks [0, 2]
  stage@0  global [1, 3] -> local [1, 3]  (ep_rank 1)   conversion picks [1, 3]
  stage@4  global [4, 6] -> local [0, 2]  (ep_rank 0)   conversion picks [0, 2]
  stage@4  global [5, 7] -> local [1, 3]  (ep_rank 1)   conversion picks [1, 3]
```

So as long as the ZeRO fragment `dp_rank` is per-stage, which is the assumption the rest of the converter already makes, `pp_degree` genuinely is not needed here. My concern was unfounded and I would rather say so than leave it sitting on the PR.

One suggestion that follows from this. `test_autoep_universal_checkpoint.py` needs real GPUs, so the mapping is only covered where CI is least likely to run. The check above is about sixty lines of pure integer arithmetic with no torch import and no distributed init, and it fails loudly if either formula drifts from `groups.py`. Adding it as a plain unit test would put a guard on the one detail that produces silently wrong momentum rather than a crash, and it would run everywhere. Happy to open that as a follow-up PR against this branch or after it lands, whichever you prefer.

Everything else in my earlier review stands, including the two smaller points: `save_checkpoint` still writes `'optimizer': None` into the per-expert files under ZeRO 1/2, and `consolidate_autoep_optimizer_states` with its silent `return` is still exported although nothing in the conversion path calls it now.

tohtana:
Hi @ErenAta16,
Thanks for the exhaustive rank-mapping check and the follow-up!

While validating native PP=2, I found two issues:

1. the AutoEP MoE save path did not record enough topology information, so a normal PP=2/TP=1 checkpoint could be inferred as PP=1/TP=2
2. `ds_to_universal` read AutoEP metadata only from the first model-state file even though each pipeline stage records only its local AutoEP layers, so later-stage expert prefixes were omitted.

I updated the branch to resolve these and resolved the conflict.

ErenAta16 (commented):
Read through the four new commits. Both issues you found are real and the fixes land; one thing in the topology commit is worth a second look.

**Why the PP=2/TP=1 misread happened**, since it is worth having written down. The old inference in `get_model_3d_descriptor` keyed everything off the Megatron layer files:

```python
num_pp0_files = len(get_files_with_prefix(file_list, f'{LAYER_FILE_PREFIX}01'))
if num_pp0_files > 0:
    tp_degree = num_pp0_files
    pp_degree = len(model_file_list) // tp_degree
else:
    tp_degree = len(model_file_list)
    pp_degree = 1
```

An AutoEP checkpoint has no `layer_01_*` files at all, so `num_pp0_files` is 0 and it always takes the `else` branch. Two `mp_rank_*` files then mean `tp_degree=2, pp_degree=1` unconditionally, which is indistinguishable from the real PP=2/TP=1 case by file count alone. Recording the degrees explicitly is the only way out of that, so `CHECKPOINT_PARALLEL_DIMS` is the right call rather than a smarter heuristic.

The validation around it is careful in the places I would have worried about. Requiring every file to carry the key once any file does (`missing_parallel_dims`) closes the half-written-checkpoint case, the cross-file agreement check catches a mixed-topology directory, and the `len(model_file_list) != pp_degree * tp_degree` check catches a truncated copy. `isinstance(value, bool) or not isinstance(value, int)` is also the right guard and not the obvious one, since `bool` is an `int` subclass and a `pp_degree` of `True` would otherwise sail through as 1.

Falling back to the old heuristic when nothing records the dims keeps existing checkpoints converting, which matters more than the new path does.

**The `param_shapes` refactor is value-equivalent.** The old code built the same list one step later:

```python
slice_shapes = []
for mp_sd in model_states:
    slice_shapes += mp_sd[PARAM_SHAPES]
```

and `param_shapes.extend(model_state[PARAM_SHAPES])` produces the identical list in the identical order, with the following `dict((k, v) for d in slice_shapes ...)` untouched. `model_states` had no other consumer in `main`, so nothing is left dangling. The `del model_state` means peak residency drops from every stage's full state to one, which for a PP=8 job is the difference between eight deserialized checkpoints in memory and one.

**Which is why the new loop in `get_model_3d_descriptor` stands out.**

```python
for model_file in model_file_list:
    model_state = torch.load(model_file, map_location='cpu', weights_only=False)
    dimensions = model_state.get(CHECKPOINT_PARALLEL_DIMS)
```

This fully deserializes every model-state file, tensors and all, to read one small dict. Memory is bounded, since the rebind on each iteration frees the previous one, so this is a time cost rather than the blowup you just removed. But it is a complete extra pass over the checkpoint before conversion starts.

The sharper part is the legacy path: a checkpoint written before this PR carries no `CHECKPOINT_PARALLEL_DIMS` anywhere, so `parallel_dims` comes back empty, the function falls through to the old heuristic, and every one of those loads was paid for nothing. That is the common case for a while after this merges.

`mmap=True` would avoid materializing the tensor storages, or the dimensions could live somewhere cheaper to read than inside the model state. Not a correctness objection, and I have not measured it against a real checkpoint, so treat it as a flag rather than a request.

Happy to run the new pipeline test file against a synthetic PP=2/TP=1 layout if that would help; say the word and I will report numbers rather than opinions.

tohtana:
Hi @ErenAta16,

Thanks for pointing this out. I've actually experienced the inefficiency resulting from loading checkpoint files, so I tried to address the concern.

As mmap's performance seems hard to predict (it heavily depends on the environment), I went with a different approach.
The checkpoint discovery now loads each model-state file only once and retains only the topology, parameter shapes, AutoEP metadata, and other global metadata needed by the conversion. I refactored the conversion path to consume that collected metadata instead of loading the same files again. I also reduced the test code.

sfc-gh-truwase (approved):

ErenAta16:
Late reply, and the PR has merged since, but I owed you an answer on the loading change so here it is against `main` rather than against the branch.

Passing on mmap was the right call. Its behaviour depends on the page cache, the filesystem and whether the file is on network storage, so it would have moved the cost around unpredictably instead of removing it, and a checkpoint converter is exactly the workload where "it depends on the environment" turns into a support thread.

What landed does remove it. `_discover_model_state_metadata` reads each `mp_rank` file once in the constructor, keeps four lightweight fields, takes the global keys from file 0 only, and drops the payload:

```python
for file_index, model_file in enumerate(self.mp_rank_files):
    model_state = torch.load(model_file, map_location=torch.device('cpu'), weights_only=False)
    metadata_by_file.append((model_file, {
        CHECKPOINT_PARALLEL_DIMS: self._lightweight_parallel_dimensions(model_state),
        PARAM_SHAPES:            self._lightweight_param_shapes(model_state),
        AUTOEP_LAYERS_KEY:       self._lightweight_autoep_metadata(model_state),
        'ds_config':             self._lightweight_ds_config(model_state),
    }))
    ...
    del model_state
```

The `del` is the part that makes it hold rather than just look tidy: without it the payloads stay alive in `metadata_by_file`'s enclosing frame for the life of the conversion, which is the shape the original problem had.

I checked whether the class still re-reads those files anywhere on this path, since there are six other `torch.load` sites in `deepspeed_checkpoint.py`. They are `get_iteration`, `get_embedding_state`, `_get_checkpoint_value`, `get_2d_parallel_state`, `get_transformer_state` and `get_final_norm_state`, and `ds_to_universal.py` calls none of them; it constructs `DeepSpeedCheckpoint(args.input_folder)` and works from the collected metadata. So on the universal-conversion path each model-state file is now opened exactly once, which is what you said and what I wanted to confirm rather than assume.

Reducing the test code was the right trade too. The rank-mapping property is the part that needed exhaustive coverage and that is arithmetic, so it does not need a multi-GPU box; the rest was paying for itself less.

Thanks for taking the report through to a merge, and for the two topology bugs you found while validating it, which were not in my report and were the more dangerous half.

tohtana:
Hi @ErenAta16,
Thank you again for reporting the issue and for all the detailed feedback! It definitely helped us improve DeepSpeed and AutoEP.

ErenAta16:
Thank you, that means a lot coming from you.

This one was genuinely enjoyable to work on, mostly because of how you handled it. You turned the first fix around in under a day, and then while validating it you went and found two things my report never mentioned: the PP=2/TP=1 misread from an AutoEP checkpoint having no `layer_01_*` files, and `ds_to_universal` reading AutoEP metadata from only the first model-state file. Neither was reachable from where I was looking. That is the difference between the fix being correct for the case I happened to hit and being correct in general, and it came from you, not from the report.

If more AutoEP or universal-checkpoint work comes up, I would be glad to be pulled in. I should be honest about the limit: I do not have a multi-GPU box, so the runtime half I cannot verify. The checkpoint-format reasoning and anything that reduces to arithmetic I can take all the way, which is how the rank mapping got done.

Congratulations on the merge, and thanks for being easy to work with on it.

PR #8203, opened 2026-08-01
Unmanaged gradient accumulation: ZeRO stage 2 support

sfc-gh-truwase:
## Summary

Extends unmanaged gradient accumulation (`managed_gradient_accumulation=false`) to **ZeRO stage 2**. Stacked on top of the stage 0/1 foundation in #8184 (base branch `sfc-gh-truwase/gas_mgmt_zero01`); review that PR first.

Unlike stage 0/1 (where `backward()` accumulates locally and `step()` performs the reduction), ZeRO stage 2 must reduce/partition gradients on **every** `backward()` to preserve its memory characteristics. This is compatible with unmanaged mode because reduce-scatter is linear: accumulating the reduced partitions across N caller-controlled backwards is equivalent to reducing once at the boundary. Micro-step tracking stays disabled and the caller still owns the boundary; only the `averaged_gradients` finalization is deferred to `step()`.

- `ZeROOptimizer.finalize_gradient_accumulation_boundary()` (stage 1/2) builds `averaged_gradients` from the accumulated `all_grad_tensors` at `step()`.
- Validation relaxed to allow stage 2 (`not partition_weights`), still rejecting stage 3 and ZeRO offload (follow-up PRs).
- **`overlap_comm` is now supported for stage 2**: its async reduction is confined to the per-backward path and the epilogue synchronizes before finalizing, so it behaves exactly as in managed mode. It remains rejected for stage 0/1 (where reduction is deferred to `step()`).

## Test plan

Validated on a 2-GPU node (full `-k Unmanaged` suite, 22 passed):

- [x] `test_unmanaged_matches_managed[2]` — unmanaged stage-2 matches managed stage-2
- [x] `test_unmanaged_varying_backward_count[2]` — variable backward count per step, stage 2
- [x] `test_unmanaged_matches_managed_overlap_comm` — unmanaged stage-2 with `overlap_comm=True` matches managed reference
- [x] `test_unmanaged_rejects_stage3`, `test_unmanaged_rejects_zero_offload` — stage 3 / offload rejected
- [x] `test_unmanaged_rejects_overlap_comm[0,1]` — overlap_comm still rejected for stage 0/1
- [x] `pre-commit` (yapf/flake8/codespell) clean

Docs (`config-json.md`, `training.rst`) updated to describe stage-2 behavior and the overlap_comm support; previewable on `rtd-staging`.


deepspeed/runtime/engine.py
```diff
@@ -3365,9 +3365,11 @@ def step(self, lr_kwargs=None):
         # Unmanaged mode: step() is the accumulation boundary.
         self._running_engine_step = True
 
-        # Unmanaged mode: backward() only accumulates locally, so reduce grads here.
+        # Unmanaged boundary: stage 2 already reduced/partitioned per backward so only finalize; stage 0/1/DDP reduce here.
         if not self.managed_gradient_accumulation():
-            if self.enable_backward_allreduce and not self.inside_no_sync_ctxt:
+            if self.zero_optimization_partition_gradients():
+                self.optimizer.finalize_gradient_accumulation_boundary()
```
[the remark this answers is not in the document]
sfc-gh-truwase: Good catch. Rejected the combination up front: `coalesce_grad_reduction()` now asserts `managed_gradient_accumulation=True` at context entry (before mutating any state), since the caller already owns the accumulation boundary via `step()` in unmanaged mode. Added `TestUnmanagedGradientAccumulationCoalesceValidation` (stage 1/2). Fixed in d480edba.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -947,6 +947,20 @@ def independent_gradient_partition_epilogue(self):
 
         see_memory_usage("End ipg_epilogue")
 
+    def finalize_gradient_accumulation_boundary(self):
+        # Unmanaged mode: grads were reduced/accumulated into all_grad_tensors each backward; finalize averaged_gradients for step().
+        assert not self.cpu_offload, "unmanaged gradient accumulation does not support ZeRO offload"
```
stas00: would it be more user friendly to be more specific here, and say optim states offload?

param offload doens't need to be mentioned since Zero 3 isn't supported, right?

stas00 (approved):

deepspeed/runtime/zero/stage_1_and_2.py
sfc-gh-truwase: Good catch — updated both places:

1. `finalize_gradient_accumulation_boundary` now says "ZeRO optimizer state offload".
2. The init-time guard in `_do_sanity_check` likewise names optimizer state offload and drops the redundant `zero_offload_param()` check (param offload is stage-3-only and already rejected by the preceding stage guard).

Pushed in 94b67229; offload-rejection tests pass with the new match strings.

PR #8204, opened 2026-08-02
(1/2) Implementing Compiler Pass for AutoTP

therealnaveenkamal:
Working on #8104 

- Added support for AutoTP. Added two primitives: copy_to_tp and reduce_from_tp - the f and g nodes.
-  Identifies column/row-parallel matmuls by the injected layer type in nn_module_stack, reading back.
- Wrote a test to verify correctness of module injection and compiler pass.

cc @tohtana


deepspeed/compile/custom_ops/tp_collectives.py
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) Microsoft Corporation.
```
tohtana: For new files, please show DeepSpeed's copyright. You can find an example in other files.

tests/unit/compile/test_tp_compile.py
```diff
@@ -0,0 +1,183 @@
...
+
+import deepspeed
+import deepspeed.comm as dist
+from deepspeed.accelerator import get_accelerator
+from deepspeed.utils import groups
+from deepspeed.utils.torch import required_torch_version
+
+from unit.common import DistributedTest
+
+pytestmark = pytest.mark.skipif(not required_torch_version(min_version=2.9),
```
tohtana: Can you clarify what limits this feature to v2.9+? Also, can we have the same check in the main code?

tohtana (changes requested):
Thank you @therealnaveenkamal, this is amazing!
I don't see any critical issue in the code. The approach that reuses existing AutoTP's patterns is great. It gives consistent results with non-DeepCompile AutoTP.

One remaining work is validating the correctness in a more realistic setting. I think it would be good to compare loss values from  existing AutoTP and this one. I did similar work for [AutoEP](https://github.com/tohtana/DeepSpeedExamples/tree/tohtana/add_auto_ep/training/expert_parallel). The harness for the verification might be useful if you don't have such a script. We should check different configs like DP1/TP4 and DP2/TP2.

I left a few comments about details. Please consider addressing them. Also, please fix the commit to pass DCO check.

tohtana:
Can you also share your plan for the the next step as this PR has 1/2 in the title.
I think it would be good to start with the combination with ZeRO. Probably we could do further optimizations based on profiling.

deepspeed/compile/passes/tp_compile.py
```diff
@@ -0,0 +1,148 @@
...
+    return issubclass(layer_type, family)
+
+
+def defer_collectives_to_compiler(model) -> None:
+    """Suppress the module-level TP collectives on layers this pass will handle in the graph.
+
+    Any tensor-parallel layer the pass cannot rewrite is rejected rather than left on the
+    module-level path.
+    """
+    for name, module in model.named_modules():
```
tohtana, line 54: Can we set the flag `defer_collectives_to_compiler` only all modules passed the check?
If this raises an error in the loop, only some modules have `defer_collectives_to_compiler=True`. But the outer code might catch the error and fallback to eager. In that case, some communication collectives will be skipped.

tests/unit/module_inject/test_tp_plan_converter.py
```diff
@@ -109,11 +108,25 @@ def test_pattern_matches_param_name(self):
...
+        wraps the MoE router, whose forward returns a tuple, and breaks the model outright.
+        """
         hf_plan = {"layers.*.q_proj": "local_colwise", "layers.*.o_proj": "rowwise"}
-        result = TPPlanConverter.convert(hf_plan)
-        assert result is None
+        specs = TPPlanConverter.convert(hf_plan)
+
+        assert specs is not None
+        by_type = {spec.partition_type for spec in specs}
+        assert PartitionType.ROW in by_type, "the supported entry should still be applied"
```
tohtana: I think this expectation (and the code) is wrong. To make TP work, COLUMN and ROW should be paired. Partially skipping the conversion breaks it.

tohtana: The issue with Llama4 MoE router should be addressed in another PR. How about making it fail when any unsupported style is found.

deepspeed/compile/passes/tp_compile.py
```diff
@@ -0,0 +1,148 @@
...
+
+COLUMN_PARALLEL_OP = torch.ops.autotp.copy_to_tp_region.default
+ROW_PARALLEL_OP = torch.ops.autotp.reduce_from_tp_region.default
+GATHER_OUTPUT_OP = torch.ops.autotp.gather_from_tp_region.default
+
+# AutoTP replaces nn.Linear with these layers and shards their weights, so the injected layer type
+# already records the partitioning decision the pass needs. 
+
+COLUMN_PARALLEL_LAYERS = (LinearLayer, SubParamLinearLayer)
+ROW_PARALLEL_LAYERS = (LinearAllreduce, SubParamLinearAllreduce)
```
tohtana, line 24: SubParamLinearAllreduce.forward() always executes its module-level row all-reduce, while the compiler pass classifies that layer as row parallel and inserts another graph all-reduce.

tohtana (changes requested):
Hi @therealnaveenkamal,
Thank you for the update! I think the earlier issues have been addressed.
I also found some correctness issues in the current code. Can you check them?

deepspeed/compile/passes/tp_compile.py
therealnaveenkamal: thanks for this, @tohtana. I've made a check for `defer_collectives_to_compiler` in `layers.py`

deepspeed/compile/passes/tp_compile.py
therealnaveenkamal: got it. now we go through all the modules and if any module is incompatible, we raise an error

tests/unit/module_inject/test_tp_plan_converter.py
therealnaveenkamal: fixed it now.

therealnaveenkamal:
Hi @tohtana - thanks for pointing out the bugs. I've made the changes. Let me know if this is okay. Thanks.

tohtana:
Hi @therealnaveenkamal, thank you for the update!
Overall the change looks okay to me, but this test failed in my env. Can you check it?

```
tests/unit/compile/test_tp_compile.py::TestAutoTPCompileMoE::test_mixtral_matches_module_injection
```

tohtana:
Hi @therealnaveenkamal,
I investigated the test failure. I think it was a bug on HF side that existed from v5.8.0 to v5.10.0.

It started with https://github.com/huggingface/transformers/pull/45621 and fixed by https://github.com/huggingface/transformers/pull/45634. I found you already pushed the version guard, but can we be more specific about versions?
I don't see any other issues. After we confirm the new version guard works, let's merge this PR.

therealnaveenkamal:
Thanks @tohtana,  Guard is now exact: the range [5.8.0, 5.10.1) lives as a constant in init_tp.py, and the test skips it.

tohtana:
Hi @therealnaveenkamal,
Great, thank you for the update!

I just found the code should work even with the affected versions as long as we choose other than `batched_mm`. I opened a small PR to your branch: https://github.com/therealnaveenkamal/DeepSpeed/pull/5/changes
Can you review it? Feel free to modifty it if you want.

therealnaveenkamal:
@tohtana sorry about that. I've merged the PR.

tohtana (approved):
@therealnaveenkamal Thank you for your great work! This is definitely a significant step for DeepCompile.
I really appreciate your contribution to DeepSpeed.

PR #8205, opened 2026-08-02
Preserve tensor learning rates in OneCycle

tohtana:
Follow up #8202:

When an optimizer starts with a tensor learning rate, OneCycle initialization replaces it with a Python scalar while applying `cycle_min_lr`. This loses the caller's tensor identity, shape, and dtype before later scheduler updates can preserve them.

This PR initializes OneCycle learning rates through the existing tensor-aware update helper, matching the path used by subsequent scheduler steps.


delock (approved):

delock:
Hi @tohtana, I think a reflective test might help to prevent new lr scheduler not using `update_lr()`, I'll experiment and ask you to review on it.  Thanks!

PR #8213, opened 2026-08-04
Add an opt-in DeepEP transport for the AutoEP expert all-to-all

yh0903:
## What this adds

An optional DeepEP transport for the AutoEP expert all-to-all, selected by environment variable and defaulting to the existing NCCL path.

```bash
DEEPSPEED_AUTOEP_COMM_BACKEND=nccl     # default, unchanged behaviour
DEEPSPEED_AUTOEP_COMM_BACKEND=deepep
DEEPSPEED_AUTOEP_COMM_SMS=<n>          # default 12
```

## Why

The expert all-to-all is the largest single cost in an AutoEP step. Replaying routing captured from real SFT runs on 16 H100s across two nodes:

| Routing | Skew (max receive / mean) | NCCL | DeepEP | Speedup |
|---|---|---|---|---|
| math, last step | 3.01 | 99.72 ms | 48.02 ms | 2.08x |
| code, last step | 2.56 | 100.52 ms | 47.91 ms | 2.10x |
| code, worst step | 5.09 | 115.75 ms | 46.40 ms | 2.49x |

Two things are worth separating. Roughly half the gain is deduplication: DeepEP sends a token once per destination rank rather than once per selected expert, which is worth about 1.29x on its own and could in principle be done without changing transport. Its kernels account for the remaining 1.61x, and they are also where the skew immunity comes from — at the most imbalanced step recorded, NCCL degrades to 115.8 ms while DeepEP stays flat.

In whole training steps, with the same model, data and step count and only the backend changed:

```
nccl:   step=454.01 ms  payload all-to-all=142.92 ms (31.5% of the step)
deepep: step=370.74 ms  payload all-to-all= 29.68 ms ( 8.0% of the step)

end-to-end: 1.225x (18.3% of the step removed)
```

## Why DeepEP v2 only

This wraps `ElasticBuffer`, the v2 API. The legacy v1 `Buffer` is deliberately not supported:

- v1 carries data over NVSHMEM and IBGDA rather than NCCL, which needs either
the `NVreg_EnableStreamMemOPs` driver parameter or the GDRCopy `/dev/gdrdrv` device. Neither is present on the cluster this was developed against, and both require administrator action rather than configuration.
- v1's published internode bandwidth is markedly lower — 43 GB/s at EP16
against 153 GB/s intranode — and the node boundary is exactly the cost this change exists to reduce. v2's hierarchical NVLink plus RDMA path targets that case directly.
- v2 can reuse existing NCCL communicators, which matters in a training step
that already has ZeRO and data-parallel groups on the fabric.

The backend name is `deepep` rather than `deepep_v2`: it names the library, and nothing about it would have to change if v1 were ever added.

## Choosing the SM budget

Communication competes with the expert GEMM for SMs, so the default was chosen by sweeping whole training steps rather than the collective in isolation:

| Comm SMs | Step time |
|---|---|
| 8 | 340 ms |
| **12** | **311 ms** |
| 16 | 353 ms |
| 24 | 360 ms |
| 32 | 391 ms |

At 8 the collective itself degrades; above 12 the step grows because communication takes SMs the rest of the step was using. Tuning this from 24 to 12 moved the end-to-end result from 1.126x to 1.225x.

## Scope and safety

`git diff -w` removes zero lines from the shipped path: the existing NCCL code is unchanged and only moves into an `else` branch. `deep_ep` is imported only when the backend is selected, so installations without it are unaffected, and an unparsable backend value warns and falls back rather than breaking a path nobody opted into. DeepEP also requires NCCL 2.30.4 or newer for GIN, which not every cluster has, so a missing package explains what it needs instead of failing deep inside buffer construction.

DeepEP replaces more than the two collectives. It takes tokens before top-k expansion and replicates them itself, groups arrivals by expert for the grouped GEMM, and reduces the weighted sum in its combine, so the expansion and reduction around the collectives are replaced as well. Its backward pass has no separate entry points: the gradient of a combine is a dispatch and the gradient of a dispatch is a combine, both replayed against the handle the forward dispatch produced.

## Limitations

The DeepEP path is disabled with a warning when folded tensor parallelism is active (`tp_size > 1`). Folded TP partitions assignments across lanes and restores them by assignment metadata, which a transport whose combine returns token-major rows cannot satisfy. Such layers fall back to the NCCL path rather than silently producing wrong output.

## Testing

13 unit tests cover backend selection (default, empty, unknown, SM budget), the preflight checks, and gradient shape conformance. Correctness was verified on 16 H100s across two nodes for every routing scenario above, including a forward-and-backward round trip through all four payload collectives.

Follow-up work, kept out of this PR so each change stands on its own: communication/compute overlap (measured 1.11x, composes with this) and deduplication on the NCCL path (1.29x, needs no new dependency).


deepspeed/module_inject/auto_ep_comm.py
```diff
@@ -0,0 +1,328 @@
...
+        return combined, combined_weights
+
+    def combine(self, rows: torch.Tensor, handle, topk_weights=None) -> torch.Tensor:
+        """Reduce expert outputs back to their tokens.
+
+        Passing ``topk_weights`` makes DeepEP weight each expert's output as it
+        reduces, which is the same weighted sum the NCCL path performs
+        separately after its combine.
+        """
+        combined, _, _ = self.buffer.combine(rows, handle=handle, topk_weights=topk_weights, num_sms=self.num_sms)
```
tohtana: `ElasticBuffer.combine(..., topk_weights=...)` does not multiply `rows` by those weights ([`ElasticBuffer.combine`](https://github.com/deepseek-ai/DeepEP/blob/01dc3aaac82068020353dce2c302e38153c0bfaa/deep_ep/buffers/elastic.py#L1046-L1107)). DeepEP reduces the activations into `combined_x` and returns the transported weights separately as `combined_topk_weights`.
Can you apply each aligned routing weight to its expert-output row before the activation reduction, or otherwise implement an equivalent weighted reduction?

deepspeed/module_inject/auto_ep_layer.py
```diff
@@ -572,6 +591,63 @@ def set_deepspeed_parallelism(
...
+        selected expert, groups arrivals by expert for the grouped GEMM, and
+        reduces the weighted sum on the way back. It therefore replaces the
+        expansion and the reduction around the collectives, not just the
+        collectives, and returns [T, H] ready for the shared tail of forward.
+        """
+        # Built on first use because the buffer is sized from the token count,
+        # which is not known until a batch arrives, and rebuilt if a later
+        # batch is larger: variable sequence lengths or a small warm-up batch
+        # would otherwise exceed a capacity fixed by the first call.
+        if self._deepep_exchange is None or tokens.shape[0] > self._deepep_exchange.num_max_tokens_per_rank:
```
tohtana: This fixes the original fixed-first-batch capacity issue, but the decision is rank-local. DeepEP requires `num_max_tokens_per_rank` to have the same value on every rank.
There are two approaches: 1) allreduce (max) the required capacity, and 2) allocate the buffer for the worst case on all ranks.
This is a memory - performance tradeoff. Do you have any idea which is better? We could refer to other frameworks.

deepspeed/module_inject/auto_ep_layer.py
```diff
@@ -572,6 +591,63 @@ def set_deepspeed_parallelism(
...
+            if self._deepep_exchange is not None:
+                self._deepep_exchange.destroy()
```
tohtana: In the following pattern, we will use a destroyed exchange.

```python
forward(b1)
forward(b2) # extend buffer
backward(b2)
backward(b1) # Use destroyed exchange
```

This actually happens in this use case:

```python
out1 = model(b1)
out2 = model(b2)
loss = out1.sum() + out2.sum()
loss.backward()
```

I'm not sure what we could do for this case. If there is no good solution, we can just detect this pattern and throw an error.

deepspeed/module_inject/auto_ep_comm.py
```diff
@@ -0,0 +1,328 @@
...
+            num_max_tokens_per_rank=num_max_tokens_per_rank,
+            hidden=hidden_size,
+            num_topk=top_k,
+            use_fp8_dispatch=False,
+            num_allocated_qps=_qps_for_sms(num_sms),
+            # Required once the EP group spans nodes: it splits the ranks into
+            # an NVLink domain and an RDMA domain rather than assuming a single
+            # flat NVLink domain.
+            allow_hybrid_mode=True,
+            explicitly_destroy=True,
```
tohtana, line 163: With `explicitly_destroy=True`, the caller needs to explicitly destroy this buffer. The only current call to `DeepEPExchange.destroy()` is the resize path.

deepspeed/module_inject/auto_ep_layer.py
```diff
@@ -572,6 +591,63 @@ def set_deepspeed_parallelism(
...
+            self._deepep_exchange = DeepEPExchange(
+                ep_group=self.ep_group,
+                num_experts=self.num_experts,
+                top_k=self.top_k,
+                hidden_size=self.hidden_size,
+                num_max_tokens_per_rank=tokens.shape[0],
+            )
+
+        received, recv_weights, exchange = deepep_dispatch(self._deepep_exchange, tokens, ro.selected_experts,
```
tohtana, line 624: DeepEP's dispatch kernel doesn't work with FP16, but the code can reach here. Can we reject FP16 settings?

tohtana (changes requested):
Hi @yh0903,
Thank you for submitting this PR! Adding DeepEP to AutoEP is exciting.

I left some comments about correctness, cleaning, etc.

deepspeed/module_inject/auto_ep_comm.py
```diff
@@ -0,0 +1,328 @@
...
+
+from __future__ import annotations
+
+import os
+
+import torch
+
+from deepspeed.utils import logger
+
+_BACKEND_ENV = "DEEPSPEED_AUTOEP_COMM_BACKEND"
```
delock: Consider move AutoEP comm backend and num SMS into AutoEP configuration. i.e.

comm_backend: Literal["comm", "deepep"]      # comm means using comm collective
comm_num_sm: int = 12

deepspeed/module_inject/auto_ep_comm.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+def _qps_for_sms(num_sms: int) -> int:
+    """Queue pairs to reserve for a given SM count.
+
+    One per SM plus a small margin for the control path. This is deliberately
+    smaller than DeepEP's automatic choice, which assumes it is the only thing
+    on the fabric.
+    """
+    return num_sms + 4
```
delock: could num qps also be configurable?   Could be called `qp_margin` and make it default to 4

deepspeed/module_inject/auto_ep_comm.py
```diff
@@ -0,0 +1,328 @@
...
+        self.last_recv_weights = recv_weights
+        return recv_x, recv_weights, handle
+
+    def dispatch_with_handle(self, tokens: torch.Tensor, handle):
+        """Replay a dispatch against a cached handle, keeping both outputs.
+
+        Used as the backward of a combine, where the weights matter as much as
+        the rows: their gradient is what reaches the router gate.
+        """
+        recv_x, _, recv_weights, _, _ = self.buffer.dispatch(
```
delock: both `dispatch` and `combine` support passing `num_qps`, however here only `dispatch` has this optional parameter.  If `num_qps` is the same as where buffer is initialized, why not omit it in `dispatch` call?  We can avoid store the `num_qps` in the class and call `self.buffer.get_theoretical_num_qps(self.num_sms)`

deepspeed/module_inject/auto_ep_layer.py
```diff
@@ -550,6 +561,14 @@ def set_deepspeed_parallelism(
 
         if folding_group_handles is not None:
             self.folding_group_handles = folding_group_handles
+            if self.comm_backend == DEEPEP_BACKEND and folding_group_handles.spec.tp_size > 1:
+                # Folded TP partitions assignments across lanes and restores
+                # them by assignment metadata, which a transport whose combine
+                # returns token-major rows cannot satisfy. Saying so beats
+                # silently running something other than what was asked for.
+                logger.warning("AutoEP: the DeepEP backend does not support folded tensor parallelism; "
```
delock: I saw a behavior incosistency here.  When DeepEP cannot be imported, it will abort.  When DeepEP does not compatible with TP folding, it will fallback.   I'll favor both use abort because fallback will make user/agent think DeepEP is used even if it fallback to NCCL, if user are not careful enough to look at the log.

deepspeed/module_inject/auto_ep_layer.py
```diff
@@ -572,6 +591,81 @@ def set_deepspeed_parallelism(
...
+        handle = exchange.last_handle
+
+        # Two quantities have to line up for the grouped GEMM, and they come
+        # from different places: the rows are the valid prefix of a buffer
+        # sized for the worst case, and the per-expert counts arrive as a
+        # prefix sum. Differencing the same prefix sum that bounds the rows
+        # keeps the two consistent.
+        prefix = handle.psum_num_recv_tokens_per_expert
+        counts = torch.diff(prefix, prepend=prefix.new_zeros(1)).to(torch.int32)
+        received = received[:int(prefix[-1].item())]
```
delock: Non-blocking nit: unnecessary D2H sync from .item() in the dispatch hot path

received[:int(prefix[-1].item())] forces a GPU→CPU synchronization before the expert GEMM. The .item() drains the GPU pipeline (~20-50μs per layer) because the CPU must wait for the dispatch kernel to finish, transfer the scalar over PCIe, and only then can enqueue the GEMM.

This trim appears unnecessary: all three GroupedExperts paths handle the valid row range internally — torch._grouped_mm and the Triton kernel use offsets = cumsum(counts) to define group boundaries (rows beyond offsets[-1] are not part of any group), and the for-loop path does its own x[:sum(counts)] trim. So padding rows in the untrimmed received buffer should be safely ignored.

Consider removing the trim and passing the full received buffer directly. Estimated impact: ~0.5% of step time. Minor, but the fix is zero-cost.

delock:
Hi @yh0903 , very glad to see DeepEP being integrated into DeepSpeed.  I have left my comments.  Thanks!

deepspeed/module_inject/auto_ep_comm.py
delock: I think this case should be caught by a cosim test -- compare results between DeepEP result and all-to-all result, they should be the same or close.  Is there a cosim test in this PR?  @yh0903

deepspeed/module_inject/auto_ep_layer.py
```diff
@@ -611,21 +615,23 @@ def _deepep_route(self, tokens: torch.Tensor, ro: "RouterOutput") -> torch.Tenso
...
-
-        if self._deepep_exchange is None or num_max_tokens_per_rank > self._deepep_exchange.num_max_tokens_per_rank:
+        # DeepEP requires the same num_max_tokens_per_rank on every rank, and
+        # construction is collective, so the size is agreed rather than taken
+        # from the local count. The agreement is cached against the shape that
+        # produced it: ranks in an expert-parallel group are handed the same
+        # micro-batch shape, so they leave and re-enter the agreement together,
+        # and a collective on every layer of every step would otherwise put a
+        # device-to-host synchronisation in the middle of the forward pass.
+        if tokens.shape[0] != self._deepep_tokens_per_rank:
```
delock: I think there still have chance that some ranks meet the condition and some ranks didn't meet this condition.  Thus some ranks may execute allreduce and some are not, leading to a stall.

How much is the allreduce cost?  We may need to live with this cost if we want to use memory smartly.

delock (approved):

delock:
@yh0903 thank you for addressing my comments.

yh0903:
> @yh0903 thank you for addressing my comments.

Thanks [@delock](https://github.com/delock) for the follow-up review and approval. Sorry I didn't response to your message timely, and actually I was typing the replies! Really appreciate your comments again, which led to multiple improvements! I should have replied to your thread to make it easier to track the latest changes, but thanks for looking into them!

deepspeed/module_inject/auto_ep_comm.py
yh0903: Thanks for catching this! You're right, I misunderstood what `topk_weights` does in `ElasticBuffer.combine`. It only transports and reduces the weights, but doesn't apply them to `rows`. I changed `_deepep_route` to apply the routing weights explicitly before or after the expert MLP based on `score_apply`, and `combine()` no longer takes them. I also added a comm-vs-DeepEP parity test for outputs and gradients, plus a check that the router gate gradient is nonzero.

deepspeed/module_inject/auto_ep_layer.py
yh0903: Good point. I ended up doing one `all_reduce(MAX)` when each layer builds its buffer, so every rank uses the same capacity. With the default `comm_max_tokens_per_rank=0`, it uses the group-wide first-batch maximum rounded to 512 tokens. For jobs where later batches can be larger, the new `comm_max_tokens_per_rank` config can provide the worst case explicitly. The buffer isn't resized after initialization; an oversized later batch now gives a clear error instead of making a rank-local resize decision.

deepspeed/module_inject/auto_ep_layer.py
yh0903: Yes, this pattern makes runtime resizing unsafe. I removed the resize path entirely. Each layer now builds one exchange and keeps it alive until engine teardown, so both forwards can still use the exchange saved with their handles during backward. If a later batch is too large, it fails before dispatch and points to `comm_max_tokens_per_rank` instead of destroying and replacing the exchange.

deepspeed/module_inject/auto_ep_comm.py
yh0903: Yes, I added a registry for the live exchanges, and `DeepSpeedEngine.destroy()` now destroys all of them in construction order. The engine destructor also calls the same cleanup as a fallback. Since the resize path is gone, buffers are no longer destroyed in the middle of training.

deepspeed/module_inject/auto_ep_layer.py
yh0903: Yes, added a dtype check at the start of `_deepep_route`. FP16 now fails before buffer construction or dispatch, with an error pointing users to bfloat16 or `comm_backend="comm"`.

deepspeed/module_inject/auto_ep_comm.py
yh0903: Done. Backend selection is no longer environment-variable based. `comm_backend` and `comm_num_sm` now live in the `expert_parallel` config alongside the rest of AutoEP, with defaults `"comm"` and `12`. `validate_autoep_config()` rejects unknown backends and non-positive SM budgets.

deepspeed/module_inject/auto_ep_comm.py
yh0903: Yes, the new `expert_parallel.comm_qp_margin` field defaults to 4 and is validated as a non-negative integer. The buffer reserves `comm_num_sm + comm_qp_margin` queue pairs.

deepspeed/module_inject/auto_ep_comm.py
yh0903: Fixed. The QP count is derived once from `num_sms + qp_margin` and passed as `num_allocated_qps` when the `ElasticBuffer` is constructed. It is no longer stored as a separate `DeepEPExchange` field, and neither `dispatch()` nor `combine()` passes a per-call `num_qps`; those calls only pass `num_sms`.

deepspeed/module_inject/auto_ep_comm.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+# Every buffer built in this process, in construction order. DeepEP buffers are
+# constructed with explicitly_destroy, so nothing reclaims them on its own, and
+# destroying them is collective: every rank has to do it in the same order.
+_LIVE_EXCHANGES: list["DeepEPExchange"] = []
+
+# DeepEP's dispatch kernel handles bfloat16 and fp8, not fp16. A half-precision
+# run would otherwise reach an assertion inside the kernel.
+SUPPORTED_DTYPES = (torch.bfloat16, torch.float32)
```
tohtana: Does DeepEP accept float32? It will throw an error at DeepEP's BF16-only kernel.

deepspeed/module_inject/auto_ep_layer.py
```diff
@@ -572,6 +590,85 @@ def set_deepspeed_parallelism(
...
+                ep_group=self.ep_group,
+                num_experts=self.num_experts,
+                top_k=self.top_k,
+                hidden_size=self.hidden_size,
+                num_max_tokens_per_rank=sized_for,
+                num_sms=self.comm_num_sm,
+                qp_margin=self.comm_qp_margin,
+            )
+
+        if tokens.shape[0] > self._deepep_exchange.num_max_tokens_per_rank:
```
tohtana, line 616: If tokens.shape[0] can differ across ranks, only some ranks will raise the error?

deepspeed/module_inject/auto_ep_comm.py
```diff
@@ -0,0 +1,331 @@
...
+    fabric: in a training step ZeRO and the data-parallel groups have already
+    taken their share, and asking for DeepEP's default exhausts them.
+    """
+    return num_sms + qp_margin
+
+
+# Every buffer built in this process, in construction order. DeepEP buffers are
+# constructed with explicitly_destroy, so nothing reclaims them on its own, and
+# destroying them is collective: every rank has to do it in the same order.
+_LIVE_EXCHANGES: list["DeepEPExchange"] = []
```
tohtana: `_LIVE_EXCHANGES` is process-global. As multiple DeepSpeed engines can be created, can we tie this to an engine?

deepspeed/runtime/engine.py
```diff
@@ -824,6 +824,11 @@ def __del__(self):
             logger.debug("DeepSpeedEngine.__del__ cleanup skipped: %s", exc, exc_info=True)
 
     def destroy(self):
+        # DeepEP buffers ask the library not to reclaim them, so they outlive
+        # the engine unless something releases them here.
+        from deepspeed.module_inject.auto_ep_comm import destroy_all_exchanges
+        destroy_all_exchanges()
```
tohtana: `DeepSpeedEngine.__del__()` calls `destroy()`, and DeepEP `ElasticBuffer.destroy()` performs a cross-rank barrier. Python does not guarantee when finalizers run on each rank, so relying on `__del__` for this cleanup can leave some ranks stuck in `destroy_all_exchanges()`. Could we document that the DeepEP path requires an explicit, coordinated `engine.destroy()`, and have the `__del__` fallback warn if DeepEP exchanges are still live instead of entering their collective teardown?

tohtana: After I investigated other cleanup in `destroy()`, we already have collective calls (I have actually experienced that destroy() got stuck. This must be the reason). So let's address this issue in a seprated PR.

tests/unit/v1/moe/test_autoep_deepep_parity.py
```diff
@@ -0,0 +1,109 @@
...
+    except Exception:
+        return False
+    return True
+
+
+def _run_one_step(backend, ep_size, seed):
+    """Build a model on ``backend``, run one step, return its output and grads."""
+    seed_everything(seed)
+
+    config = make_autoep_config(ep_size=ep_size)
```
tohtana, line 51: Doesn't `make_autoep_config()` chooses FP16 before BF16 on H100, but DeepEP test should run with BF16.
The CI skipped the test but I think the test should fail.

tohtana (changes requested):
Hi @yh0903, 
Thank you for the update! Sorry for the delay, I think the earlier issues were fixed.
I left some comments about some remaining issues with details. Can you check them?

deepspeed/module_inject/auto_ep_comm.py
yh0903: You're right. I went back to DeepEP's `ElasticBuffer.dispatch()` and it documents `x` as
bfloat16 (or the fp8 pair we don't build), and the buffer is sized in bfloat16 elements
besides — so float32 would have reached the kernel assertion instead of our guard.

Removed now: `SUPPORTED_DTYPES` is now `(torch.bfloat16, )`, with
`test_rejects_fp32` covering it. Thank you for catching it.

deepspeed/module_inject/auto_ep_layer.py
yh0903: Yes, and that's the worse failure mode: the ranks that still fit would go on into the dispatch collective and block until it times out, rather than fail.

Making the check itself collective per step would put a device-to-host sync in front of every layer, which measured 1.16x -> 0.84x end-to-end, so instead I moved the check to where the buffer is sized. That all-reduce now carries `[n, -n]` under MAX, so a single collective reports both the largest and the smallest per-rank count and every rank reads the same pair.

That closes the case where the counts are ragged from the first batch. It doesn't close a job that is uniform at first and ragged later — for that one the error message now says explicitly that the other ranks will block rather than raise. Happy to make it fully collective as well!

deepspeed/module_inject/auto_ep_comm.py
yh0903: Agreed. I now replaced the global registry with `destroy_exchanges(module)`, which walks a
single module tree, so an engine releases only what its own layers built.
`test_only_the_given_module_tree_is_released` checks that a second engine's exchange is
left alone.

deepspeed/runtime/engine.py
yh0903: Thanks for digging into that. Agreed, and I've left `destroy()` as it is here. Happy to help on the follow-up if useful.

tests/unit/v1/moe/test_autoep_deepep_parity.py
yh0903: You're right, and it would have failed rather than skipped: `mixed_precision_config()` returns fp16 for any non-CPU accelerator that supports it, so the test would have tripped our own dtype guard before it compared anything. `_run_one_step` now pins bf16.
Both backends have to run the same dtype for the comparison to mean anything anyway.

deepspeed/module_inject/auto_ep_comm.py
tohtana: Thank you for the update!

deepspeed/module_inject/auto_ep_comm.py
tohtana: Thank you, it looks good now.

deepspeed/module_inject/auto_ep_layer.py
tohtana: Thank you for the fix!

deepspeed/module_inject/auto_ep_comm.py
tohtana: Thank you for the fix!

deepspeed/module_inject/auto_ep_layer.py
tohtana: Looks good to me, thank you!

deepspeed/module_inject/auto_ep_layer.py
tohtana: Thank you for the investigation! I agree that the regression from 1.16x to 0.84x makes per-layer synchronization impractical.
I think the right approach is to derive `comm_max_tokens_per_rank` from the configuration, normally as the local micro-batch size multiplied by the maximum padded sequence length.
What do you think?

tests/unit/v1/moe/test_autoep_deepep_parity.py
tohtana: Thanks! The config is now pinned to BF16, but `hidden` is still created by `torch.randn(...)` without specifying a dtype, so it remains FP32 while the model is BF16.
In my test run, both tests failed at the first LayerNorm with `expected scalar type Float but found BFloat16`. Can you create the input using the engine dtype and rerun the parity tests?

deepspeed/module_inject/auto_ep_layer.py
yh0903: That makes sense. DeepSpeed knows the local micro-batch size but not the maximum padded sequence length reliably, so deriving it automatically could either underestimate the capacity or over-allocate from
`max_position_embeddings`.

I changed the DeepEP backend to require `comm_max_tokens_per_rank` explicitly, normally train_micro_batch_size_per_gpu * maximum padded sequence length. This removes the first-batch inference and all runtime capacity synchronization. This setting also matches in other DeepEP integration examples.

tests/unit/v1/moe/test_autoep_deepep_parity.py
yh0903: Yes, the input was still FP32. It now uses the engine's configured input dtype.

While rerunning the test, I also found that the mock's unscaled expert weights amplified BF16 reduction-order differences into unrealistically large outputs, and its hidden size was below DeepEP combine's supported vector width. Those test issues are fixed now. Both H100 parity tests now run rather than skip and pass, including forward/backward numerical parity and a nonzero router-gate gradient.

tohtana (approved):
Hi @yh0903, 
Thank you for the update! This look good to me.

PR #8223, opened 2026-08-06
Fix docstring Args entries that name a parameter the function does not take

darkdi:
Forty-one `Args:` entries name a parameter the function does not take. Docstrings only — no signature, no behaviour, no test touched.

Twenty are renames where the docstring kept the old name. The ones that stand out:

| Where | Documented | Actual |
|---|---|---|
| `RaggedUnembed.forward` | `raged_metadata` | `ragged_metadata` (typo) |
| `FlopsProfiler` | `object` | `model` |
| `attn_out_in_features` | `in_features` | `out_features` |
| `RaggedTopKGating.__call__` | `expert_assignment`, `expert_offset` | `assignments`, `offsets` |
| `MoEScatter.__call__` | `hidden_states` | `activations` |
| `RaggedEmbeddingKernel.__init__` | `fp_dtype` | `embed_dtype` |
| `InferenceEngineV2.query` | `n_tokens` | `max_request_tokens` |
| `InferenceEngineV2.serialize` | `path` | `save_path` |
| `BlockedKVCache.__init__` | `config`, `enable_offload` | `configs`, `offload` |
| `MOELayer` | `expert` | `experts` |
| `DeepSpeedCPULion.__init__` | `full_precision_optimizer_states` | `fp32_optimizer_states` |
| `BertSparseSelfAttention.forward` | `attn_mask` | `attention_mask` |
| `get_grad_norm_direct` (stage 3 and stages 1/2) | `parameters` | `params` |
| `apply_to_tensors_only` | `functional` | `function` |
| `_create_model_parallel` | `model_parallel_size` | `model_parallel_size_` |
| `prune_config` | `configs` | `config` |

Where the position in the `Args:` block lined up with the position in the signature, that is what I used to decide the mapping — for example in `RaggedTopKGating.__call__` the third and fourth documented names sit against the third and fourth parameters.

Two needed a description rather than a rename, because the old text described something that is gone:

- `PipelineEngine.load_module_state_dict` documented `state_dict (str, None): unused`. The argument is `checkpoint`, and it is used on the very next line.
- `ResourceManager.parse_results` documented `finished_experiments`, which is `self.finished_experiments`. The argument is `metric`, the key read out of each experiment's metrics file.

The remaining entries document something that is not an argument at all: `layer_id` on `DeepSpeedTransformerInference`, `DeepSpeedDiffusersAttention` and `DeepSpeedMoEInference` (it is a class attribute); `beta` on `CUDARMSPreNorm.__call__`; `q_ratio` on `BlockedRotaryEmbeddings.__init__`; `slack` and `blocks` on `BlockedKVCache.__init__`; `block_size` on `DSStateManager.__init__`; `key_padding_mask_mode` and `attn_mask_mode` on `SparseSelfAttention.forward` (both are constructor arguments); `seq_len` on `DenseSparsityConfig.__init__`, whose description was a copy of the one above it; `num_global_blocks` on `BSLongformerSparsityConfig.__init__`; `scale_factor` on `DynamicLossScaler`; `scale` on the one-bit `Adam.step` and `ZeroOneAdam.step`; `max_norm` on both `get_grad_norm_direct`; and `param` and `param_id` on `_process_selected_fp32_groups_grad`.

One was a formatting slip rather than a wrong name: `quantize_transformer_layer` had a `Note:` line indented inside its `Args:` block, so Doxygen-style readers and tooling see a parameter called `Note`. Moved out.

Every entry was opened and read against its signature. `yapf` produces no diff and `flake8` is clean on all thirty-one files, using the pinned `yapf==0.40.0` and `flake8==5.0.4` from `.pre-commit-config.yaml`.


tohtana (approved):
Hi @darkdi,
Thank you for updating the docstrings! This looks good to me.

PR #8225, opened 2026-08-06
Unmanaged gradient accumulation: ZeRO offload support

sfc-gh-truwase:
## Summary
* Extends unmanaged gradient accumulation (`managed_gradient_accumulation=false`) to **ZeRO optimizer-state and parameter offload** (CPU/NVMe). Follow-up to #8217 (ZeRO stage 3, now merged).
* Stage 2/3: grads still reduce/partition every `backward()`; `step()` finalizes deferred offload boundary work (grad norms + FP32/NVMe copy) via `finalize_gradient_accumulation_boundary()`.
* Stage 1: continues to reduce at `step()` via `allreduce_gradients()`, which already performs offload boundary finalization when the boundary flag is true.
* Pipeline parallelism, DeepCompile, Apex AMP, and stage-0/1 `overlap_comm` remain unsupported.

## Test plan
Validated on a 2-GPU node:
* [x] Full `-k Unmanaged` suite (**31 passed**), including:
  * `test_unmanaged_matches_managed_optimizer_offload[1|2|3]`
  * `test_unmanaged_matches_managed_param_offload` (stage 3)
  * existing non-offload unmanaged equivalence / varying-GAS / rejection tests
* [x] Docs updated (`config-json.md`, `training.rst`); previewable on `rtd-staging`


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -961,6 +963,38 @@ def finalize_gradient_accumulation_boundary(self):
...
 
+    def _finalize_cpu_offload_gradient_accumulation(self):
+        # Deferred boundary work: restore CPU-accumulated grads and copy norms/fp32 buffers as managed boundary would.
+        for group in self.params_in_partition:
+            for param in group:
+                if not param.requires_grad:
+                    continue
+                param_id = self.get_param_id(param)
+                if param_id in self.accumulated_grads_in_cpu:
+                    self._restore_cpu_offload_grad_to_gpu(param)
```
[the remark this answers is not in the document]
sfc-gh-truwase: Addressed in e425f836: track params reduced in the current accumulation window via `_offload_accumulated_param_ids`, finalize only those IDs, and clear the set at `step()`. Regression: `TestUnmanagedGradientAccumulationOffloadInactiveParams` (stages 1/2/3).

deepspeed/runtime/zero/stage3.py
```diff
@@ -1350,10 +1350,52 @@ def independent_gradient_partition_epilogue(self):
...
+        offload_fp32_gradients = {}
+        offload_fp32_offsets = {}
+        for param_group in self.fp16_groups:
+            for param in param_group:
+                if not param.requires_grad or param.ds_id not in self.__param_id_to_grad_partition:
+                    continue
+                contains_real_data = param.partition_numel() * self._get_param_partition_rank(param) < param.ds_numel
+                if not contains_real_data:
+                    continue
+                grad_buffer = self.__param_id_to_grad_partition[param.ds_id]
```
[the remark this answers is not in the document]
sfc-gh-truwase: Addressed in e425f836: track params reduced in the current window via `_offload_boundary_param_ids`, finalize only those IDs in `_finalize_offload_gradient_accumulation()`, and clear at `_pre_step()`. Same inactive-param regression covers stage 3.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1353,10 +1355,51 @@ def independent_gradient_partition_epilogue(self):
         self._epilogue_ran_this_backward = True
 
     def finalize_gradient_accumulation_boundary(self):
-        # Unmanaged mode: grad partitions already accumulate across backwards via __param_id_to_grad_partition; nothing to finalize for non-offload.
-        assert not self.offload_optimizer and not self.offload_param, \
-            "unmanaged gradient accumulation does not support ZeRO offload"
+        # Unmanaged mode: partitions already accumulate in __param_id_to_grad_partition; offload still needs deferred boundary copy.
         self.is_gradient_accumulation_boundary = True
```
delock: It looks like `self.is_gradient_accumulation_boundary` is not consumed in this path, is it necessary to set it to true?  Maybe add a comments in initilization of this flag is for managed code only will help understanding.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1353,10 +1355,51 @@ def independent_gradient_partition_epilogue(self):
...
+        offload_fp32_offsets = {}
+        for param_group in self.fp16_groups:
+            for param in param_group:
+                if param.ds_id not in self._offload_boundary_param_ids:
+                    continue
+                if param.ds_id not in self.__param_id_to_grad_partition:
+                    continue
+                grad_buffer = self.__param_id_to_grad_partition[param.ds_id]
+                if not get_accelerator().on_accelerator(grad_buffer):
+                    grad_buffer = grad_buffer.to(get_accelerator().current_device_name(), non_blocking=True)
```
delock, line 1398: I think whether moving grads from host to device for normalization and fp32 conversion depends on host capability.  Do we consider make it an option in the furture?

delock:
Hi @sfc-gh-truwase , I have left my comments.   One impression is `self.is_gradient_accumulation_boundary` might need some refactor.  Its a flag variable that could be injected from engine, and no longer used in stage 2/3 unmanaged code.   Proper comments and set/get interface would be better than expose as variable.

delock (approved):

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -948,9 +950,11 @@ def independent_gradient_partition_epilogue(self):
         see_memory_usage("End ipg_epilogue")
 
     def finalize_gradient_accumulation_boundary(self):
-        # Unmanaged mode: grads were reduced/accumulated into all_grad_tensors each backward; finalize averaged_gradients for step().
-        assert not self.cpu_offload, "unmanaged gradient accumulation does not support ZeRO optimizer state offload"
+        # Unmanaged mode: grads accumulated each backward; finalize for step() (averaged_gradients or offload fp32 copy).
         self.is_gradient_accumulation_boundary = True
```
delock: Same here, this variable seems not used in following path.

delock (changes requested):

deepspeed/runtime/zero/stage3.py
sfc-gh-truwase: Agree to make it an option in the future.

sfc-gh-truwase:
@delock, thanks for catching the issues with `self.is_gradient_accumulation_boundary`. It is replaced with getter/setter.

delock (approved):

PR #8226, opened 2026-08-06
Fix checkpoint rank selection for Ulysses sequence parallelism

sfc-gh-thonguyen:
## Summary

- use checkpoint model-parallel rank 0 for Ulysses sequence parallelism
- centralize checkpoint rank selection across model, expert, optimizer, and ZeRO checkpoint paths
- preserve the real model-parallel rank for tensor-parallel MPUs

## Problem

Ulysses sequence parallelism does not shard model weights, so DeepSpeed records `mp_world_size == 1`. However, the Ulysses MPU aliases `get_model_parallel_rank()` to the sequence-parallel rank. Checkpoint loading combined those values when selecting a model-state file, causing SP ranks greater than zero to index past a single non-SP checkpoint shard.

The same rank interpretation was also used for checkpoint filenames, so this change keeps save and load behavior consistent for replicated SP weights and optimizer shards.

## Tests

- `pytest tests/unit/sequence_parallelism/test_ulysses.py -k 'CheckpointRank or load_non_sequence_parallel_checkpoint' -q` (3 passed)
- `pytest tests/unit/checkpoint/test_latest_checkpoint.py -q` (2 passed)
- `pre-commit run --files deepspeed/runtime/engine.py tests/unit/sequence_parallelism/test_ulysses.py`

The distributed regression saves without SP, loads module-only with SP=2, then saves and fully resumes the SP checkpoint.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -1790,6 +1790,9 @@ class DeepSpeedEngine(Module):
         self.seq_data_parallel_group = groups._get_sequence_data_parallel_group()
         self.seq_dp_world_size = groups._get_sequence_data_parallel_world_size()
         self.mp_world_size = groups._get_model_parallel_world_size()
+        self.checkpoint_mp_rank = 0
+        if self.mp_world_size > 1:
+            self.checkpoint_mp_rank = self.mpu.get_model_parallel_rank()
         self.expert_parallel_group = groups._get_expert_parallel_group_dict()
         self.expert_data_parallel_group = groups._get_expert_data_parallel_group_dict()
         self.sequence_parallel_size = groups._get_sequence_parallel_world_size()
@@ -3944,7 +3947,11 @@ class DeepSpeedEngine(Module):
                             num_experts=1,
                             checkpoint_engine=TorchCheckpointEngine(),
                             autoep_layers=None,
-                            folding_spec=None):
+                            folding_spec=None,
+                            checkpoint_mp_rank=None):
+        if checkpoint_mp_rank is None:
+            checkpoint_mp_rank = 0 if mpu is None else mpu.get_model_parallel_rank()
+
         try:
             from deepspeed.module_inject.auto_ep_layer import AutoEPMoELayer as _AutoEPMoELayer
         except ImportError:
@@ -3970,7 +3977,7 @@ class DeepSpeedEngine(Module):
                         -1,  # -1 means ignore layer_id
                         global_expert_id,
                         tag,
-                        mpu),
+                        checkpoint_mp_rank=checkpoint_mp_rank),
                     map_location=torch.device('cpu'))
 
                 # Updating global -> local expert ids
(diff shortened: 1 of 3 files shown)


deepspeed/runtime/engine.py
```diff
@@ -4137,8 +4137,16 @@ def _get_rank_zero_ckpt_name(self, checkpoints_path, tag, mp_rank, dp_rank, bf16
         )
         return zero_ckpt_name
 
+    @staticmethod
+    def _get_checkpoint_mp_rank(mpu):
```
stas00: why pass `mpu` and not get `self.mpu` from the object?

stas00: this could even be an attribute, since it's static.
```
self.checkpoint_mp_rank
```

stas00 (approved):
Thank you for the fix, Thong. 

I have a small simplification suggestion, but otherwise looks great.

deepspeed/runtime/engine.py
sfc-gh-thonguyen: Good point. Updated in 86ab256e to cache `self.checkpoint_mp_rank` when the engine initializes: it is rank 0 when `mp_world_size == 1`, otherwise it uses `self.mpu.get_model_parallel_rank()`. All engine checkpoint paths now use that attribute. The static MoE helpers receive the already-resolved integer because they are also called by `InferenceEngine`, where there is no `DeepSpeedEngine` instance.

deepspeed/runtime/engine.py
```diff
@@ -3940,7 +3943,7 @@ def load_moe_state_dict(checkpoint_path,
                             state_dict,
                             old_moe_load,
                             model=None,
-                            mpu=None,
+                            checkpoint_mp_rank=0,
```
stas00: Are you sure it's not going to break existing code? If `load_moe_state_dict` is public then removing `mpu` might break things. But it's most likely an internal function, right?

deepspeed/runtime/engine.py
sfc-gh-thonguyen: Follow-up in 6df83ff4: I restored the original `mpu` parameter in its original position and added `checkpoint_mp_rank` as a trailing optional override. Existing keyword and positional callers therefore retain the old behavior. `InferenceEngine` is unchanged and continues to pass `mpu=self.mpu`; only `DeepSpeedEngine` passes its cached checkpoint rank for the SP-safe override. I also added a regression test covering the legacy MPU argument and explicit override.

deepspeed/runtime/engine.py
sfc-gh-thonguyen: I reverted this

deepspeed/runtime/engine.py
```diff
@@ -5035,8 +5053,11 @@ def autoep_expert_writer() -> bool:
                                     zero_partition_count=folding_spec.edp_size,
                                 )
 
-                            moe_save_path = self._get_expert_ckpt_name(save_dir, moe_layer_id, global_expert_id, tag,
-                                                                       self.mpu)
+                            moe_save_path = self._get_expert_ckpt_name(save_dir,
+                                                                       moe_layer_id,
+                                                                       global_expert_id,
+                                                                       tag,
+                                                                       checkpoint_mp_rank=self.checkpoint_mp_rank)
```
stas00, line 5060: is there any time where `checkpoint_mp_rank` arg isn't `self.checkpoint_mp_rank`?

if so it doesn't need to be passed as an arg and can be pulled from `self`, no?

sfc-gh-thonguyen: In DeepSpeedEngine save paths it is always `self.checkpoint_mp_rank`. However, `_get_expert_ckpt_name` is static and is also called by static `load_moe_state_dict`, including from `InferenceEngine`, where no `DeepSpeedEngine` instance exists. The optional argument preserves those legacy/static callers while letting `DeepSpeedEngine` override the MPU rank for Ulysses. So it cannot generally pull the value from self.

Issue #7107, opened 2025-03-04
[REQUEST] An option for SUM gradient allreduce instead of MEAN


sfc-gh-lmerrick:
**Is your feature request related to a problem? Please describe.**
In some learning problems, the correct allreduce of gradients across data-parallel workers is SUM rather than MEAN. For example, when doing contrastive learning for embedding models, it is common to allgather embeddings across all workers during the forward pass, compute a global loss, and then back-propagate, making the correct gradient reduction a SUM aggregation. Both PyTorch DDP and DeepSpeed currently appear to hard-code a `1/world_size` rescaling of gradients into the distributed backwards pass before executing a SUM allreduce, essentially hard-coding the backwards pass allreduce to be a MEAN operation.

To work around this hard-coded behavior, practitioners have to manually re-scale the loss value by `world_size` before back-propagating (e.g. as [this PyTorch community post suggests](https://discuss.pytorch.org/t/is-average-the-correct-way-for-the-gradient-in-distributeddataparallel-with-multi-nodes/34260/21)). Other workarounds include disabling or manually adjusting gradient clipping in proportion to the number of workers while simultaneously using a scale-invariant optimizer like Adam.

**Describe the solution you'd like**
I would like DeepSpeed to support for an additional configuration which would simply disable the `1/world_size` scaling, e.g. my config could look like `{"zero_optimization": {"stage": 1}, "gradient_allreduce_op": "sum"}`.

**Describe alternatives you've considered**
I have thought about trying to subclass `DeepSpeedZeroOptimizer` to modify its `average_tensor` method to not actually average anything but perform a SUM allreduce (I believe that would just involve gating [line 1122](https://github.com/deepspeedai/DeepSpeed/blob/b4177e45f9c9e1c49dc2efb88f85ad40e629d04e/deepspeed/runtime/zero/stage_1_and_2.py#L1122) on a check that the config is the default value of "MEAN" instead of "SUM"). However, this would only apply to ZeRO stage1 and stage2, and I couldn't figure out a clean way to actually use my own subclass in my training loop.

**Additional context**
I chatted a bit with @stas00 about this possible feature, who mentioned this might be of interest to @tjruwase

tjruwase:
@sfc-gh-lmerrick, this would be a useful feature. Are you able to provide a PR? I will be glad to work with you.

sfc-gh-lmerrick:
[...]

PR #8232, opened 2026-08-07
Add configurable sum gradient reduction

zupengwang:
## Summary

- add a `gradient_allreduce_op` configuration with `"mean"` as the default and `"sum"` as the new option
- support unscaled gradient sums for ZeRO stages 0, 1, and 2 across reduce-scatter, allreduce, and non-contiguous fallback paths
- reject unsupported ZeRO stage 3, ZenFlow, and DeepCompile combinations with clear configuration errors
- document the option and preserve existing mean-reduction behavior

Addresses #7107.

## Motivation

Some distributed objectives, including contrastive learning over globally gathered embeddings, require summing data-parallel gradients rather than averaging them. Today users need to rescale the loss manually to cancel DeepSpeed's world-size normalization.

This change makes the reduction semantics explicit while keeping the current behavior as the default.

## Validation

Tested on two NVIDIA GeForce RTX 3090 GPUs with PyTorch 2.13.0+cu130:

- `pytest --forked -q tests/unit/v1/zero/test_zero.py::TestGradientAllreduceOp` — 18 passed
- targeted configuration tests in `tests/unit/runtime/test_ds_config_dict.py` — 8 passed; the new ZeRO-1/2 DeepCompile cases both failed against the prior head because no error was raised
- `pytest --forked -q tests/unit/v1/zero/test_zero_coalesce_grad_reduction.py::TestCoalesceCombinations` — 12 passed
- `pytest --forked -q tests/unit/runtime/sparse_tensor/test_averaging_sparse_gradients.py` — 1 passed
- changed-file `pre-commit` hooks, including YAPF, flake8, codespell, license, `check-torchdist`, and `check-torchcuda` — passed

The distributed test matrix covers ZeRO stages 0/1/2, mean and sum reductions, reduce-scatter, gradient predivide, prescale, and non-contiguous gradient fallback.

### Real-training equivalence

A deterministic two-rank, five-step `SimpleModel` regression compares the default MEAN reduction with SUM while normalizing only the SUM backward loss by `world_size`, so both modes provide identical gradients to the optimizer.

- AdamW: ZeRO-0/1/2
- Muon: ZeRO-1/2
- targeted gradient-reduction tests: `23 passed`
- complete `tests/unit/v1/zero/test_zero.py`: `98 passed, 1 skipped`

Across all five optimizer/stage configurations and all five training steps, the observed loss, full-gradient, and full-parameter differences were zero.

loss_comparison

## Limitations

`gradient_allreduce_op="sum"` is intentionally not supported with ZeRO stage 3, ZenFlow, or DeepCompile. These combinations fail during configuration instead of silently applying mean semantics.

### DIFF
--- a/deepspeed/runtime/config.py
+++ b/deepspeed/runtime/config.py
@@ -1030,6 +1030,8 @@ class DeepSpeedConfig(object):
             raise ValueError(f"{GRADIENT_ALLREDUCE_OP}='sum' is not supported with ZeRO stage 3")
         if self.gradient_allreduce_op == GRADIENT_ALLREDUCE_OP_SUM and self.zero_config.zenflow is not None:
             raise ValueError(f"{GRADIENT_ALLREDUCE_OP}='sum' is not supported with ZenFlow")
+        if self.gradient_allreduce_op == GRADIENT_ALLREDUCE_OP_SUM and self.compile_config.deepcompile:
+            raise ValueError(f"{GRADIENT_ALLREDUCE_OP}='sum' is not supported with DeepCompile")
 
         if self.float16_config.fp16_master_weights_and_grads:
             assert self.zero_enabled and self.zero_optimization_stage in (
--- a/deepspeed/runtime/config.py
+++ b/deepspeed/runtime/config.py
@@ -231,6 +231,17 @@ def get_gradient_predivide_factor(param_dict):
     return get_scalar_param(param_dict, GRADIENT_PREDIVIDE_FACTOR, GRADIENT_PREDIVIDE_FACTOR_DEFAULT)
 
 
+def get_gradient_allreduce_op(param_dict):
+    value = get_scalar_param(param_dict, GRADIENT_ALLREDUCE_OP, GRADIENT_ALLREDUCE_OP_DEFAULT)
+    if isinstance(value, str):
+        value = value.lower()
+    if value not in GRADIENT_ALLREDUCE_OP_SUPPORTED:
+        raise ValueError(
+            f"Invalid {GRADIENT_ALLREDUCE_OP}. Supported operations: {list(GRADIENT_ALLREDUCE_OP_SUPPORTED)}. "
+            f"Got: {value!r}")
+    return value
+
+
 def get_steps_per_print(param_dict):
     return get_scalar_param(param_dict, STEPS_PER_PRINT, STEPS_PER_PRINT_DEFAULT)
 
@@ -795,6 +806,7 @@ class DeepSpeedConfig(object):
             param_dict, SEQ_PARALLEL_COMMUNICATION_DATA_TYPE, SEQ_PARALLEL_COMMUNICATION_DATA_TYPE_DEFAULT)
         self.prescale_gradients = get_prescale_gradients(param_dict)
         self.gradient_predivide_factor = get_gradient_predivide_factor(param_dict)
+        self.gradient_allreduce_op = get_gradient_allreduce_op(param_dict)
         self.sparse_gradients_enabled = get_sparse_gradients_enabled(param_dict)
 
         self.zero_config = get_zero_config(param_dict)
@@ -1012,6 +1024,12 @@ class DeepSpeedConfig(object):
             assert (self.zero_optimization_stage
                     <= ZeroStageEnum.max_stage), "DeepSpeedConfig: Maximum supported ZeRO stage is {}".format(
                         ZeroStageEnum.max_stage)
+
+        if (self.gradient_allreduce_op == GRADIENT_ALLREDUCE_OP_SUM
+                and self.zero_optimization_stage == ZeroStageEnum.weights):
+            raise ValueError(f"{GRADIENT_ALLREDUCE_OP}='sum' is not supported with ZeRO stage 3")
+        if self.gradient_allreduce_op == GRADIENT_ALLREDUCE_OP_SUM and self.zero_config.zenflow is not None:
+            raise ValueError(f"{GRADIENT_ALLREDUCE_OP}='sum' is not supported with ZenFlow")
 
         if self.float16_config.fp16_master_weights_and_grads:
             assert self.zero_enabled and self.zero_optimization_stage in (
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -59,7 +59,7 @@ from deepspeed.runtime.zero.muon.muon_optimizer import MuonWithAuxAdam
 from deepspeed.runtime.constants import \
     ROUTE_TRAIN, ROUTE_PREDICT, ROUTE_EVAL, \
     PLD_THETA, PLD_GAMMA, BFLOAT16, FP16, AMP, GRADIENT_ACCUMULATION_STEPS, \
-    DATA_PARALLEL_GROUP, GLOBAL_RANK, DDP_BFLOAT16
+    DATA_PARALLEL_GROUP, GLOBAL_RANK, DDP_BFLOAT16, GRADIENT_ALLREDUCE_OP_MEAN
 from deepspeed.runtime.zero.config import ZeroStageEnum
 from deepspeed.compression import compression_scheduler
 from deepspeed.compression.constants import \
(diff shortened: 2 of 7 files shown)


deepspeed/runtime/config.py
```diff
@@ -1013,6 +1025,12 @@ def _do_error_check(self):
                     <= ZeroStageEnum.max_stage), "DeepSpeedConfig: Maximum supported ZeRO stage is {}".format(
                         ZeroStageEnum.max_stage)
 
+        if (self.gradient_allreduce_op == GRADIENT_ALLREDUCE_OP_SUM
```
pengdurice, line 1028: any reason why zero3 is not supported at this moment?

deepspeed/runtime/engine.py
```diff
@@ -3770,11 +3772,11 @@ def sparse_allreduce(self, sparse, dp_group, dp_world_size=None):
...
             dp_world_size = dist.get_world_size(group=dp_group)
-        if self.postscale_gradients():
-            if self.gradient_average:
+        if self.gradient_average:
+            if self.postscale_gradients():
                 values.mul_(self.gradient_predivide_factor() / (dp_world_size))
-        else:
-            values.mul_(1. / (dp_world_size))
+            else:
+                values.mul_(1. / (dp_world_size))
```
pengdurice, line 3779: is this where we divide by dp world size to make sure the gradients used for optimizer is still scaled?

deepspeed/runtime/config.py
zupengwang: Yes. I limited this PR to ZeRO stages 0-2 because ZeRO-3 uses a separate reduction and partition path: `reduce_scatter_coalesced()` pre-divides by world size, and the Stage 3 optimizer currently does not receive `gradient_average`. Supporting `sum` safely would require plumbing the mode through the Stage 3 optimizer and auditing its regular, contiguous, and quantized/all-to-all reduction paths, with dedicated ZeRO-3 tests. For this PR I chose to reject the combination during config validation instead of silently returning mean semantics. I can handle ZeRO-3 in a follow-up if that scope is desired.

deepspeed/runtime/engine.py
zupengwang: Yes—this is the sparse-gradient path. With the default `mean` mode, `self.gradient_average` is true, so the values are divided by the data-parallel world size before all-gather. With `gradient_allreduce_op="sum"`, `self.gradient_average` is false, this block is skipped, and the optimizer receives the unnormalized sum. The dense and ZeRO stages 1/2 paths use the same guard.

tests/unit/v1/zero/test_zero.py
```diff
@@ -28,6 +28,58 @@
...
+@pytest.mark.parametrize("gradient_allreduce_op,expected_scale", [("mean", 1.0), ("sum", 2.0)])
+@pytest.mark.parametrize(
+    "reduce_scatter,contiguous_gradients,prescale_gradients,gradient_predivide_factor",
+    [
+        (True, True, False, 1.0),
+        (False, True, False, 2.0),
+        (False, False, True, 1.0),
+    ],
+)
+class TestGradientAllreduceOp(DistributedTest):
```
pengdurice, line 43: is this overall simplistic? maybe add a real training as well and compare the losses are close for each step with the same model with or without this knob on?

pengdurice (commented):
at the same time, maybe a loss comparison figure with our without enabling this will be useful.

pengdurice (commented):
this change should be optimizer agnostic. should we test that using different optimizers, e.g. AdamW and Muon, things all work fine?

zupengwang:
@pengdurice Thanks — I agree the original single-step SGD test was too narrow.
I added a deterministic real-training regression using two data-parallel ranks and five steps. It compares the raw forward loss at every step, the full gradients before the optimizer step, and the full FP32 parameters after every step.
One normalization detail is important here: SUM produces world_size times the MEAN gradient, so an unnormalized SUM run is not generally expected to follow the same trajectory at the same learning rate. Applying lr / world_size is also not optimizer agnostic for adaptive or nonlinear optimizers. The new test therefore leaves the forward loss unchanged and divides only the SUM backward loss by world_size, which gives both runs identical optimizer inputs. The existing low-level test remains in place and independently verifies the raw unnormalized SUM = world_size × MEAN contract.
The matrix covers AdamW with ZeRO-0/1/2 and Muon with ZeRO-1/2. Muon ZeRO-0 is intentionally excluded because the current documented/runtime-supported Muon path starts at ZeRO-1.
On two RTX 3090s, the combined targeted tests passed 23/23 and the complete changed test file passed 98 tests with one existing conditional skip. The five training configurations had zero observed loss, full-gradient, and full-parameter differences across all five steps. I also prepared the attached loss-comparison figure showing the DP-averaged raw forward losses for MEAN and normalized-backward SUM.
loss_comparison

pengdurice:
> @pengdurice Thanks — I agree the original single-step SGD test was too narrow. I added a deterministic real-training regression using two data-parallel ranks and five steps. It compares the raw forward loss at every step, the full gradients before the optimizer step, and the full FP32 parameters after every step. One normalization detail is important here: SUM produces world_size times the MEAN gradient, so an unnormalized SUM run is not generally expected to follow the same trajectory at the same learning rate. Applying lr / world_size is also not optimizer agnostic for adaptive or nonlinear optimizers. The new test therefore leaves the forward loss unchanged and divides only the SUM backward loss by world_size, which gives both runs identical optimizer inputs. The existing low-level test remains in place and independently verifies the raw unnormalized SUM = world_size × MEAN contract. The matrix covers AdamW with ZeRO-0/1/2 and Muon with ZeRO-1/2. Muon ZeRO-0 is intentionally excluded because the current documented/runtime-supported Muon path starts at ZeRO-1. On two RTX 3090s, the combined targeted tests passed 23/23 and the complete changed test file passed 98 tests with one existing conditional skip. The five training configurations had zero observed loss, full-gradient, and full-parameter differences across all five steps. I also prepared the attached loss-comparison figure showing the DP-averaged raw forward losses for MEAN and normalized-backward SUM. loss_comparison

Thank you for the reply!
Can we do this? 
1. maybe change the unit test to use SimpleModel for multiple steps of training similar to https://github.com/deepspeedai/DeepSpeed/blob/79046032e5d6800a547348f6b0c7b3e1f112e5ce/tests/unit/v1/zero/test_zero_cpu_offload_grad_accum.py#L62 then assert parameter equals with our without the knob.
2. Add the figures above to the PR description. 
After that, lgtm.

zupengwang:
@pengdurice Done — the follow-up now uses the repository `SimpleModel` for deterministic two-rank, five-step training and compares the full FP32 parameters after every step with the knob set to MEAN versus SUM (with SUM backward normalized by `world_size`). The matrix covers AdamW with ZeRO-0/1/2 and Muon with ZeRO-1/2.

I also added the loss-comparison figure and the updated validation results to the PR description. The targeted tests pass 23/23, and the complete changed test file passes 98 tests with one conditional skip. Thanks!

pengdurice:
> @pengdurice Done — the follow-up now uses the repository `SimpleModel` for deterministic two-rank, five-step training and compares the full FP32 parameters after every step with the knob set to MEAN versus SUM (with SUM backward normalized by `world_size`). The matrix covers AdamW with ZeRO-0/1/2 and Muon with ZeRO-1/2.
> 
> I also added the loss-comparison figure and the updated validation results to the PR description. The targeted tests pass 23/23, and the complete changed test file passes 98 tests with one conditional skip. Thanks!

would you mind push your change to remote? I cannot find new commit yet

zupengwang:
Thanks! The change has been pushed to the PR's fork branch (`zupengwang:feature/gradient-allreduce-sum`) and is available at https://github.com/zupengwang/DeepSpeed/commit/f612ed9605bce88b89c575b4b59d4fb344c1e1c4. The PR head is now `f612ed9605bce88b89c575b4b59d4fb344c1e1c4`, and all CI checks are green.

pengdurice (approved):
lgtm, thanks for the effort!

PR #8233, opened 2026-08-07
Split DeepCompile ZeRO-3 memory scheduler

tohtana:
Splits the DeepCompile ZeRO-3 scheduling and gather/release lifetime changes out of #8169. The exact candidate passed 10/10 tests on a single node with 2 H100 GPUs.


pengdurice:
@tohtana , thank you so much for the great work! Some questions:
did we have a chance to test it wth other passes? e.g. including z3, prefetch and selective gather? Have we observed any memory / speed differences before and after this PR?

pengdurice (approved):
thanks for the great work!

PR #8241, opened 2026-08-10
[AutoTP] Replace tp_shard process-wide globals with per-model AutoTPMeta

### LINKED ISSUE #8231: Cleanup: remove tp_shard process-wide scalar globals (num_kv_heads / tp_grain_size / ...), thread explicitly
Tracking issue for a cleanup that came up while reviewing #8185. Filed short so it's not lost; not asking for it in that PR.

### Problem

`tp_shard.py` keeps `num_kv_heads`, `tp_grain_size`, `num_attention_heads`, `n_embd` as **process-wide mutable globals** (`set_num_kv_heads` / `set_tp_grain_size` during AutoTP replacement, read back in `get_shard_size` / `get_shard_size_list`).

Because they are process-wide, **a second AutoTP model loaded into the same process overwrites them**, so the first model's later sharding/gather silently reads the wrong values. This makes it unsafe to run **more than one AutoTP model per process** — teacher/student setups with different TP sizes, online distillation, or any flow that builds a second AutoTP model (including via `init_inference`) after the first.

#8185 patches one symptom with `_freeze_partition_sizes` (resolve the split once at construction, don't re-read the globals at runtime). But the globals remain the channel by which a new model's replacement clobbers the previous one's state. Removing them is the source-level fix.

### Direction

Move that state off module globals onto per-model state — compute once from the model in `AutoTP.__init__`, thread explicitly through the sharding helpers, drop `set_*` / `get_*` / the globals. Then each model carries its own kv-head/grain state and cannot stomp on another's; multiple AutoTP models can coexist in one process. Mechanical change, many call sites, hence tracked separately.

### Related

- #8185 — uneven-sharding PR whose `_freeze_partition_sizes` this generalizes.
- #8230 — universal-checkpoint schema RFC.

delock:
## What this does

Fixes #8231.

`tp_shard` kept `num_kv_heads` / `num_attention_heads` / `n_embd` / `tp_grain_size` as **process-wide mutable globals**, written during AutoTP replacement. A second AutoTP model loaded into the same process overwrote them, so the first model's later sharding / gather / checkpoint conversion silently read the wrong values — making it unsafe to run more than one AutoTP model per process (teacher/student, online distillation, RL actor + reference).

This moves that state onto a per-model `AutoTPMeta`, computed once from the model config and threaded through every sharding helper and TP layer, so each model carries its own kv-head / grain state.

## Stacking / merge order

**Depends on #8185** (AutoTP uneven sharding). This branch is based on #8185's head and is opened as **draft** until #8185 lands — GitHub will drop #8185's commits from this diff automatically once it merges. Please **merge after #8185**.

## Changes

- `AutoTPMeta` dataclass + `from_model_config` (single source for kv-head / attn-head / hidden extraction); `get_shard_size(_list)` take it as a required arg; tp_shard globals + `set_*` / `get_*` removed.
- AutoTP threads `tp_meta` from `__init__` through every TP layer and fused-QKV helper.
- Ulysses sequence parallelism gets its own `_ulysses_num_kv_heads`, decoupled from AutoTP (the AutoTP↔Ulysses coupling is gone; Ulysses's own multi-model case is left for a separate change).
- Inference engine builds one `meta` per model (`_autotp_meta`) and threads it through the alibi head-sharding helpers; `_get_model_head_count` / `_get_model_kv_head_count` deleted.
- kv-head / attn-head attribute lists unified behind `_kv_head_count_from` / `_attention_head_count_from` (covers chatglm, falcon, llama-class, dbrx, legacy `n_head_kv`).

## Tests

Validated on 4×RTX 4080 (nccl) against #8185: full AutoTP / SP / checkpoint suite passes (127 passed); remaining failures are pre-existing env issues (transformers/HF network `client has been closed`, torch 2.12 `ProcessGroupGloo.perform_nocolor_split`, a cuda/cpu device-mismatch), each confirmed failing on the #8185 baseline too. `test_two_models_do_not_clobber_each_others_meta` is the direct regression test for #8231.


deepspeed/sequence/layer.py
```diff
@@ -12,9 +12,27 @@
...
+from deepspeed.module_inject.tp_shard import AutoTPMeta, get_shard_size_list
 from deepspeed.utils import groups
 
+# Ulysses sequence parallelism keeps its own kv-head count, memoized on the first uneven
+# all-to-all. The state lives here, independent of AutoTP.
+_ulysses_num_kv_heads = None
+
+
+def set_ulysses_num_kv_heads(num):
+    global _ulysses_num_kv_heads
```
delock, line 29: A note to Ulysses SP owner: It looks like ulysses still rely on global `num_kv_heads` to work.   Do we need to remove global variable on ulysses path as well?  Agent suggested there are two path in Ulysses and the legacy path rely on this global variable.  @sfc-gh-truwase

delock:
Hi @PKUWZP @sfc-gh-truwase @jinyouzhi This PR removes globals from AutoTP processing, which allows AutoTP for both teacher and students in OPD scenario (previously they would conflict with each other and need WA).   Can you help review this PR?  Thanks!

@sfc-gh-truwase Ulysses-SP still have globals left, we may need to address it in a seperate PR.

delock:
It is notable that tests under tests/unit/model_parallelism should be run on multi-accelerator environment.   This change is better done with a seperate PR.

delock:
Thanks @chatgpt-codex-connector — the `text_config` finding was a real gap. Fixed in 88acf89 by descending into `text_config` inside `AutoTPMeta.from_model_config` itself (rather than at the two engine call sites), so every caller — runtime engine, inference engine `_autotp_meta`, direct `AutoTP` use — shares the fix. Covered by a unit test (`test_meta_descends_into_multimodal_text_config`) and an e2e regression test (`test_multimodal_outer_config_keeps_kv_head_split`, uneven 3-kv-head split must survive a nested config).

The DCO comment was a false positive: `ba7a96e` is a merge commit (2 parents); all non-merge commits on this branch carry `Signed-off-by`, and the DCO checks are green.

deepspeed/sequence/layer.py
sfc-gh-truwase: @delock yes, lets remove ulysses global variable. 

@tohtana FYI

deepspeed/module_inject/replace_module.py
```diff
@@ -278,53 +277,34 @@ def replace_wo_policy(module, all_reduce_linears, prefix="", state_dict=None):
...
-            set_num_attention_heads(getattr(model_config, 'num_attention_heads'))
-
-        # 4.4 set tp_grain_size
-        set_tp_grain_size(config.tensor_parallel.tp_grain_size)
-
-        # 5. Set linear policies
+        # 4. Set linear policies
         _autotp.update_linear_policies()
 
         # 6. Replace modules
```
jinyouzhi: 6. -> 5.

tests/unit/runtime/tensor_parallel/test_autotp_universal_checkpoint.py
```diff
@@ -474,27 +476,22 @@ def test_sub_param_layer_materializes_zero_width_final_dimension(layer_cls):
...
+                                  mp_group=None,
+                                  tp_meta=AutoTPMeta(tp_grain_size=1))
+    layer.tp_world_size = 2
+    layer.tp_index = 1
+    frozen = layer._freeze_partition_sizes(101)
+    assert frozen == (51, 50)
+
+    # A later model carries a different grain; this layer's split is frozen from its own meta.
+    other_meta = AutoTPMeta(tp_grain_size=64)
+    assert other_meta.tp_grain_size != layer.tp_meta.tp_grain_size
```
jinyouzhi: I think it would be clearer to directly assert: `assert layer.tp_meta.tp_grain_size == 1`.

Furthermore, could we construct a second TP layer with `tp_grain_size=64` and call `_freeze_partition_sizes` on it? This would verify that creating and initializing another layer does not modify the first layer’s per-model metadata or frozen partition sizes.

If feasible, it would also be useful to cover GQA explicitly: construct the second TP layer with a different `num_kv_heads` value, then verify that the first layer still retains its original `num_kv_heads` and partition layout. This would directly test that the per-model `AutoTPMeta` state is not overwritten by another model.

delock:
Moved `tests/unit/model_parallelism` to `tests/unit/v1/autotp` per @sfc-gh-truwase's feedback on #8284 (commit 123912f). This also puts the directory inside the modal GPU workflow's existing `tests/unit/v1` scope, so `TestAutoTPMultipleModels` and the multimodal regression tests will now run on multi-accelerator CI. Closed #8284 as no longer needed.

deepspeed/module_inject/replace_module.py
delock: Fixed in 606ec52 — renumbered to `5.`.

tests/unit/runtime/tensor_parallel/test_autotp_universal_checkpoint.py
delock: Done in 606ec52. The stray `other_meta` assert is gone; the test now asserts `layer.tp_meta.tp_grain_size == 1` directly, and builds a real second `LmHeadLinearAllreduce` with `tp_grain_size=64, num_kv_heads=1` (GQA), runs `_freeze_partition_sizes` on it, then re-verifies the first layer's meta (`tp_grain_size`, `num_kv_heads`) and frozen `(51, 50)` split are untouched before the forward call.

delock:
The two GPU-CI failures from the previous run are now understood:

1. `test_qwen3_norm_grads_match_across_ranks` — a real bug in this PR (`print_dist` used without import in `register_replicated_grad_hooks`); fixed in bb63a3e. Thanks to the moved tests for catching it — this path only executes on GPU with an HF tp_plan, every CPU run skipped it.

2. `test_qwen2_tied_lm_head_falls_back_to_replicated` — upstream drift, not related to this PR: transformers main (#47579, merged 2026-08-21) now unconditionally injects `embed_tokens: "embedding_rowwise"` into `base_model_tp_plan` whenever `tie_word_embeddings=True`, and our converter's allowlist rejects the whole plan on unknown styles. Tracked in #8290. The same failure will show on any PR whose diff triggers the modal GPU workflow while it tests against transformers main.

delock:
@sfc-gh-truwase Opened #8291 to track the Ulysses `_ulysses_num_kv_heads` global removal (per-model head-count threading through `DistributedAttention` / `_SeqAllToAll` ctx, incl. the backward path that cannot recover the total from tensor shapes). Keeping it out of this PR as discussed — this PR stays scoped to the AutoTP meta.

delock:
Reverted the move of `tests/unit/model_parallelism` out of `tests/unit/v1` (e9612b3). Rationale: once this PR lands, every full-suite modal run on master would hit the transformers-main `embedding_rowwise` drift tracked in #8290 (the GPU CI failure seen on the last run), turning master's modal check red for unrelated PRs. The directory will be relocated under `tests/unit/v1` in the PR that fixes #8290.

Note the run itself proved the coverage value: it caught a real `print_dist` import NameError in this PR (fixed in bb63a3e) that every CPU run skips.

delock:
Follow-up to e9612b3: the two multi-model regression tests this PR adds now live in `tests/unit/v1/autotp/test_autotp_multiple_models.py` (0fae83e) — they are new tests, not exposed to the transformers-main drift, so they stay inside the GPU workflow's scope. The pre-existing `model_parallelism` suite moves under `tests/unit/v1` later, together with the #8290 fix.

pengdurice (approved):
thank you for the great effort!

PR #8246, opened 2026-08-12
Fix repeated FlopsProfiler metric accumulation

### LINKED ISSUE #7413: [BUG] FlopsProfiler accumulates metrics when called multiple times
Hi, I am trying to measure the floating point operations of a transformer architecture and I found that the value of MACs and FLOPs increases every time I call the same model with the same input. I would expect those metrics to stay the same. I also observe variations in latency, can you provide some intuition as why I see different values of latency if I am using the same model, the same input, the same GPU and is being synchronized before the profiling. Thanks!

Here is the code to reproduce the issue:

```
import torch
from transformers import OPTForCausalLM

model_name = 'facebook/opt-125m'
device = 'cuda'
model = OPTForCausalLM.from_pretrained(model_name, torch_dtype='auto').to(device)
model.config.use_cache = False
inputs = torch.ones([32, 1024], device=device).long()
profiler = FlopsProfiler(model)
for _ in range(3):
    torch.cuda.synchronize()
    profiler.start_profile()
    model(inputs)
    profiler.stop_profile()
    flops = profiler.get_total_flops() / 1e9
    macs = profiler.get_total_macs() / 1e9
    params = profiler.get_total_params()
    latency = profiler.get_total_duration()
    profiler.end_profile()
    print(f'''macs: {macs} G, flops: {flops} G,latency: {latency:.5f} ms''')
```

Output:
[2025-07-05 10:18:52,027] [INFO] [profiler.py:82:start_profile] Flops profiler started
[2025-07-05 10:18:52,184] [INFO] [profiler.py:230:end_profile] Flops profiler finished
macs: 6522.176274432 G, flops: 13048.706236416 G,latency: 0.14194 ms
[2025-07-05 10:18:52,184] [INFO] [profiler.py:82:start_profile] Flops profiler started
[2025-07-05 10:18:52,340] [INFO] [profiler.py:230:end_profile] Flops profiler finished
macs: 7140.651565056 G, flops: 14285.656817664 G,latency: 0.14185 ms
[2025-07-05 10:18:52,341] [INFO] [profiler.py:82:start_profile] Flops profiler started
[2025-07-05 10:18:52,499] [INFO] [profiler.py:230:end_profile] Flops profiler finished
macs: 7759.12685568 G, flops: 15522.607398912 G,latency: 0.14374 ms

Moreover, if I ignore some high level modules the output is now consistent:

```
import torch
from transformers import OPTForCausalLM
from transformers.models.opt.modeling_opt import OPTAttention, OPTDecoderLayer, OPTDecoder, OPTForCausalLM

model_name = 'facebook/opt-125m'
device = 'cuda'
model = OPTForCausalLM.from_pretrained(model_name, torch_dtype='auto').to(device)
model.config.use_cache = False
inputs = torch.ones([32, 1024], device=device).long()
profiler = FlopsProfiler(model)
for _ in range(3):
    torch.cuda.synchronize()
    profiler.start_profile(ignore_list=[OPTAttention, OPTDecoderLayer, OPTDecoder, OPTForCausalLM])
    model(inputs)
    profiler.stop_profile()
    flops = profiler.get_total_flops() / 1e9
    macs = profiler.get_total_macs() / 1e9
    params = profiler.get_total_params()
    latency = profiler.get_total_duration()
    #profiler.print_model_profile()
    profiler.end_profile()
    print(f'''macs: {macs} G, flops: {flops} G,latency: {latency:.5f} ms''')
```

Output:
[2025-07-05 10:10:11,202] [INFO] [profiler.py:82:start_profile] Flops profiler started
[2025-07-05 10:10:11,359] [INFO] [profiler.py:230:end_profile] Flops profiler finished
macs: 4048.275111936 G, flops: 8100.903911424 G,latency: 0.11131 ms
[2025-07-05 10:10:11,359] [INFO] [profiler.py:82:start_profile] Flops profiler started
[2025-07-05 10:10:11,516] [INFO] [profiler.py:230:end_profile] Flops profiler finished
macs: 4048.275111936 G, flops: 8100.903911424 G,latency: 0.11201 ms
[2025-07-05 10:10:11,517] [INFO] [profiler.py:82:start_profile] Flops profiler started
[2025-07-05 10:10:11,675] [INFO] [profiler.py:230:end_profile] Flops profiler finished
macs: 4048.275111936 G, flops: 8100.903911424 G,latency: 0.11289 ms

Versions:
torch.__version__ = '2.7.0+cu126'
transformers.__version__ = '4.52.4'
deepspeed.__version__ = '0.16.9'

Thank you!

baremetaldevx86:
## Summary

  - restore `F.scaled_dot_product_attention` after profiling
  - restore `Tensor.__matmul__` after profiling
  - restore `torch.bmm` from its correct saved implementation
  - add a CPU regression test covering repeated profiling sessions and operation restoration
  - scope the existing FP16 skip to the test that actually requires FP16

  ## Problem

  `FlopsProfiler` temporarily replaces PyTorch operations with FLOP-counting wrappers. Its cleanup path did not restore `F.scaled_dot_product_attention` or
  `Tensor.__matmul__`, causing wrappers to accumulate across profiling sessions. As a result, identical model executions reported progressively increasing FLOPs and
  MACs.

  The `torch.bmm` cleanup was also incorrect:

  ```python
  torch.bmm = old_functions[torch.matmul.__str__]
 ```
  Because torch.matmul had already been restored, this rebound torch.bmm to torch.matmul for the rest of the process. This changed normal PyTorch behavior after
  profiling: torch.bmm began accepting inputs supported by broadcasting matmul but invalid for bmm.

  ## Fix

  Restore every patched operation from its matching saved original function. The patch and cleanup paths now cover the same set of operations with matching PyTorch
  version guards.

  ## Testing

  A CPU regression test profiles the same scaled-dot-product-attention operation three times and verifies:

  - every session reports identical FLOP and MAC totals
  - F.scaled_dot_product_attention, Tensor.__matmul__, and torch.bmm are restored to their original function objects after every session

  Observed totals:

  Before:
  [(65536, 32768), (131072, 65536), (196608, 98304)]

  After:
  [(65536, 32768), (65536, 32768), (65536, 32768)]

  All pre-commit checks pass.

  Fixes #7413

### DIFF
--- a/deepspeed/profiling/flops_profiler/profiler.py
+++ b/deepspeed/profiling/flops_profiler/profiler.py
@@ -1013,14 +1013,17 @@ def _reload_functionals():
     F.interpolate = old_functions[F.interpolate.__str__]
     F.softmax = old_functions[F.softmax.__str__]
     F.embedding = old_functions[F.embedding.__str__]
+    if required_torch_version(min_version=2.0):
+        F.scaled_dot_product_attention = old_functions[F.scaled_dot_product_attention.__str__]
 
 
 def _reload_tensor_methods():
     torch.matmul = old_functions[torch.matmul.__str__]
     torch.Tensor.matmul = old_functions[torch.Tensor.matmul.__str__]
+    torch.Tensor.__matmul__ = old_functions[torch.Tensor.__matmul__.__str__]
     torch.mm = old_functions[torch.mm.__str__]
     torch.Tensor.mm = old_functions[torch.Tensor.mm.__str__]
-    torch.bmm = old_functions[torch.matmul.__str__]
+    torch.bmm = old_functions[torch.bmm.__str__]
     torch.Tensor.bmm = old_functions[torch.Tensor.bmm.__str__]
     torch.addmm = old_functions[torch.addmm.__str__]
     torch.Tensor.addmm = old_functions[torch.Tensor.addmm.__str__]
--- a/tests/unit/profiling/flops_profiler/test_flops_profiler.py
+++ b/tests/unit/profiling/flops_profiler/test_flops_profiler.py
@@ -14,9 +14,6 @@ from unit.common import DistributedTest
 from deepspeed.utils.torch import required_torch_version
 from deepspeed.accelerator import get_accelerator
 
-if torch.half not in get_accelerator().supported_dtypes():
-    pytest.skip(f"fp16 not supported, valid dtype: {get_accelerator().supported_dtypes()}", allow_module_level=True)
-
 pytestmark = pytest.mark.skipif(not required_torch_version(min_version=1.3),
                                 reason='requires Pytorch version 1.3 or above')
 
@@ -26,6 +23,34 @@ def within_range(val, target, tolerance):
 
 
 TOLERANCE = 0.05
+
+
+@pytest.mark.sequential
+@pytest.mark.skipif(not required_torch_version(min_version=2.0), reason="requires Pytorch version 2.0 or above")
+def test_repeated_profile_restores_operations():
+
+    class AttentionModel(torch.nn.Module):
+
+        def forward(self, query, key, value):
+            return torch.nn.functional.scaled_dot_product_attention(query, key, value)
+
+    model = AttentionModel()
+    inputs = [torch.randn(2, 4, 16, 8) for _ in range(3)]
+    prof = FlopsProfiler(model)
+    original_operations = (torch.nn.functional.scaled_dot_product_attention, torch.Tensor.__matmul__, torch.bmm)
+
+    profiles = []
+    for _ in range(3):
+        prof.start_profile()
+        model(*inputs)
+        prof.stop_profile()
+        profiles.append((prof.get_total_flops(), prof.get_total_macs()))
+        prof.end_profile()
+
+        restored_operations = (torch.nn.functional.scaled_dot_product_attention, torch.Tensor.__matmul__, torch.bmm)
+        assert restored_operations == original_operations
+
+    assert profiles == [(65536, 32768)] * 3
 
 
 class LeNet5(torch.nn.Module):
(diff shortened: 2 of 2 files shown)


ebarkhordar (commented):
Reproduced both sides on CPU (clean `python:3.11-slim`, torch 2.13.0+cpu), base `1d580d65` against head `997dabee`, three rounds on the SDPA model from your test:

```
base: [(65536, 32768), (131072, 65536), (196608, 98304)]
head: [(65536, 32768), (65536, 32768), (65536, 32768)]
```

The `bmm` line is more than a metrics bug, and I think the description should say so. `torch.bmm = old_functions[torch.matmul.__str__]` runs after `torch.matmul` has already been restored on the line above, so it rebinds `torch.bmm` to `torch.matmul` for the rest of the process. matmul broadcasts where bmm requires 3-D: after profiling a plain `nn.Linear` once, `torch.bmm(4x4, 4x4)` returns a 4x4 instead of raising `RuntimeError: batch1 must be a 3D tensor`. Any profiled model reaches it, not only attention ones. Head restores it correctly.

To check nothing of the same shape is left, I parsed `_patch_*` and `_reload_*`: base is 51 patched against 49 reloaded (exactly the two you add) plus the mis-keyed `bmm`; head is 51/51 with matching version guards. So that is the complete set.

One suggestion. The module is skipped entirely when the accelerator does not report fp16 (line 17), so the new test does not collect on a CPU-only runner (`1 skipped, collected 0 items` here) even though it needs neither fp16 nor a device. Hoisting it above that guard would let the `cpu-torch-latest` `-m sequential` leg run it.

baremetaldevx86:
Thanks for the thorough verification. Agreed that the incorrect `torch.bmm` restoration has broader process-wide correctness implications, not just inaccurate profiler metrics. I’ll make that explicit in the PR description.

I also addressed the CPU collection issue. Since a module-level `pytest.skip(..., allow_module_level=True)` aborts collection of the entire file even if the test is defined above it, I moved the FP16 capability check into the existing test that actually uses FP16. The repeated-profile regression now collects and passes in the CPU-only sequential test path.

tohtana (approved):
@baremetaldevx86 Thank you for the fix! This looks good to me.

PR #8254, opened 2026-08-14
Fix DeepCompile last use for unconsumed waits

tohtana:
With this PR, DeepCompile treats an unconsumed `wait_allgather` alias as its own last use to prevent `KeyError`, while preserving downstream propagation for consumed aliases.
Add synthetic FX regression coverage for both orphan-only and later-real-use cases.


pengdurice (approved):
thanks for the fix!

PR #8255, opened 2026-08-15
Enable activation offloading

pengdurice:
# Activation offloading for DeepCompile: plan against a floor the run has actually reached

## What this does

Enables the activation-offload pass to DeepCompile init_z3. The forward pass saves tensors the backward needs;
they sit in device memory for the whole step. This pass copies chosen ones to pinned host memory
during the forward and brings each back before the backward reads it.

Enable with `compile.offload_activation: true`. It replaces the prefetch and selective-gather passes
and is mutually exclusive with `offload_parameters` and `offload_opt_states` — each plans against the
whole budget on its own.

## Results

Where a job is too tight to run at all, offloading everything rescues it but costs 40%+ in step
time. Planning which activations to move recovers most of that.

Qwen3-14B, mb4, 8xH200, ZeRO-3, inductor, bf16 optimizer states, `expandable_segments`. All arms use
DeepCompile; the question is what the offload pass adds.

| seq | baseline (no offload) | forced-all (blind) | ours (planned) | ours moved | vs forced-all |
| --- | --- | --- | --- | --- | --- |
| 2048 | 1.76s | 2.50s | **1.85s** | 0GB — declines | run fits; pass correctly does nothing |
| 3072 | 5.46s | 7.81s | **5.41s** | 0GB — declines | run fits; pass correctly does nothing |
| 3328 | **OOM** | 8.67s | **7.38s** | 11.7GB of 22.0 | **15% faster** |
| 3584 | **DIED** (watchdog) | 10.80s | **9.76s** | 20.5GB of 23.7 | **9.6% faster** |
| 4096 | **OOM** | 11.17s | 11.65s | 26.8GB of 27.0 | declines to plan; safe |

Three properties, each measured:

- **It resolves the out-of-memory case.** At seq3328/3584/4096 the job does not run without it.
- **It beats moving everything, where there is room to plan.** 15% at seq3328 and 9.6% at seq3584 --
  the two lengths that need offloading at all -- moving roughly half the bytes. Copy cost is linear
  at 0.108s per GB (R²=0.96), which is where the time comes from.
- **It does nothing when nothing is needed** — at seq2048/3072. It keeps every activation resident
  and costs nothing measurable, while blind offloading costs +42%/+43%.

At seq4096 there is only ~2.3GiB of real headroom, so it correctly declines and matches forced-all
rather than dying.

The place this pass would pay is where recompute is expensive and PCIe is idle -- long sequences
where attention recompute scales O(s^2) while a copy stays O(s), or composed with recompute rather
than against it. Neither is measured here.


sfc-gh-truwase:
@pengdurice thanks for this great PR. How does it interact with the [selective activation checkpointing](https://deepspeed.readthedocs.io/en/latest/training.html#memory-optimization) introduced by [AutoSP](https://github.com/deepspeedai/DeepSpeed/pull/7860)?

sfc-gh-truwase:
@pengdurice in particular, I am curious how the offloading heuristics for this PR compares with checkpointing heuristics of AutoSP. Is it possible to combine them?

tests/unit/v1/compile/test_list_schedule.py
```diff
@@ -40,6 +41,10 @@ def _define_dc_ops():
             "reduce_grad(Tensor a, int graph_id, int id) -> Tensor",
             "free_tensors(Tensor[] tensors) -> ()",
             "end_backward(Tensor[] tensors, int graph_id, bool release_reduce_buckets = True) -> ()",
+            "offload_tensor(Tensor a, int id, int id) -> Tensor",
```
sfc-gh-truwase, line 44: Move to tests/unit/v1/compile

deepspeed/compile/config.py
```diff
@@ -22,7 +22,13 @@ class CompileConfig(DeepSpeedConfigModel):
     """ In free activation mode, activations no less than this threshold (in byte) are eagerly freed """
 
     offload_activation: bool = False
-    """ Turn on/off the activation offloading """
+    """ Move activations that the forward pass saves for the backward pass to pinned host memory,
+    and bring each one back shortly before the backward pass reads it. Only tensors of at least 10MB
+    with a fixed shape are considered, and only as many as the memory budget requires. Runs in place
+    of the prefetch/selective-gather passes and is mutually exclusive with offload_parameters and
+    offload_opt_states. Designed for gradient_accumulation_steps=1: the host buffers are reused
+    every time the compiled graph runs, so a micro-batch must finish its backward pass before the
```
sfc-gh-truwase: - Is there an assert for `gas>1`?
- Is there an assert for [unmanaged gas](https://deepspeed.readthedocs.io/en/latest/training.html#unmanaged-gradient-accumulation) mode?

pengdurice:
> @pengdurice in particular, I am curious how the offloading heuristics for this PR compares with checkpointing heuristics of AutoSP. Is it possible to combine them?

Thank you for your review!  What this heuristics does that it will offload the activations if not offloading will cause usage of memory > a budget and OOM. The AutoSP's recompute strategy decides what to recompute and what to be saved, those saved are visible to this FX graph and this PR decides if we keep them in device memory or offload to host memory. Definitely they can be combined, currently, z3 and autoSP passes are not compatible, there is a WIP branch to combine them.

tests/unit/v1/compile/test_list_schedule.py
pengdurice: great idea, done.

deepspeed/compile/config.py
pengdurice: Thank you for looking at this, after given it another thought, it occurs to me that for gas > 1 it should work without issue since the offloaded gradients are brought back for each corresponding backward to use, and after the usage, the pinned memory should be free for next forward (backed by the event records of the streams). the test with gas= 1 and 2 all passed confirming that gas > 1 is unblocking. For unmanaged gas,  this line (https://github.com/deepspeedai/DeepSpeed/blob/310eb8cb485ca30cc951f7735f5e342126ea5b55/deepspeed/runtime/engine.py#L1694) has already asserted this. Codes are changed correspondingly. Thank you!

sfc-gh-truwase (approved):

PR #8256, opened 2026-08-15
Route ZeRO/SuperOffload pin sites through accelerator pin_memory

sfc-gh-truwase:
## Summary
- Route ZeRO-1/2 `offload_states` hp/lp pins, `offload_optimizer_states`, ZeRO++ secondary shards, and SuperOffload grad buffers through `get_accelerator().pin_memory()` so `DS_PIN_MEMORY_BACKEND=native` applies.
- Use `make_copy=False` for empty scratch destinations to avoid an extra host alloc/copy under the native backend.
- Clarify in `memory.rst` that ZeRO `offload_*.pin_memory` (whether) is orthogonal to `DS_PIN_MEMORY_BACKEND` (how); env-only, no ds_config backend field.
- Add `tests/unit/v1/pin_memory/test_offload_route.py` covering `offload_optimizer_states` under native.

## Test plan
- [x] `pre-commit run --files` on touched paths (already run locally)
- [x] `DS_PIN_MEMORY_BACKEND=native` + `pytest tests/unit/v1/pin_memory/test_offload_route.py` (+ pin_memory/accelerator UTs) on H200 host with pin_memory op — **23 passed** (incl. both offload_route tests)
- [x] Smoke ZeRO-1/2 CPU offload with default torch backend — `test_offload_states.py` + `test_destroy_unpin.py` — **133 passed**
- [x] SuperOffload worker pin site: source uses `get_accelerator().pin_memory(..., make_copy=False)`; native empty-buffer pin/unpin smoke OK

Evidence: autorun `job-20260818T143819Z` on `e6dc5f9d` / `tjruwase/pin-memory-route-zero` (EXIT 0). GitHub CI also green on the tip commit.

### DIFF
--- a/accelerator/abstract_accelerator.py
+++ b/accelerator/abstract_accelerator.py
@@ -268,6 +268,9 @@ class DeepSpeedAccelerator(ABC):
     def _torch_pin_memory(self, tensor):
         return tensor.pin_memory()
 
+    def _torch_empty_pinned(self, tensor, shape):
+        return tensor.new_empty(shape, pin_memory=True)
+
     def _torch_is_pinned(self, tensor):
         return tensor.is_pinned()
 
@@ -278,7 +281,12 @@ class DeepSpeedAccelerator(ABC):
         pins = get_active_native_pinned_memory()
         if pins is not None:
             return pins.pin(tensor, make_copy=make_copy, match_shape=match_shape)
-        return self._torch_pin_memory(tensor)
+        if make_copy:
+            return self._torch_pin_memory(tensor)
+        # ``tensor`` is only a shape/dtype template here, so page-lock a fresh
+        # buffer instead of faulting it in and copying it into a second one.
+        shape = tensor.shape if match_shape else (tensor.numel(), )
+        return self._torch_empty_pinned(tensor, shape)
 
     def is_pinned(self, tensor):
         from deepspeed.utils.pin_memory import get_active_native_pinned_memory
--- a/deepspeed/runtime/superoffload/superoffload_utils.py
+++ b/deepspeed/runtime/superoffload/superoffload_utils.py
@@ -13,6 +13,7 @@ import torch.multiprocessing as mp
 import psutil
 
 from deepspeed.ops.adam import DeepSpeedCPUAdam
+from deepspeed.accelerator import get_accelerator
 from deepspeed.utils import logger
 
 
@@ -82,7 +83,8 @@ def superoffload_optimizer_worker(param_queue: mp.SimpleQueue, result_queue: mp.
         return
 
     # Pre-allocate reusable pinned memory buffer for gradients
-    pinned_grad_buffer = torch.empty(max_grad_numel, dtype=torch.float32, device='cpu', pin_memory=True)
+    pinned_grad_buffer = get_accelerator().pin_memory(torch.empty(max_grad_numel, dtype=torch.float32, device='cpu'),
+                                                      make_copy=False)
 
     while True:
         try:
--- a/deepspeed/runtime/superoffload/superoffload_utils.py
+++ b/deepspeed/runtime/superoffload/superoffload_utils.py
@@ -36,8 +36,21 @@ class EventTypes:
     ROLLBACK = "rollback"
 
 
-def superoffload_optimizer_worker(param_queue: mp.SimpleQueue, result_queue: mp.SimpleQueue,
-                                  optimizer_config: Dict[str, Any], max_grad_numel: int) -> None:
+def _allocate_worker_grad_buffer(max_grad_numel: int, pin_memory: bool) -> torch.Tensor:
+    # Scratch destination for the GPU->CPU grad copy. Pinning is optional: it
+    # speeds DMA but is not required for correctness, so honor
+    # offload_optimizer.pin_memory rather than always page-locking.
+    buffer = torch.empty(max_grad_numel, dtype=torch.float32, device='cpu')
+    if pin_memory:
+        buffer = get_accelerator().pin_memory(buffer, make_copy=False)
+    return buffer
+
+
+def superoffload_optimizer_worker(param_queue: mp.SimpleQueue,
+                                  result_queue: mp.SimpleQueue,
+                                  optimizer_config: Dict[str, Any],
+                                  max_grad_numel: int,
+                                  pin_memory: bool = True) -> None:
     """
     This function runs in a separate process and continuously processes optimization
     tasks from the parameter queue. It creates a DeepSpeedCPUAdam optimizer and
@@ -49,6 +62,7 @@ def superoffload_optimizer_worker(param_queue: mp.SimpleQueue, result_queue: mp.
         optimizer_config: Configuration dictionary for the optimizer containing
                          lr, betas, eps, weight_decay, and amsgrad parameters
         max_grad_numel: Maximum number of elements expected in gradient tensors
+        pin_memory: Whether to page-lock the reusable grad scratch buffer
     """
     cpu_tensor = torch.randn(1, device="cpu")
     cpu_param = torch.nn.Parameter(cpu_tensor)
@@ -82,9 +96,8 @@ def superoffload_optimizer_worker(param_queue: mp.SimpleQueue, result_queue: mp.
         result_queue.put({"error": error_msg})
         return
 
-    # Pre-allocate reusable pinned memory buffer for gradients
-    pinned_grad_buffer = get_accelerator().pin_memory(torch.empty(max_grad_numel, dtype=torch.float32, device='cpu'),
-                                                      make_copy=False)
+    # Pre-allocate reusable host buffer for gradients
+    grad_buffer = _allocate_worker_grad_buffer(max_grad_numel, pin_memory)
 
     while True:
         try:
@@ -118,7 +131,7 @@ def superoffload_optimizer_worker(param_queue: mp.SimpleQueue, result_queue: mp.
                 result_queue.put({"error": error_msg})
                 break
 
-            param_grad_cpu = pinned_grad_buffer[:grad_numel].view_as(param_grad)
+            param_grad_cpu = grad_buffer[:grad_numel].view_as(param_grad)
             param_grad_cpu.copy_(param_grad, non_blocking=False)
 
             fp32_param = torch.nn.Parameter(param_data)
@@ -156,10 +169,13 @@ def superoffload_optimizer_worker(param_queue: mp.SimpleQueue, result_queue: mp.
             result_queue.put({"error": error_msg})
             break
 
-    # Clean up pinned memory buffer
-    if 'pinned_grad_buffer' in locals():
-        del pinned_grad_buffer
-        logger.debug("Cleaned up pinned memory buffer")
+    # Clean up the reusable host buffer. unpin_memory is a no-op on torch and
+    # frees immediately on the native backend.
+    if 'grad_buffer' in locals():
+        if pin_memory:
+            get_accelerator().unpin_memory(grad_buffer)
+        del grad_buffer
+        logger.debug("Cleaned up worker grad buffer")
 
     logger.debug("Worker process terminated")
 
@@ -169,7 +185,8 @@ class SuperOffloadCPUOptimizer:
     def __init__(self,
                  optimizer_config: Dict[str, Any],
                  cpuadam_cores_perc: float = 0.8,
-                 max_grad_numel: int = 1000000) -> None:
+                 max_grad_numel: int = 1000000,
+                 pin_memory: bool = True) -> None:
         if not 0 < cpuadam_cores_perc <= 1:
             raise ValueError("cpuadam_cores_perc must be between 0 and 1")
 
@@ -180,7 +197,7 @@ class SuperOffloadCPUOptimizer:
 
         self.cpuadam_process = self.mp_context.Process(
             target=superoffload_optimizer_worker,
-            args=(self.param_queue, self.result_queue, optimizer_config, max_grad_numel),
+            args=(self.param_queue, self.result_queue, optimizer_config, max_grad_numel, pin_memory),
             daemon=True,
         )
         self.cpuadam_process.start()
--- a/deepspeed/runtime/zero/offload_states.py
+++ b/deepspeed/runtime/zero/offload_states.py
@@ -27,10 +27,18 @@ def offload_optimizer_states(optimizer, device, pin_memory=False, non_blocking=F
 
 
 def reload_optimizer_states(optimizer, device, non_blocking=False):
+    """Move optimizer states back to ``device``.
+
+    Returns the replaced host tensors. A non-blocking copy may still be reading
+    them on return, so callers must keep them alive until they synchronize.
+    """
+    host_buffers = []
     for state in optimizer.state.values():
         for k, v in state.items():
             if torch.is_tensor(v):
+                host_buffers.append(v)
                 state[k] = v.to(device, non_blocking=non_blocking)
+    return host_buffers
 
 
 def offload_adam_states(optimizer, device, pin_memory: bool = False, non_blocking: bool = False):
--- a/deepspeed/runtime/zero/offload_states.py
+++ b/deepspeed/runtime/zero/offload_states.py
@@ -41,6 +41,21 @@ def reload_optimizer_states(optimizer, device, non_blocking=False):
     return host_buffers
 
 
+def unpin_offloaded_optimizer_states(optimizer):
+    """Release host buffers that offloading page-locked in the optimizer state.
+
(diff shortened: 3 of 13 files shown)


delock (approved):

delock:
Hi @sfc-gh-truwase , this PR looks good to me.  There is only one thoughts, the phrase " get_accelerator().pin_memory(torch.empty(...), make_copy=False)" keeps repeating and it looks like an idiom.  Maybe it worth a place in utility.

sfc-gh-truwase:
@delock great point. I will address in a separate PR.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2941,7 +2941,8 @@ def needs_offload(target):
             if pin_memory:
                 if not hasattr(self, "hp_params_pin_buffers"):
                     self.hp_params_pin_buffers = [
-                        torch.empty_like(t, device=device).pin_memory() for t in self.single_partition_of_fp32_groups
+                        get_accelerator().pin_memory(torch.empty_like(t, device=device), make_copy=False)
```
xylian86, line 2956: I have one concern on lifetime of these native-pinned buffers. During `reload_states(non_blocking=True)`, the CPU-to-GPU copy may still be running when the HP/LP owner lists are deleted.

sfc-gh-truwase: @xylian86, can you explain a bit more? Is your concern specific to this PR or more generally about the lifetime of native pinning?

xylian86: Yes, for sure. The flow involves three steps:

Step 1. start copy
https://github.com/deepspeedai/DeepSpeed/blob/aa3914df82e8ba2d5dc2462b06c0160415e674db/deepspeed/runtime/zero/stage_1_and_2.py#L3019
Step 2. delete source memory
https://github.com/deepspeedai/DeepSpeed/blob/aa3914df82e8ba2d5dc2462b06c0160415e674db/deepspeed/runtime/zero/stage_1_and_2.py#L3021
Step 3. wait for copy to finish
https://github.com/deepspeedai/DeepSpeed/blob/aa3914df82e8ba2d5dc2462b06c0160415e674db/deepspeed/runtime/zero/stage_1_and_2.py#L3062-L3063


With PyTorch’s pinned-memory allocator, dropping the tensor does not immediately release its underlying memory. The allocator tracks the CUDA stream using the buffer and retains the allocation until the copy completes. 

My understanding is that DeepSpeed’s native allocator tracks Python object ownership but does not track the CUDA stream using the buffer. Therefore, step 2 may free the source allocation before the synchronization in step 3. 

Could we retain these source buffers until after synchronization, or add a native nonblocking GPU test to confirm that the current lifetime is safe?

sfc-gh-truwase: Good catch. 
1. Yes, we can retain the source buffers until after synchronization
2. I will add native nonblocking GPU tests 
3. I will investigate whether #8283 enables CUDA stream tracking

sfc-gh-truwase: Thanks @xylian86 — you were right, and the fix is pushed in `aebb2819`.

**1. Retain the source buffers until after synchronization.** `reload_states` now holds every host source in a local list and releases it only after the `synchronize()`:

```python
        pending_host_buffers = []
        ...
            if hasattr(self, "hp_params_pin_buffers"):
                pending_host_buffers.append(self.hp_params_pin_buffers)
                del self.hp_params_pin_buffers
        ...
        if non_blocking:
        # The copies have completed, so the host sources can be released.
        del pending_host_buffers
```

`reload_optimizer_states` had the same hazard — it replaced `state[k]`, dropping the pinned source right away — so it now returns the replaced host tensors and `reload_states` keeps them in the same list.

ZeRO-3 needs no change: `hp_params_pin_buffers`, `lp_param_contiguous_pin_buffer`, `lp_grad_partitions_flat_pin_buffers`, and the Adam offload buffers all stay reachable as attributes across reload, so nothing is freed mid-copy.

**2. Native non-blocking GPU tests.** `test_reload_states_holds_pin_buffers_until_sync` drives the real `reload_states(non_blocking=True)` with a native-pinned hp buffer and records whether the allocation is still live at the moment `synchronize()` is called. On 1x H200 it passes with the fix and fails `assert [False] == [True]` with the two source files reverted, so it genuinely catches the use-after-free. `test_reload_optimizer_states_returns_host_sources` covers the helper contract.

Full run on H200:

| Run | Result |
|-----|--------|
| `test_offload_route.py` (native) | 4 passed |
| same, fix reverted | fails as expected |
| `test_offload_states.py`, 2 GPUs, torch backend | 128 passed |
| `test_offload_states.py`, 2 GPUs, native backend | 128 passed |

**3. Does #8283 give stream tracking?** No. `cudaHostRegister` only makes the pages DMA-capable; deferred, stream-ordered reuse is a property of torch's caching host allocator, which the native (`posix_memalign` + `mlock`) allocator does not have. Registration arguably makes the hazard more visible, since the copy becomes a true async DMA rather than a synchronous staged copy. I documented the caveat in `memory.rst` so future call sites know that `non_blocking=True` out of a native-pinned buffer requires holding a reference until synchronization.

sfc-gh-truwase: @xylian86 do you have additional concerns?

deepspeed/runtime/zero/offload_states.py
```diff
@@ -19,18 +19,26 @@ def offload_optimizer_states(optimizer, device, pin_memory=False, non_blocking=F
         for k, v in state.items():
             if torch.is_tensor(v):
                 if pin_memory and v.device.type != 'cpu':
-                    pinned_buffer = torch.empty_like(v, device='cpu').pin_memory()
+                    pinned_buffer = get_accelerator().pin_memory(torch.empty_like(v, device='cpu'), make_copy=False)
```
xylian86, line 22: If I understand correctly, the previous implementation, `torch.empty(N, pin_memory=True)`, directly allocated a single pinned buffer.

With the current implementation, the default Torch backend first allocates a pageable buffer with `torch.empty(N)`, and then `pin_memory()` allocates a second pinned buffer and copies the data into it. Since `make_copy=False` is only honored by the native backend and ignored by Torch, this temporarily requires **two** full-size buffers and introduces an unnecessary copy, potentially **doubling** peak host memory usage during initialization.

deepspeed/runtime/zero/stage_1_and_2.py
xylian86: @sfc-gh-truwase  Hi Tunji, thanks for confirming and fixing it! Everything else looks good to me, except for one part where I added a new review comment in the following thread.

deepspeed/runtime/zero/offload_states.py
sfc-gh-truwase: Good point. However, previous implementation was `torch.empty_like(v, device='cpu').pin_memory()` which suffers the same memory doubling and waste copy. So this is an opportunity to improve the torch backend path to avoid these overheads.

sfc-gh-truwase: Fixed in `246c02e1` — the torch path now page-locks directly instead of pinning a copy.

```python
        if make_copy:
            return self._torch_pin_memory(tensor)
        # ``tensor`` is only a shape/dtype template here, so page-lock a fresh
        # buffer instead of faulting it in and copying it into a second one.
        shape = tensor.shape if match_shape else (tensor.numel(), )
        return self._torch_empty_pinned(tensor, shape)
```

The new `_torch_empty_pinned` hook defaults to `tensor.new_empty(shape, pin_memory=True)`. XPU and HPU override it to keep allocate-then-pin, since their `pin_memory(device=...)` form has no allocation-side equivalent, and the CPU accelerator is unaffected because it overrides `pin_memory` and returns the tensor before this branch. Both backends now honor `make_copy` and `match_shape` identically, so `memory.rst` no longer claims torch ignores them.

The `torch.empty_like(v, device='cpu')` template is never read or written on this path, so it stays virtual and only one full-size buffer becomes resident. Measured on an H200 with a 1 GiB fp32 request:

```
buffer_is_pinned=True
bytes_requested_mib=1024
peak_rss_growth_mib=1127
```

Tests: `test_pin_memory_torch_backend_make_copy_false_allocates_pinned` asserts the dispatch with a stub accelerator so it runs on CPU CI (`make_copy=True` pins a copy; `make_copy=False` allocates, flat when `match_shape=False`), and `test_pin_memory_torch_backend_no_copy_is_pinned` checks the real buffer is pinned and distinct from the input.

H200 validation at `246c02e1`: accelerator + pin_memory + offload-route unit tests **27 passed**; `test_offload_states.py` on 2 GPUs **128 passed**; `test_destroy_unpin.py` **5 passed**.

Worth noting this does not remove the template allocation itself — that needs a shape/dtype-based entry point, which is the utility @delock suggested and I will do in a follow-up PR.

xylian86 (approved):

deepspeed/runtime/zero/offload_states.py
xylian86: Got it, that makes complete sense to me. Thanks for the detailed explanation and fix.

deepspeed/runtime/superoffload/superoffload_utils.py
```diff
@@ -82,7 +83,8 @@ def superoffload_optimizer_worker(param_queue: mp.SimpleQueue, result_queue: mp.
         return
 
     # Pre-allocate reusable pinned memory buffer for gradients
-    pinned_grad_buffer = torch.empty(max_grad_numel, dtype=torch.float32, device='cpu', pin_memory=True)
+    pinned_grad_buffer = get_accelerator().pin_memory(torch.empty(max_grad_numel, dtype=torch.float32, device='cpu'),
```
tohtana: Do we want to disable pinning based on `offload_optimizer.pin_memory`?  We might not, but it would be good to raise an error if `pin_memory` is `False`.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2944,7 +2944,8 @@ def needs_offload(target):
...
```
tohtana, line 2956: When we use `engine.offload_states(pin_memory=True)` and reach `engine.destroy()` without calling `reload_states()`, `engine.destroy()` doesn't release the pinned buffers (`hp_params_pin_buffers`, `lp_params_pin_buffers`, and `self.optimizer.state[*][key]`).
This is because we have the condition in `destroy`:

```python
        if not (self.cpu_offload and self.cpu_offload_pin_memory):
            return
```

I think this is a preexisting issue but it is now related to NativePinnedMemory manager as well. Can we resolve this issue in this PR?

accelerator/abstract_accelerator.py
```diff
@@ -272,6 +272,9 @@ def LongTensor(self):
     def _torch_pin_memory(self, tensor):
         return tensor.pin_memory()
 
+    def _torch_empty_pinned(self, tensor, shape):
+        return tensor.new_empty(shape, pin_memory=True)
```
tohtana, line 276: Strangely, PyTorch 2.13 returns a "mps:0" tensor on an MPS environment even the argument `tensor`'s device is cpu when `pin_memory=True` is given. You can reproduce it with 

```python
import torch
x = torch.empty((4, 8), device="cpu")
print(x.device, x.new_empty(x.shape).device, x.new_empty(x.shape, pin_memory=True).device)
```

Shouldn't we need to override `_torch_empty_pinned()` in MPS accelerator to make sure it returns a cpu tensor?

tohtana (changes requested):
Hi @sfc-gh-truwase, I left a few comments.
These issues are not very critical, but it would be great if we could fix the memory leak one in this PR (though it is a preexisting issue).

deepspeed/runtime/superoffload/superoffload_utils.py
sfc-gh-truwase: Good question. 

@xylian86 what do you think? Should we condition memory pinning on `offload_optimizer.pin_memory` or not?

accelerator/abstract_accelerator.py
sfc-gh-truwase: Yes, I am able to repro this behavior. This looks like a bug of vanilla torch and MPS. Can you please clarify what you want us to do in this PR since it seems there are other issues with torch+MPS

```bash
>>> x.pin_memory()
  File "<python-input-11>", line 1, in <module>
    x.pin_memory()
    ~~~~~~~~~~~~^^
RuntimeError: Attempted to set the storage of a tensor on device "cpu" to a storage on different device "mps:0".  This is no longer allowed; the devices must match.
>>> x.new_empty(x.shape).pin_memory()
  File "<python-input-14>", line 1, in <module>
    x.new_empty(x.shape).pin_memory()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
RuntimeError: Attempted to set the storage of a tensor on device "cpu" to a storage on different device "mps:0".  This is no longer allowed; the devices must match.
>>>
```

tohtana: What about overriding `_torch_empty_pinned` with `pin_memory=False` and leave a comment explaining why we do it?

deepspeed/runtime/zero/stage_1_and_2.py
sfc-gh-truwase: Fixed in `05ddc1a41`. You were right: `_unpin_offload_buffers` returned immediately unless `cpu_offload and cpu_offload_pin_memory`, so `offload_states(pin_memory=True)` then `destroy()` without `reload_states()` leaked the host buffers.

`destroy()` now unpins `hp_params_pin_buffers`, `lp_params_pin_buffers`, and any CPU tensors in `optimizer.state` that `is_pinned` *before* that config check. ZeRO-3 already released its contiguous pin buffers unconditionally; it now uses the same helper for the leftover `offload_adam_states` aliases.

`test_native_destroy_frees_offload_states_pins` covers stages 2 and 3 with no ZeRO CPU offload configured. H200 `job-20260824T224008Z`:

| Run | Result |
|-----|--------|
| `test_destroy_unpin.py` with the fix | **7 passed** (both new stage-2 and stage-3 cases) |
| same test, sources reverted to `598ad17a` | **fails** `offloaded=4 after=4` on both stages |
| accelerator + pin_memory + offload-route UTs | **36 passed** |

accelerator/abstract_accelerator.py
sfc-gh-truwase: Good catch. That is broken vanilla-torch behavior on MPS (`new_empty(..., pin_memory=True)` can return `mps:0` even when the template is on CPU), so I am not changing the MPS accelerator in this PR.

I will take the MPS `_torch_empty_pinned` override in a separate PR.

tohtana: Sounds good

deepspeed/runtime/superoffload/superoffload_utils.py
xylian86: @sfc-gh-truwase  Since pinning appears to be a performance optimization rather than a correctness requirement. My preference is to honor `offload_optimizer.pin_memory`. When it is true, we should keep the current pinned allocation; when it is false, the worker should allocate a regular CPU buffer.

sfc-gh-truwase: Done in `658dc247`. SuperOffload now honors `offload_optimizer.pin_memory`:

- `true` (default): same pinned scratch buffer as before, via `get_accelerator().pin_memory(..., make_copy=False)`
- `false`: regular `torch.empty(..., device="cpu")`

The flag is passed from `self.offload_optimizer_pin_memory` into `SuperOffloadCPUOptimizer` and the worker. On worker shutdown we `unpin_memory` only when we actually pinned.

Tests: `test_superoffload_grad_buffer_unpinned_when_disabled` and `test_superoffload_grad_buffer_pinned_when_enabled` (native) both passed locally.

xylian86: Thanks!

sfc-gh-truwase:
@tohtana are there still pending issues for me to address?

tohtana:
Hi @sfc-gh-truwase,

Sorry for the delay. I just found that `destroy` can fail when it is called soon after `offload_states(non_blocking=True)` (=offloading is not finished). I don't think this is a major use case, but we can avoid it just by [adding synchronization](https://github.com/tohtana/DeepSpeed/compare/658dc24724ee41de74c5fb646042ec4b5a123da1...tohtana/pr8256-async-destroy-order-repair). Can we add it?

sfc-gh-truwase:
@tohtana great catch. Thanks for sharing a fix. Now fixed.

tohtana (approved):
Thank you Tunji, looks good to me!

PR #8259, opened 2026-08-16
Fix Triton NFS detection crash when df wraps long device names

### LINKED ISSUE #8249: IndexError in is_nfs_path() when df wraps a long device name (breaks `import deepspeed`)
## Describe the bug

`import deepspeed` raises `IndexError` and kills the process when the Triton cache directory sits on a filesystem whose **device name is longer than ~20 characters** and `df` is BusyBox rather than coreutils.

`is_nfs_path()` in `deepspeed/ops/transformer/inference/triton/matmul_ext.py` runs `df -T <path>` and reads the filesystem type from the second column of the second line:

```python
output = subprocess.check_output(['df', '-T', path], encoding='utf-8')   # line 30
lines = output.strip().split('\n')
if len(lines) > 1:                          # line 36 - lines[1] is guarded
    fs_type = lines[1].split()[1].lower()   # line 37 - .split()[1] is NOT
```

BusyBox `df` wraps a long device name onto its own line, so `lines[1]` contains only the device, `.split()` returns a single element, and `[1]` raises.

This runs at **module import scope** (`fp16_matmul = Fp16Matmul()` → `_read_autotune_table()` → `AutotuneCacheManager()` → `TritonCacheDir.warn_if_nfs()`), so the whole process dies — and the call only exists to print an informational NFS warning.

## To Reproduce

Any environment with BusyBox `df` (common in minimal/distroless container images) and a device name over ~20 characters. This is routine on AWS, where the EBS device name is derived from the volume ID:

```
$ df -T /mount
Filesystem           Type       1K-blocks      Used Available Use% Mounted on
/dev/dvol0123456789abcdef0
                     ext4       2112647088   3439616 2109191088   0% /mount
```

Second line has **1 field** → `IndexError`.

```
File ".../deepspeed/ops/transformer/inference/triton/matmul_ext.py", line 37, in is_nfs_path
    fs_type = lines[1].split()[1].lower()  # File system type is the second column
IndexError: list index out of range
```

## Expected behavior

A failure to determine the filesystem type should not be fatal — worst case the NFS warning is skipped. It certainly shouldn't be able to prevent `import deepspeed`.

## Proposed fix

Pass POSIX `-P`, which guarantees one line per filesystem and suppresses the wrap:

```diff
-        output = subprocess.check_output(['df', '-T', path], encoding='utf-8')
+        output = subprocess.check_output(['df', '-PT', path], encoding='utf-8')
```

Verified on an affected host (BusyBox df):

```
df -T  /mount  -> second line has 1 field  (device wrapped)
df -PT /mount  -> second line has 7 fields (single line)
```

BusyBox honours both `-P` and `-T`, as does coreutils.

It may also be worth wrapping the parse defensively, so any unexpected `df` output degrades to "not NFS" instead of raising:

```python
fields = lines[1].split()
fs_type = fields[1].lower() if len(fields) > 1 else ''
return 'nfs' in fs_type
```

Happy to open a PR with either or both.

## System info

- DeepSpeed: 0.17.5 (parse is unchanged through 0.19.5)
- OS: container image with BusyBox coreutils
- Python: 3.10

atirna:
## Description

`is_nfs_path()` reads the fs type from the second column of the second `df` output row. With BusyBox `df` and a device name longer than ~20 characters (common on AWS, where the EBS device is derived from the volume id), the device wraps onto its own line, the row has a single field, and the `[1]` lookup raises `IndexError`.

Since the check runs while printing an informational NFS warning during `import deepspeed`, that exception kills the whole import in BusyBox-based containers.

I changed the call to `df -PT` so POSIX output keeps every filesystem on one line (BusyBox and coreutils both honour `-P` with `-T`), and made the parse treat a short or malformed row as not-NFS instead of raising, so any other unexpected `df` output just skips the warning.

## Testing

- `pytest -q tests/unit/ops/transformer/inference/test_matmul_ext.py` — new mocked regression test covering the wrapped-device case passes, and fails with IndexError on the old parsing
- `pre-commit run --files deepspeed/ops/transformer/inference/triton/matmul_ext.py tests/unit/ops/transformer/inference/test_matmul_ext.py` — all hooks pass
- mocked POSIX `nfs4` row is still detected as NFS, so real NFS warnings keep working

Fixes #8249

### DIFF
--- a/deepspeed/ops/transformer/inference/triton/matmul_ext.py
+++ b/deepspeed/ops/transformer/inference/triton/matmul_ext.py
@@ -33,17 +33,18 @@ def is_nfs_path(path):
             break
         path = parent
 
-    # Use the 'df' command to find the file system type for the given path
+    # POSIX output keeps long device names from wrapping onto a separate line.
     try:
-        output = subprocess.check_output(['df', '-T', path], encoding='utf-8', stderr=subprocess.DEVNULL)
+        output = subprocess.check_output(['df', '-PT', path], encoding='utf-8', stderr=subprocess.DEVNULL)
     except (subprocess.CalledProcessError, FileNotFoundError):
         return False  # Command failed or 'df' not available
 
-    # Process the output of 'df -T' to check for 'nfs' in the filesystem type column
+    # Process the output of 'df -PT' to check for 'nfs' in the filesystem type column.
     lines = output.strip().split('\n')
     if len(lines) > 1:  # The first line is headers
-        fs_type = lines[1].split()[1].lower()  # File system type is the second column
-        return 'nfs' in fs_type
+        fields = lines[1].split()
+        if len(fields) > 1:
+            return 'nfs' in fields[1].lower()
     return False
 
 
--- a/tests/unit/ops/transformer/inference/test_matmul_ext.py
+++ b/tests/unit/ops/transformer/inference/test_matmul_ext.py
@@ -1,0 +1,36 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+
+import ast
+import os
+import subprocess
+from pathlib import Path
+from typing import Callable, cast
+from unittest.mock import patch
+
+MATMUL_EXT_PATH = Path(
+    __file__).resolve().parents[5] / "deepspeed" / "ops" / "transformer" / "inference" / "triton" / "matmul_ext.py"
+
+
+def load_is_nfs_path() -> Callable[[Path], bool]:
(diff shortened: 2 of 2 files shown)


pengdurice:
lgtm, @tohtana , do you mind having a second eye on it? thank you!

sfc-gh-truwase (approved):

pengdurice:
@atirna, please take a look at the ci failure.

atirna:
it looks like the runner got oom-killed rather than a real test failure.

i pushed 6f724818 anyway: branch rebased onto current master (it was 5 behind) and the regression test now imports is_nfs_path directly with a positive nfs-detection.

PR #8264, opened 2026-08-17
Fix ZeRO-3 synchronization during OPSD rollout

### LINKED ISSUE #8262: ZeRO-3 OPSD rollout deadlocks when data-parallel ranks finish generation at different lengths
## Summary

OPSD rollout can deadlock when Hugging Face `generate()` runs on a ZeRO-3-partitioned student and data-parallel ranks finish at different token positions.

One rank may emit EOS and leave generation while other ranks enter the next decode forward. The latter ranks then wait in the next ZeRO-3 parameter all-gather, while the completed rank has already entered teacher forward and waits in a different collective.

This differs from #7844: that report had all ranks hanging in `engine.step()`. Here the ranks diverge before backward/step, during student rollout and teacher scoring.

## Environment

Reproduced independently on:

- 8 x NVIDIA H200, CUDA 12.6, PyTorch 2.7.0, DeepSpeed `cf443004`
- 8 x AMD MI250 ROCm devices, ROCm 6.4.1, PyTorch 2.9.1, DeepSpeed `22969fa7`
- DeepSpeedExamples OPSD trainer with ZeRO stage 3
- Qwen2.5-0.5B-Instruct student
- Qwen2.5-Math-7B-Instruct teacher
- Different prompts on each data-parallel rank

## Steps to reproduce

1. Use the DeepSpeedExamples OPSD Hybrid Engine configuration with ZeRO stage 3.
2. Run with multiple data-parallel ranks and different prompts per rank.
3. Let `module.generate()` use the tokenizer's default EOS stopping criterion.
4. Run at least two rollout/training steps.

## Observed behavior

After a completed training step, the job permanently stops making progress. Per-rank stack traces show:

- one rank in teacher forward: `reset_step -> broadcast`;
- the other ranks in student decode: `fetch_sub_module -> all_gather_into_tensor`.

The divergent collective sequence is:

```text
rank 1:   student decode N -> EOS -> teacher forward broadcast
others:   student decode N -> student decode N+1 parameter all-gather
```

Padding generated sequences afterward cannot help because the mismatch occurs inside autoregressive generation.

## Root cause

Under ZeRO-3, each decode forward gathers partitioned parameters. Data-parallel ranks must execute the same number and order of forwards. Default per-sequence EOS termination violates that requirement when prompts produce different response lengths.

## Verified workaround

Force every rank to execute exactly `max_new_tokens` decode iterations:

```python
output_ids = module.generate(
    ...,
    eos_token_id=None,
)
```

The two-step eight-GPU reproducer completed on both H200 and MI250 after this change. A full 8-device MI250 prompt epoch and a subsequent 200-step run also completed.

## Expected behavior

The Hybrid Engine rollout should preserve an identical ZeRO-3 collective sequence across data-parallel ranks. At minimum, the HF generation path should disable rank-local early termination. A future optimized implementation could mask finished sequences while keeping all ranks in the decode loop.

LiRunGuo:
## Summary

- disable rank-local EOS early exit in the OPSD Hybrid Engine Hugging Face rollout path
- keep ZeRO-3 decode forwards and parameter collectives in the same order on every data-parallel rank
- add a CPU unit test that verifies the collective-safe generation argument

Fixes #8262.

## Why

ZeRO-3 gathers partitioned parameters during every autoregressive decode forward. If one data-parallel rank emits EOS and returns from `generate()` while another rank continues decoding, the continuing rank enters another parameter all-gather while the finished rank moves to a later collective. The job then deadlocks.

Passing `eos_token_id=None` makes every rank execute exactly `max_new_tokens` iterations. Output masking remains unchanged, so EOS and padding tokens are still excluded from downstream response loss as applicable.

## Validation

- `pytest -q tests/unit/runtime/rollout/test_hybrid_engine_rollout.py` (`10 passed`)
- `pre-commit run --files deepspeed/runtime/rollout/hybrid_engine_rollout.py tests/unit/runtime/rollout/test_hybrid_engine_rollout.py`
- 2-step distributed reproducer completed on 8 x H200 and 8 x MI250 after the fix
- full prompt epoch and a separate 200-step OPSD run completed on 8 x MI250

### DIFF
--- a/deepspeed/runtime/rollout/hybrid_engine_rollout.py
+++ b/deepspeed/runtime/rollout/hybrid_engine_rollout.py
@@ -96,6 +96,9 @@ class HybridEngineRollout(RolloutEngine):
                     prompt_ids,
                     attention_mask=prompt_attn,
                     max_new_tokens=max_new_tokens,
+                    # ZeRO-3 gathers parameters during each decode forward, so every
+                    # data-parallel rank must execute the same number of iterations.
+                    eos_token_id=None,
                     do_sample=do_sample,
                     temperature=temperature if do_sample else 1.0,
                     top_p=sampling.top_p if do_sample else 1.0,
--- a/deepspeed/runtime/rollout/hybrid_engine_rollout.py
+++ b/deepspeed/runtime/rollout/hybrid_engine_rollout.py
@@ -57,6 +57,8 @@ class HybridEngineRollout(RolloutEngine):
         pad_token_id = self.tokenizer.pad_token_id
         if pad_token_id is None:
             pad_token_id = self.tokenizer.eos_token_id
+        if pad_token_id is None:
+            raise ValueError("The tokenizer must define pad_token_id or eos_token_id")
 
         module = self.engine.module
 
@@ -112,10 +114,22 @@ class HybridEngineRollout(RolloutEngine):
             accelerator.synchronize()
             generation_end = time.perf_counter()
 
+        # Generation deliberately ignores EOS above so ZeRO-3 ranks execute
+        # the same number of parameter-gather collectives. Restore the usual
+        # generation semantics before returning: retain the first EOS in each
+        # response and replace every later token with padding.
+        output_ids, response_attn = self._pad_after_eos(
+            output_ids,
+            response_start=prompt_len,
+            eos_token_id=self.tokenizer.eos_token_id,
+            pad_token_id=pad_token_id,
+        )
+
         # Build attention mask: pad positions (both left padding from prompt
         # and right padding from EOS / shorter sequences) are 0.
         response_start = prompt_len
         attention_mask = (output_ids != pad_token_id).long()
+        attention_mask[:, response_start:] = response_attn
         for i in range(total):
             prompt_valid = request.prompt_attention_mask[i // n if B > 1 else 0]
             attention_mask[i, :prompt_len] = prompt_valid
@@ -193,6 +207,29 @@ class HybridEngineRollout(RolloutEngine):
         pre_handle = module.register_forward_pre_hook(reduce_prompt_batch, with_kwargs=True)
         post_handle = module.register_forward_hook(expand_prompt_output, with_kwargs=True)
         return pre_handle, post_handle
+
+    @staticmethod
+    def _pad_after_eos(output_ids, response_start, eos_token_id, pad_token_id):
+        """Retain the first response EOS and pad every subsequent position."""
+        response_ids = output_ids[:, response_start:]
+        response_attn = (response_ids != pad_token_id)
+
+        if eos_token_id is None or response_ids.shape[1] == 0:
+            return output_ids, response_attn.long()
+
+        eos_ids = torch.as_tensor(eos_token_id, device=response_ids.device, dtype=response_ids.dtype).flatten()
+        is_eos = (response_ids.unsqueeze(-1) == eos_ids).any(dim=-1)
+        has_eos = is_eos.any(dim=-1)
+        first_eos_idx = is_eos.long().argmax(dim=-1)
+        positions = torch.arange(response_ids.shape[1], device=response_ids.device).unsqueeze(0)
+        after_first_eos = has_eos.unsqueeze(1) & (positions > first_eos_idx.unsqueeze(1))
+        first_eos = has_eos.unsqueeze(1) & (positions == first_eos_idx.unsqueeze(1))
+
+        output_ids = output_ids.clone()
+        output_ids[:, response_start:].masked_fill_(after_first_eos, pad_token_id)
+        # EOS is a valid generated token even when pad_token_id == eos_token_id.
+        response_attn = ((response_ids != pad_token_id) | first_eos) & ~after_first_eos
+        return output_ids, response_attn.long()
 
     # ------------------------------------------------------------------
     # Graph capture decode loop (greedy only)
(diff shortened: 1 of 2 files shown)


delock:
Hi @LiRunGuo thanks for your  issue and fix.  I have some comments:
1. The fix by setting eos_token_id=None will have the hybridenginerollout keep on generating tokens instead of padding beyond eos.  This will cause a change of behavior.  A proper fix would be apply the padding (detect first EOS token and pad the rest) to ensure output identical to before the fix.
2. It will help performance if all ranks can check whether an early stop condition (EOS had been generated on every batch) had been met and all ranks can stop rollout early (not necessary rollout to max length).  We do not need to address this in this PR, but it would help if you can share your batch information (how many prompts per training iteration, how many rollouts per prompt), these information will help us prioritize our optimizations on `HybridEngineRollout`, thanks!

@PKUWZP for this discussion.

LiRunGuo:
Thanks for pointing this out. I agree that setting eos_token_id=None is only a deadlock workaround and changes the originalgeneration behavior by disabling EOS termination.
  
I will revise the fix so that EOS detection and post-EOS padding are preserved. The rollout will keep all ranks in lockstep, retain the first EOS token, and pad all subsequent positions, matching the previous output semantics.

For the current OPSD configuration, the main 8-GPU run uses micro_batch_size_per_gpu=1, gradient_accumulation_steps=1, and n_samples_per_prompt=1. With 8 data-parallel ranks, this corresponds to 8 prompts and 8 rollouts per training iteration. The 2-GPU smoke test uses 2 prompts and 2 rollouts per iteration. The verification runs use a maximum response length of 64, while the full reproduction uses 1024.

I agree that synchronized early stopping across ranks would improve performance. I will treat that as a follow-up optimization after restoring the original EOS/padding behavior.

delock:
Hi @LiRunGuo , I saw you force-pushed after comments but `eos_token_id=None` stays the same.  A gentle remind if you intend to push your fix.

Thanks for your setup information, we definely should support such use case.

delock (approved):

delock:
Hi @LiRunGuo thanks for the followup.  Can you resolve merge conflicts?  Thanks!

I'll create an issue for early stopping and assign to you, let me know if it works.

LiRunGuo:
Thanks for the reminder. I have rebased the PR onto the latest master and resolved the conflicts with the newly added HybridEngineRollout profiling changes. The updated branch preserves both profiling and synchronized fixed-length decoding with post-EOS padding. I kept synchronized early stopping out of this PR and will address it separately in #8321.

Local validation after the rebase:
- `pytest -q tests/unit/runtime/rollout/test_hybrid_engine_rollout.py`: 18 passed
- targeted pre-commit hooks: passed
- `git diff --check`: passed

LiRunGuo:
The PR is now mergeable, approved, and all available checks/DCO have passed. I attempted to merge it, but this repository uses a maintainer-controlled merge queue and my account does not have permission to add it. Could you please add #8264 to the merge queue when convenient? Thanks!

LiRunGuo:
Hi @delock, I checked the failed merge-queue run ([32981748015](https://github.com/deepspeedai/DeepSpeed/actions/runs/32981748015)). The only failures were two unrelated FusedAdam numerical-reference cases:

- `test_fused_adam_matches_reference[fp32-adam]`
- `test_fused_adam_matches_reference[fp16-adam]`

All `HybridEngineRollout` tests passed in that run. The temporary merge commit only changes `deepspeed/runtime/rollout/hybrid_engine_rollout.py` and its unit-test file; it does not touch FusedAdam or `test_adamw.py`. The full CPU unit-test job also passed previously on the PR head.

This therefore appears to be an unrelated intermittent CI/test failure rather than a regression from #8264. Could you please rerun the failed merge-group job or add the PR back to the merge queue? Thanks!

LiRunGuo:
Hi @delock, thank you for requeueing the PR. I checked the second merge-group commit (`49aba2d3741bde3bc204a39bd08e8619fb815005`), and it appears to be stuck because of GitHub Actions startup/scheduling failures rather than a code or test failure:

- [cpu-torch-latest](https://github.com/deepspeedai/DeepSpeed/actions/runs/32985534079) has remained queued without creating any jobs or steps.
- [DCO / required](https://github.com/deepspeedai/DeepSpeed/actions/runs/32985538427) and [Formatting](https://github.com/deepspeedai/DeepSpeed/actions/runs/32985533870) ended with `startup_failure`.
- [Python](https://github.com/deepspeedai/DeepSpeed/actions/runs/32985460453) was marked failed before its queued jobs started and has no test output.
- NV precompile completed successfully.

Because required workflows for this merge group have already failed, the still-queued CPU workflow cannot make the merge complete. Could you please cancel/ignore this stale merge group and add #8264 to the merge queue again? No code changes appear necessary. Thanks!

LiRunGuo:
Hi @delock, I have rebased #8264 onto the latest master (`32e301ffa`) and resolved the overlap with the newly merged shared-prefill implementation from #8296.

The updated code preserves the complete shared-prefill hook lifecycle and places the ZeRO-3-safe `eos_token_id=None` argument inside the shared generation path. Post-EOS padding is applied after generation/profiling, and both the shared-prefill helper and EOS-padding helper/tests are retained.

Local validation after the rebase:
- `pytest -q tests/unit/runtime/rollout/test_hybrid_engine_rollout.py`: 23 passed, 1 CUDA-only test skipped
- targeted pre-commit hooks: passed
- `git diff --check`: passed
- range-diff reviewed: the two original fixes are preserved, with only the expected shared-prefill context changes

The branch has been updated with `--force-with-lease`. Once the refreshed CI passes, could you please add #8264 back to the merge queue? Thanks!

PR #8265, opened 2026-08-17
Fallback for unsupported Hybrid Engine policies

### LINKED ISSUE #8263: Hybrid Engine registers partial inference policies for unsupported Qwen architectures
## Summary

DeepSpeed Hybrid Engine initialization can create a partial inference-injection path for an unsupported architecture such as Qwen2.5. Although no complete transformer replacement policy matches Qwen, generic Linear/Embedding/LayerNorm policies are still registered and recursively applied.

With ZeRO-3 partitioned parameters this fails during inference-container construction instead of falling back to the model's native `generate()` method.

## Environment

- Qwen2.5-0.5B-Instruct
- DeepSpeed Hybrid Engine enabled
- ZeRO stage 3
- Reproduced on 8 x NVIDIA H200 with DeepSpeed `cf443004`
- Independently reproduced on 8 x AMD MI250 ROCm devices with DeepSpeed `22969fa7`

## Error

```text
DeepSpeedHybridEngine.create_inference_module
  -> create_inference_containers
  -> layers.py::_mark_uc_metadata
  -> original_weight_shape = (original_out_dim, self.weight.shape[1])
IndexError: tuple index out of range
```

Qwen includes one-dimensional weights such as RMSNorm weights. The partial auxiliary-policy path reaches code that assumes a two-dimensional weight.

## Root cause

`populate_all_inference_policies()` always adds generic policies for `nn.Linear`, `nn.Embedding`, and `nn.LayerNorm`, even when the model contains no transformer layer supported by a complete DeepSpeed replacement policy.

These generic wrappers are auxiliary components of a supported transformer injection policy; by themselves they do not provide a complete Hybrid Engine inference path. Registering them alone prevents the intended native-generation fallback and can process partitioned module weights with invalid dimensional assumptions.

## Verified workaround

After collecting complete transformer policies, first check whether any module in the model matches one. If none matches, leave the inference policy map empty and use the native Hugging Face `generate()` fallback.

This preserves existing behavior for supported architectures while allowing Qwen OPSD training to initialize and run. The fallback completed an 8-device MI250 prompt epoch and a 200-step training run.

## Expected behavior

If no complete transformer replacement policy supports the model, Hybrid Engine should not register only the auxiliary policies. It should safely use native generation or report a clear unsupported-architecture message rather than failing while constructing a partial inference path.

LiRunGuo:
## Summary

- register Hybrid Engine auxiliary Linear/Embedding/LayerNorm policies only when the model contains a supported complete transformer policy
- allow unsupported architectures such as Qwen2.5 to retain the native `generate()` fallback
- add CPU-only unit coverage for supported and unsupported policy registration

Fixes #8263.

## Why

The generic wrappers are auxiliary pieces of a complete transformer injection policy. Registering them when no transformer layer matches creates a partial inference path. With a ZeRO-3-partitioned Qwen model, container construction can then reach `_mark_uc_metadata()` with a one-dimensional normalization weight and fail on `weight.shape[1]`.

The change first builds the complete-policy map and checks it against the model's module classes. If no complete policy matches, the map remains empty, no partial containers are created, and Hybrid Engine leaves the model's native generation method intact. Supported model behavior is unchanged.

## Validation

- `pytest -q tests/unit/hybrid_engine/test_he_policy.py` (`2 passed`)
- `pre-commit run --files deepspeed/runtime/hybrid_engine.py tests/unit/hybrid_engine/test_he_policy.py`
- Qwen2.5-0.5B / Qwen2.5-Math-7B OPSD completed a full prompt epoch and a separate 200-step run on 8 x MI250 with the fallback

### DIFF
--- a/deepspeed/runtime/hybrid_engine.py
+++ b/deepspeed/runtime/hybrid_engine.py
@@ -134,6 +134,12 @@ class DeepSpeedHybridEngine(DeepSpeedEngine):
                     self.inference_policies.update({orig_layer_class: (self.new_inference_container, plcy)})
             elif plcy._orig_layer_class is not None:
                 self.inference_policies.update({plcy._orig_layer_class: (self.new_inference_container, plcy)})
+
+        has_transformer_policy = any(module.__class__ in self.inference_policies for module in self.module.modules())
+        if not has_transformer_policy:
+            self.inference_policies = {}
+            return
+
         self.inference_policies.update({
             nn.Linear: (LinearLayer, ),
             nn.Embedding: (EmbeddingLayer, ),
--- a/deepspeed/runtime/hybrid_engine.py
+++ b/deepspeed/runtime/hybrid_engine.py
@@ -137,6 +137,12 @@ class DeepSpeedHybridEngine(DeepSpeedEngine):
 
         has_transformer_policy = any(module.__class__ in self.inference_policies for module in self.module.modules())
         if not has_transformer_policy:
+            logger.warning(
+                "No compatible DeepSpeed inference policy found for model type %s. "
+                "Hybrid Engine inference acceleration is unavailable; rollout will "
+                "use the model's native generate() path.",
+                type(self.module).__name__,
+            )
             self.inference_policies = {}
             return
 
--- a/tests/unit/hybrid_engine/test_he_policy.py
+++ b/tests/unit/hybrid_engine/test_he_policy.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 # DeepSpeed Team
 
+from unittest.mock import patch
+
 import torch.nn as nn
 
 import deepspeed.runtime.hybrid_engine as hybrid_engine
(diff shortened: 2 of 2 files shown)


deepspeed/runtime/hybrid_engine.py
```diff
@@ -122,6 +122,12 @@ def populate_all_inference_policies(self):
                     self.inference_policies.update({orig_layer_class: (self.new_inference_container, plcy)})
             elif plcy._orig_layer_class is not None:
                 self.inference_policies.update({plcy._orig_layer_class: (self.new_inference_container, plcy)})
+
+        has_transformer_policy = any(module.__class__ in self.inference_policies for module in self.module.modules())
+        if not has_transformer_policy:
+            self.inference_policies = {}
```
delock, line 146: Can you add a `logger.warning` here for not matching `inference_policies`?  This would give user proper expectations.

delock (changes requested):

LiRunGuo:
Hi @delock, I have addressed the requested warning in the latest update. When no compatible transformer inference policy matches, Hybrid Engine now emits a `logger.warning` explaining that inference acceleration is unavailable and rollout will use the model native `generate()` path. I also rebased the PR onto the latest `master` (including #8264). Verification on the rebased branch: `tests/unit/hybrid_engine/test_he_policy.py` passes (2/2), `git diff --check` passes, and all applicable pre-commit hooks pass. The PR is mergeable; could you please re-review it and merge/add it to the merge queue if it looks good? Thanks!

delock (approved):

delock:
Hi @LiRunGuo there is a UT failure in CI, can you take a look?  Thanks!

LiRunGuo:
Hi @delock, thanks for pointing this out. I investigated the failed CPU UT. The warning itself was emitted correctly (it appeared under `Captured stdout`), but the test asserted against `caplog.text`, which was empty because DeepSpeed custom logger output was not propagated to pytest logging capture in that CI configuration. I pushed commit `cc2aebf10`, which directly mocks and verifies the `logger.warning` call, including the warning template and model type argument, instead of depending on logger-handler configuration. Validation: the targeted tests pass (2/2), the full `tests/unit/hybrid_engine` suite passes (2 passed, 4 hardware-dependent skipped), all applicable pre-commit checks pass, and `git diff --check` passes. Could you please rerun CI and add the updated PR to the merge queue again? Thanks!

LiRunGuo:
Hi @delock, the latest Modal CI failed with exit code 137 after progressing through roughly 55% of the tests, without any assertion failure. This appears to be a transient sandbox resource/OOM termination. Could you please rerun the failed CI check and add the PR back to the merge queue if it passes? Thanks!

PR #8266, opened 2026-08-17
Fix zero reduce bucket size validation

### LINKED ISSUE #8260: [BUG] reduce_bucket_size=0 passes validation but crashes ZeRO backward with ZeroDivisionError
**Describe the bug**

`zero_optimization.reduce_bucket_size=0` is accepted by DeepSpeed's configuration schema, but ZeRO training later crashes during backward with a `ZeroDivisionError`.

The schema currently allows zero:

```python
reduce_bucket_size: int = Field(pp_int(5e8), ge=0)
```

However, the optimizer later computes:

```python
percent_of_bucket_size = (100.0 * elem_count) // self.reduce_bucket_size
```

so `0` cannot actually be used at runtime.

I first observed this on DeepSpeed 0.19.1 while boundary-testing configuration values. The same minimal reproducer also fails on the v0.19.5 tag (`cf44300453eb0af79ed84ed8f1cb49d57478bd76`). The current `master` source (`c952d92a51c8d1ac3884f6aee01f668a0932a026`) still appears to have the same `ge=0` schema and division by `self.reduce_bucket_size`.

**To Reproduce**

Minimal reproducer:

```python
import os
import torch
import deepspeed

local_rank = int(os.environ["LOCAL_RANK"])
torch.cuda.set_device(local_rank)

model = torch.nn.Sequential(
    torch.nn.Linear(16, 16),
    torch.nn.GELU(),
    torch.nn.Linear(16, 1),
).to(dtype=torch.bfloat16)
optimizer = torch.optim.AdamW(model.parameters(), lr=1e-3)

config = {
    "zero_optimization": {
        "stage": 1,
        "reduce_bucket_size": 0,
        "allgather_bucket_size": 500000000,
    },
    "bf16": {"enabled": True},
}

engine, optimizer, _, _ = deepspeed.initialize(
    model=model,
    optimizer=optimizer,
    config=config,
)

x = torch.randn(1, 16, device=engine.device, dtype=torch.bfloat16)
loss = engine(x).float().square().mean()
engine.backward(loss)
engine.step()
```

Launch with two GPUs:

```bash
CUDA_VISIBLE_DEVICES=0,1 python -m torch.distributed.run \
  --nnodes=1 --nproc-per-node=2 \
  --master-addr=127.0.0.1 --master-port=29911 \
  repro.py
```

Observed failure:

```text
File "deepspeed/runtime/base_optimizer.py", line 388, in report_ipg_memory_usage
    percent_of_bucket_size = (100.0 * elem_count) // self.reduce_bucket_size
ZeroDivisionError: float floor division by zero
```

On 0.19.1, I repeated the boundary case three independent times and got the same `ZeroDivisionError` in all 3/3 runs. I then reproduced it again with the standalone script above on v0.19.5.

**Expected behavior**

Either:

1. `reduce_bucket_size=0` should be rejected during configuration validation (for example, require `gt=0`), or
2. zero should have a defined runtime meaning and all downstream code paths should handle it safely.

The current behavior accepts the configuration and fails only later during backward.

**Relevant source**

Current master still allows zero in `deepspeed/runtime/zero/config.py` and later divides by the value in `deepspeed/runtime/base_optimizer.py`.

- https://github.com/deepspeedai/DeepSpeed/blob/c952d92a51c8d1ac3884f6aee01f668a0932a026/deepspeed/runtime/zero/config.py#L113
- https://github.com/deepspeedai/DeepSpeed/blob/c952d92a51c8d1ac3884f6aee01f668a0932a026/deepspeed/runtime/base_optimizer.py#L388

**ds_report / environment**

Reproduced environment:

```text
DeepSpeed: 0.19.1, and separately v0.19.5 source tag
PyTorch: 2.13.0+cu126
CUDA runtime reported by PyTorch: 12.6
nvcc: 12.4
Python: 3.11.15
OS: Linux 5.15.0-84-generic x86_64, glibc 2.35
GPU: 2 x NVIDIA RTX A6000 for the reproducer
Launcher: torch.distributed.run
```

No Docker image was used.

**Additional context**

I searched the existing DeepSpeed issues for `reduce_bucket_size` together with `ZeroDivisionError`/zero and did not find an issue describing this exact validation/runtime mismatch.

fwerkor:
## Summary

Reject `zero_optimization.reduce_bucket_size=0` during configuration validation instead of allowing ZeRO to fail later with a division by zero during backward.

This changes the field constraint from non-negative to strictly positive and adds a boundary regression test covering `1` as valid and `0` as invalid.

Fixes #8260

## Testing

- `python -m pytest tests/unit/runtime/zero/test_zero_config.py -q` (7 passed)
- `pre-commit run --files deepspeed/runtime/zero/config.py tests/unit/runtime/zero/test_zero_config.py`

### DIFF
--- a/deepspeed/runtime/zero/config.py
+++ b/deepspeed/runtime/zero/config.py
@@ -110,7 +110,7 @@ class DeepSpeedZeroConfig(DeepSpeedConfigModel):
     Uses reduce or reduce scatter instead of allreduce to average gradients
     """
 
-    reduce_bucket_size: int = Field(pp_int(5e8), ge=0)
+    reduce_bucket_size: int = Field(pp_int(5e8), gt=0)
     """
     Number of elements reduced/allreduced at a time. Limits the memory required
     for the allgather for large model sizes
--- a/tests/unit/runtime/zero/test_zero_config.py
+++ b/tests/unit/runtime/zero/test_zero_config.py
@@ -3,7 +3,16 @@
 
 # DeepSpeed Team
 
+import pytest
+
 from deepspeed.runtime.zero.config import DeepSpeedZeroConfig, DeepSpeedZeroOffloadParamConfig, DeepSpeedZeroOffloadOptimizerConfig
+
+
+def test_zero_config_reduce_bucket_size():
+    assert DeepSpeedZeroConfig(reduce_bucket_size=1).reduce_bucket_size == 1
+
+    with pytest.raises(ValueError):
+        DeepSpeedZeroConfig(reduce_bucket_size=0)
 
 
 def test_zero_config_deprecatedfields():


fwerkor:
Hi @tohtana, gentle ping — would appreciate a review when you have a chance. Thanks!

sfc-gh-truwase (approved):

PR #8268, opened 2026-08-17
Make the WarmupCosineLR ratio flags reach the config

vineethsaivs:
`add_tuning_arguments` declares `--warmup_min_ratio` and `--cos_min_ratio` for
WarmupCosineLR, but `get_config_from_args` routed every schedule that is not
`LRRangeTest` or `OneCycle` through `override_warmupLR_params`. So both flags
were parsed and then dropped, and the config came back carrying the WarmupLR
parameters instead:

```
$ parser = add_tuning_arguments(argparse.ArgumentParser())
$ args = parser.parse_args(["--lr_schedule", "WarmupCosineLR",
                            "--warmup_min_ratio", "0.1", "--cos_min_ratio", "0.05"])
$ get_config_from_args(args)
({'type': 'WarmupCosineLR',
  'params': {'warmup_min_lr': 0, 'warmup_max_lr': 0.001,
             'warmup_num_steps': 1000, 'warmup_type': 'log'}}, None)
```

Both ratios are gone, and building the scheduler from what is left fails,
because `WarmupCosineLR` has no min/max lr at all:

```
TypeError: WarmupCosineLR.__init__() got an unexpected keyword argument 'warmup_min_lr'
```

`WARMUP_MIN_RATIO` and `COS_MIN_RATIO` were already defined next to the other
config keys, so only the plumbing was missing.

`get_lr_from_config` had the mirror-image problem: it falls through to
`lr_params[WARMUP_MAX_LR]` for anything that is not `LRRangeTest` or
`OneCycle`. For WarmupCosineLR that returned an unrelated WarmupLR default on
a CLI-built config, and raised `KeyError: 'warmup_max_lr'` on a config
actually written for this schedule. WarmupCosineLR scales each param group's
own lr by a ratio, so there is no learning rate in its params to read, and it
now says so instead.

The JSON config path was never affected: `engine._configure_lr_scheduler`
calls `scheduler(optimizer, **scheduler_params)` straight from the config
file, so `"scheduler": {"type": "WarmupCosineLR", "params": {"warmup_min_ratio": 0.1}}`
has always worked. Only the `add_tuning_arguments` CLI helper is fixed here.

Other schedules are untouched. `LRRangeTest`, `OneCycle`, `WarmupLR` and
`WarmupDecayLR` produce the same params and the same `get_lr_from_config`
answer as before, and `test_other_schedules_keep_their_config_params` pins
that.

**Not included**, so as to keep this to one thing: `TOTAL_NUM_STEPS` is
defined but unused, and neither `WarmupDecayLR` nor `WarmupCosineLR` can be
built from `get_config_from_args` output alone because both need
`total_num_steps`, which has no CLI flag. Callers pass it themselves today.
Happy to add that separately if you would like it plumbed.

## Tests

`tests/unit/runtime/test_lr_schedulers.py` gains three plain (non-distributed)
tests next to the existing ones:

- `test_warmup_cosine_lr_config_from_args_carries_the_ratios`
- `test_warmup_cosine_lr_has_no_lr_in_its_config_params`
- `test_other_schedules_keep_their_config_params`

DeepSpeed does not build on this machine (macOS, no CUDA), so I ran the three
new test bodies against the unmodified `lr_schedules.py` from master and
against the patched one, loading the module directly:

```
before: 2 failed, 1 passed
        (KeyError: 'warmup_max_lr', and the ratios missing from the config)
after:  3 passed
```

The one that passes on both sides is the no-regression control.

`yapf --style .style.yapf --diff` is clean on both changed files, and neither
adds a line over the 119 column limit.

### DIFF
--- a/deepspeed/runtime/lr_schedules.py
+++ b/deepspeed/runtime/lr_schedules.py
@@ -195,6 +195,23 @@ def override_warmupLR_params(args, params):
         params[WARMUP_TYPE] = args.warmup_type
 
 
+def override_warmupCosineLR_params(args, params):
+    # WarmupCosineLR scales each param group's own lr by a ratio, so it takes
+    # warmup_min_ratio/cos_min_ratio and does not accept warmup_min_lr or
+    # warmup_max_lr.
+    if hasattr(args, WARMUP_NUM_STEPS) and args.warmup_num_steps is not None:
+        params[WARMUP_NUM_STEPS] = args.warmup_num_steps
+
+    if hasattr(args, WARMUP_TYPE) and args.warmup_type is not None:
+        params[WARMUP_TYPE] = args.warmup_type
+
+    if hasattr(args, WARMUP_MIN_RATIO) and args.warmup_min_ratio is not None:
+        params[WARMUP_MIN_RATIO] = args.warmup_min_ratio
+
+    if hasattr(args, COS_MIN_RATIO) and args.cos_min_ratio is not None:
+        params[COS_MIN_RATIO] = args.cos_min_ratio
+
+
 def override_params(args, params):
     # LR range test params
     override_lr_range_test_params(args, params)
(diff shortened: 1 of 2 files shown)


sfc-gh-truwase (approved):

tohtana (approved):
Hi @vineethsaivs,
Thank you for the fix! Looks good to me.

PR #7542, opened 2025-09-03
avoid setting device_id to `init_process_group`

kaixuanliu:
In some usecases such as vllm, we need to new distributed group not only on gpu, but also on cpu, if we set `device_id` here, it will prevent us from new distributed group on cpu: [L230](https://github.com/vllm-project/vllm/blob/main/vllm/distributed/parallel_state.py#L230) . This PR fixes this bug.

PR #7266, opened 2025-04-30
set `device_id` in torch's `init_process_group`

stas00:
This PR overcomes this issue:
```
[W404 00:15:21.693690333 ProcessGroupNCCL.cpp:4561] [PG ID 0 PG GUID 0 Rank 0]  using GPU 0 to perform barrier as devices used by this process are currently unknown. This can potentially cause a hang if this rank to GPU mapping is incorrect. Specify device_ids in barrier() to force use of a particular device, or call init_process_group() with a device_id.
```
by setting `device_id` to the correct device corresponding to `LOCAL_RANK` env var.

PR #8269, opened 2026-08-17
fix: Only bind device id when needed,  Fixes #8248

### LINKED ISSUE #8248: [BUG]DeepSpeed latest version fails during NCCL process-group initialization with CUDA error 600, while 0.17.2 works
## Describe the bug

With the latest DeepSpeed version, launching a single-GPU training job fails during `torch.distributed.new_group()` with an NCCL `CUDA error 600: device not ready`.

The failure occurs during DeepSpeed-launched process initialization, before the model forward pass or any training computation.

Importantly, the same environment can successfully initialize both the default NCCL process group and an additional NCCL process group without DeepSpeed. Downgrading DeepSpeed to **0.17.2** makes the exact same training command work correctly.

This suggests a regression or compatibility issue in the latest DeepSpeed launcher/runtime related to NCCL process-group initialization.

### 1. Environment

The issue was reproduced with:

```text
PyTorch: 2.13.0+cu129
CUDA: 12.9
NCCL: 2.29.7
GPU: NVIDIA GeForce RTX 3060 Laptop GPU
GPU capability: 8.6
Python: 3.12
```

### 2. Launch a training job with the DeepSpeed launcher

```bash
deepspeed --module jllm.train_pipe \
    --model Qwen3.5-27B \
    --num_train_epochs 1 \
    --gradient_accumulation_steps 1 \
    --train_data s10_Qwen3.5-27B/ \
    --init \
    --partition_method '0,6'
```

DeepSpeed launches:

```text
world_size=1
local_rank=0
CUDA_VISIBLE_DEVICES=0
```

### 3. The job fails during model-parallel initialization

The relevant traceback is:

```text
torch.distributed.new_group()
    -> _new_process_group_helper()
    -> eager_backend.eager_connect_single_device()
    -> NCCL error
```

Error:

```text
torch.distributed.DistBackendError: NCCL error in:
.../NCCLUtils.cpp:243, unhandled cuda error

ncclUnhandledCudaError: Call to CUDA function failed.

Last error:
Cuda failure 600 'device not ready'
```

The failure occurs at:

```text
jllm/core/parallel_state.py
    create_group()
        -> torch.distributed.new_group()
```

before the model starts training.

## Expected behavior

The DeepSpeed launcher should successfully initialize the NCCL process group, as it does with DeepSpeed 0.17.2.

Since the job uses only one GPU (`world_size=1`), NCCL process-group initialization should not fail with `CUDA error 600`.

## ds_report output

`ds_report` output:

```text
[Please paste the output of `ds_report` here]
```

## Screenshots

Not applicable. The complete traceback is provided above.

## System info

* OS: Ubuntu on WSL2
* GPU count and types: 1 × NVIDIA GeForce RTX 3060 Laptop GPU
* GPU capability: 8.6
* Interconnects: N/A (single GPU)
* Python: 3.12
* PyTorch: 2.13.0+cu129
* CUDA runtime: 12.9
* NCCL: 2.29.7
* DeepSpeed: latest version where the issue occurs

## Launcher context

The experiment is launched using the **DeepSpeed launcher**:

```bash
deepspeed --module jllm.train_pipe ...
```

DeepSpeed reports:

```text
WORLD INFO DICT: {'localhost': [0]}
nnodes=1
num_local_procs=1
node_rank=0
dist_world_size=1
CUDA_VISIBLE_DEVICES=0
```

## Docker context

Not using Docker.

## Additional context

I verified that the underlying PyTorch/NCCL environment works correctly without DeepSpeed.

The following minimal test succeeds in the same Python environment:

```python
import torch

print("torch:", torch.__version__)
print("cuda:", torch.version.cuda)
print("nccl:", torch.cuda.nccl.version())
print("gpu:", torch.cuda.get_device_name(0))
print("capability:", torch.cuda.get_device_capability(0))

torch.cuda.set_device(0)

dist.init_process_group(
    backend="nccl",
    init_method="tcp://127.0.0.1:29501",
    rank=0,
    world_size=1,
)

print("default PG OK")

g = dist.new_group(
    ranks=[0],
    backend="nccl",
)

print("new_group OK")

dist.destroy_process_group()
```

Output:

```text
torch: 2.13.0+cu129
cuda: 12.9
nccl: (2, 29, 7)
gpu: NVIDIA GeForce RTX 3060 Laptop GPU
capability: (8, 6)
default PG OK
new_group OK
```

Therefore, the CUDA/NCCL stack itself appears to be functional.

### DeepSpeed version comparison

The exact same training environment and command work correctly after downgrading DeepSpeed to:

```text
DeepSpeed 0.17.2
```

So the issue appears to be specific to newer DeepSpeed versions.

In summary:

```text
DeepSpeed latest
    -> torch.distributed.new_group()
    -> NCCL eager_connect_single_device()
    -> CUDA error 600: device not ready
    -> FAIL

DeepSpeed 0.17.2
    -> same environment
    -> same training command
    -> SUCCESS
```

Could this be related to a change in the DeepSpeed launcher or distributed/NCCL initialization path in versions newer than 0.17.2?

pengdurice:
Fixes #8248.

### The change

`init_process_group` now asks a small resolver instead of inlining the condition:

```python
device_id = get_init_process_group_device_id(world_size)
if device_id is not None:
    kwargs.update(device_id=device_id)
```

`DEEPSPEED_SET_DEVICE_ID` overrides the decision, parsed by a new `get_env_flag` helper:

| Value | Effect |
|---|---|
| `1` / `true` / `yes` / `on` | always bind |
| `0` / `false` / `no` / `off` | never bind |
| unset | the default above: bind iff `world_size > 1` |
| anything else | warn and ignore, so the default still decides |

Multi-rank behaviour is unchanged.

### Tests

`tests/unit/comm/test_dist.py`

### DIFF
--- a/deepspeed/comm/comm.py
+++ b/deepspeed/comm/comm.py
@@ -811,11 +811,6 @@ def init_distributed(dist_backend: Optional[str] = None,
         config: Optional (DeepSpeedConfig). DeepSpeed configuration for setting up comms options (e.g. Comms profiling)
         rank: Optional (int). The current manually specified rank. Some init_method like "tcp://" need the rank and world_size as well (see: https://pytorch.org/docs/stable/distributed.html#tcp-initialization)
         world_size: Optional (int). Desired world_size for the TCP or Shared file-system initialization.
-
-    On `cuda` with a multi-rank job, the local device is bound to the process group so that torch
-    knows which GPU this rank owns. That also switches torch to eager communicator init, which some
-    platforms cannot complete. Set the environment variable `DEEPSPEED_SET_DEVICE_ID` to 0 to never
-    bind a device, or to 1 to bind it even for a single-rank job.
     '''
     global cdb
 
--- a/deepspeed/comm/comm.py
+++ b/deepspeed/comm/comm.py
@@ -811,6 +811,11 @@ def init_distributed(dist_backend: Optional[str] = None,
         config: Optional (DeepSpeedConfig). DeepSpeed configuration for setting up comms options (e.g. Comms profiling)
         rank: Optional (int). The current manually specified rank. Some init_method like "tcp://" need the rank and world_size as well (see: https://pytorch.org/docs/stable/distributed.html#tcp-initialization)
         world_size: Optional (int). Desired world_size for the TCP or Shared file-system initialization.
+
+    On `cuda` with a multi-rank job, the local device is bound to the process group so that torch
+    knows which GPU this rank owns. That also switches torch to eager communicator init, which some
+    platforms cannot complete. Set the environment variable `DEEPSPEED_SET_DEVICE_ID` to 0 to never
+    bind a device, or to 1 to bind it even for a single-rank job.
     '''
     global cdb
 
--- a/deepspeed/comm/torch.py
+++ b/deepspeed/comm/torch.py
@@ -15,8 +15,6 @@ from ..runtime import compiler
 from deepspeed.utils.torch import required_torch_version
 import os
 
-DS_SET_DEVICE_ID = 'DEEPSPEED_SET_DEVICE_ID'
-
 DS_COMM_ALL_GATHER_OFF = False
 DS_COMM_REDUCE_SCATTER_OFF = False
 DS_COMM_BROADCAST_OFF = False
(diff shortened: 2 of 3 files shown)


sfc-gh-truwase:
@pengdurice thanks for the PR. I don't fully understand the problem. It appears that this a recent regression. Did you identify regression?

ebarkhordar (commented):
@sfc-gh-truwase on the regression question: yes, and the boundary is the one the reporter names. I traced this in git on a full clone rather than reproducing it, since I have no CUDA machine here.

The `device_id` binding entered `deepspeed/comm/torch.py` in ee286e53 (#7266), and v0.17.2 is the last release without it:

```
$ git rev-parse --is-shallow-repository
false
$ git log -G 'kwargs\.update\(device_id=' --oneline --follow -- deepspeed/comm/torch.py
ee286e53 set `device_id` in torch's `init_process_group` (#7266)
$ git tag --contains ee286e53 | sort -V | head -3
v0.17.3
v0.17.4
v0.17.5
$ git log --oneline v0.17.2..v0.17.3 -- deepspeed/comm/
ee286e53 set `device_id` in torch's `init_process_group` (#7266)
f485e136 Improvements to Communication Logger (#7404)
```

#8248 reports 0.17.2 working and later versions failing, so the two agree. The only other comm change in that window is the logger one.

Two things in that history bear on the fix. First, #7266's stated purpose was narrow: it set `device_id` to silence the `ProcessGroupNCCL` warning that "devices used by this process are currently unknown" during `barrier()`. A single-rank job has no rank-to-GPU mapping to get wrong, which is the case this PR stops binding for.

Second, this block has already been narrowed once for the same shape of reason. #7542 (08879a39, first released in v0.17.6) added the `get_accelerator().device_name() == 'cuda'` condition because binding a device blocked creating CPU process groups afterwards. So `get_init_process_group_device_id` is the second carve-out for a side effect of binding rather than a new kind of exception, and that seems worth saying in the PR description.

What I did not check: I could not run the NCCL path, so I cannot confirm that leaving the device unbound is sufficient to clear CUDA error 600 on the reporter's setup. The git evidence only establishes that the binding is what changed between the two versions they compared.

tohtana:
Do we actually know why this happens only with `world_size=1`? It seems like an issue with PyTorch side.
I hesitate a bit to add a special case without knowing the real cause.

#7266 originally tried to remove this warning and chose the second approach ("call init_process_group() with a device_id). Does anyone think using `barrier` with device_ids works? (Any thought, @stas00?) without causing the issue in #8248?

```
[W404 00:15:21.693690333 ProcessGroupNCCL.cpp:4561] [PG ID 0 PG GUID 0 Rank 0]  using GPU 0 
to perform barrier as devices used by this process are currently unknown. This can
 potentially cause a hang if this rank to GPU mapping is incorrect. Specify device_ids in
 barrier() to force use of a particular device, or call init_process_group() with a device_id.
 ```
  
While investigating this, I also found DeepSpeed ignores `device_ids` passed to `barrier`. It should be fixed regardless of the approach in this PR. I opened #8312 to address it.

**Comment by stas00** [comment]
I have never tried barrier with device id, @tohtana 

question, why would anybody try to init dist on a single gpu. I suppose to make some code work on 1 or many gpus, but nccl won't really be used on a single gpu, correct? if I remember correctly this is just a no-op on nccl level.

if the culprit is the init of nccl on a single gpu, why not solve this in a much simpler way - just add `if world_size>1`

```
    def init_process_group(self, backend, timeout, init_method, rank, world_size):
        if not torch.distributed.is_initialized():
            kwargs = dict(
                timeout=timeout,
                init_method=init_method,
                rank=rank,
                world_size=world_size,
            )

            # 1. device_id arg was added in torch==2.3
            # 2. setting device_id leads to hanging in 2.6.0<torch<2.7.1 https://github.com/pytorch/pytorch/issues/153960
            # 3. do not set device_id unless more than 1 gpu
            if world_size>1 and 'device_id' in inspect.signature(torch.distributed.init_process_group).parameters and not (
                    version.parse("2.6.0") < version.parse(torch.__version__) < version.parse("2.7.1")):
                local_rank = int(os.environ.get('LOCAL_RANK', 0))
                kwargs.update(device_id=get_accelerator().device(local_rank))
            torch.distributed.init_process_group(backend, **kwargs)
```

**Comment by pengdurice** [comment]
https://github.com/deepspeedai/DeepSpeed/issues/8248#issuecomment-5429745690 
Sorry for the delay, I did some preliminary investigations.
Hi @sfc-gh-truwase , @stas00  and @tohtana, thank you for reviewing this PR, I added some investigation to the link above and am still investigating  it and waiting on the reporter's reply. 
@stas00  to your questions: my understanding is that the communication is nearly no-op (e.g all-reduce), but the init is not. CMIIW. The DS_SET_DEVICE_ID maybe an overkill, removing it, it reduces to what you suggests.
@tohtana , to your question  about if this happens only on world-size = 1, the customer reports it works fine on world size = 2. I am waiting for more information and confirmation from them. Thank you!

**Comment by stas00** [comment]
yeah, I propose to keep things simple, your code is valid but I think a simple world>1 check would have a much smaller footprint and it's a way easier to track/understand.

**Comment by tohtana** [comment]
Thank you @stas00 for sharing your thoughts. Then we can take the simplest approach.
One corner case we need to handle is having a world size in the init URL:

```
init_method="tcp://host:port?rank=0&world_size=2"
rank=-1
world_size=-1
```

In this case, `world_size` won't reflect the real size. We should skip the nccl init only when we are sure about the world size (`!=-1`).

> @tohtana , to your question about if this happens only on world-size = 1, the customer reports it works fine on world size = 2. I am waiting for more information and confirmation from them. Thank you!

@pengdurice If they don't reply, let's move forward and merge this. I think this PR is already useful.

stas00:
before merging that `get_env_flag` shouldn't be there, surely DS already has that somewhere in utils.

no need to pollute the code.

pengdurice:
The reporter mentioned that the PR fixes their problem and I fixed according to the comments above. Feel free to take a look again. @tohtana @stas00 @sfc-gh-truwase. Or we can wait for the reporter to give more detailed information to nail down the root cause;-)

sfc-gh-truwase (approved):

pengdurice:
image

PR #8274, opened 2026-08-18
Reject invalid ZenFlow ratio and update interval boundaries

tandede:
## Summary

- require `zenflow.topk_ratio` to be strictly between 0 and 1
- require numeric `zenflow.update_interval` values to be at least 1
- add regression coverage for the invalid ratio endpoints and non-positive update intervals

## Why

ZenFlow's automatic update path normalizes selected gradients by `topk_ratio` and unselected gradients by `1 - topk_ratio`. The explicit update path selects its overlap buffer using `micro_step // update_interval`.

The previous validation accepted ratio endpoints as well as zero and negative numeric update intervals. These values could therefore trigger division by zero or invalid scheduling after training had already initialized. Rejecting them during configuration parsing provides an immediate validation error while preserving all usable configurations.

## Testing

- `python -m pytest -q tests/unit/runtime/zero/test_zero_config.py tests/unit/runtime/zenflow/test_zf_config.py tests/unit/runtime/zenflow/test_zf.py::test_split_affinity` — 21 passed
- `pre-commit run --files deepspeed/runtime/zenflow/zenflow_config.py tests/unit/runtime/zenflow/test_zf_config.py`

### DIFF
--- a/tests/unit/runtime/zenflow/test_zf_config.py
+++ b/tests/unit/runtime/zenflow/test_zf_config.py
@@ -59,6 +59,12 @@ def test_invalid_zenflow_type_raises():
         DeepSpeedZeroConfig(zenflow=123)
 
 
+@pytest.mark.parametrize("topk_ratio", [0.0, 1.0])
+def test_topk_ratio_excludes_endpoints(topk_ratio):
+    with pytest.raises(ValidationError):
+        ZenFlowConfig(topk_ratio=topk_ratio)
+
+
 def test_offload_and_zenflow_combined():
     """
     offload_optimizer and zenflow can be used together under stage 2
--- a/tests/unit/runtime/zenflow/test_zf_config.py
+++ b/tests/unit/runtime/zenflow/test_zf_config.py
@@ -65,6 +65,12 @@ def test_topk_ratio_excludes_endpoints(topk_ratio):
         ZenFlowConfig(topk_ratio=topk_ratio)
 
 
+@pytest.mark.parametrize("update_interval", [0, -5])
+def test_update_interval_rejects_non_positive_values(update_interval):
+    with pytest.raises(ValidationError):
+        ZenFlowConfig(update_interval=update_interval)
+
+
 def test_offload_and_zenflow_combined():
     """
     offload_optimizer and zenflow can be used together under stage 2
--- a/tests/unit/runtime/zenflow/test_zf_config.py
+++ b/tests/unit/runtime/zenflow/test_zf_config.py
@@ -71,6 +71,13 @@ def test_update_interval_rejects_non_positive_values(update_interval):
         ZenFlowConfig(update_interval=update_interval)
 
 
+@pytest.mark.parametrize("update_interval", [1, "auto"])
+def test_update_interval_accepts_valid_boundary_values(update_interval):
+    config = ZenFlowConfig(update_interval=update_interval)
+
+    assert config.update_interval == update_interval
+
+
 def test_offload_and_zenflow_combined():
     """
     offload_optimizer and zenflow can be used together under stage 2
--- a/deepspeed/runtime/zenflow/zenflow_config.py
+++ b/deepspeed/runtime/zenflow/zenflow_config.py
@@ -60,6 +60,9 @@ class ZenFlowConfig(DeepSpeedConfigModel):
         if isinstance(self.update_interval, str) and self.update_interval != "auto":
             raise ValueError('If update_interval is a string, it must be "auto"')
 
+        if isinstance(self.update_interval, int) and self.update_interval < 1:
+            raise ValueError("If update_interval is a number, it must be at least 1")
+
         if not isinstance(self.full_warm_up_rounds, int):
             raise ValueError('full_warm_up_rounds must be an integer')
 
--- a/deepspeed/runtime/zenflow/zenflow_config.py
+++ b/deepspeed/runtime/zenflow/zenflow_config.py
@@ -12,8 +12,8 @@ from deepspeed.runtime.config_utils import DeepSpeedConfigModel
 class ZenFlowConfig(DeepSpeedConfigModel):
     """Configuration options for ZenFlow optimization module."""
 
-    topk_ratio: float = Field(0.1, ge=0.0, le=1.0)
-    """Ratio of top-k important gradient columns to retain (range: 0.0 to 1.0)."""
+    topk_ratio: float = Field(0.1, gt=0.0, lt=1.0)
+    """Ratio of top-k important gradient columns to retain (must be greater than 0.0 and less than 1.0)."""
 
     select_strategy: str = "auto"
     """Strategy for selecting important gradient indices.
--- a/deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
+++ b/deepspeed/runtime/zenflow/zenflow_stage_1_and_2.py
@@ -44,6 +44,12 @@ SELECTIVE_OPTIMIZER_TIMERS = [
 ]
 
 
+def _num_selected_columns(num_columns, topk_ratio):
+    if num_columns == 0:
+        return 0
+    return max(1, int(num_columns * topk_ratio))
+
+
 class ZenFlowZeroOptimizer(DeepSpeedZeroOptimizer):
 
     def __init__(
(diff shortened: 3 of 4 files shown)


sfc-gh-truwase:
@Antlera can you please help review?

ebarkhordar (commented):
I ran the config at head `09641d5` in a clean `python:3.11-slim` container. Both endpoints are rejected now. The other divisor in that same expression still has no bound, and it is in this config class:

```
topk_ratio=0.0: rejected (ValidationError)
topk_ratio=1.0: rejected (ValidationError)
update_interval=0: ACCEPTED, value=0
update_interval=-5: ACCEPTED, value=-5
```

An explicit integer sets `auto_update = False` (`zenflow_stage_1_and_2.py:129-130`), and that is the branch which divides: `(self.micro_step // self.update_interval) & 1`, at `zenflow_stage_1_and_2.py:658` and `engine_stage3.py:451`. `full_warm_up_rounds` defaults to 0, so `get_overlap_step_state` reaches it on the first optimizer step. The `topk_ratio` divisions you fixed sit inside `if self.auto_update:`, so yours is the auto path and this is the explicit one.

`Field` bounds are awkward on a `Union[str, int]`, but `validate_fields` already checks this field and one clause covers it:

```python
if isinstance(self.update_interval, int) and self.update_interval < 1:
    raise ValueError('If update_interval is a number, it must be at least 1')
```

I only constructed configs. No GPU on this box, so I did not run a training step and have not watched the error happen. Happy for it to be out of scope here.

tandede:
Thanks for catching this. The explicit numeric update path does divide by `update_interval`, so zero and negative values are invalid for the same reason.

I added configuration validation requiring numeric `update_interval` values to be at least 1, together with regression coverage for `0` and `-5`. The related configuration and ZenFlow tests now report 21 passed, and the pre-commit checks pass.

ebarkhordar:
Confirmed at 2f74e7e1 in a clean container: `update_interval` now rejects `0`, `-5`, `"0"` and `0.0`, while `1` and `"auto"` still construct. Thanks for the quick turnaround.

tandede:
> Confirmed at [2f74e7e](https://github.com/deepspeedai/DeepSpeed/commit/2f74e7e1cea6ad9e64611c5ee7f66f0ee52d1798) in a clean container: `update_interval` now rejects `0`, `-5`, `"0"` and `0.0`, while `1` and `"auto"` still construct. Thanks for the quick turnaround.

Thanks for re-testing and confirming the fix! Really appreciate the thorough review.

Antlera:
Hi @tandede. Thanks for this nice PR. I think most of the code is LGTM.

One more related comment, if you are willing to take it in this PR.

Rejecting topk_ratio in {0, 1} fixes the auto-path divisors, but a legal ratio can still collapse to zero selected columns on ZeRO-1/2:

  ` num_select = int(self.topk_ratio * (end_column - start_column))`

`int()` truncates toward zero, so e.g. `topk_ratio=0.01` and 50 columns
yields 0. That later hits torch.topk(..., k=0). ZeRO-3 already guards
this with max(1, int(...)) in engine_stage3.py.

Please apply the same floor in zenflow_stage_1_and_2.py, and keep the
buffer sizes in lockstep — do not change only the two num_select lines:

- update_selected_channels: `num_select`
- _process_selected_fp32_groups_grad: `num_select`
- average_tensor: `curr_column_size (index_buffer length)`
- average_tensor: `curr_selected_reduce_size (grad_buffer length)`

Selection is per partition, so sizing with int(num_col * topk_ratio) on the full parameter can still under-allocate after a per-partition max(1, ...). Use the same per-partition formula for allocation and
selection. Skip the floor when the column/element count is already 0.

Happy for this to stay out of scope if you prefer a follow-up.

tests/unit/runtime/zenflow/test_zf_config.py
```diff
@@ -59,6 +59,18 @@ def test_invalid_zenflow_type_raises():
...
 
 
+@pytest.mark.parametrize("topk_ratio", [0.0, 1.0])
+def test_topk_ratio_excludes_endpoints(topk_ratio):
+    with pytest.raises(ValidationError):
+        ZenFlowConfig(topk_ratio=topk_ratio)
+
+
+@pytest.mark.parametrize("update_interval", [0, -5])
+def test_update_interval_rejects_non_positive_values(update_interval):
```
Antlera, line 69: Could you also add a positive case for the new lower bound? `update_interval=1`
and `update_interval="auto"` should still construct. The existing payload with
8 covers a typical value, but not the boundary this check now allows.

tandede:
> Hi @tandede. Thanks for this nice PR. I think most of the code is LGTM.
> 
> One more related comment, if you are willing to take it in this PR.
> 
> Rejecting topk_ratio in {0, 1} fixes the auto-path divisors, but a legal ratio can still collapse to zero selected columns on ZeRO-1/2:
> 
> ` num_select = int(self.topk_ratio * (end_column - start_column))`
> 
> `int()` truncates toward zero, so e.g. `topk_ratio=0.01` and 50 columns yields 0. That later hits torch.topk(..., k=0). ZeRO-3 already guards this with max(1, int(...)) in engine_stage3.py.
> 
> Please apply the same floor in zenflow_stage_1_and_2.py, and keep the buffer sizes in lockstep — do not change only the two num_select lines:
> 
> * update_selected_channels: `num_select`
> * _process_selected_fp32_groups_grad: `num_select`
> * average_tensor: `curr_column_size (index_buffer length)`
> * average_tensor: `curr_selected_reduce_size (grad_buffer length)`
> 
> Selection is per partition, so sizing with int(num_col * topk_ratio) on the full parameter can still under-allocate after a per-partition max(1, ...). Use the same per-partition formula for allocation and selection. Skip the floor when the column/element count is already 0.
> 
> Happy for this to stay out of scope if you prefer a follow-up.

Thanks, this is a good point. I addressed it in 5b5beb6f.

I added a shared per-partition selection-count helper so that every non-empty partition uses max(1, int(num_columns * topk_ratio)), while a partition with zero complete columns still selects zero.

Both num_select paths now use this calculation, and the index-buffer and gradient-buffer sizes use the same per-partition formula, keeping their allocations in lockstep with selection.

I also added regression coverage for the topk_ratio=0.01 and 50-column case, the zero-column case, and a case selecting more than one column. The changed-file pre-commit checks pass, and the targeted ZenFlow/config tests report 18 passed.

Thanks for calling out the buffer-sizing requirement as well.

tohtana:
Thank you for your contribution, @tandede!
The code looks good to me. I will approve after @Antlera gives the greenlight.

Antlera:
> Thank you for your contribution, @tandede! The code looks good to me. I will approve after @Antlera gives the greenlight.

Hi @tohtana, thanks for following up on this, will approve the changes once the positive case as I commented is added.

tandede:
> Hi @tohtana, thanks for following up on this, will approve the changes once the positive case as I commented is added.

Thanks for the clarification. I added an end-to-end positive regression test in 1bdca9f9.

The test runs the actual ZeRO-1/2 distributed training path with two ranks, a 50-column parameter, and `topk_ratio=0.01`. This exercises the per-partition nonzero selection floor together with the corresponding index and gradient buffer allocations.

The focused local tests report 13 passed and 2 GPU-dependent cases skipped. The GPU cases are now queued in CI.

Antlera:
> > Hi @tohtana, thanks for following up on this, will approve the changes once the positive case as I commented is added.
> 
> Thanks for the clarification. I added an end-to-end positive regression test in [1bdca9f](https://github.com/deepspeedai/DeepSpeed/commit/1bdca9f9fe2e58eb9c0aeb3d29f0fb816946c4ce).
> 
> The test runs the actual ZeRO-1/2 distributed training path with two ranks, a 50-column parameter, and `topk_ratio=0.01`. This exercises the per-partition nonzero selection floor together with the corresponding index and gradient buffer allocations.
> 
> The focused local tests report 13 passed and 2 GPU-dependent cases skipped. The GPU cases are now queued in CI.

Thanks. LGTM.

tests/unit/runtime/zenflow/test_zf_config.py
Antlera: Hi @tandede, could you check on the construction tests: `update_interval=1`
and `update_interval="auto"` should still construct with all the commits.

tandede: Thanks for pointing this out. I added explicit positive coverage for the allowed boundary values `update_interval=1` and `update_interval="auto"`. The test constructs `ZenFlowConfig` with both values and verifies that they are preserved.

The ZenFlow configuration test file now reports 12 passed, and the changed-file pre-commit checks pass.

Antlera: Great!

Antlera (approved):
LGTM

Antlera:
> Thank you for your contribution, @tandede! The code looks good to me. I will approve after @Antlera gives the greenlight.

Hi @tohtana just approved after the latest commit.

PR #8277, opened 2026-08-19
Fix ZeRO parameter alignment for grouped_mm

### LINKED ISSUE #8276: [BUG] DeepSeek-V3 MLA rank changes trigger 16-byte data_ptr alignment failure in grouped_mm under DeepSpeed
### Description

DeepSeek-V3 training through DeepSpeed can fail in the Hugging Face MoE `grouped_mm` path with:

```text
RuntimeError: expected data_ptr to be aligned to 16 bytes
```

The failure is triggered by otherwise executable MLA LoRA-rank configurations. The same frozen configurations complete under a PyTorch-native runner and a Transformers/Accelerate runner, while the DeepSpeed-wrapped runner fails consistently.

I observed this pattern for 10 configurations (5 `kv_lora_rank` mutations and 5 `q_lora_rank` mutations), and re-ran one representative case (`kv_lora_rank=68`) three times; it failed with the same error 3/3 times.

I am filing this in DeepSpeed because the issue is specific to the DeepSpeed execution path in our cross-runner comparison, although the immediate failing call is in Transformers/PyTorch `grouped_mm`, so the root cause may be an interaction between DeepSpeed and the HF MoE path.

### Environment

- DeepSpeed: `v0.19.1` (`510ebe58e4e4148bbf49350ad18fe291a0afb381`)
- Transformers: `v5.9.0` (`0a2757da521a7a49b8143d9e0c938f08747d682e`)
- PyTorch: `2.13.0+cu126`
- CUDA runtime reported by PyTorch: `12.6`
- GPUs: 2x NVIDIA RTX A6000
- Python: 3.11

### Representative configuration

Small DeepSeek-V3-like configuration used for the reproducer:

```text
hidden_size=512
ffn_hidden_size=1024
num_layers=4
num_attention_heads=8
num_query_groups=2
seq_length=128
vocab_size=2048
num_experts=4
moe_router_topk=2
q_lora_rank=128
kv_lora_rank=68
qk_nope_head_dim=32
qk_rope_head_dim=32
v_head_dim=64
micro_batch_size=1
global_batch_size=4
bf16=true
fp16=false
```

The only mutation from the prequalified baseline in this representative case is `kv_lora_rank: 64 -> 68`.

### Observed behavior

The failure occurs during forward in the MoE experts path:

```text
transformers/models/deepseek_v3/modeling_deepseek_v3.py:245
  hidden_states = self.experts(hidden_states, topk_indices, topk_weights).view(*orig_shape)

transformers/integrations/moe.py:441
  proj_out = _grouped_linear(...)

transformers/integrations/moe.py:369
  out = _grouped_mm(input, weight.transpose(-2, -1), offs=offs)

transformers/integrations/moe.py:333
  return torch.nn.functional.grouped_mm(input.to(weight.dtype), weight, offs=offs)

torch/nn/functional.py:7139
  return torch._grouped_mm(...)

RuntimeError: expected data_ptr to be aligned to 16 bytes
```

Both distributed ranks fail with the same message.

### Cross-runner comparison

For the same frozen mutation intents:

- PyTorch-native: completes
- Transformers/Accelerate: completes
- DeepSpeed: fails with the 16-byte `data_ptr` alignment error

This was observed for:

- `kv_lora_rank`: 60, 65, 66, 67, 68
- `q_lora_rank`: 124, 129, 130, 131, 132

### Expected behavior

Wrapping the model with DeepSpeed should not make these otherwise executable shapes produce a misaligned tensor passed into `grouped_mm`, or the integration should normalize/copy the relevant tensor before the kernel call if a stronger alignment requirement is introduced.

fwerkor:
## Summary

ZeRO stage 1/2 stores model parameters as views into flattened fp16/bf16 buffers. Individual parameter views can start at non-16-byte offsets even when the flat buffer itself is aligned, which breaks alignment-sensitive kernels such as `torch._grouped_mm`.

Pad parameter boundaries inside the ZeRO flat buffer so model parameter views remain 16-byte aligned without duplicating misaligned parameters. The padded layout is propagated through partition/gradient offsets, LP↔HP linkage, DeepCompile gradient buffers, checkpoint restore, and `zero_to_fp32` reconstruction.

Older checkpoints without parameter-alignment padding remain loadable; their compact layout is converted when restored.

## Tests

- ZeRO-1/2 BF16 regression with a deliberately misaligned parameter layout.
- Verifies zero-copy aligned flat-buffer views across optimizer steps.
- Verifies checkpoint reload with `load_module_only=True` and `load_optimizer_states=False`.
- Synced with current `master`, including the ZeRO-1/2 DeepCompile file rename.

Fixes #8276


tests/unit/v1/zero/test_stage2_flatten_on_gpu.py
```diff
@@ -31,6 +42,46 @@ def _apply_dtype_to_config(config_dict, dtype):
     # fp32: no half-precision block
 
 
+class TestStage1ParamAlignment(DistributedTest):
+    world_size = 1
+
+    def test_model_params_remain_16_byte_aligned(self):
+        if not get_accelerator().is_bf16_supported():
+            pytest.skip("bf16 is not supported on this accelerator")
```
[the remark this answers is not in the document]
fwerkor: Added

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -795,7 +795,14 @@ def _configure_moe_settings(self):
     def _update_model_bit16_weights(self, group_index):
         updated_params = self.unflatten(self.bit16_groups_flat[group_index], self.round_robin_bit16_meta[group_index])
         for p, q in zip(self.round_robin_bit16_groups[group_index], updated_params):
-            p.data = q.data
+            # Some kernels, including torch grouped_mm, require 16-byte aligned parameter pointers.
+            if q.data_ptr() % 16 == 0:
+                p.data = q.data
+            elif (p.data.shape == q.shape and p.data.dtype == q.dtype and p.data.device == q.device
+                  and p.data.data_ptr() % 16 == 0):
```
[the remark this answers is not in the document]
fwerkor: Agreed. The regression is now parameterized over ZeRO stages 1 and 2 and runs with `world_size = 2`, so it exercises the stage-2 partition/all-gather flow as well as stage 1. Both variants pass on 2x RTX A6000.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -795,7 +795,14 @@ def _configure_moe_settings(self):
...
+                p.data.copy_(q.data)
+            else:
+                p.data = q.data.clone()
```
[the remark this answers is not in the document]
fwerkor: `update_lp_params()` did all-gather into `bit16_groups_flat` without calling `_update_model_bit16_weights()`, which could leave standalone aligned parameters stale. It now resynchronizes model parameters after the gather. The regression explicitly mutates the fp32 partition, calls `update_lp_params()`, and verifies the aligned model weight still matches the flat-buffer view for both ZeRO-1 and ZeRO-2 on 2 ranks.

tohtana (changes requested):
Hi @fwerkor, thank you for submitting this PR!

This is useful, but I think this approach has two challenges:

- This has too much impact about the performance and memory usage. All misaligned parameters will be duplicated.
- This can easily cause a bug about synchronization. Actually the current code leaves the flat buffer stale when a checkpoint is loaded, so the first optimizer step silently overwrites loaded misaligned parameters. (I reproduced this on ZeRO-1/2 with both `load_module_only=True` and `load_optimizer_states=False`)

It might be okay to accept this as an opt-in, but I still feel the performance impact is too big. I think a long term solution is to pad parameters in the flat buffer. It will need more changes, but the performance/memory impact will be small. 
Can you share your thoughts?

fwerkor:
> Hi @fwerkor, thank you for submitting this PR!
> 
> This is useful, but I think this approach has two challenges:
> 
> * This has too much impact about the performance and memory usage. All misaligned parameters will be duplicated.
> * This can easily cause a bug about synchronization. Actually the current code leaves the flat buffer stale when a checkpoint is loaded, so the first optimizer step silently overwrites loaded misaligned parameters. (I reproduced this on ZeRO-1/2 with both `load_module_only=True` and `load_optimizer_states=False`)
> 
> It might be okay to accept this as an opt-in, but I still feel the performance impact is too big. I think a long term solution is to pad parameters in the flat buffer. It will need more changes, but the performance/memory impact will be small. Can you share your thoughts?

Thanks, agreed. I switched to padding the ZeRO flat buffer and synced with current master. The previous GPU CI was killed by the 1-hour timeout (exit 137) at ~76%, with no failures before that.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -434,19 +440,34 @@ def _enforce_cpu_offload():
                     meta_tensors.append(torch.zeros_like(param.cpu_data, device="meta"))
             self.round_robin_bit16_meta.append(meta_tensors)
 
+            param_padding = []
+            param_offsets = []
+            current_offset = 0
+            for param_index, param in enumerate(round_robin_tensors):
+                param_offsets.append(current_offset)
+                current_offset += param.numel()
```
tohtana: When `flatten_on_accelerator` is false, `param.data` has already been replaced with a one-element placeholder above, so `param.numel()` here returns 1 rather than the original parameter size. However, the flat buffer is built from `param.cpu_data`, which still contains the full tensor.

For example, with two 64-element BF16 parameters, this records offsets `[0, 8]`, while the actual tensors start at `[0, 71]` with the generated padding. `_update_model_bit16_weights()` then binds the second parameter to the wrong slice, silently corrupting its values. This is reproducible with a minimal CPU reproduction of the layout and rebinding code.

How about computing the layout from `meta_tensors`, which preserve the original sizes on both paths, and add a regression test that forces the CPU flatten fallback?

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2141,6 +2214,72 @@ def get_flat_partition(self,
...
+            buffer_idx += tensor.numel()
+            param_id = self.get_param_id(tensor_list[i])
+            tensor_offset = self.grad_start_offset[param_group_idx][partition_id][param_id]
+            dest_offset = self.grad_partition_insertion_offset[param_group_idx][partition_id][param_id]
+            num_elements = min(tensor.numel() - tensor_offset, partition_size - dest_offset)
+
+            if dest_offset > current_size:
+                pad = torch.zeros(dest_offset - current_size, dtype=dtype, device=device)
+                pad._zero_padding = True
+                flat_tensor_list.append(pad)
```
tohtana, line 2256: Adding padding entries here also changes the indexing of `averaged_gradients`, since this list is stored there when `return_tensor_list=True`. However, `_init_lp_to_hp_mapping()` still assigns `_index_in_param_group` by counting only parameters, and `tensor_fragment.get_lp_grad_fragment()` uses that index directly into the gradient list.

For example, with one rank and BF16 parameter sizes `[1, 64]`, the gradient list has entry sizes `[1, 7, 64]`. The second parameter gets index 1, so `safe_get_full_grad()` selects the 7-element padding instead of its 64-element gradient and fails with a size mismatch. I confirmed this using the extracted methods with CPU tensors.

Can we make the gradient accessors use a mapping that accounts for padding, and cover both `safe_get_full_grad()` and `safe_set_full_grad()` with a regression test?

tohtana (changes requested):
Thank you for switching to padding the flat buffer and adding the checkpoint reload regression tests! This addresses my original concerns.
I found two additional correctness issues in the updated implementation, detailed inline below.
This PR can have a significant impact on broad usage of DeepSpeed. Can we validate the correctness by running a simple training loop?

fwerkor:
Thanks, addressed both issues.

- The CPU flatten fallback now computes parameter offsets and padding from the preserved metadata tensors, so it no longer depends on the temporary one-element parameter placeholders.
- The LP/HP gradient mapping now accounts for internal padding entries, fixing `safe_get_full_grad()` and `safe_set_full_grad()` for padded layouts.

I also added regression tests for the CPU flatten fallback with a short training loop, and for `safe_get_full_grad()` / `safe_set_full_grad()` under internal alignment padding, covering both ZeRO-1 and ZeRO-2.

The existing alignment/checkpoint regression still passes as well.

deepspeed/compile/init_z1_and_2.py
```diff
@@ -56,34 +57,32 @@ def _build_flat_partition_grad_views(optimizer, group_idx):
...
+        if num_elements > 0:
+            view = flat_buffer.narrow(0, dest_offset, int(num_elements))
+            if source_offset == 0 and num_elements == tensor.numel():
+                view = view.view(tensor.shape)
+            views.append(view)
+            current_size = dest_offset + int(num_elements)
 
     if current_size < partition_size:
-        views.append(flat_buffer.narrow(0, current_size, partition_size - current_size))
+        padding = flat_buffer.narrow(0, current_size, partition_size - current_size)
```
tohtana, line 93: This fixes the non-offload mapping, but `offload_gradient_dict` still excludes padding. With internal padding, index 2 accesses a two-entry list and `safe_get_full_grad()` raises `IndexError` in ZeRO-1/2. Could we keep the offload index consistent with its list?

deepspeed/compile/init_z1_and_2.py
```diff
@@ -102,7 +101,12 @@ def init_z1_and_2(engine, backend, compile_config, compile_kwargs, schedule=None
     if use_z2:
         grad_buffer = {}
         for i, group in enumerate(optimizer.bit16_groups):
-            grad_buffer[i] = [p.clone().detach() for p in _build_partition_grad_views(optimizer, i)]
+            partition_grad_views = _build_partition_grad_views(optimizer, i)
+            grad_buffer[i] = [p.clone().detach() for p in partition_grad_views]
```
tohtana: Could we preserve `_zero_padding` on these clones? Losing it shifts the gradient-to-parameter pairing in the norm calculation. In an isolated two-rank test using the production norm function, the norm changed from `8.0622578` to `6.0827627`, affecting clipping.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2645,22 +2780,139 @@ def state_dict(self):
...
+            } for group in saved_optimizer_state['param_groups']]
+        }
+
+        for i, param_group in enumerate(saved_optimizer_state['param_groups']):
+            param_id = param_group['params'][0]
+            saved_state = saved_optimizer_state['state'].get(param_id, {})
+            converted_param_state = {}
+
+            for key, value in saved_state.items():
+                if torch.is_tensor(value) and value.dim() > 0:
```
tohtana: Could we restrict redistribution to states matching the saved parameter partition shape? Non-scalar tensors can also be metadata: a one-element `tensor_step` causes `IndexError` when loading pre-padding checkpoints in both ZeRO-1 and ZeRO-2.

tohtana (changes requested):
Thank you for addressing my previous comments! The changes look good to me.
I left a few more comments (two of them already existed before your recent update).

fwerkor:
Thanks, addressed all three comments. I fixed the offload gradient indexing, preserved `_zero_padding` on cloned gradients, and restricted legacy optimizer-state redistribution to partition-shaped tensors. Added a ZeRO-1/2 regression for tensor metadata as well.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2679,11 +2933,14 @@ def refresh_fp32_params(self):
         self._restore_from_bit16_weights()
 
     # Extract optimizer state for current partition from merged states of all partitions
-    def _partition_base_optimizer_state(self, state_key, all_partition_states, group_id):
+    def _partition_base_optimizer_state(self, state_key, all_partition_states, group_id, unpadded_layout=False):
         partition_id = dist.get_rank(group=self.real_dp_process_group[group_id])
         alignment = self.nccl_start_alignment_factor * dist.get_world_size(group=self.real_dp_process_group[group_id])
         if torch.is_tensor(all_partition_states[0]):
-            flat_merged_partitions = self.flatten_dense_tensors_aligned(all_partition_states, alignment)
+            if unpadded_layout and all_partition_states[0].dim() > 0:
```
tohtana: The same pre-existing metadata issue remains here: a one-element `tensor_step` passes `dim() > 0` and enters `_expand_unpadded_flat_group()`. Could we apply the same partition-shape check as in the rigid-path fix?

tohtana (changes requested):
Thanks for the fixes! They look good to me.
I left one comment about a similar issue to the previous one.

As we are not super confident in the correctness of all the features, can we make this opt-in?
I opened a [PR](https://github.com/fwerkor/DeepSpeed/pull/1) to this PR branch. If you think it useful, please review and merge it. It would also be good if you could implement it separately.

deepspeed/runtime/zero/stage_1_and_2.py
fwerkor: Fixed in bdb741440 (now included in the PR head). Elastic restore now only redistributes tensor states whose shape matches the saved SINGLE_PARTITION_OF_FP32_GROUPS partition shape, so tensor metadata such as a one-element tensor_step is preserved instead of entering _expand_unpadded_flat_group(). Added a ZeRO-1/2 regression for the elastic pre-padding checkpoint path; the full TestStage12ParamAlignment suite passes (24/24).

fwerkor:
Thanks — I merged your opt-in PR into this branch and also addressed the elastic `tensor_step` metadata case from the remaining thread. The current head includes both fixes and the ZeRO-1/2 parameter-alignment regression suite passes 24/24. Please take another look when convenient.

tohtana (approved):
Thank you @fwerkor! Everything looks good to me. I appreciate your patience.
Let's merge this now.

PR #8278, opened 2026-08-19
Enable optimized Adam backend for MuonWithAuxAdam optimizer

jinyouzhi:
### Motivation

This pull request aim to extend the Muon optimizer auxiliary Adam(w) optimizer to support more built-in optimized implementation such as `FusedAdam`.

###  Changes

* Extracted Adam/AdamW backend selection logic into a new `_select_adam_optimizer` method, improving modularity and allowing consistent backend selection for both Adam and Muon optimizers. 
* Keep original naive implementation as fallback path and extend to support weight-decay (AdamW)
* Support `CPUAdam` and `Zenflow` 

### Tests
```
pytest tests/unit/ops/muon/test_muon_partial_training.py
```

图片

tested on RTX 5090: 128 Tensors 16,384 FP32 elements, warmup 30, steps 200.


tests/unit/ops/muon/test_muon_partial_training.py
```diff
@@ -214,3 +218,184 @@ def test_muon_adam_learning_rate_overrides(self, optimizer_params, expected_muon
...
+        }),
+        (_AdamSelectionEngine(cpu_offload=True, zenflow=True), {}, True, ZenFlowCPUAdam, {
+            "adamw_mode": True,
+            "fp32_optimizer_states": True,
+            "overlap_step": True
+        }),
+    ],
+)
+def test_adam_backend_selection(engine, parameters, adam_w_mode, expected_class, expected_kwargs):
+    optimizer_class, optimizer_kwargs = DeepSpeedEngine._select_adam_optimizer(engine, parameters, adam_w_mode)
```
sfc-gh-truwase: Why not this? 
```suggestion
    optimizer_class, optimizer_kwargs = engine._select_adam_optimizer(parameters, adam_w_mode)
```

tests/unit/ops/muon/test_muon_partial_training.py
```diff
@@ -20,10 +20,14 @@
 ensuring empty parameter groups are properly handled.
 """
 
+import torch
 import torch.nn as nn
 import deepspeed
 import pytest
 from unit.common import DistributedTest
+from deepspeed.ops.adam import DeepSpeedCPUAdam, FusedAdam, ZenFlowCPUAdam
+from deepspeed.runtime.engine import DeepSpeedEngine
```
sfc-gh-truwase, line 29: It seems strange to import `DeepSpeedEngine` class to access an engine method instead of using the engine object itself. Would it be better to implement `_select_adam_optimizer` as a utility belong in `deepspeed.runtime.utils` or `deepspeed.utils`?

tests/unit/ops/muon/test_muon_partial_training.py
jinyouzhi: Yes, it's strange. Thank you for your suggestions.

tests/unit/ops/muon/test_muon_partial_training.py
jinyouzhi: Good idea! I will move `_select_adam_optimizer` funtction into utility.

jinyouzhi: However, would moving _select_adam_optimizer out of engine.py make the code less readable? (Because only reference is also in engine.py) @sfc-gh-truwase

tests/unit/ops/muon/test_muon_partial_training.py
jinyouzhi: Fixed.

tests/unit/ops/muon/test_muon_partial_training.py
sfc-gh-truwase: Great question. My reading of `_select_adam_optimizer` is that it is extracting the optimizer configuration of the engine which originates from the `ds_config`. Thus,  `_select_adam_optimizer` **ideally** belongs in a utility because it does not use engine runtime logic (e.g., [get_lr](https://github.com/deepspeedai/DeepSpeed/blob/80c8e5ba4283b600a2744b5d04b773da1a112a52/deepspeed/runtime/engine.py#L3605)), instead it is uses configuration logic (e.g., [bf16_optimizer_states](https://github.com/deepspeedai/DeepSpeed/blob/80c8e5ba4283b600a2744b5d04b773da1a112a52/deepspeed/runtime/engine.py#L1334)). However, because engine currently intermingles configuration and runtime information (poor code quality), significant refactoring is required to place `_select_adam_optimizer` appropriately. So, my proposal is to 
1. Rename `_select_adam_optimizer` to  more appropriately to `get_optimizer_configuration`
2. Drop the `_` to indicate that is an external facing method
3. Access directly as the engine object method rather than the class method

sfc-gh-truwase: For extra brownie points, you could submit a followup PR [consolidate](https://github.com/deepspeedai/DeepSpeed/blob/80c8e5ba4283b600a2744b5d04b773da1a112a52/deepspeed/runtime/engine.py#L284-L285) `self.config` and `self._config` and refactoring them out of  `engine.py`. That would be greatly appreciated :)

sfc-gh-truwase:
@PKUWZP FYI

tests/unit/ops/muon/test_muon_partial_training.py
```diff
@@ -20,10 +20,13 @@
 ensuring empty parameter groups are properly handled.
 """
 
+import torch
```
sfc-gh-truwase, line 23: Relocate to file to tests/unit/v1/ops/muon

tests/unit/ops/muon/test_muon_partial_training.py
jinyouzhi: Thank you very much! very kindly and valuable advice.

tests/unit/ops/muon/test_muon_partial_training.py
jinyouzhi: done

sfc-gh-truwase (approved):

tests/unit/ops/muon/test_muon_partial_training.py
jinyouzhi: Updated the implementation following your suggestion. @sfc-gh-truwase

tests/unit/ops/muon/test_muon_partial_training.py
jinyouzhi: @sfc-gh-truwase Sorry, I need to revert this modification. `engine` here is an instance of the local `_AdamSelectionEngine` test double, not a fully constructed `DeepSpeedEngine`. Because of that, we need to call DeepSpeedEngine's method directly (`DeepSpeedEngine.get_optimizer_configuration(engine, ...)`) rather than as a bound method on engine.

Fully instantiating a real `DeepSpeedEngine` (via `deepspeed.initialize()`) would require a model, config, and distributed backend — overhead that's disproportionate to what this test is actually checking, and it would distract from the test's real purpose (verifying Adam-backend selection logic).

pengdurice (commented):
Thank you for the great PR. I think this changes warrants an end to end test with loss value and curves reported on a realistic training (e.g. a 128m model for 100 steps) between baseline and this change. LMK what you think.

jinyouzhi:
> Thank you for the great PR. I think this changes warrants an end to end test with loss value and curves reported on a realistic training (e.g. a 128m model for 100 steps) between baseline and this change. LMK what you think.

Thank you for your suggestion. I agree that an end-to-end training loss comparison would be necessary. I’ll set up a training-loss curve check focusing on the auxiliary Adam backend with Muon. Do you have a particular model or configuration you would recommend?

pengdurice:
> > Thank you for the great PR. I think this changes warrants an end to end test with loss value and curves reported on a realistic training (e.g. a 128m model for 100 steps) between baseline and this change. LMK what you think.
> 
> Thank you for your suggestion. I agree that an end-to-end training loss comparison would be necessary. I’ll set up a training-loss curve check focusing on the auxiliary Adam backend with Muon. Do you have a particular model or configuration you would recommend?

Thank you!
Let's maybe use a small model e.g. 128mb ish? and make sure the naming of the parameters are recognizable by how we split between muon and Adam (https://github.com/deepspeedai/DeepSpeed/blob/aa3914df82e8ba2d5dc2462b06c0160415e674db/deepspeed/__init__.py#L84)
test baseline and the possible arms after the PR (fused Adam, cpu Adam etc)
Also, make sure the lr are set appropriately (not sure if the grad clipping issue is solved or not), the seed is set so for each arm same data example is used by different arms at the same step. 
LMK what you think, thank you!

tohtana:
Hi @jinyouzhi,
Thank you for submitting this PR! This is a significant improvement. The change looks okay to me, but I want to give you a heads-up.

Before this PR, non-Muon groups under the Muon optimizer use the inline Adam path. After this PR is merged, the same Muon configuration automatically selects FusedAdam. A pre-PR checkpoint therefore loads under the same user-facing optimizer and configuration, but its first resumed step fails because the new backend expects fields absent from the saved state (e.g. `KeyError: 'bias_correction'`).
Is there a good way to avoid it?

jinyouzhi:
> Hi @jinyouzhi, Thank you for submitting this PR! This is a significant improvement. The change looks okay to me, but I want to give you a heads-up.
> 
> Before this PR, non-Muon groups under the Muon optimizer use the inline Adam path. After this PR is merged, the same Muon configuration automatically selects FusedAdam. A pre-PR checkpoint therefore loads under the same user-facing optimizer and configuration, but its first resumed step fails because the new backend expects fields absent from the saved state (e.g. `KeyError: 'bias_correction'`). Is there a good way to avoid it?

Nice catch, @tohtana! Thanks, I missed that. I’ve pushed a fix in commit [cf0f900](https://github.com/deepspeedai/DeepSpeed/commit/cf0f9008fddcb8d644058a2994a24d07b25e3148).

jinyouzhi:
> > > Thank you for the great PR. I think this changes warrants an end to end test with loss value and curves reported on a realistic training (e.g. a 128m model for 100 steps) between baseline and this change. LMK what you think.
> > 
> > 
> > Thank you for your suggestion. I agree that an end-to-end training loss comparison would be necessary. I’ll set up a training-loss curve check focusing on the auxiliary Adam backend with Muon. Do you have a particular model or configuration you would recommend?
> 
> Thank you! Let's maybe use a small model e.g. 128mb ish? and make sure the naming of the parameters are recognizable by how we split between muon and Adam (
> 
> https://github.com/deepspeedai/DeepSpeed/blob/aa3914df82e8ba2d5dc2462b06c0160415e674db/deepspeed/__init__.py#L84
> 
> )
> test baseline and the possible arms after the PR (fused Adam, cpu Adam etc)
> Also, make sure the lr are set appropriately (not sure if the grad clipping issue is solved or not), the seed is set so for each arm same data example is used by different arms at the same step.
> LMK what you think, thank you!

I am still working on it. Quick update, the losses of FusedAdam and PyTorch impl is close to base's, there's a gap about CPUAdam, still investigate into it. Thank you for your advice.

jinyouzhi:
```
Dataset: tatsu-lab/alpaca
Tokenizer: /mnt/disk3/hf_models/Qwen3-8B-FP8
Seed: 1234
Steps: 100
Sequence length: 128
Vocab size:   151,669
Hidden size:   1,024
Layers:            7
Total params: 318,116,981
≈ 318M parameters
Weight decay: 0
GPU: RTX 5090 D

┌────────────────────┬──────────────────────┬─────────────┬─────────────┬─────────────────────┐
│ Configuration      │ Backend              │  First Loss │  Final Loss │ Max Diff vs. Pre-PR │
├────────────────────┼──────────────────────┼─────────────┼─────────────┼─────────────────────┤
│ Pre-PR (24402c3be) │ Original inline Adam │ 11.93690586 │ 11.51892757 │                   0 │
├────────────────────┼──────────────────────┼─────────────┼─────────────┼─────────────────────┤
│ New PR             │ FusedAdam            │ 11.93690586 │ 11.51893520 │             1.14e-5 │
├────────────────────┼──────────────────────┼─────────────┼─────────────┼─────────────────────┤
│ New PR             │ PyTorch AdamW        │ 11.93690586 │ 11.51893520 │             1.05e-5 │
├────────────────────┼──────────────────────┼─────────────┼─────────────┼─────────────────────┤
│ New PR             │ CPUAdam offload      │ 11.93690586 │ 11.46116829 │             9.37e-2 │
└────────────────────┴──────────────────────┴─────────────┴─────────────┴─────────────────────┘
```

CPUAdam itself is supported. However, Muon CPU offload is not currently correctness-safe because ZeRO’s flat CPU partitioning path bypasses Muon’s full-matrix update semantics. I plan to address the Muon CPU offload issues in a separate PR.

pengdurice (approved):

pengdurice:
> ```
> Dataset: tatsu-lab/alpaca
> Tokenizer: /mnt/disk3/hf_models/Qwen3-8B-FP8
> Seed: 1234
> Steps: 100
> Sequence length: 128
> Vocab size:   151,669
> Hidden size:   1,024
> Layers:            7
> Total params: 318,116,981
> ≈ 318M parameters
> Weight decay: 0
> GPU: RTX 5090 D
> 
> ┌────────────────────┬──────────────────────┬─────────────┬─────────────┬─────────────────────┐
> │ Configuration      │ Backend              │  First Loss │  Final Loss │ Max Diff vs. Pre-PR │
> ├────────────────────┼──────────────────────┼─────────────┼─────────────┼─────────────────────┤
> │ Pre-PR (24402c3be) │ Original inline Adam │ 11.93690586 │ 11.51892757 │                   0 │
> ├────────────────────┼──────────────────────┼─────────────┼─────────────┼─────────────────────┤
> │ New PR             │ FusedAdam            │ 11.93690586 │ 11.51893520 │             1.14e-5 │
> ├────────────────────┼──────────────────────┼─────────────┼─────────────┼─────────────────────┤
> │ New PR             │ PyTorch AdamW        │ 11.93690586 │ 11.51893520 │             1.05e-5 │
> ├────────────────────┼──────────────────────┼─────────────┼─────────────┼─────────────────────┤
> │ New PR             │ CPUAdam offload      │ 11.93690586 │ 11.46116829 │             9.37e-2 │
> └────────────────────┴──────────────────────┴─────────────┴─────────────┴─────────────────────┘
> ```
> 
> CPUAdam itself is supported. However, Muon CPU offload is not currently correctness-safe because ZeRO’s flat CPU partitioning path bypasses Muon’s full-matrix update semantics. I plan to address the Muon CPU offload issues in a separate PR.

thank you for the great work! Look forward  to the PR fixing Muon CPU offload issue.

PR #8280, opened 2026-08-19
Fix ZeRO-1/2 with zero-sized parameters

### LINKED ISSUE #8279: [BUG] ZeRO-1/2 fail on trailing zero-sized trainable parameters
**Describe the bug**
DeepSpeed ZeRO stage 1 and stage 2 fail during backward when a zero-sized trainable parameter appears at the end of a parameter group and participates in autograd.

A minimal example using `nn.Linear(4, 0, bias=False)` fails with:

```text
KeyError: 2
```

at:

```text
deepspeed/runtime/zero/stage_1_and_2.py
partition_ids = self.param_to_partition_ids[i][param_id]
```

The same model works with ZeRO stage 0 and with PyTorch DDP. The failure also reproduces with a single GPU and FP32, so it does not depend on NCCL world size or mixed precision.

I reproduced this on DeepSpeed 0.19.1 and current `master` (`24402c3be79406b1ff10f6e851e1fbc41dfb5e3e`).

**To Reproduce**

Save as `repro.py`:

```python
import os
import torch
import torch.nn as nn
import deepspeed

local_rank = int(os.environ.get("LOCAL_RANK", 0))
torch.cuda.set_device(local_rank)


class Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.main = nn.Linear(4, 4)
        self.zero = nn.Linear(4, 0, bias=False)

    def forward(self, x):
        return self.main(x).pow(2).mean() + self.zero(x).sum() * 0


model = Model().cuda()
optimizer = torch.optim.AdamW(model.parameters(), lr=1e-4)

config = {
    "zero_optimization": {
        "stage": 1,
        "overlap_comm": False,
        "reduce_scatter": True,
    },
    "steps_per_print": 1000,
}

engine, optimizer, _, _ = deepspeed.initialize(
    model=model,
    optimizer=optimizer,
    config=config,
)

x = torch.randn(2, 4, device=engine.device)
loss = engine(x)
engine.backward(loss)
engine.step()
```

Run on one GPU:

```bash
CUDA_VISIBLE_DEVICES=0 python -m torch.distributed.run \
  --nnodes=1 --nproc-per-node=1 \
  --master-addr=127.0.0.1 --master-port=29670 \
  repro.py
```

Observed with ZeRO stage 1 and 2:

```text
File "deepspeed/runtime/zero/stage_1_and_2.py", line 1265, in average_tensor
    partition_ids = self.param_to_partition_ids[i][param_id]
KeyError: 2
```

Controls:

- ZeRO stage 0: passes.
- PyTorch DDP with the same zero-sized parameter: passes.
- ZeRO stage 1, one GPU, FP32: fails with the same `KeyError`.
- ZeRO stage 2, one GPU, FP32: fails with the same `KeyError`.
- If the zero-sized parameter does not participate in backward, the run passes.

The internal parameter ids for the minimal model are `0`, `1`, and `2`, but after ZeRO initialization the mapping is:

```text
param_to_partition_ids = {0: {0: [0], 1: [0]}}
```

The trailing zero-sized parameter (`param_id == 2`) is therefore absent from `param_to_partition_ids`, while its empty gradient can still enter the reduction bucket during backward. The lookup then raises `KeyError: 2`.

As a root-cause check, skipping `param.numel() == 0` entries in the reduction loop makes the same reproducer complete successfully.

**Expected behavior**
ZeRO stage 1/2 should handle zero-sized trainable parameters consistently with PyTorch, or reject them explicitly during initialization with a clear validation error. It should not fail later in gradient reduction with an internal `KeyError`.

**ds_report output**

```text
torch version .................... 2.13.0+cu126
deepspeed info ................... 0.19.1
torch cuda version ............... 12.6
nvcc version ..................... 12.4
shared memory (/dev/shm) size .... 503.77 GB
```

The issue also reproduces against current DeepSpeed `master` at `24402c3be79406b1ff10f6e851e1fbc41dfb5e3e` using the same environment.

**Screenshots**
Not applicable; the text reproducer and traceback are included above.

- OS: Linux 5.15.0-84-generic x86_64
- GPU count and types: tested on NVIDIA RTX A6000; minimal reproducer requires only one GPU
- Interconnects: not applicable to the single-GPU reproducer
- Python version: 3.11.15
- PyTorch: 2.13.0+cu126
- CUDA runtime: 12.6
- DeepSpeed: 0.19.1 and current master `24402c3be79406b1ff10f6e851e1fbc41dfb5e3e`

**Launcher context**
Launched with `python -m torch.distributed.run`. The minimal reproducer uses one local process/GPU.

**Docker context**
No Docker image is required.

**Additional context**
This was originally observed in a larger model with a zero-sized output projection, where the same failure appeared as `KeyError: 110`. The standalone example above reduces it to a single trailing zero-sized parameter and reproduces the same ZeRO mapping failure.

fwerkor:
## Summary
- skip zero-sized parameters before ZeRO-1/2 gradient reduction
- add regression coverage for stages 1 and 2

## Testing
- `pytest -q tests/unit/runtime/zero/test_zero_empty_param.py -s`
- original #8279 reproducer on A6000 with PyTorch 2.13.0+cu126 for ZeRO-1 and ZeRO-2
- `pre-commit run --files deepspeed/runtime/zero/stage_1_and_2.py tests/unit/runtime/zero/test_zero_empty_param.py`

Fixes #8279

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -1192,6 +1192,9 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
     ############### Independent Partition Gradient ########################
     def reduce_independent_p_g_buckets_and_remove_grads(self, param, i):
 
+        if param.numel() == 0:
+            return
+
         grad_reduc = self.get_gradient_for_reduction(param)
         comm_dtype = self.get_param_comm_dtype(param)
         bucket = self.ipg_buckets[comm_dtype]


sfc-gh-truwase (approved):

tests/unit/runtime/zero/test_zero_empty_param.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Microsoft Corporation.
```
sfc-gh-truwase: Remove this UT since it is not representative of real use cases.

fwerkor: OK

PR #8284, opened 2026-08-21
Run tests/unit/model_parallelism in the modal GPU workflow

delock:
## What this does

`tests/unit/model_parallelism` was never executed in PR CI:

- **cpu-torch-latest** collects it but skips every test: the whole directory is multi-rank `DistributedTest` (world_size 2/4), and `DistributedExec._launch_procs` (`tests/unit/common.py`) skips them because a CPU runner reports `device_count() == 1`.
- **modal-torch-latest** (2xL40S GPU) has the devices to run them, but its diff-driven selector only considered `tests/unit/v1`.

This came to light in #8241, whose regression test for #8231 (`TestAutoTPMultipleModels`) is in that directory and therefore ran nowhere in CI.

## Changes

- `ci/tests_fetcher.py`: add `tests/unit/model_parallelism` to the modal workflow `test_scopes`.
- `ci/torch_latest.py`: derive `_TEST_FILE_RE` and the all-mode selection check from the workflow's `test_scopes` (imported from `tests_fetcher`) instead of hardcoding `tests/unit/v1`, so the selector and the validator share one source of truth.
- `ci/test_torch_latest.py`: update the self-tests for the two-scope reality (all-mode selection content, a model_parallelism subset target, new invalid cases, `count=` assertion).

## Validation

- `python3 ci/test_tests_fetcher.py` — 16/16 passed
- `python3 ci/test_torch_latest.py` — 20/20 passed
- End-to-end locally: with a scratch commit touching `deepspeed/module_inject/tp_shard.py`, `ci/tests_fetcher.py --workflow modal-torch-latest` selects all 6 `model_parallelism` test files (mode=subset, 45 targets), and `ci/torch_latest.py validate-selection --mode subset` accepts the list.

## Notes

- The collect-tests job checks out the **base** revision of the CI scripts (`pull_request_target`), so this only takes effect for other PRs once it lands on master; this PR's own run validates the selection plumbing but will still select v1-only tests for its own diff.
- `tp_size=4` tests will still be skipped on the 2-GPU modal sandbox via the existing device-count gate; world_size=2 tests (including `TestAutoTPMultipleModels`) will run for real.

### DIFF
--- a/ci/test_torch_latest.py
+++ b/ci/test_torch_latest.py
@@ -37,6 +37,9 @@ def _selection_file(content: str) -> tuple[Path, Path]:
     return root, path
 
 
+_ALL_MODE_SELECTION = "tests/unit/v1\ntests/unit/model_parallelism\n"
+
+
 def _valid_env(path: Path, **overrides: str) -> dict[str, str]:
     values = {
         "GITHUB_EVENT_NAME": "pull_request_target",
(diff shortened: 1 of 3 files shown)


sfc-gh-truwase:
@delock please move `tests/unit/model_parallelism` to `tests/unit/v1/autotp`. We are using `tests/unit/v1` for tests to preserve.

delock:
Closing in favor of the simpler path: moved `tests/unit/model_parallelism` to `tests/unit/v1/autotp` in #8241 (delock/deepspeed@123912f), since `tests/unit/v1` is already the modal GPU workflow's selection scope and the tests-to-preserve tree. The xpu workflow path was updated to match. No CI-scope changes needed anymore.

delock:
@sfc-gh-truwase Done — moved to `tests/unit/v1/autotp` in #8241 (commit 123912f), closing this in favor of that.

PR #8292, opened 2026-08-22
Make DeepSpeed train on Apple Silicon (MPS) with ZeRO 0-3

PKUWZP:
## Summary

The MPS accelerator was a stub: memory queries returned `None`, no communication backend was set, fp16/bf16 were reported unsupported, and every op builder resolved to `NotImplementedBuilder`. `deepspeed.initialize` + one training step failed for every ZeRO stage on an Apple Silicon machine.

This PR is the first step of Apple Silicon support: make single-device training work end to end with pure-PyTorch ops. Metal kernels come later and plug into the `op_builder/mps` classes added here.

### Changes

- **`accelerator/mps_accelerator.py`** — real `torch.mps` memory stats, fp16/bf16 support (bf16 gated on macOS 14+), `torch.mps.Event`, `gloo` as the comm backend, and `is_synchronized_device() = True` (MPS has a single in-order command queue and no public streams or `record_stream`). Unified memory makes `pin_memory` a no-op (torch's `pin_memory()` also raises under MPS).
- **`deepspeed/comm/torch.py`** — gloo cannot operate on MPS tensors (even at world size 1), so collectives stage MPS tensors through CPU copies via a `stage_on_cpu` decorator; async ops copy back on `wait()`.
- **`accelerator/abstract_accelerator.py` + `runtime/zero`** — MPS has no fp64. Gradient-norm accumulation now picks its dtype via a new concrete `is_fp64_supported()` (default `True`) and `get_norm_dtype()` instead of hard-coded `.double()`.
- **`op_builder/mps/`** — new backend package (`MPSOpBuilder`, `NotImplementedBuilder`, `FusedAdamBuilder`). `FusedAdam` is implemented with `torch._foreach_*` ops and mirrors the math in `csrc/adam/multi_tensor_adam.cu`, following the HPU precedent of Python-backed builders.
- **`tests/unit/common.py`** — MPS must use `spawn` (Metal's compiler service is lost in `forkserver` children, which hangs the harness) and reports its device count via the accelerator.
- **`tests/unit/ops/adam/test_adamw.py`** — `test_fused_adam_matches_torch` checks `FusedAdam` against `torch.optim.Adam/AdamW` on the active accelerator (fp32/bf16 × Adam/AdamW), so it also guards the CUDA kernel.

### Verified on an M5 Max (macOS 26.3, torch 2.13.0)

- ZeRO 0/1/2/3 × fp32/bf16/fp16 train end to end with `deepspeed.initialize` (single process).
- MPS `FusedAdam` matches `torch.optim` to 2e-7 in fp32.
- `DS_ACCELERATOR=mps pytest unit/runtime/test_ds_config_dict.py unit/runtime/test_ds_initialize.py unit/runtime/half_precision/test_fp16.py unit/runtime/half_precision/test_dynamic_loss_scale.py unit/runtime/zero/test_zero_grad_clip.py unit/runtime/zero/test_zero_context.py unit/checkpoint/test_zero_optimizer.py`: 132 passed, 0 failed, 126 skipped (multi-device tests; `device_count() == 1`).

### Known limitations / follow-ups

- bf16 `FusedAdam` differs from the CUDA kernel by ~1 bf16 ulp (CUDA computes in fp32 and stores bf16; the `_foreach` path rounds in bf16).
- The CPU-staged gloo path is only exercised at world size 1 here; multi-Mac runs are untested.
- Follow-ups: macOS arm64 CI workflow, arm64 build of CPU Adam for ZeRO-Offload, Metal kernels via `torch.mps.compile_shader`, and an Apple Silicon tutorial page.

### DIFF
--- a/accelerator/abstract_accelerator.py
+++ b/accelerator/abstract_accelerator.py
@@ -174,6 +174,10 @@ class DeepSpeedAccelerator(ABC):
     def is_fp16_supported(self):
         ...
 
+    # Not abstract: nearly every accelerator supports fp64, so only those that do not need to override.
+    def is_fp64_supported(self):
+        return True
+
     @abc.abstractmethod
     def supported_dtypes(self):
         ...
--- a/accelerator/mps_accelerator.py
+++ b/accelerator/mps_accelerator.py
@@ -19,11 +19,14 @@ class MPS_Accelerator(DeepSpeedAccelerator):
 
     def __init__(self):
         self._name = "mps"
-        self._communication_backend_name = None
+        # MPS has no native collective backend; gloo is the only torch backend available on macOS.
+        self._communication_backend_name = "gloo"
         self._compile_backend = "inductor"
 
     def is_synchronized_device(self):
-        return False
+        # MPS runs everything on a single in-order command queue and exposes no user-visible
+        # streams, so DeepSpeed never needs to synchronize between streams on this device.
+        return True
 
     def use_host_timers(self):
         # Event timers are not supported on MPS
@@ -90,7 +93,8 @@ class MPS_Accelerator(DeepSpeedAccelerator):
         return None
 
     def stream(self, stream):
-        return None
+        from deepspeed.runtime.utils import noop_context
+        return noop_context()
 
     def current_stream(self, device_index=None):
         return None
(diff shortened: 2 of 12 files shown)


PKUWZP:
Superseded by #8293 (same branch pushed directly to the upstream repo).

Issue #8290, opened 2026-08-22
HF transformers main injects 'embedding_rowwise' into tp_plan for tied-embedding models; AutoTP rejects the whole plan


delock:
## Description

Two days ago transformers merged [#47579 "TP dtensor API inference + training"](https://github.com/huggingface/transformers/pull/47579), which makes `PretrainedConfig.__init__` inject a new tp_plan entry for tied-embedding models:

```python
# transformers/src/transformers/configuration_utils.py (main, since 861f4c41)
if getattr(self, "tie_word_embeddings", False) and self.base_model_tp_plan is not None:
    self.base_model_tp_plan = {
        **self.base_model_tp_plan,
        "embed_tokens": "embedding_rowwise",
    }
```

`embedding_rowwise` means sharding the embedding along the vocab dimension (their new tied `embed_tokens`/`lm_head` TP semantics).

DeepSpeed's converter is a strict allowlist and rejects the whole plan on any unknown style:

```python
# deepspeed/module_inject/tp_plan_converter.py
SUPPORTED_STYLES = {"colwise", "colwise_rep", "colwise_gather_output", "rowwise", "replicated_with_grad_allreduce"}
...
raise ValueError(f"HuggingFace tp_plan contains unsupported partition style(s): {sorted(unsupported)}. ...")
```

As a result, **any tied-embedding HF model** (Qwen2/Qwen3 small, Llama, Gemma, ...) now fails AutoTP initialization against transformers main:

```
ValueError: HuggingFace tp_plan contains unsupported partition style(s): ['embedding_rowwise']. Applying only the
supported entries could shard one half of a column/row pair, so the plan is rejected as a whole. Provide an explicit
'tensor_parallel.partition_config' for this model instead.
```

This is not tied to a particular DeepSpeed change — it surfaced in the GPU CI of #8241 simply because that PR moved `tests/unit/model_parallelism` into the modal GPU workflow (which tests against transformers `main`), and `test_tp_plan_real_models.py::test_qwen2_tied_lm_head_falls_back_to_replicated` (a `tie_word_embeddings=True` Qwen2Config) started failing there.

## Repro

```python
import torch, deepspeed
from transformers import AutoModelForCausalLM, Qwen2Config   # transformers main, >= 861f4c41 (2026-08-21)

config = Qwen2Config(vocab_size=1000, hidden_size=128, intermediate_size=256,
                     num_hidden_layers=1, num_attention_heads=4, num_key_value_heads=4,
[...]

PR #8294, opened 2026-08-23
[AutoTP] Convert HF embedding_rowwise tp_plan entries to SKIP specs

ebarkhordar:
## Problem

huggingface/transformers#47579 (on `main` since `861f4c41`, 2026-08-21) makes `PretrainedConfig.__init__` inject `"embed_tokens": "embedding_rowwise"` into `base_model_tp_plan` whenever `tie_word_embeddings` is true. `SUPPORTED_STYLES` is a strict allowlist and `convert()` raises on any style outside it, rejecting the whole plan, so AutoTP plan conversion now fails for every tied-embedding model (Qwen2/Qwen3, Llama, Gemma) built against transformers `main`.

## Fix

Recognize `embedding_rowwise` and convert it to a `SKIP` spec, which is option 2 in #8290: the entry is understood and the embedding is deliberately left replicated.

What follows is the behaviour DeepSpeed already implements rather than a new policy. `lm_head` still converts to a gathered column spec, and `_configure_gathered_column_tie_fallbacks` then sees that `lm_head.weight is embed_tokens.weight` and leaves both modules replicated, logging that coupled vocabulary-parallel embedding is not supported yet. A tied model is therefore left in the shape it has on a transformers release without the injection, with both modules replicated and the tie intact.

The entry maps to `SKIP` with `grad_allreduce` left false, unlike `replicated_with_grad_allreduce`. The parameter is never split, so `register_replicated_grad_hooks` must not register an all-reduce for it.

Styles that are still unknown continue to reject the whole plan. `test_unsupported_style_rejects_whole_plan` is unchanged and still passes.

## Verification

Run on CPU in a container at `edaa7221`, against transformers `main` (5.16.0.dev0) and torch 2.13.0+cpu.

- The two added tests fail on master with the reported `ValueError` and pass with this change.
- `tests/unit/module_inject/` and `tests/unit/runtime/test_tp_plan_extraction.py`: 47 passed, on Python 3.11 and on 3.12.
- `pre-commit run --files` on the three changed files passes yapf, check-torchdist, check-license and codespell; flake8 5.0.4 exits 0 on them under Python 3.11.
- Not verified here: `test_qwen2_tied_lm_head_falls_back_to_replicated`, which needs 2 GPUs. That is the test #8290 reports as failing and the one this change is meant to restore.

## Two things worth deciding separately

Scoping this to `embedding_rowwise` leaves the next transformers-side style to fail the same way, since the injection is unconditional and the allowlist is deny-by-default against a vocabulary DeepSpeed does not own. A general rule for unknown styles looks like a maintainer call rather than something to settle here.

Related to that, the `convert()` docstring says entries with an unsupported style become SKIP specs instead of invalidating the plan, but no code path does that, and none does after this change either: an unsupported style still raises before the loop is reached. The docstring and the raise arrived together in #8204, so I have left both alone. Happy to follow up once you have picked the policy.

Refs #8290. This covers the conversion failure only, and does not implement vocabulary-parallel tied embeddings (option 1 or 3 in that issue), so I have not used a closing keyword.

### DIFF
--- a/deepspeed/module_inject/tp_plan_converter.py
+++ b/deepspeed/module_inject/tp_plan_converter.py
@@ -6,8 +6,11 @@
 from typing import List, Dict, Optional
 from .autotp_config import TPLayerSpec, PartitionType
 
-SUPPORTED_STYLES = {"colwise", "colwise_rep", "colwise_gather_output", "rowwise", "replicated_with_grad_allreduce"}
+SUPPORTED_STYLES = {
+    "colwise", "colwise_rep", "colwise_gather_output", "rowwise", "replicated_with_grad_allreduce", "embedding_rowwise"
+}
 # `colwise_rep` was renamed to `colwise_gather_output` in huggingface/transformers#42809.
+# `embedding_rowwise` is injected for every tied-embedding model since huggingface/transformers#47579.
 
 
 class TPPlanConverter:
@@ -50,6 +53,10 @@ class TPPlanConverter:
                 gather_output = partition_style != "colwise"
             elif partition_style == "rowwise":
                 partition_type = PartitionType.ROW
+            elif partition_style == "embedding_rowwise":
+                # Vocabulary-parallel embeddings are not supported yet, so the embedding stays whole and
+                # the gathered-column tie fallback keeps any LM head tied to it replicated as well.
+                partition_type = PartitionType.SKIP
             else:  # replicated_with_grad_allreduce, the only other supported style
                 # The parameter stays whole; only its gradient needs summing across the group.
                 partition_type = PartitionType.SKIP
--- a/deepspeed/module_inject/tp_plan_converter.py
+++ b/deepspeed/module_inject/tp_plan_converter.py
@@ -20,15 +20,21 @@ class TPPlanConverter:
     def convert(hf_tp_plan: Dict[str, str]) -> Optional[List[TPLayerSpec]]:
         """Convert HF tp_plan to DeepSpeed layer specs.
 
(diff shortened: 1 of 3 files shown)


delock:
Hi @ebarkhordar , thanks for your timely fix to this issue.  For your opens:
1. Unknown entries should lead to an explicit fail, allowing concious decision made on each new entry.  Let's see whether this strategy works.
2. Yes, the docstring needs to be updated to reflect this strategy, can you update doc string in this PR?  Thanks!

ebarkhordar:
Thanks, both are settled in `0aa0c41`.

1. No change. An unknown style still raises and rejects the whole plan, and `test_unsupported_style_rejects_whole_plan` still pins that.

2. The `convert()` docstring now describes that strategy instead of the SKIP-on-unsupported behaviour it claimed, which no code path has implemented since the raise landed in #8204. It also names the two supported styles that do become SKIP specs, `embedding_rowwise` and `replicated_with_grad_allreduce`. One more line in the same docstring was stale and I fixed it there rather than leaving it for later: the function returns None only for an empty plan, not when no entry is convertible.

The commit touches that docstring and nothing else. yapf, check-torchdist, check-license and codespell pass on the file, and flake8 5.0.4 exits 0 on it under Python 3.11.

delock (approved):

delock:
Hi @ebarkhordar , thanks for your timely fix.  Now we can move tied tests to modal GPU CI.   I have a question for you, do you think we should persue option 1 in the long run?  What is pros and cons for option 1?  Thanks!

PR #8299, opened 2026-08-23
Fix AutoTP metadata updates for unsharded modules

### LINKED ISSUE #8285: [AutoTP] update_mp_params shards vision-tower attributes (patch_embed.embed_dim) without sharding weights, breaking multimodal forward
## Description

When AutoTP (tp_size > 1) is applied to a multimodal VLM (e.g. Qwen2-VL), `AutoTP.update_mp_params` (`deepspeed/module_inject/auto_tp.py`, the `param_list` with `n_heads`/`embed_dim`/`hidden_size`/...) recursively halves dimension attributes on **every** module it walks, including modules whose weights AutoTP does **not** shard — the vision tower's `PatchEmbed` (a `Conv3d`, not covered by the linear replacement policies).

Result: `patch_embed.embed_dim` becomes 640 while the Conv3d weight keeps its full 1280 output channels, so `PatchEmbed.forward`'s `.view(-1, self.embed_dim)` silently reshapes `[N, 1280]` into `[2N, 640]`, and the very first vision-block `LayerNorm(normalized_shape=[1280])` fails:

```
RuntimeError: Given normalized_shape=[1280], expected input with shape [*, 1280], but got input of size [2048, 640]
```

This is a **single-model** issue: it reproduces with one multimodal model and `tensor_parallel.autotp_size=2` via `deepspeed.initialize`. It is unrelated to (and pre-dates) the multi-model globals fix in #8241 — verified identical on both `e79cdf2` (pre-fix) and `88acf89` (post-fix) of that branch.

## Repro (2 GPUs)

```python
import torch, deepspeed
import torch.distributed as tdist
from transformers import Qwen2VLForConditionalGeneration

tdist.init_process_group("nccl")
torch.cuda.set_device(tdist.get_rank())

m = Qwen2VLForConditionalGeneration.from_pretrained("Qwen/Qwen2-VL-2B-Instruct", torch_dtype=torch.bfloat16)
ds_config = {
    "tensor_parallel": {"autotp_size": 2},
    "bf16": {"enabled": True},
    "zero_optimization": {"stage": 0},
}
engine = deepspeed.initialize(model=m, config=ds_config)[0]
pe = engine.module.model.visual.patch_embed
print(f"embed_dim {1280} -> {pe.embed_dim}, proj out_ch={pe.proj.weight.shape[0]}")
# prints: embed_dim 1280 -> 640, proj out_ch=1280   <-- attribute halved, weight not
```

Then any forward pass with images (`model.generate(**processor(images=..., text=...))`) dies in `visual.blocks.0.norm1` with the LayerNorm shape error above.

Language-model-side AutoTP on the same model works correctly (q_proj/k_proj shards and per-model meta are right); only the vision tower is corrupted.

## Environment

- DeepSpeed: master-era branch of #8241 (`e79cdf2` and `88acf89`); behavior identical on both
- transformers 5.14.0.dev0, torch 2.12.1+cu130
- 2× RTX 5090, Qwen2-VL-2B-Instruct

## Suggested directions

- Only call `update_mp_params` for modules whose weights were actually sharded (or that sit under an AutoTP-replaced parent), or
- skip vision towers / non-replaced subtree roots (e.g. guard on `param.dim() >= 3` already exists for fused experts; Conv3d weights are 5D but `PatchEmbed` itself has no parameters so the current guard misses it), or
- match attributes to the module's actual (unsharded) weights before rewriting them.

Happy to turn the repro into a unit test if useful.

zupengwang:
## Summary

- defer module metadata updates until recursive AutoTP replacement has completed
- update logical dimensions only when a subtree contains parameters carrying AutoTP TP metadata
- preserve metadata for replicated vision modules and high-dimensional weights while retaining existing sharded Linear behavior
- add regression coverage for PatchEmbed-style Conv3d wrappers, unsharded attention subtrees, and genuinely sharded parameters

Fixes #8285.

## Problem

`update_mp_params()` previously ran before Linear replacement and scaled logical attributes for every traversed module. In multimodal models this changed replicated vision metadata, such as Qwen2-VL PatchEmbed `embed_dim`, even though the corresponding Conv3d weight was not partitioned. The metadata then disagreed with the full weight shape and broke vision forward.

## Validation

Current master baseline `e70b90352e93e07862c45ebb946d75b813f025a2` reproduces on 2-GPU NCCL with `patch_embed.embed_dim=60` while `proj.out_channels=120`.

- CPU targeted: `7 passed, 20 deselected`
- 2-process Gloo metadata check: unsharded `num_heads=16`, sharded `hidden_size=6` on both ranks
- 2-GPU custom-pattern suite: `27 passed`
- 4-GPU AutoTP/ZeRO adjacent suite: `33 passed, 2 skipped, 24 deselected`
- synthetic training on 1/2/3/5 RTX 3090 GPUs: forward, backward, and optimizer step all complete with zero cross-rank output, PatchEmbed gradient, and PatchEmbed parameter difference
- 5-GPU synthetic, 3 runs with 3 warmup and 20 measured steps each: `5.008 ms` mean step latency, `409,216 tokens/s`, `28,170,752` peak allocated bytes, `52,428,800` peak reserved bytes
- Qwen2-VL-2B AutoTP=2 vision forward: PatchEmbed `1280 -> 1280`, vision heads `16 -> 16`, language q_proj `[1536,1536] -> [768,1536]`, zero cross-rank output difference
- changed-file pre-commit hooks, `DS_BUILD_OPS=0 python setup.py check`, `compileall`, and `git diff --check` pass

### DIFF
--- a/deepspeed/module_inject/auto_tp.py
+++ b/deepspeed/module_inject/auto_tp.py
@@ -622,10 +622,10 @@ class AutoTP():
                 f"parameters, e.g. {registered[0]!r}",
                 ranks=[0])
 
-    def update_mp_params(self, child, name=None, only_if_sharded=False):
+    def update_mp_params(self, child, name=None):
         if getattr(child, "replaced", False) == True:
             return
-        if only_if_sharded and not any(hasattr(param, DS_AUTOTP_UC_META) for param in child.parameters()):
+        if not any(hasattr(param, DS_AUTOTP_UC_META) for param in child.parameters()):
             setattr(child, "replaced", True)
             return
         tp_index = dist.get_rank(group=self.mp_group) if self.mp_group is not None else 0
@@ -689,7 +689,7 @@ class AutoTP():
                 setattr(autoep_layer, child_name, self.linear_policies[key](child, full_name, self.conv_linear_layer))
             else:
                 self._replace_module(child, full_name, "")
-                self.update_mp_params(child, full_name, only_if_sharded=True)
+                self.update_mp_params(child, full_name)
 
     def _replace_module(self, r_module, prev_name='', prev_class_name=''):
         if prev_name == '' and prev_class_name == '':
@@ -738,7 +738,7 @@ class AutoTP():
                         setattr(r_module, name, new_child)
                 else:
                     self._replace_module(child, name, class_name)
-                    self.update_mp_params(child, full_name, only_if_sharded=True)
+                    self.update_mp_params(child, full_name)
             # Traditional path: use linear_policies for type-based routing
             elif child.__class__ in self.linear_policies:
                 setattr(r_module, name, self.linear_policies[child.__class__](child, prev_name + '.' + name,
@@ -758,7 +758,7 @@ class AutoTP():
                 self._replace_module(child, name, class_name)
                 # Descendants have now been replaced and carry universal-checkpoint TP metadata.
                 # Keep logical dimensions whole when no parameter in this subtree was actually sharded.
-                self.update_mp_params(child, name, only_if_sharded=True)
+                self.update_mp_params(child, name)
         return r_module
 
     @staticmethod
--- a/deepspeed/module_inject/auto_tp.py
+++ b/deepspeed/module_inject/auto_tp.py
@@ -622,15 +622,20 @@ class AutoTP():
                 f"parameters, e.g. {registered[0]!r}",
                 ranks=[0])
 
-    def update_mp_params(self, child, name=None):
+    def update_mp_params(self, child, name=None, only_if_sharded=False):
         if getattr(child, "replaced", False) == True:
             return
+        if only_if_sharded and not any(hasattr(param, DS_AUTOTP_UC_META) for param in child.parameters()):
+            setattr(child, "replaced", True)
+            return
         tp_index = dist.get_rank(group=self.mp_group) if self.mp_group is not None else 0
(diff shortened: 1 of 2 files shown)


deepspeed/module_inject/auto_tp.py
```diff
@@ -622,15 +622,20 @@ def hook(grad):
                 f"parameters, e.g. {registered[0]!r}",
                 ranks=[0])
 
-    def update_mp_params(self, child, name=None):
+    def update_mp_params(self, child, name=None, only_if_sharded=False):
```
delock: Does `only_if_sharded` switch needed if all call site pass `True`?  We could avoid this condition.

zupengwang: Good point. Since every call site requires the metadata guard, the switch was redundant. I removed `only_if_sharded` and made `update_mp_params()` always skip metadata updates when the subtree has no parameter carrying `DS_AUTOTP_UC_META`. The direct unit tests now mark their synthetic sharded parameters explicitly.

Validated on physical GPUs 4 and 5:
- `tests/unit/model_parallelism/test_autotp_custom_patterns.py`: 27 passed
- Qwen2-VL-2B AutoTP: vision output rank max absolute difference 0.0, PatchEmbed metadata preserved, and the language `q_proj` sharded as expected

Updated in `2cc5f6a7a0bc125850adef8848545e1eefd27e1d`.

delock (approved):

PR #8300, opened 2026-08-23
[Apple Silicon Support Phase 1] Add Metal FusedAdam kernel and CPU Adam build for Apple Silicon

PKUWZP:
## Summary

This PR implements the Phase 1 work of Apple Silicon support for DeepSpeed (follow-up to #8293, which made single-device ZeRO 1–3 training work with pure-PyTorch ops). This PR adds the **op-builder layer**: 

- Metal kernels compiled at runtime, 
- a first real Metal kernel (FusedAdam), 
- and the C++ CPU Adam build so ZeRO-Offload works on Macs.

### Changes

- **`op_builder/mps/builder.py`** — new `MetalOpBuilder`. Subclasses list `.metal` files in `metal_sources()`; the shader is compiled at `load()` through `torch.mps.compile_shader`, which dispatches kernels on PyTorch's own MPS command stream. No Xcode project, `.metallib` packaging, or C++ extension build is involved. `is_compatible()` additionally requires `torch.mps.compile_shader`.
- **`csrc/mps/fused_adam.metal` + `op_builder/mps/fused_adam.py`** — `FusedAdam` becomes a Metal kernel (one launch per tensor). It does all math in fp32 and stores in the parameter dtype, the same contract as `csrc/adam/multi_tensor_adam.cu`; this also closes the bf16 ulp gap noted in #8293. The `torch._foreach_*` implementation remains as a fallback for torch builds without `compile_shader` and for non-contiguous tensors.
- **`op_builder/mps/cpu_adam.py`** — builds `csrc/adam/cpu_adam*.cpp` with the system clang (`-D__SCALAR__` on arm64). Apple clang has no OpenMP, so the build uses Homebrew `libomp` when `brew --prefix libomp` resolves and omits it otherwise; both paths verified. This enables `DeepSpeedCPUAdam` and therefore ZeRO-Offload on Apple Silicon. Unified memory means offloading does not copy parameters between separate memories.
- **`tests/unit/ops/adam/test_adamw.py`** — `test_fused_adam_matches_reference` checks `FusedAdam` against an explicit fp32-math / storage-dtype-rounding reference for fp32, bf16, and fp16 × Adam/AdamW (replaces the `torch.optim` comparison from #8293, whose bf16 reference computes in bf16 and is a worse baseline). Tolerance is 8 ulp of the storage dtype at tensor scale, which covers measured fp32 op-order drift over 5 steps.
- **`tests/unit/ops/adam/test_cpu_adam.py`, `test_hybrid_adam.py`** — `py-cpuinfo` has no `vendor_id_raw` on Apple Silicon; use `.get()`.
- **`MANIFEST.in`** — ship `.metal` sources. **Docs** — accelerator setup guide updated for offload and the Metal/OpenMP notes.

### Verified on an M5 Max (macOS 26.3, torch 2.13.0)

- `DeepSpeedCPUAdam` matches `torch.optim` to ~1e-6; ZeRO-Offload trains end to end for stage 1/2/3 × fp32/bf16/fp16 (optimizer offload; plus param offload for stage 3).
- Metal `FusedAdam` vs foreach fallback, 4×100k params: 0.08 vs 0.27 ms/step (fp32), 0.03 vs 0.16 ms/step (bf16). Both implementations pass the new reference test in all 6 cases.
- `DS_ACCELERATOR=mps pytest unit/ops/adam/test_cpu_adam.py unit/ops/adam/test_hybrid_adam.py unit/ops/adam/test_adamw.py`: 74 passed, 7 skipped.
- `op_builder.mps` imports with torch absent (the sdist/install-smoke path).

### Follow-ups

- macOS arm64 CI workflow so these paths are exercised upstream.
- Further Metal kernels (quantizer for ZeRO++).

### DIFF
--- a/csrc/mps/fused_adam.metal
+++ b/csrc/mps/fused_adam.metal
@@ -49,7 +49,10 @@ template [[host_name("fused_adam_half")]] kernel void fused_adam<half>(
     device half*, device const half*, device half*, device half*,
     constant float&, constant float&, constant float&, constant float&, constant float&,
     constant float&, constant float&, constant float&, uint);
+// bfloat is a Metal 3.1 type (macOS 14); older toolchains still get the float and half kernels.
+#if __METAL_VERSION__ >= 310
 template [[host_name("fused_adam_bfloat")]] kernel void fused_adam<bfloat>(
     device bfloat*, device const bfloat*, device bfloat*, device bfloat*,
     constant float&, constant float&, constant float&, constant float&, constant float&,
     constant float&, constant float&, constant float&, uint);
+#endif
--- a/csrc/mps/fused_adam.metal
+++ b/csrc/mps/fused_adam.metal
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: Apache-2.0
+
+// DeepSpeed Team
+
+// Fused Adam / AdamW step for one tensor. Mirrors csrc/adam/multi_tensor_adam.cu: all math is done in
+// fp32 regardless of the storage dtype, so fp16/bf16 parameters see the same numerics as on CUDA.
+
+#include <metal_stdlib>
+using namespace metal;
+
+template <typename T>
+kernel void fused_adam(device T* param [[buffer(0)]],
+                       device const T* grad [[buffer(1)]],
+                       device T* exp_avg [[buffer(2)]],
+                       device T* exp_avg_sq [[buffer(3)]],
+                       constant float& lr [[buffer(4)]],
+                       constant float& beta1 [[buffer(5)]],
+                       constant float& beta2 [[buffer(6)]],
+                       constant float& epsilon [[buffer(7)]],
+                       constant float& weight_decay [[buffer(8)]],
+                       constant float& bias_correction1 [[buffer(9)]],
+                       constant float& bias_correction2 [[buffer(10)]],
+                       constant float& adam_w_mode [[buffer(11)]],
+                       uint i [[thread_position_in_grid]])
+{
+    float g = float(grad[i]);
+    float p = float(param[i]);
+    float m = float(exp_avg[i]);
+    float v = float(exp_avg_sq[i]);
+
+    // L2 mode folds weight decay into the gradient; AdamW mode applies it to the parameter.
+    if (adam_w_mode == 0.0f) { g += weight_decay * p; }
+    m = beta1 * m + (1.0f - beta1) * g;
+    v = beta2 * v + (1.0f - beta2) * g * g;
+    float denom = sqrt(v / bias_correction2) + epsilon;
+    float update = (m / bias_correction1) / denom;
+    if (adam_w_mode != 0.0f) { update += weight_decay * p; }
+
+    param[i] = T(p - lr * update);
+    exp_avg[i] = T(m);
+    exp_avg_sq[i] = T(v);
+}
+
+template [[host_name("fused_adam_float")]] kernel void fused_adam<float>(
+    device float*, device const float*, device float*, device float*,
+    constant float&, constant float&, constant float&, constant float&, constant float&,
+    constant float&, constant float&, constant float&, uint);
+template [[host_name("fused_adam_half")]] kernel void fused_adam<half>(
+    device half*, device const half*, device half*, device half*,
+    constant float&, constant float&, constant float&, constant float&, constant float&,
+    constant float&, constant float&, constant float&, uint);
+template [[host_name("fused_adam_bfloat")]] kernel void fused_adam<bfloat>(
+    device bfloat*, device const bfloat*, device bfloat*, device bfloat*,
+    constant float&, constant float&, constant float&, constant float&, constant float&,
+    constant float&, constant float&, constant float&, uint);
--- a/csrc/mps/fused_adam.metal
+++ b/csrc/mps/fused_adam.metal
@@ -20,7 +20,7 @@ kernel void fused_adam(device T* param [[buffer(0)]],
                        constant float& weight_decay [[buffer(8)]],
                        constant float& bias_correction1 [[buffer(9)]],
                        constant float& bias_correction2 [[buffer(10)]],
-                       constant float& adam_w_mode [[buffer(11)]],
+                       constant uint& adam_w_mode [[buffer(11)]],
                        uint i [[thread_position_in_grid]])
 {
     float g = float(grad[i]);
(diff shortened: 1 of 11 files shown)


csrc/mps/fused_adam.metal
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
PKUWZP: DCO/commit policy checks are passed. Resolving this comment.

csrc/mps/fused_adam.metal
```diff
@@ -0,0 +1,55 @@
...
+    constant float&, constant float&, constant float&, constant float&, constant float&,
+    constant float&, constant float&, constant float&, uint);
+template [[host_name("fused_adam_half")]] kernel void fused_adam<half>(
+    device half*, device const half*, device half*, device half*,
+    constant float&, constant float&, constant float&, constant float&, constant float&,
+    constant float&, constant float&, constant float&, uint);
+template [[host_name("fused_adam_bfloat")]] kernel void fused_adam<bfloat>(
+    device bfloat*, device const bfloat*, device bfloat*, device bfloat*,
+    constant float&, constant float&, constant float&, constant float&, constant float&,
+    constant float&, constant float&, constant float&, uint);
```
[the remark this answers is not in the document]
PKUWZP: Good catch. Fixed in d026abd36: the `bfloat` specialization is now guarded by `#if __METAL_VERSION__ >= 310`, `load()` only registers the bf16 entry point when the library exposes it (bf16 params use the foreach path otherwise), and a shader compile failure logs once and keeps the foreach fallback instead of raising. Verified all three scenarios against the reference test: normal, an emulated pre-3.1 toolchain (float/half kernels only), and a simulated compile failure.

tests/unit/ops/adam/test_adamw.py
```diff
@@ -80,32 +80,50 @@ def test(self,
...
         ds_optimizer.step()
 
-    # bf16 storage rounds differently depending on where intermediates are kept, so allow one ulp.
-    atol = 1e-5 if dtype == torch.float else 2e-2
-    for ref_param, ds_param in zip(ref_params, ds_params):
-        torch.testing.assert_close(ds_param.float(), ref_param.float(), atol=atol, rtol=0)
+    # fp32 operation order differs between implementations and accumulates over steps, so allow a
+    # few ulps of the storage dtype at the scale of the tensor (per-element rtol is too strict near zero).
+    for ds_param, ref_param in zip(ds_params, ref_params):
+        atol = 8 * torch.finfo(dtype).eps * ref_param.abs().max().item()
```
delock, line 130: Is this drift impact fp32 only?  Should we retain original atol for bf16?

delock: What is the new `atol` value compared to old `atol` value (1e-5)?  An example might help to see how much is relaxed here.

csrc/mps/fused_adam.metal
```diff
@@ -0,0 +1,58 @@
...
+                       constant float& adam_w_mode [[buffer(11)]],
+                       uint i [[thread_position_in_grid]])
+{
+    float g = float(grad[i]);
+    float p = float(param[i]);
+    float m = float(exp_avg[i]);
+    float v = float(exp_avg_sq[i]);
+
+    // L2 mode folds weight decay into the gradient; AdamW mode applies it to the parameter.
+    if (adam_w_mode == 0.0f) { g += weight_decay * p; }
```
delock: why adam_w_mode is a float?

docs/_tutorials/accelerator-setup-guide.md
```diff
@@ -306,7 +306,9 @@ Launch a single-process job as usual; no hostfile is needed:
 ```
 deepspeed --num_gpus 1 train.py --deepspeed --deepspeed_config ds_config.json
 ```
-ZeRO stages 0 through 3 are supported with fp32, fp16, and bf16 (bf16 requires macOS 14 or newer). The fused Adam optimizer runs as a PyTorch implementation on MPS; ZeRO-Offload (`DeepSpeedCPUAdam`) is not yet available on this backend.
+ZeRO stages 0 through 3 are supported with fp32, fp16, and bf16 (bf16 requires macOS 14 or newer), with or without ZeRO-Offload.
+
+The fused Adam optimizer is a Metal kernel compiled at first use through `torch.mps.compile_shader`; no Xcode project or C++ build is involved. ZeRO-Offload uses the C++ `DeepSpeedCPUAdam` kernel, which is built just-in-time with the system clang. Apple's clang has no OpenMP, so the kernel is single-threaded unless Homebrew's `libomp` is installed (`brew install libomp`), in which case it is picked up automatically. Because Apple Silicon has unified memory, offloading to the CPU optimizer does not copy parameters between separate memories.
```
delock: One question is for unified memory whether offloading to CPU optimizer is necessary.

op_builder/mps/builder.py
```diff
@@ -27,3 +25,30 @@ def include_paths(self):
     def is_compatible(self, verbose=False):
         import torch
         return hasattr(torch.backends, "mps") and torch.backends.mps.is_available()
+
+
+class MetalOpBuilder(MPSOpBuilder):
```
delock, line 30: is that true that ops on Metal are not compiled and saved on disk cache?  Is it temporary or its nature of Metal?

tests/unit/ops/adam/test_adamw.py
PKUWZP: The fp32 op-order drift affects every dtype (both implementations do their math in fp32), but for bf16/fp16 the dominant term in any honest bound is the storage rounding: one boundary flip moves an element by a full storage ulp, which dwarfs the drift. The old 2e-2 was calibrated against a different reference (torch.optim running bf16 math) that this PR replaces, so it isn't directly comparable — measured agreement against the new fp32-math reference is ~3e-5 for bf16, far inside the bound. I've added the concrete numbers to the comment in 34f7cb794; happy to tighten bf16/fp16 to fewer ulps if you'd prefer a snugger bound.

PKUWZP: Good idea — added to the code comment in 34f7cb794. Concretely, for this test's data (|param| ~ 3 after 5 steps): fp32 atol evaluates to ~3e-6 (tighter than the old 1e-5), bf16 to ~0.2, fp16 to ~2.5e-2, while the measured implementation-vs-reference differences are ~1e-6 (fp32) and ~3e-5 (bf16). So fp32 got stricter, and the loose-looking bf16 bound is headroom over a much smaller observed error.

csrc/mps/fused_adam.metal
PKUWZP: No good reason — at the time I hadn't verified that the shader binding marshals Python ints. It does: fixed in 34f7cb794, adam_w_mode is now constant uint& and the Python side passes int(adam_w_mode).

docs/_tutorials/accelerator-setup-guide.md
PKUWZP: Fair question — on unified memory offload is not about capacity at all (optimizer states occupy the same DRAM either way). What it still buys: Metal caps a process's GPU working set below total RAM (torch.mps.recommended_max_memory, ~75% here), and CPU-held optimizer state stays outside that budget; the step also runs on the CPU cores. So: unnecessary when the model fits the working-set budget, useful when that limit binds. Reworded the doc to say exactly this in 34f7cb794.

op_builder/mps/builder.py
PKUWZP: It's the nature of the Metal workflow rather than a temporary gap: torch.mps.compile_shader compiles these small kernels in milliseconds at first load, and macOS's Metal framework maintains its own per-app on-disk cache of compiled pipelines, so a torch-extensions style build cache would add complexity without saving anything. Documented in the MetalOpBuilder docstring in 34f7cb794.

delock (approved):

PR #8301, opened 2026-08-23
Fix nonblocking P2P staging on MPS

FU-max-boop:
## Summary

- route `deepspeed.comm.isend` and `irecv` through the backend's nonblocking methods
- treat point-to-point MPS staging as inherently asynchronous
- keep staged send buffers alive and copy received data back exactly once when the underlying `Work` completes
- preserve `Work` polling, result, future, and error semantics while restoring original tensor identities

This follows the unresolved MPS review note on #8293: https://github.com/deepspeedai/DeepSpeed/pull/8293#discussion_r3838066086. The existing wrapper only detects collectives with an `async_op` argument, but `isend` and `irecv` are asynchronous by contract and `irecv` has no such parameter. In addition, the public wrappers currently dispatch them to blocking backend methods.

## Validation

- `pytest tests/unit/comm/test_p2p.py` — 16 passed
- `pytest --forked tests/unit/comm/test_p2p.py` — 16 passed
- `pre-commit run --files deepspeed/comm/comm.py deepspeed/comm/torch.py tests/unit/comm/test_p2p.py`

The tests use deterministic fake MPS tensors and work handles to cover wait, polling, result and future completion, failure paths, no-grad copy-back, exactly-once behavior, and staging lifetime.

### DIFF
--- a/deepspeed/comm/comm.py
+++ b/deepspeed/comm/comm.py
@@ -384,13 +384,13 @@ def recv(tensor, src=None, group=None, tag=0, prof=False, log_name='recv', debug
 @timed_op
 def isend(tensor, dst, group=None, tag=0, prof=False, log_name='isend', debug=get_caller_func()):
     global cdb
-    return cdb.send(tensor=tensor, dst=dst, group=group, tag=tag)
+    return cdb.isend(tensor=tensor, dst=dst, group=group, tag=tag)
 
 
 @timed_op
 def irecv(tensor, src=None, group=None, tag=0, prof=False, log_name='irecv', debug=get_caller_func()):
     global cdb
-    return cdb.recv(tensor=tensor, src=src, group=group, tag=tag)
+    return cdb.irecv(tensor=tensor, src=src, group=group, tag=tag)
 
 
 @timed_op
--- a/deepspeed/comm/torch.py
+++ b/deepspeed/comm/torch.py
@@ -7,6 +7,8 @@ import deepspeed
 from deepspeed import utils
 from packaging import version
 import inspect
+from threading import Lock
+import weakref
 
 from .utils import *
 from .backend import *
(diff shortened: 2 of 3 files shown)


PKUWZP (changes requested):
Thanks so much for the PR, a couple of things here: 

- The `comm.py` rerouting changes `isend/irecv` semantics for every backend (NCCL included) — it revives previously-dead async paths in the 1-bit comm modules. It's a global behavior change riding in an "MPS staging" PR; it should be called out prominently and watched on the **CUDA CI, not just macOS**.

- `StagedWork` grew locks, a weakref identity map, future wrapping, and `__getattr__` delegation (~100 lines) to preserve `result()/get_future_result()` semantics that no current DeepSpeed caller uses. I feel it add unnecessary complexity — callers only use `.wait()` and occasionally `is_completed()`. Can we drop the future/result machinery until something needs it.

PKUWZP:
@FU-max-boop Also I found that tests are all based on FakeTensor, FakeWork, monkeypatched staging predicate etc. — they elaborate implementation details (copy counts, weakref retention) rather than behavior, and none of them tests real torch.distributed. A 2-rank gloo test with real Work objects (like the one I ran) would be stronger and could run on any CI, not just macOS.

FU-max-boop:
Thanks @PKUWZP for the detailed review and for carrying the narrower implementation in #8303. I agree with both review points: #8301's locks/weakrefs/future/result machinery is larger than the current call surface justifies, and its fake-heavy tests do not replace a real two-rank `torch.distributed` regression.

#8303 now carries the core routing change, focused MPS staging coverage, and a real two-rank Gloo path, including the CPU skip fix in `1504ea6`. To avoid maintaining two competing implementations, I'm closing #8301 as superseded by the active #8303.

This is coordination, not a claim that #8303 has merged. I'll leave `fix/nonblocking-p2p-staging` exactly as-is: no branch deletion and no force-push, so the work remains available for reference or reopening if needed. I'll take only the remaining minimal Work-contract questions into a separate maintainer-agreed follow-up after #8303 settles.

Thanks again for the concrete feedback.

PR #8304, opened 2026-08-24
Honor --include/--exclude in the SLURM launcher

vineethsaivs:
## Problem

`SlurmRunner.get_cmd()` forwards DeepSpeed's own `--include` / `--exclude` strings to `srun` and sizes the job from the unfiltered hostfile. With a two-node, four-slot hostfile:

```
deepspeed --launcher slurm --include worker-1:0,2 train.py
```

builds

```
srun -n 8 --include worker-1:0,2 --export=ALL python -u train.py
```

Two things are wrong with that line:

1. `srun` has no `--include` option, so it exits with `unrecognized option '--include'` and the job never starts.
2. `-n 8` counts every slot in the pool. The user asked for 2.

`--exclude` is broken the same way. `srun` does accept that flag, but DeepSpeed's `NAME[:SLOT[,SLOT ...]]` syntax is not a slurm hostlist (`--exclude worker-1:0` names no node slurm knows), and `-n` still ignores the filter, so `--exclude worker-1` asks for 8 tasks on the 4 slots that are left.

## Fix

`runner.py` already resolves both flags into `active_resources` before calling `get_cmd()`, and it passes that dict in as the second argument. The other runners either use it (`PDSHRunner`) or explicitly reject the flags in `validate_args()`; `SlurmRunner` was the only one accepting them and then ignoring the resolved result.

So take the process count and the host list from `active_resources`, and give `srun` `--nodelist`, which is the flag it actually has:

```
srun -n 2 --nodelist worker-1 --nodes 1 --export=ALL python -u train.py
```

`--nodelist` alone is only an upper bound. SchedMD documents that "a lower node or processor count may only require a subset of the supplied node list", so `-n 7 --nodelist worker-0,worker-1` could put all seven ranks on `worker-0` and silently drop a host the filter kept. `--nodes` pins the count (a single number is used as both minimum and maximum). It cannot collide with the `--nodes` the branch below emits, because `runner.py` raises `Cannot specify num_nodes/gpus with include/exclude`, so `args.num_nodes` is always `-1` on this path.

Slot-level selection still cannot be expressed to `srun`, but the task count and the node set are now what the user asked for, instead of an invalid command line.

## Wider scope, stated rather than left implicit

Sizing from `active_resources` is unconditional, so it also moves `-n` for `--num_gpus` and `--num_nodes`, which `runner.py` trims the same dict for (L521 and L529):

```
                           base    head
control (no flags)         -n 8    -n 8
--include worker-1:0,2     -n 8    -n 2
--exclude worker-1:0       -n 8    -n 7
--num_gpus 2               -n 8    -n 4
--num_nodes 1              -n 8    -n 4
```

The bottom two are corrections rather than regressions: `world_info_base64` is encoded from that same trimmed dict at L543, so `-n 8` against a four-rank world info was already inconsistent. Both are now covered by tests.

To pin them without a hand-rolled copy of `main()`'s trim in the test, those lines move verbatim into `apply_num_nodes_and_gpus()` next to `parse_inclusion_exclusion`, and `main()` calls it. Pure extraction, no behaviour change.

## Test

Added `test_slurm_runner_resource_filter` to `tests/unit/launcher/test_multinode_runner.py`, covering `--include` with slots, `--exclude` with slots, and `--exclude` of a whole host, asserting the host list, the node count and the task count for each. `test_slurm_runner` also now asserts the unfiltered count and passes `get_cmd()` the slot lists that `runner.py` really hands it, rather than the raw hostfile counts.

`test_slurm_runner_num_nodes_and_gpus` covers the two flags above, asserting `-n 4` for each plus that the resource-filter branch stays silent, so `--nodelist` is absent and `--nodes` appears exactly once.

```
# before the fix, with the new tests
5 failed, 7 passed

# after
12 passed
```

The rest of `tests/unit/launcher/` is unchanged at `9 failed, 23 passed` on a clean tree and `9 failed, 28 passed` here; the 9 are pre-existing `test_user_args.py` failures in this environment because the `deepspeed` console script is not installed, and they reproduce identically on a clean tree.

`yapf --style .style.yapf -d` and `flake8 --config .flake8` are clean on both changed files.

### DIFF
--- a/deepspeed/launcher/multinode_runner.py
+++ b/deepspeed/launcher/multinode_runner.py
@@ -355,11 +355,31 @@ class SlurmRunner(MultiNodeRunner):
     def name(self):
         return 'slurm'
 
+    @staticmethod
+    def _assert_expressible_in_srun(active_resources):
+        """srun places tasks by count, not by device id.
+
+        It can run N tasks on a named set of hosts, but it cannot pin them to
+        particular slots, and -n gives it no way to put fewer tasks on one host
+        than another. Fail here rather than launch a job that ignores the filter.
+        """
+        slot_counts = set()
+        for hostname, slots in active_resources.items():
+            if list(slots) != list(range(len(slots))):
+                raise ValueError(f"slurm backend cannot select specific device ids, got "
+                                 f"{list(slots)} on {hostname}. Filter whole hosts, or keep the "
+                                 f"first N slots on every host.")
+            slot_counts.add(len(slots))
+        if len(slot_counts) > 1:
+            counts = {hostname: len(slots) for hostname, slots in active_resources.items()}
+            raise ValueError(f"slurm backend needs the same slot count on every host, got {counts}.")
+
     def get_cmd(self, environment, active_resources):
         assert not getattr(self.args, 'detect_nvlink_pairs',
                            False), "slurm backend does not support remapping visible devices"
+        self._assert_expressible_in_srun(active_resources)
         # --include/--exclude are already resolved into active_resources, so counting the
-        # whole pool here would ask srun for slots the user filtered out.
+        # whole pool would ask srun for slots the user filtered out.
         total_process_count = sum(len(slots) for slots in active_resources.values())
         srun_cmd = [
             'srun',
@@ -371,23 +391,19 @@ class SlurmRunner(MultiNodeRunner):
             srun_cmd += ['--comment', self.args.slurm_comment]
 
         if self.args.include != "" or self.args.exclude != "":
-            # srun has no --include, and the NAME[:SLOT,...] syntax DeepSpeed accepts is not
-            # a slurm hostlist, so name the hosts that survived the filter instead.
-            active_hosts = ",".join(active_resources.keys())
+            # srun has no --include, and DeepSpeed's NAME[:SLOT,...] syntax is not a slurm
+            # hostlist, so name the hosts that survived the filter.
             srun_cmd.append('--nodelist')
-            srun_cmd.append(active_hosts)
-            # --nodelist alone is only an upper bound: srun documents that a lower task count
-            # "may only require a subset of the supplied node list", so it could pack every
-            # rank onto one host and silently drop a host the filter kept. --nodes pins the
-            # count, and runner.py forbids --num_nodes alongside a resource filter, so the
-            # branch below cannot also set it.
+            srun_cmd.append(",".join(active_resources.keys()))
+            # --nodelist is only an upper bound: srun may satisfy -n from a subset of it.
             srun_cmd.append('--nodes')
             srun_cmd.append(f'{len(active_resources)}')
         if self.args.num_nodes > 0:
             srun_cmd.append('--nodes')
             srun_cmd.append(f'{self.args.num_nodes}')
         if self.args.num_gpus > 0:
-            srun_cmd.append('--gpus')
+            # --num_gpus is per node, so --gpus (a job total) would under-request.
+            srun_cmd.append('--gpus-per-node')
             srun_cmd.append(f'{self.args.num_gpus}')
 
         exports = '--export=ALL'
--- a/deepspeed/launcher/multinode_runner.py
+++ b/deepspeed/launcher/multinode_runner.py
@@ -358,7 +358,9 @@ class SlurmRunner(MultiNodeRunner):
     def get_cmd(self, environment, active_resources):
         assert not getattr(self.args, 'detect_nvlink_pairs',
                            False), "slurm backend does not support remapping visible devices"
-        total_process_count = sum(self.resource_pool.values())
+        # --include/--exclude are already resolved into active_resources, so counting the
+        # whole pool here would ask srun for slots the user filtered out.
+        total_process_count = sum(len(slots) for slots in active_resources.values())
         srun_cmd = [
             'srun',
             '-n',
@@ -368,12 +370,19 @@ class SlurmRunner(MultiNodeRunner):
         if getattr(self.args, 'slurm_comment', ''):
             srun_cmd += ['--comment', self.args.slurm_comment]
 
-        if self.args.include != "":
-            srun_cmd.append('--include')
-            srun_cmd.append(f'{self.args.include}')
-        if self.args.exclude != "":
-            srun_cmd.append('--exclude')
-            srun_cmd.append(f'{self.args.exclude}')
+        if self.args.include != "" or self.args.exclude != "":
+            # srun has no --include, and the NAME[:SLOT,...] syntax DeepSpeed accepts is not
+            # a slurm hostlist, so name the hosts that survived the filter instead.
+            active_hosts = ",".join(active_resources.keys())
+            srun_cmd.append('--nodelist')
+            srun_cmd.append(active_hosts)
+            # --nodelist alone is only an upper bound: srun documents that a lower task count
+            # "may only require a subset of the supplied node list", so it could pack every
+            # rank onto one host and silently drop a host the filter kept. --nodes pins the
+            # count, and runner.py forbids --num_nodes alongside a resource filter, so the
+            # branch below cannot also set it.
+            srun_cmd.append('--nodes')
+            srun_cmd.append(f'{len(active_resources)}')
         if self.args.num_nodes > 0:
             srun_cmd.append('--nodes')
             srun_cmd.append(f'{self.args.num_nodes}')
--- a/deepspeed/launcher/multinode_runner.py
+++ b/deepspeed/launcher/multinode_runner.py
@@ -51,6 +51,16 @@ class MultiNodeRunner(ABC):
     def validate_args(self):
         """Validate self.args"""
 
+    @classmethod
+    def validate_active_resources(cls, active_resources):
+        """Check a resolved --include/--exclude filter against what this backend can express.
+
+        runner.main() calls this before it picks between a local and a multi-node launch. A
+        filter can narrow the pool to one host, which takes the local path and never builds a
+        backend, so a backend that cannot honor the filter has to say so from here or the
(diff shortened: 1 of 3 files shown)


vineethsaivs:
Good catch, and it is right. `--nodelist` is only an upper bound: srun's documentation says "a lower node or processor count may only require a subset of the supplied node list", so `--exclude worker-1:0` producing `srun -n 7 --nodelist worker-0,worker-1` could put all seven ranks on `worker-0` and silently drop a host the filter kept, which is the same class of bug this PR is fixing.

Fixed in `8fa2478` by also passing `--nodes` with the count of filtered hosts. A single number there is used as both the minimum and the maximum, so the host count is pinned rather than merely permitted:

```
--include worker-1:0,2   ->  srun -n 2 --nodelist worker-1 --nodes 1 ...
--exclude worker-1:0     ->  srun -n 7 --nodelist worker-0,worker-1 --nodes 2 ...
--exclude worker-1       ->  srun -n 4 --nodelist worker-0 --nodes 1 ...
```

It cannot collide with the `--nodes` the `args.num_nodes` branch below emits: `runner.py` raises `Cannot specify num_nodes/gpus with include/exclude`, so `num_nodes` is always `-1` whenever this branch runs.

The three parametrized cases now assert the node count alongside the host list and the task count: 3 failed / 7 passed against the unpatched source, 10 passed after. The unfiltered path is byte-for-byte unchanged and still emits no `--nodelist` or `--nodes` at all.

deepspeed/launcher/multinode_runner.py
```diff
@@ -358,7 +358,9 @@ def name(self):
     def get_cmd(self, environment, active_resources):
         assert not getattr(self.args, 'detect_nvlink_pairs',
                            False), "slurm backend does not support remapping visible devices"
-        total_process_count = sum(self.resource_pool.values())
+        # --include/--exclude are already resolved into active_resources, so counting the
+        # whole pool here would ask srun for slots the user filtered out.
+        total_process_count = sum(len(slots) for slots in active_resources.values())
```
ebarkhordar, line 383: I ran this against 8fa24782 in a clean container. The `--include` path does what the description says. One thing the description does not name: this line is unconditional, so it also changes `-n` for `--num_gpus` and `--num_nodes`.

`runner.py` trims `active_resources` for those two flags as well (`--num_nodes` at runner.py:521, `--num_gpus` at runner.py:529), and `get_cmd` now sizes the job from `active_resources` instead of `self.resource_pool`. Building the srun command at this SHA and at the merge base aa3914df, on the same two-node four-slot pool the launcher tests use:

```
                           base    head
control (no flags)         -n 8    -n 8
--include worker-1:0,2     -n 8    -n 2
--num_gpus 2               -n 8    -n 4
--num_nodes 1              -n 8    -n 4
```

The bottom two look correct to me rather than wrong: `world_info_base64` is encoded from the same trimmed `active_resources` at runner.py:543, so `-n 8` against a 4 rank world info was already inconsistent. My point is only that they are unpinned. `test_slurm_runner_resource_filter` parametrizes include and exclude, and `test_slurm_runner` asserts the unfiltered 8, so nothing covers the two flags that now also move.

Two more cases in that parametrize would cover it, though they need main()'s trim replicated the way `parse_inclusion_exclusion` already is, since it happens after it. Naming the wider scope in the description would work too.

How I got the numbers: I called `parse_inclusion_exclusion` and then replicated runner.py:521-533 in a probe, then called `SlurmRunner.get_cmd`. I did not run `main()`, and I have no slurm cluster here, so the `--nodelist` and `--nodes` semantics in your description are the only part I did not check.

vineethsaivs:
Thanks, this is right and I have covered it in `9a3bf84`.

You are correct that the line is unconditional and that `--num_gpus` / `--num_nodes` move with it, and correct that the new values are the consistent ones: `world_info_base64` is encoded from the same trimmed dict at runner.py:543, so `-n 8` against a 4-rank world info was the inconsistency. I reproduced your table exactly on the two-node four-slot pool:

```
                           base    head
control (no flags)         -n 8    -n 8
--include worker-1:0,2     -n 8    -n 2
--exclude worker-1:0       -n 8    -n 7
--num_gpus 2               -n 8    -n 4
--num_nodes 1              -n 8    -n 4
```

Did both things you offered rather than picking one.

**Pinned.** New `test_slurm_runner_num_nodes_and_gpus` parametrizes `--num_gpus 2` and `--num_nodes 1`, asserting `-n 4` for each plus that the resource-filter branch stays silent, so `--nodelist` is absent and `--nodes` appears exactly once (the `--num_nodes` case is the one that could have grown a second one). Against the unpatched `multinode_runner.py` both fail with `assert '8' == '4'`, which is the gap you found.

**Not replicated in the test.** You are right that replicating main()'s trim in a test would drift. `main()` lines 521-533 move verbatim into `apply_num_nodes_and_gpus()` next to `parse_inclusion_exclusion`, and `main()` calls it; no behaviour change, and the test now builds the same dict main() hands the backends instead of a hand-rolled copy.

**Named in the description.** The PR body and commit message now say that sizing from `active_resources` moves `-n` for those two flags too, and why that is the consistent direction.

Whole file: 5 failed / 7 passed before, 12 passed after. Whole `tests/unit/launcher/`: 9 failed / 23 passed on a clean tree, 9 failed / 28 passed here, the 9 being the pre-existing `test_user_args.py` failures where the `deepspeed` console script is not installed.

On the part you flagged as unchecked: I have no slurm cluster either, so the `--nodelist` / `--nodes` semantics come from the srun manual rather than a run. The specific sentence the `--nodes` addition rests on is "a lower node or processor count may only require a subset of the supplied node list", which is why `--nodelist` alone is not enough, and "if only one number is specified, this is used as both the minimum and maximum node count" for pinning with a bare `--nodes N`.

ebarkhordar:
That covers it. Moving the trim into `apply_num_nodes_and_gpus()` and calling it from `main()` is the better of the two options, since the test now exercises the same dict the backends get rather than a copy that can drift, and `test_slurm_runner_num_nodes_and_gpus` pins the two cases that were unpinned. Nothing further from me.

sfc-gh-truwase (approved):

tohtana (changes requested):
Thank you for the improvement! The overall direction seems good to me.

However, I found the following issues:
- The GPU indices aren't properly used. `--include worker-1:0,2` will be interpreted as "2 processes on the node". The same goes for `--exclude worker-1:0`. This doesn't pass how many processes should be created on each node.
- `--num_gpus` is passed to slurm but the semantics is different. For example, when setting `--num_gpus 2` for two nodes, it will request 2 GPUs for 4 processes.

I don't think we need to handle all these cases, but it would be better to raise an error when the given setting is not supported.

vineethsaivs:
Both points fixed in 98465c8.

**Device ids and uneven counts now raise** instead of launching a job that ignores the filter. `srun` places tasks by count, so it cannot pin them to slots or give one host fewer than another. `SlurmRunner._assert_expressible_in_srun` rejects any host whose surviving slots are not `0..n-1`, and any run where hosts differ in slot count. So `--include worker-1:0,2` and `--exclude worker-1:0` now error; whole-host filters and an equal prefix on every host still work.

**`--num_gpus` is now `--gpus-per-node`, not `--gpus`.** `--gpus` is a job total, which is exactly the 2-for-4 under-request you described.

`pytest tests/unit/launcher/test_multinode_runner.py -k slurm`: 8 passed, including the 3 new rejection cases.

deepspeed/launcher/multinode_runner.py
vineethsaivs: Right, and covered now: `test_slurm_runner_num_nodes_and_gpus` parametrizes `--num_gpus 2` and `--num_nodes 1` and pins `-n 4` for both, using `apply_num_nodes_and_gpus` (split out of `main()`) so the trim happens the same way.

`--num_gpus` also changed in 98465c8: it now maps to `--gpus-per-node`, not `--gpus`. `--gpus` is a job total, so the old form asked for 2 GPUs to run 4 tasks.

tohtana:
Thank you for the update, @vineethsaivs! The backend fixes address most of my previous feedback. However, the `--include worker-1:0,2` case I raised still bypasses the new validation: without `--force_multi`, `runner.main()` falls back to local launch instead of invoking `srun`.
Please handle this before the local fallback and add a `runner.main` regression test for the command without `--force_multi`.

vineethsaivs:
Fixed in 6234d29, thanks for pinning that down.

You were right, and it is worse than the check being skipped. Without `--force_multi`, `--include worker-1:0,2` leaves one host, so `main()` takes the local path and actually launches:

```
python -u -m deepspeed.launcher.launch --world_info=eyJ3b3JrZXItMSI6IFswLCAyXX0= ... train.py
```

The job runs on the submit node with `{"worker-1": [0, 2]}`, and `--launcher slurm` is dropped without a word.

What I changed: `MultiNodeRunner` gets a `validate_active_resources()` hook, a no-op by default. `main()` calls it on the requested launcher right after the filter is resolved and before the local-vs-multi decision, so both paths report the same error. `SlurmRunner`'s existing check became that hook, so `get_cmd()` still runs it on the multi-node path.

Tests, as asked: `test_runner_main_rejects_unsupported_filter` drives `runner.main()` with and without `--force_multi`. It also stubs `subprocess.Popen`, so if the guard ever regresses the test fails the assertion instead of launching a real job. `test_validate_active_resources_default_is_a_no_op` pins that the other five backends still accept any filter.

One thing worth your call: this rejects the command rather than sending it to srun. Making `--launcher slurm` plus a filter naming a remote host imply multi-node would also fix it, but that changes what `--force_multi` is for, so I did not assume it. Say the word if you would rather have that.

yapf 0.40.0 and flake8 with the repo config are clean. The full `tests/unit/launcher/test_multinode_runner.py` is 17 passed locally.

tohtana (approved):

PR #8306, opened 2026-08-24
Recognize Qwen3.5's RMSNorm variants in AutoTP module loading

### LINKED ISSUE #7947: Has DeepSpeed supported Qwen 3.5 yet?

promptsmith1990:
Fixes #7947.

## Problem

`Loading.is_load_module()` in `auto_tp.py` gates whether a leaf module's parameters get loaded from a raw state dict during AutoTP-based checkpoint loading (both call sites — `auto_tp.py` and `replace_module.py` — use it the same way). It matches by an exact class-name allowlist that gets a new entry each time a model family ships its own RMSNorm class, but Qwen3.5's classes were never added.

`Qwen3_5RMSNorm`, `Qwen3_5RMSNormGated` (dense), `Qwen3_5MoeRMSNorm` and `Qwen3_5MoeRMSNormGated` (MoE) each own a `weight` `nn.Parameter`, same shape as every other listed `*RMSNorm` class — without an allowlist entry, `is_load_module()` returns `False` for them and that weight is never loaded from the checkpoint, silently left at its random init value instead.

## Verification

Verified against the real modeling code (transformers 5.3.0, the version reported in the issue): built a small `Qwen3_5TextModel` and confirmed `Qwen3_5RMSNorm`/`Qwen3_5RMSNormGated` instances only carry a `weight` parameter, matching the pattern of the already-listed `LlamaRMSNorm` et al. Confirmed the MoE variants carry the same single `weight` parameter.

**Deliberately not adding `Qwen3_5TextRotaryEmbedding`**: it carries no parameters, only two buffers (`inv_freq`, `original_inv_freq`), and both call sites already load any child's buffers unconditionally (`if len(child._buffers) != 0`) regardless of `is_load_module()` — so listing it would be a no-op, not a functional fix. (Happy to add it anyway for consistency with the codebase's existing `Phi3RotaryEmbedding`/`YuanRotaryEmbedding` entries if maintainers prefer explicitness over strict minimality here — wasn't sure which this repo's convention favors.)

## Test

`tests/unit/module_inject/test_auto_tp_is_load_module.py` (new). Uses name-matched dummy `nn.Module` stand-ins rather than real `transformers` classes, since `is_load_module()` matches by class name only and the dev requirement (`transformers>=4.51.3`) predates Qwen3.5 — importing the real classes would make the test depend on a newer transformers than the repo's own pinned minimum. Covers all four new allowlist entries plus one negative case for an unrelated class name.

```
$ pytest tests/unit/module_inject/test_auto_tp_is_load_module.py -v
...
5 passed in 1.70s
```

`pre-commit run --files deepspeed/module_inject/auto_tp.py tests/unit/module_inject/test_auto_tp_is_load_module.py` (yapf, flake8, check-license, codespell) is clean.

---
Prepared with AI assistance under my review; the root cause, the parameter-vs-buffer distinction behind the RotaryEmbedding decision, and the test were verified locally before opening this PR.


delock:
Hi @promptsmith1990 , thanks for your fix.  Does this PR fix issue #7947, so that finetuning Qwen 3.5 family with AutoTP is possible, or it just fix the stated Qwen3.5 RMSNorm variants issue?  Thanks!

tests/unit/module_inject/test_auto_tp_is_load_module.py
```diff
@@ -0,0 +1,31 @@
...
+# Matches Loading.is_load_module() by class name only, so a lightweight stand-in
+# with the right name exercises the same allowlist check real Qwen3.5 modules
+# hit, without depending on a transformers version new enough to ship Qwen3.5.
+def _make_module(class_name):
+    return type(class_name, (nn.Module, ), {})()
+
+
+class TestIsLoadModule:
+
+    @pytest.mark.parametrize("class_name", [
```
delock: I think we should either remove this test or verify all elements in `is_load_module`.  Given that it is unlikely that an item be taken out of `is_load_module` with out causing attention, I suggest not having this test at all.

delock:
I independently reproduced this end-to-end (4×RTX 4080, torch 2.12.1+cu130, transformers 5.15.0.dev0) and can confirm both the bug on `master` and that this PR fixes it.

**Method** — tiny real `Qwen3_5ForCausalLM` (2 layers: `linear_attention` + `full_attention`, so both `Qwen3_5RMSNorm` and `Qwen3_5RMSNormGated` are exercised): model A (seed=1) provides the raw `state_dict` ("checkpoint"); model B is built under `init_empty_weights()` and passed through `AutoTP._replace_module(B.model, state_dict=sd, prefix="model")`. Afterwards each norm instance is checked for (a) no longer being meta and (b) matching the checkpoint. `q_proj`/`up_proj` serve as a control group to prove the state_dict path itself works.

**Results**

| run | build | world | control linear | qwen3_5 norms | verdict |
|---|---|---|---|---|---|
| CPU | master (`edaa72212`) | 1 | loaded | 0/8, all `still_meta` | BUG_REPRODUCED |
| GPU | master (`9311fd5`, latest) | 1 | loaded | 0/8, all `still_meta` | BUG_REPRODUCED |
| GPU | master (`9311fd5`) | 2 (torchrun) | loaded (slices match) | 0/8, all `still_meta` | BUG_REPRODUCED |
| CPU | master + this PR | 1 | loaded | 8/8 match checkpoint | FIXED |
| GPU | master + this PR | 1 | loaded | 8/8 match checkpoint | FIXED |
| GPU | master + this PR | 2 (torchrun) | loaded | 8/8 match checkpoint | FIXED |

Broken on master: `model.norm`, both layers' `input_layernorm`/`post_attention_layernorm`, `self_attn.{q,k}_norm` (`Qwen3_5RMSNorm`) and `linear_attn.norm` (`Qwen3_5RMSNormGated`). Norms are replicated (not sliced) under TP, so world=2 behaves identically per rank.

**Two findings worth recording** (from setting up the repro):

1. `Loading.load()` only fills weights that are still on the meta device (`if module.weight.data.is_meta:` with no else branch). So on `master` the skipped norms don't get random values — they **stay on meta device**, which fails at the first forward pass. That matches an "throws an error" report better than a silent numeric drift, which makes the link to #7947 more plausible than it looked. It also means this gate only bites workflows that use meta/empty init (e.g. fast checkpoint loading); a normally-initialized model loads nothing through this path at all.
2. `Loading.load()` does exact key lookup (`state_dict[prefix + 'weight']`) while the outer gate at both call sites is a substring match (`any(checking_key in item ...)`), so `prefix` must line up exactly with the state_dict keys — a misaligned prefix is silently masked by the outer gate. Not something this PR needs to address, just a trap I hit while reproducing.

Note: the e2e above covers the dense classes; the two MoE entries (`Qwen3_5MoeRMSNorm`/`Qwen3_5MoeRMSNormGated`) were verified by source inspection only (same single `weight` parameter, same shape pattern as the other listed `*RMSNorm` classes).

The four allowlist entries change exactly this behavior and nothing else in my runs — thanks for the fix.

tohtana (changes requested):
Please use the new copyright header showing DeepSpeed team for new files.

```

# DeepSpeed Team
```

delock:
@promptsmith1990 any updates?  Thanks!

promptsmith1990:
> @promptsmith1990 any updates? Thanks!

Hi! Thanks for your review, really appreciate it. Working on your comments.

tests/unit/module_inject/test_auto_tp_is_load_module.py
promptsmith1990: Removed the test file entirely per your suggestion — it only covered the four new Qwen3.5 entries rather than the full allow-list, and as you noted, removals from `is_load_module` would be obvious in code review. The functional change in `auto_tp.py` stands unchanged.

promptsmith1990:
Thank you for your review! **Reviewer feedback addressed:**

- **@delock**: Removed the test file entirely — it only covered the 4 new Qwen3.5 entries rather than the full allow-list, making it misleadingly partial. As you noted, any future removal from `is_load_module` would be caught in code review. The functional fix in `auto_tp.py` is unchanged.

- **@tohtana**: The new file with the old Microsoft copyright header is now gone, so no new files with incorrect headers remain. The modified `auto_tp.py` is an existing file with its existing header.

- **Signed-off-by** trailer included in the new commit.

promptsmith1990:
> Hi @promptsmith1990 , thanks for your fix. Does this PR fix issue #7947, so that finetuning Qwen 3.5 family with AutoTP is possible, or it just fix the stated Qwen3.5 RMSNorm variants issue? Thanks!

Hi @thuanaislab — good question. This PR specifically fixes Loading.is_load_module() in auto_tp.py, which controls whether a module's parameters get loaded from the checkpoint during AutoTP-based model loading. Without this fix, Qwen3.5's RMSNorm layers (Qwen3_5RMSNorm, Qwen3_5RMSNormGated, Qwen3_5MoeRMSNorm, Qwen3_5MoeRMSNormGated) silently start with randomly initialized weights instead of checkpoint values — which is the root cause of the failure reported in https://github.com/deepspeedai/DeepSpeed/issues/7947 when using transformers ≥ 5.3.0.

So yes, it addresses the is_load_module gap that was blocking Qwen3.5 + AutoTP.

tohtana (approved):
Thank you for the update, @promptsmith1990! Looks good to me.

PR #8307, opened 2026-08-24
Honor adam_w_mode in the CPU multi_tensor_adam binding

PKUWZP:
## Summary

The CPU `fused_adam` extension created its `Adam_Optimizer` once with default arguments and ignored the `mode` parameter entirely (`csrc/cpu/adam/fused_adam.cpp`), so `FusedAdam(adam_w_mode=False)` on the CPU backend always applied decoupled (AdamW) weight decay instead of L2. Everything else (`lr`, betas, `eps`, `weight_decay`, bias correction) is already passed per call via `ds_adam_step`; only the AdamW-vs-L2 flag is fixed at construction. The fix keeps one optimizer instance per mode (`mode 1 == AdamW`, matching the CUDA kernel's `ADAM_MODE_1`).

Also trims `test_fused_adam_matches_torch` to fp32: its bf16 cases compared against `torch.optim` running bf16 math, while the fused kernels compute in fp32 — never a valid reference. Low-precision dtypes get an explicit fp32-math reference test in the FusedAdam rework (#8300).

## How this surfaced

Split out of #8303 at @delock's request: after a master merge, cpu-torch-latest failed on `test_fused_adam_matches_torch[fp32-adam]` (98.7% of elements mismatched — systematic, not tolerance noise), and the investigation traced it to this binding. The fix was verified green on cpu-torch-latest in #8303's CI (run 32695...) before being extracted here.

## Validation

- `test_fused_adam_matches_torch[fp32-adam]` / `[fp32-adamw]` now genuinely exercise both decay modes against `torch.optim.Adam` / `AdamW` on the active accelerator.
- cpu-torch-latest passed with this exact change as part of #8303's branch; this PR carries it alone.


csrc/cpu/adam/fused_adam.cpp
```diff
@@ -19,13 +19,16 @@ void multi_tensor_adam(int chunk_size,
                        const int bias_correction,
                        const float weight_decay)
 {
-    static bool initialized = false;
-    if (!initialized) {
-        create_adam_optimizer(0);
-        initialized = true;
+    // ds_adam_step reads lr/betas/eps/weight_decay per call; only AdamW-vs-L2 is fixed at
+    // construction, so keep one optimizer instance per mode (mode 1 == AdamW, as in CUDA).
+    static bool initialized[2] = {false, false};
```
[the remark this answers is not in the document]
PKUWZP: We have sign-off trailer and the DCO is passed.

FU-max-boop:
CI triage note: modal-torch-latest tested the exact head 1e17e84f2b3f0918553404cad692b11f78ba7c0f. Because csrc/** is a run-all trigger, it ran the full tests/unit/v1 suite (1232 items), not the changed CPU Adam test. The run reached 97% without an Adam-related failure, then exited 137 almost exactly at the controller's 3600-second Modal Sandbox lifetime; the unrelated test_offload_activation case was marked failed only after its xdist worker was forcibly terminated.

The same 1232-item full suite passed on the base SHA in 2479.20 seconds ([run 32692693729](https://github.com/deepspeedai/DeepSpeed/actions/runs/32692693729/job/97329130272)), and cpu-torch-latest passed on this exact head ([job 97354322902](https://github.com/deepspeedai/DeepSpeed/actions/runs/32701660791/job/97354322902)). The failed [Modal job](https://github.com/deepspeedai/DeepSpeed/actions/runs/32701815058/job/97354856062) therefore looks like a full-suite lifetime timeout rather than evidence of an Adam regression. A rerun should distinguish runtime/order variance; if it recurs, the shared 3600-second Sandbox lifetime may need more headroom.

csrc/cpu/adam/fused_adam.cpp
```diff
@@ -19,13 +19,16 @@ void multi_tensor_adam(int chunk_size,
...
+    const int optimizer_id = mode;
+    if (!initialized[mode]) {
+        create_adam_optimizer(optimizer_id, 1e-3f, 0.9f, 0.999f, 1e-8f, 0.0f, mode == 1);
```
delock: Are these  '1e-3f, 0.9f, 0.999f, 1e-8f, 0.0f' value place holder?  If they are better define a macro as place holder and put it here.

tohtana (changes requested):
The fix looks good to me, thank you @PKUWZP! This is important to unblock the CI.

I noticed that the new regression test is not run in CPU CI on hosts without fp16 support. This means its coverage can vary depending on the underlying GitHub Actions runner hardware. As I understand it, this bug had existed for a long time but remained hidden because the runners being used did not support fp16, causing the entire test module to be skipped.

Could we scope the fp16 skip to `TestAdamConfigs` only, while leaving `test_fused_adam_matches_torch` eligible to run on CPU regardless of fp16 support?

PKUWZP:
> The fix looks good to me, thank you @PKUWZP! This is important to unblock the CI.
> 
> I noticed that the new regression test is not run in CPU CI on hosts without fp16 support. This means its coverage can vary depending on the underlying GitHub Actions runner hardware. As I understand it, this bug had existed for a long time but remained hidden because the runners being used did not support fp16, causing the entire test module to be skipped.
> 
> Could we scope the fp16 skip to `TestAdamConfigs` only, while leaving `test_fused_adam_matches_torch` eligible to run on CPU regardless of fp16 support?

Let me fix it today.

csrc/cpu/adam/fused_adam.cpp
PKUWZP: They were placeholders, yes — update_state and IncrementStep overwrite every one of them on the first step (IncrementStep takes its full-reset branch whenever the incoming betas differ from the stored ones). Done in a4c407e24: a single named constant kPlaceholderHyperparam = 0.0f with a comment saying exactly that, so the values no longer look meaningful.

PKUWZP:
@tohtana done in a4c407e24 — the module-level fp16 skip is removed and scoped onto TestAdamConfigs only (its config hard-codes fp16 enabled). After merging master, the dtype-parametrized reference test from #8300 (which replaced test_fused_adam_matches_torch) now runs on CPU runners regardless of fp16 support, with its own per-dtype skips — so this regression can't hide behind runner hardware again.

Also merged master to resolve the branch conflicts (took master's reference-based test, kept the kernel fix), and re modal-torch-latest: @FU-max-boop's triage looks right — the failed run hit the 3600s sandbox lifetime at 97% of the full 1232-item suite with no Adam-related failure; the new push re-runs it.

tohtana (approved):
Thank you for the update, @PKUWZP! Looks good to me.

Issue #7842, opened 2026-02-10
[BUG] DistributedAttention produces numerically incorrect results vs its own SeqAllToAll-equivalent baseline when sequence length is not divisible by world size


griffinstalha:
**Describe the bug**
`deepspeed.sequence.layer.DistributedAttention` returns numerically incorrect outputs compared to a reference implementation that reproduces the exact internal logic of `DistributedAttention.forward` using `_SeqAllToAll.apply` + `scaled_dot_product_attention`.

This happens deterministically when:

- Using sequence-first layout [S, B, H, D]
- Sequence length S is not divisible by world_size
- Running on 2 GPUs with NCCL
- Using PyTorch SDPA as the local attention function

The discrepancy is large:

```
MAX_ABS_DIFF: 0.680379
MEAN_ABS_DIFF: 0.139780
ALLCLOSE: False
```

This indicates silent numerical corruption inside `DistributedAttention` when handling uneven sequence splits.
No exception is raised.

**To Reproduce**

1. Use 2 GPUs
2. Install latest DeepSpeed + Torch 2.2
3. Save the testcase below as deepspeed_testcase.py 
4. Run:

`PYTHONUNBUFFERED=1 torchrun --nproc_per_node=2 deepspeed_testcase.py
`
You will see:
```
MAX_ABS_DIFF: ~0.6
ALLCLOSE: False
Test Passed ✅
```

**Testcase**

```
import os
import sys
import time


def _skip(reason: str):
    print(f"SKIP_ENV: {reason}", flush=True)
    sys.exit(0)


def _pass():
    print("Test Passed ✅", flush=True)
    sys.exit(0)


def _fail():
    print("Test Failed ❌", flush=True)
    sys.exit(0)


def _harness_error(e: BaseException):
    print(f"HARNESS_ERROR: {type(e).__name__}: {e}", flush=True)
    sys.exit(1)


def _get_int_env(name: str, default=None):
    v = os.environ.get(name, None)
    if v is None:
        return default
    try:
        return int(str(v).strip())
    except Exception:
        return default


def _barrier_with_timeout(group, timeout_s: float = 20.0):
    import torch
    import torch.distributed as dist

    start = time.time()
    x = torch.ones(1, device=torch.device("cuda"))
    while True:
        try:
            dist.all_reduce(x, op=dist.ReduceOp.SUM, group=group)
            return True
        except Exception:
            return False
        if time.time() - start > timeout_s:
            return False


def main():
    try:
[...]

PR #8317, opened 2026-08-25
Fix the seq-first Ulysses all2all output layout

vineethsaivs:
## Symptom

`_generate_layout_params` (`deepspeed/sequence/layer.py`) builds the `reshape` target for every `all2all` in `DistributedAttention`. For the seq-first layout (`batch_dim_idx=1`, `(s, b, n, h)`) with `scatter_idx < 2` it returns

```python
post_all2all_res_shape = [bs, seq_world_size * global_seq_len, num_local_head // seq_world_size, head_dim]
```

That is the `batch_dim_idx=0` / `scatter_idx >= 2` shape: batch first, sequence multiplied, heads divided. This direction does the opposite, it scatters the sequence and gathers the heads, so the result should be `[global_seq_len // seq_world_size, bs, seq_world_size * num_local_head, head_dim]`.

The element count still matches whenever `num_local_head` is divisible by `seq_world_size`, so the `reshape` succeeds and silently returns a transposed, mis-strided tensor. When it is not divisible the floor division makes a dimension `0` and the reshape raises `shape '[...]' is invalid for input of size ...`.

## Root cause

A copy of the wrong sibling branch during a refactor. Before #6750 extracted this function, `post_all2all` computed the shape inline and this case read:

```python
        else:
            # s, b, n, h
            if scatter_idx < 2:
                output = input.permute(1, 2, 0, 3, 4).contiguous()
                output = output.reshape(seq_len // seq_world_size, bs, seq_world_size * num_head,
                                        head_dim).contiguous()
```

The permute survived the refactor unchanged; only the reshape target was replaced, with the shape from the `batch_dim_idx=0`/`scatter_idx >= 2` branch.

## Reachability

`DistributedAttention.__init__` defaults to `gather_idx=0`, and the output-projection `all2all` swaps the two indices:

```python
output = _SeqAllToAll.apply(self.spg, context_layer, self.gather_idx, self.scatter_idx, batch_dim_idx, ...)
```

so it runs with `scatter_idx=0`. `_SeqAllToAll.backward` swaps them too, so the backward of the q/k/v `all2all`s takes the same path. Both are the seq-first `(s, b, n, h)` layout used by Megatron-DeepSpeed.

## Why the existing tests miss it

- `TestUlyssesAll2All` only runs `batch_dim_idx = 0`.
- `TestUlyssesAll2All_odd` does cover `batch_dim_idx = 1`, but its first call has `num_heads % seq_world_size != 0`, which calls `set_num_kv_heads(...)`. From then on `get_num_kv_heads() is not None` routes every later call to `uneven_heads_all2all`, so `_generate_layout_params` is never reached.
- Both are `DistributedTest` classes behind `skip_on_arch(min_arch=8)`, so neither runs in the CPU CI.

## Fix

One line, restoring the pre-#6750 shape, plus a two-line comment naming the direction.

## Test

`_generate_layout_params`, `pre_all2all_fun` and `post_all2all` are pure, so `TestUlyssesAll2AllLayout` drives them with an emulated `all_to_all_single` (rank `i` sends chunk `j` of dim 0 to rank `j`) and checks that both directions land the right `(sequence, head)` shard of a known tensor. No process group, no accelerator, so it runs in the CPU CI, where this branch currently has no coverage at all.

```
# before, tests/
$ pytest unit/sequence_parallelism/test_ulysses.py -k Layout
FAILED TestUlyssesAll2AllLayout::test_head_to_seq_parallel[2-1] - AssertionError: rank 0 got torch.Size([2, 12, 1, 4]), expected torch.Size([3, 2, 4, 4])
FAILED TestUlyssesAll2AllLayout::test_head_to_seq_parallel[4-1] - RuntimeError: shape '[2, 48, 0, 4]' is invalid for input of size 192
2 failed, 6 passed

# after
8 passed

# whole file, after
8 passed, 43 skipped        (the 43 are the DistributedTest classes, no GPU here)
```

`yapf --style .style.yapf` and `flake8` are clean on both files.

### DIFF
--- a/deepspeed/sequence/layer.py
+++ b/deepspeed/sequence/layer.py
@@ -50,7 +50,9 @@ def _generate_layout_params(scatter_idx, batch_dim_idx, seq_world_size, input):
             pre_all2all_permute_idx = None
 
             post_all2all_permute_idx = (1, 2, 0, 3, 4)
-            post_all2all_res_shape = [bs, seq_world_size * global_seq_len, num_local_head // seq_world_size, head_dim]
+            # seq-first layout: the all2all scatters the sequence and gathers the heads, so the
+            # result keeps bs in dim 1 with a local sequence and every rank's heads.
+            post_all2all_res_shape = [global_seq_len // seq_world_size, bs, seq_world_size * num_local_head, head_dim]
         else:
             local_seq_len, bs, num_total_head, head_dim = input.shape
             assert num_total_head % seq_world_size == 0, f"Number of heads ({num_total_head}) must be divisible by the sequence parallel size ({seq_world_size})!"
--- a/tests/unit/sequence_parallelism/test_ulysses.py
+++ b/tests/unit/sequence_parallelism/test_ulysses.py
@@ -11,7 +11,7 @@ from deepspeed import initialize
 import deepspeed.runtime.sequence_parallel.parallel_state_sp as sp_mpu
 from transformers import AutoModel
 from unit.common import DistributedTest
-from deepspeed.sequence.layer import _SeqAllToAll
+from deepspeed.sequence.layer import _SeqAllToAll, _generate_layout_params, post_all2all, pre_all2all_fun
 from deepspeed.sequence.fpdt_layer import _FPDTGPUOffloadingAttentionImpl_, FPDT_InputConstruct
 from unit.util import skip_on_arch
 from unit.simple_model import *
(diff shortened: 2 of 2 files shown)


vineethsaivs:
Flagging some prior art I only found today: #7868 proposes the same one-line change to `post_all2all_res_shape`, as part of a larger draft that also pads uneven sequence lengths and touches `fpdt_layer.py`, `sequence/__init__.py` and `utils/groups.py`. That draft has been conflicting and unchanged since 2026-02-22, and the issue it was filed against (#7842) was closed out by its reporter after the mismatch turned out to be their baseline rather than DeepSpeed.

I did not know about it when I opened this. Since the shape bug is real independently of the padding work, this PR keeps to just that one line plus a CPU regression test that fails on master (2 failed, 6 passed) and passes with the fix (8 passed). Happy to close this in favour of #7868 if a maintainer would rather revive the larger change.

tohtana (approved):
Hi @vineethsaivs,
This is a very high-impact fix! And the fix looks correct to me.
I really appreciate you submitting this PR.

tohtana:
Thank you @vineethsaivs for your comments about the prior PR (#7868). I have missed it.
I think #7868 is still worth merging, even without the shape fix from this PR, since it includes several other fixes.

Sorry, @harshang03, I should have been more careful about the overlap. Please feel free to revive #7868 and continue working on the remaining issues.

PR #8322, opened 2026-08-26
XPU: register native pinned host memory with the SYCL runtime

### LINKED ISSUE #8315: [REQUEST] XPU: register native pinned host memory with the device runtime (follow-up to #8283)

#8283 taught the native pinned-memory allocator (`DS_PIN_MEMORY_BACKEND=native`, `posix_memalign` + `mlock`) to register host memory with the device runtime so that `copy_(..., non_blocking=True)` can go through the DMA engine. On CUDA this closed a large bandwidth gap (H200: ~10-17 GB/s with mlock only vs. ~53-55 GB/s after `cudaHostRegister`).

Registration goes through the new `register_host_memory` / `unregister_host_memory` accelerator hooks, but only `CUDA_Accelerator` overrides them. `XPU_Accelerator` inherits the base-class no-op, so on Intel GPUs native pinned buffers stay mlock-only and H2D/D2H transfers run at pageable bandwidth even with `DS_PIN_MEMORY_REGISTER_DEVICE=1`. The degradation is silent: registration just returns `False` and nothing logs.

- Implement `register_host_memory` / `unregister_host_memory` in `accelerator/xpu_accelerator.py`, e.g. via Level Zero `zeMemRegisterHostMemory` / `zeMemUnregisterMemory` on the allocation range, honoring the same `DS_PIN_MEMORY_REGISTER_DEVICE` opt-out and the graceful "log once and continue with mlock" fallback on failure.
- Extend `tests/unit/v1/pin_memory/test_pin_memory.py` with an XPU equivalent of the CUDA cudart-mock test.
- Optionally extend `benchmarks/pin_memory/h2d_d2h_bench.py` so the before/after H2D/D2H bandwidth can be quantified on XPU.

- Keep mlock-only on XPU (status quo): functionally correct, but leaves H2D/D2H at pageable speeds for native-pin users on Intel GPUs.

**Additional context**

- Hooks added in #8283: no-op defaults in `accelerator/abstract_accelerator.py`, CUDA implementation in `accelerator/cuda_accelerator.py` (`torch.cuda.cudart().cudaHostRegister` / `cudaHostUnregister`).
- Call sites: `NativePinnedMemory.pin` / `_release` in `deepspeed/utils/pin_memory.py`.

delock:
Fixes #8315.

## Summary

- Implement `register_host_memory` / `unregister_host_memory` in `XPU_Accelerator` via SYCL's `prepare_for_device_copy` / `release_from_device_copy` (the SYCL analogue of `cudaHostRegister`/`cudaHostUnregister`). Previously XPU inherited the base-class no-ops, so `DS_PIN_MEMORY_BACKEND=native` buffers stayed mlock-only on Intel GPUs.
- Honors the existing `DS_PIN_MEMORY_REGISTER_DEVICE` opt-out and falls back to "log once and continue with mlock" when the symbols cannot be resolved.
- Make `benchmarks/pin_memory/h2d_d2h_bench.py` accelerator-agnostic (it hardcoded a CUDA check and torch's CUDA-only `pin_memory=True` path) so the arms can be measured on XPU.

## Implementation note: resolving the SYCL entry points

Neither SYCL nor torch exposes a C entry point for these functions, so the mangled C++ symbols are resolved with `ctypes`. Two constraints, both found the hard way on an Arc Pro B60:

**Which library.** Opening the `libsycl.so` development symlink `dlopen`s a *second* SYCL runtime alongside the one torch already loaded. Both end up with their own driver state and the process aborts at interpreter shutdown with `UR_RESULT_ERROR_UNINITIALIZED`. We therefore only ever bind, with `RTLD_NOLOAD`, to the runtime already mapped into the process.

**Not by soname.** The soname is version-dependent -- oneAPI 2025.x ships `libsycl.so.8`, 2026.x ships `libsycl.so.9` (both are present on the test machine). Hardcoding `.so.8` would silently disable registration after a oneAPI upgrade. The path is instead read from `/proc/self/maps`, which tracks the version and also disambiguates between multiple oneAPI installs.

The mangled names themselves are stable: they live in SYCL's `_V1` inline ABI namespace and are byte-identical in `libsycl.so.8` and `libsycl.so.9`, verified with `nm` on both. Symbols disappearing is handled as the mlock-only fallback rather than an error.

Two tests pin this down (`test_xpu_sycl_lookup_reuses_torch_runtime`, `test_xpu_sycl_lookup_without_mapped_runtime`) so the lookup cannot regress to loading a second runtime or a hardcoded soname.

The `sycl_queue` attribute torch exposes as an `int` is a `sycl::queue*` -- confirmed against torch's own AOTI runtime, which dereferences it as `*queue_ptr` -- and pointer/reference share an ABI, so passing it to a `const sycl::queue&` parameter is sound.

## Testing

Intel Arc Pro B60 (4 cards, `ZE_AFFINITY_MASK=3`), torch 2.10.0+xpu, oneAPI 2025.3 + 2026.0 both installed:

- `pytest tests/unit/v1/pin_memory/test_pin_memory.py` -- 21 passed.
- `benchmarks/pin_memory/h2d_d2h_bench.py` runs all three arms to completion and exits 0 (it aborted at shutdown before the `RTLD_NOLOAD` fix).

Bandwidth on this particular machine is **not** a useful signal: the host is a VM whose PCIe path tops out around 1.8 GB/s, and all arms -- including plain pageable -- already sit at that ceiling, so registration has no headroom to show a gain:

```
arm,size_mib,h2d_gbps,d2h_gbps
torch,256,1.82,1.81
native-unregistered,256,1.80,1.78
native-registered,256,1.61,1.81
```

The path is nevertheless exercised: a standalone probe against non-mlocked `malloc` memory (which is not at the mlock-assisted ceiling) moved D2H from 1.53 to 1.78 GB/s after `prepare_for_device_copy`, matching torch-pinned. Confirming the intended H2D/D2H win would need a bare-metal Intel GPU host with a full-bandwidth PCIe link.

## Known limitation

SYCL reports registration failures by throwing `sycl::exception`, which cannot cross the `ctypes` boundary -- a bad address returns normally rather than raising. So unlike the CUDA path (which checks the `cudaError_t`), a failed registration here degrades silently to mlock-only speed instead of being reported. Making this observable would need a small C++ shim that catches the exception and returns a status code; I left it out to keep the change dependency-free, but am happy to add it if reviewers prefer.

### DIFF
--- a/accelerator/xpu_accelerator.py
+++ b/accelerator/xpu_accelerator.py
@@ -3,6 +3,7 @@
 
 # DeepSpeed Team
 
+import ctypes
 import torch
 from deepspeed.accelerator.abstract_accelerator import DeepSpeedAccelerator
 import functools
@@ -14,6 +15,29 @@ try:
     oneccl_imported_p = True
 except ImportError as e:
     oneccl_imported_p = False
+
+# Mangled names of ``sycl::ext::oneapi::experimental::prepare_for_device_copy(const void*, size_t, const queue&)``
+# and ``release_from_device_copy(const void*, const queue&)``. These are the SYCL
+# equivalent of ``cudaHostRegister``/``cudaHostUnregister``. Neither SYCL nor torch
+# exposes a C entry point for them, so the C++ symbols are resolved directly.
+SYCL_PREPARE_FOR_DEVICE_COPY = "_ZN4sycl3_V13ext6oneapi12experimental23prepare_for_device_copyEPKvmRKNS0_5queueE"
+SYCL_RELEASE_FROM_DEVICE_COPY = "_ZN4sycl3_V13ext6oneapi12experimental24release_from_device_copyEPKvRKNS0_5queueE"
+
+
+@functools.lru_cache(maxsize=None)
+def _sycl_host_copy_funcs():
+    """Resolve the SYCL host-registration functions, or return None when unavailable."""
+    try:
+        libsycl = ctypes.CDLL("libsycl.so", mode=ctypes.RTLD_GLOBAL)
+        prepare = getattr(libsycl, SYCL_PREPARE_FOR_DEVICE_COPY)
+        release = getattr(libsycl, SYCL_RELEASE_FROM_DEVICE_COPY)
+    except (OSError, AttributeError):
+        return None
+    prepare.restype = None
+    prepare.argtypes = [ctypes.c_void_p, ctypes.c_size_t, ctypes.c_void_p]
+    release.restype = None
+    release.argtypes = [ctypes.c_void_p, ctypes.c_void_p]
+    return prepare, release
 
 
 class XPU_Accelerator(DeepSpeedAccelerator):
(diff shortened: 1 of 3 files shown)


delock:
This feature (register host memory) depends on oneapi 2026.2.  Need to wait release of oneapi > 2026.1

PR #8326, opened 2026-08-26
Add an opt-in fused weighted restore for AutoEP

yh0903:
## Summary
- Add an opt-in AutoEP `combine_impl="fused_weighted_sum"` path while keeping the existing eager weighted reduction as the default.
- Restore `[tokens * top_k, hidden]` expert rows directly to `[batch, sequence, hidden]` in one Triton pass with FP32 weighting and accumulation, including expert-row and routing-score gradients.
- Fail fast for unsupported CUDA/dtype/score/AutoTP/expert-TP configurations, preserve higher-order autograd through a differentiable fallback, and document the experimental path.

## Performance
- Canonical-shape H100 microbenchmark: 2.98x forward and 1.87x forward+backward, saving about 0.215 ms per layer.
- Qwen3-30B-A3B, 48 layers, EP16, 2x8 H100: pooled median improved by 0.97%, with a 95% CI of `[-1.12%, +3.02%]`; this is not yet a statistically conclusive end-to-end win.
- Peak reserved memory decreased by 96 MiB, matching the removed 64 MiB FP32 weighted intermediate and 32 MiB assignment buffer.
- The experimental expert reorder was removed after launch-amortized measurements showed no benefit, so this PR changes only the weighted restore.

## Testing Done
- [x] Unit tests added/updated
- [x] Integration tests pass
- [x] Manual testing performed

- Repository pre-commit hooks passed for all nine changed files.
- H100 fused token-op suite: 23 passed, including top-k 2/4/6/8, non-power-of-two hidden sizes, input-contract validation, forward/backward parity, and double backward.
- H100 configuration tests: 5 passed for standard, folded AutoTP, expert tensor parallelism, and score-application validation.
- Full-step eager/fused parity: 3 passed, covering loss, output, input gradients, router/expert gradients, optimizer parameter deltas, activation checkpointing on/off, EP2, and local experts.
- Earlier final restore-only sweep: 121 kernel/config tests passed before the additional review-driven safety cases were added.


deepspeed/moe/autoep_fused_token_ops.py
```diff
@@ -0,0 +1,339 @@
+# Copyright (c) DeepSpeed Team.
```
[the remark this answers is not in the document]
yh0903: Not applicable: the reviewed commit `c02ea19b5f` contains `Signed-off-by: yh0903 <helloyu0903@gmail.com>`. Every non-merge commit in this PR has a Signed-off-by trailer; the only unsigned commit is the merge from master, which is exempt. The DCO check is passing.

yh0903:
The failing `cpu-torch-latest` job is an upstream baseline failure. The current `master` run at `32e301ff` fails the same two cases (`test_fused_adam_matches_reference[fp32-adam]` and `[fp16-adam]`): https://github.com/deepspeedai/DeepSpeed/actions/runs/33029999569. This PR does not touch Adam code. The dedicated fix is tracked in #8307; I will rerun this PR after the upstream issue is resolved.

deepspeed/moe/autoep_fused_token_ops.py
hwchen2017: Can you move this file to deepspeed/ops/triton_ops?

deepspeed/moe/autoep_fused_token_ops.py
```diff
@@ -0,0 +1,339 @@
...
+if _IS_ROCM_PYTORCH:
+    _TRITON_AVAILABLE = False
+else:
+    try:
+        import triton
+        import triton.language as tl
+
+        _TRITON_AVAILABLE = True
+    except ImportError:
+        _TRITON_AVAILABLE = False
```
hwchen2017: Reuse the triton availability check here: https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/ops/triton_ops/_triton.py

deepspeed/moe/autoep_fused_token_ops.py
```diff
@@ -0,0 +1,339 @@
...
+
+SUPPORTED_ROW_DTYPES = (torch.bfloat16, torch.float16)
```
hwchen2017: Can you also support float32?

tests/unit/v1/moe/test_autoep_fused_token_ops.py
```diff
@@ -0,0 +1,146 @@
+# Copyright (c) DeepSpeed Team.
```
hwchen2017: Move this file to tests/unit/v1/ops/triton_ops

deepspeed/moe/autoep_fused_token_ops.py
```diff
@@ -0,0 +1,339 @@
...
+        raise RuntimeError(f"fused weighted restore expects token_indices_sorted to contain {expected_rows} "
+                           f"assignments, got shape {tuple(token_indices_sorted.shape)}.")
+    if token_indices_sorted.dtype not in (torch.int32, torch.int64):
+        raise RuntimeError("fused weighted restore expects token_indices_sorted to use int32 or int64 indices, got "
+                           f"{token_indices_sorted.dtype}.")
+    if not torch.is_floating_point(top_scores):
+        raise RuntimeError(f"fused weighted restore expects floating-point top_scores, got {top_scores.dtype}.")
+    if combined_rows.device != top_scores.device or combined_rows.device != token_indices_sorted.device:
+        raise RuntimeError("fused weighted restore expects rows, scores, and indices on the same device, got "
+                           f"{combined_rows.device}, {top_scores.device}, and {token_indices_sorted.device}.")
```
hwchen2017: Some of the sanity checks are over-conservative. For example, it's impossible to have `bsz < 0 or seqlen < 0 or hidden < 0` here. Other code before here will throw an error. These checks will increase CPU overhead.

deepspeed/ops/triton_ops/autoep_fused_token_ops.py
```diff
@@ -0,0 +1,339 @@
...
+        n_tokens, hidden = top_scores.shape[0], combined_rows.shape[-1]
+        output = torch.empty((n_tokens, hidden), dtype=combined_rows.dtype, device=combined_rows.device)
+
+        ctx.save_for_backward(combined_rows, top_scores, inverse)
+        ctx.top_k = top_k
+        if n_tokens == 0 or hidden == 0:
+            return output
+
+        k_padded = _padded_top_k(top_k)
+        block_hidden = _block_hidden(hidden, slots=k_padded)
```
hwchen2017, line 213: Does multi thread blocks produce better performance compared to single thread block for 1 token?

deepspeed/moe/autoep_fused_token_ops.py
```diff
@@ -0,0 +1,339 @@
...
+        combined_rows, top_scores, inverse = ctx.saved_tensors
+        grad_output = grad_output.contiguous()
+
+        if torch.is_grad_enabled():
+            grad_rows, grad_scores = _differentiable_backward(grad_output, combined_rows, top_scores, inverse,
+                                                              ctx.top_k)
+            return grad_rows, grad_scores, None, None
+
+        # Zero initialization keeps malformed direct calls deterministic.
+        grad_rows = torch.zeros_like(combined_rows)
```
hwchen2017: Does the tensor need to be initialized to zero? The following kernel will overwrite it.

deepspeed/moe/autoep_fused_token_ops.py
```diff
@@ -0,0 +1,339 @@
...
+        grad_scores = torch.empty_like(top_scores)
+
+        n_tokens, hidden = top_scores.shape[0], combined_rows.shape[-1]
+        if n_tokens == 0 or hidden == 0:
+            return grad_rows, torch.zeros_like(top_scores), None, None
```
hwchen2017: We probably don't need to consider such case.

deepspeed/moe/autoep_fused_token_ops.py
yh0903: Yes! I moved the implementation to `deepspeed/ops/triton_ops/autoep_fused_token_ops.py` and updated the AutoEP and parity-test imports.

deepspeed/moe/autoep_fused_token_ops.py
yh0903: The op now imports `_TRITON_AVAILABLE`, `triton`, and `tl` from `deepspeed.ops.triton_ops._triton`. I retained the explicit ROCm rejection because this path has only been validated on NVIDIA CUDA so far.

deepspeed/moe/autoep_fused_token_ops.py
yh0903: Float32 rows are now supported alongside float16 and bfloat16.

tests/unit/v1/moe/test_autoep_fused_token_ops.py
yh0903: Moved! Thanks for pointing to the location.

deepspeed/moe/autoep_fused_token_ops.py
yh0903: Agreed. I removed the per-call shape, dimension, dtype, index, and device checks from the fused restore hot path. These values are produced internally by the router, `argsort`, and expert path. The cached one-time compatibility check before collectives remains in place.

deepspeed/ops/triton_ops/autoep_fused_token_ops.py
yh0903: I measured both launch layouts on H100 with top-k 8 and `BLOCK_H=256`:

| Tokens | Hidden | Multi-block | Single program | Result |
|---:|---:|---:|---:|---|
| 1 | 2048 | 15.406 us | 13.928 us | single program 10.6% faster |
| 1 | 4096 | 15.347 us | 24.926 us | multi-block 1.62x faster |
| 24 | 2048 | 15.402 us | 13.925 us | single program 10.6% faster |
| 1024 | 2048 | 78.583 us | 80.890 us | multi-block 2.9% faster |

The single-program layout is slightly better for very small H=2048 workloads, with an absolute difference of about 1.5 us. The current multi-block layout is faster at the canonical training shape and substantially faster for larger hidden sizes, so I kept it rather than adding a second dynamically selected kernel.

deepspeed/moe/autoep_fused_token_ops.py
yh0903: Agreed. `token_indices_sorted` is a complete assignment permutation, so the backward kernel overwrites every `grad_rows` row exactly once. I changed it to `empty_like`; the inverse-permutation allocation now uses `empty` for the same reason.

deepspeed/moe/autoep_fused_token_ops.py
yh0903: Thanks for pointing out! Just double checked that normal AutoEP execution supplies non-empty hidden states and a positive configured top-k. I removed the zero-token/zero-hidden branches from the forward and backward paths.

deepspeed/ops/triton_ops/autoep_fused_token_ops.py
hwchen2017: Can you try from 4K to 16K tokens?

deepspeed/ops/triton_ops/autoep_fused_token_ops.py
```diff
@@ -0,0 +1,295 @@
...
+            BLOCK_H=block_hidden,
+        )
+        return output
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        combined_rows, top_scores, inverse = ctx.saved_tensors
+        grad_output = grad_output.contiguous()
+
+        if torch.is_grad_enabled():
```
hwchen2017: Calculating a gradient of a gradient is uncommon in model training. We don' need to consider such edge case if there is no demonstrated need.

deepspeed/ops/triton_ops/autoep_fused_token_ops.py
yh0903: Sure, I extended the H100 comparison to the requested token counts, keeping `top_k=8`, `hidden=2048`, and `BLOCK_H=256` fixed:

| Tokens | Multi-block | Single program | Multi-block speedup |
|---:|---:|---:|---:|
| 4,096 | 0.302345 ms | 0.310745 ms | 2.8% |
| 8,192 | 0.600182 ms | 0.606852 ms | 1.1% |
| 16,384 | 1.194984 ms | 1.198649 ms | 0.3% |

The two layouts converge as the larger workloads saturate memory bandwidth, but multi-block remains slightly faster at all three requested sizes. I kept the existing launch layout.

deepspeed/ops/triton_ops/autoep_fused_token_ops.py
yh0903: Thank you, and I agreed. I removed the higher-order differentiable fallback, the `torch.is_grad_enabled()` branch, and the dedicated double-backward test. The normal first-order Triton forward and backward paths are
unchanged.

tohtana (approved):
Hi @yh0903,
Thank you for the great improvement on AutoEP. I also reviewed and this PR now looks good to me.
Let's merge after @hwchen2017 also gives the green light.

hwchen2017 (approved):

yh0903:
Thanks @hwchen2017 @tohtana for reviewing! I just rebased with master branch and resolved the conflicts. Now CIs are all green, could you help merge this PR when convenient? Thank you!

PR #8331, opened 2026-08-27
Add an opt-in ZeRO-1/2 gradient norm fast path

yh0903:
## Summary

- add `zero_optimization.compute_grad_norm`, defaulting to `true`
- allow ZeRO Stage 1/2 GPU optimizers to skip an unused global gradient norm while preserving overflow detection
- skip the gradient multiply when clipping is disabled and the effective loss scale is exactly 1
- return `None` from `get_global_grad_norm()` when norm computation is explicitly disabled
- fail fast for gradient clipping, optimizer offload, ZenFlow, unsupported ZeRO stages, checkpoint-restored clipping, and the dedicated ZeRO-1 BF16 optimizer with FP32 gradient accumulation

## Safety contract

Finite/overflow checking is unchanged. This does not remove the required non-finite scan; it only removes work with no mathematical consumer.

The default configuration and numerical behavior are unchanged. Identity unscaling is skipped automatically when clipping is disabled and the effective loss scale is exactly 1. Gradient-norm computation remains enabled by default because its cached result is externally observable through `get_global_grad_norm()`.

## Performance

Qwen3-30B-A3B, 48 layers, EP16 on 2x8 H100, sequence length 1024, BF16, ZeRO-1, activation checkpointing enabled:

| Metric | Baseline | Fast path | Change |
|---|---:|---:|---:|
| Median step | 987.92 ms | 966.67 ms | -2.2% |
| Optimizer update | 84.93 ms | 59.82 ms | -29.6% |
| Global norm | 20.30 ms | 0 ms | removed |
| Identity unscale | 5.09 ms | 0.03 ms | removed |

The paired block used a benchmark-only control that restores the pre-change multiply-by-one path. Peak allocated and reserved memory were unchanged. Two additional rotated blocks isolating the opt-in norm change were both positive (+3.3% and +0.6% E2E); their optimizer-phase saving was stable at about 20 ms, while full-step variance remained larger.

## Testing Done

- rebased onto current `master`
- H100 targeted suite: 20 passed, covering Stage 1/2 update parity, overflow skip behavior, `get_global_grad_norm()`, clipping/offload/ZenFlow guards, checkpoint clipping, identity/non-identity scaling, and the ZeRO-1 BF16 optimizer with FP32 gradient accumulation
- changed-file pre-commit hooks passed; flake8 5.0.4 was run separately under Python 3.13 because its pyflakes dependency is incompatible with Python 3.14
- current-head multi-GPU H100 suite: **20 passed, 0 failed, 0 skipped**
- full repository CI still requires maintainer approval to run on the fork PR

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -2068,6 +2068,9 @@ class DeepSpeedEngine(Module):
                     logger.warning("**** You are using ZeRO with an untested optimizer, proceed with caution *****")
             if model_dtype == torch.bfloat16 and grad_accum_dtype == torch.float32 and self.zero_optimization_stage(
             ) == 1 and not self.zero_cpu_offload():
+                if not self.zero_compute_grad_norm():
+                    raise ValueError("zero_optimization.compute_grad_norm=false does not support ZeRO Stage 1 with "
+                                     "BF16 parameters and FP32 gradient accumulation")
                 return BFLOAT16
             return ZERO_OPTIMIZATION
         elif amp_enabled:
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -23,7 +23,7 @@ from contextvars import ContextVar
 from threading import Lock
 from weakref import ref
 
-from typing import Callable, Dict, Union, Iterable, Container, List
+from typing import Callable, Dict, Union, Iterable, Container, List, Optional
 
 import deepspeed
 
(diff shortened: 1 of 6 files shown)


yh0903:
CI status note:

- The required Modal job timed out twice at the 75-minute GitHub job limit; neither run reported a test assertion failure.
- In the latest run, the Modal sandbox did not begin provisioning until about 47 minutes after the controller started. Pytest then reached 29% with passing tests before GitHub canceled the job at the hard timeout: https://github.com/deepspeedai/DeepSpeed/actions/runs/33746999940
- The previous run spent about 70 minutes waiting for the sandbox before setup began: https://github.com/deepspeedai/DeepSpeed/actions/runs/33734091483
- The rebased PR head was also validated independently on H100: 20 targeted tests passed, including ZeRO Stage 1/2 parity, overflow handling, unsupported configuration guards, and the BF16 + FP32 gradient-accumulation regression case. OpenConnect execution: `fecf8204f9e7648be914`.
- Changed-file pre-commit hooks and flake8 5.0.4 passed.

The remaining red required check appears to be CI provisioning/runtime capacity rather than a code failure and needs a maintainer rerun or a longer Modal job timeout.

tohtana (approved):
Hi @yh0903, 
Thank you for the improvement! Sorry for the delay of review. 
This looks good to me.

PR #8334, opened 2026-08-27
Stop the curriculum schedule starting below min_difficulty

vineethsaivs:
## Symptom

The curriculum schedule can start below the `min_difficulty` it is configured with, and for one pair the tutorial itself recommends it starts at **0**.

```python
from deepspeed.runtime.data_pipeline.curriculum_scheduler import CurriculumScheduler

def sched(min_d, step):
    return CurriculumScheduler({
        "min_difficulty": min_d, "max_difficulty": 1024, "schedule_type": "fixed_linear",
        "schedule_config": {"total_curriculum_step": 100, "difficulty_step": step},
    })

for min_d, step in [(8, 8), (8, 16), (64, 16), (1, 8), (10, 8)]:
    s = sched(min_d, step)
    print(min_d, step, [s.get_difficulty(i) for i in range(6)])
```

| `min_difficulty` | `difficulty_step` | first six steps |
| --- | --- | --- |
| 8 | 8 | `[8, 16, 24, 32, 48, 56]` |
| **8** | **16** | **`[0, 16, 16, 32, 48, 48]`** |
| 64 | 16 | `[64, 64, 80, 80, 96, 112]` |
| **1** | **8** | **`[0, 8, 16, 24, 40, 48]`** |
| **10** | **8** | **`[8, 16, 24, 40, 48, 56]`** |

`min_difficulty=8` with `difficulty_step=16` is not a contrived pair. The tutorial recommends "starting with `min_difficulty` at 8 (million-scale models) or 64 (billion-scale models)" and separately "we usually set [`difficulty_step`] to 8 (for FP16 data) or 16 (for INT8 data)". A million-scale model on INT8 data lands on exactly that combination, and its first training step gets a sequence length of 0.

## Root cause

`__fixed_root_get_difficulty`, which serves both `fixed_linear` (root degree 1) and `fixed_root`, floors the interpolated value to a multiple of `difficulty_step` and then clamps only the top:

```python
next_difficulty -= (next_difficulty % s_state[CURRICULUM_LEARNING_SCHEDULE_DIFFICULTY_STEP])
next_difficulty = min(next_difficulty, self.state[CURRICULUM_LEARNING_MAX_DIFFICULTY])
```

At step 0 the interpolation is exactly `min_difficulty`, so the floor subtracts `min_difficulty % difficulty_step` and there is nothing to stop it going under. The tutorial's own formula for this schedule is `((step/total)**(1/root_degree)) * (max_difficulty - min_difficulty) + min_difficulty`, which starts at `min_difficulty`.

## Fix

Clamp the bottom the way the top already is, one line.

This does not introduce a new exception to the "difficulty is a multiple of `difficulty_step`" rule: the existing top clamp already returns `max_difficulty` verbatim when it is not a multiple. With `max_difficulty=1000` and `difficulty_step=16` the schedule returns 1000, not 992, once it runs out. Both endpoints being the configured values rather than multiples of the step is the behaviour this function already has at one end.

`__fixed_discrete_get_difficulty` picks from an explicit list and is untouched.

## Test

Two tests in `tests/unit/runtime/test_data_efficiency.py`, both plain CPU tests rather than `DistributedTest`, since `CurriculumScheduler` needs neither an accelerator nor a process group:

- `test_curriculum_never_starts_below_min_difficulty`, parametrized over `fixed_linear` and `fixed_root` and over five `(min_difficulty, difficulty_step)` pairs including the tutorial's own recommendations, checks the first twenty steps stay within `[min_difficulty, max_difficulty]`.
- `test_curriculum_endpoints_are_the_configured_values` pins both ends with a `max_difficulty` that is not a multiple of `difficulty_step`. Its top-end assertion passes on master too, which is what makes it the control for the argument above.

Against master: **7 failed, 4 passed, 6 skipped** (`assert 0 >= 8`, `assert 8 >= 10`, `assert 0 == 8`). With the fix: **11 passed, 6 skipped**. The 6 skipped are the file's pre-existing `DistributedTest` cases, which need 2 GPUs; a pristine checkout reports the same 6 skips and nothing else.

yapf and flake8 clean, with yapf making no changes to either file.

### DIFF
--- a/deepspeed/runtime/data_pipeline/curriculum_scheduler.py
+++ b/deepspeed/runtime/data_pipeline/curriculum_scheduler.py
@@ -137,6 +137,12 @@ class CurriculumScheduler(object):
             (self.state[CURRICULUM_LEARNING_MAX_DIFFICULTY] - self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY]) +
             self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY])
         next_difficulty -= (next_difficulty % s_state[CURRICULUM_LEARNING_SCHEDULE_DIFFICULTY_STEP])
+        # Flooring to a multiple of difficulty_step can land below the configured start,
+        # and does for pairs the tutorial itself recommends: min_difficulty 8 with
+        # difficulty_step 16 gives 0 on the first step, which is a zero-length sequence
+        # for the seqlen metric. Clamp the bottom the way the top already is; both ends
+        # are the configured values rather than multiples of the step.
+        next_difficulty = max(next_difficulty, self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY])
         next_difficulty = min(next_difficulty, self.state[CURRICULUM_LEARNING_MAX_DIFFICULTY])
         return next_difficulty
 
--- a/deepspeed/runtime/data_pipeline/curriculum_scheduler.py
+++ b/deepspeed/runtime/data_pipeline/curriculum_scheduler.py
@@ -137,12 +137,15 @@ class CurriculumScheduler(object):
             (self.state[CURRICULUM_LEARNING_MAX_DIFFICULTY] - self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY]) +
             self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY])
         next_difficulty -= (next_difficulty % s_state[CURRICULUM_LEARNING_SCHEDULE_DIFFICULTY_STEP])
-        # Flooring to a multiple of difficulty_step can land below the configured start,
-        # and does for pairs the tutorial itself recommends: min_difficulty 8 with
-        # difficulty_step 16 gives 0 on the first step, which is a zero-length sequence
-        # for the seqlen metric. Clamp the bottom the way the top already is; both ends
-        # are the configured values rather than multiples of the step.
-        next_difficulty = max(next_difficulty, self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY])
+        # Flooring to a multiple of difficulty_step can land below the configured start, and
+        # does for pairs the tutorial itself recommends: min_difficulty 8 with difficulty_step
+        # 16 gives 0 on the first step, which is a zero-length sequence for the seqlen metric.
+        # Raise the floor to the first multiple of the step at or above min_difficulty rather
+        # than to min_difficulty itself: the constructor warns that every difficulty has to be
+        # a multiple of the step (8 for FP16, 16 for INT8), and a min_difficulty that is not
+        # one cannot also be the first difficulty. This way both contracts hold.
+        step = s_state[CURRICULUM_LEARNING_SCHEDULE_DIFFICULTY_STEP]
+        next_difficulty = max(next_difficulty, -(-self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY] // step) * step)
         next_difficulty = min(next_difficulty, self.state[CURRICULUM_LEARNING_MAX_DIFFICULTY])
         return next_difficulty
 
--- a/tests/unit/runtime/test_data_efficiency.py
+++ b/tests/unit/runtime/test_data_efficiency.py
@@ -10,6 +10,7 @@ from deepspeed.accelerator import get_accelerator
 import pytest
 from unit.common import DistributedTest
 from unit.simple_model import Curriculum_SimpleModel, SimpleModel, random_dataloader, random_dataset
+from deepspeed.runtime.data_pipeline.curriculum_scheduler import CurriculumScheduler
 
 
 class MPU():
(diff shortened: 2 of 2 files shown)


deepspeed/runtime/data_pipeline/curriculum_scheduler.py
```diff
@@ -137,6 +137,12 @@ def __fixed_root_get_difficulty(self, global_steps, root_degree=None):
             (self.state[CURRICULUM_LEARNING_MAX_DIFFICULTY] - self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY]) +
             self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY])
         next_difficulty -= (next_difficulty % s_state[CURRICULUM_LEARNING_SCHEDULE_DIFFICULTY_STEP])
+        # Flooring to a multiple of difficulty_step can land below the configured start,
+        # and does for pairs the tutorial itself recommends: min_difficulty 8 with
+        # difficulty_step 16 gives 0 on the first step, which is a zero-length sequence
+        # for the seqlen metric. Clamp the bottom the way the top already is; both ends
+        # are the configured values rather than multiples of the step.
+        next_difficulty = max(next_difficulty, self.state[CURRICULUM_LEARNING_MIN_DIFFICULTY])
```
ebarkhordar: I ran this against `6505f2ff` in a clean container. The bug reproduces at the merge base `92843ad7`: `min_difficulty=8` with `difficulty_step=16` does start at 0.

The clamp buys that with the other invariant this file documents, in the same pair you picked as motivation. Both constructor branches warn that the difficulty "must be a multiple of this difficulty_step ... (requires multiple of 8 (FP16) or 16 (INT8))", and flooring is what guaranteed it. Clamping up to `min_difficulty` returns a non-multiple whenever `min_difficulty % difficulty_step != 0`:

```
first six difficulties, fixed_linear, max=1024, total_curriculum_step=100

              base 92843ad7           head 6505f2ff
min=8  step=16 [0, 16, 16, 32, 48...]  [8, 16, 16, 32, 48...]   8 is not a multiple of 16
min=10 step=8  [8, 16, 24, 40, 48...]  [10, 16, 24, 40, 48...]  10 is not a multiple of 8
```

So the million-scale INT8 run in your description gets a first-step seqlen of 8 where the constructor says it needs a multiple of 16. Better than 0, and I am not arguing for the old behaviour. What is worth deciding first is that the `difficulty_step % 8` warning does not fire here, since 16 is a multiple of 8, so that user gets a misaligned first step with no diagnostic.

Clamping instead to the smallest multiple at or above `min_difficulty` keeps both: `-(-min_difficulty // difficulty_step) * difficulty_step`, which is 16 for the 8/16 pair.

Which contract did you mean to win, "starts at `min_difficulty`" or "every difficulty is a multiple of `difficulty_step`"? They cannot both hold when `min_difficulty` is not a multiple of the step, and `test_curriculum_endpoints_are_the_configured_values` pins the first.

I only exercised `get_difficulty`, not a training loop.

vineethsaivs:
You are right, and this is the better fix. "Every difficulty is a multiple of `difficulty_step`" is the contract that should win, because the constructor states it as a hardware requirement (8 for FP16, 16 for INT8) while "starts at exactly `min_difficulty`" is only an implication of the name. When `min_difficulty` is not a multiple the two cannot both hold, and my clamp picked the wrong one: the 8/16 pair I used as motivation got a first difficulty of 8 where the constructor asks for 16.

Pushed `4ab6172` with your `-(-min_difficulty // difficulty_step) * difficulty_step`.

I checked all three behaviours over steps 0..99 for six `(min_difficulty, difficulty_step)` pairs, counting violations of each contract:

```
                   below min_difficulty                 not a multiple of step
master             (8,16) (10,8) (1,8) (100,64)         none
clamp to min       none                                 (8,16) (10,8) (1,8) (100,64)
clamp to aligned   none                                 none
```

Same four configurations both times, which is what I should have noticed: the pairs where flooring undershoots are exactly the pairs where `min_difficulty` is not aligned. Aligning the floor is the only one of the three with no violation of either contract.

First six difficulties, `fixed_linear`, max=1024, total_curriculum_step=100:

```
              master                   clamp to min             clamp to aligned
min=8  step=16 [0, 16, 16, 32, 48, 48]  [8, 16, 16, 32, 48, 48]  [16, 16, 16, 32, 48, 48]
min=10 step=8  [8, 16, 24, 40, 48, 56]  [10, 16, 24, 40, 48, 56] [16, 16, 24, 40, 48, 56]
min=100 step=64 [64, 64, 64, 64, 128]   [100, 100, 100, 100, 128] [128, 128, 128, 128, 128]
```

On the missing diagnostic: agreed that the `difficulty_step % 8` warning cannot catch this, since it only looks at the step. I have left the warning alone here rather than widen the PR, because with the aligned floor there is no longer a misaligned value to warn about; the user just starts one step higher than they wrote. If you would rather the constructor also said so out loud when `min_difficulty % difficulty_step != 0`, I am happy to add it in a follow-up.

Test changes, in `tests/unit/runtime/test_data_efficiency.py`:

- `test_curriculum_never_starts_below_min_difficulty` now asserts step alignment beside the floor, sweeps 0..99 instead of 0..19, and adds the `(100, 64)` pair. Against the previous head that alignment assertion fails on 8 of the 12 parametrizations, so it pins exactly the regression you found.
- `test_curriculum_endpoints_are_the_configured_values` was pinning `get_difficulty(0) == 8`, which was the wrong contract, so it is split: `test_curriculum_first_difficulty_is_the_aligned_min` (16 for the 8/16 pair, and 64 unchanged for an already-aligned 64/16) and `test_curriculum_tops_out_at_the_configured_max` (the top end is untouched by this change).

```
$ pytest tests/unit/runtime/test_data_efficiency.py -k "curriculum_never_starts or aligned_min or tops_out" -q
14 passed, 6 deselected

# against the previous head
9 failed, 5 passed
```

`yapf --diff` is clean on both files. Thanks for running it and for spotting that the two contracts collide; I would have shipped the wrong one.

ebarkhordar:
That matches what I measured, and your grid answers the part mine could not: the pairs where flooring undershoots are exactly the pairs where `min_difficulty` is not aligned, so the aligned floor is the only one of the three that satisfies both contracts. I have not re-run against `4ab6172`.

On the follow-up: I would leave the constructor warning alone. With the aligned floor there is no misaligned difficulty left to warn about, so a check on `min_difficulty % difficulty_step != 0` would fire on a config that now behaves correctly, and the first difficulty being one step above what the user wrote is visible in the schedule itself.

vineethsaivs:
Agreed on the warning, and that settles it: nothing further from me on this PR. Your reasoning is the one I could not decide between, that a `min_difficulty % difficulty_step != 0` check would now fire on a config that behaves correctly, and the schedule already shows the user where it starts.

Numbers at `4ab6172`, so you do not have to re-run it:

```
$ pytest tests/unit/runtime/test_data_efficiency.py -k curriculum -q
14 passed, 6 skipped

# against the previous head, the clamp to min_difficulty
9 failed, 5 passed
```

The 9 are the alignment assertion on 8 of the 12 `test_curriculum_never_starts_below_min_difficulty` parametrizations plus `test_curriculum_first_difficulty_is_the_aligned_min`, which is the regression you found, pinned. `yapf --diff` and `flake8` are clean on both files.

tohtana (approved):
@vineethsaivs Thank you for the imporvement! Looks good to me.

Issue #8330, opened 2026-08-27
[BUG] DeepSpeed kernel injection combined with CUDA graphs aborts at startup


Redempt1onzzZZ:
**Describe the bug**
DeepSpeed accepts the combination of kernel injection and CUDA graphs, then aborts during startup with an internal assertion. The same behavior was reproduced on DeepSeek-R1-Distill-Llama-8B and Meta-Llama-3.1-8B-Instruct, while both baselines serve generation successfully.

**To Reproduce**

1. Save the following as a shell command and run it in a shell on a machine with a free CUDA device.
2. Run the command in the target Python environment with the listed dependencies installed.
3. Observe the startup or generation result.

~~~bash
import torch
import deepspeed
from transformers import AutoModelForCausalLM, AutoTokenizer

model_name = "/root/.cache/huggingface/hub/models--deepseek-ai--DeepSeek-R1-Distill-Llama-8B/snapshots/6a6f4aa4197940add57724a7707d069478df56b1"
tokenizer = AutoTokenizer.from_pretrained(model_name)
model = AutoModelForCausalLM.from_pretrained(
    model_name, torch_dtype=torch.bfloat16
).to("cuda")
engine = deepspeed.init_inference(
    model,
    config={
        "return_tuple": False,
        "enable_cuda_graph": True,
        "replace_with_kernel_inject": True,
    },
)
print("DeepSpeed inference engine initialized")
~~~

**Actual behavior**

Startup terminates before the endpoint becomes healthy with AssertionError: cuda graph is not supported with this model, please disable.

**Expected behavior**

The incompatible combination should be rejected during explicit validation with an actionable configuration error rather than an internal assertion.

**System info (please complete the following information):**

- OS: Ubuntu 22.04.3 LTS
- GPU count and types: 1 x NVIDIA H100 80GB HBM3
- DeepSpeed version: 0.19.1
- DeepSpeed-MII: not used
- PyTorch: 2.9.1+cu128
- Transformers: 5.15.0
- CUDA reported by PyTorch: 12.8
- nvcc: 13.3
- Python version: 3.12.12
- Target runtime: /root/anaconda3/envs/deepspeed128

**Docker context**
No Docker image was used. The reproduction ran directly on Ubuntu 22.04.3 LTS in the target conda environment.

**Additional context**

alanhuangyoo:
[...]

PR #8336, opened 2026-08-28
Replace cuda_graph assertion with explicit ValueError; fix custom_op type annotations

chakshu-dhannawat:
This PR addresses #8330 by replacing the internal assertion in `replace_with_policy` with a clear `ValueError` when both `replace_with_kernel_inject=True` and `enable_cuda_graph=True` are requested for a model whose policy does not support CUDA graphs.

Changes:
- `deepspeed/module_inject/replace_module.py`: raise an actionable `ValueError` instead of asserting.
- `tests/unit/inference/test_inference_config.py`: add a regression test using a tiny Llama model that verifies the new error message.
- `deepspeed/compile/custom_ops/tp_collectives.py`: replace `list[int]` annotations with `typing.List[int]` so `import deepspeed` works on PyTorch 2.6+ (the new regression test needs this to import cleanly).

The regression test was verified on a single H100 with:
```
PYTHONPATH=tests/unit pytest tests/unit/inference/test_inference_config.py::TestInferenceCudaGraphConfig::test_cuda_graph_with_kernel_inject_raises -m inference -v
```

### DIFF
--- a/deepspeed/compile/custom_ops/tp_collectives.py
+++ b/deepspeed/compile/custom_ops/tp_collectives.py
@@ -3,6 +3,8 @@
 
 # DeepSpeed Team
 
+from typing import List
+
 import torch
 import deepspeed.comm as dist
 from deepspeed.utils import groups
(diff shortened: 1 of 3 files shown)


chakshu-dhannawat:
Addressed the codex feedback: the regression test now moves the model to `get_accelerator().device_name()` instead of hard-coding `cuda`, so non-CUDA accelerator runs don't fail before the ValueError check.

tohtana (approved):
Thank you for the improvement, @chakshu-dhannawat! Looks good to me.

PR #8342, opened 2026-08-28
Fix sequence overlap backward gradient permutation

taking-lying-flat:
## Description

Sequence-parallel overlap backward must wait for the asynchronous Q/K all-to-all and return the post-permuted gradients before the Q/K producers run backward. Returning those gradients from hooks on the producers' parent nodes instead can change gradient sizes for transpose/view inputs or bypass the producers' backward transformations.

Apply @tohtana's patch from https://github.com/taking-lying-flat/DeepSpeed/pull/1, preserving his authorship. Register hooks directly on the Q/K tensors and return the completed, post-permuted gradients. Also use `ctx.stream is not None` so the stream check does not invoke CUDA stream equality operators.

The included regression test compares synchronous and overlap gradients for transpose-produced Q/K/V using a mocked two-rank all-to-all with the real reshape, permutation, and autograd paths.

## Validation

- Included regression test: `.venv-cherry/bin/python -m pytest --forked -q tests/unit/v1/sequence_parallelism/test_overlap_grad_permutation.py` — 1 passed.
- Additional local regression matrix: 10 passed across CPU and CUDA, covering scaling, transpose, view, shared packed-QKV unbind, and leaf tensors. The asynchronous collective mock defers writing its output until `wait()`; gradients match synchronous execution exactly. CUDA cases use real CUDA streams; multi-GPU NCCL was not tested locally.
- Pre-commit checks passed for both changed files.
- Restored the missing sign-off on the earlier test-removal commit; all PR commits now include sign-offs.

Merged current master to resolve the stream-check conflict and follow the test-directory migration. Removed the test mock for `get_num_kv_heads`, which upstream deleted. All validation above passed again after integration; CUDA hardware: NVIDIA RTX A1000 Laptop GPU (4 GB), PyTorch 2.13.0+cu130.


tohtana (changes requested):
Hi @taking-lying-flat, 
Great catch! This is an important fix.

However, it seems to have an issue: The added return grad still returns from prehooks attached to `query.grad_fn.next_functions[0][0]` and `key.grad_fn.next_functions[0][0]`. This replaces gradients after the Q/K producer’s backward transformation. So it raises a hook size-change error when Q/K are produced by a transpose. 

Separately, I found an existing issue with assert `ctx.stream != None`. This rejects a valid CUDA stream (This issue was not introduced by this PR).

I opened a [PR](https://github.com/taking-lying-flat/DeepSpeed/pull/1) on your fork to fix these issues. Can you review and consider merging it? You can update this branch directly if it's more convenient.

taking-lying-flat:
@tohtana Thank you very much for reviewing this and for taking the time to prepare a fix. I really appreciate your help, and I'm sorry the original change missed the producer-backward ordering issue.

I've pulled in your full patch from https://github.com/taking-lying-flat/DeepSpeed/pull/1, including the regression test, and preserved your authorship. Your test passes, and an additional local matrix passes all 10 CPU/CUDA cases covering scaling, transpose, view, shared QKV unbind, and leaf tensors. The CUDA cases use real streams with mocked all-to-all communication. Pre-commit checks also pass, and I've corrected the missing DCO sign-off.

Would you please take another look and, if everything looks good and the required checks pass, help merge this PR? I'd be very grateful. Thank you again for helping get this fix right!

Update: I also merged current master, resolved the stream-check conflict, and moved the regression test to `tests/unit/v1/sequence_parallelism/`, removing the mock for the deleted `get_num_kv_heads` helper. The included test and all 10 additional CPU/CUDA cases pass again after integration.

tohtana (approved):
Thank you for merging, @taking-lying-flat!

PR #8343, opened 2026-08-28
Validate positive inference and HybridEngine max output tokens

### LINKED ISSUE #8339: [BUG] DeepSpeed accepts negative max_out_tokens and fails generation requests at runtime
**Describe the bug**

~~~
import torch
import deepspeed

).to("cuda")
)
with torch.no_grad():
~~~

**Expected behavior**

**Actual behavior**

**Expected behavior**

- OS: Ubuntu 22.04.3 LTS
- DeepSpeed version: 0.19.1
- DeepSpeed-MII: not used
- PyTorch: 2.9.1+cu128
- Transformers: 5.15.0
- nvcc: 13.3
- Python version: 3.12.12

**Docker context**

taking-lying-flat:
## Description

Reject zero and negative `max_out_tokens` during initial configuration validation. These values previously passed parsing and failed later during generation or model conversion.

Both `DeepSpeedInferenceConfig` (including its `max_tokens` alias) and `HybridEngineConfig` now require a positive value. HybridEngine therefore rejects invalid values when parsing the top-level `DeepSpeedConfig`, before constructing an inference configuration during model conversion. The defaults remain 1024 for inference and 512 for HybridEngine.

Regression coverage is included in the existing inference and runtime configuration test files. Inference tests reject -1 and 0 through both field names. HybridEngine tests construct the top-level `DeepSpeedConfig` with HybridEngine enabled, reject -1 and 0, accept 1/512/1024, and preserve the default of 512. This covers the parsed configuration consumed by `DeepSpeedHybridEngine.new_inference_container`.

Fixes #8339.

## Testing

- `python -m pytest -q -m 'inference or not inference' tests/unit/runtime/test_ds_config_model.py tests/unit/inference/test_inference_config.py --tb=short`: **25 passed**, including the regression coverage and existing CUDA Graph test. Executed on an NVIDIA RTX A1000 Laptop GPU (4 GiB, driver 595.84), Python 3.12.13, PyTorch 2.13.0+cu130.
- Confirmed that the HybridEngine -1/0 rejection tests fail without the new constraint and pass with it.
- `pre-commit run --files deepspeed/inference/config.py deepspeed/runtime/config.py tests/unit/inference/test_inference_config.py tests/unit/runtime/test_ds_config_model.py`: passed.
- `python ci/tests_fetcher.py --base origin/master`: selects the full `tests/unit/v1` scope because `deepspeed/runtime/config.py` is shared runtime configuration.

No full HybridEngine model conversion or generation run was performed.


deepspeed/inference/config.py
```diff
@@ -258,7 +258,7 @@ class DeepSpeedInferenceConfig(DeepSpeedConfigModel):
 
     config: Optional[Dict] = Field(None, alias="args")  # todo: really no need for this field if we can refactor
 
-    max_out_tokens: int = Field(1024, alias="max_tokens")
+    max_out_tokens: int = Field(1024, alias="max_tokens", gt=0)
```
ebarkhordar, line 261: I ran this against `81b1d26` in a clean container (`python:3.11-slim`, torch 2.14.0+cpu, pydantic 2.13.5). The constraint holds for the reported case: `max_out_tokens` and the `max_tokens` alias both reject -1 and 0.

One route into this field does not pass through it. `HybridEngineConfig.max_out_tokens` (`deepspeed/runtime/config.py:517`) is an unbounded `int = 512`, and `hybrid_engine.py:105-106` copies it straight into the field constrained here:

```python
max_out_tokens=self._config.hybrid_engine.max_out_tokens,
min_out_tokens=self._config.hybrid_engine.max_out_tokens,
```

So `{"hybrid_engine": {"enabled": true, "max_out_tokens": -1}}` still passes config parsing, and the rejection lands later, during model conversion. Running those two constructions in order (traceback tail trimmed):

```
hybrid_engine.max_out_tokens = -1
  File "/src/deepspeed/runtime/config_utils.py", line 57, in __init__
    super().__init__(**data)
pydantic_core._pydantic_core.ValidationError: 1 validation error for DeepSpeedInferenceConfig
max_out_tokens
  Input should be greater than 0 [type=greater_than, input_value=-1, input_type=int]
```

The same `gt=0` on `runtime/config.py:517` would keep that case at config validation, where this PR puts the other one.

I only built the configs. I did not run the HybridEngine path itself, which needs a GPU and a model.

taking-lying-flat: Thanks for checking this and catching the HybridEngine entry point, @ebarkhordar!

Added `Field(512, gt=0)` to `HybridEngineConfig.max_out_tokens` in 217e45373. The new tests construct the top-level `DeepSpeedConfig` with HybridEngine enabled, so -1 and 0 must now fail during initial config parsing. They also cover accepted positive values (1, 512, and 1024) and the unchanged default of 512.

Confirmed the two rejection cases fail before the fix and pass afterward. The runtime config tests (15) and inference config tests (9) all pass, as do pre-commit checks. This validation covers config parsing; I have not run a full HybridEngine model conversion or generation path.

ebarkhordar: That is the line, thanks. Testing it through the top-level `DeepSpeedConfig` rather than constructing `HybridEngineConfig` directly is the better choice too, since `hybrid_engine.py:105` reads `self._config.hybrid_engine.max_out_tokens` off the parsed config, so that is the path a real run takes.

tohtana (approved):
Hi @taking-lying-flat,
Thank you for submitting this PR! This looks good to me.

PR #8347, opened 2026-08-29
Include the last sample when packing variable-size batches

vineethsaivs:
### The bug

`batch_by_seqlens` (`deepspeed/runtime/data_pipeline/data_sampling/variable_batch_size_and_lr.py`) packs samples into microbatches by growing a candidate slice and keeping the widest one that stays under the token cap:

```python
valid_batch_end = batch_init
for batch_end in range(batch_init + equal_size_multiple, len(metrics), equal_size_multiple):
    batch = metrics[batch_init:batch_end]
    ...
    if is_batch_valid:
        valid_batch_end = batch_end

if batch_init == valid_batch_end: break  # last batch is not valid (size zero), so we are done
```

`batch_end` is the exclusive end of `metrics[batch_init:batch_end]`, and `range` stops before its own end, so `batch_end` never reaches `len(metrics)`. The widest slice ever offered is `metrics[batch_init:len(metrics) - 1]`, and the last sample of the dataset is never placed in a microbatch. When `batch_init` finally lands on that sample the range is empty, `valid_batch_end` stays equal to `batch_init`, and the `break` comment reads it as "no more data".

Five samples that comfortably fit in one microbatch:

```python
>>> microbatch_ids, _, _ = batch_by_seqlens(seqlens=[1, 2, 3, 4, 5], max_tokens=100, effective_batch_size=1)
>>> sorted(i for _, ids in microbatch_ids for i in ids)
[0, 1, 2, 3]
```

It is silent, and it is every epoch, not a one-off: the sample is dropped from training for good with `sequence_picking_order="dataloader"` or `"seqlen"`, and a different sample is dropped each epoch with `"random"`.

### The fix

Stop at `len(metrics) + 1` so the full-width slice is offered. Nothing else moves: `is_microbatch_valid` still applies the token cap and the min/max batch sizes, and the trim to a whole number of effective batches still decides what survives.

### Test

`tests/unit/runtime/test_variable_batch_size_and_lr.py`, the first tests for this file. Plain CPU pytest functions, no accelerator needed.

- `test_every_sample_is_batched`, parametrized over four dataset sizes and token caps, asserts the batched sample ids are exactly `range(n_samples)`.
- `test_microbatches_stay_within_max_tokens` asserts the same coverage on ragged sequence lengths and that no microbatch exceeds `max_tokens`, so the packing is not bought at the cap's expense.

Before the change:

```
5 failed in 6.14s
```

After:

```
5 passed in 7.10s
```

I also swept the other modes (`sequence_picking_order` in dataloader/seqlen/random x `required_microbatches_of_same_size` in False/True x `effective_batch_size` in 1/2/4, 24 ragged samples, `max_tokens=15`): no duplicated sample ids anywhere, every microbatch within the cap, and full coverage except where the deliberate trim to a whole number of effective batches removes the tail.

`yapf --diff` and `flake8` are clean on both files.

### DIFF
--- a/tests/unit/runtime/test_variable_batch_size_and_lr.py
+++ b/tests/unit/runtime/test_variable_batch_size_and_lr.py
@@ -12,16 +12,73 @@ def batched_sample_ids(microbatch_ids):
     return sorted(sample_id for _, sample_ids in microbatch_ids for sample_id in sample_ids)
 
 
+@pytest.mark.parametrize("required_microbatches_of_same_size", [False, True])
 @pytest.mark.parametrize("n_samples,max_tokens", [(5, 100), (6, 10), (8, 30), (9, 20)])
-def test_every_sample_is_batched(n_samples, max_tokens):
-    """No sample may be left out of the epoch. `batch_end` is the exclusive end of
-    `metrics[batch_init:batch_end]`, so a range stopping at `len(metrics)` never
-    offered the slice that reaches the last sample and it was silently dropped."""
+def test_every_sample_is_batched(n_samples, max_tokens, required_microbatches_of_same_size):
+    """With one microbatch per batch nothing is aligned away and the trim below is a no-op,
+    so the epoch has to contain every sample. `batch_end` is the exclusive end of
+    `metrics[batch_init:batch_end]`, so a range stopping at `len(metrics)` never offered the
+    slice that reaches the last sample and it was silently dropped."""
     seqlens = [5] * n_samples
 
-    microbatch_ids, _, _ = batch_by_seqlens(seqlens=seqlens, max_tokens=max_tokens, effective_batch_size=1)
+    microbatch_ids, _, _ = batch_by_seqlens(
+        seqlens=seqlens,
+        max_tokens=max_tokens,
+        effective_batch_size=1,
+        required_microbatches_of_same_size=required_microbatches_of_same_size,
+    )
 
     assert batched_sample_ids(microbatch_ids) == list(range(n_samples))
+
+
+@pytest.mark.parametrize("effective_batch_size,required_microbatches_of_same_size,n_samples,max_tokens", [
+    (2, True, 6, 10),
+    (2, True, 8, 30),
+    (2, True, 10, 10),
+    (2, False, 8, 30),
+    (4, True, 8, 30),
+])
+def test_larger_effective_batches_also_reach_the_last_sample(effective_batch_size, required_microbatches_of_same_size,
+                                                             n_samples, max_tokens):
+    """The loop steps by `equal_size_multiple`, which is the effective batch size when
+    microbatches must match, so the exclusive end matters at every step size and not only at
+    one. These are the shapes where the last stride lands exactly on `len(metrics)`: each of
+    them lost its tail before the fix, half the dataset in the last case."""
+    seqlens = [5] * n_samples
+
+    microbatch_ids, _, _ = batch_by_seqlens(
+        seqlens=seqlens,
+        max_tokens=max_tokens,
+        effective_batch_size=effective_batch_size,
+        required_microbatches_of_same_size=required_microbatches_of_same_size,
+    )
+
+    assert batched_sample_ids(microbatch_ids) == list(range(n_samples))
+
+
+@pytest.mark.parametrize("effective_batch_size", [1, 2, 4])
+@pytest.mark.parametrize("required_microbatches_of_same_size", [False, True])
+@pytest.mark.parametrize("n_samples,max_tokens", [(9, 10), (10, 10), (12, 10)])
+def test_only_an_unalignable_tail_is_dropped(effective_batch_size, required_microbatches_of_same_size, n_samples,
+                                             max_tokens):
+    """Above one microbatch per batch the epoch is deliberately allowed to fall short: the
+    microbatch list is trimmed to a multiple of the effective batch size so every dataloader
+    gets the same count. What has to hold in every mode is that the samples which survive are
+    the leading ones in order, with nothing repeated and nothing over the token cap."""
+    seqlens = [5] * n_samples
+
+    microbatch_ids, _, _ = batch_by_seqlens(
+        seqlens=seqlens,
+        max_tokens=max_tokens,
+        effective_batch_size=effective_batch_size,
+        required_microbatches_of_same_size=required_microbatches_of_same_size,
+    )
+
+    sample_ids = batched_sample_ids(microbatch_ids)
+    assert sample_ids == list(range(len(sample_ids))), "a sample was dropped from the middle, not the tail"
+    assert len(microbatch_ids) % effective_batch_size == 0, "every dataloader must get the same microbatch count"
+    for _, ids_in_microbatch in microbatch_ids:
+        assert sum(seqlens[sample_id] for sample_id in ids_in_microbatch) <= max_tokens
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/runtime/test_variable_batch_size_and_lr.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@pytest.mark.parametrize("n_samples,max_tokens", [(5, 100), (6, 10), (8, 30), (9, 20)])
+def test_every_sample_is_batched(n_samples, max_tokens):
+    """No sample may be left out of the epoch. `batch_end` is the exclusive end of
+    `metrics[batch_init:batch_end]`, so a range stopping at `len(metrics)` never
+    offered the slice that reaches the last sample and it was silently dropped."""
+    seqlens = [5] * n_samples
+
+    microbatch_ids, _, _ = batch_by_seqlens(seqlens=seqlens, max_tokens=max_tokens, effective_batch_size=1)
```
ebarkhordar: I ran this against `f2d4b959` in a clean container (python 3.11.16, torch 2.13.0+cpu). The `+ 1` is right and it does recover the last sample for the shape this test covers, but the two arguments the test pins are the ones that decide whether the tail survives.

`batch_by_seqlens` called directly, sample ids missing from all microbatches, head vs merge base `f4ee787d`:

```
                                                  head   base
A ebs=1, required=False,  5 samples, max_tok=100  none   [4]
B ebs=2, required=False,  5 samples, max_tok=5    [4]    [4]
C ebs=2, required=True,   5 samples, max_tok=10   [4]    [4]
D ebs=4, required=True,  10 samples, max_tok=10   [8,9]  [8,9]
```

Only A changes. In C and D, `required_microbatches_of_same_size=True` sets `equal_size_multiple = effective_batch_size` (line 87), so `range` steps by more than one and still stops short of `len(metrics)` when the remainder is not a multiple. Line 427 selects that path for a `PipelineEngine`. In B the loop does offer the last sample, and line 119 then trims `microbatches` to a multiple of `effective_batch_size` and discards it.

Both look deliberate to me, so I am not suggesting you change them. What I would change is the claim: "No sample may be left out of the epoch" is a universal, while the parametrize varies `n_samples` and `max_tokens` and pins the other two. Adding them, and asserting what holds in each mode, would pin the fix where it applies.

I did not test `sequence_picking_order` other than the default.

vineethsaivs:
Agreed on the claim, and thank you for pinning the two arguments I left at their defaults. "No sample may be left out of the epoch" is not true above `effective_batch_size=1`, because the trim on line 119 is deliberate, so the test should not have asserted it as a universal. Pushed `f22b2b4`.

One correction to the grid, though, and it matters for where the fix applies. Running your four configurations plus the shapes the parametrize already used, head vs merge base `f4ee787d`, **six rows change, not one**:

```
ebs same   n   mt |  missing HEAD    missing BASE   changed
  1 False 10   10 |            []             [9]   yes
  1 False  9   10 |            []             [8]   yes
  1 False  5  100 |            []             [4]   yes     <- your row A
  1 False  8   30 |            []             [7]   yes
  1 False  6   10 |            []             [5]   yes
  1  True  (same five rows, same result)              yes
  2 False 10   10 |        [8, 9]          [8, 9]
  2 False  9   10 |           [8]             [8]           <- your row B
  2 False  8   30 |            []             [7]   yes
  2 False  6   10 |        [4, 5]          [4, 5]
  2  True 10   10 |            []          [8, 9]   yes
  2  True  9   10 |           [8]             [8]           <- your row C
  2  True  8   30 |            []          [6, 7]   yes
  2  True  6   10 |            []          [4, 5]   yes
  4  True 10   10 |        [8, 9]          [8, 9]           <- your row D
  4  True  8   30 |            []    [4, 5, 6, 7]   yes     <- half the dataset
  4  True  6   10 |        [4, 5]          [4, 5]
```

Your rows B, C and D are all shapes where the tail was unalignable anyway, so they are unchanged for the reasons you gave. But `equal_size_multiple` being larger does not make the exclusive end stop mattering: it matters whenever the last stride lands exactly on `len(metrics)`, which is a property of `n_samples` and `max_tokens` together with the step, not of the step alone. `(ebs=4, required=True, 8 samples, max_tokens=30)` is the sharpest one: the base drops samples 4 through 7, half the dataset, and the head keeps all eight.

So the fix is not `effective_batch_size=1` only. What is genuinely mode-specific is the guarantee, which is what I have rewritten the test around:

- `test_every_sample_is_batched` now parametrizes `required_microbatches_of_same_size` over both values and passes `effective_batch_size=1` explicitly. This is the mode where the universal holds, because `equal_size_multiple` is 1 and the trim is a no-op.
- `test_larger_effective_batches_also_reach_the_last_sample` is new and covers the five larger-stride rows above that the fix recovers, including the `(4, True, 8, 30)` case.
- `test_only_an_unalignable_tail_is_dropped` is new and sweeps all three `effective_batch_size` values and both `required_microbatches_of_same_size` values. It asserts what is true in every mode rather than a universal: the surviving samples are the leading ones in order, nothing is repeated, `len(microbatch_ids)` is a multiple of the effective batch size, and no microbatch exceeds `max_tokens`. That is the contract the trim implements, so it pins the deliberate behaviour instead of arguing with it.

```
$ pytest tests/unit/runtime/test_variable_batch_size_and_lr.py -q
32 passed

# against the merge base
14 failed, 18 passed
```

`yapf --diff` is clean. I have left `sequence_picking_order` at its default too; the fix is in the packing loop, which runs after the ordering, so I did not think it added coverage, but say the word if you would like it swept as well.

ebarkhordar:
You are right, and my grid was too narrow. I re-ran your sharpest row myself, merge base `f4ee787d` against the PR head with the single line reverted, same container:

```
                     missing HEAD    missing BASE
4 True  n=8  mt=30             []    [4, 5, 6, 7]
2 True  n=8  mt=30             []          [6, 7]
2 True  n=10 mt=10             []          [8, 9]
2 False n=8  mt=30             []             [7]
```

So the exclusive end matters whenever the last stride lands exactly on `len(metrics)`, which is not a property of the step alone. My rows B, C and D were all unalignable tails, and I generalized from them to the whole larger-stride case instead of treating them as the special case they are.

Pinning the per-mode contract in `test_only_an_unalignable_tail_is_dropped` rather than a universal is the right shape, and leaving `sequence_picking_order` at its default is fine by me for the reason you gave: the ordering runs before the packing loop.

vineethsaivs:
Thank you for re-running it rather than taking my table on faith, and for the `sequence_picking_order` call. Nothing outstanding on my side: `f22b2b4` is the current head, `32 passed` on it and `14 failed, 18 passed` against the merge base.

tests/unit/runtime/test_variable_batch_size_and_lr.py
vineethsaivs: Agreed, and fixed in f22b2b4f. Sorry for the slow reply.

The universal claim is gone. The tests now parametrize both arguments and assert what actually holds per mode:
- `test_every_sample_is_batched`: `effective_batch_size=1`, both modes, nothing is trimmed, so every sample must appear.
- `test_larger_effective_batches_also_reach_the_last_sample`: your case D shape and four others where the last stride lands exactly on `len(metrics)`.
- `test_only_an_unalignable_tail_is_dropped`: your cases B and C, where the trim is deliberate. Asserts the survivors are the leading ones in order, the count is a multiple of the effective batch size, and nothing exceeds `max_tokens`.

PR #8350, opened 2026-08-29
perf(rollout): profile prefill and decode forwards

nathon-lee:
## Summary
Extend the opt‑in HybridEngineRollout profiling with:
- prefill_forward_ms
- decode_forward_ms
- generation_overhead_ms
- num_decode_forwards

Accelerator events are recorded around top‑level model forwards without synchronizing every decode step. Existing profiling fields and rollout behavior remain unchanged.

For CUDA graph generation, forward timings are reported as unavailable because graph replay bypasses model forward hooks.

Follow‑up to #8197.

## Validation
- Focused rollout tests: 25 passed, 1 skipped
- Changed‑file pre‑commit hooks: passed
- py_compile: passed
- git diff --check: passed

## GPU Environment
- GPU: NVIDIA RTX A4500, 20,470 MiB
- Driver: 570.195.03
- CUDA: 12.8
- PyTorch: 2.9.1+cu128
- Transformers: 4.40.2
- Model: facebook/opt‑6.7b
- Dtype: FP16
- World size: 1
- ZeRO stage: 0
- Batch size: 1
- Samples per prompt: 4
- Prompt length: 512
- Warmup: 5
- Iterations: 20
- Response Length: 32

| Metric | Baseline | Shared prefill | Change |
|---|---|---|---|
| Prefill forward | 438.91 ms | 130.15 ms | -70.35% |
| Decode forward | 863.04 ms | 865.98 ms | +0.34% |
| Generation overhead | 9.60 ms | 14.19 ms | +47.72% |
| Generation | 1311.55 ms | 1010.32 ms | -22.97% |
| Total | 1312.11 ms | 1010.90 ms | -22.96% |
| Throughput | 97.55 tok/s | 126.62 tok/s | +29.80% |
| Peak memory | 13144.63 MiB | 13178.63 MiB | +34 MiB |

The measurements show that shared prefill reduces prefill latency by about 70%, while decode latency remains nearly unchanged.

Thank you for reviewing!


deepspeed/runtime/rollout/hybrid_engine_rollout.py
```diff
@@ -21,6 +21,48 @@
...
+        self.active_start = self._record_time()
+
+    def _end_forward(self, _module, _args, _output):
+        end = self._record_time()
+        self.measurements.append((self.active_start, end))
+        self.active_start = None
+
+    def _record_time(self):
+        if self.use_host_timer:
+            return time.perf_counter()
```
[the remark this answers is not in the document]
nathon-lee: resolved

tests/unit/runtime/rollout/test_hybrid_engine_rollout.py
```diff
@@ -113,6 +131,83 @@ def test_generate_records_profile_when_enabled(mock_get_accelerator, mock_perf_c
     assert mock_get_accelerator.return_value.synchronize.call_count == 4
 
 
+@patch("deepspeed.runtime.rollout.hybrid_engine_rollout.time.perf_counter")
```
delock: It occurs to me we don't need to have these UTs which actually running on CPU workflow.  Remove the tests instead of mocking will be better choice.

nathon-lee: Thanks @delock. I removed the CPU-only mocked forward-profiler unit tests. The production profiling implementation and profile schema remain unchanged; accelerator-event timing will be validated in the GPU profiling run.

delock (approved):

PR #8351, opened 2026-08-29
Assert the scattered sequence length is divisible by the SP size

### LINKED ISSUE #7384: [BUG] Ulysses DistributedAttention silently produces incorrect output when #GPUs does not divide global sequence length
**Describe the bug**
When using `deepspeed.sequence.DistributedAttention` on sequences with length N where `N % G != 0` and `G` is the number of GPUs the output will silently be corrupted without throwing an error.

**To Reproduce**

1. Save this as `ds_distributed_att_varlen.py`
```
import os
import deepspeed
import torch
import torch.nn.functional as F
from torch.distributed.device_mesh import init_device_mesh

from deepspeed.sequence.layer import DistributedAttention

def _gather_variable_tensor(t: torch.Tensor, dim: int = 0, group=None):
    """Gather tensors with different size in ``dim`` using padding + all_gather.

    Returns the concatenated tensor on rank-0 and ``None`` on the others.
    """
    if group is None:
        group = dist.group.WORLD

    world_size_local = dist.get_world_size(group)
    rank = dist.get_rank(group)

    # 1. Share the size of ``dim`` across ranks.
    local_len = torch.tensor([t.shape[dim]], device=t.device, dtype=torch.long)
    len_list = [torch.zeros_like(local_len) for _ in range(world_size_local)]
    dist.all_gather(len_list, local_len, group=group)
    lens = [int(l.item()) for l in len_list]
    max_len = max(lens)

    # 2. Pad to the maximum length so that shapes match.
    pad_shape = list(t.shape)
    pad_shape[dim] = max_len
    t_padded = torch.zeros(pad_shape, dtype=t.dtype, device=t.device)
    slc = [slice(None)] * t.ndim
    slc[dim] = slice(0, t.shape[dim])
    t_padded[tuple(slc)] = t

    # 3. All-gather the padded tensors.
    gather_list = [torch.empty_like(t_padded) for _ in range(world_size_local)]
    dist.all_gather(gather_list, t_padded, group=group)

    # 4. On rank-0, unpad and concatenate. Other ranks drop the data.
    if rank == 0:
        parts = []
        for i, g in enumerate(gather_list):
            slc_i = [slice(None)] * t.ndim
            slc_i[dim] = slice(0, lens[i])
            # Keep the data on the original device (no .cpu()).
            parts.append(g[tuple(slc_i)])
        return torch.cat(parts, dim=dim)
    else:
        return None


def context_parallel_sdpa_example(world_size: int, rank: int):
    print(f"Rank {rank}, world_size {world_size}")
    assert torch.cuda.is_available()
    assert dist.is_nccl_available()
    torch.cuda.manual_seed(0)

    deepspeed.init_distributed("nccl")

    device_mesh = init_device_mesh(
        device_type="cuda", mesh_shape=(world_size,), mesh_dim_names=("cp",)
    )
    group = device_mesh.get_group("cp")

    device = torch.device("cuda", rank)

    batch = 8
    nheads = 8
    qkv_len = world_size * 2 + 1  # Make not a multiple of #GPUs.
    dim = 64
    # backend = SDPBackend.CUDNN_ATTENTION
    torch.backends.cuda.enable_math_sdp(False)

    dtype = torch.float32

    qkv = [
        torch.rand(
            (batch, nheads, qkv_len, dim),
            dtype=dtype,
            requires_grad=True,
            device=device,
        )
        for _ in range(3)
    ]
    out = F.scaled_dot_product_attention(*qkv, is_causal=False)
    # make a clean copy of QKV for output comparison
    cp_qkv = [t.detach().clone() for t in qkv]

    # Broadcast to all ranks
    for t in cp_qkv:
        dist.broadcast(t, src=0, group=group)
    print(f"[RANK {rank}] Broadcasted QKV")

    cp_qkv = [torch.tensor_split(t, world_size, dim=2)[rank] for t in cp_qkv]

    for t in cp_qkv:
        print(f"[RANK {rank}] Chunked QKV shape: {t.shape}")

    # Run distributed attention
    dist_attn = DistributedAttention(F.scaled_dot_product_attention, group, gather_idx=2, scatter_idx=1)
    cp_out = dist_attn(*cp_qkv, is_causal=False, batch_dim_idx=0)
    print(f"[RANK {rank}] DistributedAttention ran.")

    cp_out_full = _gather_variable_tensor(cp_out, dim=2, group=group)
    print(f"[RANK {rank}] Gathered output across ranks.") 

    # Validate equivalence with the baseline implementation.
    if rank == 0:
        print(cp_out_full.device, out.device)
        print(
            f"Rank {rank}: cp_out_full.shape {cp_out_full.shape}, baseline out.shape {out.shape}"
        )
        print(cp_out_full - out)
        assert torch.allclose(
            cp_out_full,
            out,
            atol=(1e-08 if dtype == torch.float32 else 1e-08 * world_size),
        )

        # If we reached here the test has passed on this rank.
        print(f"Rank {rank}: DistributedAttention test passed ✅")


if __name__ == "__main__":
    rank = int(os.environ.get("RANK", 0))
    world_size = int(os.environ.get("WORLD_SIZE", 1))

    context_parallel_sdpa_example(world_size, rank)

    dist.destroy_process_group()
```

2. Run the above script with `torchrun --nproc_per_device 2 ds_distributed_att_varlen.py`.

**Expected behavior**
I would have expected for an error to be thrown if this is outside the specification of the implementation or the output should actually be correct and coincide with the single GPU version.

**ds_report output**
```
cpu_lion ............... [NO] ....... [OKAY]
dc ..................... [NO] ....... [OKAY]
 [WARNING]  Please specify the CUTLASS repo directory as environment variable $CUTLASS_PATH
evoformer_attn ......... [NO] ....... [NO]
 [WARNING]  NVIDIA Inference is only supported on Ampere and newer architectures
 [WARNING]  FP Quantizer is using an untested triton version (3.3.1), only 2.3.(0, 1) and 3.0.0 are known to be compatible with these kernels
fp_quantizer ........... [NO] ....... [NO]
fused_lion ............. [NO] ....... [OKAY]
compiler_compat/ld: 
gds .................... [NO] ....... [NO]
inference_core_ops ..... [NO] ....... [OKAY]
cutlass_ops ............ [NO] ....... [OKAY]
ragged_device_ops ...... [NO] ....... [OKAY]
ragged_ops ............. [NO] ....... [OKAY]
 [WARNING]  sparse_attn requires a torch version >= 1.5 and < 2.0 but detected 2.7
 [WARNING]  using untested triton version (3.3.1), only 1.0.0 is known to be compatible
torch version .................... 2.7.1+cu126
deepspeed info ................... 0.17.1, unknown, unknown
torch cuda version ............... 12.6
nvcc version ..................... 12.1
deepspeed wheel compiled w. ...... torch 2.3, cuda 12.1
```

**Screenshots**

 - OS: Linux version 5.4.0-80-generic (buildd@lcy01-amd64-030) (gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04)) #90-Ubuntu SMP Fri Jul 9 22:49:44 UTC 2021
 - Tested with 2x and 8x V100 
 - Interconnects: NVLink
 - Python version: 3.11

**Launcher context**
`torchrun`

**Docker context**
Running on barebones node.

**Additional context**
-

adarshsm:
Refs #7384. @stas00 — this is the "original implementation" half of your reply, plus the
answer to what you asked.

## The new implementation does already check

`UlyssesSPDataLoaderAdapter.refill()` validates it before sharding
(`deepspeed/runtime/sequence_parallel/ulysses_sp.py`):

```python
if seq_length % self.sp_world_size != 0:
    raise ValueError(f"batch's seqlen={seq_length} isn't divisible by sp-size={self.sp_world_size}")
```

So the new path is fine. `deepspeed/sequence/layer.py` has no equivalent.

## What this changes

`_generate_layout_params` asserts the head count divides evenly in both branches that
scatter heads:

```python
assert num_total_head % seq_world_size == 0, f"Number of heads ({num_total_head}) must be divisible by the sequence parallel size ({seq_world_size})!"
```

The two branches that scatter the *sequence* had no equivalent — they just computed
`global_seq_len // seq_world_size`, so a length that is not a multiple of the SP size was
silently truncated and the all2all buffers then disagreed. This adds the matching
assertion, so it now fails the same way a head mismatch already does:

```
AssertionError: Sequence length (9) must be divisible by the sequence parallel size (2)!
```

## What it does not cover, and why

**This does not fully close #7384**, which is why the commit says `Refs` rather than
`Fixes`.

The reporter pre-shards with `torch.tensor_split(t, world_size, dim=2)[rank]` and passes
`scatter_idx=1, gather_idx=2`. With a global length of 9 over 2 ranks that hands rank 0 a
size-5 shard and rank 1 a size-4 shard, so the ranks disagree on the **gather** dimension.
No rank can detect that locally — each one's own shard looks perfectly valid — and the
mismatch only surfaces inside `all_to_all_single`, where gloo aborts the process group with
no Python traceback at all.

Catching that case needs a cross-rank check, e.g. an `all_gather` of the local shape
compared across ranks before the all2all. That is a collective on every all2all call, so it
is a performance trade I did not want to make on your behalf. Happy to add it behind a flag,
or only under a debug/validation mode, if you want it.

## Verification

Two ranks over gloo on CPU:

| case | before | after |
|---|---|---|
| seq 8, world 2, full sequence at the scattered dim | works, matches single-device attention | unchanged |
| seq 9, world 2, full sequence at the scattered dim | process group aborts, no message | `AssertionError` naming both numbers |
| seq 9, world 2, pre-sharded unevenly (the report) | aborts | still aborts — see above |

The divisible case was checked against single-device `scaled_dot_product_attention` on the
same input to confirm the harness itself is sound before trusting the failing case.

`yapf --style .style.yapf` reports no diff; the two new lines are the same shape and length
as the head assertions directly below them.

I could not run `tests/unit/sequence_parallelism/` here — no NVIDIA GPU on this machine — so
that is worth a look in CI.

### DIFF
--- a/deepspeed/sequence/layer.py
+++ b/deepspeed/sequence/layer.py
@@ -30,6 +30,7 @@ def _generate_layout_params(scatter_idx, batch_dim_idx, seq_world_size, input):
     if batch_dim_idx == 0:
         if scatter_idx < 2:
             bs, global_seq_len, num_local_head, head_dim = input.shape
+            assert global_seq_len % seq_world_size == 0, f"Sequence length ({global_seq_len}) must be divisible by the sequence parallel size ({seq_world_size})!"
             pre_all2all_inp_shape = [bs, seq_world_size, global_seq_len // seq_world_size, num_local_head, head_dim]
             pre_all2all_permute_idx = (1, 0, 2, 3, 4)
 
@@ -46,6 +47,7 @@ def _generate_layout_params(scatter_idx, batch_dim_idx, seq_world_size, input):
     else:
         if scatter_idx < 2:
             global_seq_len, bs, num_local_head, head_dim = input.shape
+            assert global_seq_len % seq_world_size == 0, f"Sequence length ({global_seq_len}) must be divisible by the sequence parallel size ({seq_world_size})!"
             pre_all2all_inp_shape = [seq_world_size, global_seq_len // seq_world_size, bs, num_local_head, head_dim]
             pre_all2all_permute_idx = None
 


stas00:
I don't know if it's worthwhile investing into an outdated implementation. Perhaps it should be deprecated in favor of the new version that has no such bugs?

But in any case if you are fixing the old implementation please tag the original authors. Perhaps they can also decide what to do with it.

adarshsm:
You're right, and I should have reached that conclusion myself before opening this.

Your first reply already contained the answer — you suspected the new implementation handled it, and my own investigation confirmed it does (`UlyssesSPDataLoaderAdapter.refill()` raises before sharding). At that point the honest read was that the remaining gap is in a path you would rather retire, not one worth hardening. I had the disqualifying evidence and opened the PR anyway, which puts a review burden on you for marginal value. Compounding it, the guard does not even close #7384: the reported failure is ranks disagreeing on the gather dimension after an uneven pre-shard, which no local assertion can detect.

Closing this. If you do want the deprecation route instead — a warning on `deepspeed.sequence.DistributedAttention` pointing at `runtime/sequence_parallel`, plus a docs note — I am happy to write that, and it would make the old path's status explicit for the frameworks still on it. Just say the word and I will open it against whatever removal timeline you want.

Thanks for the straight answer rather than letting it sit.

PR #8352, opened 2026-08-29
perf(rollout): combine shared prefill with CUDA Graph decode

nathon-lee:
## Summary

Enable native HybridEngine CUDA Graph decode for fixed-length rollout generation and allow it to compose with shared prefill.

The rollout-level StaticCache graph path remains incompatible with shared prefill.

Depends on #8350.

## Validation

- Shared-prefill graph/eager output equivalence: 1 passed
- Changed-file pre-commit hooks: passed
- OPT-6.7B FP16, RTX A4500, world size 1
- Batch 1, samples 4, prompt 512, response 128
- Warmup 5, iterations 20

| Metric | Eager | Shared | Graph | Shared + graph |
|---|---:|---:|---:|---:|
| Prefill (ms) | 455.899 | 133.497 | 456.884 | 133.824 |
| Decode (ms) | 3606.478 | 3606.113 | 3518.189 | 3516.792 |
| Total (ms) | 4104.628 | 3784.307 | 4012.964 | 3692.591 |
| Throughput (tok/s) | 124.738 | 135.296 | 127.587 | 138.656 |
| Peak memory (MiB) | 13144.629 | 13178.629 | 13203.820 | 13237.820 |

Compared with eager, the combined path reduces total latency by 10.04% and improves throughput by 11.16%, with about 93 MiB additional peak memory.

Thank you for reviewing!

### DIFF
--- a/deepspeed/runtime/hybrid_engine.py
+++ b/deepspeed/runtime/hybrid_engine.py
@@ -198,8 +198,6 @@ class DeepSpeedHybridEngine(DeepSpeedEngine):
             raise RuntimeError("Shared prefill does not support inference tensor parallelism")
         if hybrid_config.release_inference_cache:
             raise RuntimeError("Shared prefill does not support release_inference_cache")
-        if hybrid_config.enable_cuda_graph:
-            raise RuntimeError("Shared prefill does not support CUDA graph capture")
         if len(self._inference_containers) == 0:
             raise RuntimeError("Shared prefill requires HybridEngine inference containers")
 
--- a/deepspeed/runtime/rollout/hybrid_engine_rollout.py
+++ b/deepspeed/runtime/rollout/hybrid_engine_rollout.py
@@ -27,10 +27,16 @@ class _ForwardProfiler:
     def __init__(self, accelerator):
         self.event_type = accelerator.Event
         self.use_host_timer = accelerator.use_host_timers()
+        # Host intervals are only reliable when there is no asynchronous device
+        # work, as on CPU. MPS exposes events but cannot time them, so its host
+        # intervals would measure enqueue latency rather than model execution.
+        self.is_available = not self.use_host_timer or self.event_type is None
         self.active_start = None
         self.measurements = []
 
     def register(self, module):
+        if not self.is_available:
+            return ()
         pre_handle = module.register_forward_pre_hook(self._start_forward)
         post_handle = module.register_forward_hook(self._end_forward)
         return pre_handle, post_handle
--- a/deepspeed/runtime/rollout/hybrid_engine_rollout.py
+++ b/deepspeed/runtime/rollout/hybrid_engine_rollout.py
@@ -152,6 +152,7 @@ class HybridEngineRollout(RolloutEngine):
                     prompt_ids,
                     attention_mask=prompt_attn,
                     max_new_tokens=max_new_tokens,
+                    min_new_tokens=max_new_tokens,
                     # ZeRO-3 gathers parameters during each decode forward, so every
                     # data-parallel rank must execute the same number of iterations.
                     eos_token_id=None,
--- a/deepspeed/runtime/rollout/hybrid_engine_rollout.py
+++ b/deepspeed/runtime/rollout/hybrid_engine_rollout.py
@@ -21,6 +21,48 @@ from deepspeed.accelerator import get_accelerator
 from deepspeed.runtime.rollout.base import RolloutBatch, RolloutEngine, RolloutRequest, SamplingConfig
 
 
+class _ForwardProfiler:
+    """Measure top-level model forwards without synchronizing each call."""
+
+    def __init__(self, accelerator):
+        self.event_type = accelerator.Event
+        self.use_host_timer = accelerator.use_host_timers()
+        self.active_start = None
+        self.measurements = []
+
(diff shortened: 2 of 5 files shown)


nathon-lee:
# Controlled Rollout Performance Attribution on NVIDIA H200

## Experimental setup

- Hardware: NVIDIA H200
- Model: GPT-2 (`/data/models/gpt2`)
- Precision: BF16
- Tensor parallelism: 1 GPU per backend
- Batch size: 1
- Prompt length: 128 tokens
- Response length: 32 tokens
- Sampling: greedy decoding (`temperature=0`, `top_p=1`)
- Warmup: 2 iterations
- Measured iterations: 5
- Seed: 1234
- Prefix/radix cache: disabled
- All successful runs returned exactly 32 tokens
- Engine startup and CUDA Graph capture time are excluded from steady-state throughput

The experiments used the same model, input shape, sampling parameters, and token budget. For the parallel repeat, OPD, vLLM, and SGLang were placed on separate idle H200 GPUs.

## Results

| Configuration | OPD / DeepSpeed HybridEngine | verl / vLLM | slime / SGLang |
|---|---:|---:|---:|
| Eager execution, no cache | 222.3 tok/s | 207.7 tok/s | 205.1 tok/s |
| SGLang CUDA Graph enabled | 218.6 tok/s | 210.2 tok/s | **964.6 tok/s** |
| vLLM non-eager path enabled | 222.4 tok/s | **766.6 tok/s** | 208.2 tok/s |
| Mean latency, eager/no-cache | 144.0 ms | 154.1 ms | 156.0 ms |
| Mean latency, SGLang Graph | 146.4 ms | 152.2 ms | **33.2 ms** |
| Mean latency, vLLM Graph/compile | 143.9 ms | **41.8 ms** | 153.7 ms |

All successful backend runs completed with exit code 0.

## Attribution

With CUDA Graph disabled and caching disabled, all three backends were in the same performance range: approximately 205–222 tok/s. OPD was slightly faster than both vLLM and SGLang in this controlled eager baseline.

Enabling SGLang CUDA Graph increased SGLang throughput from 205.1 tok/s to 964.6 tok/s, a 4.70x improvement. The OPD and vLLM controls remained around 219–210 tok/s.

Enabling the vLLM non-eager path increased vLLM throughput from 207.7 tok/s to 766.6 tok/s, a 3.69x improvement. In this vLLM configuration, the gain includes both torch.compile and CUDA Graph execution. The OPD and SGLang controls remained around 222–208 tok/s.

These results indicate that, for this batch-1 decode workload, host-side kernel launch and dispatch overhead is a dominant bottleneck. The large performance gap observed in the original comparison is primarily an execution-mode difference, rather than an inherent model, hardware, or sampling advantage of SGLang.

## OPD CUDA Graph status

The OPD CUDA Graph path was tested through the explicit `HybridEngineRollout(use_graph_capture=True)` path. The run did not produce a throughput result because GPT-2 failed during the prefill stage with:

```text
TypeError: 'StaticCache' object is not subscriptable
```

This is a Transformers/StaticCache compatibility issue in the current OPD Graph path, not a measured performance regression.

## Limitations

- The workload uses a small GPT-2 model and batch size 1, which intentionally emphasizes CPU/kernel-launch overhead.
- The vLLM result combines torch.compile and CUDA Graph effects; it should not be interpreted as a pure CUDA Graph-only measurement.
- Engine-specific optimized kernels remain enabled. Therefore, within-backend Graph-on versus Graph-off deltas are the strongest attribution signal.
- Results should be reproduced with larger models, larger batches, and variable sequence lengths before making a general throughput claim.

## Why this CUDA Graph PR matters

This PR addresses a performance-critical execution path rather than a speculative micro-optimization.

In a controlled NVIDIA H200 batch-1 decode experiment with the same model, prompt length, response length, sampling parameters, and disabled prefix/radix caching, all eager baselines achieved only 205–222 tok/s. Enabling SGLang CUDA Graph increased throughput from 205.1 tok/s to 964.6 tok/s (4.70x). Enabling the vLLM non-eager path, which activates torch.compile and CUDA Graph execution, increased throughput from 207.7 tok/s to 766.6 tok/s (3.69x), while the other backend controls remained unchanged.

This isolates CUDA Graph/compiled execution as a major factor in the batch-1 decode bottleneck: the GPU is not compute-bound, and host-side kernel launch overhead dominates. The current OPD Graph path still exposes a concrete `StaticCache` compatibility issue, so merging this PR would address both an observed performance gap and a reproducible integration blocker. I suggest merging the feature behind the existing opt-in flag, followed by follow-up work for dynamic sequence lengths, graph bucketing, and broader model coverage.

Given the controlled attribution results above, I would appreciate a review and merge of this PR. The feature is opt-in, the performance benefit is measurable under a reproducible workload, and the remaining limitations are clearly isolated for follow-up work.

## Reproduction artifacts

- `parallel_fair_eager_nocache_20260922T1200Z/`
- `parallel_vllm_graph_nocache_20260922T1210Z/`
- `attribution_sglang_graph_nocache_20260922T1110Z/`
- `attribution_opd_graph_real_20260922T1140Z/runner.log`
- `run_attribution_parallel_20260922.sh`

nathon-lee:
image

nathon-lee:
image

nathon-lee:
image

PR #8359, opened 2026-08-29
Overlap AutoEP split planning with token packing

yh0903:
## Summary
- add an opt-in `expert_parallel.async_split_plan` path for standard AutoEP
- keep count AllToAll and split-size computation on the caller stream; only pinned metadata D2H uses the cached copy stream, while token sorting and packing stay on the caller stream
- wait only immediately before payload dispatch, preserve the existing synchronous default, and reject unsupported AutoEP+AutoTP folding
- start asynchronous planning only for the standard communication backend and drain newly owned pending work after failed forwards

## Validation
- [x] 113 targeted CPU cases pass, including inference-mode/no-grad warmup followed by training
- [x] Current-head two-H100 inference-to-training parity: 4/4 cases pass, with no failures, errors, or skips
- [x] `pre-commit run --files` passed for all changed files
- [x] Current-head GPU coverage includes activation checkpointing on/off and default/non-default caller streams, each after inference-mode warmup
- [x] parity covers loss, output, input gradient, router gradients, expert gradients, and optimizer parameter deltas
- [x] clean E2E ABBA+BAAB completed on two upstream-stack and two PR1+PR2 combined-stack allocations
- [x] sparse split-plan attribution completed separately from the clean E2E runs

### Historical clean E2E (original broad-overlap design)

These timings predate the D2H-only revision and do not establish a speedup for the current implementation.

These runs used the pre-review PR head implementation, Qwen3-30B-A3B, 48 layers, EP16, sequence length 1024, activation checkpointing, and the same checkpoint/input/seed. `time-step-phases` and the component profiler were completely disabled.

Each allocation discarded one fixed warm arm, then ran ABBA followed by BAAB: four runs per arm, with 30 measured steps per run.

| Stack | Allocation | Sync median | Async median | Delta |
| --- | --- | ---: | ---: | ---: |
| upstream baseline | 1 | 975.59 ms | 974.10 ms | -1.49 ms (-0.2%) |
| upstream baseline | 2 | 989.63 ms | 978.11 ms | -11.52 ms (-1.2%) |
| PR1+PR2 combined | 1 | 990.45 ms | 976.60 ms | -13.85 ms (-1.4%) |
| PR1+PR2 combined | 2 | 983.81 ms | 972.47 ms | -11.34 ms (-1.2%) |

All four allocations favored asynchronous split planning. The median of the two allocation medians improves by 6.51 ms (0.7%) on the upstream stack and 12.60 ms (1.3%) on the PR1+PR2 combined stack. With only two allocations per stack, these are performance estimates rather than a claim of a fixed universal speedup.

### Sparse attribution

A separate combined-stack run sampled only the first MoE layer. Activation checkpointing invoked that layer twice per step (forward and replay), so warmup and aggregation account for two records per step.

| Metric | Median per invocation | Median summed per step |
| --- | ---: | ---: |
| count A2A | 0.690 ms | 1.501 ms |
| count A2A through D2H completion | 0.728 ms | 1.579 ms |
| split reductions/stack plus D2H | 0.108 ms | 0.175 ms |
| sort/pack | 0.250 ms | 0.397 ms |
| residual host wait before dispatch | 2.989 ms | 6.080 ms |

The original D2H range includes split reductions and stacking, rather than isolating the metadata copy. Residual host wait also includes outstanding GPU dependencies and rank-arrival effects. These inclusive intervals overlap and do not isolate pure NCCL or peer-arrival cost; this change does not remove the full historical split-sync envelope.

An alternative implementation using `async_op=True` plus `Work.wait()` regressed a same-allocation L48 comparison by 2.1%, so that variant was removed.

## Inference-mode review follow-up

Commit `d6874193cab3ad3a8a1c09d95f515dd7874885c8` allocates the persistent split buffer, including pinning, inside `torch.inference_mode(False)`. A first inference-mode validation or warmup forward can therefore be followed by a training forward/backward using the same cached buffer. The mode override is limited to its one-time allocation.

The regression test reproduces the reported in-place inference-tensor error before the fix, while its `no_grad` control passes. With the fix, all 113 selected CPU cases and changed-file pre-commit pass. The four GPU cases retain output, input/router/expert gradient and FP32 optimizer-update comparisons against the synchronous path after inference-mode warmup.

GPU validation of this exact commit passed all four cases on 2 × NVIDIA H100 80GB HBM3, using PyTorch 2.10.0.2+cu130, CUDA 13.0 and NCCL 2.28.9. Each case uses one MoE layer, EP2/FP16, an inference-mode warmup and three training updates, with checkpointing on/off crossed with default/non-default caller streams. There were zero failures, errors or skips. This is a targeted correctness regression, not a performance measurement.


PKUWZP:
@yh0903 Thanks for submitting the PR.  It seems that the D2H copy is not a bottleneck, do you have any insights on how to resolve the true bottlenecks?

tohtana (approved):
Thank you @yh0903! The code change looks good to me. 
When I tried in a very small setting (2 GPU/4 layers/EP2/FP16/4096 tokens/rank), the performance was ~9% slower. This is an opt-in feature and I don't think this is a blocking issue, but we'll need to figure out when this feature would work well.

deepspeed/module_inject/auto_ep_layer.py
```diff
@@ -163,6 +200,58 @@ def _split_plan_from_expert_counts(
...
+            send_counts,
+            group=ep_group,
+        )
+        received_counts = received_counts_flat.view(ep_size, num_local_experts)
+        local_counts = received_counts.sum(dim=0)
+        device_splits = torch.stack((
+            count_matrix.sum(dim=1),
+            received_counts.sum(dim=1),
+        ))
+        host_splits.copy_(device_splits, non_blocking=True)
```
hwchen2017, line 243: It tries to overlap the some tensor computation, all-to-all comm, and gpu to cpu data copy with the sort/packing. But computation kernel can not really be overlapped.  And the all-to-all can take much longer time than the spot/packing. That may explain the performance regression. 

From my profiling, the opportunity here is to make H2D memory copy async and overlappe with some computational kernels, which can reduce the GPU-CPU sync overhead. This is on my to-do-lists, and I’m glad you identified the same opportunity.

I prefer to have such design:
```
Router -> Count AllToAll -> Compute split sizes -> Main stream: Sort/pack --------> token AllToAll
                                                -> Copy stream: Metadata D2H ------>
                                                                                   | Sync
```
I think it's less likely to have perf regression.

hwchen2017 (changes requested):

yh0903:
> Thank you @yh0903! The code change looks good to me. When I tried in a very small setting (2 GPU/4 layers/EP2/FP16/4096 tokens/rank), the performance was ~9% slower. This is an opt-in feature and I don't think this is a blocking issue, but we'll need to figure out when this feature would work well.

Thanks, @tohtana. I tested a similar small configuration on two H100s (4 layers, EP2, FP16, 4096 tokens/rank) and also observed a modest regression (about 1% in median step time in the initial D2H-only test).
 
I think it's because the metadata payload is tiny at EP2 and the packing window is short, so the extra stream/event and submission overhead could outweigh the time hidden by overlap. I agree that this should remain opt-in, with the benefit evaluated for the target workload likely for relatively larger models.

yh0903:
> @yh0903 Thanks for submitting the PR. It seems that the D2H copy is not a bottleneck, do you have any insights on how to resolve the true bottlenecks?

Thanks @PKUWZP! I agree that the metadata copy itself is unlikely to explain the main bottlenecks. The original profiling range also included split reductions and stacking, so it was not an isolated measurement of the D2H copy.

In parallel, we have been addressing other sources of overhead: #8326  fuses weighted token restore/combine; #8331 skips unused gradient-norm computation and identity unscaling while retaining overflow checks; and #8380 adds regional compilation of the non-MoE parts of AutoEP decoder layers to reduce eager execution overhead, while keeping routing, expert computation and communication eager.

This PR specifically targets the exposed host wait for split metadata. The remaining communication and checkpoint/recompute costs need separate profiling to determine where further optimization would help most.

deepspeed/module_inject/auto_ep_layer.py
yh0903: Thanks for the insightful feedback @hwchen2017 ! I've updated the implementation to follow your proposed scheduling: count AllToAll and split-size computation stay on the caller stream, and only the pinned metadata D2H copy runs on the copy stream, overlapping with token sorting and packing. The host waits for copy completion immediately before token AllToAll.

I also changed the pinned buffer to int64 to avoid an implicit dtype conversion, reused the dependency event to reduce setup overhead, and removed an unused count reduction. Buffer lifetime tracking and failed-forward cleanup are preserved.
The feature remains disabled by default, since narrowing the overlap does not guarantee a speedup for small workloads. Thanks again for suggesting this design!

tohtana:
Thank you for the clarification, @yh0903!
@hwchen2017 Does the update address your concern?

hwchen2017:
Looks good to me. I ask AI to review the code again, and here is one comment from it:

With  `async_split_plan=True`  and EP > 1, a first forward inside  `torch.inference_mode()`  creates the cached pinned buffer as an inference tensor. The next forward outside that context fails at `host_splits.copy_()`  with  `Inplace update to inference tensor outside InferenceMode is not allowed` . This breaks an inference-mode validation or warm-up pass followed by training; the synchronous path is unaffected.

Allocate the persistent buffer inside ` torch.inference_mode(False)` so its lifetime is independent of the mode used for the first forward.

@yh0903 Can you fix the minor problem?

yh0903:
> Looks good to me. I ask AI to review the code again, and here is one comment from it:
> 
> With `async_split_plan=True` and EP > 1, a first forward inside `torch.inference_mode()` creates the cached pinned buffer as an inference tensor. The next forward outside that context fails at `host_splits.copy_()` with `Inplace update to inference tensor outside InferenceMode is not allowed` . This breaks an inference-mode validation or warm-up pass followed by training; the synchronous path is unaffected.
> 
> Allocate the persistent buffer inside ` torch.inference_mode(False)` so its lifetime is independent of the mode used for the first forward.
> 
> @yh0903 Can you fix the minor problem?

Thanks for catching this! I reproduced the error and fixed it in latest commit by wrapping both the cached host buffer allocation and pinning in `torch.inference_mode(False)`. This keeps the buffer writable when training follows an inference-mode warmup, while leaving the overlap scheduling unchanged.

hwchen2017 (approved):

yh0903:
Thank you @hwchen2017 @tohtana!! All CIs are green now, could you help merge the PR if it looks good? Thanks

PR #8365, opened 2026-08-30
feat(cpu-adam): add ARM SVE update kernel

xylian86:
`DeepSpeedCPUAdam` has hand-vectorized AVX2 and AVX-512 implementations, but AArch64 systems currently use the scalar fallback. This leaves CPU optimizer performance on NVIDIA Grace and other SVE-capable Arm CPUs below the available memory and vector throughput, which directly affects ZeRO-Offload workloads.

## Summary

- add a vector-length-agnostic ARM SVE path for FP32 `DeepSpeedCPUAdam`
- retain the scalar fallback for unsupported data types and non-SVE Arm CPUs

## Performance

Three warmup steps, ten timed steps per trial, median of three trials, FP32 parameters/states.

| Parameters | Scalar CPU-Adam | SVE CPU-Adam | Speedup |
|---:|---:|---:|---:|
| 1B | 0.163 s | 0.125 s | 1.30x |
| 2B | 0.322 s | 0.215 s | 1.50x |
| 4B | 0.755 s | 0.487 s | 1.55x |


op_builder/builder.py
```diff
@@ -506,6 +510,8 @@ def simd_width(self):
                 return '-D__AVX512__'
             elif 'avx2' in cpu_info['flags']:
                 return '-D__AVX256__'
+        elif (cpu_info['arch'] or '').startswith('ARM') and 'sve' in cpu_info['flags']:
+            return '-D__SVE__'
```
delock, line 514: Is macro `__SVE__` actually used in the code?

tests/unit/ops/test_op_builder.py
```diff
@@ -49,6 +49,23 @@ def make_builder(**overrides):
...
+    ({
+        'arch': 'ARM_8',
+        'flags': ['sve', 'sve2']
+    }, '-D__SVE__'),
+    ({
+        'arch': 'ARM_8',
+        'flags': []
+    }, '-D__SCALAR__'),
+])
+def test_simd_width_detects_arm_sve(cpu_info, expected):
```
delock: This test has strong coupling with implementation, should remove this test.

delock:
Hi @xylian86 , thanks for your PR.  Overall the change looks good to me, except some minor questions I left in comments, thanks!

delock (approved):

op_builder/builder.py
xylian86: Good catch. Addressed in 0d17e165: all SVE guards now require both the builder-defined __SVE__ macro and the compiler-defined __ARM_FEATURE_SVE.

tests/unit/ops/test_op_builder.py
xylian86: Yes, I have removed it.

xylian86:
> Hi @xylian86 , thanks for your PR. Overall the change looks good to me, except some minor questions I left in comments, thanks!

Thanks for the review! I’ve addressed both comments.

PR #8370, opened 2026-08-30
Validate positive inference min output tokens

AmirF194:
## Description

`DeepSpeedInferenceConfig.min_out_tokens` has no lower-bound validation, so a negative or
zero value passes Pydantic construction and is only ever caught later, deep in
`GenWorkSpace`'s C++ workspace allocation (`inference_context.h`), which takes it as an
unsigned parameter. #8343 added the same `gt=0` constraint to the architecturally
identical sibling field `max_out_tokens`; this completes it on `min_out_tokens`, which
that PR does not touch.

Refs #8339 (reports the `max_out_tokens` half of this gap, already being fixed by #8343).

## Testing

Verified with a plain Pydantic construction check, no model or GPU required:

- Before: `DeepSpeedInferenceConfig(min_out_tokens=-1)` and `DeepSpeedInferenceConfig(min_out_tokens=0)` both construct without error.
- After: both raise `ValidationError`, and `min_out_tokens=50` (or the default `1`) still constructs normally.
- New parametrized test in `tests/unit/inference/test_inference_config.py` (mirrors #8343's shape for `max_out_tokens`) fails on unmodified `master` and passes on this branch; ran with `pytest -m inference tests/unit/inference/test_inference_config.py`.
- `pre-commit run --files deepspeed/inference/config.py tests/unit/inference/test_inference_config.py`: all hooks pass.
- Not run: the CUDA/ROCm `GenWorkSpace` path itself, since this VPS has no GPU. The C++ guard at `inference_context.h:150` is untouched and stays as a defense in depth; this PR only stops the invalid value earlier, at config time.

### DIFF
--- a/deepspeed/inference/config.py
+++ b/deepspeed/inference/config.py
@@ -265,7 +265,7 @@ class DeepSpeedInferenceConfig(DeepSpeedConfigModel):
     to the required token-length required for your use-case.
     """
 
-    min_out_tokens: int = Field(1, alias="min_tokens")
+    min_out_tokens: int = Field(1, alias="min_tokens", gt=0)
     """
     This argument communicates to the runtime the minimum number of tokens you
     expect you will need to generate. This will cause the runtime to error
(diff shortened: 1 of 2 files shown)


AmirF194:
Five of the six checks here are stuck on the fork-approval gate (action_required), so nothing but DCO has actually run. Approving those would let CI clear before review.

PR #8372, opened 2026-08-31
docs: add test discipline rules to agent guidelines

delock:
## Motivation

AI coding agents increasingly submit tests alongside features, and a recurring failure mode of such contributions is tests coupled to the implementation rather than the contract:

- asserting private method return values or exact internal strings,
- re-implementing the logic under test as the test's own reference,
- mocks that stand in for internals instead of a documented collaborator contract.

Such tests pass trivially, break on harmless refactors (or worse, survive bugs they should catch), and give false coverage confidence.

## What this PR adds

A `### Test discipline` section to the agent guidelines (`AGENTS.md` + `CLAUDE.md`, kept in sync as the file header requires) with operational rules:

1. A test is well-formed only if a different correct implementation of the same contract passes it.
2. Name the concrete incorrect behavior a test would catch before writing it; if none can be named, do not write it.
3. Assert observable outcomes through public/stable interfaces; white-box pinning only for specific fixed bugs, justified in a comment.
4. Anchor to an external oracle or an independently derived reference instead of re-implementing the logic under test.
5. Mocks must stand in for a collaborator's documented contract (schema, protocol), never for internals of the module under test.

The rules are deliberately example-free to keep the guide lean.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,7 +33,7 @@
 - Anchor to an external oracle or an independently derived reference instead of re-implementing the logic under test.
 - Mocks must stand in for a collaborator's documented contract (schema, protocol), never for internals of the module under test.
 - Changes that affect the external contract at the training-loop or inference level require integration tests, not just unit tests (e.g. a minimal training loop with `SimpleModel`).
-- Integration tests must be executed on an actual device, not merely written: report the execution hardware spec and results in the PR.
+- Integration tests must be executed on actual devices, not merely written: report the execution hardware spec and results in the PR.
 
 ## Tool Caveats
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,6 +33,7 @@
 - Anchor to an external oracle or an independently derived reference instead of re-implementing the logic under test.
 - Mocks must stand in for a collaborator's documented contract (schema, protocol), never for internals of the module under test.
 - Changes that affect the external contract at the training-loop or inference level require integration tests, not just unit tests (e.g. a minimal training loop with `SimpleModel`).
+- Integration tests must be executed on an actual device, not merely written: report the execution hardware spec and results in the PR.
 
 ## Tool Caveats
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,14 @@
 - Comments should explain **why**, not **what**. Describe the purpose and reasoning, not the mechanics that the code already shows.
 - New features must include corresponding tests and documentation updates.
 
+### Test discipline
+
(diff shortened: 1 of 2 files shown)


pengdurice:
Thank you for adding this! Do we want to add something like: if the change is substantially large, add integration test? i.e. if we make changes on optimizer, instead of just unit test the optimizer, we should add a minimal training loop with a SimpleModel like training-framework-benchmarking-v5/third_party/DeepSpeed/tests/unit/checkpoint/test_universal_checkpoint.py

delock:
> Thank you for adding this! Do we want to add something like: if the change is substantially large, add integration test? i.e. if we make changes on optimizer, instead of just unit test the optimizer, we should add a minimal training loop with a SimpleModel like training-framework-benchmarking-v5/third_party/DeepSpeed/tests/unit/checkpoint/test_universal_checkpoint.py

Hi  @pengdurice , add integration test is a good idea.  Let me add it.  Thanks!

delock:
Hi @pengdurice , I added a line for integration test and tried to clearly define substaintial change.  Can you take a look?  Thanks!

pengdurice (commented):
one last question, does the agent read the PR description? I think so? do we need to add requests of adding verification on at least some hardware when the change is substantial?

AGENTS.md
```diff
@@ -25,6 +25,16 @@
...
 
+### Test discipline
+
+- Tests verify contracts, not implementations: a test is well-formed only if a different correct implementation of the same contract passes it.
+- Before writing a test, name the concrete incorrect behavior it would catch; if you cannot name one, do not write it.
+- Assert on observable outcomes through public/stable interfaces; do not assert private method return values or exact internal strings unless pinning a specific fixed bug (justify in a comment).
+- Anchor to an external oracle or an independently derived reference instead of re-implementing the logic under test.
+- Mocks must stand in for a collaborator's documented contract (schema, protocol), never for internals of the module under test.
+- Changes that affect the external contract at the training-loop or inference level require integration tests, not just unit tests (e.g. a minimal training loop with `SimpleModel`).
+- Integration tests must be executed on an actual device, not merely written: report the execution hardware spec and results in the PR.
```
pengdurice: nit: an actual device --> actual devices. Thank you!

delock:
> one last question, does the agent read the PR description? I think so? do we need to add requests of adding verification on at least some hardware when the change is substantial?

I believe most morden agents would be able to get PR description with tool call.   I add a line to require agents to report the hardware spec and result used for local validation when submit integration test.

pengdurice (approved):
Thank you for adding this test rules!

PR #8381, opened 2026-09-01
[DON'T MERGE] Run multi-rank CPU unit tests in CI via LOCAL_SIZE

delock:
_**This is an on going work. Note I'll rebase when revealed bugs are fixed. The goal is fix all issues exposed by this CI, then turn on multi-rank test in CPU workflow.**_

## Problem

On the `cpu-torch-latest` runner (single-socket `ubuntu-24.04`), every unit test that needs more than one rank is skipped:

```
SKIPPED tests/unit/common.py:275: Skipping test because not enough GPUs are available: 2 required, 1 available
```

Root cause: `CPU_Accelerator.device_count()` reports the number of NUMA nodes (`accelerator/cpu_accelerator.py`), which is 1 on the single-socket runner, and `DistributedExec._launch_procs()` in `tests/unit/common.py` gates process count on `device_count()`. Multi-rank CPU tests do not actually need one device per rank — ranks are ordinary processes communicating over gloo.

## Change

Set `LOCAL_SIZE=4` for the `unit-tests` job. `device_count()` reads `LOCAL_SIZE` first, so the launch gate now admits `world_size<=4` tests. This is the same signal the DeepSpeed launcher sets for spawned processes; the unit-test harness re-sets `LOCAL_SIZE` per worker in `_dist_run()`, so the CI-level value only affects the gate and cannot leak into test bodies.

## Evidence this is safe

- `TestDistIsendIrecv` (`tests/unit/comm/test_dist.py`, world_size=2) and the autotp universal-checkpoint test (`tests/unit/checkpoint/test_autotp_uc_checkpoint.py`, world_size=4) already bypass the per-device gate for CPU and run green in this very CI job.
- `LOCAL_SIZE` has a single reader in the codebase (`CPU_Accelerator.device_count()`); the launcher only writes it.

## Expected impact

~207 world_size=2 and ~85 world_size=4 tests that are currently skipped will now execute on CPU CI (world_size>=8 stays skipped). Some of them may have latent failures — this PR intentionally surfaces them so they can be triaged.

### DIFF
--- a/.github/workflows/cpu-torch-latest.yml
+++ b/.github/workflows/cpu-torch-latest.yml
@@ -81,6 +81,12 @@ jobs:
     runs-on: ubuntu-24.04
 
     env:
+      # The runner is single-socket, so CPU_Accelerator.device_count() reports 1
+      # NUMA node and the per-device gate in tests/unit/common.py skips every
+      # multi-rank test. CPU ranks are plain processes over gloo, so advertise 4
+      # local devices to let world_size<=4 tests run. The test harness re-sets
+      # LOCAL_SIZE per worker, so this value only affects the launch gate.
+      LOCAL_SIZE: '4'
       DEFAULT_TORCH_PRESET: '2.10.0-cpu'
       DEFAULT_TRANSFORMERS_SOURCE: 'git'
       # Manual PyPI fallback only; scheduled and default manual runs use Git.
--- a/.github/workflows/cpu-torch-latest.yml
+++ b/.github/workflows/cpu-torch-latest.yml
@@ -284,5 +284,12 @@ jobs:
         run: |
           unset TORCH_CUDA_ARCH_LIST # only jit compile for current arch
           cd tests
-          HF_HOME=/tmp/hf_home/ pytest $PYTEST_OPTS --forked -n 4 unit/ --torch_ver="$TORCH_TEST_VERSION"
-          HF_HOME=/tmp/hf_home/ pytest $PYTEST_OPTS --forked -m 'sequential' unit/ --torch_ver="$TORCH_TEST_VERSION"
+          # The suite is split so each half gets fresh xdist workers: multi-rank pool
+          # teardown eventually wedges a worker, and one 3400-test process never
+          # reaches its summary inside the 6h job limit. maxfail is raised so every
+          # failure is listed, and timeout caps each half.
+          overall=0
+          timeout 150m pytest $PYTEST_OPTS --maxfail=100000 --forked -n 4 unit/ --ignore=unit/v1 --torch_ver="$TORCH_TEST_VERSION" || overall=$?
+          timeout 150m pytest $PYTEST_OPTS --maxfail=100000 --forked -n 4 unit/v1 --torch_ver="$TORCH_TEST_VERSION" || overall=$?
+          HF_HOME=/tmp/hf_home/ pytest $PYTEST_OPTS --forked -m 'sequential' unit/ --torch_ver="$TORCH_TEST_VERSION" || overall=$?
+          exit $overall
--- a/.github/workflows/cpu-torch-latest.yml
+++ b/.github/workflows/cpu-torch-latest.yml
@@ -87,6 +87,10 @@ jobs:
       # local devices to let world_size<=4 tests run. The test harness re-sets
       # LOCAL_SIZE per worker, so this value only affects the launch gate.
       LOCAL_SIZE: '4'
+      # Multi-rank tests churn mp pools for hours; reused pools eventually hang in
+      # cleanup and stall workers until the 6h job limit. Fresh pools per test are
+      # slower but let the suite finish (knob documented in tests/unit/common.py).
+      DS_DISABLE_REUSE_DIST_ENV: '1'
       DEFAULT_TORCH_PRESET: '2.10.0-cpu'
       DEFAULT_TRANSFORMERS_SOURCE: 'git'
       # Manual PyPI fallback only; scheduled and default manual runs use Git.


delock:
This is an on going work.  Note I'll rebase when revealed bugs are fixed.  The goal is fix all issues exposed by this CI, then turn on multi-rank test in CPU workflow.

delock:
## Experiment results: enabling multi-rank CPU tests via `LOCAL_SIZE=4`

Baseline comparison: same-commit master run (all multi-rank tests skipped) vs this branch. **Zero regressions**: every failure is a test that was previously skipped.

### What now runs
~1800 previously-skipped multi-rank tests execute on the CPU runner (gloo). Final complete numbers from the last run (split invocations):
- `unit/v1` half (complete): **916 passed / 131 failed / 206 skipped**
- sequential (complete): 54 passed / 9 failed
- non-v1 half: killed by the 150m guard (still too slow on 4 vCPU), 59 failures listed inline

### Failure inventory & root causes (triaged)
| Count | Test | Root cause |
|---|---|---|
| 92 | `v1/zero/test_offload_states.py` | Asserts `memory_allocated()` drops after offload — a VRAM concept. `CPU_Accelerator.memory_allocated()` returns RSS, which never shrinks on free. Test/accelerator semantic gap; needs discussion (gate on device semantics or a better CPU metric). |
| 24 | `v1/zero/test_zero_autocast.py` | Hardcoded `dist_backend='nccl'` + a CUDA/NCCL-oriented bf16 gate. Needs accelerator-derived backend + capability-based skip. |
| 25 | ulysses SP tests | Numerical mismatches on CPU (real correctness questions, need deep dive). |
| 16 | onebit optimizer tests | `NoneType.size` at `onebit/adam.py:108` — runtime bug on CPU path. |
| ~20 | pipe / zeropp / moe-checkpoint / coalesce | Smaller follow-ups, same GPU-assumption patterns. |

### Test-side fixes already in this branch (validated by CI)
- All 5 DDP reference sites route through `wrap_ddp_reference()` (CPU modules must not pin `device_ids`) — dropped this class from 114 failures to 7.
- `reduce_boolean_flags` / autotp tests use `current_device_name()` instead of `current_device()` (a LOCAL_RANK string on CPU, not a device).
- fp16-config tests get a capability skipif (`not get_accelerator().is_fp16_supported()`) — GH `ubuntu-24.04` runners are hardware-heterogeneous w.r.t. AVX512-FP16, so hardcoded fp16 was a runner lottery.
- `fork_rng` probes the device module for a per-device RNG instead of matching accelerator names.

### CI infrastructure findings (need maintainer decisions)
1. **`--maxfail=100` in `PYTEST_OPTS` + no timeout in `DistributedExec._close_pool` = permanent wedge**: when the 100th failure trips the interrupt, teardown (`pool.starmap(_dist_destroy)`, `close/join`) can block forever on wedged gloo peers; the run then dies at the 6h job limit without printing any summary. This is what cancelled the first three attempts. (This branch works around it by splitting the suite and raising maxfail.)
2. **Suite size vs 4 vCPU**: the full multi-rank suite does not fit one 6h job. Options: split invocations (done here), lower `-n`, or a dedicated larger runner.
3. `CPU_Accelerator.device_count()`'s NUMA-node semantics remain the root cause of the original gap; the long-term fix could be a test-harness-level exemption for CPU (ranks are processes, not devices).

Happy to split the commits into separate PRs (workflow change / mechanical test fixes / triage follow-ups) per maintainer preference — the failure inventory above is intended as the working list.

delock:
Given the CPU UT would run ~hrs after enabling LOCAL_SIZE=4, I wouldn't suggest to turn on this in master.  However it is beneficial to use this branch to reveal problems in CPU accelerator and fix them.  I'm mark this PR as do not merge.

delock:
## Status update

### Fixes extracted so far (spin-off PRs)

Investigating the failures exposed by this CI change surfaced three real bugs, each with a standalone fix:

1. **`split_half_float_double()` silently drops all gradient buckets on CPU** — legacy type strings built from the accelerator device name never match CPU tensors (`torch.FloatTensor` has no device prefix), so with `contiguous_gradients=False` + `reduce_scatter=False` ZeRO-1/2 skips the gradient all-reduce entirely; `mean` configs mask it, `sum` is off by world_size. Fix in **#8382** (also fixes the `TestCoalesceCollectiveCount` "baseline issued no collectives" failure).
2. **Test-harness comparison helper crashes on CPU** — `reduce_boolean_flags()` passed `current_device()` (a bare rank string) to torch and used a 0-dim all-gather that gloo rejects. The `current_device_name()` part already landed on master via #8397; the rest is in #8382.
3. **ZeRO-3 coalesced parameter gather only waits for the last async handle** — gloo handles are independent and CPU `synchronize()` is a no-op, so earlier gathers can still be in flight when `param.data` is read, exposing uninitialized memory as NaN. Reproduced flaky (9/20 → 0/20 with the fix); a separate PR for `partition_parameters.py` is being prepared.

### Current failures on this branch (84 total: 75 in the `unit/v1` half + 9 sequential)

- ~50: CPU gating missing for GPU-only tests (bf16-requires-NCCL checks, fp16-unsupported, remaining DDP `device_ids` usages in `test_zero_autocast.py`)
- 15: HF download `PermissionError` for `bert-base-uncased` — runner environment issue, not code
- ~5: covered by #8382 + the ZeRO-3 wait fix above
- 16: `test_offload_states` remainder after #8409's gating; needs triage

### Current focus: phase-1 timeout / worker wedge

The job takes **4h05m** and the first half (`unit/` non-v1, `timeout 150m`) never finishes — it reaches 94% and a worker wedges for ~40min until the timeout kills it, losing that half's results. From the job log:

- **gw1** claims `TestDistInferenceAllReduce::test[128-dtype2]` (**fp16**) at 09:54:17 and produces no output for the remaining 111 minutes;
- gw0/gw2 wedge later, but the tests they were running were later **passed by other workers** — secondary casualties of state left by the first wedge;
- gw3 runs alone until the timeout.

Prime suspect is `dist.inference_all_reduce()` on an fp16 tensor over CPU/gloo. Investigating next; fixing this should also bring the job back to ~2h45m and recover phase-1 results.

delock:
## Progress update: full-suite run on `2654cac05` (4.5h, all three phases completed)

Comparison against the previous complete run (`f2793e70e`):

| Test | Prev | Now | Cause |
|---|---|---|---|
| `v1/zero/test_offload_states` | 92 | 16 | #8409 (via the master merge) cleared the 76 RSS-delta failures. The remaining 16 are all `TestDynamicOffloadStatesZero3` hitting `State OffloadStateTypeEnum.contiguous_grad_buffer must be removed after offload_states()` — a real state-lifecycle assertion, not a memory-metrics artifact |
| `sequence_parallelism/test_ulysses` | 17 | **0** | the shm-allreduce segment isolation commit (89ca1635b) fixed all of them |
| `v1/zero/test_zero_autocast` | 24 | 24 | two independent GPU assumptions: bf16 cases raise (not skip) in `bf16_required_version_check()` (NCCL/CUDA gate); fp16 cases die on the baseline `DDP(device_ids=[i])`; plus a hardcoded `dist_backend='nccl'` |
| `runtime/pipe` + `pipe/test_pipe_module` | 13 | 13 | gradient numerics (`Gradients differ ... use_reentrant=False`, iteration 2) — real correctness questions |
| others | ~40 | ~40 | no_sync 8, zero_user_backward 7, coalesce 7, zeropp 9 (sequential), checkpoint ~12 + 48 fixture ERRORs, ulysses_alst 5, offloadpp 2, intX 1 |

Totals: **v1 half 57 failed / 1122 passed / 292 skipped** (was 131/923/258); sequential 9 failed; non-v1 half 41 inline failures + 48 fixture ERRORs.

CI infra notes:
- `huggingface-cli` is now a hard-exit deprecation stub; switched to `hf download`.
- `bert-base-uncased` is Xet-backed and the anonymous `xet-read-token` request 404s; `HF_HUB_DISABLE_XET=1` falls back to the classic resolve endpoint.
- With those two fixes the job runs the full suite to completion again.

Next targets: autocast (24) — backend derivation + NCCL-scoped bf16 gate + the baseline DDP device_ids pinning; then the ZeRO-3 `contiguous_grad_buffer` lifecycle question.

PR #8382, opened 2026-09-01
Fix silent gradient-bucket drop on CPU in ZeRO-1/2 non-contiguous reduction

delock:
## Problem

While investigating multi-rank CPU CI (deepspeedai/DeepSpeed#8381), `TestGradientAllreduceOp` failed for two variants (`sum` + `prescale_gradients=True` + ZeRO-1/2 + `contiguous_gradients=False` + `reduce_scatter=False`) with results off by a factor of world_size, and `TestUnmanagedGradientAccumulationOffload` showed a numeric mismatch. Separately, any test comparing tensors across ranks on CPU crashed with `RuntimeError: Invalid device string` or a gloo allgather error.

## Root cause

`split_half_float_double()` built legacy type strings from the accelerator device name (`torch.{device}.FloatTensor`). CPU tensors report `torch.FloatTensor` with **no device prefix**, so no tensor ever matched: the bucket list came back empty and **the gradient all-reduce was silently skipped** — every rank updated on its local gradients.

- With `gradient_allreduce_op=mean` this stays invisible: identical per-rank gradients make the unreduced local value equal the mean.
- With `sum` the result is wrong by a factor of world_size.

The same variants pass on GPU CI (all 18 `TestGradientAllreduceOp` variants and 5 `TestGradientAllreduceOpTraining` variants are green on `modal-torch-latest`), because `torch.cuda.FloatTensor` matches on CUDA — the bug is CPU-specific, and CPU multi-rank runs were the first to reach this code.

The harness crashes were two stacked bugs in `reduce_boolean_flags()` (tests/unit/common.py): `current_device()` returns a bare rank index (a `LOCAL_RANK` string on CPU) that torch cannot parse as a device, and gloo rejects 0-dim inputs to `all_gather_into_tensor`.

## Fix

- Group gradient buckets by `t.dtype` instead of legacy type strings, making bucketing device-independent (behavior-equivalent on CUDA, where the legacy strings matched).
- Use `current_device_name()` (valid on every accelerator) and carry the flag as a 1-element tensor.

## Verification (CPU/gloo, world sizes 2–3)

| Test | Before | After |
|---|---|---|
| `TestGradientAllreduceOp` (18 variants) | 2 failed | 18/18 pass |
| `TestGradientAllreduceOpTraining` (5) | crash in comparison helper | pass (3 locally + muon skips fp16) |
| `TestUnmanagedGradientAccumulationOffload` + `InactiveParams` (6) | crash + numeric mismatch | 6/6 pass |
| `TestZero3ParamPartitioningBase` / `TestZeroToFP32` (regression smoke) | — | pass |

GPU behavior is unchanged by construction (same buckets, same order); GPU CI will re-confirm.

## Follow-up

With the comparison helper fixed, `test_unmanaged_varying_backward_count[3]` (ZeRO-3) can now reach its numeric comparison and reveals a pre-existing unmanaged-vs-managed mismatch on CPU that was previously masked by the harness crash. Out of scope here.


deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -64,14 +64,14 @@ def input(msg):
...
-    ]
+    # Legacy type strings omit the device prefix on CPU ("torch.FloatTensor"), so
+    # building them from the accelerator device name matches nothing on CPU and
+    # silently drops every gradient bucket, skipping the all-reduce entirely.
+    # Compare dtypes directly so the buckets are device-independent.
+    dtypes = [torch.half, torch.float, torch.double, torch.bfloat16]
     buckets = []
     for i, dtype in enumerate(dtypes):
-        bucket = [t for t in tensors if t.type() == dtype]
+        bucket = [t for t in tensors if t.dtype == dtype]
```
ebarkhordar: `t.type()` encodes layout as well as dtype, so this also changes membership for sparse gradients, not just on CPU. `sparse.type()` is `torch.sparse.FloatTensor`, which matches none of the four legacy strings on any device, so sparse grads were dropped from every bucket. `t.dtype == torch.float` does match them, and `allreduce_bucket` opens with `self.flatten(bucket)`, which is `_flatten_dense_tensors`.

```
>>> sparse = torch.sparse_coo_tensor(torch.tensor([[0, 2]]), torch.tensor([1., 2.]), (4,))
>>> sparse.type(), sparse.dtype
('torch.sparse.FloatTensor', torch.float32)
>>> _flatten_dense_tensors([torch.zeros(4), sparse])
RuntimeError: unsupported memory format option Contiguous
```

That is torch 2.14.0+cpu in a clean container, against this branch at ad3c27e.

I have not checked whether a sparse grad can actually reach here. ZeRO 1/2 has no sparse-tensor handling of its own, and the `is_sparse` routing at `engine.py:3824` is on the non-ZeRO path, so it may be unreachable. If it is reachable, an `nn.Embedding(sparse=True)` model on ZeRO-1/2 with `contiguous_gradients=False` moves from a silently skipped all-reduce to a hard failure.

Adding `and not t.is_sparse` on line 74 would preserve the old membership exactly, if you would rather keep this scoped to the device prefix.

delock: @ebarkhordar thanks for pointing this out.  I have amend the check to exclude sparse type.

ebarkhordar: `is_sparse` is COO-only, so the CSR family still gets in. A `sparse_csr` grad has `is_sparse == False`, lands in the float bucket, and `flatten` then raises `Sparse CSR tensors do not have is_contiguous`. Same for csc, bsr and bsc.

Not a regression, to be clear. On those four layouts `t.type()` itself raises `RuntimeError: Unimplemented backend SparseCsrCPU`, so the old code died a few lines earlier instead. Both fail, and I still have not found a path that gets any sparse grad here in the first place.

`t.layout == torch.strided` on line 75 covers all five layouts in one predicate, if you want the exclusion the comment on 70-71 describes.

torch 2.14.0+cpu in a clean container, at 03adfbb.

delock: @ebarkhordar thanks for the catch, fixed according to your suggestion.

pengdurice (approved):

Issue #8252, opened 2026-08-14
Define affine IR for checkpointing


delock:
This sub-issue (feature request) tracks the task of define IR spec for non-semantic checkpointing.   The idea is describe shard operations as affine operations, this will help use achieve goal of both phase 1 (UC_info for UCP conversion) and phase 2 (direct checkpointing transfer between topologies)

The outcome is defined as a spec document where we can implement UC_info description (both in memory structure and in file format), checkpoint conversion in phase 1, and checkpoint transfer component in phase 2.

Achyuthan-S:
@delock Consolidating my part of the #8230 discussion here as requested, updated for your last two replies.

**The boundary.** The IR describes; the planner decides. Concretely the IR should be a relation, not a function — for each element of F it names the set of shard elements that legitimately hold it, and the planner resolves that relation into a function by choosing sources under a cost model. Anything the IR resolves, the planner can no longer optimize. Cost model, scheduling and replica choice stay outside.

**On all-reduce / average — agreed, and I think dropping it is structurally necessary rather than just simpler.** I had argued for an optional combine op on a piece. Your read is better, and the repo supports it: `parameter_to_average_patterns` has exactly three references in DeepSpeed — the constant, the import, and the `.get` in `merge_tp_slices` — with no producer in tree, and `collect_autotp_universal_checkpoint_info` never emits it.

There is also a structural reason to drop it. A reduce is not invertible: N source elements collapse into one element of F and M_s^-1 cannot recover them. If the IR admits arithmetic, then invertibility — which you named as an intentional requirement — holds only over whichever subset of the language happens to be view-only. Removing the combine op is what makes M_s^-1 well-defined for the entire language. So: pieces are pure views, no arithmetic. A legacy checkpoint carrying the pattern lowers to replicated-with-assert-equal in the task-4 translator, or is refused explicitly.

**On subset + holding ranks.** I think this collapses your extension and the geometry/placement split into a single concept: a piece is an affine view of F plus a location set — the ranks, and later tiers, that hold that view.

- TP sharded: one location per piece
- TP replicated: one piece, N locations — exactly the freedom case 2 needs preserved
- ZeRO-1/3: the DP partition subdivides the byte range; the location set is the ranks holding it
- AutoEP: expert pieces carry location sets over the EP group
- ZeRO offload: the same field refines to a memory tier rather than needing separate placement machinery, which matches section 4.1

[...]

Issue #8230, opened 2026-08-07
[RFC] Universal checkpoint: non-semantic per-parameter geometric shard map


delock:
This is a design proposal (RFC) that came out of reviewing #8185 (uneven sharding) and #8168 (ZeRO-3 checkpoint consolidation).

The universal-checkpoint conversion/restore schema is currently **semantic** (regex-classified categories or semantic driven during restore), which limits it to a *closed* set of layouts — anything outside that set gets marked `unsupported` and refused, even though the model can shard and train those layouts fine at runtime. I'd like to propose moving to a **non-semantic, per-parameter geometric shard map**, so any layout a model can shard becomes convertible, and the converter stops being model-specific. Full proposal below; feedback very welcome before any implementation.

---

- **Status:** Draft
- **Author:** delock
- **Related:** PR #8185 (uneven sharding), PR #8168 (ZeRO-3 checkpoint consolidation)

## 1. Summary

Replace the current **semantic, regex-category-based** conversion/restore schema in the
universal checkpoint (UC) pipeline with a **non-semantic, per-parameter geometric shard
map**. Each parameter records the geometric correspondence between its tensor-parallel
shards and its full logical tensor. The converter and restorer *execute* this map; they no
longer *classify* parameters by architectural role. This makes conversion universal (no
`unsupported` set), model-agnostic, and extensible without converter code changes.

## 2. Motivation

The universal checkpoint is DeepSpeed's topology-agnostic interchange format: it stores
each parameter as a full logical tensor so a model trained with one TP/PP/DP grid can be
restored with another.

To produce and consume UC, the converter (`deepspeed/checkpoint/ds_to_universal.py`) and
the restorer (`deepspeed/checkpoint/universal_checkpoint.py`) must know how each parameter
was sharded across TP ranks. Today this knowledge is encoded **semantically**, in
`universal_checkpoint_info` (`uc_info`), as a set of regex-classified categories:

- `TP_REPLICATED_PARAMETER_PATTERNS`
- `PARAMETER_WITH_ROW_PARALLELISM_PATTERNS`
- `VOCABULARY_PARAMETER_PATTERNS`
- `PARAMETER_WITH_SUB_PARAMS` (+ `SUB_PARAM_SHARD_WIDTHS`)
- `AUTOTP_UNSUPPORTED_PARAMETER_PATTERNS`  （Adding from PR 8185）

The merge/slice logic branches on which category a parameter name matches (see
[...]

PR #8385, opened 2026-09-01
Describe universal checkpoint shards as affine maps

Achyuthan-S:
Implements steps 1–3 of the staging plan in #8252: the IR structure, the lowering from today's metadata, and the converter reading it. Emitting the map from `collect_autotp_universal_checkpoint_info` is step 4 and will be a separate PR, per @delock's suggestion.

**No conversion changes.** Nothing writes an affine map yet, so the new branch in `merge_tp_slices` is never taken and every checkpoint converts exactly as it does today. The tests are what give the work its value at this stage: they require the map to reproduce the existing arithmetic before anything depends on it.

### Why

Universal checkpoint decides how to merge a parameter by matching its *name* against regex categories — vocabulary, row-parallel, fused sub-parameters. A layout no category describes cannot be converted at all, which is what `AUTOTP_UNSUPPORTED_PARAMETER_PATTERNS` records. This describes the layout geometrically instead, so the question becomes where the bytes are rather than what the parameter means.

`deepspeed/checkpoint/affine_ir_spec.md` is the specification, developed in #8252 and #8230. `affine.py` references it.

### What is here

**`AffinePiece`** — a block of elements, recording where it sits in the full tensor **and** where it sits in the shard, with `shape` shared. Each side is `torch.as_strided`'s argument list, so a piece is executable with no interpretation step.

A piece also carries:

- `locations`, the ranks holding it, so a converter can read a replicated block from whichever rank is cheapest rather than a designated owner
- `scale`, the factor the shard holds the block by. A row-parallel layer pre-divides its replicated bias by the world size so the all-reduced sum adds the bias once; the divisor changes with the world size, so a checkpoint that cannot record it cannot be restored at a different TP degree without a rule naming which parameters are biases.

Scaling is admitted where averaging is not, and the line is invertibility rather than arithmetic: a scale is `1 -> 1` and reverses, a reduce is `N -> 1` and does not.

**`ParamAffineMap`** — `extract` and `rebuild`, which are the same loop with the copy reversed, plus coverage and homogeneity validation and the on-disk form.

**Lowering constructors** for the layouts the converter already handles: `replicated_map`, `contiguous_split_map`, `sub_param_map`. Row and column parallelism differ only in stride, so one constructor covers both — which is why the recorded concat dimension becomes redundant.

### Tests

39 cases, all plain pytest: the partition functions take an explicit rank, so none of this needs a process group or an accelerator, and the module runs in about two seconds on any runner.

**Parity** — each constructor must reproduce `merge_tp_slices`' own arithmetic exactly, including an uneven `[3, 3, 2, 2]` split (where `torch.chunk` and AutoTP disagree) and sub-parameters of different sizes, none dividing evenly by the tp degree.

**Coverage** — the four layouts `AUTOTP_UNSUPPORTED_PARAMETER_PATTERNS` currently refuses (`bigcodetype`, `codegentype`, Yuan shared-QK value and o_proj) are all describable. Pieces are derived by running the **real** partition functions on a marker tensor, then validated against independent random data. That second step is the actual test: if pieces derived from markers reproduce a random tensor's shard bit-exactly, the layout is a pure view rather than something data-dependent.

### Two things worth knowing about

**A shard is not its pieces concatenated in order.** For a column split the shard interleaves them row by row, so an implementation built on the concatenation assumption reproduces row-split layouts correctly and silently transposes column-split ones. A round-trip test on the Yuan o_proj case caught this, and it is why both ends of a piece are recorded.

**Piece offsets are storage offsets**, because that is what `as_strided` takes. A loaded shard is frequently a *view* into a larger buffer — CodeGen's rank 1 begins at offset 96 of a 192-element buffer — so applying a piece to it directly reads from the wrong place, silently. `_flat_buffer` normalises this.

### One boundary for review

The category branches write per-category keys into the converted checkpoint (`CAT_DIM`, `PARAM_N_SUB_PARAMS`, `SUB_PARAM_SHAPE`) that the restore path reads. The affine branch cannot reconstruct those, and arguably should not: the geometry is what a restoring job needs and it is not tied to a category. So a checkpoint converted through the map carries the map instead. This is inert until step 4, but it is the compatibility question I would most like a second opinion on.

Related: #8252, #8230. Builds on #8185 (aa3914df8).

cc @delock

### DIFF
--- a/deepspeed/checkpoint/affine.py
+++ b/deepspeed/checkpoint/affine.py
@@ -1,4 +1,4 @@
-# Copyright (c) Microsoft Corporation.
+# Copyright (c) DeepSpeed Team.
 # SPDX-License-Identifier: Apache-2.0
 
 # DeepSpeed Team
--- a/deepspeed/checkpoint/affine.py
+++ b/deepspeed/checkpoint/affine.py
@@ -1,0 +1,205 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+"""Affine description of how a parameter is sharded.
+
+Universal checkpoint currently decides how to merge a parameter by matching its *name*
+against regex categories (vocabulary, row-parallel, fused sub-parameters, ...), so a
+layout that no category describes cannot be converted at all. This module describes a
+shard geometrically instead: each rank holds a list of affine views of the full tensor,
+and the views alone say how to rebuild it.
+
+The representation is deliberately view-only. A piece never combines several elements of
+the full tensor, which is what keeps the mapping invertible: given the full tensor a rank's
+shard can be produced, and given the shards the full tensor can be rebuilt.
+"""
+
+import torch
+
+__all__ = ['AffinePiece', 'ParamAffineMap']
+
+
+class AffinePiece:
+    """A block of elements, described where it lives in the full tensor and in the shard.
+
+    A piece holds the same elements in the same arrangement on both sides, so ``shape`` is
+    shared and only the offset and strides differ. Describing the shard side explicitly is
+    what lets a piece land somewhere other than the end of the shard: a shard split along
+    a column interleaves its pieces row by row, so "append each piece in turn" is not
+    enough to say where the elements go.
+
+    The offsets and strides are the arguments of ``torch.as_strided``, so a piece can be
+    applied to a tensor without any interpretation step.
+
+    ``locations`` is a set of ranks rather than a single owner because a piece may be
+    replicated. Naming one owner here would be a scheduling decision, and the cheapest
+    source depends on the topology, which this description does not know about.
+    """
+
+    __slots__ = ('shape', 'source_offset', 'source_strides', 'dest_offset', 'dest_strides', 'locations')
+
+    def __init__(self, shape, source_offset, source_strides, dest_offset, dest_strides, locations):
+        self.shape = tuple(int(dim) for dim in shape)
+        self.source_offset = int(source_offset)
+        self.source_strides = tuple(int(stride) for stride in source_strides)
+        self.dest_offset = int(dest_offset)
+        self.dest_strides = tuple(int(stride) for stride in dest_strides)
+        self.locations = frozenset(int(rank) for rank in locations)
+
+    @property
+    def numel(self):
+        count = 1
+        for dim in self.shape:
+            count *= dim
+        return count
+
+    def source_view(self, full_param):
+        """This piece's elements as a view of the full parameter, without copying."""
+        return torch.as_strided(full_param,
+                                size=self.shape,
+                                stride=self.source_strides,
+                                storage_offset=self.source_offset)
+
+    def dest_view(self, shard):
+        """This piece's elements as a view of the shard that holds it."""
+        return torch.as_strided(shard, size=self.shape, stride=self.dest_strides, storage_offset=self.dest_offset)
+
+    def source_offsets(self):
+        """Yield the offset into the full tensor of every element this piece covers.
+
+        Used to check coverage. Pieces are small relative to the parameter, so walking
+        them elementwise is affordable and avoids assuming anything about their shape.
+        """
+        return self._offsets(self.source_offset, self.source_strides)
+
+    def _offsets(self, base, strides):
+        if not self.shape:
+            yield base
+            return
+
+        index = [0] * len(self.shape)
+        while True:
+            offset = base
+            for axis, position in enumerate(index):
+                offset += position * strides[axis]
+            yield offset
+
+            axis = len(self.shape) - 1
+            while axis >= 0:
+                index[axis] += 1
+                if index[axis] < self.shape[axis]:
+                    break
+                index[axis] = 0
+                axis -= 1
+            if axis < 0:
+                return
+
+    def __repr__(self):
+        return (f'AffinePiece(shape={self.shape}, source=({self.source_offset}, {self.source_strides}), '
+                f'dest=({self.dest_offset}, {self.dest_strides}), locations={sorted(self.locations)})')
+
+    def __eq__(self, other):
+        if not isinstance(other, AffinePiece):
+            return NotImplemented
+        return (self.shape == other.shape and self.source_offset == other.source_offset
+                and self.source_strides == other.source_strides and self.dest_offset == other.dest_offset
+                and self.dest_strides == other.dest_strides and self.locations == other.locations)
+
+    def __hash__(self):
+        return hash(
+            (self.shape, self.source_offset, self.source_strides, self.dest_offset, self.dest_strides, self.locations))
+
+
+class ParamAffineMap:
+    """How one parameter is spread over a tensor-parallel group.
+
+    ``shard_shapes`` gives each rank the shape of the tensor it holds, and
+    ``pieces_by_rank`` says which blocks of the full parameter make it up.
+    """
+
+    def __init__(self, logical_shape, shard_shapes, pieces_by_rank):
+        self.logical_shape = tuple(int(dim) for dim in logical_shape)
+        self.shard_shapes = {int(rank): tuple(int(dim) for dim in shape) for rank, shape in shard_shapes.items()}
+        self.pieces_by_rank = {int(rank): list(pieces) for rank, pieces in pieces_by_rank.items()}
+
+    @property
+    def numel(self):
+        return _product(self.logical_shape)
+
+    def uncovered_offsets(self):
+        """Return the offsets of the full tensor that no rank holds.
+
+        A non-empty result means the parameter cannot be rebuilt from its shards, so this
+        is the check that has to pass before a map is used for conversion.
+        """
+        covered = set()
+        for pieces in self.pieces_by_rank.values():
+            for piece in pieces:
+                covered.update(piece.source_offsets())
+        return sorted(set(range(self.numel)) - covered)
+
+    def validate(self):
+        missing = self.uncovered_offsets()
+        assert not missing, (f'Affine map for a parameter of shape {self.logical_shape} leaves '
+                             f'{len(missing)} element(s) uncovered, starting at offset {missing[0]}, '
+                             'so the parameter cannot be rebuilt from its shards.')
+
+        for rank, pieces in self.pieces_by_rank.items():
+            held = sum(piece.numel for piece in pieces)
+            expected = _product(self.shard_shapes[rank])
+            assert held == expected, (f'Rank {rank} holds a shard of {expected} elements but its pieces '
+                                      f'account for {held}.')
+
+    def rebuild(self, shards):
+        """Rebuild the full parameter from per-rank shards, using the pieces as the plan.
+
+        ``shards`` maps a rank to its shard. Where pieces overlap the data is identical by
+        construction, so writing them in any order gives the same result.
+        """
+        self.validate()
+        any_shard = next(iter(shards.values()))
+        full_param = torch.empty(self.numel, dtype=any_shard.dtype, device=any_shard.device)
+
+        for rank, pieces in self.pieces_by_rank.items():
+            flat_shard = _flat_buffer(shards[rank])
+            for piece in pieces:
+                piece.source_view(full_param).copy_(piece.dest_view(flat_shard))
+
+        return full_param.view(self.logical_shape)
+
+    def extract(self, full_param, rank):
+        """Produce one rank's shard from the full parameter. The inverse of ``rebuild``."""
+        flat_param = _flat_buffer(full_param)
+        shard_shape = self.shard_shapes[rank]
+        shard = torch.empty(_product(shard_shape), dtype=full_param.dtype, device=full_param.device)
+
+        for piece in self.pieces_by_rank[rank]:
+            piece.dest_view(shard).copy_(piece.source_view(flat_param))
+
+        return shard.view(shard_shape)
+
+    def __repr__(self):
+        counts = {rank: len(pieces) for rank, pieces in sorted(self.pieces_by_rank.items())}
+        return f'ParamAffineMap(logical_shape={self.logical_shape}, pieces_per_rank={counts})'
+
+
+def _flat_buffer(tensor):
+    """Flatten ``tensor`` into a buffer whose storage starts at its first element.
+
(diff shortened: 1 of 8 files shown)


Achyuthan-S:
The modal-torch-latest failure is a dependency-install error, not a test failure — THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE while downloading torch/CUDA wheels, before any test runs. This branch passed the same workflow two hours ago (run 33527949501) and the diff touches no requirements or CI files. Looks transient — could someone re-run it? @delock .

Achyuthan-S:
@delock ., The modal-torch-latest failures aren't from this change. The first was a pip hash mismatch while installing torch/CUDA wheels (THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE), before any test ran; the second was the job being cancelled at the 75-minute mark. Neither reported a test failure, and the diff touches no requirements or CI files. This branch passed the same workflow green in 62m on the first push (run 33527949501). 

Happy to rebase or re-run if that helps.

 Also correcting the description: it says 31 test cases, now 39 after the review fixes.

deepspeed/checkpoint/affine.py
```diff
@@ -0,0 +1,455 @@
+# Copyright (c) Microsoft Corporation.
```
delock: Remove Microsoft license header.

deepspeed/checkpoint/affine_ir_spec.md
```diff
@@ -0,0 +1,544 @@
...
+
+## 2. The representation
+
+### 2.1 Piece
+
+A **piece** is a block of elements, described by where it sits in the full tensor `F` and
+where it sits in the shard:
+
+```python
+Piece = (shape, source_offset, source_strides, dest_offset, dest_strides, locations)
```
delock: scale is list down below bit does not appear here.

delock: `scale` should appear in this line.

delock:
Hi @Achyuthan-S thanks for your PR.  I have read the spec part and left my comments.  Nice catch to optimizer states scaling where we didn't covered early.  I like the json structure and how that simplify the code in 7.1

Modal failure is due to a different reason and we are working on it and hopefully it won't block PRs again.  I'll come back to review the rest part next week.   Also need to ponder on the optimizer state scaling part a little more.  The rest part of the spec document itself looks fine to me.  Its better if you have an example on optimizer state scaling to help understanding.  Thanks!

deepspeed/checkpoint/affine.py
Achyuthan-S: Done. Switched both new files to # Copyright (c) DeepSpeed Team. to match what files added recently carry. Confirmed with scripts/check-license.py — the Microsoft line is commented out in its COPYRIGHT list, so only the SPDX and DeepSpeed Team lines are actually required.

deepspeed/checkpoint/affine_ir_spec.md
Achyuthan-S: Fixed — it was in the notes below but not in the tuple itself.

Achyuthan-S:
> Hi @Achyuthan-S thanks for your PR. I have read the spec part and left my comments. Nice catch to optimizer states scaling where we didn't covered early. I like the json structure and how that simplify the code in 7.1
> 
> Modal failure is due to a different reason and we are working on it and hopefully it won't block PRs again. I'll come back to review the rest part next week. Also need to ponder on the optimizer state scaling part a little more. The rest part of the spec document itself looks fine to me. Its better if you have an example on optimizer state scaling to help understanding. Thanks!

Thanks @delock. All three addressed in the latest push, and good to know the modal failure is being handled separately.

   Optimizer state example — added to §2.1 as a worked table, for a row-parallel bias at world size 4 (scale = 1/4). A shard holds full * s**power and conversion recovers full = shard / s**power:

    | state      | power | s**power | a rank holds | F recovers to          |
    |------------|-------|----------|--------------|------------------------|
    | fp32       | 1     | 1/4      | 2.0          | 8.0 — the logical bias |
    | exp_avg    | -1    | 4        | 8.0          | 2.0                    |
    | exp_avg_sq | -2    | 16       | 32.0         | 2.0                    |

    The moments move the opposite way to the parameter, and the second twice as far, because the optimizer trains p = b/4 and ∂L/∂b = (∂L/∂p)·(1/4). Using the parameter's own factor for all three would multiply exp_avg by 4 where it should be divided — off by 16× — and change the trajectory after a resume without any visible error. The numbers in that table are the actual output of ParamAffineMap.rebuild, not hand-derived.

 One thing worth flagging while you ponder it: those powers are derived, not measured. They follow from the chain rule, and the round-trip is tested, but nothing yet trains a model with a scaled bias, converts, resumes, and checks the trajectory matches. §5 and §8.2 are verified against the real partition functions; this part is reasoned. If you'd rather the IR simply refused a scaled optimizer state until that end-to-end check exists, that's a smaller and more defensible surface — happy to go that way.

Also correcting the description: it said 31 test cases, now 39 after the review fixes.

0z5a:
@Achyuthan-S @delock Following the train→convert→resume gap discussed here and in #8230/#8252, I ran actual checkpoint tests on top of `35eb2e2`.

[Follow-up commit](https://github.com/0z5a/DeepSpeed/commit/e753a02b72dd18bc220f31ecac5ce5f6c1a89c54) adds a small column/row AutoTP model with real FP32 ZeRO-1 Adam training: TP2 saves after four steps, converts through either the legacy or affine path, and resumes for four more steps at TP1/TP2. It compares logits, losses, gradients, FP32 parameters, both moments, and step counters with uninterrupted training. Affine metadata is supplied explicitly through checkpoint client state, since the producer is a later step. Native `LinearAllreduce` adds its bias after reduction, so that bias's scale in this test is **1**.

The test found two pre-existing loader problems in **both** conversion paths:

- TP2→TP1 invoked the Megatron weight merger and failed looking for `attention.dense.weight`.
- `map_to_flat_opt_states` rebound the FP32 parameter fragment to a moment buffer, instead of rebinding the corresponding optimizer fragment.

The commit fixes these. All four resume cases passed on **2× RTX 4000 Ada 20 GB, torch 2.13.0+cu130, Python 3.12.14 / NCCL**. CPU/Gloo also passed those four cases, 71 affine/AutoTP unit regressions, and four existing fragment/DP-resize regressions. Modified-file pre-commit checks passed.

```bash
pytest -sv tests/unit/checkpoint/test_autotp_uc_checkpoint.py::TestAffineUniversalCheckpointResume
```

There is a separate non-unit-scale result: [standalone real-checkpoint reproducer and recorded output](https://gist.github.com/0z5a/f0c5e39166d92a5d59fb0a8778530d3f). This is an explicitly parameterized `p = 0.5*b` bias on one CPU/Gloo rank, with clipping disabled and a plain torch Adam oracle; it is **not native AutoTP bias**. The trained parameter and moments convert/restore correctly with the current powers, but loading also restores source-coordinate `lr=0.015, eps=0.1`, whereas the target coordinates need `lr=0.03, eps=0.05`. The next update has max parameter error **0.0087981**, growing to **0.0272084** after four updates. Restoring those target-coordinate group hyperparameters gives zero parameter error at every step in this probe.

So this supports the moment powers for that explicit reparameterization, but not general scaled-optimizer resume with unchanged group hyperparameters. Before enabling that path, the contract needs to cover optimizer hyperparameters and replicated-gradient semantics, or reject unsupported scaled optimizer states as you suggested. The follow-up commit is available for cherry-picking; its regression scope is the native scale-one path.

Achyuthan-S:
@0z5a Thanks for running this — an end-to-end resume was the one thing the spec asserted without measuring, and it's good to have it closed.

I reproduced the hyperparameter result independently with a plain Adam oracle, and it matches yours exactly. Keeping the source `lr`/`eps`, max parameter error is 0.0068 after one update growing to 0.0160 after four; transforming to `lr / s` and `eps * s` gives **exactly zero at every step**. Your `lr 0.015 → 0.03, eps 0.1 → 0.05` is what the chain rule gives: the optimizer trains `p`, so matching `Δb = Δp / s` needs `lr_b = lr_p / s` and `eps_b = eps_p * s`.

The part worth stating clearly, since it's easy to read this as the powers being wrong: **they aren't.** Your probe confirms the parameter and both moments convert and restore correctly under the powers in §2.1. The divergence is entirely from the optimizer's own hyperparameters, which sit a layer below parameter geometry.

That makes this the case I raised with @delock last week — that the powers are derived rather than measured, and that refusing scaled optimizer states is the smaller and more defensible surface until the contract covers them. Your data settles which way to go, so I've taken it: `rebuild`/`extract` now raise `NotImplementedError` on a scaled optimizer state, naming the reason. A scaled *parameter* still converts normally, since that path is correct. §2.1 records the constraint and the `lr / s`, `eps * s` transform a future contract will need. The factors themselves stay pinned in tests — they're what that contract builds on.

On the two loader bugs: those are pre-existing in both conversion paths and independent of this change, and the `map_to_flat_opt_states` one looks like the more serious of the two — rebinding a parameter fragment to a moment buffer would corrupt a resume with nothing to show for it. I'd suggest sending them as their own PR rather than folding them in here: they're a bug fix in shared code, they'll review faster on their own, and they shouldn't wait on an IR discussion. Happy to review it.

If you'd rather they ride along with this one, I'll cherry-pick with attribution — either way works, and thanks again for the thorough write-up.

Achyuthan-S:
@delock could you kick off CI when you get a chance? Modal hasn't managed to complete a run on this branch yet, so the suite hasn't actually been exercised.

0z5a:
> @0z5a Thanks for running this — an end-to-end resume was the one thing the spec asserted without measuring, and it's good to have it closed.
> 
> 
> 
> I reproduced the hyperparameter result independently with a plain Adam oracle, and it matches yours exactly. Keeping the source `lr`/`eps`, max parameter error is 0.0068 after one update growing to 0.0160 after four; transforming to `lr / s` and `eps * s` gives **exactly zero at every step**. Your `lr 0.015 → 0.03, eps 0.1 → 0.05` is what the chain rule gives: the optimizer trains `p`, so matching `Δb = Δp / s` needs `lr_b = lr_p / s` and `eps_b = eps_p * s`.
> 
> 
> 
> The part worth stating clearly, since it's easy to read this as the powers being wrong: **they aren't.** Your probe confirms the parameter and both moments convert and restore correctly under the powers in §2.1. The divergence is entirely from the optimizer's own hyperparameters, which sit a layer below parameter geometry.
> 
> 
> 
> That makes this the case I raised with @delock last week — that the powers are derived rather than measured, and that refusing scaled optimizer states is the smaller and more defensible surface until the contract covers them. Your data settles which way to go, so I've taken it: `rebuild`/`extract` now raise `NotImplementedError` on a scaled optimizer state, naming the reason. A scaled *parameter* still converts normally, since that path is correct. §2.1 records the constraint and the `lr / s`, `eps * s` transform a future contract will need. The factors themselves stay pinned in tests — they're what that contract builds on.
> 
> 
> 
> On the two loader bugs: those are pre-existing in both conversion paths and independent of this change, and the `map_to_flat_opt_states` one looks like the more serious of the two — rebinding a parameter fragment to a moment buffer would corrupt a resume with nothing to show for it. I'd suggest sending them as their own PR rather than folding them in here: they're a bug fix in shared code, they'll review faster on their own, and they shouldn't wait on an IR discussion. Happy to review it.
> 
> 
> 
> If you'd rather they ride along with this one, I'll cherry-pick with attribution — either way works, and thanks again for the thorough write-up.

Cherry-picking with attribution works for me — thanks!

delock:
Hi @Achyuthan-S , thanks for provide the example.  @0z5a thanks for running the checkpoint test.  This discussion reveals `scale!=1` should be a case for rollout end checkpoint only.  During inference bias could be moved before allreduce and scale could help.  In the training end (where optimizer and lr/eps needs to be scaled), keep scale=1 should be better choice.  My rational is handling scale with optimizer would increase software complexity, handling scale in inference would catch most potential performance benefit.

In three map, why only `contiguous_split_map` accept scale as parameter, can you explain?

The intergration test is a good idea, thanks @0z5a for contribution and thanks @Achyuthan-S for cherrypicking.

delock:
> @delock could you kick off CI when you get a chance? Modal hasn't managed to complete a run on this branch yet, so the suite hasn't actually been exercised.

@Achyuthan-S I started one round of test.  For modal test, they will be tested when this PR enter merge queue.  We will see the result when we start the merge process.

delock (approved):

Achyuthan-S:
@delock Answers below, plus two changes pushed since your approval — flagging so you know a re-approval may be needed.

**On the three maps.** `scale` belongs on `replicated_map`, and I've moved it there. The layouts that pre-divide a value hold it whole on every rank: Yuan's o_proj and the last conv layer both replicate the bias divided by the world size, so the all-reduced sum adds it exactly once. The weight beside them is what gets split, and it's unscaled — so a split constructor has nothing to scale, and `contiguous_split_map` and `sub_param_map` now take no `scale` argument. There's a test covering the replicated case.

**On `scale != 1` being rollout-only** — agreed, and the code already lands there. Refusing scaled optimizer states means a training checkpoint, which carries `exp_avg`/`exp_avg_sq`, cannot have `scale != 1`; an inference or rollout export carries only the parameter, so it can. Happy to state that as an explicit rule in §2.1 if you'd like it written down rather than implied.

**Cherry-picked @0z5a's fixes** as agreed, with attribution. Worth noting this changes the PR's scope: `engine.py` and `tensor_fragment.py` are live paths, so it is no longer purely additive. I've updated the description to say so. Locally on CPU/gloo their four resume cases pass alongside the 40 affine tests — 43 passed.

**On the merge queue** — both attempts ended with `SandboxStartTimeout: Sandbox did not start within 1800s, so no test ran`, which the job reports as a capacity problem rather than a test failure. All 12 PR checks are green. Ready to re-queue whenever it's worth another attempt.

Achyuthan-S:
@delock The current queue entry predates the two commits I pushed just after you queued — its squash is the earlier 5-file version, without the scale move or @0z5a's fixes. Worth dequeuing and re-adding on dd5c57f so those land.

delock:
Hi @Achyuthan-S ，push new commit should deque a merge request, but for some reason this time it didn't happen.  I'll review #8477 and 0z5a's #8474, thank you!

PR #8388, opened 2026-09-01
Offload a saved view when it is the last value holding its storage

### LINKED ISSUE #8387: DeepCompile offload_activation skips saved views when the aliased base is not saved
## Symptom

DeepCompile activation offload (all-floats: `compile.offload_activation=true`, `offload_activation_hidden_states_only=false`) OOMs at long sequence instead of moving more activations. Failures report `planned_offloads=None` because the process dies at global step 0, before the WARMUP planner runs.

## Cause

`_eligible_activations` drops any saved node whose `target` is in `get_no_copy_ops()` (`view`, `permute`, `slice`, `expand`, `detach`, …) with no liveness check on the alias root:

https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/compile/passes/offload_activation.py

That skip is correct when the **base is also saved** (copying the view frees nothing). It is wrong when AOT saves only the view: the base is not in the saved list, so the view is the last keeper of that storage. Offloading it would free real bytes.

## Repro

```bash
pytest tests/unit/v1/compile/test_offload_activation.py::test_eligible_includes_saved_view_when_base_is_not_saved -v
```

This test is `xfail(strict=True)`: it asserts the view is eligible and `_skipped["alias"] == 0`. Today the view is dropped.

Contrast (passes; skip is correct when the base is also saved):

```bash
pytest tests/unit/v1/compile/test_offload_activation.py::test_eligible_skips_saved_view_when_base_is_also_saved -v
```

Graph used by both: `x -> relu(base) -> aten.view(viewed) -> sum(out)`, AOT-shaped outputs `(out, viewed)` vs `(out, base, viewed)`.

## Scale evidence (Qwen3.5-9B, 2xH200, ZeRO-3 GPU Adam, no HF GC)

- Alias skip was ~75% of saved-for-backward bytes; the floor still grew ~5 GB / 1k tokens/GPU.
- OOM at 4K mbs=5, 8K mbs=3, 16K mbs=2, 32K mbs=1 (`metrics/qwen35_gpuadam_mbs_scale*`).
- Stacks: `dc.offload_tensor`, FLA `empty_like(v)`, inductor workspace, with ~137 GiB already live. Death is step 0 with the floor already applied.

## Fix direction

Offload a saved alias when no other saved node keeps the same storage. Do not drop the existing skip for “base is also saved.”

pengdurice:
# Offload a saved view when it is the last value holding its storage

Fixes deepspeedai/DeepSpeed#8387

## Problem

`_eligible_activations` in `deepspeed/compile/passes/offload_activation.py` dropped every
saved-for-backward value whose target is an aliasing op (`view`, `permute`, `slice`, `expand`,
`detach`, and every other aten op whose schema declares an aliasing tensor return):

```python
# A value that only aliases another tensor shares its storage, so copying it out frees
# nothing while the tensor it aliases is still live.
if node.target in no_copy_ops:
    _skipped["alias"] += _skip_bytes(node)
    continue
```

There was no check on whether anything else still held that storage.

The rule is correct when the tensor the view came from is a weight, a graph input, or another
saved value. It is wrong when AOTAutograd saves only the view. The base node is then dead, but its
allocation is not: the returned view still points at it, so the caching allocator cannot reclaim
the block until the backward pass reads the view. Moving the view to the host is what releases the
whole allocation.

## The fix

Three helpers, and the single alias skip becomes two narrower ones.

| Helper | What it does |
|---|---|
| `_alias_root` | Follows a node back through aliasing ops to the node that allocated the storage. A node whose target is not an aliasing op is its own root. Results are cached per call. |
| `_storage_keeper_counts` | Counts, per storage root, the values that outlive the forward pass: graph placeholders (the caller owns those), values returned to the caller, and every saved value. |
| `_has_reloadable_layout` | True when the tensor is non-overlapping and dense. |

```python
if node.target in no_copy_ops:
    if storage_keepers[_alias_root(node, no_copy_ops, alias_roots)] > 1:
        _skipped["alias"] += _skip_bytes(node)
        continue
    if not _has_reloadable_layout(node):
        _skipped["alias_layout"] += _skip_bytes(node)
        continue
```

The existing skip for "the base is also saved" is kept, exactly as the issue asks. What changed is
that it is now conditional on a liveness count instead of unconditional.

## Tests

Added to `tests/unit/v1/compile/test_offload_activation.py`. The first two are the tests named in
the issue; neither existed in the tree or on `master`, so both are written as ordinary tests rather
than one of them as `xfail(strict=True)`.

| Test | Asserts |
|---|---|
| `test_eligible_includes_saved_view_when_base_is_not_saved` | the view is eligible and `_skipped["alias"] == 0` |
| `test_eligible_skips_saved_view_when_base_is_also_saved` | the view is skipped, only the base is eligible |
| `test_eligible_skips_a_saved_view_the_host_copy_would_not_reproduce` | an expanded view is skipped under `alias_layout` |

The graph both of the first two use is the one the issue specifies: `x -> relu(base) ->
aten.view(viewed) -> sum(out)`, with AOT-shaped outputs `(out, viewed)` and `(out, base, viewed)`.

The pre-existing `test_fwd_skips_values_that_alias_another_tensor` still passes unchanged. In that
graph the view's root is a placeholder, so it is still skipped.

### Results

| Run | Result |
|---|---|
| `tests/unit/v1/compile/test_offload_activation.py`, 1 GPU | 32 passed, 2 skipped (they need world_size=2) |
| `TestOffloadActivation::test_offload_activation_correctness`, 2 GPUs | 2 passed |

torch 2.6.0+cu124, Python 3.10. The end-to-end test compares losses with offloading on and off at
`DS_DC_OFFLOAD_ACT_MIN_SIZE_MB=0`, so it now exercises the newly eligible views. `yapf --diff` and
`flake8 --max-line-length=120` are clean on both changed files.

## Measured in a real training run

Full detail and provenance in `RUN_RESULTS.md`. Qwen3-14B, 8xH200 (139.80 GiB usable), ZeRO-3,
tiled loss 8, `expandable_segments:True`. Two frozen trees differing in exactly one file; every
cell asserts its own import path and prints the file's sha256 before running.

**Sequence 4096, micro-batch 4** — both trees complete, so this measures the size of the bug:

| | baseline | fixed |
|---|---|---|
| bytes skipped under `alias`, per graph | 49.66 GB | **13.42 GB** |
| floor moved | 26.8 GB, 160 values | **63.0 GB, 280 values** |
| peak | 134.97 GB | 131.45 GB |
| steady step | 11.49 s | **10.95 s** |

Total saved for backward is about 76.7 GB per rank, so the old rule was refusing 65% of it.
With the budget forced to nothing so only the floor runs, peak falls from 134.74 GB to **99.98 GB**.

**Sequence 4096, micro-batch 5, floor only** — the configuration the issue reports dying at global
step 0:

| | baseline | fixed |
|---|---|---|
| bytes skipped under `alias` | 62.08 GB | **16.78 GB** |
| floor moved | 33.8 GB, 362 values | **79.1 GB, 482 values** |
| **steps completed** | **0 of 12** | **12 of 12** |
| verdict | **out of memory**, 603 MiB free | **completed**, peak 119.20 GB |

Same node type, same configuration, same instruction to move everything possible. One file differs.

### DIFF
--- a/deepspeed/compile/passes/offload_activation.py
+++ b/deepspeed/compile/passes/offload_activation.py
@@ -3,6 +3,7 @@
 
 # DeepSpeed Team
 
+import functools
 import os
 import time
 from collections import OrderedDict, defaultdict
@@ -61,6 +62,12 @@ _activation_ops_lib = None
 # The backward graph receives those tensors as placeholders carrying the same names, which is how
 # the two halves of the pass find each other.
 _offload_plans: Dict[int, "OrderedDict[str, Tuple[int, int]]"] = {}
+
+# Bytes each planned value keeps allocated if it stays on the device, keyed the same way. This is
+# not the size of the copy: a saved view holds its whole base allocation alive, so a 4KB row of a
+# 4MB tensor costs 4MB of residency and 4KB of copy. The planner spends headroom in residency
+# bytes, while the backward pass schedules its copies in copy bytes.
+_resident_bytes: Dict[int, Dict[str, int]] = {}
 
 # Value ids identify a host buffer inside the C++ executor. They never repeat, so a buffer holding
 # a tensor of one shape is never reused for another.
@@ -279,6 +286,27 @@ def _skip_bytes(node) -> int:
         return 0
 
 
+@functools.lru_cache
+def _aliasing_ops():
+    """Ops whose output shares storage with an input, for the purpose of tracking that storage.
+
+    get_no_copy_ops() reads the aten schemas, and aten._unsafe_view declares a fresh tensor return
+    even though it hands back a view -- that declaration is the entire point of the op. It shares
+    storage all the same, and AOTAutograd emits it after nearly every matmul, so this pass has to
+    know about it. It is added here rather than in get_no_copy_ops() because that set also decides
+    where the ZeRO-3 passes release parameters, and this pass has no business changing that.
+    """
+    return frozenset(get_no_copy_ops() | {torch.ops.aten._unsafe_view.default})
+
+
+def _zero3_gathered_param_ops():
+    """The op that produces a ZeRO-3 gathered parameter buffer, empty if DeepCompile is not built."""
+    try:
+        return {torch.ops.dc.allgather_param.default}
+    except (AttributeError, RuntimeError):
+        return set()
+
+
 def _alias_root(node: Node, no_copy_ops, cache: Dict[Node, Node]) -> Node:
     """The node that allocated the storage `node` reads.
 
@@ -311,11 +339,20 @@ def _storage_keeper_counts(graph: Graph, saved_nodes: List[Node], returned_to_ca
     """How many values that outlive the forward pass hold each storage.
 
     Moving a saved value to the host releases its memory only if nothing else still points at that
-    storage. Three kinds of value still do: the graph's own inputs, which the caller owns; the
-    values the graph returns to the caller; and every other saved value.
+    storage. Four kinds of value still do:
+
+    - the graph's own inputs, which the caller owns;
+    - get_attr nodes, whose tensor the GraphModule holds for the life of the process (attention
+      masks, rotary embedding tables, and the other constants inductor bakes in);
+    - ZeRO-3 gathered parameters, whose buffer belongs to ZeRO's own registry and is released by
+      release_param, not by this graph. The forward graph reaches one of these through
+      dc.wait_allgather, which is an aliasing op, so without this the gathered weight looks like
+      an ordinary activation with nothing else holding it;
+    - the values the graph returns to the caller, and every other saved value.
     """
     counts = defaultdict(int)
-    keepers = [node for node in graph.nodes if node.op == "placeholder"]
+    gather_ops = _zero3_gathered_param_ops()
+    keepers = [node for node in graph.nodes if node.op in ("placeholder", "get_attr") or node.target in gather_ops]
     keepers.extend(returned_to_caller)
     keepers.extend(saved_nodes)
     for keeper in keepers:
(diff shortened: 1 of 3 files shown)


deepspeed/compile/passes/offload_activation.py
```diff
@@ -339,7 +471,11 @@ def _eligible_activations(graph: Graph, graph_id: int, num_fwd_outputs, param_ma
         if size is None or size < min_size:
```
tohtana, line 471: This check still skips a saved view based on the amount of data in the view. For example, a view may reference 4 MiB of data within a 32 MiB GPU allocation. If no other tensor references that allocation, keeping the view keeps all 32 MiB allocated.

Offloading the view would copy only 4 MiB to CPU and allow the full 32 MiB GPU allocation to be freed. However, this check rejects it because 4 MiB is below the default 5 MiB threshold.

Please feel free to refer to [this supplemental PR](https://github.com/pengdurice/DeepSpeed/pull/2).

deepspeed/compile/passes/offload_activation.py
```diff
@@ -279,8 +286,120 @@ def _skip_bytes(node) -> int:
...
+def _has_non_overlapping_storage(node: Node) -> bool:
+    """Whether the tensor's elements each occupy their own place in storage.
+
+    Both the host buffer and the reloaded device tensor are made with `empty_like`, which allocates
+    one element per logical element. A tensor whose elements overlap holds fewer elements of
+    storage than it has entries -- `expand` is the ordinary case, repeating a row with a stride of
+    zero -- so the round trip would copy and allocate several times what the value actually keeps
+    alive. An expanded row of 1000 floats seen as 4x1000 copies 16KB each way to release 4KB.
+
+    Strides that are merely non-contiguous are fine and are deliberately allowed. `empty_like` does
```
tohtana, line 372: I'm not sure it causes an actual issue, but it would be safe to restore the original strides when reloading the tensor onto the GPU. `empty_like` packs stepped slices into a dense buffer, while the reload nodes retain the original tensor metadata.
I think we can save the original shape and strides before offloading, then allocate the destination GPU tensor with those strides, for example using `at::empty_strided`, before copying the values back. The CPU buffer can remain compact.

The reload memory calculation should account for the storage required by the restored strides, including gaps between elements. This would make the runtime layout match the layout expected by the compiled backward.

tohtana (changes requested):
Thank you @pengdurice for the update! I left a few comments about some cornor cases.

deepspeed/compile/passes/offload_activation.py
pengdurice: thank you for the suggestion, the supplemental PR is accepted.

deepspeed/compile/passes/offload_activation.py
pengdurice: I tested whether the different strides give wrong results. They do not, the values before and after are identical.

On the cost. The CPU buffer stays the same size either way, so we move the same number of bytes. What grows is the GPU tensor we allocate when we reload.

Take one chunk of a fused QKV tensor. It holds 384 real values. With its original strides, the first and last value sit 1104 elements apart. Today empty_like allocates 384 elements. empty_strided with the original strides would allocate 1104. That is 2.9 times more GPU memory, on every reload. This pass exists to lower GPU memory, so that is a real trade.

I agree your version is safer, what do you think about the tradeoff?

tohtana: Thanks for testing this and explaining the memory tradeoff! 
Your concern about the memory usage seems valid to me. So let's go with your approach and keep the current packed reload behavior without restoring the original strides.
If we encounter layout-related correctness issues later, we can revisit this with storage-level offloading and reloading that preserves sharing between views. I don't think we need to take on that additional complexity in this PR.

tohtana (approved):

PR #8396, opened 2026-09-02
Fix ZeRO checkpoint DP merge on scalar optimizer state

Kayvan-Zahiri:
`merge_state` concatenates every tensor it meets. With `elastic_checkpoint=False` (the default) ZeRO 1/2 saves `self.optimizer.state_dict()` verbatim (`stage_1_and_2.py:2628`), so per-parameter state carries torch's `step`, a 0-dim tensor for Adam/AdamW. Reshaping that checkpoint to a smaller `dp_degree` sends the shards through `merge_state` and dies:

```
RuntimeError: zero-dimensional tensor (at position 0) cannot be concatenated
```

raised at `reshape_utils.py:111` from `ZeROCheckpoint.get_state_for_rank`. Replicated scalars are now returned unchanged, mirroring the `value.dim() > 0` guard in `stage_1_and_2.py` and the `state_name != "step"` skip in `_strip_tensor_paddings`.

Same file: `merge_state_dict` iterated only `dict_b`, silently dropping keys held by `dict_a` alone, and passed `[str(key)]` rather than the accumulated path, so the mismatch diagnostic printed `groups` instead of `opt.groups`.

`merge_state` had no test coverage. The three added tests fail on master with exactly those symptoms. `test_reshape_checkpoint.py` + `test_convert_checkpoint.py`: 12 passed / 1 skipped before, 15 passed / 1 skipped after. `pre-commit run --files` passes on both changed files.

### DIFF
--- a/deepspeed/checkpoint/reshape_utils.py
+++ b/deepspeed/checkpoint/reshape_utils.py
@@ -80,10 +80,14 @@ def _key_list_to_string(key_list):
 def merge_state_dict(dict_a, dict_b, key_list):
     merged_dict = type(dict_a)({})
 
-    for key, value in dict_b.items():
-        if key in dict_a.keys():
-            merged_dict[key] = merge_state(dict_a[key], dict_b[key], [str(key)])
+    for key, value in dict_a.items():
+        if key in dict_b:
+            merged_dict[key] = merge_state(value, dict_b[key], key_list + [str(key)])
         else:
+            merged_dict[key] = value
+
+    for key, value in dict_b.items():
+        if key not in dict_a:
             merged_dict[key] = value
 
     return merged_dict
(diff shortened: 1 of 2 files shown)


Kayvan-Zahiri:
The sign-off is already there. The commit message on `35e294e` ends with:

```
Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01DQughgr95y4B9H1jfaQH8o
Signed-off-by: kzahiri1 <kzahiri@dons.usfca.edu>
```

`git log -1 --format=%B` shows it, and the DCO check on this PR passes, which is the
gate that enforces AGENTS.md:L8. I suspect the trailer was missed because it sits after
two other trailers rather than last-but-one.

Nothing to recreate here, but happy to reorder the trailers so `Signed-off-by` comes
first if maintainers prefer that.

PR #8406, opened 2026-09-03
Checkpoint the curriculum data sampler's own RNG state

### LINKED ISSUE #8405: [BUG] Curriculum data sampler checkpoints the global numpy RNG, not its own, so resume replays the sampling stream
**Describe the bug**

`DeepSpeedDataSampler` draws from its own generator, `self.np_rng = np.random.default_rng(seed)`, but `state_dict()` stores `np.random.get_state()` and `load_state_dict()` restores it. That is the global legacy `RandomState`, not the generator the sampler uses, so the saved value is the same whether the sampler has drawn nothing or a thousand batches.

The effect shows up on resume. `__init__` seeds `self.np_rng` fresh, `load_state_dict` never touches it, and the sampler goes back to the top of its own stream: same cluster mix per step, same shuffles out of `get_new_cluster` and `reshuffle_clusters`. Restoring the global state also moves whatever else in the process is drawing from `np.random`.

**To Reproduce**

```python
from deepspeed.runtime.data_pipeline.config import get_data_efficiency_config
from deepspeed.runtime.data_pipeline.data_sampling.data_sampler import DeepSpeedDataSampler

metric = {"index_to_sample_path": "dummy", "index_to_metric_path": "dummy", "difficulty_type": "value",
          "clustering_type": "single_cluster", "min_difficulty": 8, "max_difficulty": 80,
          "schedule_type": "fixed_linear",
          "schedule_config": {"total_curriculum_step": 100, "difficulty_step": 8}}
config = get_data_efficiency_config({"data_efficiency": {"enabled": True, "seed": 1234, "data_sampling": {
    "enabled": True, "curriculum_learning": {"enabled": True, "data_cluster_path": "/tmp/clusters",
                                             "curriculum_metrics": {"dummy": metric}}}}})

def sampler():
    s = DeepSpeedDataSampler(config, 100, 8, 0, 1, None, 1, global_rank=0)
    s.data_clusters, s.data_cluster_sizes = [None] * 4, [10, 20, 30, 40]
    return s

saved = sampler()
before = saved.state_dict()["np_rng_state"]
first = [saved.sample_from_clusters().tolist() for _ in range(3)]
print("saved rng state changed by 3 draws:", repr(saved.state_dict()["np_rng_state"]) != repr(before))

resumed = sampler()
resumed.load_state_dict(saved.state_dict())
print("resumed draw 1:", resumed.sample_from_clusters().tolist(), " original draw 1:", first[0])
```

On master at `493dafa`, python 3.12.14, torch 2.14.0+cpu:

```
saved rng state changed by 3 draws: False
resumed draw 1: [0, 3, 3, 2]  original draw 1: [0, 3, 3, 2]
```

**Expected behavior**

A resumed run continues the sampling stream instead of replaying it, and loading a sampler state leaves the global numpy RNG alone.

**Additional context**

Both lines go back to #2585, which added the library. I found this reading the code, not from a training run. PR on the way.

ebarkhordar:
`DeepSpeedDataSampler` draws from `self.np_rng`, the generator it seeds in `__init__`, but `state_dict()` saved `np.random.get_state()` and `load_state_dict()` restored that. The global legacy `RandomState` was never the RNG the sampler used, so the generator stayed out of the checkpoint and a resumed run started its stream again from the seed: same cluster mix per step, same shuffles.

Saves and restores `self.np_rng.bit_generator.state` instead. A state dict written by an older version carries the old tuple, and that shape still goes through `np.random.set_state` so old checkpoints keep loading the way they did.

Added two tests for the resume path and one that loads an old-style state dict. The two behaviour ones fail on master.

Fixes #8405

### DIFF
--- a/deepspeed/runtime/data_pipeline/data_sampling/data_sampler.py
+++ b/deepspeed/runtime/data_pipeline/data_sampling/data_sampler.py
@@ -321,7 +321,7 @@ class DeepSpeedDataSampler(object):
             CURRICULUM_LEARNING_CURRENT_DIFFICULTIES: self.current_difficulties,
             CURRICULUM_LEARNING_DATA_CLUSTER_PATHS: self.data_cluster_paths,
             CURRICULUM_LEARNING_DATA_CLUSTER_CURRENT_POSITION: self.data_cluster_current_position,
-            CURRICULUM_LEARNING_NP_RNG_STATE: np.random.get_state()
+            CURRICULUM_LEARNING_NP_RNG_STATE: self.np_rng.bit_generator.state
         }
 
     def load_state_dict(self, state_dict):
@@ -331,7 +331,13 @@ class DeepSpeedDataSampler(object):
         self.current_difficulties = state_dict[CURRICULUM_LEARNING_CURRENT_DIFFICULTIES]
         self.data_cluster_paths = state_dict[CURRICULUM_LEARNING_DATA_CLUSTER_PATHS]
         self.data_cluster_current_position = state_dict[CURRICULUM_LEARNING_DATA_CLUSTER_CURRENT_POSITION]
-        np.random.set_state(state_dict[CURRICULUM_LEARNING_NP_RNG_STATE])
+        np_rng_state = state_dict[CURRICULUM_LEARNING_NP_RNG_STATE]
+        if isinstance(np_rng_state, dict):
+            self.np_rng.bit_generator.state = np_rng_state
+        else:
+            # checkpoints written before this field held self.np_rng carry the tuple from
+            # np.random.get_state(), so restore that the way those checkpoints expect
+            np.random.set_state(np_rng_state)
         cluster_root_path = self.data_efficiency_config[DATA_SAMPLING][CURRICULUM_LEARNING][
             CURRICULUM_LEARNING_CLUSTER_PATH]
         # Backward compatibility: previously data_cluster_paths were stored as
--- a/tests/unit/runtime/test_data_efficiency.py
+++ b/tests/unit/runtime/test_data_efficiency.py
@@ -5,12 +5,16 @@
 
 import torch
 import os
+import numpy as np
 import deepspeed
 from deepspeed.accelerator import get_accelerator
 import pytest
 from unit.common import DistributedTest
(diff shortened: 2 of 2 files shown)


ebarkhordar:
On the RNG assertion: `bit_generator.state` is numpy's own interface for a `Generator`'s state, and it is exactly the value the checkpoint now stores, so that line compares the field against its source rather than against an internal encoding.

It also pins something the resume test does not, which is *which* generator gets checkpointed. Putting the sampler back on the process-global `np.random` for both the draws and the state would keep `test_curriculum_sampler_resumes_its_rng_stream` green and bring back the shared-state coupling this change is about. So I'd rather keep it.

On the engine-level resume test: `TestDataEfficiency::test_curriculum_learning` already drives the engine, and a `save_checkpoint`/`load_checkpoint` comparison alongside it would be worth having. I have not run one on a GPU, so adding it here would mean shipping a test I cannot report a result for. If you want it in this PR, say so and I will write it.

ebarkhordar:
#8425 opened a couple of hours ago with the same source change as this one, legacy-tuple branch included, so what actually differs is the tests. This PR adds three to `test_data_efficiency.py`: that `state_dict` stores the sampler's own `bit_generator.state`, that a reload continues that stream instead of restarting it from the seed, and that an old-style state dict still loads.

Updated 2026-09-08: there are now three PRs on #8405, since #8460 opened this morning. All three make the same one-file change, saving and restoring `self.np_rng.bit_generator.state` with a branch for the old legacy tuple.

Updated 2026-09-14: #8460 is closed, withdrawn by its own author, so the offer above no longer points anywhere. Two are left, this one and #8425, and only this one carries tests.

The reason he gave was #8489, where the Data Efficiency library is named a candidate for deprecation for lack of bandwidth. If that is the direction, say so and I will close this. If it stays, the tests are the part worth keeping, since nothing else in the suite catches the sampler silently replaying its stream after a resume.

Updated 2026-09-17: #8566 merged and answers the question above. It drops the legacy top-level `curriculum_learning` config and leaves the Data Efficiency path alone, and its own description says `data_sampler.py` is unaffected, so I'll keep this open.

Rebased past it. The only conflict was the import line in `test_data_efficiency.py`, since the legacy tests took `Curriculum_SimpleModel` with them. All three tests pass on the new base, and the two that pin the fix still fail with `data_sampler.py` reverted to master.

PR #8234, opened 2026-08-07
Add adaptive gather-buffer pooling for DeepCompile ZeRO-3

tohtana:
(#8233 needs to be merged before this PR)
Splits the native gather-buffer pool and allocator-pressure recovery changes out of #8169. This is stacked on the scheduler PR and passed 12/12 tests on a single node with 2 H100 GPUs.

PR #8418, opened 2026-09-04
DeepCompile: reuse ZeRO-3 all-gather memory

tohtana:
This PR supersedes #8234 and is stacked on #8233 (#8233 should be merged first)

## Problem

#8234 used separate memory-management mechanisms for normal and prefetched gathers. Because the scheduler already knows when each all-gather result is created and released, this PR replaces those mechanisms with a single reusable memory area.

## Approach

The scheduler determines when an all-gather result is no longer needed and reuses the same memory for later gathers, including prefetched ones. Forward and backward use separate reuse plans and arena storage, and DeepSpeed falls back to normal allocation whenever reuse is unsafe.


pengdurice:
@tohtana , thanks for the great contribution, feel free to share some test results and fix the conflicts!

PR #8424, opened 2026-09-05
Partition AutoEP expert parameters per layer under ZeRO-3

tohtana:
### Problem

When a model created under `zero.Init` is passed to `deepspeed.initialize`, AutoEP creates new local-expert tensors while replacing MoE layers. ZeRO-3 previously converted those tensors only after every layer had been replaced, so full replacement-expert storage accumulated across layers and increased the initialization peak.

This addresses the initialization-memory behavior described as problem 1 in #8353. It does not change the independent debug-map retention behavior covered by #8356.

### Approach

After each AutoEP layer is repacked, copied, tagged, and installed, resolve its expert-data-parallel group and immediately pass only its newly created expert parameters through the existing ZeRO conversion helper. Optimizer classification and grouping use the experts' ZeRO logical shape and size after partitioning. Existing source gathers, router/shared/dense placement, and batched output-recorder retargeting remain unchanged. The later ZeRO setup already skips parameters converted by this step.

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -84,7 +84,10 @@ dist = None
 def set_optimizer_flags(config_class: DeepSpeedConfig, model: torch.nn.Module) -> None:
     if config_class.optimizer_name == MUON_OPTIMIZER:
         for name, p in model.named_parameters():
-            if p.ndim >= 2 and not any(keyword in name.lower() for keyword in ("embed", "lm_head")):
+            is_partitioned_autoep_expert = (getattr(p, "ds_zero_placement_family", None) == "autoep_expert"
+                                            and hasattr(p, "ds_shape"))
+            ndim = len(p.ds_shape) if is_partitioned_autoep_expert else p.ndim
+            if ndim >= 2 and not any(keyword in name.lower() for keyword in ("embed", "lm_head")):
                 setattr(p, "use_muon", True)
             else:
                 setattr(p, "use_muon", False)
--- a/deepspeed/module_inject/auto_ep.py
+++ b/deepspeed/module_inject/auto_ep.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 import math
 import re
 from collections import OrderedDict
-from typing import Literal
+from typing import Callable, Literal
 
 import torch
 import torch.nn as nn
@@ -544,11 +544,14 @@ class AutoEP:
         specs: list[MoELayerSpec],
         ep_size: int,
         ep_rank: int,
+        on_moe_layer_replaced: Callable[[nn.Module], None] | None = None,
     ) -> None:
         """Replace multiple MoE modules and batch post-replacement recorder retargeting."""
         replacements: list[tuple[MoELayerSpec, nn.Module]] = []
         for spec in specs:
             replacement = self._replace_moe_layer_without_retarget(spec, ep_size, ep_rank)
+            if on_moe_layer_replaced is not None:
+                on_moe_layer_replaced(replacement)
             replacements.append((spec, replacement))
             logger.info(f"AutoEP: replaced '{spec.moe_module_name}' with AutoEPMoELayer "
                         f"(ep_size={ep_size}, ep_rank={ep_rank}, "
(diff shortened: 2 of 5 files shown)


deepspeed/__init__.py
```diff
@@ -84,7 +84,10 @@ def _parse_version(version_str):
 def set_optimizer_flags(config_class: DeepSpeedConfig, model: torch.nn.Module) -> None:
     if config_class.optimizer_name == MUON_OPTIMIZER:
         for name, p in model.named_parameters():
-            if p.ndim >= 2 and not any(keyword in name.lower() for keyword in ("embed", "lm_head")):
+            is_partitioned_autoep_expert = (getattr(p, "ds_zero_placement_family", None) == "autoep_expert"
+                                            and hasattr(p, "ds_shape"))
+            ndim = len(p.ds_shape) if is_partitioned_autoep_expert else p.ndim
```
pengdurice, line 89: this can be a bigger bug. regardless of this work: under zero.Init, models are already partitioned and p.ndim = 1 even for 2+ dimension tensors causing all parameters to not use muon. Do we want to take this opportunity to fix this?

tohtana: Thank you @pengdurice for reviewing! You’re right that this affects ordinary ZeRO-3 parameters as well as AutoEP experts: under `zero.Init.

The fix has already landed in #8438. I merged master into this branch and retained its `ds_shape`-based classification. I also updated the AutoEP regression test to verify that both expert matrices and ordinary partitioned matrices remain eligible for Muon.

tohtana:
During validation, I found that discarded ZeRO-3 source parameters could retain their GPU storage after AutoEP replaced the original layers.
ZeRO attaches functions to each parameter that capture the same parameter. A simplified version of the relevant code looks like this:

```python
def _convert_to_deepspeed_param(self, param):
    def partition(param_list=None):
        if param_list is None:
            param_list = [param]  # The function holds a reference to param.
        self._partition(param_list)
    param.partition = partition  # param holds a reference to the function.
```

This creates a cycle: `param → partition function → captured param`. After AutoEP removes the original layer from the model, these internal references can remain, keeping the parameter’s reference count above zero. Ordinary reference counting therefore cannot immediately destroy the parameter or release its associated GPU storage.

Python’s cyclic garbage collector can reclaim this cycle once no external references remain. In the validation run, collection was delayed enough for discarded source weights to accumulate across successive replacements.
So I added `gc.collect()` after each replacement’s expert parameters are converted to ZeRO-3, before constructing the next replacement. This explicitly collects otherwise-unreachable source parameters and makes their GPU storage available for reuse by PyTorch.
Note that the collection is not called from the training loop, including forward, backward, or optimizer steps.**

The reference-cycle mechanism already exists in ZeRO-3; AutoEP exposes its memory impact by replacing and discarding parameters during initialization. Avoiding these cycles would be another approach, but would require a broader change to ZeRO’s parameter lifetime management.

pengdurice:
> During validation, I found that discarded ZeRO-3 source parameters could retain their GPU storage after AutoEP replaced the original layers. ZeRO attaches functions to each parameter that capture the same parameter. A simplified version of the relevant code looks like this:
> 
> ```python
> def _convert_to_deepspeed_param(self, param):
>     def partition(param_list=None):
>         if param_list is None:
>             param_list = [param]  # The function holds a reference to param.
>         self._partition(param_list)
>     param.partition = partition  # param holds a reference to the function.
> ```
> 
> This creates a cycle: `param → partition function → captured param`. After AutoEP removes the original layer from the model, these internal references can remain, keeping the parameter’s reference count above zero. Ordinary reference counting therefore cannot immediately destroy the parameter or release its associated GPU storage.
> 
> Python’s cyclic garbage collector can reclaim this cycle once no external references remain. In the validation run, collection was delayed enough for discarded source weights to accumulate across successive replacements. So I added `gc.collect()` after each replacement’s expert parameters are converted to ZeRO-3, before constructing the next replacement. This explicitly collects otherwise-unreachable source parameters and makes their GPU storage available for reuse by PyTorch. Note that the collection is not called from the training loop, including forward, backward, or optimizer steps.**
> 
> The reference-cycle mechanism already exists in ZeRO-3; AutoEP exposes its memory impact by replacing and discarding parameters during initialization. Avoiding these cycles would be another approach, but would require a broader change to ZeRO’s parameter lifetime management.

Thank you for finding this deeper issue. the gc collect is a good idea since it is less intrusive. Let's fix it later if this still remains an issue.

pengdurice (approved):
LGTM, thank you!

PR #8428, opened 2026-09-06
Fix missing () on is_bf16_supported in ZeRO-3 dtype fallback (silent bf16 on non-bf16 hardware)

promptsmith1990:
Spotted a missing `()` in the dtype fallback chain in `_set_dtype`:

```python
self.dtype = dtype or torch.float16 if get_accelerator().is_fp16_supported()
    else torch.bfloat16 if get_accelerator().is_bf16_supported   # <-- no call
    else torch.float32
```

`is_bf16_supported` is a method on every accelerator, so without the parens you're checking the method object itself, which is always truthy. The `torch.float32` branch can never be reached — hardware that doesn't support fp16 or bf16 silently ends up with bfloat16 instead of float32.

### DIFF
--- a/deepspeed/runtime/zero/partition_parameters.py
+++ b/deepspeed/runtime/zero/partition_parameters.py
@@ -401,7 +401,7 @@ class InsertPostInitMethodToModuleSubClasses(object):
                 self.dtype = torch.float
         else:
             self.dtype = dtype or torch.float16 if get_accelerator().is_fp16_supported(
-            ) else torch.bfloat16 if get_accelerator().is_bf16_supported else torch.float32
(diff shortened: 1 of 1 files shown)


alanhuangyoo (commented):
Good catch. There's one more thing on the same line though: `or` binds tighter than the conditional, so it parses as `(dtype or torch.float16) if fp16 else ...`, and when the accelerator reports no fp16 an explicit `zero.Init(dtype=torch.float32)` still ends up bf16 (checked by stubbing `is_fp16_supported` to False, both on master and this branch). Wrapping the fallback, `dtype or (torch.float16 if ... else torch.bfloat16 if ... else torch.float32)`, keeps the dtype you pass in.

Issue #8432, opened 2026-09-06
[BUG] Muon + fp16 does not train under ZeRO 1/2: the first loss-scale overflow permanently poisons the momentum buffer


alanhuangyoo:
## Summary

Muon under ZeRO 1/2 with fp16 does not train. The first loss-scale overflow folds an overflowed gradient into Muon's momentum buffer, the buffer stays `NaN` for the rest of the run, and every subsequent step therefore overflows too. The loss scaler backs off until it gives up:

```
Exception: Current loss scale already at minimum - cannot decrease scale anymore. Exiting run.
```

No parameter is ever updated. This is with the configuration `tests/unit/ops/muon/test_muon.py` itself uses, only run for longer.

## Isolation

Same configuration, only the initial loss scale differs. 30 steps, 2 GPUs, ZeRO-2, `SimpleModel(hidden_dim=128, nlayers=5)`, `lr=0.05`:

| `initial_scale_power` | parameters changed | momentum finite | outcome |
| ---: | ---: | --- | --- |
| 16 (the default, 65536) | **0 / 10** | **False** | dies at minimum loss scale |
| 0 (scale 1, no overflow) | **10 / 10** | True | trains normally |

So the failure is entirely the overflow interaction, not Muon itself.

## Mechanism

An overflowed step is supposed to be discarded. Muon consumes the gradient before that happens.

`get_flat_partition` (`stage_1_and_2.py`) calls `muon_update` on the accumulated gradient while filling the partition, which runs in `independent_gradient_partition_epilogue`, i.e. before the overflow check that decides whether to skip the step. `muon_update` then does:

```python
momentum.lerp_(grad, 1 - beta)                      # inf/nan enters the momentum
update = grad.lerp_(momentum, beta) if nesterov     # ...and is written back into grad
```

Both halves matter. The first poisons the momentum permanently, since nothing ever resets it. The second writes `NaN` back into the gradient in place, so the overflow check fires again on the next step regardless of what the loss scale has been reduced to — the failure sustains itself.

Step trace, default scale:

```
step   1: loss_scale=65536    overflow=True  changed=0/10  momentum_finite=False
step   2: loss_scale=32768    overflow=True  changed=0/10  momentum_finite=False
step  10: loss_scale=128      overflow=True  changed=0/10  momentum_finite=False
...        loss scale reaches the minimum and the run raises
```

## Why the suite is green

Two things hide it.

[...]

PR #8433, opened 2026-09-06
[muon] Reconcile the momentum dtype when a checkpoint is restored

### LINKED ISSUE #7746: [BUG] Dtype mismatch (bf16 vs fp32) when resuming Muon optimizer from checkpoint
## Bug Description

When resuming training from a checkpoint using `bf16` and the Muon optimizer, a `RuntimeError` occurs due to a dtype mismatch.
- Model parameters and gradients are in `bf16`.
- Optimizer state (`momentum_buffer`) is loaded from the checkpoint as `fp32`.
- The mismatch happens when Muon tries to apply updates (e.g., `lerp_`) between `fp32` momentum buffers and `bf16` gradients.

## Minimal Reproducible Example

```python
import torch
import os
import deepspeed

torch.cuda.set_device(int(os.environ["LOCAL_RANK"]))

def train_step(model_engine, x, y):
    output = model_engine(x)
    loss = ((output - y) ** 2).mean()
    model_engine.backward(loss)
    model_engine.step()

hidden_size = 64
out_size = 1
dtype = torch.bfloat16

# Setup dummy data and model
x = torch.randn(16, hidden_size, dtype=dtype).cuda()
y = torch.randn(16, out_size, dtype=dtype).cuda()
model = torch.nn.Linear(hidden_size, out_size)

ds_config = {
    "bf16": {"enabled": True},
    "zero_optimization": {"stage": 2},
    "optimizer": {"type": "Muon", "params": {"lr": 1e-3}},
    "train_batch_size": 4,
    "train_micro_batch_size_per_gpu": 1
}

model_engine, optimizer, _, _ = deepspeed.initialize(model=model, config=ds_config)


train_step(model_engine, x, y)

model_engine.save_checkpoint("./test_checkpoint")
model_engine.load_checkpoint("./test_checkpoint")

# Resume training -> trigger error
train_step(model_engine, x, y)
```

**Run command:**
```bash
deepspeed --include localhost:0,1,2,3 example.py
```

**Error trace:**
```
[rank2]: torch._dynamo.exc.TorchRuntimeError: Failed running call_method lerp_(*(FakeTensor(..., device='cuda:2', size=(1, 64)), FakeTensor(..., device='cuda:2', size=(1, 64), dtype=torch.bfloat16), 0.050000000000000044), **{}):
[rank2]: expected dtype torch.float32 for `end`, but got dtype torch.bfloat16

[rank2]: from user code:
[rank2]:    File "/usr/local/lib/python3.12/dist-packages/deepspeed/runtime/zero/muon/original_muon.py", line 72, in torch_dynamo_resume_in_muon_update_at_71
[rank2]:     momentum.lerp_(grad, 1 - beta)
```

## Environment
- GPU: `4 H100`
- Docker image: `nvidia/pytorch:25.03-py3`
- torch: `2.7.0a0+7c8ec84dab.nv25.3`
- deepspeed: `0.18.3`

## My Workaround

Debugging showed that `momentum_buffer` is loaded as an `fp32` tensor, which conflicts with `bf16` gradients in [muon_update](https://github.com/deepspeedai/DeepSpeed/blob/v0.18.3/deepspeed/runtime/zero/stage_1_and_2.py#L1923-L1926). I found that manually converting the buffer to `bf16` right after `load_checkpoint()` fixes the crash:

```python
for tensor_key, values_dict in optimizer.optimizer.state.items():
    for key, tensor_value in values_dict.items():
        if key == "momentum_buffer":
            values_dict[key] = tensor_value.to(dtype=torch.bfloat16)
```

This resolves the error, but I'm unsure if forcing `momentum_buffer` to `bf16` is numerically stable or the intended behavior, given that optimizer states are typically kept in `fp32`. Opening this issue to find the proper fix.

alanhuangyoo:
Closes #7746.

## Problem

Resuming a Muon run under ZeRO 1/2 raises as soon as the first gradient arrives:

```
RuntimeError: expected dtype torch.float32 for `end`, but got dtype torch.bfloat16
```

ZeRO 1/2 keep Muon's momentum in a hand-flattened tensor under `optimizer.state[flatten_copy]["momentum_buffer"]`, allocated in the gradient accumulation dtype. A checkpoint stores optimizer state in fp32, so a restored buffer comes back in fp32 while the gradients it is combined with are not, and `muon_update`'s `momentum.lerp_(grad)` requires the two to match.

`get_flat_partition` only asked whether a buffer existed:

```python
if "momentum_buffer" not in self.optimizer.state[flatten_copy] and ...:
    self.optimizer.state[flatten_copy]["momentum_buffer"] = ...
```

so it left the restored one alone. Traced across a resume:

```
fresh run          momentum torch.bfloat16
after initialize   no buffer yet
after load         momentum torch.float32     <- crashes on the next step
```

## Fix

The buffer is converted to the dtype it is about to be combined with. Discarding and reallocating it would also stop the crash, but the momentum a resume just restored is the reason the checkpoint carries it — starting from zero is a different run, and Muon's update is built on that momentum.

## Verification

2 × H20, ZeRO 1/2/3, an uninterrupted run against one interrupted at the halfway point and resumed in a fresh engine:

| stage | master | this branch |
| --- | --- | --- |
| 1 | `RuntimeError` | momentum restored (0.003303 → 0.003303), max param norm drift **0.0** |
| 2 | `RuntimeError` | momentum restored (0.003303 → 0.003303), max param norm drift **0.0** |
| 3 | passes | unchanged |

Drift 0.0 is the part worth noting: the resumed run does not merely survive, it produces the same parameters as the run it resumed.

Stage 3 keeps its momentum through a different path (`_create_momentum_buffer`) and was never affected.

## Tests

`tests/unit/ops/muon/test_muon_checkpoint.py`, ZeRO 1/2/3 × bf16/fp16. Four of the six fail on the parent commit:

```
FAILED test_resumes_with_its_momentum[bf16-1]
FAILED test_resumes_with_its_momentum[bf16-2]
FAILED test_resumes_with_its_momentum[fp16-1]
FAILED test_resumes_with_its_momentum[fp16-2]
4 failed, 2 passed
```

and all six pass here. It asserts both halves: that the restored momentum matches what was saved, and that continuing from the checkpoint lands on the same parameters as never having stopped.

It is a new file rather than a case in `test_muon.py` because that module skips itself entirely where fp16 is unsupported, which would take the bf16 coverage with it.

## Why the suite could not have caught this

`tests/unit/ops/muon/test_muon.py` never saves or loads a checkpoint. It also runs fp16 at the default loss scale, where — separately from this PR, filed as #8432 — every step overflows, so no momentum buffer is ever allocated in a dtype that could disagree with anything.

The fp16 cases here set `initial_scale_power: 4` for that reason: this test is about what a checkpoint carries, not about the scaler.

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -2149,12 +2149,23 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
         if (not self.optimizer.state[flatten_copy]) and getattr(
                 tensor_list[0], 'use_muon', False) and 'muon' in self.optimizer.__class__.__name__.lower():
             self.optimizer.state[flatten_copy] = {}
-        if "momentum_buffer" not in self.optimizer.state[flatten_copy] and getattr(
-                tensor_list[0], 'use_muon', False) and 'muon' in self.optimizer.__class__.__name__.lower():
-            # need to check the total # of elements in the parameters in this group and this partition
-            total_size = sum([t.numel() for t in tensor_list])
-            flatten_bf_list = [torch.zeros([total_size], dtype=dtype, device=device)]
-            self.optimizer.state[flatten_copy]["momentum_buffer"] = self.flatten(flatten_bf_list)
+        if getattr(tensor_list[0], 'use_muon', False) and 'muon' in self.optimizer.__class__.__name__.lower():
+            momentum_buffer = self.optimizer.state[flatten_copy].get("momentum_buffer")
+            if momentum_buffer is None:
+                # need to check the total # of elements in the parameters in this group and this partition
+                total_size = sum([t.numel() for t in tensor_list])
+                flatten_bf_list = [torch.zeros([total_size], dtype=dtype, device=device)]
+                self.optimizer.state[flatten_copy]["momentum_buffer"] = self.flatten(flatten_bf_list)
+            elif momentum_buffer.dtype != dtype:
+                # A restored buffer arrives in the dtype the checkpoint holds optimizer state
+                # in, which is fp32, while the gradients it is combined with are in the
+                # gradient accumulation dtype. muon_update does momentum.lerp_(grad), which
+                # requires both to match, so resuming a bf16 run raised:
+                #   RuntimeError: expected dtype torch.float32 for `end`, but got dtype
+                #   torch.bfloat16
+                # Convert rather than reallocate: the momentum a resume just restored is the
+                # reason the checkpoint carries it.
+                self.optimizer.state[flatten_copy]["momentum_buffer"] = momentum_buffer.to(dtype=dtype, device=device)
 
         buffer_idx = 0
         for i, tensor in enumerate(tensor_list):
--- a/tests/unit/ops/muon/test_muon_checkpoint.py
+++ b/tests/unit/ops/muon/test_muon_checkpoint.py
@@ -1,0 +1,119 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# DeepSpeed Team
+"""Muon has to survive a checkpoint round trip, in the dtype the run is using.
+
+ZeRO 1/2 keep Muon's momentum in a hand-flattened tensor under
+``optimizer.state[flatten_copy]["momentum_buffer"]``, allocated in the gradient
+accumulation dtype. A checkpoint stores optimizer state in fp32, so a restored buffer
+comes back in a different dtype than the gradients it is combined with, and
+``muon_update``'s ``momentum.lerp_(grad)`` requires the two to match.
+
+The existing Muon suite covers neither half of this: it runs in fp16, where the two
+dtypes coincide, and it never saves or loads.
+"""
+
+import pytest
+import torch
+
+import deepspeed
+from deepspeed.accelerator import get_accelerator
+from unit.common import DistributedTest
+from unit.simple_model import SimpleModel
+
+
+def _config(zero_stage, dtype):
+    return {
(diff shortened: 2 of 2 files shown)


delock (approved):

alanhuangyoo:
@delock Thanks for merging this. One thing surfaced right after, which I have opened #8483 for.

This PR's reconciliation reached only one of the two copies of the momentum-buffer setup. `get_flat_partition` delegates to `_get_flat_partition_unpadded` whenever a group has no round-robin padding:

```python
if not any(self.round_robin_bit16_padding[param_group_idx]):
    return self._get_flat_partition_unpadded(...)
```

and that copy still allocates the buffer without reconciling the dtype. So **this PR's own regression tests fail on `master` as merged** whenever a group takes the unpadded path:

```
master (b726f4ed)   4 failed, 2 passed    TestMuonCheckpointRoundTrip::test_resumes_with_its_momentum
                    RuntimeError: expected dtype torch.float32 for `end`, but got dtype torch.bfloat16
```

That is my miss, not a review miss — I wrote the fix against the copy I had been reading and did not check whether the tests were exercising the other one.

#8483 has both paths calling a single `_muon_momentum_buffer` rather than carrying the reconciliation twice, which is what let them diverge in the first place. It also takes the dtype from the gradient rather than from the configured accumulation dtype, because those differ at stage 2 — `partition_gradients` turns `use_grad_accum_attribute` off there, so the gradients stay in the parameter dtype. That is a second, unrelated way to hit the same `lerp_` error, on a fresh run with no checkpoint at all.

With #8483 the six cases here pass.

PR #8434, opened 2026-09-06
Fix SequenceTiledCompute backward for empty trailing shards

taking-lying-flat:
## Description

`sequence_tiled_compute` can finish forward and then raise `IndexError` in backward when `torch.chunk` returns fewer pieces than requested. For `seqlen=5, shards=4`, forward calls the function on slices of lengths `[2, 2, 1, 0]`, while backward creates only three chunks and still indexes four. This also occurs without ZeRO.

Save the forward slice step and reuse it to slice every sharded input during backward. Clamp trailing empty-slice offsets to the sequence length so the gradient `narrow()` calls remain valid. This preserves the existing forward calls, requested shard count, reduction behavior, and final-shard gradient signaling.

The regression test runs two SGD steps against an untiled PyTorch MLP and compares outputs, input gradients, parameter gradients, and updated weights. It covers batch sizes 1/2, `(seqlen, shards)` pairs `(5, 4)`, `(2, 4)`, `(7, 4)`, and `(8, 4)`, multiple sharded inputs, and tensor/sum/mean outputs. The callback supports empty inputs, as required by the existing forward behavior.

## Validation

Hardware: NVIDIA RTX A1000 Laptop GPU, 4 GiB, driver 595.84; Python 3.12.13, PyTorch 2.13.0+cu130, CUDA 13.0.

- Before the fix, the 12 empty-tail regression variants reproduced the backward `IndexError`; the 12 control variants passed.
- `DS_ACCELERATOR=cuda python -m pytest -q tests/unit/ulysses_alst/test_tiled_compute.py -k sequence_tiled_compute_shard_boundaries --tb=short`: **24 passed** in FP32, including both optimizer steps.
- Existing `TestTiledCompute` ZeRO-2 cases: **2 passed** on the GPU.
- Existing ZeRO-3 cases: default communication buffers exhausted this 4 GiB GPU while initializing the third model. Reran with a local-only `reduce_bucket_size=65536` override: **2 passed**. The override is not included in this PR.
- `pre-commit run --files deepspeed/runtime/sequence_parallel/ulysses_sp.py tests/unit/ulysses_alst/test_tiled_compute.py`: passed.
- `python ci/tests_fetcher.py --base origin/master`: no affected tests in its configured `tests/unit/v1` scope; the ALST tests above were run explicitly.

### DIFF
--- a/deepspeed/runtime/sequence_parallel/ulysses_sp.py
+++ b/deepspeed/runtime/sequence_parallel/ulysses_sp.py
@@ -821,6 +821,7 @@ class SequenceTiledCompute(torch.autograd.Function):
 
         with torch.no_grad():
             shard_step = math.ceil(seqlen / shards)
+            ctx.shard_step = shard_step
             output_shards = []
 
             for i in range(shards):
@@ -877,8 +878,6 @@ class SequenceTiledCompute(torch.autograd.Function):
         else:
             grad_requiring_tensor_grad = torch.empty_like(grad_requiring_tensor)
 
-        kwargs_to_shard_shards = {k: list(torch.chunk(v, chunks=shards, dim=1)) for k, v in kwargs_to_shard.items()}
-
         for i in range(shards):
             # when fn involves one or more model weights deepspeed will normally push a grad to
             # reduce per sub-module call, so since we only want it to add a grad for the last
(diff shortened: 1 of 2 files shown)


pengdurice:
@stas00 , hey waiting on your call to see if SequenceTiledCompute is still used.

stas00:
Thank you for the tag, @pengdurice 

It was an attempt to come up with a generic sequence tiling autograd function, but I found it too mind boggling with all the abstraction and extremely difficult to understand, so ended up coding specific classes like TiledMLP and TiledFusedLogitsLoss, which are so much easier to understand.

So I don't know if anybody is attempting to use SequenceTiledCompute

@taking-lying-flat, thank you for the fix and test. If I may ask are you using SequenceTiledCompute for some practical purpose? I would love to hear your context

stas00 (approved):

taking-lying-flat:
Thanks for explaining the background, @stas00! I found this while reviewing the code and reproduced the backward failure with empty trailing shards. I wasn't working from a specific training use case; I wanted to fix the inconsistency between forward and backward slicing and add a regression test to cover it.

Your explanation of the move to `TiledMLP` and `TiledFusedLogitsLoss` is helpful. Thank you for reviewing and approving the fix!

PR #8384, opened 2026-09-01
[muon] Per-head Newton-Schulz for attention projections

alanhuangyoo:
Implements #8367 — per-head Newton–Schulz for attention projections, as proposed there.

Scope grew since I opened this: it started as the kernel only, but the metadata and config half
turned out not to depend on the two questions I left on the issue, so it is all here. Points 1–4
of your proposal, plus unit tests for 5; the convergence run is below under "what is not here".

## 1. Kernel

`muon_update` gains `num_heads`. With it set, an attention projection of shape
`[num_heads * head_dim, in_features]` is viewed as `[num_heads, head_dim, in_features]` and
Newton–Schulz runs on that batch, so each head is orthogonalized against itself instead of
sharing one update direction with every other head — the coupled-block behaviour Kimi K3
(arXiv:2607.24653 §2.5) and GLM-5 Muon Split (arXiv:2602.15763) both move away from. The existing
`max(1, m/n)**0.5` scaling is applied per head block.

As you said, mostly a view: both kernels are already batch-safe, and `muon_update` already had a
batched branch with per-block scaling for expert groups. This reuses that path.

## 2. Metadata

`set_optimizer_flags` tags `muon_num_heads` alongside `use_muon`, so it follows the pattern
already there and does not require AutoTP to be on. Head counts come from
`AutoTPMeta.from_model_config`, the repo's existing reader for them, and the per-head width from
the config:

| leaf | heads | per-head width |
| --- | --- | --- |
| `q_proj`, `query`, `wq` | `num_attention_heads` | `head_dim` |
| `k_proj`, `v_proj`, `key`, `value`, `wk`, `wv` | `num_key_value_heads` | `head_dim` |
| `q_b_proj`, and `q_proj` on an MLA config | `num_attention_heads` | `qk_nope + qk_rope` |
| `kv_b_proj` | `num_attention_heads` | `qk_nope + v_head_dim` |

Different counts under GQA, and using the query count for k/v would silently split them wrong.
MLA does not use `head_dim` for either up-projection, so a `num_heads * head_dim` check rejects
both — on GLM-5.2 that is 16 × 64 = 1024 against a real 4096. Without a `q_lora_rank` there is no
`q_a`/`q_b` pair and the query up-projection is a plain `q_proj` at the same MLA width
(DeepSeek-V2-Lite).

Matching is on the leaf module name, not the path, so `dense` in `attention.output.dense` cannot
[...]

PR #8436, opened 2026-09-06
[muon] Per-head Muon for linear-attention layers: read head geometry from the owning module

### LINKED ISSUE #8420: [Per-Head Muon] Head geometry that lives outside num_attention_heads/head_dim: Kimi-K3 KDA layers and GLM-5.2's DSA indexer
Follow-up to #8367 / #8384, opened at @delock's suggestion so the layout question has its own place to be discussed.

## What is not covered today

#8384 tags a projection for per-head Newton-Schulz by reading the head count from `AutoTPMeta` and the per-head width from the config. That covers standard attention and MLA. Two head-blocked matrices in the very models #8384 was tested against are left on the full-matrix path, because their head geometry is not in the fields the tagger reads.

**Kimi-K3's linear-attention layers.** `inference-optimization/Kimi-K3-0.40B` is a hybrid: `linear_attn_config` gives `full_attn_layers: [4, 8]` and `kda_layers: [1, 2, 3, 5, 6, 7]`, so 2 of its 8 language layers are MLA and 6 are linear attention with convolution. Instantiated, 318 Muon parameters, 4 tagged — both MLA layers, correctly:

| layer kind | parameter | shape | today |
| --- | --- | --- | --- |
| MLA (layers 4, 8) | `q_b_proj` | (768, 256) | per-head, 8 × 96 |
| MLA (layers 4, 8) | `kv_b_proj` | (1024, 128) | per-head, 8 × 128 |
| KDA (6 layers) | `q_proj` / `k_proj` / `v_proj` | (256, 1024) | full-matrix |

The KDA projections are genuinely head-blocked. From the model's own code:

```python
# modeling_kimi_k3_linear.py
self.head_dim = config.linear_attn_config["head_dim"]      # 32
self.num_heads = config.linear_attn_config["num_heads"]    # 8
projection_k_size = self.head_k_dim * self.num_k_heads
self.q_proj = nn.Linear(self.hidden_size, projection_k_size, bias=False)
```

8 × 32 = 256, which is the shape. They are rejected only because `num_attention_heads` (8) and `head_dim` (74) give 8 × 74 = 592, so the shape check declines rather than the tagger recognising them. The head count happens to be right and the width happens to be wrong.

**GLM-5.2's sparse-attention indexer.** `inference-optimization/GLM-5.2-0.8B-A0.8B` has `indexer.wq_b` at (512, 512), blocked on dim 0 by `index_n_heads` 8 × `index_head_dim` 64:

```python
# modeling_glm_moe_dsa.py
self.wq_b = nn.Linear(self.q_lora_rank, self.n_heads * self.head_dim, bias=False)
```

Its count is in `index_n_heads`, which `AutoTPMeta` does not read. The indexer's other two matrices are not candidates: `wk` is (64, 2048), a single shared K with no head axis, and `weights_proj` is (8, 2048), one row per head.

## The questions this needs answered first

1. Does Muon Split apply to linear-attention heads in K3, or is the split specific to full attention?
2. Does it apply to the DSA indexer in GLM-5, or only to the attention proper?

Both are model questions rather than implementation ones, which is why #8384 declines rather than guessing. If either answer is yes, the code change is small; if the answer is no, the current behaviour is already correct and this issue closes.

## If the answer is yes

The general fix is probably to read the head count off the owning module rather than adding config keys one architecture at a time. Every case above carries it as an attribute:

| module | attributes |
| --- | --- |
| `KimiLinearAttention` | `num_heads`, `head_dim` (from `linear_attn_config`) |
| `GlmMoeDsaIndexer` | `n_heads`, `head_dim` (from `index_*`) |
| MLA / standard attention | `num_heads`, and a width derivable per kind |

Walking `named_modules()` to map a parameter to its owner and asking the module is architecture-agnostic, and the shape confirmation in #8384 stays as the guard. It is a larger change than #8384 and would want its own review, which is why it is not folded in there.

## Not in scope

Whether per-head helps on these layers is a separate question from whether the split is well-defined on them. The measurements in #8384 show the mechanism only has something to act on once heads have differentiated: on trained checkpoints the gradient carries a 6–8x median head-norm imbalance that the whole-matrix path reduces to ~3x and per-head takes to ~1.1x, while both mini checkpoints ship untrained weights whose heads are interchangeable.

I am happy to take this if it is wanted.

alanhuangyoo:
Closes #8420. Stacked on #8384 — that branch is the base, so the diff to review is the single commit 622c28de3; the rest is #8384.

@delock answered the two model questions in #8420: Muon Split applies to KDA, and not to GLM's indexer. This implements exactly that.

## The problem

`#8384` derives candidate geometries from the config: head counts through `AutoTPMeta`, per-head widths from `head_dim` or the MLA fields. Kimi-K3's linear-attention layers are built from `linear_attn_config` instead:

```python
# modeling_kimi_k3_linear.py, KimiDeltaAttention
self.head_dim  = config.linear_attn_config["head_dim"]    # 32
self.num_heads = config.linear_attn_config["num_heads"]   # 8
self.q_proj = nn.Linear(self.hidden_size, self.head_k_dim * self.num_k_heads, bias=False)
```

`q_proj` is (256, 1024) = 8 x 32. Nothing at the top level says 32: `head_dim` is 74 and `qk_nope + qk_rope` is 96, so every candidate failed the width check. The head count was right and the width was wrong, so the projections were declined rather than recognized.

## The change

Map the parameter to the module that owns it and ask that module for the geometry it was built with. It is one more candidate, not an override — `#8384`'s shape confirmation still decides, and a module that disagrees with the config on the head count is still ambiguous and still skipped.

Only modules that say they are attention are asked. That is what keeps the indexer out: `GlmMoeDsaIndexer.wq_b` is (512, 512) and `index_n_heads 8 * index_head_dim 64` is exactly 512, so reading `n_heads`/`head_dim` off any module carrying them would tag it. It is declined by the module, not by the shape, and there is a test asserting the geometry does confirm so that the reason stays honest.

The cost of that choice is stated in a test: Qwen3-Next's `Qwen3NextGatedDeltaNet` does not say attention, so it keeps the full-matrix path. That is `#8384`'s behaviour rather than a regression, and widening it is one marker. The alternative — ask every module, name the ones to skip — tags the indexer by default, which is the opposite of what #8420 concluded.

## Both models, instantiated

`inference-optimization/Kimi-K3-0.40B`, 318 Muon parameters:

| owner | parameter | shape | before | after |
| `KimiMLAAttention` (2 layers) | `q_b_proj` | (768, 256) | 8 heads (`mla-q`) | unchanged |
| `KimiMLAAttention` (2 layers) | `kv_b_proj` | (1024, 128) | 8 heads (`mla-kv`) | unchanged |
| `KimiDeltaAttention` (6 layers) | `q_proj` | (256, 1024) | full matrix | **8 heads (`owner-q`)** |
| `KimiDeltaAttention` (6 layers) | `k_proj` | (256, 1024) | full matrix | **8 heads (`owner-k`)** |
| `KimiDeltaAttention` (6 layers) | `v_proj` | (256, 1024) | full matrix | **8 heads (`owner-v`)** |
| `KimiDeltaAttention` | `g_proj` | (256, 1024) | full matrix | unchanged |
| `KimiDeltaAttention` | `f_b_proj`, `b_proj`, `o_proj` | — | full matrix | unchanged |

4 tagged -> 22. `g_proj` is the case worth noting: same 256 x 1024 shape as `q_proj`, in the same module, and it stays whole because its leaf name is not a projection name. There is a test for it.

`inference-optimization/GLM-5.2-0.8B-A0.8B`, 69 Muon parameters — unchanged at 12:

```
TAGGED:   q_b_proj (4096, 512) x6 -> 16 (mla-q)
          kv_b_proj (5120, 128) x6 -> 16 (mla-kv)
DECLINED: wq_b (512, 512) x3   width-mismatch
          wk (64, 2048) x3     width-mismatch
```

## End to end, real checkpoints

bf16, ZeRO-1, `per_head_muon: true`, 6 steps, loaded weights rather than `from_config` (the KDA layer has an uninitialized `dt_bias`, so a randomly initialized hybrid gives NaN on step 0 regardless of the optimizer):

```
Kimi-K3-0.40B   tagged=22 (18 KDA + 4 MLA)   loss 20.00 -> 18.25   all params finite
GLM-5.2-0.8B    tagged=12  indexer_tagged=0  loss 12.29 ->  9.98   all params finite
```

## Tests

12 cases added to `tests/unit/runtime/zero/test_per_head_muon_tagging.py`, 54 passing. Reverting the code change fails exactly the four that assert the new behaviour:

```
FAILED test_linear_attention_geometry_comes_from_the_owning_module[q_proj] - assert None == 8
FAILED test_linear_attention_geometry_comes_from_the_owning_module[k_proj] - assert None == 8
FAILED test_linear_attention_geometry_comes_from_the_owning_module[v_proj] - assert None == 8
FAILED test_the_value_projection_uses_the_value_width                      - assert None == 8
```

The other eight pass both before and after by design — they pin that nothing widened: the indexer, a module that does not say attention, geometry that does not match the shape, the non-projection matrices of a KDA layer, and a standard Llama where the config and the module agree.

The two existing tests that assert Kimi's KDA is declined on a config-only model are unchanged and still pass. A `SimpleNamespace` config has no modules, so they now document the narrower thing they were always testing: the config alone cannot describe this layout.

Whether per-head helps on linear-attention heads is a separate question from whether the split is well-defined on them, and #8384 has the measurements on that.


alanhuangyoo:
Rebased onto the updated #8384 branch (0a1fbe1). Two fixes landed in the base since this was opened, both from the same review pass and both affecting what this PR sits on:

- **Tensor parallelism.** `set_optimizer_flags` runs before AutoTP partitions, and AutoTP replaces the parameter's `.data` in place, so the head count made against the whole model rode onto a shard and Newton-Schulz ran on half-heads. The count is now re-derived from the shard, which also makes per-head exact under TP — `torch.equal`, not close.
- **`zero.Init`.** Under ZeRO-3's `zero.Init` every parameter reports as 1-D, so nothing confirmed and the flag raised. Shapes are read from `ds_shape` now.

Neither changes what this PR does: Kimi-K3 still goes 4 tagged -> 22, GLM-5.2 stays at 12 with the indexer untouched, and the whole per-head suite is 70 passed.

The commit to review here is still the single one on top.

alanhuangyoo:
Merged the base branch in (`fbf60f6`) — this had gone to CONFLICTING too, for the same reason: #8438 landed `_layer_shape` in the block of `set_optimizer_flags` this stack rewrites. No conflict to resolve here, the resolution was made once on #8384 below.

Re-ran on the merge result, 2×H20:

- `tests/unit/runtime/zero/test_per_head_muon.py` — 72 passed (60 from the base branch plus this one's linear-attention cases)
- `tests/unit/ops/muon/test_per_head_muon_accelerator.py` — 17 passed, 2 GPU
- master's `test_muon_use_muon_under_zero_init.py` + `test_muon_without_zero_optimizer.py` — 14 passed

This still stacks on #8384 and wants that reviewed first.

alanhuangyoo:
@delock #8384 is in, so this is now only the linear-attention part: 3 files, +252/-6 against master (head geometry read from the owning module for KDA, plus tests and docs). I merged master after the squash, and `test_per_head_muon.py` passes on CPU (70 passed). Ready for a look when you have time.

delock:
**Collision case: when `num_k_heads * head_k_dim == num_attention_heads * head_dim`, the config candidate exact-matches the width and wins the tie, so the owner fallback is never consulted — the projection gets tagged with the config's head count even though the module was built with different ones.**

In `_resolve_attention_head_count`, the config-derived candidates come first and `_confirm` returns the first exact width match, so a linear-attention projection whose width collides with `num_attention_heads * head_dim` is resolved in favor of the config. Both factorizations are shape-valid, so the update is then split into blocks that cut across the layer's true head boundaries — silently, since nothing fails.

This makes the owner fallback reachable only when the widths differ, which is also the only case where it's demonstrably needed; the colliding case is exactly what neither the KDA tests nor the indexer guard covers.

My suggestion would be maintain a white list for llinear attention for the time being, and maintain it later if linear attention become prevalent.  Annotate model name for each entry in the white list to help tracking.

docs/_pages/config-json.md
```diff
@@ -80,6 +80,15 @@ What is tagged, and what deliberately is not:
 | `o_proj` and other output projections | no | the head structure is on the input dimension, so splitting dim 0 would cut across the wrong axis |
 | fused `qkv_proj` / `query_key_value` / `c_attn` / `wqkv` | no | the three sections do not share a head count under GQA |
 | MLA `q_a_proj`, `kv_a_proj_with_mqa` | no | down-projections mixing latent and rope components, with no head structure |
+| linear-attention `q_proj` / `k_proj` / `v_proj` | yes | blocked by the head count the attention module was built with, which for hybrids such as Kimi-K3 is not `num_attention_heads * head_dim` |
```
delock: Is it possible to list models list (Kimi-K3/GLM5) that per head muon is supported for their linear attentions here.

What I'm thinking is developer who care about other models with linear attention can take a look here and know they need to do some work explicitly.

delock:
@alanhuangyoo thanks for your continuous work on per-head muon.  I have left my comments.

alanhuangyoo:
@delock Thanks, you're right about the collision: with the config candidates first, an exact width tie went to the config. Both changes are in 64bfdcbb5:

- **Whitelist.** `_LINEAR_ATTENTION_OWNERS` in `deepspeed/__init__.py` maps class name to the model it was checked on. It currently lists `KimiDeltaAttention` for Kimi-K3. The old "class name says attention" gate is gone.
- **The listed module decides.** For a listed owner, the config is not consulted, so the collision can't happen. `test_a_listed_module_wins_when_the_config_width_collides` covers config 8 x 32 against module 4 x 64, both 256 rows, and expects 4 heads. It fails with the config-first order.
- `test_an_unlisted_module_is_not_asked` pins that a module with the same attributes but a different class name (`GatedDeltaNet`, `SomeOtherLinearAttention`) keeps the full-matrix path.

`test_per_head_muon.py` passes on CPU.

docs/_pages/config-json.md
alanhuangyoo: Done in 64bfdcbb5. There's now a table of supported models with their linear-attention module (Kimi-K3 / `KimiDeltaAttention`), a row saying linear attention in any other model stays on the full-matrix path, and a note on how to add one. GLM-5.2's DSA indexer is named in the "not tagged" row. GLM-5.2's MLA projections are already covered through the config, as in #8384.

delock (approved):

Issue #7713, opened 2025-12-04
[BUG] The different lr for Muon doesn't work?


Zazexy:
I highly appreciate the improvements you've made in supporting the Muon optimizer. However, I've run into an issue when trying to set the individual `muon_lr` and `adam_lr` within the configuration JSON.
I've found that the optimizer still defaults to setting a single `lr` for both the Muon parameter group and the Adam parameter group. I've verified that I am on the latest commit, as the relevant processing code for this feature branch is present in my local `engine.py`.
All other settings in the configuration are working as expected, but the specific `muon_lr` and `adam_lr` settings are not taking effect. Has anyone else encountered this issue?
    "optimizer": {
        "type": "Muon",
        "params": {
            "lr": 1e-4,
            "betas": [0.9, 0.95],
            "eps": 1e-8,
            "weight_decay": 0.01,
            "muon_lr": 2e-2,
            "adam_lr": 1e-5
        }

KeeProMise:
@Zazexy Hello! Thank you for providing the detailed information. According to your description, you have set `muon_lr` and `adam_lr` in the configuration JSON, but you found that these settings did not take effect, and the optimizer still uses a single `lr` value. I noticed that there is indeed logic in the code where `muon_lr` and `adam_lr` override `lr`. To further confirm the issue, could you please confirm whether you are using the latest version of the code? Additionally, I suggest trying to apply the Pull Request (PR) I previously submitted to see if the issue still persists. If the problem remains, please provide more debugging information so that we can better locate and resolve the issue. Looking forward to your reply!

Zazexy:
Thanks for your patient reply! I do many checks about the version and I pull your PR as well. I found the reason is that I'm using `accelerate` in my project. It looks like some default settings of optimizer just override the config. What I do is create a dummy optimizer and afford it to accelerator. Result is the other parameters like `eps` and `betas` are set to optimizer, however, no matter `lr` or `muon_lr` is override by acclerate.
Is there any method to apply muon with deepspeed in accelerate? 
`optimizer = DummyOptim(params_to_optimize)
lr_scheduler = DummyScheduler(
            optimizer, total_num_steps=config.trainer.max_train_steps, warmup_num_steps=config.trainer.lr_warmup_steps * accelerator.num_processes
        )
train_module, optimizer, train_dataloader, lr_scheduler = accelerator.prepare(
        train_module, optimizer, train_dataloader, lr_scheduler
    )`

delock:
[...]

Issue #7657, opened 2025-10-29
[REQUEST] Muon Optimizer - Different LR for Different Groups


DongmingShenDS:
**Is your feature request related to a problem? Please describe.**
When using Muon optimizer, now the method is to setup that in config. Params will be grouped into muon & adam automatically per #7555. However, there's a single LR that passed into both groups - this is not the standard pattern. We should have different LR for the muon group and the adam group (usually, muon should have larger LR). I wonder if there's a way to enable this behavior easier from my side. If not, then we probably need to add a support for different LR. 

**Describe the solution you'd like**
Easiest way to achieve this is probably to have "muon_lr" and "adam_lr" instead of a single "lr" key. Then in different groups, fetch accordingly.

delock:
The problem statement is when optimizer is marked as "muon", actually both adam and muon optimizer will be used -- muon for >2D weights for hidden layers, adam for the rest.  By default "lr" will be learning rate for both optimizer, and this will not be sufficient for advanced users.

The solution could be use "lr" as default learning rate.  And "muon_lr" and "adam_lr" as override.   If one of them is specified, the specified value will be used instead of default learning rate, this makes these two config item optional.

@PKUWZP @sfc-gh-truwase FYI.

sowndappan5:
I agree. Since Muon internally splits parameters into Muon and Adam groups, exposing separate muon_lr and adam_lr configuration options would better align with common Muon training recipes and remove the need for custom optimizer setup. It would also make DeepSpeed's Muon integration more flexible while keeping the user experience simple.

sfc-gh-truwase:
@DongmingShenDS and @sowndappan5 thanks for the proposal of `adam_lr` and `moun_lr` overrides. Are you able to submit a PR? Thanks!

sowndappan5:
Thanks! Yes, I can take a look and work on a PR for this enhancement.

I'll first investigate how the Muon and Adam parameter groups are created and how learning rates are currently propagated through the configuration and optimizer setup. If the change is straightforward and backward-compatible, I'll prepare a PR with support for separate `muon_lr` and `adam_lr` overrides.

I'll post my findings here as I progress.

sowndappan5:
[...]

PR #8440, opened 2026-09-06
Muon silently discards the param groups it is given

alanhuangyoo:
Muon's param groups are built by DeepSpeed rather than taken from the caller, because which half of Muon a parameter belongs to is a property of the parameter. Doing that flattened away whatever the incoming groups set.

## What happens

`_configure_basic_optimizer` collapsed `model_parameters` into one list and rebuilt two groups from the config:

```python
# Flatten param group dicts (created by MoE/EP) into a raw parameter list
all_params = []
for item in model_parameters:
    if isinstance(item, dict):
        all_params.extend(item['params'])
    ...
```

The comment says MoE/EP, and for those groups it is harmless — the MoE identity rides on `param.group_name` and `split_params_into_different_moe_groups_for_optimizer` rebuilds them afterwards. But the same line eats a caller's own groups, and every other optimizer here receives `model_parameters` unchanged, so for them a group's `lr` and `weight_decay` reach the optimizer.

The pattern this breaks is the one in most training recipes: no weight decay on biases and norms. Passing the usual two groups, `wd 0.1` and `0.0`, `lr 1e-3` and `1e-4`, against a config of `lr 5e-4, weight_decay 0.01`:

```
AdamW   group 1: lr=1.0e-03 wd=0.1      group 2: lr=1.0e-04 wd=0.0
Muon    muon-params: lr=5.0e-04 wd=0.01  adam-params: lr=5.0e-04 wd=0.01
```

Everything falls back to the config values. The parameters the caller excluded from weight decay are decayed at 0.01 anyway, and nothing says so.

## The change

Split each incoming group into its Muon and Adam halves and carry that group's settings onto both. Settings resolve most-specific-last: the config's shared value, then `muon_lr` / `adam_lr`, then whatever the group itself sets.

After:

```
Muon    group0-muon-params: lr=1.0e-03 wd=0.1   group1-adam-params: lr=1.0e-04 wd=0.0
```

Also raise on a parameter with no `use_muon` attribute. Today that logs an error and then dies two lines later on `p.use_muon` with an `AttributeError`, so the helpful message is followed by an unhelpful traceback.

## Backward compatibility

A plain parameter list — the common case, and every existing test — is unchanged, group names included. Verified byte-for-byte against master:

```
master     at init  muon-params: lr=2.000e-02  adam-params: lr=1.000e-05
this PR    at init  muon-params: lr=2.000e-02  adam-params: lr=1.000e-05
```

Names only gain a prefix when there is more than one incoming group, which is required anyway: MoE regrouping keys its buckets by `name`, so two groups must not collide on one. Nothing in the repo reads `'muon-params'` / `'adam-params'`.

`muon_lr` / `adam_lr` still override the shared `lr`, and now lose to a group that sets its own `lr`, which is the same precedence every other optimizer has.

## Tests

`tests/unit/runtime/zero/test_muon_param_groups.py`, 11 cases. Ten call the grouping directly (it is a staticmethod, so they need no engine and no GPU): the historical names and values for a plain list, `muon_lr`/`adam_lr` overriding, a group keeping its own weight decay and its own lr, a group without an lr still taking `muon_lr`, a mixed group splitting in two with both halves keeping its settings, distinct names across groups, each half getting only the keys its optimizer accepts, frozen parameters left out, and the untagged-parameter error.

The eleventh is the regression test: a real `deepspeed.initialize` with the two-group no-decay pattern. On master it fails with

```
AssertionError: the groups the user passed were not honoured: [0.01, 0.01]
```

and passes here with `[0.0, 0.1]`. The other ten exercise a helper that does not exist on master, so they fail there for the trivial reason; the end-to-end one is the one that fails for the right reason.

yapf and flake8 clean.

Related: #7657 asked for different learning rates for the Muon and Adam halves, which `muon_lr`/`adam_lr` covers at the config level; this makes the general per-group form work too. #7713 is a different way the configured Muon lr fails to reach the optimizer (an LR scheduler broadcasting one scalar over both groups) and is not addressed here.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -2187,6 +2187,71 @@ class DeepSpeedEngine(Module):
             return None, {}
         return FusedAdam, {'adam_w_mode': adam_w_mode}
 
+    # Which of the optimizer's config keys each half of a Muon param group accepts. Muon takes a
+    # momentum and a Newton-Schulz method; the auxiliary Adam takes betas and eps.
+    _MUON_HALF_KEYS = ("lr", "momentum", "weight_decay", "ns_method")
+    _ADAM_HALF_KEYS = ("lr", "betas", "eps", "weight_decay")
+
+    @staticmethod
+    def _muon_half_defaults(optimizer_parameters, keys, lr_override):
+        """Config-level settings for one half, with muon_lr / adam_lr overriding the shared lr."""
+        defaults = {key: optimizer_parameters[key] for key in keys if key in optimizer_parameters}
+        if lr_override in optimizer_parameters:
+            defaults["lr"] = optimizer_parameters[lr_override]
+        return defaults
+
+    @staticmethod
+    def _muon_param_groups(model_parameters, optimizer_parameters):
+        """Split each incoming param group into its Muon and Adam halves.
+
+        Muon has to build its own groups, because which half a parameter belongs to is a
+        property of the parameter rather than of the config. The incoming groups still have to
+        survive that: every other optimizer here receives `model_parameters` unchanged, so a
+        group's own `lr` or `weight_decay` reaches it. Flattening the groups into one list threw
+        those away, and the no-weight-decay-on-biases-and-norms grouping that most training
+        recipes use was silently ignored - the parameters the user excluded were decayed at the
+        config's rate instead, with nothing reported.
+
+        Settings are resolved most-specific-last: the config's shared value, then `muon_lr` /
+        `adam_lr`, then whatever the group itself sets.
+        """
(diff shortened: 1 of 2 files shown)


delock (approved):

delock:
Hi @alanhuangyoo thanks for your PR.
We have slack channel in project front page.  If you join that channel, we may create a discussion group for muon to share usage of muon and bugs need to fix and features wanted.  That could help community collaboration.

alanhuangyoo:
@delock Sorry for the slow reply — this slipped past me. Joined the workspace just now, happy to help with a Muon channel.

PR #8451, opened 2026-09-07
Add opt-in AutoEP Python GC policy

yh0903:
## Summary

Add an experimental, opt-in AutoEP policy for disabling automatic Python cyclic garbage collection during training:

```json
{
  "expert_parallel": {
    "enabled": true,
    "python_gc_policy": "disable_during_training"
  }
}
```

The default is `"default"` and does not change Python GC settings. Opting in affects the entire process, not just one engine. Python reference counting remains active.

## Motivation

Fixed-routing Qwen3-30B-A3B EP16 profiling found repeated single-rank generation-2 GC pauses of 234–327 ms. Other ranks then waited for the paused rank at the next DeepEP dispatch, exposing the host-side pause as collective wait time.

Across 20 measured steps, 27 generation-2 collections accumulated 7.60 s; all long collections occurred during forward. Generation-0/1 collections had maximum durations of only 1.04/2.31 ms.

## Implementation

- When the first managed engine has finished initialization, collect once and disable automatic cyclic GC if it was previously enabled.
- Reference-count the process-wide policy across multiple engines, preserving a pre-disabled GC state.
- Restore the original GC state when the last managed engine is explicitly destroyed.
- Restore previously enabled GC in forked workers and use generation tokens so inherited engines cannot release child-process leases.
- Release the GC lease even if another engine teardown operation raises.
- Expose `engine.collect_python_gc()` for explicit collection at application-selected safe boundaries.

## AutoEP-only performance

The isolated comparison used one 16-GPU H100 allocation, a 48-layer model, fixed routing, dual warmup, and the order `default → managed → managed → default`. Each arm measured 20 steps, giving 40 measured steps per policy.

| Policy | Full measured-window mean per step | Median of arm medians | Median of arm p95s | Steps >800 ms | Steps >1 s |
|---|---:|---:|---:|---:|---:|
| Default | 688.57 ms | 535.13 ms | 1072.52 ms | 11/40 | 4/40 |
| Managed GC | 526.42 ms | 522.89 ms | 551.12 ms | 0/40 | 0/40 |

The full measured-window mean decreased by 162.14 ms per step, or 23.5%, primarily through fewer slow steps rather than a comparable reduction in median step time. Both paired measured-window contrasts favored the managed policy: 166.37 ms and 157.92 ms per step.

- Peak allocated/reserved GPU memory was unchanged: 45,106,782,720 / 50,899,976,192 bytes.
- Observed routes were identical across all arms.
- Maximum paired measured-window loss differences were 0.00266 and 0.01439, within the experiment's 0.02 tolerance.

These are short-run, workload-specific observations from the tested baseline, not a guaranteed speedup across models or subsequent implementation changes. This comparison does not establish long-run training equivalence.

## Testing Done

- Targeted config parsing and validation tests.
- GC manager lifecycle tests covering multiple engines, pre-disabled GC, explicit collection, fork restoration, stale pre-fork leases, and reentrant finalizers.
- Changed files pass repository pre-commit hooks.
- The AutoEP-only distributed ABBA experiment checked routing, loss tolerance, GC state, GPU peak memory, and tail latency.

## Usage and limitations

Disabling automatic cyclic GC defers collection; it does not eliminate the work or fix live references that retain tensors. Cyclic objects can accumulate during long runs, including cycles that retain GPU tensors. Applications should monitor host and device memory and call `engine.collect_python_gc()` at coordinated, application-selected safe boundaries, such as after checkpointing. The cost of these explicit collections must still be included when assessing overall training throughput.

Call `engine.destroy()` when an engine is no longer needed. Restoration does not rely on garbage collection of the engine itself. This PR does not add automatic periodic collection, and the short experiment above does not establish long-run memory stability.

This policy is independent of the DeepEP local-preparation cleanup in #8423.

### DIFF
--- a/deepspeed/__init__.py
+++ b/deepspeed/__init__.py
@@ -258,6 +258,8 @@ def initialize(
     # Restore zero.Init context if necessary
     zero.partition_parameters.restore_init_context()
 
+    engine._configure_python_gc()
+
     return_items = [
         engine,
         engine.optimizer,
--- a/deepspeed/module_inject/auto_ep_config.py
+++ b/deepspeed/module_inject/auto_ep_config.py
@@ -62,6 +62,7 @@ def parse_autoep_config(param_dict: dict) -> AutoEPConfig:
     config.comm_num_sm = param_dict.get("comm_num_sm", 12)
     config.comm_qp_margin = param_dict.get("comm_qp_margin", 4)
     config.comm_max_tokens_per_rank = param_dict.get("comm_max_tokens_per_rank", 0)
+    config.python_gc_policy = param_dict.get("python_gc_policy", "default")
     config.num_expert_groups = param_dict.get("num_expert_groups", None)
     config.num_limited_groups = param_dict.get("num_limited_groups", None)
     config.score_func = param_dict.get("score_func", "auto")
@@ -118,6 +119,11 @@ def validate_autoep_config(
 
     if not isinstance(config.validate_folding_routing, bool):
         raise ValueError("expert_parallel.validate_folding_routing must be a boolean")
+
+    valid_python_gc_policies = ("default", "disable_during_training")
+    if config.python_gc_policy not in valid_python_gc_policies:
+        raise ValueError(f"python_gc_policy must be one of {valid_python_gc_policies}, "
+                         f"got {config.python_gc_policy!r}")
 
     if not config.enabled:
         return
--- a/deepspeed/module_inject/auto_ep_presets/base.py
+++ b/deepspeed/module_inject/auto_ep_presets/base.py
@@ -115,6 +115,7 @@ class AutoEPConfig:
     comm_num_sm: int = 12
     comm_qp_margin: int = 4
     comm_max_tokens_per_rank: int = 0
+    python_gc_policy: Literal["default", "disable_during_training"] = "default"
     num_expert_groups: int | None = None
     num_limited_groups: int | None = None
     score_func: Literal["auto", "softmax", "sigmoid"] = "auto"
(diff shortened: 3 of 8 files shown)


tohtana (changes requested):
I have two suggestions about the configuration API:

1. Could we move this option outside the `expert_parallel` section? Although the motivating workload uses AutoEP, disabling automatic Python cyclic GC is a process-wide setting and does not inherently depend on expert parallelism. Exposing it as a general DeepSpeed option would better reflect its scope.
2. Would a boolean `disable_python_gc` option be clearer than `python_gc_policy`? The current choices are effectively “leave the existing GC behavior unchanged” and “disable automatic cyclic GC.” Unless there are concrete plans for additional policies, a boolean seems sufficient and more direct. Here, `false` should preserve the existing GC state rather than explicitly enable GC.

yh0903:
> I have two suggestions about the configuration API:
> 
> 1. Could we move this option outside the `expert_parallel` section? Although the motivating workload uses AutoEP, disabling automatic Python cyclic GC is a process-wide setting and does not inherently depend on expert parallelism. Exposing it as a general DeepSpeed option would better reflect its scope.
> 2. Would a boolean `disable_python_gc` option be clearer than `python_gc_policy`? The current choices are effectively “leave the existing GC behavior unchanged” and “disable automatic cyclic GC.” Unless there are concrete plans for additional policies, a boolean seems sufficient and more direct. Here, `false` should preserve the existing GC state rather than explicitly enable GC.

Thanks for the review, and for pushing on the scope of this option  @tohtana!  I think it's clearly better this way.
1. I moved out of `expert_parallel`. It's now a top-level DeepSpeed option, parsed in `runtime/config.py` alongside the other top-level scalars, with the constant in `runtime/constants.py`. It no longer depends on AutoEP being enabled at all.  auto_ep_config.py  and the AutoEP preset no longer carry it, and the AutoEP unit tests that asserted on it moved to the runtime config tests, since it isn't an AutoEP concern any more.
2. Now a boolean.  `"disable_python_gc": true | false` , defaulting to  false . The old `python_gc_policy` string and its validation are gone.

tohtana (approved):

PR #8454, opened 2026-09-08
fix(inference): reject non-positive max_out_tokens at config validation

### LINKED ISSUE #8339: [BUG] DeepSpeed accepts negative max_out_tokens and fails generation requests at runtime

chakshu-dhannawat:
Fixes #8339

DeepSpeed currently accepts a negative (or zero) value for `max_out_tokens` during inference config parsing. The failure only surfaces later at generation time with a confusing "Input with size N exceeds maximum length of -1" message.

This PR adds a small pydantic field validator on `max_out_tokens` so that non-positive values raise a clear `ValueError` when the config is constructed.

Changes:
- `deepspeed/inference/config.py`: add `validate_max_out_tokens` field validator.
- `tests/unit/inference/test_inference_config.py`: add CPU-only regression tests for -1, 0, and positive values.

Verified:
- `CUDA_VISIBLE_DEVICES="" python -m pytest tests/unit/inference/test_inference_config.py -m inference -k TestInferenceConfigValidation -v` passes (3/3).
- `pre-commit run --files deepspeed/inference/config.py tests/unit/inference/test_inference_config.py` passes (yapf, flake8, license, codespell, check-torchdist, etc.).

### DIFF
--- a/deepspeed/inference/config.py
+++ b/deepspeed/inference/config.py
@@ -310,6 +310,12 @@ class DeepSpeedInferenceConfig(DeepSpeedConfigModel):
             return field_value
         raise TypeError(f"Invalid type for dtype: {type(field_value)}")
 
+    @field_validator("max_out_tokens")
+    def validate_max_out_tokens(cls, field_value, values):
+        if field_value <= 0:
+            raise ValueError(f"max_out_tokens must be a positive integer, got {field_value}")
+        return field_value
+
     @field_validator("moe")
     def moe_backward_compat(cls, field_value, values):
         if isinstance(field_value, bool):
--- a/tests/unit/inference/test_inference_config.py
+++ b/tests/unit/inference/test_inference_config.py
@@ -82,3 +82,27 @@ class TestInferenceConfig(DistributedTest):
             config = DeepSpeedInferenceConfig(moe=value)
             assert isinstance(config.moe, DeepSpeedMoEConfig)
             assert config.moe.enabled == value
+
+
+@pytest.mark.inference
+class TestInferenceConfigValidation:
+    """CPU-only validation tests for DeepSpeedInferenceConfig."""
+
+    def test_negative_max_out_tokens_rejected(self):
+        # Regression test for https://github.com/deepspeedai/DeepSpeed/issues/8339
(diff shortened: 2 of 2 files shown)


ebarkhordar (commented):
I ran this against `e368cb00` and against current master `468a8ad1` in a clean container (python:3.11-slim, torch 2.14.0+cpu, pydantic 2.13.5).

This validation is already on master. #8343 landed as `7c1b0f2c` on 2026-09-14 and closed #8339, putting the constraint on the field itself:

```python
max_out_tokens: int = Field(1024, alias="max_tokens", gt=0)
```

At master all three cases covered here are rejected, the alias included: `max_out_tokens=-1`, `max_out_tokens=0` and `max_tokens=-1` each raise `Input should be greater than 0`.

One thing worth knowing before you rebase. The merge is clean, but `gt=0` is applied during field validation and `@field_validator` defaults to `mode="after"`, so the constraint raises first and `validate_max_out_tokens` never runs. Two of the three tests here then fail on the `match=` regex:

```
$ python -m pytest tests/unit/inference/test_inference_config.py -m inference -k TestInferenceConfigValidation
FAILED ...::test_negative_max_out_tokens_rejected - AssertionError: Regex pattern did not match.
  Expected regex: 'max_out_tokens must be a positive integer'
  Actual message: 'Input should be greater than 0 [type=greater_than, input_value=-1, ...]'
2 failed, 1 passed, 10 deselected
```

Looks superseded to me.

chakshu-dhannawat:
Thanks for the PR. As reviewer ebarkhordar noted, the fix already landed in #8343, so this is no longer needed. Closing as superseded.

PR #8460, opened 2026-09-08
Checkpoint the RNG the curriculum sampler actually draws from

### LINKED ISSUE #8405: [BUG] Curriculum data sampler checkpoints the global numpy RNG, not its own, so resume replays the sampling stream

alanhuangyoo:
Fixes #8405, reported by @ebarkhordar.

## Problem

Every draw in `DeepSpeedDataSampler` goes through `self.np_rng`, a `np.random.default_rng` seeded in `__init__`:

```python
self.np_rng = np.random.default_rng(self.data_efficiency_config[DATA_EFFICIENCY_SEED])   # :64
...
self.np_rng.shuffle(new_cluster)                                                          # :217  get_new_cluster
self.np_rng.choice(num_clusters, self.global_batch_size, replace=True, p=weights)         # :236  sample_from_clusters
self.np_rng.shuffle(cluster)                                                              # :246  reshuffle_clusters
self.np_rng.shuffle(batch)                                                                # :289
```

but the checkpoint stored and restored something else entirely:

```python
CURRICULUM_LEARNING_NP_RNG_STATE: np.random.get_state()      # :324  state_dict
np.random.set_state(state_dict[CURRICULUM_LEARNING_NP_RNG_STATE])   # :334  load_state_dict
```

That is the process-wide legacy `RandomState`, which this sampler never touches. So the saved value reads the same whether the sampler has drawn nothing or a thousand batches, and on resume `__init__` re-seeds `self.np_rng` while `load_state_dict` leaves it alone — the sampler restarts its own stream, reproducing the same cluster mix per step and the same shuffles it produced the first time. Restoring the global state also moved whatever else in the process draws from `np.random`.

## Solution

Save and restore `self.np_rng.bit_generator.state`.

A checkpoint written before this holds the legacy tuple, and there is no sampler stream recorded in it to restore. `_load_np_rng_state` recognises the shape, warns once on rank 0, and leaves the sampler on its configured seed — which is where those checkpoints already resumed — rather than refusing to load.

## Verification

`tests/unit/runtime/data_pipeline/test_curriculum_sampler_rng.py`, four cases:

| test | on master | here |
| --- | --- | --- |
| the saved state moves as the sampler draws | ❌ identical before and after 3 draws | ✅ |
| resume continues the stream rather than replaying it | ❌ `[[0, 3, 3, 2], ...] != [[2, 1, 3, 2], ...]` | ✅ |
| loading leaves the global numpy RNG alone | ✅ | ✅ |
| a legacy checkpoint still loads | ✅ | ✅ |

The resume test goes through `load_state_dict` rather than poking the generator, so it fails on master for the behaviour rather than on a type error, and it asserts that a sampler on a fresh seed does *not* already agree — otherwise it would prove nothing.

```
1×H20:  4 passed   (this branch)
        2 failed, 2 passed   (master, same test file)
```

`pre-commit run` passes on both files.

### DIFF
--- a/deepspeed/runtime/data_pipeline/data_sampling/data_sampler.py
+++ b/deepspeed/runtime/data_pipeline/data_sampling/data_sampler.py
@@ -313,6 +313,22 @@ class DeepSpeedDataSampler(object):
                 self.consumed_samples += len(current_batch)
                 current_batch = []
 
+    def _load_np_rng_state(self, rng_state):
+        """Restore `self.np_rng`, tolerating checkpoints written before this was fixed.
+
+        Older checkpoints hold the tuple `np.random.get_state()` returns for the
+        legacy global RandomState. There is no sampler stream recorded in it, so the
+        best that can be done is to leave `self.np_rng` on its fresh seed -- the same
+        position those checkpoints already resumed at -- rather than fail to load.
+        """
+        if isinstance(rng_state, dict):
+            self.np_rng.bit_generator.state = rng_state
+            return
+        if self.global_rank == 0:
+            logger.warning("Curriculum learning checkpoint stores the legacy global numpy RNG state, "
+                           "which does not describe this sampler's own stream. Resuming its sampling "
+                           "from the configured seed.")
+
     def state_dict(self):
         return {
             CURRICULUM_LEARNING_BATCH: self.batch,
@@ -321,7 +337,13 @@ class DeepSpeedDataSampler(object):
             CURRICULUM_LEARNING_CURRENT_DIFFICULTIES: self.current_difficulties,
             CURRICULUM_LEARNING_DATA_CLUSTER_PATHS: self.data_cluster_paths,
             CURRICULUM_LEARNING_DATA_CLUSTER_CURRENT_POSITION: self.data_cluster_current_position,
-            CURRICULUM_LEARNING_NP_RNG_STATE: np.random.get_state()
+            # `self.np_rng` is what every draw in this class goes through
+            # (sample_from_clusters, get_new_cluster, reshuffle_clusters), so its
+            # bit generator's state is the one worth saving. `np.random.get_state()`
+            # returns the process-wide legacy RandomState, which this sampler never
+            # touches: it reads the same whether the sampler has drawn nothing or a
+            # thousand batches.
+            CURRICULUM_LEARNING_NP_RNG_STATE: self.np_rng.bit_generator.state
         }
 
     def load_state_dict(self, state_dict):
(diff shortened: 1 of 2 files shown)


alanhuangyoo:
Closing this one. #8489 lists the Data Efficiency library as a deprecation candidate for lack of bandwidth, so a checkpointing fix to its sampler probably isn't worth anyone's review time now. If Data Efficiency ends up staying, I'm happy to reopen. The change is a single file plus its test.

PR #8464, opened 2026-09-08
Fix Muon optimizer under ZeRO CPU offload and bound gather buffers

jinyouzhi:
### Motivation

In PR #8278, we extended the auxiliary optimizer in `MuonWithAuxAdam` to support optimized backends such as `CPUAdam`. However, Muon with CPU offload under ZeRO stages had critical correctness issues and performance/memory pitfalls:
1. **Geometric structure loss under ZeRO CPU offload**: Parameters and gradients flattened into 1D partitions lost matrix geometric properties, bypassing Newton-Schulz polar decomposition and first-order momentum tracking.
2. **Device mismatch & duplicate update under ZeRO-3 offload**: When offload is enabled, the momentum buffer resides in CPU memory while `param.grad` is on accelerator devices. Calling `_apply_distributed_muon_update` during backward caused cross-device runtime errors and double-updating.
3. **Multi-GPU ZeRO-1/2 single-buffer gather nesting**: Reconstructing single-buffer partitions returned nested lists, causing `TypeError: zeros_like()` crashes on the first step.
4. **Buffer allocation spikes & memory fragmentation**: Lacking bounded cache management for all-gather scratch buffers led to unbounded VRAM allocations.

### Changes

- **Shape reconstruction & metadata propagation**:
  - Unflatten 2D/3D logical parameters before Newton-Schulz polar decomposition, slice back to rank partitions, and preserve 1D auxiliary Adam parameters (`CPUAdam`) intact.
- **ZeRO-3 optimizer offload support**:
  - Schedule Muon updates during `_get_norm_groups()` and guard `_apply_distributed_muon_update` during backward when `offload_optimizer` is enabled.
- **Bounded scratch buffers with LRU eviction**:
  - Introduce `_muon_allgather_buffers` LRU cache (capped at 256MB) using `OrderedDict` to prevent VRAM spikes and memory fragmentation.
  - Properly unwrap single-buffer gathered partitions and clear buffers in `destroy()`.

### Tests

- All 23 Muon unit tests pass:
  ```bash
  pytest tests/unit/ops/muon/test_muon.py -k 'not TestMuonConfigs'
  ```
  Including:
  - `TestMuonOptimizerOffload` (ZeRO stages 1, 2, and 3 with CPU offload)
  - `TestMuonAllGatherBufferLifecycle`
  - `TestMuonZero12NumericalCorrectness`
- Verified pre-commit checks pass (`yapf`, `flake8`, `codespell`, `check-torchdist`, `check-license`).


delock:
hi @jinyouzhi whenever this is ready for review just turn it from draft and I'll start review, thanks!

jinyouzhi:
> hi @jinyouzhi whenever this is ready for review just turn it from draft and I'll start review, thanks!

Thank you for your remind! Just rebase with latest master and ready to review.

alanhuangyoo (commented):
Nice, this fixes it. I ran ZeRO-1/2/3 with CPU offload against ZeRO-0 on 2×H20 (same batch on every rank, no clipping) and it matches bit for bit, while master is about 5e-2 off at ZeRO-1/2.

One thing to sort out: #8609 does the universal checkpoint part too, and `_flat_optimizer_states` here skips any state that isn't partition-sized. Without offload, ZeRO-1/2 keep Muon's momentum whole per param, so it gets dropped quietly on conversion (#8609 handles that). Would you be ok leaving the UC changes to #8609 so this one stays about offload? Your offload momentum is partition-sized, so it converts fine there as is.

jinyouzhi:
> Nice, this fixes it. I ran ZeRO-1/2/3 with CPU offload against ZeRO-0 on 2×H20 (same batch on every rank, no clipping) and it matches bit for bit, while master is about 5e-2 off at ZeRO-1/2.
> 
> One thing to sort out: #8609 does the universal checkpoint part too, and `_flat_optimizer_states` here skips any state that isn't partition-sized. Without offload, ZeRO-1/2 keep Muon's momentum whole per param, so it gets dropped quietly on conversion (#8609 handles that). Would you be ok leaving the UC changes to #8609 so this one stays about offload? Your offload momentum is partition-sized, so it converts fine there as is.

Thank you very much for your thorough review. I’m confident that your tests and suggestions will greatly accelerate fixing this issue.

I completely agree with your proposed approach to handling the momentum conversion. This was originally intended as a supplementary improvement while validating the offload fix, so dividing the work this way should allow us both to stay focused on our respective areas.

alanhuangyoo:
Great, thanks! I'll make sure #8609 still handles the offload layout once this lands.

deepspeed/runtime/swap_tensor/optimizer_utils.py
```diff
@@ -207,10 +207,26 @@ def purge_state(self):
...
+            assert len(swap_buffers) == aio_handle.wait()
+        if swap_info.unswapped_gradients:
+            swap_info.write_unswapped_gradients(src_buffer=parameter.grad)
 
     def is_swappable_tensor(self, tensor=None, numel=None):
         assert tensor is not None or numel is not None, "Either tensor or numel must be provided"
         if tensor is not None:
+            # Callers can pin an optimizer state in memory (e.g. the Muon momentum buffer under
+            # save_muon_momentum_buffer_in_memory) by tagging the tensor, which excludes it from swapping.
+            if not getattr(tensor, "swappable", True) or getattr(tensor, "is_resident", False):
```
delock: what is the scenario when `swappable` is used and what is the scenario for `is_resident`?

jinyouzhi: Good question. `is_resident` is redundant, I’ll remove it and keep `swappable=False`.

jinyouzhi: done @delock

deepspeed/runtime/swap_tensor/optimizer_utils.py
```diff
@@ -207,10 +207,25 @@ def purge_state(self):
...
+    def _writeback_gradients(self, swap_info, parameter, aio_handle):
+        """Persist a parameter's gradient partitions, whichever way they are stored."""
+        if swap_info.swapped_gradients:
+            param_gradients = swap_info.swapped_gradients.values()
+            swap_buffers = [parameter.grad.narrow(0, grad.offset, grad.length) for grad in param_gradients]
+            swap_paths = [grad.path for grad in param_gradients]
+            swap_out_tensors(aio_handle, swap_buffers, swap_paths)
+            assert len(swap_buffers) == aio_handle.wait()
+        if swap_info.unswapped_gradients:
+            swap_info.write_unswapped_gradients(src_buffer=parameter.grad)
```
delock, line 223: do we need to call swap_info.release_unswapped_gradients() after this line?

jinyouzhi: No, not at this intermediate writeback point. These small gradient fragments have no NVMe backing file: `write_unswapped_gradients()` updates their CPU copies with the Muon-transformed gradients, and `_prepare_sub_group()` subsequently swaps the subgroup in again for the actual optimizer step. Clearing them here would lose those updates; when all fragments are unswapped it can also leave the next swap-in without a gradient buffer.

They are already released by the normal post-step `swap_out_optimizer_state()` in both `PartitionedOptimizerSwapper` and `PipelinedOptimizerSwapper`.

I have added a lifecycle comment and strengthened the all-unswapped-fragments regression locally: two training steps are compared with an independent full-gradient Muon reference for both swappers. Both cases pass on an RTX 5090 D using real NVMe I/O. These follow-up changes are not pushed yet.

jinyouzhi: No, not at this intermediate writeback point. These small gradient fragments have no NVMe backing file: `write_unswapped_gradients()` updates their CPU copies with the Muon-transformed gradients, and `_prepare_sub_group()` subsequently swaps the subgroup in again for the actual optimizer step. Clearing them here would lose those updates; when all fragments are unswapped it can also leave the next swap-in without a gradient buffer.

They are already released by the normal post-step `swap_out_optimizer_state()` in both `PartitionedOptimizerSwapper` and `PipelinedOptimizerSwapper`.

I have added a lifecycle comment and strengthened the all-unswapped-fragments regression locally: two training steps are compared with an independent full-gradient Muon reference for both swappers. Both cases pass on an RTX 5090 D using real NVMe I/O.

delock (approved):

PR #8467, opened 2026-09-09
A zero-element parameter loses its shape when it is bound to a flat buffer

alanhuangyoo:
A zero-element trainable parameter lost its shape when a wrapper bound it to a flat buffer. After that, the module's own forward failed. `nn.Linear(8, 0, bias=False)` in a model, `deepspeed.initialize`, one step:

```
RuntimeError: size mismatch, got input (1), mat (1x8), vec (0)
```

torch's `unflatten_dense_tensors` special-cases `numel == 0` and returns a fresh 1-D `zeros({0})` rather than a view of the requested shape, so `p.data = q.data` turned `(0, 8)` into `(0,)`. This happened with `FP16_Optimizer` (fp16 or bf16 at stage 0) and with `BF16_Optimizer` (bf16 at stage 1 with fp32 grad accumulation).

## The change

As discussed with @sfc-gh-truwase, zero-element parameters are now left out of the optimizer groups, the same way frozen parameters are. They have nothing to optimize.

- `is_optimized_parameter(param)` in `runtime/utils.py`: `requires_grad and numel > 0`, using `ds_numel` under ZeRO-3.
- Every wrapper uses it to build its groups: `FP16_Optimizer`, `FP16_UnfusedOptimizer`, `BF16_Optimizer`, ZeRO-1/2 (`stage_1_and_2.py`) and ZeRO-3 (`_get_trainable_parameter_groups`).
- The checkpoint side uses the same predicate: `_get_zero_frozen_param_attributes` and the frozen-fragment load in `load_module_state_dict`. `param_shapes` is built from the optimizer groups, so with the old `requires_grad` rule a filtered zero-element parameter would be in neither `param_shapes` nor `frozen_param_shapes`, and `zero_to_fp32` would drop it. It is now recorded with the frozen parameters and rebuilt with its shape.

Grad hooks and the other bookkeeping in ZeRO-1/2 and ZeRO-3 iterate the wrappers' own groups, so they follow the filter without further changes.

## Testing

On 2×H20, `tests/unit/runtime/zero/test_zero_numel_param_shape.py` (17 passed):
- `TestZeroNumelParameterShape`: the shape survives init, one step and a second step for fp16 stage 0, bf16 stage 0, bf16 stage 1 with fp32 accumulation, ZeRO-1 and ZeRO-2.
- `TestZeroNumelParameterCheckpoint`: ZeRO-2 and ZeRO-3, save → `get_fp32_state_dict_from_zero_checkpoint` → strict `load_state_dict`, with the zero-element parameter at `(0, 8)`. With the optimizer filter but the old frozen rule in `engine.py`, both fail with `KeyError: 'extra'`.

`tests/unit/checkpoint/test_zero_optimizer.py -k frozen`: 13 passed.

### DIFF
--- a/deepspeed/runtime/bf16_optimizer.py
+++ b/deepspeed/runtime/bf16_optimizer.py
@@ -12,9 +12,10 @@ from deepspeed.runtime.constants import PIPE_REPLICATED
 from deepspeed.runtime.base_optimizer import ZeROOptimizer
 from packaging import version as pkg_version
 from deepspeed.git_version_info import version
-from deepspeed.runtime.utils import (get_global_norm_of_tensors, clip_tensors_by_global_norm, DummyOptim,
-                                     align_dense_tensors, all_gather_dp_groups, is_model_parallel_parameter,
-                                     see_memory_usage, graph_process, get_norm_with_moe_layers)
+from deepspeed.runtime.utils import (bind_flat_views, get_global_norm_of_tensors, clip_tensors_by_global_norm,
+                                     DummyOptim, align_dense_tensors, all_gather_dp_groups,
+                                     is_model_parallel_parameter, see_memory_usage, graph_process,
+                                     get_norm_with_moe_layers)
 from deepspeed.utils import link_hp_params, lazy_init_hp_params_optimizer_state, fragment_address, groups
 from deepspeed.moe.utils import is_moe_param, is_moe_param_group
 from deepspeed.utils.bwc import bwc_tensor_model_parallel_rank
@@ -293,8 +294,7 @@ class BF16_Optimizer(ZeROOptimizer):
 
     def _update_storage_to_flattened_tensor(self, tensor_list, flat_tensor):
         updated_params = self.unflatten(flat_tensor, tensor_list)
-        for p, q in zip(tensor_list, updated_params):
-            p.data = q.data
+        bind_flat_views(tensor_list, updated_params)
 
     def _flatten_dense_tensors_aligned(self, tensor_list, alignment):
         return self.flatten(align_dense_tensors(tensor_list, alignment))
--- a/deepspeed/runtime/fp16/fused_optimizer.py
+++ b/deepspeed/runtime/fp16/fused_optimizer.py
@@ -10,7 +10,8 @@ This file is adapted from FP16_Optimizer in NVIDIA/apex
 import torch
 from torch._utils import _flatten_dense_tensors, _unflatten_dense_tensors
 from deepspeed.runtime.base_optimizer import DeepSpeedOptimizer
-from deepspeed.runtime.utils import get_global_norm, get_flattened_grad_norm, CheckOverflow, get_weight_norm, get_norm_with_moe_layers, is_model_parallel_parameter
+from deepspeed.runtime.utils import (bind_flat_views, get_global_norm, get_flattened_grad_norm, CheckOverflow,
+                                     get_weight_norm, get_norm_with_moe_layers, is_model_parallel_parameter)
 from deepspeed.runtime.fp16.loss_scaler import LossScaleConfig, LossScaleProfile
 from deepspeed.utils import logger, log_dist
 from deepspeed.utils.torch import required_torch_version
@@ -92,8 +93,7 @@ class FP16_Optimizer(DeepSpeedOptimizer):
             self.fp16_groups_flat.append(_flatten_dense_tensors([p.clone().detach() for p in self.fp16_groups[i]]))
             # set model fp16 weight to slices of flattened buffer
             updated_params = _unflatten_dense_tensors(self.fp16_groups_flat[i], self.fp16_groups[i])
-            for p, q in zip(self.fp16_groups[i], updated_params):
-                p.data = q.data
+            bind_flat_views(self.fp16_groups[i], updated_params)
             # init master weight, flattened
             self.fp32_groups_flat.append(self.fp16_groups_flat[i].clone().float().detach())
             # modify optimizer of have flat master weight
@@ -187,8 +187,7 @@ class FP16_Optimizer(DeepSpeedOptimizer):
         # TODO: we probably don't need this? just to be safe
         for i in range(len(norm_groups)):
             updated_params = _unflatten_dense_tensors(self.fp16_groups_flat[i], self.fp16_groups[i])
-            for p, q in zip(self.fp16_groups[i], updated_params):
-                p.data = q.data
+            bind_flat_views(self.fp16_groups[i], updated_params)
         return self.overflow
 
     def set_lr(self, lr):
@@ -354,6 +353,11 @@ class FP16_Optimizer(DeepSpeedOptimizer):
         for i in range(len(self.fp16_groups)):
             updated_params = _unflatten_dense_tensors(self.fp32_groups_flat[i], self.fp16_groups[i])
             for p, q in zip(self.fp16_groups[i], updated_params):
+                if p.numel() == 0:
+                    # See bind_flat_views: `q` is a 1-D zeros({0}) here, not a view of
+                    # `p`'s shape, so this copy would raise on the shape mismatch. There
+                    # are no elements to copy either way.
+                    continue
                 p.data.copy_(q.data)
         self.has_executed_step = True
         if self.timers:
--- a/deepspeed/runtime/utils.py
+++ b/deepspeed/runtime/utils.py
@@ -823,6 +823,21 @@ def empty_cache():
     get_accelerator().reset_peak_memory_stats()
 
 
+def bind_flat_views(tensors, views):
+    """Point each tensor at its view of a flat buffer, skipping zero-element ones.
+
+    torch's ``unflatten_dense_tensors`` special-cases ``numel == 0`` and returns a
+    freshly allocated 1-D ``zeros({0})`` rather than a view of the requested shape,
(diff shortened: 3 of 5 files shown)


alanhuangyoo:
Scope widened after opening this: ZeRO-1/2 was not the only place. `FP16_Optimizer` and `BF16_Optimizer` bind parameters to a flat buffer with the same two lines, and drop the same shape — so the failure does not need ZeRO at all. Any run with fp16 or bf16 enabled and a zero-element parameter breaks on the first forward after `initialize`.

Measured on an H20 against unmodified master (`cbd303e`):

```
fp16 + stage 0               FP16_Optimizer   shape after init (0,)   step: RuntimeError
bf16 + stage 0               FP16_Optimizer   shape after init (0,)   step: RuntimeError
bf16 + stage 1 + fp32 accum  BF16_Optimizer   shape after init (0,)   step: RuntimeError
bf16 + ZeRO-1                                 shape after init (0,)   step: RuntimeError
bf16 + ZeRO-2                                 shape after init (0,)   step: RuntimeError
```

`09d6f71` moves the guard into `bind_flat_views` in `runtime/utils.py` so the reason is written once, and calls it from all four binding sites. `FP16_Optimizer.step` has a fifth site that copies rather than rebinds; without a guard there that copy would *start* raising on the `(0, 8)` vs `(0,)` mismatch once the earlier sites stop corrupting the shape, so it takes the same skip inline — worth a look, since it is the one place the fix could have introduced a new failure rather than removed one.

Test is now parametrized over those five configurations: 15 passed here, 15 failed on master. Regression on the same box with `DS_SKIP_CUDA_CHECK=1` (`test_stage2_flatten_on_gpu`, `test_zero_tensor_fragment`, `test_zero_coalesce_grad_reduction`): 174 passed, 75 skipped, 0 failed.

ebarkhordar (commented):
I ran your tests at 990c49cd and at the merge base b5e000c4c in a clean container (torch 2.8.0+cpu, `deepspeed.__file__` printed each run). CPU, so this is the four non-fp16 cases.

Base b5e000c4c: 6 failed, 6 passed. `bf16_stage0` and `bf16_stage1_fp32_accum` fail all three, with the signature from your description:

```
RuntimeError: size mismatch, got input (1), mat (1x8), vec (0)
```

`zero1` and `zero2` pass at the base, so the #8277 attribution holds up. Head 990c49cd: 12 passed. The parameter `bind_flat_views` skips also matches the bound one, `empty.weight` (0, 8) bfloat16 cpu beside `dense.weight` (8, 8) bfloat16 cpu, so skipping does not strand it on another dtype or device.

On "every wrapper that flattens the parameters": I parsed the package for unflatten call sites rather than grepping. 16, of which the ones that bind the result back are your three, `stage_1_and_2.py:857` (already `narrow().view()` from #8277), and four in `stage3.py`. Those four bind `fp16_partitioned_groups`, which is `[param.ds_tensor ...]` at `stage3.py:899` and 1-D, so there is no shape there to lose. The set looks closed.

One question. Stage 3 is the wrapper missing from your table, and `EmptyTailModel` under it fails at BOTH refs, identically:

```
AssertionError: {'id': 1, 'status': 'NOT_AVAILABLE', 'numel': 0,
 'ds_numel': 0, 'shape': (0,), 'ds_shape': (0, 8), 'persist': True}
```

`ds_shape` is intact, so that is a different bug and your diff does not touch `stage3.py`. Worth a line saying stage 3 is out of scope?

No GPU here, so fp16 and anything multi-rank is untested by me.

ebarkhordar:
Answering my own question, since I should have checked before asking it: #8375 is already open for the stage 3 case, with the same assertion text, and its description says the ZeRO-1/2 shape loss is fixed separately. So the scope split here is deliberate and nothing needs adding to this body. The measured half of my comment stands.

sfc-gh-truwase:
@alanhuangyoo thanks for the PR. What are zero-element trainable parameters used for? Can they be filtered out during optimizer construction similar to frozen parameters?
https://github.com/deepspeedai/DeepSpeed/blob/7075430aa754eb97ecede4cfb2a14ad497d9ab40/deepspeed/runtime/zero/stage_1_and_2.py#L396-L400

alanhuangyoo:
@sfc-gh-truwase Two honest answers.

**Where they come from.** I found this by auditing the flatten/bind path after #8277, not from a user report. The sources I know of are config-driven: a block whose width is a config value that can be zero (register/extra tokens, an optional projection or task head), and shard-level empties, where a small dimension split across tensor- or expert-parallel ranks leaves some ranks with nothing. In every case the module is still in the graph and its forward still runs, which is why the shape matters — `F.linear` dispatches to `addmv` once the weight comes back 1-D.

**On filtering them out.** It would fix this failure too, and for frozen parameters it is clearly right. Two things make me prefer the narrow fix here:

- ZeRO-1/2 already made the other choice. #8277 kept zero-element parameters in the flat group and fixed the binding with `narrow().view(meta.shape)`, and `stage_1_and_2.py` carries explicit `numel() == 0` handling (`:1261`, `:1326`). If the other three wrappers filter instead, the same model has a parameter in optimizer state at stage 1/2 and not at stage 0, which is a harder rule to explain than "every wrapper keeps the shape".
- Filtering changes the flat buffer. Dropping a parameter from `fp16_groups` / `bf16_groups` shifts the offsets of everything after it, so checkpoints written before and after the change no longer line up, and `param_idx_in_group` consumers (ZenFlow reads `bit16_groups[i][param_idx_in_group]`) shift with it. That is a migration, not a 3-line fix.

The current change is three lines per wrapper through one shared helper, and it leaves what appears in optimizer state and checkpoints exactly as it is today.

If you would still rather filter, I am happy to do it as a follow-up covering all four wrappers plus the checkpoint mapping — but I would not want this PR to be the place where stage 0 and stage 1/2 start disagreeing.

sfc-gh-truwase:
@alanhuangyoo thanks for the detailed explanation. It is very helpful.  

- While the about checkpoint BC is valid and appreciated, it is unlikely such checkpoints exist since they would have already run into this issue during training/creation. 
- ZeRO-1/2 solution shouldn't be a constraint if a more optimal choice later emerges. I think the right approach is to globally apply the new solution. 
- The impact of changing offsets on consumers like `ZenFlow` is worth deeper analysis. My intuition is that this shouldn't be problem since frozen params, which are similar, are have been filtered out for a long time without issues.  
-  It seems counter-intuitive for zero-sized parameters, which are thus non-trainable, to propagate into optimizer state. I am concerned this confusion will only magnify as new codes build on it. 

I prefer this PR applies the filtering solution globally. I am very curious about the checkpointing mapping analysis and recommended solution.

alanhuangyoo:
@sfc-gh-truwase Makes sense, I'll switch this PR to filtering. Here's the checkpoint analysis first, since that's where filtering causes trouble:

- `_get_zero_param_shapes()` is built from the optimizer's bit16/fp16/bf16 groups, and `_get_zero_frozen_param_attributes()` is built from `not param.requires_grad`. A zero-element trainable parameter filtered out of the groups would land in neither. `zero_to_fp32` would then rebuild a state dict without it, and `load_state_dict(strict=True)` on that state dict would fail.
- My recommendation is a single predicate for "the optimizer owns this parameter": `requires_grad and numel > 0` (`ds_numel` under ZeRO-3). Every wrapper would use it when building its groups, and `_get_zero_frozen_param_attributes` would use it too, so zero-element parameters get recorded as frozen (their shape plus an empty fragment). `zero_to_fp32` and checkpoint loading already handle frozen parameters, so both would rebuild the parameter with its shape without any change to the converters.
- The places in ZeRO-1/2 and ZeRO-3 that check `requires_grad` separately (gradient hooks, `param_id` maps, the norm) need the same predicate. Otherwise a hook fires for a parameter that has no slot. I'll change those too.
- On old checkpoints, agreed, they can't contain such parameters.

For tests: the shape survives init and step for stage 0 with fp16 and bf16 and for stages 1, 2 and 3, plus a save → `zero_to_fp32` → strict load round trip.

alanhuangyoo:
@sfc-gh-truwase Switched to filtering in f2d7d91 and updated the description.

`is_optimized_parameter` (`requires_grad and numel > 0`) now decides group membership in all five wrappers, and it also decides what gets recorded as frozen in checkpoints. Without that second part, `zero_to_fp32` drops the parameter: the new checkpoint test fails with `KeyError: 'extra'` if only the optimizer side is changed. 17 new tests plus the frozen checkpoint tests pass on 2×H20.

sfc-gh-truwase (approved):

alanhuangyoo:
@sfc-gh-truwase Thanks. The red CPU job was my test, not the change: the fp16 stage 0 cases raised "Type fp16 is not supported" on the CPU runner. They skip on unsupported dtypes now (f2a23ba5a). The push probably turned auto-merge off, sorry about that.

alanhuangyoo:
@sfc-gh-truwase Merged master to clear the conflict (9cb907b). It was import lines only: master dropped the PLD, eigenvalue and LoCo imports, and this branch adds `is_optimized_parameter`. The 17 tests pass again on 2×H20. Could you re-trigger CI and merge when it's green?

alanhuangyoo:
@sfc-gh-truwase It's been green since the master merge on 9/19. Is anything else needed before it can go in?

sfc-gh-truwase:
@alanhuangyoo apologies for the delay. CI is running.

PR #8474, opened 2026-09-10
Fix universal checkpoint resume across AutoTP sizes

0z5a:
Universal checkpoints saved by AutoTP can fail or restore corrupted FP32 parameter mappings when training resumes. TP2 -> TP1 enters the Megatron-specific model-state merger even though universal weights are restored separately from `zero/`. TP2 -> TP2 can overwrite the FP32 parameter mapping while flattening Adam moments. This fixes metadata-rank selection for universal loading and updates each optimizer-state mapping without replacing the FP32 parameter mapping.

This carries forward my [original fix and integration tests](https://github.com/0z5a/DeepSpeed/commit/e753a02b72dd18bc220f31ecac5ce5f6c1a89c54), which were cherry-picked as `dd5c57f` into #8385 after its merge-queue snapshot. The merged squash `1190946` contains the affine IR/converter work but omits these three files. The follow-up applies only the two loader fixes and the existing resume tests on that merged base.

The integration test trains a real column/row AutoTP FP32 model with ZeRO-1 and torch Adam for four steps at TP2, saves and converts its checkpoint through legacy or affine metadata, then resumes four steps at TP1 or TP2. It compares logits, loss, gradients, FP32 parameters, both Adam moments, and step counts against uninterrupted training. The affine maps are injected by the test; production metadata emission remains separate work under #8252. Native row-parallel bias uses scale 1 here; this does not enable general non-unit-scale optimizer resume.

Validation on the merged base plus this patch, using Apple M5 CPU/Gloo, Python 3.12.13 and torch 2.14.0:

- All four resume cases fail on unmodified `1190946`: TP1 hits Megatron key validation; TP2 has incorrect FP32 mappings. All four pass with this patch.
- 71 affine/AutoTP unit regressions pass.
- Four existing tensor-fragment and universal DP-resize regressions pass.
- Modified-file pre-commit checks pass.

The same original patch also previously passed all four resume cases on two NVIDIA RTX 4000 Ada GPUs (20 GB each), Python 3.12.14, torch 2.13.0+cu130 and NCCL, as recorded in [the original validation](https://github.com/deepspeedai/DeepSpeed/pull/8385#issuecomment-5598904008). That GPU run used the pre-merge base; the fresh follow-up validation above is CPU/Gloo.

```bash
pytest -sv tests/unit/checkpoint/test_autotp_uc_checkpoint.py::TestAffineUniversalCheckpointResume
```

Related: #8252, #8230. Follow-up to #8385.

### DIFF
--- a/deepspeed/runtime/engine.py
+++ b/deepspeed/runtime/engine.py
@@ -4779,8 +4779,15 @@ class DeepSpeedEngine(Module):
 
         is_pipe_parallel = isinstance(self.module, PipelineModule)
 
-        load_path, checkpoint, _ = sd_loader.load(self.mp_world_size,
-                                                  self.checkpoint_mp_rank,
+        checkpoint_mp_world_size = self.mp_world_size
+        checkpoint_mp_rank = self.checkpoint_mp_rank
+        if self.load_universal_checkpoint():
+            # UC restores weights from zero/. Read metadata from the corresponding
+            # source rank without invoking Megatron's model-specific weight merger.
+            checkpoint_mp_rank = checkpoint_mp_rank * len(ckpt_list) // checkpoint_mp_world_size
+            checkpoint_mp_world_size = len(ckpt_list)
+        load_path, checkpoint, _ = sd_loader.load(checkpoint_mp_world_size,
+                                                  checkpoint_mp_rank,
                                                   is_pipe_parallel=is_pipe_parallel)
 
         if checkpoint is None:
--- a/deepspeed/utils/tensor_fragment.py
+++ b/deepspeed/utils/tensor_fragment.py
@@ -79,7 +79,7 @@ def map_to_flat_opt_states(flat_hp_tensor, lp_tensors, optim_state, opt_keys):
                 hp_fragment_address = lp._hp_mapping.get_hp_fragment_address()
                 hp_fragment = buffer.narrow(0, hp_fragment_address.start, hp_fragment_address.numel)
                 hp_fragment.data.copy_(lp._hp_mapping.get_hp_fragment(optim_state_key=key).data)
-                lp._hp_mapping.hp_fragment = hp_fragment
+                lp._hp_mapping.optim_fragment[key] = hp_fragment
 
         optim_state[hp_param][key] = buffer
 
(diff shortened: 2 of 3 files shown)


0z5a:
@delock Could you review this follow-up? It restores the two AutoTP universal-checkpoint loader fixes and the four save-convert-resume integration cases omitted from the #8385 merge-queue snapshot. On the merged base, all four cases fail before the patch and pass after it; 75 related regressions also pass. The scope remains native scale-1 training, with production affine metadata emission left to the separate step-4 work.

Achyuthan-S:
Reviewed, and reproduced both halves of your claim independently on the merged base (`1190946`), CPU/gloo with `DS_ACCELERATOR=cpu LOCAL_SIZE=4`.

**Applying only your test file, with the production changes withheld: all four cases fail.** The reasons split cleanly along the two bugs, which is a nice confirmation that the fixes are independent rather than one masking the other:

| case | failure without the fix |
|---|---|
| `tp1-legacy`, `tp1-affine` | `key: attention.dense.weight is not found in the checkpoint` |
| `tp2-legacy`, `tp2-affine` | `snapshot 0, fc1.weight/fp32: Tensor-likes are not close!` |

The TP1 pair hits the Megatron merger, the TP2 pair hits the fp32 mapping. **Adding both production changes: 4 passed** — 43 with the affine suite alongside. I checked the file contents between stages to be sure the production files really were untouched in the first run, since that is the part of the experiment that is easy to get wrong.

Worth spelling out why the `tensor_fragment.py` line matters, for whoever merges this. The loop runs once per optimizer key, and `get_hp_fragment(optim_state_key=key)` returns `optim_fragment[key]`. Assigning to `hp_fragment` overwrote the FP32 master-weight pointer with a moment buffer slice, so after the loop every `update_lp()` / `update_hp()` read and wrote the wrong tensor. One line, and it corrupts a resume with nothing to signal it — which is exactly the `Tensor-likes are not close` signature above.

The `engine.py` change is scoped behind `load_universal_checkpoint()`, so the non-UC path is untouched.

On the Codex comment: `Signed-off-by` is present on `7be3255` and the DCO check passes — it looks like the diff is being read without the commit trailers. Same false positive it produced on #8385 and #8477.

LGTM.

@delock this fixes a live bug on master rather than adding anything: universal-checkpoint resume across TP sizes currently fails in all four combinations, so it may be worth taking ahead of #8477.

delock (approved):

delock:
Thanks @0z5a for this fix!

PR #8477, opened 2026-09-10
Carry the affine scale on the replicated map, not the split

Achyuthan-S:
Follow-up to #8385. The scale move answering @delock's question was pushed after the merge queue had already snapshotted the branch, so it did not land with the rest.

The layouts that pre-divide a value hold it whole on every rank: Yuan's o_proj and the last conv layer both replicate the bias divided by the world size, so the all-reduced sum adds it exactly once. The weight beside them is what gets split, and it is unscaled — so no in-tree layout scales a split, and the split constructors no longer take the argument.

Adds a test covering the replicated case.

### DIFF
--- a/deepspeed/checkpoint/affine.py
+++ b/deepspeed/checkpoint/affine.py
@@ -356,11 +356,16 @@ def _row_major_strides(shape):
     return tuple(strides)
 
 
-def replicated_map(shape, tp_degree):
+def replicated_map(shape, tp_degree, scale=1.0):
     """Every rank holds the whole parameter.
 
     One piece, named by every rank, which is what lets a converter read it from whichever
     rank is cheapest rather than from a designated owner.
+
+    ``scale`` belongs here rather than on a split, because the layouts that pre-divide a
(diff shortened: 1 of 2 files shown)


delock (approved):

delock:
@Achyuthan-S can you fix formatting error?  Thanks!

Achyuthan-S:
@delock I don't think there's a formatting issue to fix — yapf crashed on the runner rather than reporting a diff:

```
from yapf_third_party._ylib2to3 import pygram
  ...pgen2/grammar.py
EOFError: Ran out of input
```

That's yapf's lib2to3 grammar cache being truncated in the pre-commit env. There's no `would reformat` line anywhere in the run, and every other hook passed.

I re-checked locally with the pinned v0.40.0 against the branch including your merge of master — no diff on either file, and those are the only two `.py` files this PR touches. The PR's own formatting check is green too.

Happy to push an empty commit if that's easier than a re-queue, but I don't believe there's anything to correct.

Issue #8469, opened 2026-09-10
[Bug] AutoEP FP32 gradient clipping averages norms across ranks with different expert ownership


gss10282023:
### Summary

DeepSpeed's FP32 gradient-clipping path computes the wrong global norm when AutoEP places different experts on different ranks. It averages rank-local norms as if every rank held the same parameters, but under AutoEP the ranks own different expert parameters.

Redistributing the same unique expert gradients across ranks can therefore change the clipping coefficient and the resulting update when clipping is active.

### System info

**Environment:** DeepSpeed commit [cf443004](https://github.com/deepspeedai/DeepSpeed/blob/cf44300453eb0af79ed84ed8f1cb49d57478bd76/deepspeed/runtime/utils.py), two CUDA ranks, FP32 gradients, ZeRO-0, L2 clipping, `mpu=None`.

The results below are from the pinned version. Source inspection on 2026-09-10 found the same calculation in [0166cc87](https://github.com/deepspeedai/DeepSpeed/blob/0166cc878b106546adc78214da952f1aa6f5d02b/deepspeed/runtime/utils.py); the full experiment has not been rerun on that commit.

Separate current-branch CPU validation: Ubuntu 26.04 LTS (x86_64), Python 3.11.15, PyTorch 2.13.0+cpu, Intel Xeon Cascadelake, two local CPU/Gloo processes, source baseline `0166cc87` and PR #8470. No GPU or inter-node connection is used in that CPU run.

### To Reproduce

The test enters through `DeepSpeedEngine.clip_fp32_gradients()` and represents the same four unique expert gradients in two layouts:

- **EP size 1:** all four experts replicated on both ranks.
- **EP size 2:** each rank owns two of the four experts.

The unique gradient vector and the clipping threshold are fixed. The expected coefficient comes directly from the L2 norm of the unique vector, counting each expert once.

| Comparison | Absolute difference in clipping coefficient |
| --- | ---: |
| Replicated vs. owner-sharded, current code | `0.2000005` |
| Replicated vs. owner-sharded, with repair | `0` |
| Fully replicated control | `0` |

The replicated layout matches the direct reference; the owner-sharded layout does not. The discrepancy is identical across three repeated blocks, and the repair was confirmed on a second host.

[...]

PR #8478, opened 2026-09-10
Count each expert once in the fp32 gradient-clipping norm

### LINKED ISSUE #8475: [Bug] AutoEP FP32 gradient clipping averages norms across ranks with different expert ownership
### Summary

DeepSpeed's FP32 gradient-clipping path computes the wrong global norm when AutoEP places different experts on different ranks. It averages rank-local norms as if every rank held the same parameters, but under AutoEP the ranks own different expert parameters.

Redistributing the same unique expert gradients across ranks can therefore change the clipping coefficient and the resulting update when clipping is active.

### System info

**Environment:** DeepSpeed commit [cf443004](https://github.com/deepspeedai/DeepSpeed/blob/cf44300453eb0af79ed84ed8f1cb49d57478bd76/deepspeed/runtime/utils.py), two CUDA ranks, FP32 gradients, ZeRO-0, L2 clipping, `mpu=None`.

The results below are from the pinned version. Source inspection on 2026-09-10 found the same calculation in [0166cc87](https://github.com/deepspeedai/DeepSpeed/blob/0166cc878b106546adc78214da952f1aa6f5d02b/deepspeed/runtime/utils.py); the full experiment has not been rerun on that commit.

Separate current-branch CPU validation: Ubuntu 26.04 LTS (x86_64), Python 3.11.15, PyTorch 2.13.0+cpu, Intel Xeon Cascadelake, two local CPU/Gloo processes, source baseline `0166cc87` and PR #8476. No GPU or inter-node connection is used in that CPU run.

### To Reproduce

The test enters through `DeepSpeedEngine.clip_fp32_gradients()` and represents the same four unique expert gradients in two layouts:

- **EP size 1:** all four experts replicated on both ranks.
- **EP size 2:** each rank owns two of the four experts.

The unique gradient vector and the clipping threshold are fixed. The expected coefficient comes directly from the L2 norm of the unique vector, counting each expert once.

| Comparison | Absolute difference in clipping coefficient |
| --- | ---: |
| Replicated vs. owner-sharded, current code | `0.2000005` |
| Replicated vs. owner-sharded, with repair | `0` |
| Fully replicated control | `0` |

The replicated layout matches the direct reference; the owner-sharded layout does not. The discrepancy is identical across three repeated blocks, and the repair was confirmed on a second host.

The regression code is in [PR #8476](https://github.com/deepspeedai/DeepSpeed/blob/8288645c1cef4c03bb65fb316fb6473bcb13f02f/tests/unit/v1/moe/test_autoep_grad_parity.py). `TestAutoEPFP32Clipping` exercises the native engine against an independent unique-gradient reference. [PR #8476](https://github.com/deepspeedai/DeepSpeed/pull/8476) includes the exact two-process CPU/Gloo command; all eight cases passed with the repair, and the owner-sharded case fails on unmodified `0166cc87`.

### Expected behavior

Replicated and owner-sharded layouts of the same unique gradients should give the same global L2 norm, clipping coefficient, and optimizer update. Dense parameters should count once despite replication, and each expert should count once according to its ownership.

### Root cause

`clip_grad_norm_` takes a local square root before averaging across the data-parallel group. When ranks own different experts, the average of local norms does not reconstruct the norm of the concatenated unique gradients.

Replication counts differ as well: dense parameters live on every DP rank, while AutoEP expert parameters live on only `dp_world_size / ep_size` ranks. Their contributions need different weights for each logical parameter to count exactly once.

### Proposed fix

Record the EP size on each AutoEP expert parameter. For the FP32 L2 path:

1. Sum squared gradients locally, dividing each parameter's contribution by its replica count.
2. Sum those contributions across the DP group.
3. Take a single square root after the reduction and compute the clipping coefficient.

The tested patch adds the ownership metadata in `auto_ep_layer.py` and implements this reduction in `runtime/utils.py`. Its validated scope is the ZeRO-0, `mpu=None`, L2 case above.

### ds_report output

Collected with `DS_ACCELERATOR=cpu python -m deepspeed.env_report` from the current PR checkout. This describes the separate CPU validation environment; it is not the pinned two-GPU environment. Local installation paths have been replaced with placeholders.

<details>
<summary>CPU environment report</summary>

```text
deepspeed_not_implemented  [NO] ....... [OKAY]
deepspeed_ccl_comm ..... [NO] ....... [OKAY]
deepspeed_shm_comm ..... [NO] ....... [OKAY]
pin_memory ............. [NO] ....... [OKAY]
torch install path ............... ['<local dependency path>']
torch version .................... 2.13.0+cpu
deepspeed install path ........... ['<DeepSpeed checkout>/deepspeed']
deepspeed info ................... 0.19.6, [none], [none]
deepspeed wheel compiled w. ...... torch 0.0 
shared memory (/dev/shm) size .... 30.27 GB
```

</details>

### Launcher context

The current regression uses `python -m torch.distributed.run --standalone --nproc-per-node=2` with Gloo. The exact driver is included in PR #8476.

### Docker context

The current CPU regression ran in an existing Ubuntu environment; no separately specified Docker image is required by the reproducer. The pinned GPU report does not specify an image tag.

alanhuangyoo:
Fixes #8475.

(Originally opened against #8469, which @gss10282023 withdrew and resubmitted as #8475 while this was being written. The discussion about how this and #8476 relate is on #8469 and continues on #8475.)

## The measurement

`clip_grad_norm_` ends by averaging the per-rank norms over the data parallel group:

```python
pg = groups._get_data_parallel_group()
scaled_norm = total_norm * 1.0 / float(dist.get_world_size(group=pg))
dist.all_reduce(scaled_norm_tensor, group=pg)
```

That reconstructs a global norm only while every rank holds the same parameters. Expert parameters are not replicated, so each rank's norm describes a different set of experts. Four expert gradients across two ranks, measured on 2×H20:

| layout | global norm |
| --- | ---: |
| reference — each expert counted once | **10.954452** (`sqrt(120)`) |
| replicated on both ranks | 10.954452 |
| one half per rank | **7.236068** |

`sqrt(20)` and `sqrt(100)` averaged to 7.236 instead of `sqrt(20 + 100)`. The same gradients, 34% apart, decided only by where the experts happen to live. `clip_coef` is `max_norm / total_norm`, so wherever clipping is active this changes the update.

@gss10282023 reported this (#8475) and measured a 0.2 shift in the clipping coefficient on their setup; the numbers above are an independent reproduction from a different direction.

## Why the fix is the repo's own

`get_norm_with_moe_layers` already does it correctly — it takes the non-expert norm plus the expert gradients keyed by group, reduces each expert group over its own `moe_ep_group`, and combines as `(Σ norm_i ** p) ** (1/p)`. **Both the bf16 and the fp16 optimizer already call it.**

The fp32 path instead inlines the approximation. That approximation also exists under its own name, `get_norm_with_moe_layers_fast`, carrying this comment:

> This implementation standardizes the grad_norm across ranks. A more precise implementation can be found in 'get_norm_with_moe_layers'.

and that named copy has **no callers at all** — so the warning never reaches the one place still doing it. This PR keeps expert gradients out of the replicated norm and folds them in through the precise helper, leaving fp32 the same shape as the other two optimizers.

## Scope

**Unchanged when no expert parameter is present** — that branch is the original code, not touched. Every non-MoE fp32 run behaves exactly as before, which the existing suite confirms (below).

Two guards worth naming:

- An expert parameter carrying no `group_name` has no expert parallel group to reduce over. `TestClipGradNorm::test_gather` builds one by hand exactly like that, so such a call stays on the original path rather than guessing a group.
- `get_norm_with_moe_layers` reports a non-finite norm as `-1`. Used as-is that would make `clip_coef` negative and flip the sign of every gradient, so it is mapped to `inf` — which keeps what the averaging path already did, scale to zero.

## Testing

2×H20. New case `test_sharded_experts_count_once` asserts the **norm value**, not that a code path was reached:

```
branch                        10 passed          (9 existing + 1 new)
master + this branch's test    1 failed
    AssertionError: global norm 7.236067771911621 should be 10.954451560974121
                    regardless of expert placement
```

It fails on `master` for the defect itself, not because it is new.

`tests/unit/runtime/test_runtime_utils.py` + `tests/unit/v1/moe/test_moe.py`, comparing failure *sets*:

```
master   7 failed, 29 passed, 14 skipped
branch   6 failed, 30 passed, 14 skipped

only failing on master               -> test_sharded_experts_count_once   (the new one)
failing on both                      -> 6
```

The 6 are pre-existing on this box — `ImportError: libcudart.so.13`, unrelated to this change.

## Note

#8470 proposed a fix for the same issue and was closed by its author without discussion. This takes a different route: rather than repairing the averaging, it routes the fp32 path through the combiner the other two optimizers already use, so there is one implementation of the correct thing instead of two.

### DIFF
--- a/deepspeed/runtime/utils.py
+++ b/deepspeed/runtime/utils.py
@@ -389,9 +389,28 @@ def clip_grad_norm_(parameters, max_norm, norm_type=2, mpu=None):
         parameters = [parameters]
     parameters = list(filter(lambda p: p.grad is not None, parameters))
     norm_type = float(norm_type)
+
+    # Expert parameters are not replicated across the data parallel group, so the
+    # rank-averaging below cannot reconstruct a global norm from them: each rank's
+    # norm covers a different set of experts. Combine those the way the bf16 and
+    # fp16 optimizers already do, over the expert parallel group.
+    expert_tensors: Dict[str, List[torch.Tensor]] = {}
+    for p in parameters:
+        if not is_moe_param(p):
+            continue
+        # An expert parameter built by hand may carry no group name, and without one
+        # there is no expert parallel group to reduce over. Leave every such call on
+        # the original path rather than guessing which group it belongs to.
+        group_name = getattr(p, "group_name", None)
+        if group_name is None:
+            expert_tensors = {}
+            break
+        expert_tensors.setdefault(group_name, []).append(p.grad.data)
+    norm_parameters = [p for p in parameters if not is_moe_param(p)] if expert_tensors else parameters
+
     all_norms = []
     if norm_type == inf:
-        for p in parameters:
+        for p in norm_parameters:
             all_norms.append(p.grad.data.abs().max().float())
         total_norm = torch.stack(all_norms).max()
         total_norm = total_norm.to(get_accelerator().current_device_name())
@@ -400,7 +419,7 @@ def clip_grad_norm_(parameters, max_norm, norm_type=2, mpu=None):
             dist.all_reduce(total_norm, op=dist.ReduceOp.MAX, group=mpu.get_model_parallel_group())
     else:
         total_norm = 0
-        for p in parameters:
+        for p in norm_parameters:
             if mpu is not None:
                 if (mpu.get_model_parallel_rank() == 0) or is_model_parallel_parameter(p):
                     param_norm = p.grad.data.detach().float().norm(norm_type)
@@ -421,14 +440,29 @@ def clip_grad_norm_(parameters, max_norm, norm_type=2, mpu=None):
             dist.all_reduce(total_norm, op=dist.ReduceOp.SUM, group=mpu.get_model_parallel_group())
         total_norm = total_norm.pow(1. / norm_type)
 
-    # Need to average total_norm across different GPUs due to the presence of moe params
-    pg = groups._get_data_parallel_group()
-    scaled_norm = total_norm * 1.0 / float(dist.get_world_size(group=pg))
-    scaled_norm_tensor = scaled_norm
+    if expert_tensors:
+        # `total_norm` now covers only the replicated parameters, so it is already the
+        # same on every rank. Fold the experts in over their own group, counting each
+        # expert once, rather than averaging norms that describe different parameters.
+        moe_norm = get_norm_with_moe_layers(total_norm,
+                                            mpu=mpu,
+                                            expert_tensors=expert_tensors,
+                                            norm_type=norm_type)
+        # That helper reports a non-finite norm as -1. Left as-is it would make
+        # `clip_coef` negative and flip the sign of every gradient; inf keeps the
+        # behaviour the averaging path already had, which is to scale them to zero.
+        if moe_norm == -1:
+            moe_norm = float('inf')
+        total_norm = torch.tensor([float(moe_norm)], device=parameters[0].device, dtype=torch.float)
+    else:
+        # Need to average total_norm across different GPUs due to the presence of moe params
+        pg = groups._get_data_parallel_group()
+        scaled_norm = total_norm * 1.0 / float(dist.get_world_size(group=pg))
+        scaled_norm_tensor = scaled_norm
 
-    dist.all_reduce(scaled_norm_tensor, group=pg)
-    total_norm = scaled_norm_tensor
-    total_norm = total_norm.to(parameters[0].device)
+        dist.all_reduce(scaled_norm_tensor, group=pg)
+        total_norm = scaled_norm_tensor
+        total_norm = total_norm.to(parameters[0].device)
 
     max_norm = torch.tensor([float(max_norm)], device=total_norm.device)
     clip_coef = max_norm / (total_norm + 1e-6)
--- a/deepspeed/runtime/utils.py
+++ b/deepspeed/runtime/utils.py
@@ -399,25 +399,33 @@ def clip_grad_norm_(parameters, max_norm, norm_type=2, mpu=None):
     # The two branches below issue different collectives, so this decision has to
     # come out the same on every rank -- and a rank whose experts received no tokens
     # this step has expert parameters but no expert gradients.
-    # A rank can legitimately own no expert at all -- under pipeline parallelism, say --
-    # so reading ownership locally is not enough to guarantee the same choice everywhere.
-    # One scalar settles it. The second slot carries "an expert here has no group name",
-    # which has no expert parallel group to reduce over, so such a call stays on the
-    # original path rather than guessing which group the parameter belongs to.
-    vote = torch.tensor(
-        [
-            float(any(is_moe_param(p) and getattr(p, "group_name", None) is not None for p in parameters)),
-            float(any(is_moe_param(p) and getattr(p, "group_name", None) is None for p in parameters)),
-        ],
-        device=get_accelerator().current_device_name(),
-        dtype=torch.float)
-    dist.all_reduce(vote, group=groups._get_data_parallel_group())
-    if vote[0].item() > 0 and vote[1].item() == 0:
-        # The registry is built identically on every rank, so these names agree without
-        # a second collective even where the local parameters do not.
-        expert_group_names = sorted(getattr(groups, "_EXPERT_PARALLEL_GROUP", None) or {})
-    else:
-        expert_group_names = []
+    # The registry of expert groups is built identically on every rank, so an empty one
+    # means no MoE is configured anywhere and the choice is already made: take the
+    # original path, with no collective and no host sync. This is a per-step call from
+    # `_take_model_step`, so the common case must not pay for the MoE one.
+    registry = sorted(getattr(groups, "_EXPERT_PARALLEL_GROUP", None) or {})
+    expert_group_names = []
+    if registry:
+        # A rank can legitimately own no expert at all -- under pipeline parallelism, a
+        # stage holding only dense layers -- so reading ownership locally cannot guarantee
+        # the same choice everywhere. One scalar settles it, and it has to be reduced over
+        # the world rather than the data parallel group: under pipeline parallelism that
+        # group is one stage wide (`PipelineParallelGrid.get_data_parallel_group`), while
+        # the expert norm below reduces over the pipeline group, which spans stages. A
+        # per-stage vote let an expert stage and a dense stage issue different collectives
+        # on that group. The second slot carries "an expert here has no group name", which
+        # has no expert parallel group to reduce over, so such a call stays on the
+        # original path rather than guessing which group the parameter belongs to.
+        vote = torch.tensor(
+            [
+                float(any(is_moe_param(p) and getattr(p, "group_name", None) is not None for p in parameters)),
+                float(any(is_moe_param(p) and getattr(p, "group_name", None) is None for p in parameters)),
+            ],
+            device=get_accelerator().current_device_name(),
+            dtype=torch.float)
+        dist.all_reduce(vote, group=groups._clone_world_group())
+        if vote[0].item() > 0 and vote[1].item() == 0:
+            expert_group_names = registry
 
     parameters = list(filter(lambda p: p.grad is not None, parameters))
 
--- a/deepspeed/runtime/utils.py
+++ b/deepspeed/runtime/utils.py
@@ -387,32 +387,62 @@ def clip_grad_norm_(parameters, max_norm, norm_type=2, mpu=None):
     """
     if isinstance(parameters, torch.Tensor):
         parameters = [parameters]
-    parameters = list(filter(lambda p: p.grad is not None, parameters))
+    parameters = list(parameters)
     norm_type = float(norm_type)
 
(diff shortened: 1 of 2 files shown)


alanhuangyoo:
**Do not review this as it stands — two defects found by @gss10282023 on #8469, both of which I reproduced.**

1. **Collective deadlock** when one rank holds no expert gradient. I branch on whether *this rank* has expert gradients and the two branches issue different collectives, so a rank whose experts got no tokens this step waits on a peer that took the other path. `p.grad is not None` filtering makes that ordinary in MoE training, not a corner case.

2. **Exception** on expert-only `norm_type=inf`: `norm_parameters` is empty and `torch.stack([])` raises.

Marking this as a draft while I fix both. The branch decision has to be rank-independent — `_EXPERT_PARALLEL_GROUP` is built identically everywhere, so iterating its group names and contributing a zero tensor where this rank holds nothing keeps the collectives symmetric no matter which experts got tokens.

There is also #8476 from @gss10282023 covering the AutoEP FP32 L2 path. We are discussing on #8469 whether these should be one change; if a maintainer prefers a single PR, theirs came first.

alanhuangyoo:
@gss10282023 Both fixed in `338ed51`, and both reproduced on 2×H20 before I touched anything.

**Deadlock.** Your diagnosis was exactly right, and your own patch already had the answer in a comment I should have read more carefully:

> Inspect ownership before filtering gradients: an unused expert on one rank must not make that rank choose a different collective from the others.

I was reading ownership *after* `p.grad is not None` filtering. Ownership now comes from the parameters.

That alone was not enough, though — your case was "no expert **gradients**", and there is a harder one: a rank owning no expert **parameter** at all, which happens under pipeline parallelism. Local inspection cannot settle that in kind, so the choice is now made by one all-reduced flag, with group names taken from `_EXPERT_PARALLEL_GROUP` (built identically on every rank). A rank holding nothing for a group contributes a zero, which adds nothing to a sum of p-th powers and nothing to a max of absolute values.

Measured, same harness that hung before:

```
uneven  (rank1 owns no expert at all)     both ranks return 4.4721
nograd  (rank1's expert got no tokens)    both ranks return 4.4721
inf     (expert-only, norm_type=inf)      both ranks return 2.0000
```

`sqrt(2**2 + 4**2)` — the non-expert 2.0 replicated, plus one expert of norm 4.0 owned by rank 0. The original sharded-expert number is unchanged at 10.954452.

**Expert-only inf norm** was `torch.stack([])` with nothing replicated left to take a max over; zero is the identity there.

Three tests added for your cases. On `master`, run individually, all four fail with `AssertionError` — no hangs, since master has no expert branch to diverge on. (Run together they cascade into `TimeoutError`: a failed assertion in one rank of a `DistributedTest` leaves the other waiting. Worth knowing if you see that in your own runs.)

On the split: having read #8476, I do not think either subsumes the other. Yours keys on `ds_zero_placement_family == "autoep_expert"` with its own `ds_autoep_ep_size`, `norm_type == 2` and `mpu is None`; mine keys on `is_moe_param`/`group_name` and routes through `get_norm_with_moe_layers` for every norm type. Different families, different coverage.

What I would rather not do is leave `clip_grad_norm_` with two special cases bolted on. If AutoEP experts can be given a `group_name` and an entry in `_EXPERT_PARALLEL_GROUP`, the helper route covers both and the ownership weighting falls out of the group size — but you know the AutoEP placement model far better than I do, so tell me whether that holds. If it does not, I will take yours as the shape and fold legacy MoE into it.

Un-drafting since the two defects are closed, but I am in no hurry for this to land ahead of that conversation.

deepspeed/runtime/utils.py
```diff
@@ -387,20 +387,69 @@ def clip_grad_norm_(parameters, max_norm, norm_type=2, mpu=None):
...
+    # which has no expert parallel group to reduce over, so such a call stays on the
+    # original path rather than guessing which group the parameter belongs to.
+    vote = torch.tensor(
+        [
+            float(any(is_moe_param(p) and getattr(p, "group_name", None) is not None for p in parameters)),
+            float(any(is_moe_param(p) and getattr(p, "group_name", None) is None for p in parameters)),
+        ],
+        device=get_accelerator().current_device_name(),
+        dtype=torch.float)
+    dist.all_reduce(vote, group=groups._get_data_parallel_group())
```
ebarkhordar: I read this against 338ed51 rather than running it, so this is a question about the branch selector, not a repro.

Your comment above says the all-reduced flag is what settles the pipeline-parallel case, where a rank owns no expert parameter at all. The flag is reduced over `groups._get_data_parallel_group()`, and under pipeline parallelism that group does not span stages:

- `deepspeed/__init__.py:258` passes `model.mpu()` for a `PipelineModule`, which is the `PipelineParallelGrid` (`runtime/pipe/module.py:566`)
- `runtime/engine.py:2048` copies it into `groups.mpu`
- `utils/groups.py:707-711` then returns `mpu.get_data_parallel_group()`
- `runtime/pipe/topology.py:436` documents that as "The group of ranks within the same stage of all pipelines"

So each stage votes among itself. If experts sit on some stages and not others, two stages can reach different `expert_group_names` and issue different collectives, which is the divergence the vote is there to prevent. Your harness is 2 ranks with no pipe dimension, so it would not show this either way. Is `_clone_world_group()` what you want here, or does something already keep MoE inside a single stage?

One smaller thing on the same lines. `clip_grad_norm_` is a per-step call from `_take_model_step`, so the vote costs an all-reduce and two `.item()` syncs on every fp32 clip. With no MoE configured `_EXPERT_PARALLEL_GROUP` is empty and `sorted(... or {})` gives `[]` whichever way the vote goes, so reading the registry locally first would skip it, on the same "built identically on every rank" argument you use two lines down.

alanhuangyoo: You are right on both, and the first one is worse than a hole in a corner: it fails in the exact case the comment named as its reason for existing. Fixed in `1a761e8`.

**I checked your chain before changing anything**, since I could not reproduce it — my harness has no pipe dimension, so it would not show this either way, as you said.

- `PipelineParallelGrid.get_data_parallel_group` — "The group of ranks within the same stage of all pipelines". One stage wide.
- `get_global_norm_of_tensors`, with `mpu` set and MoE-TP disabled, takes `elif bwc_pipeline_parallel_world_size(mpu) > 1` and reduces over `bwc_pipeline_parallel_group(mpu)` → `get_pipe_parallel_group()`, "The group of ranks within the same pipeline". **That one spans stages.**

So a stage that votes "experts" issues that pipeline-group all-reduce inside `get_norm_with_moe_layers`, and a stage that votes "no experts" never does — mismatched calls on one cross-stage communicator. I also checked nothing refuses MoE with a `PipelineModule` (no assert in `pipe/engine.py` or `pipe/module.py`), and the pipeline branch in the norm code only makes sense if the combination is meant to work, so I do not think anything keeps MoE inside one stage. The vote now reduces over `_clone_world_group()`. Its lazy `new_group` is itself a collective, but entry is decided by the registry, which every rank holds identically, so all ranks reach it together.

This is verified by reading, not by running. If you have a PP box, the case worth a real run is experts on some stages and not others.

**On the cost**, you were right and it is now skipped without MoE. Counted per call on the non-MoE path, 2 ranks:

```
master      all_reduce=1  item=0
338ed51     all_reduce=2  item=1
1a761e8     all_reduce=1  item=0
```

One correction to the count, for the record: without MoE it was one `.item()` rather than two, because `vote[0].item() > 0 and ...` short-circuits when the first slot is zero. Still a host sync on every fp32 clip, so the conclusion stands.

`test_runtime_utils.py`: 13 passed, and the three cases from @gss10282023 (idle experts, a rank owning no expert, expert-only inf) still return on both ranks.

ebarkhordar: Both read right to me at `1a761e8`.

The part I wanted to check is the new early exit, since entry to the collective now depends on a local read where it used to be unconditional. It holds. `_EXPERT_PARALLEL_GROUP` has exactly two writers, `_create_expert_and_data_parallel` (`utils/groups.py:393,403,461`) and `_create_expert_data_and_model_parallel` (`:588`), and every write sits behind an `if rank in ranks`. The enclosing loops partition the whole world though, in both functions and on all three topology paths, so a rank that calls either one always lands in exactly one expert group. On top of that `dist.new_group` has to be entered by every rank in the job, so a run where one rank built expert groups and another skipped them would have hung at construction rather than reaching a clip. Empty and non-empty cannot disagree across ranks at that point.

The dense stages the world vote now sends down the expert path are covered by the `setdefault(group_name, [zero])` from `338ed51`, and `_get_expert_parallel_group(exp_name)` resolves on those ranks for the same reason.

Read, not run. Still no PP box here, so the experts-on-some-stages case is as untested by me as by you.

alanhuangyoo:
@tohtana @sfc-gh-truwase Could one of you take a look when you have time? The branch merges cleanly with master. @ebarkhordar reviewed both revisions and ran the cases (thread above): the per-step cost is back to master's, and the pipeline-parallel vote now reduces over the world group.

With MoE and fp32 gradient clipping, expert gradients were counted twice in the global norm, so clipping triggered early and scaled every update too small.

PR #8479, opened 2026-09-10
Recognize a norm by its shape, not by whether its class name was listed

### LINKED ISSUE #8447: AutoTP loads module weights by hardcoded class name, so 219 of 233 transformers norm classes are silently left on meta
## Summary

`Loading.is_load_module` decides whether AutoTP loads a module's weights from the checkpoint, and it decides by **class name against a hardcoded list of 26**. A norm class not on that list never has its weights loaded: on the meta-device path its parameters stay on meta.

219 of the 233 norm classes in transformers 5.16.1 are not on the list, including every Gemma generation, every GLM, Granite, Olmo, MiniMax, Nemotron and Zamba.

## Demonstration

Two norm classes with identical structure — same base, same `weight` parameter, same forward — differing only in name. Model built under `torch.device("meta")`, then run through the AutoTP replacement with a state dict that gives both norms 7.0:

```
RESULT checkpoint says 7.0 for both norms
  LlamaRMSNorm  (on the list):  value=7.0
  GemmaRMSNorm  (not on it):    STILL ON META
```

The gate itself, on classes taken from transformers:

```
  LlamaRMSNorm             is_load_module=True
  Qwen3_5RMSNormGated      is_load_module=True
  GemmaRMSNorm             is_load_module=False
  Gemma3RMSNorm            is_load_module=False
  CohereLayerNorm          is_load_module=False
  DeepseekV32RMSNorm       is_load_module=False
  MiniMaxRMSNorm           is_load_module=False
  GraniteRMSNorm           is_load_module=False
```

## Why nothing catches it

```python
# deepspeed/module_inject/auto_tp.py
if Loading.is_load_module(child) and self.state_dict is not None:
    if any(checking_key in item for item in self.state_dict):
        Loading.load(child, self.state_dict, checking_key, self.mp_group)
    else:
        continue
```

`Loading.load` is the only thing that materializes a meta parameter here, and it is behind the gate. A norm has no children, so falling through to the recursion reaches nothing. The one leftover-meta case that *is* handled downstream is tied embeddings:

```python
# deepspeed/module_inject/replace_module.py
if embedding_weight is not None and hasattr(module, "lm_head") and ... module.lm_head.weight.is_meta:
    module.lm_head.weight = embedding_weight
```

which is a targeted fix for one module rather than a general sweep — if there were a general one, that line would not be needed.

## What is missing

Restricting to families people actually run, 47 classes:

```
Gemma2RMSNorm, Gemma3RMSNorm, Gemma3nRMSNorm, Gemma4RMSNorm, Gemma4UnifiedRMSNorm, GemmaRMSNorm,
Glm4MoeLiteRMSNorm, Glm4MoeRMSNorm, Glm4RMSNorm, Glm4vMoeRMSNorm, Glm4vMoeTextRMSNorm, Glm4vRMSNorm,
Glm5NextRMSNorm, Glm5NextTextRMSNorm, Glm5NextTextRMSNormGated, Glm5NextTextUnweightedRMSNorm,
GlmImageRMSNorm, GlmMoeDsaRMSNorm, GlmOcrRMSNorm, GlmRMSNorm,
Granite4VisionTextRMSNorm, GraniteMoeHybridRMSNorm, GraniteMoeHybridRMSNormGated, GraniteMoeRMSNorm,
GraniteMoeSWARMSNorm, GraniteMoeSharedRMSNorm, GraniteRMSNorm, GraniteSWARMSNorm,
MiniCPM3RMSNorm, MiniMaxM2RMSNorm, MiniMaxM3VLRMSNorm, MiniMaxRMSNorm, NemotronHRMSNorm,
Olmo2RMSNorm, Olmo3RMSNorm, OlmoHybridRMSNorm, OlmoHybridRMSNormGated, OlmoLayerNorm, OlmoeRMSNorm,
Phi4MultimodalRMSNorm, RecurrentGemmaRMSNorm, T5Gemma2RMSNorm, T5GemmaRMSNorm, VaultGemmaRMSNorm,
Zamba2RMSNorm, Zamba2RMSNormGated, ZambaRMSNorm
```

`GlmMoeDsaRMSNorm` is the norm of `inference-optimization/GLM-5.2-0.8B-A0.8B`, one of the models used in this repo's own testing.

## Where this leaves the maintenance model

#8306 added four names for Qwen3.5, which is the right fix under the current design and also shows the problem with it: the list has to grow by hand for every model family and every generation of it, and when it does not, the failure is silent rather than an error. transformers gains norm classes faster than the list does.

Two directions, and I would rather have your call than pick one:

1. **Decide structurally instead of by name.** The question the gate is really asking is "is this a leaf module whose parameters the checkpoint has and nothing else will replace". That is answerable from the module and the state dict — no list. It is the bigger change and needs care about what should *not* be loaded here.
2. **Fail loudly instead of silently.** Keep the list, but after replacement, check for parameters still on meta and raise naming them. That does not fix any model, but it converts a silent wrong-weights run into an error that says which class to add — which is what the current design needs to be usable.

Doing 2 is small and would have surfaced this immediately; 1 removes the maintenance entirely. I am happy to write either.

## What I did not verify

I demonstrated the gate and the meta parameter surviving replacement, not an end-to-end wrong generation from a real Gemma checkpoint — I do not have that setup here. If someone has an AutoTP inference run against a checkpoint for any family above, the quick check is whether any parameter is still `.is_meta` after `init_inference` returns.

## Reproduction

```python
import torch, deepspeed
from torch import nn
from deepspeed.module_inject.auto_tp import AutoTP
from deepspeed.module_inject.layers import set_autotp_mode

def norm_cls(name):
    def __init__(self, h=8):
        nn.Module.__init__(self); self.weight = nn.Parameter(torch.ones(h))
    return type(name, (nn.Module,), {"__init__": __init__})

class M(nn.Module):
    def __init__(self, h=8):
        super().__init__()
        self.lin = nn.Linear(h, h, bias=False)
        self.listed_norm = norm_cls("LlamaRMSNorm")(h)
        self.unlisted_norm = norm_cls("GemmaRMSNorm")(h)

with torch.device("meta"):
    model = M()
sd = {"model.lin.weight": torch.eye(8),
      "model.listed_norm.weight": torch.full((8,), 7.0),
      "model.unlisted_norm.weight": torch.full((8,), 7.0)}

set_autotp_mode(training=True)
autotp = AutoTP(module=model, all_reduce_linears=(), prefix="model", state_dict=sd,
                linear_layer_setting=(nn.Linear, nn.Embedding), orig_layer_impl=None, training_mode=True)
autotp.mp_size, autotp.mp_group = 1, None
autotp.update_linear_policies()
autotp._replace_module(model)

print(model.listed_norm.weight.is_meta, model.unlisted_norm.weight.is_meta)  # False True
```

alanhuangyoo:
Fixes #8447.

## The gate

`Loading.is_load_module` decides whether AutoTP loads a module's weights from the checkpoint:

```python
def is_load_module(module):
    load_layers = [nn.Linear, nn.Embedding, nn.LayerNorm]
    load_layer_names = ["LPLayerNorm", "SharedEmbedding", ..., "MoEGate"]   # 26 names
    return module.__class__ in load_layers or module._get_name() in load_layer_names
```

A norm whose class is not on that list returns `False`, `_replace_module` skips it, and on the meta-device path its weight is never materialized. It stays on meta, and nothing is logged.

transformers defines one norm class per architecture, so a list of names can only ever cover a fraction of them. Measured by scanning every `modeling_*.py` in transformers 5.12.1 for norm classes and asking the gate about each — using **one shared RMSNorm body so only the class name differs**, which is the whole point:

| | `master` | this branch |
| --- | ---: | ---: |
| norm class names found | 336 | 336 |
| recognized | **14** | **336** |
| silently skipped | **322** | **0** |

The skipped set includes every Gemma generation, every GLM, Granite, Olmo, MiniMax, Nemotron and Zamba.

## The change

Recognize a norm by shape rather than by name: a **leaf module whose only parameter is a 1-D `weight`** (plus an optional bias) is elementwise-affine, which is exactly what `Loading.load` already knows how to copy.

The name list is left in place, so nothing that matched before stops matching.

## It admits norms and nothing else

This is the half that needed checking, and it is identical on both sides:

| | `master` | this branch |
| --- | --- | --- |
| `nn.Linear` | True | True |
| `nn.Embedding` | True | True |
| `nn.LayerNorm` | True | True |
| `nn.Conv1d` | False | False |
| `nn.MultiheadAttention` | False | False |
| `nn.Sequential` (container) | False | False |
| leaf with a 2-D `weight` | False | False |

`nn.Linear` and `nn.Embedding` carry 2-D weights and are matched by class above, so they cannot start matching by shape; a container owns no parameters of its own; and a leaf carrying anything besides `weight`/`bias` is not a plain norm.

## Testing

2×H20, `tests/unit/v1/autotp/`. Two new cases, both asserting **which modules the gate accepts**, not that a code path ran:

```
branch                         2 passed
master + this branch's tests   1 failed
    AssertionError: a norm must be recognized by its shape,
                    not by whether its class name was listed
```

The second test passes on `master` too — deliberately: it pins the refusals so a future widening of the rule has to break something.

Full file, comparing failure *sets*:

```
master   17 failed, 12 passed
branch   16 failed, 13 passed

only failing on master               -> test_is_load_module_recognizes_a_norm_it_has_never_heard_of
failing on both                      -> 16
```

The 16 are pre-existing on this box — `ImportError: libcudart.so.13`, unrelated to this change.

### DIFF
--- a/deepspeed/module_inject/auto_tp.py
+++ b/deepspeed/module_inject/auto_tp.py
@@ -146,7 +146,28 @@ class Loading():
             "Qwen3_5MoeRMSNormGated", "DeepseekV2RMSNorm", "DeepseekV3RMSNorm", "DeepseekV2YarnRotaryEmbedding",
             "DeepseekV3YarnRotaryEmbedding", "MoEGate"
         ]
-        return module.__class__ in load_layers or module._get_name() in load_layer_names
+        if module.__class__ in load_layers or module._get_name() in load_layer_names:
+            return True
+        # Every norm above is on the list by class name, and transformers defines one
+        # norm class per architecture -- 233 of them in 5.16.1, of which this list names
+        # 14. The rest reach `_replace_module` and are skipped, so on the meta-device
+        # path their weights are never materialized and stay on meta.
+        #
+        # Recognise them by shape instead: a leaf module whose sole parameter is a 1-D
+        # `weight` is elementwise-affine, which is what `Loading.load` already knows how
+        # to copy. `nn.Linear` and `nn.Embedding` carry 2-D weights and are matched
+        # above, so this admits norms and nothing that wants a different load path.
+        return Loading._is_elementwise_affine_leaf(module)
+
+    def _is_elementwise_affine_leaf(module):
+        if next(module.children(), None) is not None:
+            return False
+        weight = getattr(module, "weight", None)
+        if not isinstance(weight, torch.nn.Parameter) or weight.dim() != 1:
+            return False
+        # A bias, if any, has to match; anything else means this is not a plain norm.
+        extra = {name for name, _ in module.named_parameters(recurse=False)} - {"weight", "bias"}
+        return not extra
 
     def load_buffer(module, state_dict, prefix):
         for name in module._buffers.keys():
--- a/tests/unit/v1/autotp/test_autotp_custom_patterns.py
+++ b/tests/unit/v1/autotp/test_autotp_custom_patterns.py
@@ -989,3 +989,64 @@ class TestAutoTPFusedWeights(DistributedTest):
 
 # Multi-model (teacher + student in one process) and multimodal-config regressions live in
 # tests/unit/v1/autotp/test_autotp_multiple_models.py so they run on the GPU workflow too.
+
+
+class _WeightOnlyNorm(nn.Module):
+    """A norm with the shape every transformers RMSNorm has: one 1-D `weight`."""
+
+    def __init__(self, size=8):
+        super().__init__()
+        self.weight = nn.Parameter(torch.ones(size))
+
+    def forward(self, hidden_states):
+        return hidden_states * self.weight
+
+
+def test_is_load_module_recognizes_a_norm_it_has_never_heard_of():
+    """Two norms with identical structure, differing only in class name (#8447).
+
+    `Loading.is_load_module` gated on a hardcoded list of 26 class names, so a norm
+    whose architecture was not on it returned False, `_replace_module` skipped it,
+    and on the meta-device path its weight was never materialized. transformers
+    defines one norm class per architecture, so the list named a small fraction of
+    them and the rest were silently left uninitialized.
+    """
+    from deepspeed.module_inject.auto_tp import Loading
+
+    on_the_list = type("LlamaRMSNorm", (_WeightOnlyNorm, ), {})()
+    not_on_the_list = type("GemmaRMSNorm", (_WeightOnlyNorm, ), {})()
+
+    assert Loading.is_load_module(on_the_list)
+    assert Loading.is_load_module(not_on_the_list), (
+        "a norm must be recognized by its shape, not by whether its class name was listed")
+
+
+def test_is_load_module_still_refuses_what_it_refused_before():
+    """The shape test must admit norms and nothing that wants a different load path."""
+    from deepspeed.module_inject.auto_tp import Loading
+
+    # 2-D weights: these are matched by class above and must not start matching by shape.
+    assert Loading.is_load_module(nn.Linear(4, 4))
+    assert Loading.is_load_module(nn.Embedding(4, 4))
+    assert Loading.is_load_module(nn.LayerNorm(4))
+
+    class _TwoDimLeaf(nn.Module):
+
+        def __init__(self):
+            super().__init__()
+            self.weight = nn.Parameter(torch.ones(4, 4))
+
+    assert not Loading.is_load_module(_TwoDimLeaf())
+    assert not Loading.is_load_module(nn.Conv1d(2, 2, 3))
+    assert not Loading.is_load_module(nn.MultiheadAttention(4, 2))
+    # A container owns its children's parameters but none of its own.
+    assert not Loading.is_load_module(nn.Sequential(nn.Linear(4, 4)))
+
+    class _NormWithExtraParam(_WeightOnlyNorm):
+
+        def __init__(self):
+            super().__init__()
+            self.scale = nn.Parameter(torch.ones(8))
+
+    assert not Loading.is_load_module(_NormWithExtraParam())
+


ebarkhordar (commented):
I ran the gate against `404f8172` in a clean container (torch 2.14.0+cpu, `auto_tp.py` imported from the PR tree). I wanted to check the "admits norms and nothing else" half by enumerating `torch.nn` rather than sampling it, and for weights the claim holds. Of 42 constructible classes, the ones whose verdict changes between the merge base and this branch are `BatchNorm1d/2d/3d`, `SyncBatchNorm`, `GroupNorm`, `PReLU` and torch's own `RMSNorm`. All of them are affine along dim 0, so `mp_replace.copy` replicates them, which is what you want.

The BatchNorm family is worth a second look, because it is the only newly admitted set that keeps state in buffers. When a module is admitted and its prefix is absent from the state_dict, `replace_module.py:695` and `auto_tp.py:766` take the `else: continue`, and that skips the `Loading.load_buffer` call two lines below it. That call is the only thing that moves a buffer off meta.

```
meta-init model with a BatchNorm1d child, state_dict carrying no 'norm.' key
                          running_mean.is_meta   running_var.is_meta
bb7ad0d82 (merge base)    False                  False
404f8172  (this branch)   True                   True
```

A state_dict that does not carry a prefix is ordinary once a checkpoint is sharded, since `replace_module()` loads one file into `sd` and hands that to `_replace_module`. I drove `_replace_module` directly rather than a real sharded model, so I am reporting the path and the measurement and not a user report.

The hole is the caller's, not yours. On the merge base a named rotary entry already reaches it, and `inv_freq` stays on meta at both SHAs. What your change decides is how many modules get there.

One edit keeps it in scope, and it is safe because `_is_elementwise_affine_leaf` already requires a leaf, so nothing else below the `continue` applies:

```python
    else:
        if len(child._buffers) != 0:
            Loading.load_buffer(child, state_dict, checking_key)
        continue
```

alanhuangyoo:
@ebarkhordar Thanks, that's right, and I could reproduce it the same way: a meta-init `BatchNorm1d` child with no `norm.` key in the state_dict stays on meta on 404f8172 and not on the merge base.

Applied your edit in 292bed80 in both places (`replace_module._replace_module` and `AutoTP._replace_module`): when an admitted module's prefix is missing, load its buffers before the `continue`. That is what the merge base already did for BatchNorm, since it went down the buffer path unadmitted.

Added `test_admitted_norm_absent_from_the_state_dict_still_gets_its_buffers_off_meta`, which drives `_replace_module` the way you did. It fails on 404f8172 and passes now; `test_autotp_custom_patterns.py` is 10 passed / 20 skipped on CPU. The test only covers the `replace_module` call site; the `AutoTP` one got the same three lines.

ebarkhordar:
I re-ran 292bed80 against bb7ad0d82 and 404f81726 in the same container as before (dscpu:1, torch 2.14.0+cpu, provenance `auto_tp __file__=/src/deepspeed/module_inject/auto_tp.py`). The BatchNorm case is fixed and your new test's assertions hold.

One thing the new call reaches that neither of those refs did. The `else` branch now runs for every admitted module whose prefix is missing, and six names already on `load_layer_names` are rotary embeddings whose buffers are non-persistent, so they are in no shard at all. `load_buffer` materializes them with `torch.empty_like` and then has nothing to copy:

```
                          bb7ad0d8   404f8172   292bed80
BatchNorm1d running_var   off meta   meta       off meta
Phi3Rotary  inv_freq      meta       meta       off meta, uninitialised
```

For a rotary built with `torch.arange(1, 5)`, `inv_freq` at 292bed80 comes back `[3.02e-19, 3.17e-41, 0.0, 0.0]`. On both earlier refs it stayed on meta and raised on first use, so this turns a loud failure into a plausible wrong number.

Gating the new call on persistence restores bb7ad0d8 for those and leaves your fix alone:

```python
if any(n not in child._non_persistent_buffers_set for n in child._buffers):
    Loading.load_buffer(child, state_dict, checking_key)
```

With that in both files, `inv_freq` is back on meta, BatchNorm is still off meta, and `test_autotp_custom_patterns.py` is 10 passed / 20 skipped either way.

Same caveat as last time: I drove `_replace_module` directly, not an end-to-end sharded load. The rotary is a stand-in named `Phi3RotaryEmbedding` so the name list matches; the real one registers `inv_freq` as `nn.Buffer(..., persistent=False)` (`modeling_phi3.py:82`). I did not check the other five.

alanhuangyoo:
@ebarkhordar Right, turning a loud meta failure into a plausible wrong number is worse than what I was fixing. I reproduced it: on 292bed80, a meta `Phi3RotaryEmbedding` with a non-persistent `inv_freq` and no key in the shard comes back off meta.

Your gate is now in both call sites (121cd03). I added `test_non_persistent_buffers_of_an_absent_module_stay_on_meta` next to the BatchNorm test. It fails on 292bed80 and passes now, and the BatchNorm test still passes. `test_autotp_custom_patterns.py` is 11 passed / 20 skipped on CPU.

The gate doesn't cover a module that has both persistent and non-persistent buffers: its non-persistent ones would still be materialized. The in-shard path just below already does that today for any module with buffers, so I left it out of this PR.

ebarkhordar:
Verified at 121cd03d in a clean container (python:3.12-slim, torch 2.14.0+cpu, provenance printed as `/src/deepspeed/module_inject/auto_tp.py`). Same script at four SHAs, meta-init model whose shard carries only `fc.*`:

```
                    bb7ad0d8(base)  404f8172      292bed80      121cd03d
rotary, absent      meta            meta          materialized  meta
batchnorm, absent   materialized    meta          materialized  materialized
```

`test_non_persistent_buffers_of_an_absent_module_stay_on_meta` fails on 292bed80 and passes on 121cd03d, the BatchNorm test passes at both, and the file is 11 passed / 20 skipped here too.

On the mixed case, one correction to where the line sits. I measured the in-shard path at the merge base and it does materialize a non-persistent buffer there, so that half holds. The else path does not: at bb7ad0d8 a listed module absent from the shard left every buffer on meta, so for a module carrying both kinds this would be a change against master rather than something master already does. I could not name a class on `load_layer_names` that has both today, so this is not a request to widen the PR.

alanhuangyoo:
Thanks for checking all four SHAs. You're right about the else path: at the merge base, a listed module absent from the shard left all its buffers on meta, so the mixed case would be new behavior, not something master already did. Nothing on `load_layer_names` has both kinds today. If one ever does, the fix is to gate per buffer.

alanhuangyoo:
@hwchen2017 @loadams Could one of you take a look when you have time? It merges cleanly with master. @ebarkhordar verified 121cd03 against master and the previous two commits (table above).

`is_load_module` recognizes norms only by class name. Measured against transformers 5.12, 322 of 336 norm classes weren't on the list, so on the meta-device path their weights were never loaded. That covers every Gemma generation, GLM, Granite, Olmo and others.

Issue #8461, opened 2026-09-08
Muon cannot run under BF16_Optimizer: flat partitions with no Newton-Schulz applied anywhere


alanhuangyoo:
Split out of #8442 at @delock's suggestion, so the refusal that PR adds has somewhere to be lifted from.

## What is refused

`{"bf16": {"enabled": true}, "data_types": {"grad_accum_dtype": "fp32"}, "zero_optimization": {"stage": 1}}` with Muon now raises `ZeRORuntimeException` instead of training. That combination selects `BF16_Optimizer`, and Muon cannot run under it.

## Why

`MuonWithAuxAdam.step` decides what it has been handed by shape:

```python
if p.dim() < 2:
    # A flat ZeRO partition -- the gradient already holds the update
    update = p.grad
else:
    # The weight itself, so nothing has orthogonalized it
    update = muon_update(p.grad, state["momentum_buffer"], ...)
```

`BF16_Optimizer` replaces the param groups with flat fp32 partitions, exactly like the ZeRO 1/2 and 3 optimizers, so every parameter reads as 1-D and `step` takes the first branch. But unlike them it never applies Muon anywhere: `DeepSpeedZeroOptimizer` and `_Stage3` orthogonalize while the partition is filled, and `BF16_Optimizer` has no such call. So the first branch's premise -- "ZeRO already did the update" -- is false, and the step is plain SGD with momentum.

Nothing in the flat partition records the original 2-D shapes, so `step` cannot recover them to orthogonalize there.

## What lifting it would take

Either:

1. **Apply Muon in `BF16_Optimizer` where the partition is filled**, matching what `DeepSpeedZeroOptimizer.get_flat_partition` does -- it has the parameter list, so the shapes are reachable at that point; or
2. **Carry the shapes onto the flat partition** so `step` can orthogonalize per original tensor, which is a wider change and duplicates work the ZeRO optimizers already do in the right place.

(1) looks like the right shape to me, but I have not written it and would rather not guess at the ordering constraints inside `BF16_Optimizer`'s accumulation window.

## How it was found

Re-reading my own claim in #8442. That PR's description said Muon runs correctly under "every wrapper", which was generalised from four configurations I had measured; `BF16_Optimizer` is a fifth and behaves differently. I posted the hole on the PR before it was reviewed rather than letting it stand.

[...]

PR #8483, opened 2026-09-11
Give Muon's momentum the dtype of the gradient it is combined with

alanhuangyoo:
## The bug

`muon_update` does `momentum.lerp_(grad)`, which requires both operands to have the same dtype. The momentum buffer is allocated in the configured gradient accumulation dtype — but the gradients only arrive in that dtype while `use_grad_accum_attribute` is on:

```python
self.use_separate_grad_accum = (self.dtype != self.gradient_accumulation_dtype)
if self.use_separate_grad_accum and not self.partition_gradients:
    self.use_grad_accum_attribute = True
```

`partition_gradients` is true at stage 2, so the attribute path is off there and `get_param_gradient_attribute` returns `param.grad` in the **parameter** dtype instead. bf16 with `grad_accum_dtype: fp32` at ZeRO stage 2 therefore cannot run Muon at all:

```
RuntimeError: expected dtype torch.float32 for `end`, but got dtype torch.bfloat16
```

The dtype now comes from the gradient rather than from the configuration.

## Why #8433 did not already cover this

There are two copies of the buffer setup. `get_flat_partition` delegates to `_get_flat_partition_unpadded` for any group with no round-robin padding:

```python
if not any(self.round_robin_bit16_padding[param_group_idx]):
    return self._get_flat_partition_unpadded(...)
```

#8433 reconciled a restored checkpoint's buffer dtype in the first copy. **The second never got it** — and it is the one this configuration takes. I found this by adding a debug print to the code I assumed was running and watching it never fire; the traceback goes through `_get_flat_partition_unpadded`, not `get_flat_partition`.

So rather than add the same reconciliation a second time — duplication is what caused the miss — both paths now call one `_muon_momentum_buffer`, which allocates in the gradient's dtype and converts a restored buffer into it.

**That means this also repairs #8433's own case on the unpadded path.** Its regression tests fail on current `master`, which carries #8433 (`b726f4ed`):

```
master   4 failed, 2 passed      TestMuonCheckpointRoundTrip::test_resumes_with_its_momentum
branch   6 passed

master's failure: RuntimeError: expected dtype torch.float32 for `end`,
                  but got dtype torch.bfloat16
```

I verified `master` really does contain #8433 before believing that, rather than assuming a stale checkout.

## Testing

2×H20.

New case, `TestMuonSeparateGradAccumDtype`:

```
branch   1 passed
master   1 failed
```

It parametrizes stage 2 only — stage 1 with the same combination selects `BF16_Optimizer`, which refuses Muon for a different reason (#8461). My first version parametrized both and failed on the branch too; that was the test's fault, not the fix's.

`tests/unit/ops/muon/`, comparing failure *sets*:

```
master   70 failed, 83 passed
branch   66 failed, 87 passed

only failing on master               -> the 4 TestMuonCheckpointRoundTrip cases above
failing on both                      -> 66
```

The 66 are pre-existing on this box — `CUDAMismatchException` between the installed toolkit and the one torch was built against, unrelated to this change.

### DIFF
--- a/deepspeed/runtime/zero/stage_1_and_2.py
+++ b/deepspeed/runtime/zero/stage_1_and_2.py
@@ -2231,26 +2231,8 @@ class DeepSpeedZeroOptimizer(ZeROOptimizer):
         current_size = 0
         # find the flatten copy in the optimizer's state
         flatten_copy = self.optimizer.param_groups[param_group_idx]['params'][0]
-        if (not self.optimizer.state[flatten_copy]) and getattr(
-                tensor_list[0], 'use_muon', False) and 'muon' in self.optimizer.__class__.__name__.lower():
-            self.optimizer.state[flatten_copy] = {}
-        if getattr(tensor_list[0], 'use_muon', False) and 'muon' in self.optimizer.__class__.__name__.lower():
-            momentum_buffer = self.optimizer.state[flatten_copy].get("momentum_buffer")
-            if momentum_buffer is None:
-                # need to check the total # of elements in the parameters in this group and this partition
-                total_size = sum([t.numel() for t in tensor_list])
-                flatten_bf_list = [torch.zeros([total_size], dtype=dtype, device=device)]
-                self.optimizer.state[flatten_copy]["momentum_buffer"] = self.flatten(flatten_bf_list)
-            elif momentum_buffer.dtype != dtype:
-                # A restored buffer arrives in the dtype the checkpoint holds optimizer state
-                # in, which is fp32, while the gradients it is combined with are in the
-                # gradient accumulation dtype. muon_update does momentum.lerp_(grad), which
-                # requires both to match, so resuming a bf16 run raised:
-                #   RuntimeError: expected dtype torch.float32 for `end`, but got dtype
-                #   torch.bfloat16
-                # Convert rather than reallocate: the momentum a resume just restored is the
-                # reason the checkpoint carries it.
-                self.optimizer.state[flatten_copy]["momentum_buffer"] = momentum_buffer.to(dtype=dtype, device=device)
+        if self._is_muon_group(tensor_list):
+            self._muon_momentum_buffer(tensor_list, param_group_idx, dtype, device)
 
         partition_id = dist.get_rank(group=self.real_dp_process_group[param_group_idx])
         buffer_idx = 0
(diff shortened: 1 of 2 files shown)


delock:
Hi @alanhuangyoo I think the reason that #8433 are not caught is because these tests (which requires >=2 accelerators) skipped on CPU test.  Can you check which test needs >=2 accelerator tests?  We may need to move these tests to v1 to ensure they will not be on SKIP list.   Thanks!

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2298,6 +2280,43 @@ def get_flat_partition(self,
...
+        grads = self.all_grad_tensors.get(param_group_idx)
+        momentum_dtype = grads[0].dtype if grads else dtype
+
+        state = self.optimizer.state[flatten_copy]
+        buffer = state.get("momentum_buffer")
+        if buffer is None:
+            total_size = sum([t.numel() for t in tensor_list])
+            buffer = self.flatten([torch.zeros([total_size], dtype=momentum_dtype, device=device)])
+            state["momentum_buffer"] = buffer
+        elif buffer.dtype != momentum_dtype or buffer.device != torch.device(device):
```
delock: Is the second condition (device in-consistency) a real scenario?   Would the buffer come from CPU?  If it is a dead condition better remove.

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2298,6 +2280,43 @@ def get_flat_partition(self,
...
+        if not self.optimizer.state[flatten_copy]:
+            self.optimizer.state[flatten_copy] = {}
+        grads = self.all_grad_tensors.get(param_group_idx)
+        momentum_dtype = grads[0].dtype if grads else dtype
+
+        state = self.optimizer.state[flatten_copy]
+        buffer = state.get("momentum_buffer")
+        if buffer is None:
+            total_size = sum([t.numel() for t in tensor_list])
+            buffer = self.flatten([torch.zeros([total_size], dtype=momentum_dtype, device=device)])
```
delock: it looks like flatten here is redudant.  This line could be simpler

deepspeed/runtime/zero/stage_1_and_2.py
```diff
@@ -2231,26 +2231,8 @@ def get_flat_partition(self,
...
-                # A restored buffer arrives in the dtype the checkpoint holds optimizer state
-                # in, which is fp32, while the gradients it is combined with are in the
-                # gradient accumulation dtype. muon_update does momentum.lerp_(grad), which
-                # requires both to match, so resuming a bf16 run raised:
-                #   RuntimeError: expected dtype torch.float32 for `end`, but got dtype
-                #   torch.bfloat16
-                # Convert rather than reallocate: the momentum a resume just restored is the
-                # reason the checkpoint carries it.
-                self.optimizer.state[flatten_copy]["momentum_buffer"] = momentum_buffer.to(dtype=dtype, device=device)
+        if self._is_muon_group(tensor_list):
```
delock, line 2240: Did current test also cover padded path?  If not better cover it with one test as guard.  Does not need to be too comprehensive.

delock (changes requested):

alanhuangyoo:
@delock Checked. The CPU runner has one device, so every `DistributedTest` with `world_size = 2` under `tests/unit/ops/muon` got skipped there:
- all of `test_muon_checkpoint.py`, which is where #8433's tests live
- `TestMuonSurvivesLossScaleBackoff` in `test_muon_overflow.py`
- `TestMuonConfigs`, `TestGramNewtonSchulz` and `TestMuonZero12NumericalCorrectness` in `test_muon.py`

In 862174a I moved those three files to `tests/unit/v1/ops/muon/`. On 2×H20, `test_muon_checkpoint.py` and `test_muon_overflow.py` pass. In `test_muon.py`, only the cases that JIT-build fused Adam / cpu_adam failed, on a CUDA version mismatch in my environment. The rest pass.

deepspeed/runtime/zero/stage_1_and_2.py
alanhuangyoo: Not a real case. Both callers pass the current accelerator device, cpu_offload returns before `get_flat_partition`, and `load_state_dict` puts restored state on the parameter's device. Removed it, so only the dtype check is left.

deepspeed/runtime/zero/stage_1_and_2.py
alanhuangyoo: Right, it was a no-op on one 1-D tensor. It is a plain `torch.zeros` now.

deepspeed/runtime/zero/stage_1_and_2.py
alanhuangyoo: It did not. Added a case with `parameter_alignment` on and 63×63 weights, and it asserts the group is actually padded. Both cases fail on master with the fp32/bf16 `lerp` error and pass here (2×H20).

delock (approved):

PR #8496, opened 2026-09-13
Keep ZenFlow Adam step counters in float32

vineethsaivs:
ZenFlow's selective Adam counter stops advancing at 256 in BF16 and 2048 in FP16, freezing bias correction.

Initialize counters in float32. Direct optimizer `load_state_dict()` also promotes old FP16/BF16 counters, preserving float64 counters and saved values. This restore coverage is for the selective optimizers used through PyTorch's API: DeepSpeed's engine does not currently checkpoint their separate state.

Test: `DS_ACCELERATOR=cpu PYTHONPATH=. python -m pytest tests/unit/runtime/zenflow/test_zf.py -k counter -q`. All 13 counter tests pass on Apple M2 Pro CPU, including direct state restoration and resumed updates. Previously lost step counts cannot be reconstructed.

### DIFF
--- a/deepspeed/ops/adam/zenflow_torch_adam.py
+++ b/deepspeed/ops/adam/zenflow_torch_adam.py
@@ -42,6 +42,13 @@ if not _ZENFLOW_AVAILABLE:
 
 class ZenFlowSelectiveAdamW(torch.optim.AdamW):
 
+    def __setstate__(self, state):
+        super().__setstate__(state)
+        for param_state in self.state.values():
+            step = param_state.get("step")
+            if torch.is_tensor(step) and step.dtype in (torch.float16, torch.bfloat16):
+                param_state["step"] = step.float()
+
     def __init__(self, *args, offload=False, bucket_size=5e8, **kwargs):
         if not _ZENFLOW_AVAILABLE:
             raise RuntimeError("ZenFlow features are not available with PyTorch < 2.0. "
@@ -115,7 +122,7 @@ class ZenFlowSelectiveAdamW(torch.optim.AdamW):
 
                     state = self.state.setdefault(param, {})
                     if len(state) == 0:
-                        state["step"] = torch.zeros((), dtype=param.dtype, device=selected_param.device)
+                        state["step"] = torch.zeros((), dtype=torch.float32, device=selected_param.device)
                         state["exp_avg"] = torch.zeros_like(selected_param)
                         state["exp_avg_sq"] = torch.zeros_like(selected_param)
                         if amsgrad:
@@ -218,7 +225,7 @@ class ZenFlowSelectiveAdamW(torch.optim.AdamW):
 
                     state = self.state.setdefault(param, {})
                     if len(state) == 0:
-                        state["step"] = torch.zeros((), dtype=param.dtype, device=selected_param.device)
+                        state["step"] = torch.zeros((), dtype=torch.float32, device=selected_param.device)
                         if amsgrad:
                             state["max_exp_avg_sq"] = torch.zeros_like(selected_param)
                         if not self.offload:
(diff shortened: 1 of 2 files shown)


deepspeed/ops/adam/zenflow_torch_adam.py
```diff
@@ -42,6 +42,13 @@ def wrapper(fn):
 
 class ZenFlowSelectiveAdamW(torch.optim.AdamW):
 
+    def __setstate__(self, state):
+        super().__setstate__(state)
+        for param_state in self.state.values():
+            step = param_state.get("step")
+            if torch.is_tensor(step) and step.dtype in (torch.float16, torch.bfloat16):
+                param_state["step"] = step.float()
```
ebarkhordar, line 50: I read this at `4e23970` and went looking for where a bf16 step tensor comes from inside DeepSpeed, to see which half of the patch each case needs.

The creation half lands where it should: `zenflow_stage_1_and_2.py:81` builds the selective optimizer over `self.bit16_groups`, so `param.dtype` is bf16 or fp16 for every real ZenFlow run, not just a hand-built one.

The `__setstate__` half is what I could not place. Which path hands these classes a checkpoint whose selective `step` is bf16? `ZenFlowZeroOptimizer` defines no `state_dict` and no `load_state_dict` of its own, so it inherits `DeepSpeedZeroOptimizer.state_dict`, which serializes `self.optimizer`. `selective_optimizer` is a separate instance attached beside it (`engine_stage3.py:120` does the same for stage 3), and `selective_optimizer.state_dict(` and `selective_optimizer.load_state_dict(` occur zero times in the package.

If that reading is right, the selective `step`, `exp_avg` and `exp_avg_sq` never enter a DeepSpeed checkpoint, so `__setstate__` can only fire for someone driving `ZenFlowSelectiveAdamW` through torch APIs directly, and a ZenFlow run resumed from a checkpoint restarts its selective moments at zero whatever the dtype was. That would make the second half a different bug from the first one rather than the resume story for it.

I did not run this. The above is from reading the class bodies at that sha and searching the package for the two calls, so it is a question and not a claim about behaviour.

sfc-gh-truwase: Good catch. I am curious why `step` is not an integer?

ebarkhordar: It is a tensor because the vendored `_single_tensor_adamw` in this file takes `state_steps: List[Tensor]` and increments it in place at line 739 (`step_t += 1`), and it is float rather than int because that is torch's own scalar-state convention, not a ZenFlow choice. `torch/optim/adam.py` at v2.8.0, which `AdamW` subclasses, lazily creates it as:

```python
state["step"] = (
    torch.zeros((), dtype=_get_scalar_dtype(is_fused=group["fused"]), device=p.device)
    if group["capturable"] or group["fused"]
    else torch.tensor(0.0, dtype=_get_scalar_dtype())
)
```

`_get_scalar_dtype` (`torch/optim/optimizer.py:211`) returns `torch.float32`, or `float64` when the default dtype is float64, and `float32` when fused. There is no integer branch. The capturable path needs it floating: at lines 749 to 757 here the step feeds `1 - beta1**step`, `lr / bias_correction1` and `step_size.neg()` as tensor math, and upstream checks `state["step"].requires_grad` in differentiable mode (`adam.py:195`).

That also explains why upstream does not already catch what this PR fixes. `Adam.__setstate__` runs a normalizing pass over loaded state, but it is guarded on `not torch.is_tensor(p_state["step"])` (`adam.py:126`), so it repairs an int or a python float and lets a step that is already a tensor through with whatever dtype it carries. A bf16 step tensor, which is what the old `dtype=param.dtype` sites produced, is precisely the shape that guard skips. So the `__setstate__` added here fills a gap in upstream's normalizer rather than duplicating it.

I read this rather than ran it: torch v2.8.0 source, and the class bodies at `4e23970`.

vineethsaivs: The restore tests exercise the selective optimizers directly; DeepSpeed's engine checkpoint does not save them. I've clarified that limit in the description, and all 13 counter tests pass.

Issue #8462, opened 2026-09-08
[FEATURE] Align decode fronts and trim dead zones in continuous batching


nathon-lee:
## Context

The current continuous-batching rollout path assumes equal prompt widths and
uses a shared logical decode position for active rows. This keeps the initial
prototype simple, but it can require a larger cache buffer than necessary when
prompt lengths vary.

For rollout workloads, we do not need to preserve serving-oriented TTFT/TPOT
ordering. This gives us an opportunity to reorder prompts internally and
reduce unused cache space.

## Proposed direction

1. Sort prompts from longest to shortest internally and right-align the initial
   prompts so that their decode fronts are aligned.
2. When a row retires, admit a pending prompt only when its prefill can fit
   behind the current decode front.
3. Track each active row's logical span start and position independently.
4. Periodically detect dead zones before the earliest active span and trim them
   by shifting active KV rows, attention metadata, and write positions to the
   left.
5. Restore the original request-row order in the returned `RolloutBatch`.

## Correctness requirements

- Support mixed prompt widths.
- Preserve token-level equality with the sequential eager reference.
- Cover EOS retirement and prompt refill.
- Verify KV-cache contents, attention masks, logical positions, and write
  positions after trimming.
- Preserve the original request-row order in the output.

## Performance requirements

Report:

- peak cache length;
- cache memory usage;
- bytes moved during trimming;
- trimming frequency and latency;
- end-to-end rollout throughput.

The cache capacity should be bounded by the maximum prompt length plus the
shared generation budget, subject to the selected scheduling policy.

## Scope

Follow-up to DeepSpeed PR #8368. The current equal-width StaticCache path remains
the baseline until this optimization is implemented and validated. 
cc @delock  Appreciate your insights. Happy to hear any feedback or recommendations.

PR #8497, opened 2026-09-13
feat(rollout): align decode fronts and trim dead zones

nathon-lee:
## Summary

This PR adds an opt-in decode-front alignment path for continuous-batching rollouts.

When `align_decode_fronts=False`, the existing equal-width padded layout remains the default. Requests may have different effective lengths in their attention masks, but their physical decode fronts are not right-aligned or compacted.

When `align_decode_fronts=True`, the rollout:

* derives effective prompt lengths from attention masks;
* orders requests from longest to shortest;
* right-aligns prefill spans;
* tracks logical positions and physical write positions separately;
* trims dead cache columns with attention metadata and write positions;
* restores the original request order in the returned batch.

The scheduler no longer exposes `admit_if`. Trimming preserves enough physical cache span for prompts admitted during the current refill.

### Test Results
- ✅ 14 tests passed
- ✅ Changed-file pre-commit checks passed

### Device-Backed Validation
- **GPU:** NVIDIA RTX 4000 Ada, 20 GB
- **PyTorch:** 2.9.1+cu128
- **CUDA:** 12.8
- **Transformers:** 4.57.1
- **Model:** Qwen/Qwen2.5-3B-Instruct
- **Precision:** FP32

> The aligned and equal-width outputs matched the single-request eager oracle token by token.

#### Workload 1: 32 mixed left-padded requests
* **Prompt shape:** `[32, 20]`
* **max_new_tokens:** 16

| Case | Time | Throughput |
| :--- | :--- | :--- |
| **no-CB** | 1183.46 ms | 432.63 tok/s |
| **CB equal-width, batch 32** | 1418.99 ms | 360.82 tok/s |
| **aligned CB, batch 32** | 2600.53 ms | 196.88 tok/s |
| **dynamic CB, continuous_batch_size=8** | 4315.39 ms | 118.65 tok/s |

> The optimization is workload-dependent. For uniform short prompts, aligned CB is slower because per-request prefill and cache-management overhead dominate. For long-tail prompt lengths, aligned CB can be much faster by avoiding padded prefill computation.

#### Workload 2: Long-tail workload
* **Composition:** 1 × 512-token prompt and 31 × 16-token prompts (padded to `[32, 512]`)

| Case | Mean time | Mean throughput |
| :--- | :--- | :--- |
| **no-CB** | 9539.69 ms | 53.67 tok/s |
| **CB equal-width, batch 32** | 9668.58 ms | 52.96 tok/s |
| **aligned CB, batch 32** | 3711.05 ms | 137.97 tok/s |

> The aligned path was **2.57× faster** than no-CB for this long-tail workload by avoiding repeated prefill computation on padded rows. These measurements are workload-dependent and are not presented as a general CB speedup claim.

> The observed 2.57x improvement comes from avoiding padded prefill work in the long-tail prompt workload, not from cache trimming. The workload had trim_count=0, so it does not provide real-model trim overhead data. Trimming is covered by the synthetic regression test, while a real trim measurement would require a workload with staggered EOS or per-request token budgets.

### Notes
The tested workloads did not trigger real-model trimming:
```text
trim_count=0
trimmed_columns=0
trim_latency_ms=0.0
trim_bytes_moved=0
```

### Applicability and Limitations

| Workload | Expected behavior | Main reason |
| --- | --- | --- |
| Short prompts with similar lengths | May be slower | Scheduler, cache-management, and per-request prefill overhead dominate |
| Highly variable prompt lengths | Can improve throughput | Reduces padded prefill computation |
| Dynamic request arrivals | Can improve GPU utilization | Released slots can be refilled with pending requests |
| Staggered request completion or EOS retirement | Enables trimming benefits | Dead cache columns can be reclaimed |
| Fixed batch with similar-length requests | Standard static batching may be faster | It avoids continuous-batching overhead |
| No early request completion | No real trimming benefit | The trimming path is not exercised |

*BF16 token-exactness is not claimed; separate BF16 runs showed numerical token divergence.*

Related to #8462 and follow-up to #8368.


deepspeed/runtime/rollout/continuous_batching.py
```diff
@@ -52,9 +52,11 @@ class ContinuousBatchScheduler:
     All submitted requests share the scheduler's ``max_new_tokens`` budget;
     requests may still retire earlier when the model emits EOS.
     ``schedule`` performs admission/retirement without advancing tokens.
-    ``advance`` represents one decode step for every active request and also
-    retires requests whose token budget has been consumed. A caller may pass
-    explicit finished IDs when the model emits EOS before that budget.
+    ``admit_if`` can defer the FIFO head until the caller's cache layout can
```
delock, line 55: It occurs to me if the prompts are sorted with decendent order, the condition of `admit_if` will never happen.  Because we always have longer prompt in some slots to prevent the buffer being over trimmed for new (shorter) prompt to be filled in.  Consider remove admit_if for simplicity.

deepspeed/runtime/rollout/hybrid_engine_rollout.py
```diff
@@ -80,6 +80,7 @@ class HybridEngineRolloutConfig:
     use_graph_capture: bool = False
     enable_profiling: bool = False
     use_shared_prefill: bool = False
+    align_decode_fronts: bool = False
```
delock, line 83: Is it true that if `align_decode_fronts`  is false, the prompt always have to have same length?   Or we allow staggered decode fronts?   Current name sound like we support staggered decode fronts.

delock:
Hi @nathon-lee do you have performance of CB with this optimization vs no CB?   Could be performance of 32 different prompts rollout, then measure the rollout end2end performance.   This could show how much performance improvement CB can bring us, and could reveal real trim overhead with your optimization.

nathon-lee:
> Hi @nathon-lee do you have performance of CB with this optimization vs no CB? Could be performance of 32 different prompts rollout, then measure the rollout end2end performance. This could show how much performance improvement CB can bring us, and could reveal real trim overhead with your optimization.

Hi @delock  I did test this before, but unfortunately I didn’t save the results. I’ll run the 32-prompt E2E comparison again this weekend and share the numbers.

nathon-lee:
Hi, @delock 

### Modified
I removed the admission gate from the scheduler/rollout API and added a 32-request padded-prompt refill/trim regression with continuous_batch_size=8.

Cache trimming now preserves enough physical span for every prompt admitted in the current refill, so an admitted padded prompt cannot receive an empty cache slice. align_decode_fronts=False remains documented as the equal‑width
padded‑prompt baseline.

I also fixed the real Qwen prefill/decode mismatch for models whose cache‑class capability is unknown. Continuous prefill now obtains the first token and prompt KV through the model's standard greedy generation path, and subsequent continuous decode applies the model generation configuration's repetition penalty instead of using a bare argmax.

### Validated: uniform mixed‑prompt workload
On an NVIDIA RTX 4000 Ada (20 GB), PyTorch 2.9.1+cu128, CUDA 12.8, Transformers 4.57.1, and Qwen/Qwen2.5‑3B‑Instruct in FP32, I ran 32 mixed left‑padded requests with prompt shape `[32, 20]` and max_new_tokens=16. All outputs were token‑exact against the single‑request eager oracle:

| Case | Time | Throughput | Exact match |
| --- | ---: | ---: | --- |
| no‑CB | 1183.46 ms | 432.63 tok/s | Yes |
| CB equal‑width, batch 32 | 1418.99 ms | 360.82 tok/s | Yes |
| aligned CB, batch 32 | 2600.53 ms | 196.88 tok/s | Yes |
| dynamic CB, continuous_batch_size=8 | 4315.39 ms | 118.65 tok/s | Yes |

The dynamic run completed 32/32 non‑empty responses without an empty `torch.cat`; it took 60 decode steps. This workload did not trigger trimming (trim_count=0, trimmed_columns=0, trim_latency_ms=0.0, trim_bytes_moved=0).

These uniform short‑prompt measurements are not a CB speedup claim. There is no padded‑prefill reduction to offset scheduler, KV‑copy, mask/cache maintenance, and Python decode overhead. Aligned prefill is also currently performed per request, which adds forward‑call overhead for short prompts. The batch‑8 dynamic case has only eight active rows and four refill waves while all requests arrive at once, so it is not comparable to fixed batch‑32 utilization.

### Validated: aligned long‑tail workload
I also measured a deliberately long‑tail workload where aligned prefill has the expected benefit. The prompts were constructed from a repeated non‑special token ID: one 512‑token row and 31 16‑token rows, left‑padded to `[32, 512]`, with max_new_tokens=16. Both repetitions produced 512 generated tokens, 32/32 non‑empty responses, and token‑exact outputs against the single‑request eager oracle.

| Case | Mean time (2 runs) | Mean throughput | Per‑run throughput |
| --- | ---: | ---: | ---: |
| no‑CB | 9539.69 ms | 53.67 tok/s | 53.86, 53.49 |
| CB equal‑width, batch 32 | 9668.58 ms | 52.96 tok/s | 53.13, 52.78 |
| aligned CB, batch 32 | 3711.05 ms | 137.97 tok/s | 138.08, 137.86 |

Aligned CB is **2.57× faster (+157.1%)** than no‑CB for this long‑tail workload. The gain comes from avoiding repeated prefill computation on 31 padded rows; it is workload‑dependent, not a general CB speedup claim. This workload also did not trigger trimming: trim_count=0, trimmed_columns=0, trim_latency_ms=0.0, and trim_bytes_moved=0.

The focused scheduler/rollout suite passes 14 tests, and changed‑file pre‑commit checks pass.

### Pending validation
BF16 is not being presented as token‑exact correctness. Its previously observed token divergence remains a separate limitation and needs independent disclosure/validation.

deepspeed/runtime/rollout/hybrid_engine_rollout.py
nathon-lee: I agree the name could be slightly misleading since the alignment path is opt-in; I’ll consider clarifying the naming or documentation further.

nathon-lee: Thanks for pointing this out, @delock.

When `align_decode_fronts=False`, all requests must use the same padded prompt width. Their effective prompt lengths may still differ according to the attention masks, so the legacy path keeps staggered logical decode positions. It does not physically right-align or compact the decode fronts.

When `align_decode_fronts=True`, the rollout uses the effective prompt lengths to physically align decode fronts and trim dead cache columns.

I clarified this distinction in the configuration docstring and the inference documentation.

tests/unit/runtime/rollout/test_hybrid_engine_rollout.py
```diff
@@ -205,6 +208,79 @@ def forward(self, input_ids, attention_mask, past_key_values=None, use_cache=Tru
     assert (1, 3) in model.calls
 
 
+def test_aligned_continuous_generation_supports_mixed_effective_prompt_lengths():
```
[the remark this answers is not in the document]
nathon-lee: Hi @delock,

I completed the device-backed validation on an NVIDIA RTX 4000 Ada with Qwen/Qwen2.5-3B-Instruct, and the CB outputs matched the single-request eager oracle token by token. The commands and results are included in the
latest PR comment.

The Codex review asks for an added device-backed integration test. I could not find an existing model-checkpoint-dependent integration test pattern in this area of the repository.

Would the external validation be sufficient for this PR? If a committed model-dependent integration test is needed, I can prepare it as a follow-up PR.

nathon-lee:
I’ll construct an additional workload with staggered EOS and measure the real trimming overhead, including trim count, latency, and bytes moved.

nathon-lee:
## Performance, trim-overhead, and cache-capacity validation

The measurements now separate three effects:

1. **Aligned prefill** can improve long-tail prompt workloads by avoiding padded prefill work.
2. **Trimming is not a throughput optimization.** It introduces measurable KV/metadata copy overhead.
3. **Trimming provides a capacity benefit** for staggered-completion workloads: it can keep a continuous batch within a smaller static cache span when the equivalent untrimmed execution would exhaust that span.

#### 1. Aligned-prefill measurements

These runs used an NVIDIA RTX 4000 Ada (20 GB), FP32, PyTorch 2.9.1+cu128, CUDA 12.8, Transformers 4.57.1, and `Qwen/Qwen2.5-3B-Instruct`. All outputs were token-exact against the single-request eager oracle.

| Workload | no-CB | Equal-width CB | Aligned CB | Conclusion |
| --- | ---: | ---: | ---: | --- |
| 32 mixed left-padded requests, `[32, 20]`, `max_new_tokens=16` | 1183.46 ms / 432.63 tok/s | 1418.99 ms / 360.82 tok/s | 2600.53 ms / 196.88 tok/s | Short uniform prompts do not amortize aligned-prefill and scheduler/cache-management overhead. |
| 1 x 512-token prompt + 31 x 16-token prompts, padded to `[32, 512]`, `max_new_tokens=16` | 9539.69 ms / 53.67 tok/s | 9668.58 ms / 52.96 tok/s | 3711.05 ms / 137.97 tok/s | Aligned CB is **2.57x faster (+157.1%)** by avoiding padded prefill for the 31 short rows. |

Neither workload triggered trimming (`trim_count=0`), so the aligned-prefill improvement must not be attributed to trimming.

#### 2. Isolated trim-overhead measurement

For a controlled trim/no-trim comparison, I added `enable_cache_trimming` (default: `True`). Disabling it preserves aligned scheduling and decode-front layout while skipping left-prefix trimming.

This workload used an NVIDIA RTX A5000 (24 GB), BF16, Transformers 4.57.1, and `Qwen/Qwen2.5-0.5B-Instruct`. It has 32 left-padded requests with effective prompt lengths from 64 down to 33 and `continuous_batch_size=8`. A deterministic 2-5-token EOS schedule creates staggered completion and reliably exercises trimming. After one warm-up, three runs per case used the same 96-token cache capacity and produced token-exact outputs.

| Case | Mean E2E time | Mean throughput | Trim count | Bytes moved | Cumulative trim latency |
| --- | ---: | ---: | ---: | ---: | ---: |
| Trim enabled | 1157.62 ms | 96.86 tok/s | 12 | 109,188,480 | 23.68 ms |
| Trim disabled | 1087.39 ms | 103.00 tok/s | 0 | 0 | 0.00 ms |

With capacity held constant, trimming adds about **6.5%** E2E time for this deliberately trim-heavy workload. It moves about **8.68 MiB per trim** and costs about **1.97 ms per trim** on average.

#### 3. Capacity-reclamation measurement

I then used the same staggered-EOS pattern with `max_new_tokens=8` to test the actual capacity benefit.

| Case | Static cache span | KV cache + metadata allocation | Peak cache span | Result |
| --- | ---: | ---: | ---: | --- |
| Trim enabled | 72 | 7,082,560 bytes | 65 | Completed |
| Trim disabled | 72 | 7,082,560 bytes | - | `cache exhausted before active requests retired` |
| Trim disabled | 80 | 7,869,504 bytes | 80 | Completed |

The trim-enabled 72-token run and the untrimmed 80-token run are token-exact. For this workload, trimming reduces the required cache span by **8 tokens (10%)** and the preallocated KV-cache/metadata footprint by approximately **11.1%**. The trim-enabled run performed 12 trims, reclaimed 42 columns, moved 80,858,496 bytes, and recorded 25.52 ms cumulative trim latency.

### Conclusion

Aligned decode-front processing and cache trimming solve different problems:

- **Aligned prefill** can provide substantial performance gains for highly variable prompt lengths by removing padded-prefill work.
- **Trimming is a cache-span reclamation mechanism, not a throughput optimization.** It trades KV and attention-metadata copy overhead for the ability to continue staggered-EOS/refill workloads within a smaller bounded static cache span.
- When the untrimmed workload already fits in the available cache, disabling trim is faster in these measurements.
- When cache capacity is constrained, trim can make a workload complete that would otherwise exhaust the configured cache span.

The deterministic EOS schedule is intended to isolate and reproducibly exercise the trim path; these results are not presented as a claim about the natural EOS distribution of production traffic.

Validation: the trim-enabled and no-trim paths are covered by rollout unit tests, including output equivalence and trim telemetry. `tests/unit/runtime/rollout/test_hybrid_engine_rollout.py` passes 37 tests, and changed-file pre-commit checks pass.

delock:
@nathon-lee so in rollout case if we leave enough KV cache size then no trimming is better choice, this is the conclusion from trimming comparison data right?

nathon-lee:
@delock  I also agree to a large extent: if the KV cache can sufficiently cover peak demand, skipping trimming will result in faster performance. However, in scenarios with limited cache capacity or significant variations in request completion times, trimming still holds value for capacity recovery.

delock:
> @delock I also agree to a large extent: if the KV cache can sufficiently cover peak demand, skipping trimming will result in faster performance. However, in scenarios with limited cache capacity or significant variations in request completion times, trimming still holds value for capacity recovery.

Agree.  We should set trimming default as false.  If user use this path and hit OOM and have to reduce cache capacity, they will investigate and will find the trimming option.  On the other hand, if user never need to reduce cache capacity, user will bear trim overhead and not knowing it.  So rational action might be set trimming default as false.  Is it possible to direct user to turn on trimming or increase cache capacity, when cache capacity exthausted?

nathon-lee:
> > @delock I also agree to a large extent: if the KV cache can sufficiently cover peak demand, skipping trimming will result in faster performance. However, in scenarios with limited cache capacity or significant variations in request completion times, trimming still holds value for capacity recovery.
> 
> Agree. We should set trimming default as false. If user use this path and hit OOM and have to reduce cache capacity, they will investigate and will find the trimming option. On the other hand, if user never need to reduce cache capacity, user will bear trim overhead and not knowing it. So rational action might be set trimming default as false. Is it possible to direct user to turn on trimming or increase cache capacity, when cache capacity exthausted?

Agree. I will set cache trimming to disabled by default so users with sufficient KV-cache capacity do not pay the copy overhead.

On capacity exhaustion, I will make the failure actionable: the error will explicitly recommend either increasing the configured cache capacity, or enabling trimming for workloads where requests retire at different times and dead cache columns can be reclaimed.

I am also considering an opt-in auto mode: it would preserve the no-trim fast path and trim only when the next cache write would otherwise exhaust capacity and there is a reclaimable dead prefix. It would not try to recover after a real OOM, because trimming cannot help when no dead prefix exists. I’ll keep the initial change minimal unless we agree that auto belongs in this PR.

PR #8517, opened 2026-09-14
Restore memory-mapped datasets when unpickling

vineethsaivs:
Unpickling `MMapIndexedDataset` raises `TypeError`, preventing spawned DataLoader workers from reading it. Adding a default alone also silently re-enables a full-file warmup in every worker.

Save the dataset path and `skip_warmup` together, and restore both. Legacy path-only states retain the original default.

Test: `DS_ACCELERATOR=cpu PYTHONPATH=. python -m pytest tests/unit/runtime/test_data.py -k 'mmap or repeating_loader' -q`. Six tests pass, covering both warmup settings, signed/unsigned storage, legacy states and a real spawned DataLoader. The two skipped-warmup cases fail before the follow-up. Changed-file pre-commit passes. Apple M2 Pro CPU.


deepspeed/runtime/data_pipeline/data_sampling/indexed_dataset.py
```diff
@@ -491,7 +491,7 @@ def __getstate__(self):
     def __setstate__(self, state):
         self._do_init(state)
 
-    def _do_init(self, path, skip_warmup):
+    def _do_init(self, path, skip_warmup=False):
```
ebarkhordar, line 494: I ran this against `344e22bb10e7f97c339040e0c4e6729d427d4c2c` in a clean `python:3.11-slim` container (torch 2.8.0+cpu, `requirements/requirements.txt`). The round trip works and the new tests pin what they say.

The default fixes the crash but drops the flag. `__getstate__` on line 488 returns `self._path` alone, so a dataset opened with `skip_warmup=True` comes back from the round trip with warmup on. I traced the real `_warmup_mmap_file` (the wrapper calls through, it does not replace it) on a 4.9 MB corpus:

```
construct with skip_warmup=True  -> warmup calls: []
pickle round trip of that object -> warmup calls: [('corpus.idx', 240050), ('corpus.bin', 5120000)]
__getstate__ returns: '/tmp/ds/corpus'
round trip correct?  len 20000 == 20000, item equal True
```

That matters here because every construction site in the tree passes `skip_warmup=True`: `data_sampler.py:95,100,221,250,348`, `data_analyzer.py:238,252,354,403,421,422`, `variable_batch_size_and_lr.py:398`. So on the spawn path your second test covers, each worker sweeps the whole `.bin` sequentially in 100 MB chunks, which is the cost those call sites set the flag to avoid. It is quiet apart from two `warming up ...` prints per worker.

Carrying the flag is about the same size as the default:

    def __getstate__(self):
        return self._path, self._skip_warmup

    def __setstate__(self, state):
        self._do_init(*state)

plus `self._skip_warmup = skip_warmup` in `_do_init`. Nothing else in the tree pickles this class, so the state shape stays contained to this file and your new test.

vineethsaivs: Fixed in 44e5b52a. Pickling now preserves `skip_warmup`; legacy path-only states still load. All six focused tests pass, including a spawned DataLoader.

ebarkhordar: Confirmed at 44e5b52a in a clean container (torch 2.8.0+cpu), tracing the real `_warmup_mmap_file` through a call-through wrapper rather than asserting on the flag.

```
344e22bb  PICKLE ROUND TRIP  warmup calls: ['ds.idx', 'ds.bin']
44e5b52a  PICKLE ROUND TRIP  warmup calls: []
44e5b52a  LEGACY str state   warmup calls: ['ds.idx', 'ds.bin'] | data ok: True
```

The legacy branch is the part I was least sure of and it behaves the way old pickles did, so nothing written by an earlier version changes meaning. `_do_init` still has only the two call sites, `__init__` and `__setstate__`, so the new field cannot be bypassed.

I did not run the wider data pipeline suite and nothing here was on GPU.

PR #8524, opened 2026-09-15
fix(zenflow): use optimizer_z3.device instead of hardcoded cuda

li-lizhe:
## Summary

The ZenFlow optimizer module in `engine_stage3.py` hardcodes `device='cuda'` in 4 locations when creating tensors (`grad_sum_buffer`, `selected_indices`, `sum_buffer`, and `critic_sum_buffer`). This causes a crash on non-CUDA accelerators (NPU, XPU, HPU, etc.) with `Torch not compiled with CUDA enabled`.

## Root cause

```python
optimizer_z3.grad_sum_buffer = torch.zeros(total_chunk_size, dtype=optimizer_z3.dtype, device='cuda')
```

The `DeepSpeedZeroOptimizer_Stage3` already has a `device` attribute set via `get_accelerator().current_device_name()`, which correctly reports the active accelerator. DeepSpeed has a device-agnostic accelerator abstraction (`deepspeed.accelerator.get_accelerator()`) used throughout the codebase.

## Fix

Replace all 4 hardcoded `device='cuda'` with `device=optimizer_z3.device`.

```diff
-    optimizer_z3.grad_sum_buffer = torch.zeros(total_chunk_size, dtype=optimizer_z3.dtype, device='cuda')
+    optimizer_z3.grad_sum_buffer = torch.zeros(total_chunk_size, dtype=optimizer_z3.dtype, device=optimizer_z3.device)
```

## Testing

On a non-CUDA accelerator (Ascend NPU), the hardcoded `device='cuda'` raises:
`RuntimeError: Torch not compiled with CUDA enabled`

With this fix, tensor creation uses `optimizer_z3.device` which resolves to the active accelerator, and ZenFlow selection logic runs correctly.

On CUDA, `optimizer_z3.device` is `cuda` (via `get_accelerator().current_device_name()`), so behavior is unchanged.

## Related

Consistent with DeepSpeed's existing device-agnostic design via `get_accelerator()`.


pengdurice (approved):

li-lizhe:
Thanks for the review and for updating the branch, @pengdurice.

Follow-up pushed (`36f1c790`) that addresses the P1 point from the automated review:

- `grad_sum_buffer`, `param.selected_indices`, `sum_buffer` and `critic_sum_buffer` now use `get_accelerator().current_device_name()` instead of `optimizer_z3.device`. Those buffers are handed to `dist.all_gather` on the data-parallel process group, and `stage3.py` sets `self.device` to `OffloadDeviceEnum.cpu` whenever ZeRO optimizer offload is enabled, so the previous expression would have put them on the host in the offload configuration used by `tests/unit/runtime/zenflow/test_zf.py`.
- Wrapped one long tensor-creation line so the yapf formatting check passes (`COLUMN_LIMIT = 119`).

Verification: `yapf --style .style.yapf -d` is clean and the file compiles. I did not run the ZeRO-3 ZenFlow offload path on hardware - `tests/unit/runtime/zenflow/test_zf.py` returns early when the accelerator device is CPU, and I have no CUDA/NPU accelerator in this environment - so the multi-process `all_gather` path is untested on my side.

li-lizhe:
One note: the `pull_request` workflow runs for this commit are sitting in `action_required` (fork PR), so the checks will only start once a maintainer approves the run.

PR #8525, opened 2026-09-15
fix(data_pipeline): default device to active accelerator in DistributedDataAnalyzer

li-lizhe:
## Summary

`DistributedDataAnalyzer.__init__` hardcodes `device='cuda'` as the default value. When users create an analyzer without passing an explicit device (common in data pipelines that use `LOCAL_RANK` or other auto-detection), the default `'cuda'` causes tensor creation on a device that may not be available.

On non-CUDA accelerators (NPU, XPU, HPU, etc.):
```python
RuntimeError: Torch not compiled with CUDA enabled
```

## Fix

Set the default to `None` and resolve via `get_accelerator().device_name()`:

```python
- device='cuda',
+ device=None,
...
- self.device = device
+ self.device = device if device is not None else get_accelerator().device_name()
```

`get_accelerator()` is DeepSpeed's existing device-agnostic abstraction that correctly reports the active accelerator (CUDA, NPU, XPU, HPU, etc).

## Impact

- Callers that pass an explicit `device` argument are **unaffected**
- On CUDA systems, `get_accelerator().device_name()` returns `'cuda'` — behavior is **unchanged**
- On NPU/XPU/HPU, the analyzer now auto-detects the correct device

## Testing

On CUDA: `device=None` resolves to `'cuda'` via `get_accelerator()`, identical to previous behavior.
On Ascend NPU: before fix → crash (`Torch not compiled with CUDA enabled`); after fix → analyzer runs on the active NPU device.


li-lizhe:
Also pushed the same follow-up to this branch (`d9810b0e`). This branch carries the `engine_stage3.py` commit from #8524, so it had the same P1 from the automated review; the buffers now follow `get_accelerator().current_device_name()` instead of `optimizer_z3.device`. Nothing else changed in the data_pipeline fix. Ready for review when you have a moment.

PR #8528, opened 2026-09-15
Keep tensor scheduler state independent of optimizer updates

vineethsaivs:
Using a live optimizer LR tensor as a schedule boundary makes later rates drift. Returning that same tensor from `get_last_lr()` also changes values callers already captured.

Clone tensor-valued bounds and the values saved after each scheduler step. Captured rates remain stable, and modifying them cannot change the optimizer LR.

Test: `DS_ACCELERATOR=cpu PYTHONPATH=. python -m pytest tests/unit/runtime/test_lr_schedulers.py -k 'tensor_snapshot or tensor_lr_bound' -q`. Thirteen CPU tests pass, including real SGD updates. The five new snapshot cases fail before this follow-up. Changed-file pre-commit passes. Apple M2 Pro CPU; no distributed run.


ebarkhordar (commented):
I ran this at `a720309e` in a clean container (`python:3.12-slim`, torch 2.14.0+cpu, `pip install --no-deps -e .`). The bound fix does what it says. There is one more alias of the same shape just below it that the PR leaves in place.

`update_lr` returns `[group['lr'] for group in param_groups]` at lr_schedules.py:283, and all four schedulers store that as `_last_lr` and hand it straight back from `get_last_lr()`. On the float path :282 rebinds, so a caller gets a snapshot. On the tensor path :280 fills the live optimizer tensor in place, so the list the caller is holding moves under them:

```
float lr  : captured get_last_lr() = [0.1] -> after one more step() the same object reads [0.1]
tensor lr : captured get_last_lr() = [0.1] -> after one more step() the same object reads [0.2]
```

That is WarmupLR with `warmup_min_lr=lr`, `warmup_max_lr=0.3`, `warmup_num_steps=2`, captured after `step(0)` and read again after `step(1)`.

It is the same float versus tensor asymmetry your `_format_param` change closes for bounds, so it looks in scope here: a `.clone()` at :283, or in `get_last_lr`.

On where to put it, the only in-tree consumer is `variable_batch_size_and_lr.py:288`, which restores `get_last_lr()` onto the param groups before stepping the base scheduler. It does not look sensitive either way, since `scale_lr` at :154 rebinds rather than filling in place, so its saved list holds the unscaled value under both behaviours. I read that path, I did not run it.

Not checked: GPU, and I did not run the wider suite.

vineethsaivs:
Fixed in 3edc4af5. `get_last_lr()` now keeps a snapshot across later steps, and modifying it cannot change the optimizer LR. All 13 focused CPU tests pass.

ebarkhordar:
Confirmed at 3edc4af5 in a clean container (torch 2.8.0+cpu). I checked the thing most likely to break, which is whether cloning the reported value costs you the in-place update the tensor path exists for. It does not: the optimizer still holds the caller's own tensor and still fills it in place.

```
a720309e  WarmupLR  captured_is_snapshot=False  user_tensor_still_bound=True  mutate_reported_safe=False
3edc4af5  WarmupLR  captured_is_snapshot=True   user_tensor_still_bound=True  mutate_reported_safe=True
```

Same on WarmupDecayLR, LRRangeTest, OneCycle and WarmupCosineLR, and the float path still reports a plain float.

I also re-read `variable_batch_size_and_lr.py` step(), where `get_last_lr()` is written back onto the param groups: it rebinds through `scale_lr` on the next loop either way, so the clone does not change what that path ends up with. I read it and did not run it, and nothing here was on GPU.

vineethsaivs:
Thanks for running it down, the in-place path was the part I was least sure survived the clone.

PR #8530, opened 2026-09-15
Fix ZeRO-3 checkpoint conversion with --debug

vineethsaivs:
The checkpoint converter crashes with `--debug` on ZeRO-3 checkpoints.

`_zero3_merge_trainable_params` receives each rank's optimizer groups as a list and the debug block treats it as a tensor, reading `.shape` and `.numel()` off the list. Print the per-group shapes instead, the way the ZeRO-2 debug block above it already does, and drop the `avail_numel` reassignment: it recomputed, wrongly, a value the top of the function already has.

Tests: `DS_ACCELERATOR=cpu python -m pytest tests/unit/checkpoint/test_convert_checkpoint.py -k 'not TestCheckpointConvert' -q`. Debug export fails before. Five converter tests pass after. A separate real ZeRO-3 Adam step with two parameter groups, checkpoint save and debug export produces matching weights. Changed-file pre-commit passes. Apple M2 Pro CPU, PyTorch 2.10, one-rank Gloo; no GPU run.


vineethsaivs:
Could a maintainer approve CI for this PR? The focused regression tests and repository lint checks pass locally.

alanhuangyoo (commented):
Ran your test on CPU both ways: master dies at the same `AttributeError: 'list' object has no attribute 'shape'`, passes on this branch.

One thing worth adding to the description: the `avail_numel` line you removed wasn't only crashing. It replaced the correct group sum from the top of the function with `fp32_flat_groups[0].numel()`, and `avail_numel` is what the `offset != avail_numel` check at the end compares against, so with `--debug` that check was looking at the wrong number too.

The zero-2 debug block at :317 already prints shapes per group, so this just brings stage 3 in line.

vineethsaivs:
Good catch, added to the description: the shadowed `avail_numel` also fed the `offset != avail_numel` sanity check at the end, so under `--debug` that check was comparing against the wrong number. Thanks for running it.

vineethsaivs:
Correcting myself: that reassignment never actually ran. The `.shape` print above it is the first statement in the `debug` block and raises the same AttributeError, so the final `offset != avail_numel` check always used the correct sum from the top of the function. It was a stale line, not a wrong comparison. I have taken that claim back out of the description.

alanhuangyoo:
You're right, the print dies first so that line never ran. My mistake, thanks for checking.

PR #8283, opened 2026-08-20
Register native pinned host memory with CUDA for GPU DMA

sfc-gh-truwase:
## Summary
- Native pin (`posix_memalign` + `mlock`) is device-independent and sufficient for DeepNVMe, but the GPU DMA engine does not treat `mlock` pages as pinned. After allocation, optionally call `cudaHostRegister` (and `cudaHostUnregister` before free) via accelerator hooks so `copy_(..., non_blocking=True)` can DMA.
- Default-on via `DS_PIN_MEMORY_REGISTER_DEVICE` (`1`/`true`/`yes`/`on`; `0` keeps mlock-only). Registration failure logs once and continues with mlock. CPU accelerators no-op.
- Adds `benchmarks/pin_memory/h2d_d2h_bench.py` (torch vs native-unregistered vs native-registered).

## H200 H2D/D2H (1 GPU)
Host `tunji-h200-n1g2-ds-pin-0`, NVIDIA H200, `--sizes-mib 4 64 256 --warmup 10 --iters 50`, autorun `job-20260820T155549Z`.

| Arm | Size (MiB) | H2D (GB/s) | D2H (GB/s) | `torch.is_pinned` |
|-----|------------|------------|------------|-------------------|
| torch | 4 | 53.20 | 52.96 | True |
| torch | 64 | 55.32 | 52.49 | True |
| torch | 256 | 55.41 | 52.30 | True |
| native-unregistered | 4 | 16.95 | 14.05 | False |
| native-unregistered | 64 | 16.45 | 15.45 | False |
| native-unregistered | 256 | 9.98 | 16.42 | False |
| native-registered | 4 | 53.16 | 52.78 | True |
| native-registered | 64 | 55.29 | 54.79 | True |
| native-registered | 256 | 55.40 | 54.94 | True |

Registered native matches torch bandwidth (~53–55 GB/s). `mlock` without `cudaHostRegister` stays in the ~10–17 GB/s pageable range and reports `torch.is_pinned=False`.

## Test plan
- [x] `pre-commit run --files` on touched paths
- [x] `pytest tests/unit/v1/pin_memory/test_pin_memory.py` **16 passed** (register on/off, CPU no-op, CUDA cudart mock, register failure keeps mlock, GC unregister, invalid env)
- [x] GPU autorun `job-20260820T155549Z`: pin_memory + accelerator tests **25 passed**, H2D/D2H bench exit 0

Made with [Cursor](https://cursor.com)

PR #8532, opened 2026-09-16
Register native pinned host memory with the CANN runtime for NPU

### LINKED ISSUE #8531: [REQUEST]  Register native pinned host memory with device runtime for NPU (follow-up to #8283 and #8315)
DeepSpeed's native pinned-memory backend (`DS_PIN_MEMORY_BACKEND=native`) page-locks
host buffers with `posix_memalign` + `mlock`, and since #8283 it can additionally
register each buffer with the device runtime so asynchronous copies can use the DMA
engine. That change made a large difference where it shipped: on an H200, H2D
bandwidth went from ~10-17 GB/s (mlock alone) to ~53-55 GB/s after registration.

The registration capability, however, is a per-accelerator hook, and today only the
CUDA accelerator implements it. Every other platform falls through to an empty
default: with `DS_PIN_MEMORY_REGISTER_DEVICE=1` on an Ascend NPU, the hook returns
`False` silently, the buffer remains mlock-only, and the native backend performs
like pageable memory. This is not a hypothetical gap — on an Ascend 910B4
(CANN 8.5.0) registering the same native buffer lifts 4 MiB H2D from ~10 GB/s to
~23 GB/s, and 64 MiB registered copies hold ~23.4 GB/s while mlock-only
fluctuates between ~13 and ~22.5 GB/s.

The XPU platform faces the identical issue, already filed as #8315. That NPU and
XPU users are independently hitting the same blind spot makes the point: #8283's
benefit is gated per platform, and the non-CUDA platforms that run an increasing
share of real training jobs (Huawei Ascend in particular) silently lose 2x+ on
every H2D copy until someone implements the two hooks for them. This issue covers
the NPU side; #8315 covers XPU.

Implement `register_host_memory` / `unregister_host_memory` on `NPU_Accelerator`
(`accelerator/npu_accelerator.py`) via the NPU runtime binding torch_npu ships:

- `torch.npu.npurt()` returns the NPU runtime-API module with
  `npuHostRegister(addr, size, flags)` / `npuHostUnregister(addr)`, thin wrappers
  over `aclrtHostRegisterV2` / `aclrtHostUnregister`. It initializes the runtime if
  needed and fails with an explicit `RuntimeError` when the build does not include
  it.
- Failures should not disturb training: if torch_npu is older than 2.9.0, or the
  build has no npurt, or the registration call returns a non-zero code, log a
  warning once and continue with plain mlock memory (the existing
  `DS_PIN_MEMORY_REGISTER_DEVICE` switch keeps acting as the opt-out).
- Extend `tests/unit/v1/pin_memory/test_pin_memory.py` with NPU cases mirroring
  the CUDA cudart-mock test, plus coverage for the availability fallbacks.
- The before/after bandwidth can be measured on NPU with
  `benchmarks/pin_memory/h2d_d2h_bench.py` once it is made accelerator-agnostic.

- Leave NPU as-is: native pin stays mlock-only on Ascend, which is functionally
  correct but pays the full pageable-copy penalty this feature was created to
  remove.
- Give the NPU native backend an `aclrtMallocHost`-style allocator: true pinned
  allocation instead of post-hoc registration, but it would replace the
  device-independent allocator wholesale and is a much bigger change than the
  two-hook contract #8283 established.

**Additional context**
- Hook definitions and call sites: #8283 introduced the hooks (empty defaults in
  `accelerator/abstract_accelerator.py`, CUDA implementation via
  `cudaHostRegister` / `cudaHostUnregister`) and invokes them from
  `NativePinnedMemory.pin` / `_release` in `deepspeed/utils/pin_memory.py`.
- XPU counterpart: #8315, with an implementation attempt in PR #8322 (blocked on
  oneAPI 2026.2). Resolving both follow-ups completes the picture for the
  non-CUDA accelerators DeepSpeed supports.
- Validation environment: CANN 8.5.0, torch 2.10.0+cpu, torch_npu 2.10.0,
  Ascend 910B4 (aarch64).

VenusTZZ:
Fixes #8531.

Follow-up to #8283; XPU counterpart tracked in #8315 (attempted in #8322).

## Summary

- Implement `register_host_memory` / `unregister_host_memory` on `NPU_Accelerator`via the NPU runtime API torch_npu ships: `torch.npu.npurt()` returns the runtime-API module with `npuHostRegister` / `npuHostUnregister`, thin wrappers over `aclrtHostRegisterV2` / `aclrtHostUnregister`. Previously `NPU_Accelerator` inherited the base-class no-ops, so `DS_PIN_MEMORY_BACKEND=native` buffers stayed mlock-only on Ascend NPUs and asynchronous copies ran at pageable bandwidth.
- Degrades to "log once and continue with mlock" on every failure path and keeps honoring the existing `DS_PIN_MEMORY_REGISTER_DEVICE` opt-out.
- Extends `tests/unit/v1/pin_memory/test_pin_memory.py` with NPU cases mirroring the CUDA cudart-mock tests, including the availability fallbacks.
- Makes `benchmarks/pin_memory/h2d_d2h_bench.py` accelerator-agnostic (it hardcoded a CUDA check and torch's `pin_memory=True` fast path) so the before/after bandwidth could be measured on NPU.
- Documents the NPU registration path in `docs/code-docs/source/memory.rst`.

## Implementation notes

**The npurt binding.** torch_npu exposes `torch.npu.npurt()`, which initializes the runtime if needed and returns the runtime-API module. Using it avoids binding `libascendcl.so` ourselves and raises a clear `RuntimeError` when the build lacks the module. Two defenses gate the hook, both falling back to mlock-only with a one-time warning:

- no `torch.npu.npurt` attribute (the hook checks for its presence),
- `npurt()` failing to initialize the runtime.

**Flag choice: `ACL_HOST_REG_MAPPED`.** The `flags` parameter of `npuHostRegister` is the CANN `ACL_HOST_REG_*` bitmask (per the npurt docstring and `acl_rt.h`): `MAPPED` (0x2) page-locks the range and adds a device mapping, `PINNED` (0x10000000) only page-locks. Measured through the bench methodology at 64 MiB (10 process-level A/B runs): `PINNED`-only registrations fell back to mlock-only H2D speed (~8-9 GB/s) in 5/5 runs, while `MAPPED` kept ~23 GB/s (5/6). `MAPPED` additionally requires 4K-aligned addresses (a 64B-offset probe fails with `ACL_ERROR_RT_DRV_INTERNAL_ERROR`, 507899), which DeepSpeed's native allocator guarantees (`posix_memalign(..., sysconf(_SC_PAGESIZE), ...)`); sizes need no page alignment (probed 400 B to 1 MiB+3, rc=0). As a guard against unaligned callers, the hook extends the range down to the page boundary, padding the size accordingly, instead of relying on the driver's opaque internal error.

**Failure semantics.** A non-zero return code from `npuHostRegister` logs once and returns `False`, so `NativePinnedMemory` keeps the buffer mlock-only and does not track the address. A failed `npuHostUnregister` raises (like the CUDA hook), which makes `NativePinnedMemory` keep the allocation alive so the driver never holds a registration for pages later reused by `malloc`.

## Bandwidth on Ascend 910B4

`benchmarks/pin_memory/h2d_d2h_bench.py`, one full run of the #8283 methodology (`--sizes-mib 4 64 --warmup 10 --iters 50`), CANN 8.5.0, torch 2.10.0+cpu, torch_npu 2.10.0, Ascend 910B4 (aarch64), shared box:

| Arm | Size (MiB) | H2D (GB/s) | D2H (GB/s) | torch.is_pinned |
|-----|------------|------------|------------|-----------------|
| torch | 4 | 10.4 | 21.1 | True |
| torch | 64 | 16.5 | 21.0 | True |
| native-unregistered | 4 | 9.7 | 9.4 | False |
| native-unregistered | 64 | 21.1 | 19.6 | False |
| native-registered | 4 | **22.7** | 20.6 | True |
| native-registered | 64 | **23.4** | 20.9 | True |

After registration, `torch.is_pinned()` on a `torch.frombuffer` view of the range returns `True` — the torch_npu runtime recognizes the registered range — so the DMA path is confirmed to engage, not just inferred from the numbers.

Registration lifts H2D bandwidth from ~9.7 to ~22.7 GB/s at 4 MiB (2.3x) and holds ~23.4 GB/s at 64 MiB, with D2H improving as well (9.4 → 20.6 and 19.6 → 20.9 GB/s). Registered copies match or exceed the torch `pin_memory()` arm at every size.

## Test plan

- [x] `pre-commit run --files` on touched paths — yapf, flake8, check-torchdist, check-torchcuda, check-license, codespell all pass
- [x] `pytest tests/unit/v1/pin_memory/test_pin_memory.py` — **24 passed** on Ascend 910B4 (18 existing incl. the CUDA cudart-mock tests + 6 new NPU cases: npurt register/unregister calls, unaligned address extended to the page boundary, non-zero rc degrades to mlock, unregister failure raises, missing npurt is a no-op, availability gate table)
- [x] `benchmarks/pin_memory/h2d_d2h_bench.py` on Ascend 910B4 — all three arms run to completion and exit 0 at 4/64 MiB

## Known limitation

Some torch_npu builds may ship without the npurt module. The `hasattr` check degrades to mlock-only with a one-time warning, so registration is only available when the binding is present.


accelerator/npu_accelerator.py
```diff
@@ -13,6 +13,49 @@
...
+
+    torch.npu.npurt() returns the runtime-API module exposing
+    npuHostRegister/npuHostUnregister; the binding is maintained with
+    torch_npu and initializes the runtime itself. Host registration is
+    gated on torch_npu >= 2.9.0, the minimum supported version.
+    """
+    version = _torch_npu_version()
+    if version is None:
+        return None, "unable to determine the installed torch_npu version"
+    if version < _MIN_TORCH_NPU_REGISTER_VERSION:
```
hipudding: Do we need to check torch_npu' version?  I think check npurt is enough.

accelerator/npu_accelerator.py
```diff
@@ -152,6 +195,43 @@ def total_memory(self, device_index=None):
     def available_memory(self, device_index=None):
         return self.total_memory(device_index) - self.memory_allocated(device_index)
 
+    # Host memory registration
+    def register_host_memory(self, address, num_bytes):
+        # Register natively pinned (posix_memalign + mlock) host memory with the
+        # ACL runtime so torch's async copies can use the DMA engine. npurt
+        # initializes the runtime itself, so no set_device ordering is needed.
+        if address % 4096:
```
hipudding: We can align address to 4k, remember ping memory always contains all the buffer, and add num_bytes after align.

accelerator/npu_accelerator.py
VenusTZZ: Agreed — the version gate could reject builds that actually ship npurt, and the `hasattr` check alone covers the case we care about. Removed it in b7e6784d; the hook now falls back based only on whether `torch.npu.npurt` is present (plus `npurt()` initializing successfully).

accelerator/npu_accelerator.py
VenusTZZ: Agreed — Done in bb2e5967 — the hook rounds the address down to the 4K page boundary and pads num_bytes by the same offset, so the registered range always covers the original buffer exactly. An already-aligned address passes through unchanged (offset 0), and unregister_host_memory applies the same rounding so both sides stay consistent.

accelerator/npu_accelerator.py
```diff
@@ -152,6 +179,50 @@ def total_memory(self, device_index=None):
...
+        # ACL runtime so torch's async copies can use the DMA engine. npurt
+        # initializes the runtime itself, so no set_device ordering is needed.
+        # MAPPED registration requires 4K-aligned addresses (per the
+        # aclrtHostRegisterV2 API reference cited on ACL_HOST_REG_MAPPED). The
+        # native allocator always yields aligned addresses; if a caller passes
+        # an unaligned one, extend the range down to the page boundary so
+        # registration still succeeds instead of failing on the driver's opaque
+        # internal error. An already-aligned address passes through unchanged
+        # (offset 0), and unregister_host_memory rounds down identically.
+        offset = address % 4096
```
hipudding: It's better to make align as a function, register and unregister function can call align func. Avoid inconsistent

tests/unit/v1/pin_memory/test_pin_memory.py
```diff
@@ -284,3 +286,115 @@ def unregister_host_memory(self, address):
...
+    accelerator = NPU_Accelerator.__new__(NPU_Accelerator)
+
+    # 4096: page-aligned addresses register without any range extension.
+    assert accelerator.register_host_memory(4096, 4096) is True
+    accelerator.unregister_host_memory(4096)
+    assert registered == [(4096, 4096, npu_accelerator.ACL_HOST_REG_MAPPED)]
+    assert unregistered == [4096]
+
+
+def test_npu_unaligned_address_is_extended_to_page_boundary(monkeypatch):
```
hipudding: Should test unaligned address and aligned address.

hipudding (commented):
Most of these changes already very good, though there are a few minor points I suggest revising.

tests/unit/v1/pin_memory/test_pin_memory.py
VenusTZZ: Done in 8ba87d19 — the two tests were consolidated into one parametrized case (`test_npu_register_aligns_to_page_boundary`) covering both addresses: the page-aligned address (4096) registers unchanged, the unaligned one (4096+1234) is extended down to the boundary with a matching size pad, and unregister rounds down identically in both cases.

accelerator/npu_accelerator.py
VenusTZZ: Thanks! Done in 8ba87d19 — extracted `_align_to_page_boundary(address)` and both hooks now call it: `register_host_memory` uses the aligned address plus the returned offset for the size pad, and `unregister_host_memory` passes the aligned address back to the driver. The rounding now lives in one place, so the two can't drift apart.

tests/unit/v1/pin_memory/test_pin_memory.py
```diff
@@ -284,3 +286,100 @@ def unregister_host_memory(self, address):
...
+    monkeypatch.setattr(npu_accelerator, "_npu_host_copy_funcs", lambda: (None, "test"))
+    accelerator = NPU_Accelerator.__new__(NPU_Accelerator)
+
+    assert accelerator.register_host_memory(4096, 4096) is False
+    assert accelerator.unregister_host_memory(4096) is None
+
+
+def test_npu_host_copy_lookup_gates(monkeypatch):
+    # A build lacking npurt must not resolve, with a reason saying so.
+    monkeypatch.delattr(torch.npu, "npurt", raising=False)
```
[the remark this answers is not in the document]
VenusTZZ: Done in 91d7f015 — the test now installs a stand-in torch.npu via monkeypatch before touching npurt, so the gate logic also runs on builds without the backend.

VenusTZZ:
Hi @delock, this PR follows up on #8283 by implementing the register_host_memory / unregister_host_memory hooks for the NPU accelerator (the XPU counterpart is tracked in #8315). Would you be able to take a look when you have a chance? Thanks!

docs/code-docs/source/memory.rst
```diff
@@ -413,8 +413,10 @@ Native device registration
 
 Native allocations are device-independent ``mlock`` buffers. On CUDA systems,
 DeepSpeed additionally calls ``cudaHostRegister`` so PyTorch can use them for
-asynchronous H2D/D2H DMA. Device registration is enabled by default and can be
-disabled for comparison or debugging:
+asynchronous H2D/D2H DMA. On Ascend NPU systems, DeepSpeed additionally calls
```
delock: instead of enumerate what each accelerator do, its better to generize the original CUDA behavior description to make it work for more general concept 'accelerator'.

tests/unit/v1/pin_memory/test_pin_memory.py
```diff
@@ -284,3 +286,106 @@ def unregister_host_memory(self, address):
...
     assert begin not in native_pins._device_registered
+
+
+@pytest.mark.parametrize(
+    "address, expected_address, expected_bytes",
+    [
+        (4096, 4096, 4096),  # page-aligned: registers unchanged
+        (4096 + 1234, 4096, 4096 + 1234),  # unaligned: extended down with a matching size pad
+    ])
+def test_npu_register_aligns_to_page_boundary(monkeypatch, address, expected_address, expected_bytes):
```
delock: page alignment requirement may not be specific to NPU.  There should be an interface defined in abstract interface to query pin_memory alignments (default is 1 and define as 4096 for NPU).  Then this function can test against this alignment if not 1).  In that way this test could be more generic

docs/code-docs/source/memory.rst
VenusTZZ: Thanks @delock, agreed — the per-device enumeration doesn't scale. I've rewritten the paragraph to be 
device-agnostic: native buffers are registered through the accelerator's host-memory registration hook, and accelerators declare their required alignment via pin_memory_alignment. The cudaHostRegister / aclrtHostRegisterV2 specifics are gone. See bca3d15b (docs/code-docs/source/memory.rst).

tests/unit/v1/pin_memory/test_pin_memory.py
VenusTZZ: Good point — the alignment requirement isn't inherently NPU-specific. I've added 
pin_memory_alignment() to the abstract accelerator interface (default 1 = no requirement; NPU declares 4096), and moved the round-down-and-pad logic out of the NPU accelerator into NativePinnedMemory, which applies the declared alignment. The test is now generic: parametrized over the alignment (pass-through when 1, aligned when > 1) and driven through the public pin/unpin flow. The NPU 4096 declaration itself is pinned by a small test, and all pin-memory/accelerator unit tests pass with pre-commit clean. See ff9e91b9 (accelerator/abstract_accelerator.py, accelerator/npu_accelerator.py, deepspeed/utils/pin_memory.py, tests/unit/v1/pin_memory/test_pin_memory.py).

VenusTZZ: The NPU host-copy lookup test is now guarded with skipif and is skipped where torch_npu is absent (stubbing torch.npu proved unreliable across torch versions and broke the CPU CI), see bba5a0f .The fix is on the npu-register-host-memory branch. Happy to move these to a vendor directory instead if that's the direction the team prefers.

delock (approved):

Issue #8173, opened 2026-07-24
[RFC] Support sharding LM heads and adopting Online Softmax


jinyouzhi:
## Motivation

Online softmax and in-place computation could significantly save memory, especially for OPSD teacher TP sharding.

- Extend AutoTP capabilities by integrating Liger Kernel (sharding LM heads and adopting Online Softmax) #8104 

## Keypoints

- Support the sharding LM heads with `gather_output=False` of ColumnParallelLinear
- Enabling parallel CE loss feature (integrating Liger kernel)
- Profiling the GPU footprint to evaluate 
- Unify the untied and tied path for LM heads

## Revised Roadmap

| PR | Scope | Status |
|---|---|---|
| #8146 | Enable `gather_output=True` for column-parallel `lm_head` | Merged |
| #8185 | Uneven sharding + universal checkpoint; remove uneven-dim rowwise fallback | Merged |
| PR-A (#8302) | Lift hardcoded `name == "lm_head"` routing; training defaults to column parallelism; reject unsupported row-parallel training | Merged |
| PR-B (#8309) | `gather_output=False` + vocab-parallel cross entropy, pure PyTorch reference implementation | WIP |
| PR-C | `VocabParallelEmbedding`; unify tied and untied `lm_head` paths; fix #8290 | To do |
| PR-D | Optional Liger fused CE acceleration + memory profiling | To do |
| PR-E | Implement row-parallel `lm_head` training, including input slicing and backward `dx` all-reduce | To do |

## Proposed PR Sequencing

```
#8146 ──► #8185 ──┬─► PR-A  Lift hardcoded routing; independently mergeable
                  │
                  └─► PR-B  gather_output=False + vocab-parallel CE
                        ├─► PR-C  VocabParallelEmbedding / tied-untied unification
                        └─► PR-D  Liger integration + profiling

PR-A ──► PR-E  Row-parallel lm_head training
```

PR-A is small and low-risk, and has no code-level dependency on `gather_output`;
it can be submitted against master in parallel. PR-B is the core of this RFC —
we suggest landing the pure PyTorch implementation first so that numerical
correctness can be reviewed independently, and deferring the Liger kernel
integration to PR-D.

jinyouzhi:
[...]

PR #8538, opened 2026-09-16
[AutoTP] Support eager row-parallel output head training

0z5a:
## Summary

Implements PR-E from #8173: eager training for explicitly row-parallel, untied `lm_head` / `embed_out` output heads. This PR is independent of the vocabulary-parallel CE work in PR-B #8309 and contains no PR-D changes.

- Slice replicated activations along the last dimension using the frozen weight-shard geometry, including uneven hidden dimensions.
- Reuse the existing row-parallel forward output reduction and replicated bias behavior.
- Reconstruct and all-reduce the complete input gradient in backward so preceding replicated layers receive the correct gradient.
- Preserve the original parameter identity, existing inference specialization, and default gathered-column training behavior.
- Reject tied output weights at TP > 1, reshaped/non-input-dimension partitions, and deferred compiler collectives for this path.

## Validation

Executed on jk01 with two NVIDIA L20 GPUs (46068 MiB each), in an isolated container: Python 3.12.3, PyTorch 2.13.0+cu130, CUDA 13.0, Transformers 5.14.1.

- Partition/config regression: **17 passed**.
- Real two-rank DeepSpeed-engine integration: **three configurations each completed five optimizer updates**, with two accumulated micro-batches per update (SGD LR 0.05, ZeRO stage 0, FP32, clipping explicitly disabled in both paths).
- Configurations cover `lm_head` / `embed_out`, bias / no bias, hidden dimensions 32 / 35, and flattened / sequence-shaped inputs.
- Compared logits, CE loss, input gradients, replicated projection gradients, local head gradients, and updated weights against an independent unsharded PyTorch model after every update. Also checked the exact SGD update and that weights actually changed.
- Changed-file pre-commit checks and `git diff --check` passed.

The integration method `TestRowParallelOutputHeadTraining.test_five_optimizer_steps_match_unsharded_reference` was executed through two-rank torchrun on the real devices. It can also be run through pytest:

```sh
PYTHONPATH=.:tests python3 -m pytest tests/unit/module_inject/test_tp_partition_config_path.py -q
PYTHONPATH=.:tests python3 -m pytest tests/unit/v1/autotp/test_autotp_training.py::TestRowParallelOutputHeadTraining -v -s
```

This validates the minimal eager training contract, not large-model convergence, mixed-precision/ZeRO checkpoint round trips, or DeepCompile. Coupled tied-embedding sharding remains outside this PR's scope.


deepspeed/module_inject/auto_tp.py
```diff
@@ -464,13 +464,20 @@ def _create_row_parallel_layer(self, module, spec: TPLayerSpec, name: str):
         """Create row-parallel layer (AllReduce after forward)."""
         if self.conv_linear_layer:
             return Conv_LinearALlreduce(module, self.mp_group, name=name, tp_meta=self.tp_meta)
-        if name == "lm_head" or name == 'embed_out':
+        if name.rsplit('.', 1)[-1] in ("lm_head", "embed_out"):
```
jinyouzhi: https://github.com/deepspeedai/DeepSpeed/blob/ef7d23f4258f3469846f35864a7cb19c2c23b5fc/deepspeed/module_inject/auto_tp.py#L543
Can we reuse like it?

deepspeed/module_inject/layers.py
```diff
@@ -231,6 +231,28 @@ def backward(ctx: Any, grad_output: torch.Tensor) -> Tuple[None, torch.Tensor, N
         return None, grad_output, None
 
 
+class ScatterToTensorParallelRegion(torch.autograd.Function):
```
jinyouzhi, line 235: Could `ScatterToTensorParallelRegion` be considered the dual of `GatherFromTensorParallelRegion`? If so, can we use this relationship to design unit tests?

0z5a:
Reused `_is_lm_head_name`, added Scatter/Gather round-trip tests for outputs and gradients (including uneven splits), and resolved the main-branch conflicts. Two-L20 validation: 8 passed, including three five-step training comparisons. All changed-file pre-commit checks passed. Recreated the commit with the required sign-off.

jinyouzhi:
LGTM

delock (approved):

PR #8552, opened 2026-09-16
Preserve convolution shapes and scaling in Muon updates

vineethsaivs:
Muon crashes when updating convolution weights with four-dimensional gradients.

The overflow guard combines a flattened update with the original four-dimensional gradient. The aspect-ratio correction also reads the kernel dimensions instead of the flattened matrix. Restore the original shape before the overflow guard and derive the scale from the matrix being orthogonalized.

Test: `TORCHDYNAMO_DISABLE=1 DS_ACCELERATOR=cpu PYTHONPATH=. python -m pytest tests/unit/v1/ops/muon/test_muon_overflow.py -k 'convolution or overflowed_gradient or finite_gradient' -q`. Four convolution regressions fail before. Six CPU tests pass after, including two Conv2d training steps with both Newton-Schulz methods and overflow checks. Changed-file pre-commit passes. Executed on Apple M2 Pro CPU; no CUDA or distributed training run.


vineethsaivs:
Could you approve the CI run? The six CPU tests and changed-file pre-commit checks pass.

tests/unit/v1/ops/muon/test_muon_overflow.py
pengdurice, line 1: is this a good place for the new test?

vineethsaivs: No, and I moved them. `test_muon_overflow.py` is about the overflow invariant; only the shape-preservation half of this change fits there. Both convolution tests now live in `tests/unit/v1/ops/muon/test_muon_conv.py` in f5837006b966526af29ae77c660f421ed7bd8f15. 6 passed, 3 skipped on CPU; pre-commit clean.

pengdurice:
please fix the failed tests

vineethsaivs:
Fixed in 66ef97fcad37ca6fe6ae07cbc1c60cd13aada44b. The failure was my test, not the change.

`test_convolution_training_matches_flattened_muon` compared two 2-step training trajectories, and `zeropower_via_newtonschulz5` runs in bfloat16 where the accelerator supports it. The two shapes reach the iteration through different layouts, so the rounding diverges and feeds back through the weights: 1.97e-3 absolute on CI against float32 default tolerances. It passed locally because the gap is platform dependent, which is exactly what a test should not be.

Replaced it with the contract itself, which is exact everywhere:

- `muon_update` on a rank-4 gradient versus its flattened copy, asserted with `torch.equal`. Both go through `update.view(len(update), -1)`, so the iteration gets a bit-identical matrix and there is no tolerance to pick.
- the scale read off `[out, in*kh*kw]` rather than the kernel, asserted at 2 percent. The two readings are 1.633 and 1.225, so the old code misses by 25 percent and bfloat16 rounding (eps about 7.8e-3) is nowhere near it.
- a Conv2d step that just has to complete and stay finite, which is the crash this PR fixes.

Ten tests, and each half of the change now fails on its own: the shape half raises on unpatched code, the scale half fails the 2 percent assertion. Changed-file pre-commit passes.

alanhuangyoo (commented):
Thanks, the fix looks right to me. On 2×H20 everything passes except `test_convolution_scale_comes_from_the_flattened_matrix[gram]`, and only when the whole file runs. I think it's torch.compile on CPU: once `muon_update` gets recompiled with dynamic shapes, `max(1, a / b)` seems to use integer division, so the scale ends up √2 instead of √(32/12). `max(1, float(update.size(-2)) / update.size(-1))` fixes it here.

The expert and per-head branches on master have the same issue, I can send a small PR for those.

alanhuangyoo:
Small correction, `float()` isn't quite enough: when the graph breaks inside `muon_update` (the standard NS does), the max still gets cast to int64. `math.sqrt(rows / cols) if rows > cols else 1.0` works in every case I tried, it's `_aspect_ratio_scale` in #8641 if you want to reuse it.

pengdurice (approved):

pengdurice:
fix the test

alanhuangyoo:
The two failures look like the truncation from #8641: 0.2265 / 0.1961 = 1.155 = sqrt(32/12) / sqrt(2), so one of the calls got its scale from a recompiled graph. Using `math.sqrt(rows / cols) if rows > cols else 1.0` for that line should fix it, or `torch._dynamo.reset()` at the start of the test if you'd rather keep the line as is here.

pengdurice:
> The two failures look like the truncation from #8641: 0.2265 / 0.1961 = 1.155 = sqrt(32/12) / sqrt(2), so one of the calls got its scale from a recompiled graph. Using `math.sqrt(rows / cols) if rows > cols else 1.0` for that line should fix it, or `torch._dynamo.reset()` at the start of the test if you'd rather keep the line as is here.

Got you, then it seems that the fix in #8641 applied to full matrix should be the way to go? It's fine either PR is checked in first to me;-) Maybe you can fix it on your PR and let this one rebase on yours, wdyt?

vineethsaivs:
Reproduced on CPU: in a full-file run both scale tests get sqrt(2) instead of sqrt(32/12). The division is float, but after the dynamic-shape recompile Inductor's C++ codegen casts both `Max` arguments to `int64_t`, so your cast fixes it and the file passes 10/10 here. Please go ahead with the expert and per-head branches, thanks.

alanhuangyoo:
Sounds good, pushed it to #8641: the full-matrix line uses the same helper now, still on `grad`'s dims, so switching it to `update` stays with this PR.

PR #8565, opened 2026-09-17
Fix memory leak and NULL-deref in shm_initialize

tritsystem:
## Summary
Two related bugs in `shm_initialize()` (`csrc/cpu/comm/shm.cpp`), both found via static analysis and manually verified against the real code before fixing.

### 1. Memory leak (CWE-401)
```c
workspace_buf = (struct allreduce_workspace*)malloc(sizeof(struct allreduce_workspace));
snprintf(shm_name, NAME_BUF_SIZE, "%s_%d", shm_name_prefix, rank);
shared_create(&allreduce_buffer, shm_name, workspace_buf, sizeof(struct allreduce_workspace));
workspace_buf = (struct allreduce_workspace*)allreduce_buffer.bytes;  // original pointer overwritten here
```
`shared_create()` only reads `workspace_buf`'s contents (via a `write()` call that seeds the shared-memory segment) — it never stores or frees the pointer it's given. Once `workspace_buf` is reassigned to `allreduce_buffer.bytes`, the original `malloc()`'d block is unreachable for the rest of the process. Verified by reading `shared_create()`'s implementation directly and confirming there is no `free()` call anywhere in `shm.cpp`.

**Fix:** free the original buffer right after `shared_create()` returns (its contents have already been copied out at that point), before `workspace_buf` is repointed at the shared mapping.

### 2. NULL pointer dereference on unchecked calloc (CWE-476)
```c
workspace = (struct allreduce_workspace**)calloc(size, sizeof(struct allreduce_workspace*));
symmetric_buffer[0] = (char**)calloc(size, sizeof(char*));
symmetric_buffer[1] = (char**)calloc(size, sizeof(char*));
distributed_buffer[0] = (char**)calloc(size, sizeof(char*));
distributed_buffer[1] = (char**)calloc(size, sizeof(char*));
// (no NULL check anywhere)
```
All five are dereferenced unconditionally in the loop immediately after (`workspace[i] = ...`, `workspace[i]->buffer`, `symmetric_buffer[*][i] = ...`, etc.). `calloc()` can return `NULL` under memory pressure; any of these failing would then be a NULL-pointer write. Verified there is no existing check on any of these five variables anywhere in the file.

**Fix:** a single guard checking all five allocations before the loop uses any of them, matching this file's existing fail-and-return style (`shared_create()`'s own printf-then-return error handling).

## Test plan
- [x] `clang-format` (repo's own `.clang-format`) — no reformatting needed for either commit, diffs already conform
- [x] `codespell` — clean on both commits
- [x] Manually traced `shared_create()`'s implementation and confirmed no `free()` anywhere in the file (leak fix)
- [x] Grepped the whole file for any existing NULL check on `workspace`/`symmetric_buffer`/`distributed_buffer` — none found (NULL-deref fix)
- [x] Syntax-checked each modified block in isolation against realistic stand-in types matching the real struct/function shapes (`g++ -fsyntax-only`, clean)
- [ ] Could not build/run the full CUDA/PyTorch extension in this environment to exercise `shm_initialize()` at runtime — flagging this limitation explicitly rather than claiming a test run that didn't happen


csrc/cpu/comm/shm.cpp
```diff
@@ -362,6 +362,11 @@ void shm_initialize(int size, int rank, char* addr_string, char* port_string)
     workspace_buf = (struct allreduce_workspace*)malloc(sizeof(struct allreduce_workspace));
```
ebarkhordar, line 362: I ran this against `89ca1a04` in a clean `gcc:13` container, compiling `SharedData`, `shared_open()` and `shared_create()` from `shm.cpp:42-85` verbatim plus the real `allreduce_workspace` struct.

The guard covers the five `calloc()`s at `:375-379`, which ask for 320 bytes in total at world size 8. The `malloc()` on this line asks for one `allreduce_workspace`, measured at 69206024 bytes, and it is still unchecked. That is the allocation here that fails under memory pressure.

It escapes the guard because a NULL from `:362` never reaches a NULL dereference. It goes into `shared_create()`:

```
write(d, NULL, 69206024) -> -1 errno=14 (Bad address)
nbytes = write(...) assigned -> 18446744073709551615 ; truthy in `if (nbytes = write(...))`
shared_open's shm_open -> d=4, so it takes the SUCCESS branch
mmap(NULL, 18446744073709551615) -> 0xffffffffffffffff (MAP_FAILED) errno=12
```

The assignment inside the `if` at `:73` turns -1 into `(size_t)-1`, which is truthy, so `shared_open()` runs with that length. Its `shm_open()` succeeds, since the segment was just created, so `:57` assigns `data->bytes = bytes` unconditionally and stores `MAP_FAILED`. `descriptor` is a live fd, so the `descriptor == -1` sentinel never fires. Driving both arms through the real functions, only `workspace_buf` forced to NULL, `:371` writes through `(void*)-1` and the process takes SIGSEGV; the malloc-succeeds arm exits 0.

Pre-existing, and your diff does not introduce it, but it is the same CWE-476 in this function, 13 lines above the guard. One edit: check `workspace_buf` straight after this line and bail the way the new guard does.

Not checked: I did not build the extension or call `shm_initialize()` itself, and the second arm forces the NULL rather than exhausting memory.

tritsystem: Thank you for the thorough trace here — really appreciate you walking through the exact chain (the `write()` EFAULT, the `-1` -> `size_t` reassignment into `SIZE_MAX`, and `shared_open()` storing `MAP_FAILED` unconditionally). That's a much clearer root-cause explanation than I had.

I reproduced it independently before pushing anything: extracted `shared_open()`/`shared_create()` verbatim into a standalone repro and forced the same NULL-malloc path. Unguarded, it segfaults exactly as you described; with the guard, it bails cleanly.

Applied your suggested fix in a7782f865, checking `workspace_buf` right after the `malloc()` and bailing the same way the existing `calloc()` guard below it does. Thanks again for catching this — let me know if you'd like anything adjusted.

ebarkhordar: Confirmed at `a7782f865`. I re-ran the same two-arm harness in a clean `gcc:13` container against the guarded and unguarded versions of this prefix, with only `workspace_buf` forced to NULL:

```
unguarded / malloc ok    exit=0
unguarded / malloc NULL  exit=139   (SIGSEGV at the states[0] write)
guarded   / malloc ok    exit=0
guarded   / malloc NULL  exit=0     "malloc failed to allocate workspace_buf", returns before shared_create()
```

The guard sits above `shared_create()`, so the NULL no longer reaches `write()` and the `nbytes` reassignment never happens. The malloc-succeeds arm is unchanged.

sfc-gh-truwase (approved):

sfc-gh-truwase:
@tritsystem thanks for the PR. Can you please address the DCO blocker?

csrc/cpu/comm/shm.cpp
```diff
@@ -360,8 +360,25 @@ void shm_initialize(int size, int rank, char* addr_string, char* port_string)
     struct allreduce_workspace* workspace_buf;
     struct allreduce_workspace* workspace_buf_other;
     workspace_buf = (struct allreduce_workspace*)malloc(sizeof(struct allreduce_workspace));
+    // malloc() can return NULL under memory pressure. Left unchecked, the NULL flows
```
delock, line 363: Don't need lengthy explaination here, can you remove these 8 lines of comments here?  The code itself explains well.

delock: Hi @tritsystem , can you follow up with this comment?  Thanks!

tritsystem: Thanks @delock, and sorry for the slow follow-up. I removed the 8-line comment in 8683537; the guard reads fine on its own. The change is comment-only: compiling the old and new file with comments stripped gives identical output. The failure chain it described is still in the discussion above and in 8d07b4e's commit message if anyone needs it. I left the shorter comments on the `free()` ownership and the `calloc` guard alone since you only mentioned this block, but I'm happy to trim those too.

delock (approved):

PR #8569, opened 2026-09-17
[CI] Raise the Modal Sandbox test budget so the full tests/unit/v1 suite can finish

0z5a:
# Raise the Modal Sandbox test budget so the full `tests/unit/v1` suite can finish

`modal-torch-latest` runs the **entire** `tests/unit/v1` suite (`mode=all`) on 2×L40S inside a
single Modal Sandbox (`push` and `merge_group` have no trustworthy base diff, so the selector
always falls back to all). That suite now needs longer than the Sandbox's `timeout`, so recent
runs either die on the deadline or finish with seconds to spare.

## Symptom

`Run tests in isolated Modal Sandbox` ends with:

```
RuntimeError: run pytest failed with exit code 137
```

Exit code 137 is `128 + 9` (SIGKILL) and lands exactly on `SANDBOX_TIMEOUT_SECONDS = 4200`:

| Event | Time (UTC) |
|---|---|
| `Sandbox started after 40s` | 2026-09-17T07:43:17 |
| first pytest result | 2026-09-17T07:46:13 |
| SIGKILL (`exit 137`) | 2026-09-17T08:53:24 |
| elapsed | **4207s ≈ the 4200s budget** |

pytest was healthy and at ~90% with all four workers still producing results — no `FAILED`,
no `ERROR`, no OOM. It simply ran out of wall clock.

## This is not specific to any one PR

Same-mode runs in the last ~24h, counting only those that actually created a Sandbox:

| Run | Event | Result | Duration |
|---|---|---|---|
| 07:43 | merge_group | failure | 70:20 (137) |
| 06:50 | merge_group | failure | 76:30 (137) |
| 06:38 | **push (master)** | failure | 70:25 (137) |
| 06:23 | merge_group | failure | 78:02 (137) |
| 04:47 | merge_group | failure | 70:23 (137) |
| 04:04 | **push (master)** | failure | 76:31 (137) |
| 02:56 | merge_group | success | 67:09 — **2:51 of margin left** |

`master`'s own push runs fail the same way, and the one passing run squeaked in 2m51s before the
cap. Runs that "passed" in ~0s never created a Sandbox at all (`mode=none`, or the Sandbox is
skipped for `pull_request_target`).

## Cost of the suite

Summing per-worker busy time for the failing run: 4 × 67.2 min of test time (16,124s total), i.e.
**~4031s ≈ 67.2 min even with perfect four-way balance**, plus ~3 min to acquire the Sandbox and
run the install chain. A few files dominate:

| File | Machine time |
|---|---|
| `tests/unit/v1/zero/test_zero_user_backward.py` | 59.0 min |
| `tests/unit/v1/ops/muon/test_muon.py` | 44.0 min |
| `tests/unit/v1/zero/test_offload_states.py` | 27.6 min |
| `tests/unit/v1/zero/test_zero.py` | 27.5 min |

## Change

- `ci/torch_latest.py`: `SANDBOX_TIMEOUT_SECONDS` `4200` → `5400`, with a comment recording that
  the Sandbox clock covers acquisition + install + pytest (so future budgets are not estimated
  from pytest time alone).
- `ci/test_torch_latest.py`: update the assertion that pins that constant.

Budget after the change, against the unchanged `deploy.timeout-minutes: 105`:

| Item | Value |
|---|---|
| install/setup overhead (observed) | ≈ 3.1 min |
| pytest (observed) | ≈ 67 min |
| new budget | 90 min |
| pytest time available | ≈ 86.9 min |
| projected job wall clock | ≈ 93 min ≤ 105 min |

## Note for reviewers / merge order

Both jobs check out the **trusted base**:

```yaml
ref: ${{ github.event.pull_request.base.sha || github.event.merge_group.base_sha || github.sha }}
```

so a PR that changes `ci/` cannot fix its own Modal run — the Sandbox always executes the base
commit's `ci/torch_latest.py`. This fix therefore only takes effect once it is on `master`; every
open PR's `merge_group` run picks it up from there.

## Verification

- `git apply` / `patch -p1 --dry-run` against the base tree is clean.
- `build_sandbox_kwargs` self-check (equivalent to
  `ci/test_torch_latest.py::test_sandbox_kwargs_are_fixed_and_secret_free`) passes and reports
  `timeout == 5400`, `gpu == "l40s:2"`, `acquire == 1800`.

## Follow-ups intentionally left out

- Sharding the full suite across multiple Sandboxes, or diff-selecting on `merge_group` via
  `merge_group.base_sha` instead of `all`, is the durable fix — one Sandbox on 2 GPUs is now
  near its physical ceiling.
- `pytest -n 4` uses the default `--dist load`; `--dist loadfile` or more workers may shorten the
  critical path, but four workers already share two GPUs, so that needs measurement.


0z5a:
@/tmp/close8569.md

0z5a:
Superseded and reverted: the CI budget change was briefly committed to #8526 on request, then removed again after the maintainer asked for it to be reverted so the sandbox behavior can be investigated separately. This standalone PR is no longer needed, so it is closed.

The standalone commit still exists as `dd1725d29fa44c8a99d9b4eecd60734a2cfaab54` on branch `ci/raise-modal-sandbox-timeout` and applies cleanly to `master`, in case the budget change is wanted later as its own PR. Reopen or ask and I will take it from there.

PR #8570, opened 2026-09-17
Raise Modal sandbox timeout to absorb full-suite runtime

delock:
## What

Raise `SANDBOX_TIMEOUT_SECONDS` in `ci/torch_latest.py` from 4200s to 5400s.

## Why

The master-push full suite currently needs ~62 min of pytest on top of sandbox setup, which leaves almost no headroom under the 4200s sandbox limit. Recent master runs are being SIGKILLed (pytest exit code 137) near completion whenever pytest runs a few minutes slower than average (Modal hardware/zone variance), e.g. run 35114915337 died at 96% with the sandbox alive for ~4255s.

The deploy job timeout was already raised to 105 min, so a 5400s (90 min) sandbox timeout plus worst-case ~8 min of controller/image overhead still fits with margin.

## Validation

- `pre-commit run --files ci/torch_latest.py` passes.
- No behavior change beyond the timeout constant; merge-queue entries pick up the new value because the controller is checked out from the merge candidate SHA.


delock:
Long term plan is move multi-rank + non GPU mandatory back to CPU workflow.  And take furhter action to reduce modal usage.  The former part is an on going task from #8381 .  If you have thoughts and ideas, please leave comments in that PR.  Thanks!

PR #8571, opened 2026-09-17
Raise Modal sandbox timeout to absorb full-suite runtime

jinyouzhi:
The master-push full suite now needs ~62 min of pytest on top of sandbox setup, which leaves almost no headroom under the 4200s sandbox limit: runs are SIGKILLed (exit 137) near completion whenever pytest is a few minutes slower than average. Raise the sandbox timeout to 5400s; the deploy job timeout (105 min) already covers the added margin.
@delock


delock (approved):

delock:
Long term plan is move multi-rank + non GPU mandatory back to CPU workflow. And take furhter action to reduce modal usage. The former part is an on going task from https://github.com/deepspeedai/DeepSpeed/pull/8381 . We will also need cto address UT number increase.  If you have thoughts and ideas, please leave comments in that PR. Thanks!

PR #8573, opened 2026-09-17
Keep runtime configuration validation active under python -O

### LINKED ISSUE #8572: [BUG] Runtime configuration validation is bypassed under python -O
## Bug: Runtime configuration validation is bypassed under `python -O`

`deepspeed/runtime/config.py` uses Python `assert` statements for several runtime configuration validation checks.

When Python is executed with optimization enabled (`python -O`), assertion statements are removed. As a result, invalid DeepSpeed configurations can bypass validation.

### Minimal reproduction

```bash
python -O -c 'from deepspeed.runtime.config import DeepSpeedConfig; DeepSpeedConfig({"train_batch_size": 0})'
```

With normal Python execution, the invalid configuration triggers the existing validation error. With `python -O`, the assertion is removed and the invalid configuration can be accepted.

The same issue affects other validation paths in `DeepSpeedConfig`, including:

* invalid batch-size values
* inconsistent batch-size arithmetic
* simultaneously enabling fp16 and bf16
* missing batch-size configuration
* invalid ZeRO stages
* invalid fp16/bf16 ZeRO-related configuration

### Proposed fix

Replace the affected runtime `assert` statements with explicit condition checks and exceptions so that configuration validation is enforced regardless of Python optimization mode.

The existing validation conditions and error messages should be preserved where practical.

Regression tests should verify the affected validation paths and ensure they remain active when running with `python -O`.

### Related work

Related configuration-validation discussions/issues include #3982, #7852, and #8260, as well as PR #8336. These address different validation problems but provide related context for explicit configuration validation.

I did not find an existing issue describing this specific `python -O` assertion-removal problem.

Saisharathchandranandnetha:
Fixes #8572.

Running Python with `-O` removes the assertions used by `DeepSpeedConfig`, allowing invalid batch sizes, inconsistent batch arithmetic, missing batch settings, and incompatible precision options to bypass validation. Replace those assertions with explicit checks and `raise AssertionError` so validation remains active while preserving the existing exception type, conditions, and error messages.

Regression coverage exercises zero, negative, and NaN batch values; inconsistent batch arithmetic; simultaneous fp16/bf16; missing batch settings; and fp16/bf16 ZeRO requirements. A subprocess regression runs actual optimized Python during a normal test run.

Validation:
- `python -m pytest tests/unit/runtime/test_ds_config_dict.py -q -k "TestConfigValidation"`: 16 passed.
- The same command with `python -O`: 16 passed.
- Final focused runs including the optimized-Python subprocess regression: 17 passed in each mode.
- Full test file in both modes: 54 passed, 4 skipped, 12 environment failures. The failures occur in multiprocessing forkserver setup because the execution sandbox blocks Unix-socket binding, before the affected test bodies run. The optimized-Python subprocess regression passes.
- Test environment: Intel Core i5-12450HX CPU, Python 3.14.7, PyTorch 2.14.0+cu130; no accessible GPU. Distributed/GPU integration remains unverified.
- All applicable repository pre-commit hooks passed for the two changed files. Used a temporary config with identical hooks and Python 3.11 as the hook interpreter because pinned flake8 5.0.4 crashes on Python 3.14; the repository hook configuration is unchanged.
- `git diff --check`: passed.


ebarkhordar (commented):
I ran this against c196d410 in a clean container (python:3.11-slim, torch 2.14.0+cpu, CPU accelerator). The conversion is complete for the file it targets: parsing `deepspeed/runtime/config.py` with `ast` gives 13 `assert` statements at merge-base 327190be and 0 at this head.

Three config-validation asserts reached from the same `DeepSpeedConfig(param_dict)` call survive, which matters because #8572 names "other validation paths in `DeepSpeedConfig`" as in scope:

- `data_pipeline/config.py:88`, curriculum learning on with no `curriculum_metrics`, and `:108`, dynamic batching on with no `max_tokens`, both reached from `config.py:654` `get_data_efficiency_config`
- `zero/config.py:399`, partial offload outside stage 3, reached from `config.py:599` `get_zero_config` and its `offload_ratio_check` validator

Two of the three take the invalid config outright under `-O`:

```
__debug__=True
A curriculum, no metrics : AssertionError: Curriculum learning is enabled, curriculum_metrics must be specified
B dynamic, no max_tokens : AssertionError: Dynamic batching is enabled, so max_tokens must be specified
C partial offload stage 1: ValidationError: Assertion failed, Partial offloading only supported for ZeRO Stage 3.

__debug__=False (python -O, same head)
A curriculum, no metrics : NO EXCEPTION, config accepted
B dynamic, no max_tokens : KeyError: 'max_tokens'
C partial offload stage 1: NO EXCEPTION, config accepted
```

`train_batch_size: 0` raises under `-O` in that same run, so the two silences are the flag and not my fixture.

The same treatment at those three sites would close them. `test_config_validation_optimized_python` pins only the batch-size path, so it stays green either way.

I did not test a GPU or distributed path.

ebarkhordar (commented):
I ran 7ecfd3f3 in a clean container (python:3.10-slim, torch 2.4.1 cpu, `pip install --no-deps .`). The three checks I raised earlier now behave identically with and without `-O`.

On whether #8572's "other validation paths in DeepSpeedConfig" are now closed: I traced every deepspeed function executed while building a valid `DeepSpeedConfig` (130 of them), then parsed those for `ast.Assert` rather than grepping. Five bare asserts remain. One is `comm.get_rank`, a runtime precondition rather than config validation. The other four still vanish under `-O`:

```
                                                    normal        -O
train_batch_size=0 (converted by this PR)           raises      raises    <- control
autotuning/config.py:39   empty results_dir         raises      ACCEPTED
autotuning/config.py:41   empty exps_dir            raises      ACCEPTED
swap_tensor/aio_config.py:39  GDS on non-CUDA       raises      ACCEPTED
config_utils.py:89  deprecated + replacement both   raises      ACCEPTED
```

The first row is a positive control: it raises under `-O` too, so the harness detects a raise while `__debug__` is False. Every case raises under normal python, so each precondition does fire.

I would put `config_utils.py:89` first. It sits in `DeepSpeedConfigModel._process_deprecated_field`, so it guards every subconfig declaring a deprecated field, not one config. Under `-O`, passing both `cpu_offload_param` and `offload_param` writes the deprecated value over the explicit one and reports nothing.

The same conversion you used would cover all four. I did not test any GPU or distributed path, and I checked config construction only.

Saisharathchandranandnetha:
Thanks for the review and for catching these missed validation paths, @ebarkhordar.

You're right that the curriculum learning metrics, dynamic batching `max_tokens`, and ZeRO Stage 3 partial-offloading checks relied on `assert` statements, which can be removed under `python -O`.

I've replaced those checks with explicit exception raising. I also identified two additional assert-based validation paths in autotuning and model checkpointing and updated those as well.

I expanded `tests/unit/runtime/test_ds_config_dict.py` with regression coverage, including a subprocess test that runs the invalid configurations under an optimized Python interpreter (`-O`). I also verified the test's effectiveness through mutation testing: restoring the original assert-based behavior caused the regression test to fail as expected. (All 24 validation tests passed under both standard and `-O` conditions).

The updated changes are committed with sign-off. The diff passes `git diff --check`. I've included the additional fixes and regression tests in my branch for review.

Thanks again for pointing these out!

Saisharathchandranandnetha:
Thanks for the additional audit, @ebarkhordar. I appreciate you tracing the remaining assertions through the `DeepSpeedConfig` construction path.

I've already addressed the autotuning checks in the latest commits. I'm now checking the remaining `swap_tensor/aio_config.py` and `config_utils.py` paths, with regression coverage under `python -O`.

I'll update this thread with the verified changes and test results once those checks are complete.

PR #8575, opened 2026-09-17
Describe the GPTBigCode and Yuan shared-QK layouts

Achyuthan-S:
Three of the four layouts in `AUTOTP_UNSUPPORTED_PARAMETER_PATTERNS` now describe themselves, so universal checkpoint conversion stops refusing them. Follows #8519, which made AutoTP layers emit affine maps for the layouts a generic rule already covered.

`codegentype` is not included — its reshape-and-interleave needs more than a segment list or a block selection, and it is the next piece of work.

### What was actually missing

Both refusals came from the same limit: a parameter carries one partition dimension, which cannot say that one block of it is sharded while the next is held whole, nor that a rank takes blocks which are not adjacent.

**GPTBigCode** splits the query rows across ranks and hands every rank the whole key/value block. `segmented_map` describes a parameter as an ordered list of segments, each split or replicated, so the two end up as pieces that differ in which ranks hold them rather than in kind. That is the case the per-piece location set was introduced for in #8385.

**Yuan shared-QK** gives a rank the value heads that pair with its query heads, plus the matching second-half heads. Those land in two runs rather than one span, so `block_gather_map` takes a selection of equal-sized blocks and merges consecutive ids into single pieces.

### Why the tests are worth reading

A layer builds its map analytically, from head counts and the tp degree. The test harness derives one independently by running the **real** partition function on a marker tensor and reading back where each slice came from. The tests require the two to agree, so what ships is checked against what the partition code actually does rather than against a reading of it.

I confirmed they fail when the map is wrong: corrupting `segmented_map` so the replicated segment claims a single owner produces two failures rather than passing quietly.

`shared_qk_value_head_ids` is extracted from `shard_value_with_share_qk` and used by both the partition and the map, so the description cannot drift from the layout it describes.

### What stays unsupported, deliberately

- **A layout with no head counts.** Describing Yuan needs `num_kv_heads`; without it the layer publishes the refusal rather than guessing, and there is a test for that.
- **The Yuan o_proj bias.** Its partition pre-divides by the world size and replicates, which needs a scaled piece — and that code path does not currently produce a usable bias, so there is no working layout to describe.

### Validation

 Validated on CPU/gloo (DS_ACCELERATOR=cpu LOCAL_SIZE=4): 90 passed across the affine suite, the producer and coverage tests, the resume matrix and tests/unit/runtime/tensor_parallel/.

Related: #8252, #8230.

cc @delock


deepspeed/module_inject/fusedqkv_utils.py
```diff
@@ -215,6 +215,27 @@ def _transpose_fused_qkvw(src, mp_size, fused_qkv_type=None, module=None):
...
+    """
+    head_per_rank = num_heads // world_size
+    q_head_start = rank * head_per_rank
+    v_head_ids = []
+    index = 0
+    while index < head_per_rank:
+        v_head_ids.append(q_head_start // 2)
+        q_head_start += 2
+        index += 2
+    v_head_ids.extend([head + num_heads // 2 for head in v_head_ids])
```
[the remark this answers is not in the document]
Achyuthan-S: Confirmed and fixed in 5f84b034. Twelve heads over four ranks gives each rank three, and the pairing then overlaps rather than partitions — heads 1, 4, 7 and 10 land on two ranks each:

     rank 0: [0, 1, 6, 7]
     rank 1: [1, 2, 7, 8]
     rank 2: [3, 4, 9, 10]
     rank 3: [4, 5, 10, 11]

    Nothing is missing in this case, but the overlap is the problem: a piece naming a single owner contradicts the one beside it, and my map published exactly that. Both Yuan builders now check the selection partitions the head range and return no map otherwise, so the layout stays unsupported rather than being described inconsistently. Test added with the 12/4 config, asserting first that the pairing really does overlap so it can't quietly stop testing anything.

    Worth noting the pairing itself is unchanged — this only governs whether a map is published for it.

deepspeed/module_inject/layers.py
```diff
@@ -1193,6 +1208,22 @@ def _freeze_partition_sizes(self, total_size):
...
+        """
+        from deepspeed.checkpoint.affine import segmented_map
+        from deepspeed.module_inject.fusedqkv_utils import get_fused_qkv_type
+
+        if get_fused_qkv_type(self.fused_module.module) != 'bigcodetype':
+            return None
+        n_embd = self.tp_meta.n_embd
+        if not shape or n_embd is None or n_embd >= shape[0]:
+            return None
+        return segmented_map(shape, [(n_embd, False), (shape[0] - n_embd, True)], 0, self.tp_world_size)
```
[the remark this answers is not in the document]
Achyuthan-S: Fixed in 5f84b034: segmented_map now requires explicit per-rank widths for split segments and raises if they don't sum to the segment, and the layer passes get_shard_size_list(...) — the same call the partition makes. Test covers the uneven config and asserts up front that the widths differ from an even split.

 My tests missed both of these because every config I'd picked happened to be one where the two agree.

deepspeed/module_inject/layers.py
```diff
@@ -1193,6 +1208,29 @@ def _freeze_partition_sizes(self, total_size):
         if self._subparam_shard_widths is not None:
             set_fused_qkv_shard_state(self.fused_module.module, self._subparam_shard_widths, self.tp_index)
 
+    def _segmented_affine_map(self, shape):
```
delock, line 1227: **Missing wiring test for GPTBigCode**

`test_yuan_weight_is_no_longer_refused_by_conversion` verifies the full wiring for Yuan (the layer actually publishes `AFFINE_MAP_PARAMS` and is removed from `AUTOTP_UNSUPPORTED_PARAMETER_PATTERNS`), but there is no equivalent for `bigcodetype`. The geometry tests validate `segmented_map` itself, but nothing fails if `fused_LinearLayer._segmented_affine_map` never gets called (e.g. a hook-name typo or `get_fused_qkv_type` returning something else in that context) — the layout would silently stay refused while all tests stay green.

Could you mirror the Yuan test for a `fused_LinearLayer` with a `GPTBigCodeBlock` module? The pattern should transfer directly.

deepspeed/module_inject/layers.py
```diff
@@ -1240,6 +1278,29 @@ class Yuan_LinearAllreduce(LinearAllreduce):
     _unsupported_uc_reason = ("Yuan shared-QK tensor parallelism selects noncontiguous head groups that universal "
                               "checkpoint conversion cannot currently describe")
 
+    def _shared_qk_affine_map(self, shape):
```
delock: **Duplicated `_shared_qk_affine_map` in the two Yuan classes (non-blocking)**

`Yuan_LinearAllreduce` and `Yuan_LinearLayer` carry near-identical copies of `_shared_qk_affine_map` (guards, completeness check, and its comment included); the only real difference is the partition axis. This is essential rather than incidental duplication — both copies encode the same invariant (the head selection must partition the head set) and will need to change together. The `len(shape) <= 0` leftover in the other copy shows how they can drift. A shared helper taking `partition_dim` would collapse them to one-liners at each call site.

deepspeed/checkpoint/affine.py
```diff
@@ -480,3 +486,118 @@ def _piece_from_dict(entry):
                        dest_strides=dest_strides,
                        locations=entry['locations'],
                        scale=entry.get('scale', 1.0))
+
+
+def segmented_map(shape, segments, partition_dim, tp_degree, split_widths=None):
```
delock: **Consider making `split_widths` required (non-blocking)**

The only production caller (`fused_LinearLayer._segmented_affine_map`) already passes explicit widths; the `split_widths=None` fallback is only exercised by one test. Since an inferred split can read the wrong rows while still covering the tensor (the exact bug the second commit fixed), removing the fallback would close that door for future callers. `_even_split_sizes` has no other callers, so it can be deleted along with the fallback — the one test call site just passes explicit widths.

deepspeed/module_inject/layers.py
```diff
@@ -1274,6 +1340,29 @@ class Yuan_LinearLayer(LinearLayer):
...
```
delock: **Dead condition in `Yuan_LinearLayer._shared_qk_affine_map` (non-blocking)**

`len(shape) <= 0` is always false for a tensor shape (`ndim >= 1`). It looks like a copy-paste leftover from the `len(shape) <= 1` guard in `Yuan_LinearAllreduce` (which protects the `shape[1]` access). Harmless here since `shape[0]` is valid for 1-D, but the check as written does nothing — either drop it or restore the intended bound.

delock (changes requested):
Thanks for the PR — the shared-source design (`shared_qk_value_head_ids` used by both the partition and the map) and the marker-tensor verification methodology are excellent.

One blocking item (missing GPTBigCode wiring test) and three non-blocking suggestions below.

delock:
Hi @Achyuthan-S , thanks for the PR.   Being able to handle GPTBigCode and Yuan really show the power of this work.  The comments is left above.  Thanks!

Achyuthan-S:
@delock All four addressed in f8b9022.

   The wiring test — added and confirmed it earns its place: renaming _segmented_affine_map makes it fail while the geometry tests stay green, which is the scenario you described.

   The duplication — collapsed into one module-level helper taking partition_dim. Your point about it being essential rather than incidental was borne out: the len(shape) <= 0 leftover disappeared as a consequence of merging them, rather than needing its own fix.

   split_widths required — done, and _even_split_sizes deleted with it. Agreed that leaving the fallback open invites exactly the bug the second commit fixed.

 90 passed across the affine suite, the producer and coverage tests, the resume matrix and tests/unit/runtime/tensor_parallel/.

delock (approved):

Issue #7100, opened 2025-03-02
Support DualPipe training


ehartford:
I would like to propose adding support for DualPipe in DeepSpeed. Developed by DeepSeek AI, DualPipe is a bidirectional pipeline parallelism algorithm that enhances training efficiency by fully overlapping forward and backward computation-communication phases, reducing pipeline bubbles, and optimizing resource utilization. This approach allows simultaneous processing of micro-batches from both ends of the pipeline, minimizing idle times and improving throughput, especially beneficial for training large-scale models. Integrating DualPipe into DeepSpeed would provide users with an advanced option for pipeline parallelism, complementing existing features and potentially improving training efficiency for large models. Since DualPipe is implemented with PyTorch 2.0 and above, it should integrate smoothly into DeepSpeed's existing architecture.

https://github.com/deepseek-ai/DualPipe

phoeenniixx:
Hello @loadams, @ehartford, is this issue being worked at? I would really like to give it a go if you say so :)
I really would like to contribute to DeepSpeed...

ehartford:
Nobody has taken it up yet

phoeenniixx:
Thanks! I will start working on this. Although I might need some time... I have my mid sems coming up :)

PR #8576, opened 2026-09-17
Add `DualPipeV`

phoeenniixx:
Implements `DualPipeV` from `DeepSeek` team
Part of #7100.

This adds DualPipeV, the V-shaped schedule from [DualPipe](https://github.com/deepseek-ai/DualPipe), as an opt-in alternative to 1F1B. Existing pipeline users are not affected. A model opts in by building a `DualPipeVModule` instead of a `PipelineModule`.

```python
from deepspeed.pipe import DualPipeVModule

net = DualPipeVModule(layers=layers, num_stages=4, loss_fn=loss_fn)
engine, _, _, _ = deepspeed.initialize(config=config, model=net, model_parameters=net.parameters())
loss = engine.train_batch()
```

## Changes
- Adds `DualPipeVModule`, which cuts the model into `2P` stages and builds stages `r` and `2P-1-r` on rank `r`.
- Adds `DualPipeVEngine`, a child of `PipelineEngine`. It  replaces only the send and receive methods with batched `isend`/`irecv`.
- Adds `DualPipeVSchedule`, which follows the eight steps of the reference `dualpipev.py`.
- Adds `P2POp` and `batch_isend_irecv` to `deepspeed.comm`.
- `PipelineEngine` gets three small hook methods for the schedules and the loss rank, and `PipelineModule` reads `self.num_stages` when partitioning. Neither changes 1F1B behaviour.
- Tests for the schedule protocol, and an end-to-end test against an unpartitioned `nn.Sequential`.

I also tried it using a small experiment on Kaggle's 2 T4 using [this scipt](https://gist.github.com/phoeenniixx/245535be06ae731d2a8329590f43dca1). See the results [here](https://www.kaggle.com/code/arian32/notebook561d1e71d8)
| schedule | step time (s) | samples/s | peak mem rank 0 (MB) | peak mem rank 1 (MB) | final loss |
|---|---|---|---|---|---|
| 1F1B (PipelineModule) | 0.465 | 68.9 | 621 | 525 | 1.000726 |
| DualPipeV (DualPipeVModule) | 0.439 | 72.9 | 761 | 909 | 1.000726 |

## Limitations
- `gradient_accumulation_steps` must be at least twice the number of pipeline ranks.
- `dynamic_shape`, `pipe_partitioned`, `grad_partitioned` and boolean attention masks are not supported.
- No forward/backward overlap and no deferred weight gradients yet, so the speedup is smaller than in the DualPipe README.

## Follow-ups
If this approach is accepted, I plan to raise follow-up PRs for:
- The bidirectional DualPipe schedule. It can reuse parts of `DualPipeVEngine` (maybe move the common parts to a common base class in future?) and needs a second `steps()` generator plus data loading and loss on both end ranks.
- Forward/backward overlap and the zero-bubble weight-gradient split.
- Tutorial and API reference docs.


phoeenniixx:
FYI @delock @ehartford

phoeenniixx:
Hi @delock, can you please review this PR whenever you have some time.
Thanks

delock:
Hi @phoeenniixx , thanks for introducing dualpipev into deepspeed.  I have a few comments:
1. For stage partitioning, is it possible to allow user pass list of layer list as a way of manual partitioning?
2. Need to update documentation for dualpipev as well.
3. For new file (duappipev.py) Microsoft copyright head should not be used.  Follow suggestion from AGENTS.md
4. batch_isend_irecv lackes @timed_op, is it intended?
5. The per-phase activation-checkpointing path (_precompute_checkpointable_values cache reset) is never exercised: the e2e test uses activation_checkpoint_interval: 0. Could you add a case with interval > 0?
6. [Optional] self.loss_model at engine.py:225 has been write-only since 2022 ("use the loss fn directly from the module") — feel free to drop it in passing, or leave for a separate cleanup.

Issue #1454, opened 2021-10-13
Issues with running inference on quantized model


HamidShojanazeri:
I have a question on MoQ usage and appreciate the help. I have run the [example](https://github.com/microsoft/DeepSpeedExamples/tree/master/MoQ) that train a bert-base model with MoQ, then trying to load it back for inference using the code snippet below.

```
import torch
import copy
from transformers.models.bert import BertLayer
from transformers import (
    AutoConfig,
    AutoModelForSequenceClassification,
    AutoTokenizer)
device = 'cuda'
model = AutoModelForSequenceClassification.from_pretrained('../DeepSpeedExamples/MoQ/output-8bits/qnli/checkpoint-9000/')
tokenizer = AutoTokenizer.from_pretrained('bert-base-cased')

inputs = tokenizer("Hello, my dog is cute", return_tensors="pt")
input_ids = inputs["input_ids"].to(device)
attention_mask = inputs["attention_mask"].to(device)
test_model = copy.deepcopy(model).to('cuda')
import deepspeed
import deepspeed.module_inject as module_inject
injection_policy={BertLayer:
                    module_inject.HFBertLayerPolicy}
ds_engine = deepspeed.init_inference(model,
                            mp_size=1,
                            dtype=torch.int8,
                            replace_method='auto',
                            quantization_setting=8,
                            injection_policy=injection_policy)
model = ds_engine.module
output = model(input_ids,attention_mask=attention_mask) # or output = model(input_ids,attention_mask)
```
This runs into an issue that complains about missing input_mask in the inputs, [logs](https://gist.github.com/HamidShojanazeri/fddae1c09aaf4fdfdd6d247566f1e661), 
```
TypeError: compute_attention() missing 1 required positional argument: 'input_mask'
```
I was wondering if the setting for inference engine is  missing something and what would be the best practice to run inference on a MOQ quantized model. Thanks.

--- Update, it seems like a bug that [compute_attention(qkv_out)](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/ops/transformer/inference/transformer_inference.py#L304) is missing input_mask, added and could pass the error.  

[...]

PR #5505, opened 2024-05-08
inference: remove unused _validate_args function

nelyahu:

PR #8578, opened 2026-09-18
Accept torch.int8 in init_inference's dtype validation gate

### LINKED ISSUE #8577: init_inference rejects dtype=torch.int8 on every accelerator
**Describe the bug**
`deepspeed.init_inference(model, dtype=torch.int8)` raises `ValueError: Data type torch.int8 is
not supported by <device> accelerator` unconditionally, on every accelerator DeepSpeed ships
(checked `cpu`, `cuda`, `hpu`, `mlu`, `xpu` — none list `torch.int8` in `supported_dtypes()`).
This is a regression: two unrelated cleanup PRs (#6528, "add bfloat16 to inference support
dtypes", 2024-09-27; and #5505, "remove unused _validate_args function", 2025-01-07) combined to
silently tighten `InferenceEngine.__init__`'s dtype gate (`deepspeed/inference/engine.py:74`) from
its original fp16-only check to a blanket `supported_dtypes()` check that was never meant to
describe DeepSpeed's own int8 weight-quantization dtype. `docs/_tutorials/inference-tutorial.md`
("Datatypes and Quantized Models", rewritten as recently as #8535) still documents int8 as
supported.

Found while investigating #1454 — the `compute_attention()`/`input_mask` crash reported there is
already fixed upstream, but the underlying "run int8 quantized inference" workflow that issue was
trying to use is broken by this separate, still-live gap.

**To Reproduce**
```python
import torch, deepspeed
deepspeed.init_inference(torch.nn.Linear(4, 4), dtype=torch.int8)
```
No GPU needed — the gate itself is device-agnostic and raises identically on CPU.

**Expected behavior**
`init_inference` should accept `dtype=torch.int8` as it did before #6528, since DeepSpeed's own
int8 weight-quantization path (`DeepSpeedSelfAttention`/`DeepSpeedMLP` in
`deepspeed/ops/transformer/inference/ds_attention.py`) still special-cases `torch.int8`
construction today — that code is presently unreachable dead code because the engine can no
longer be constructed with this dtype at all.

**System info:**
 - Reproduces identically on every accelerator backend (`cpu`, `cuda`, `hpu`, `mlu`, `xpu`) since
   the dtype check is evaluated before any accelerator-specific code runs.
 - Python 3.12, torch 2.14 (CPU repro); not version-specific — same gate on current `master`.

**Additional context**
A one-line fix (exempt `torch.int8` from the `supported_dtypes()` gate, matching the code's
pre-#6528 scope) plus a regression test is ready; opening a PR alongside this issue.

udsy19:
Fixes #8577

Related to #1454 (the original `compute_attention()`/`input_mask` TypeError reported there is
already fixed upstream — `deepspeed/ops/transformer/inference/ds_attention.py` has required
`input_mask` on `compute_attention()` for years — so this does not use a `Fixes #1454` keyword;
it closes a different, still-live gap in the same int8/quantized-inference workflow the issue is
about, filed separately as #8577).

## Problem
`InferenceEngine.__init__` (`deepspeed/inference/engine.py:74`) rejects any `dtype` not in
`get_accelerator().supported_dtypes()`. That per-accelerator list — checked across every
implementation that defines it (cpu, cuda, hpu, mlu, xpu) — has never included `torch.int8`. So
`deepspeed.init_inference(model, dtype=torch.int8)` raises `ValueError: Data type torch.int8 is
not supported by <device> accelerator` unconditionally, on every accelerator DeepSpeed ships,
regardless of what hardware is actually running.

Impact: crash-on-valid-input. Who reaches it / triggered by: anyone following
`docs/_tutorials/inference-tutorial.md`'s own "Datatypes and Quantized Models" section (last
touched yesterday by #8535, which still asserts "DeepSpeed inference supports fp32, fp16 and int8
parameters" and "INT8 inference quantization remains supported"), or the `dtype=torch.int8`
fixture already present in `tests/unit/inference/test_checkpoint_sharding.py:54-58`. What is
observed: a hard `ValueError` at engine construction instead of a working (or even attempted)
int8 inference engine.

This is a regression, not a deliberate removal of int8 support:
- Before #6528 ("add bfloat16 to inference support dtypes", 2024-09-27), this same check only
  rejected an unsupported `torch.half`; it never gated `torch.int8` at all. #6528's own stated
  intent was narrowly "to allow running inference tasks using bfloat16" and its diff never
  mentions int8.
- A separate, since-deleted `_validate_args()` explicitly listed
  `supported_dtypes = [None, torch.half, torch.int8, torch.float, torch.bfloat16]` right up until
  it was removed as unused dead code by #5505 (2025-01-07) — independent evidence int8 was still
  considered valid after #6528 landed, just checked by a function nobody called.
- `DeepSpeedSelfAttention`/`DeepSpeedMLP` (`deepspeed/ops/transformer/inference/ds_attention.py`)
  still special-case `config.dtype == torch.int8` for parameter dtype construction today — dead
  code for a dtype the engine can no longer be constructed with.
- `tests/unit/inference/test_checkpoint_sharding.py`'s `dtype` fixture already defensively
  `pytest.skip`s when `dtype not in get_accelerator().supported_dtypes()`, which, given the
  enumeration above, means its `int8` parametrization has been silently skipped on every
  accelerator since #6528, quietly losing that coverage.

## Fix
Exempt `torch.int8` from the `get_accelerator().supported_dtypes()` gate, restoring the pre-#6528
scope of this one check. This does not touch the accelerator capability lists themselves (they
describe genuine hardware tensor-dtype support and are used elsewhere too) — it only stops that
list from also gating DeepSpeed's own int8 weight-quantization dtype, which is not what it was
written to describe.

## Testing
Added `test_int8_dtype_accepted` to `tests/unit/inference/test_inference_config.py::TestInferenceConfig`.

Negative control (revert only `deepspeed/inference/engine.py` to `upstream/master`, keep the new
test):
```
FAILED ...test_int8_dtype_accepted - ValueError: Data type torch.int8 is not supported by cpu accelerator
1 failed
```
Restored: `1 passed`. Full `test_inference_config.py`: unchanged pre-existing results before/after
this diff (10 passed, 1 pre-existing failure — `TestInferenceCudaGraphConfig::test_cuda_graph_with_kernel_inject_raises`
fails identically on unmodified `upstream/master` because this CPU lacks `torch.float16` support,
unrelated to `torch.int8`/this change).

`yapf`/`flake8`/`codespell` clean on both changed files.

Signed-off-by: Udaya Tejas <udayatejas2004@gmail.com>


deepspeed/inference/engine.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, model, config):
         if hasattr(self.module, "config"):
             TransformerPolicy.hf_model_config = self.module.config
 
-        if config.dtype not in get_accelerator().supported_dtypes():
+        if config.dtype not in get_accelerator().supported_dtypes() and config.dtype != torch.int8:
```
ebarkhordar, line 74: I ran this against `e57bdeda67e6430658c2f8cd542331544cd0f6cd` in a clean `python:3.11-slim` container, CPU accelerator, torch 2.14.0+cpu.

The gate change does let int8 through, but the path it opens does not quantize. `_convert_to_dtype` (`engine.py:517`) still keeps the int8 branch behind `if False:`, and no remaining branch matches int8, so the module keeps the dtype it arrived with:

```
dtype=torch.int8      config.dtype=torch.int8       weight: torch.float32 -> torch.float32
dtype=torch.bfloat16  config.dtype=torch.bfloat16   weight: torch.float32 -> torch.bfloat16
```

bfloat16 is a control, so the weight read is not just always printing float32. On base `fe8c4b10` the same script raises `ValueError: Data type torch.int8 is not supported by cpu accelerator`. So with no checkpoint and `replace_with_kernel_inject` at its default `False`, this swaps a loud error for an engine reporting `_config.dtype == torch.int8` over fp32 weights.

Both places that do act on int8 are unreached there: `engine.py:480` inside `_load_checkpoint`, and `replace_module.py:201` inside `replace_transformer_layer`, which `__init__` calls only for one of its three injection modes (lines 135-169).

One correction to the body: `279bf743d` does delete the `if False:` block, but it is not an ancestor of master, so it never landed. `git blame` still puts that line at `b5d18a6ab`.

`test_int8_dtype_accepted` asserts `_config.dtype` only, which passes whether or not anything quantized. Asserting the weight dtype the engine produces would pin the real behaviour, and if that is fp32 then narrowing the exemption to the cases that consume int8 seems better than opening it for all.

CPU only, no checkpoint, no kernel injection.

ebarkhordar: Thanks for picking this up. I ran ce500b21 in a clean container (python:3.11-slim, torch 2.14.0+cpu, `pip install .` from the PR head, reported version 0.19.8+ce500b21e). The narrowed gate refuses one configuration that does reach the quantizer.

`tensor_parallel.mpu` is a fourth way into AutoTP. The gate at engine.py:84 reads `config.tensor_parallel.tp_size`, but an mpu caller does not set that field: it is written at engine.py:125 from `dist.get_world_size(group=self.mpu.get_model_parallel_group())`, after the gate has already raised. So `init_inference(model, dtype=torch.int8, tensor_parallel={"mpu": mpu})` is rejected even when that group would have given tp_size > 1 and mode 3 would have run.

Measured on 2 ranks (gloo, world_size 2, an mpu stub returning the default group). Only dtype differs between the two runs:

```
world_size = 2
[bf16 + mpu] AssertionError: Not able to determine model policy automatically.
             Please provide policy.          <- auto_tp.py:260, inside AutoTP.tp_parser
[int8 + mpu] ValueError: Data type torch.int8 requires kernel injection or a
             replacement policy ...
```

The bf16 run gets past the gate and into the AutoTP branch. The int8 run never gets there. That assertion is my toy model having no AutoTP policy, which is what makes it a usable marker that the branch ran at all.

Your new test docstring already names the case: "AutoTP (tensor_parallel.tp_size > 1 or tensor_parallel.mpu)". The guard checks only the first half. One more term covers it:

```python
if not (config.injection_policy or config.replace_with_kernel_inject
        or config.tensor_parallel.tp_size > 1 or config.tensor_parallel.mpu):
```

I did not run a real model-parallel mpu, so what tp_size becomes for a genuine TP group is read from engine.py:125 rather than executed.

Issue #8489, opened 2026-09-11
Deprecate unused DeepSpeed features


sfc-gh-truwase:
## Summary

Tracking issue for retiring unused / unmaintained DeepSpeed features so the runtime, docs, tests, and packaging stay consistent.

Goal: remove code paths that are no longer supported, then sweep the leftover docs, extras, CODEOWNERS, and tests so they cannot be re-enabled by accident.

- Landed in [#8490](https://github.com/deepspeedai/DeepSpeed/pull/8490) (`71d316d60`, 2026-09-12): **57 files, +169 / −8,298**.
- Follow-up (local, not yet merged): MoQ / eigenvalue, 1-bit comm backends, Progressive Layer Dropping, Apex AMP, and ZeRO-3 elastic checkpoint. Working tree vs `master`: **48 files, +104 / −2,814**. Combined vs pre-#8490: **95 files, +253 / −11,092**.

## Coverage statistics

Line counts are from `git diff --numstat`. **Code** = runtime/library sources; **tests** = `tests/`; **docs** = `docs/` plus `CODEOWNERS` / issue templates / CI where noted.

### Removed in #8490

| Feature | Code − | Tests − | Docs − | Net |
|---|---:|---:|---:|---:|
| Compression library | 2,430 | 301 | 455 | −3,166 |
| 1-bit / 0-1 Adam optimizers | 1,132 | 1,255 | 745 | −3,131 |
| MiCS ZeRO-3 | 650 | 225 | 34 | −909 |
| Nebula checkpoint engine | 229 | 0 | 0 | −229 |
| Runtime wiring (`config.py` / `engine.py` / launchers / CODEOWNERS) | 264 | 0 | 3 | −208 |
| Shared docs (config-json, training pages, nav) | 0 | 0 | 513 | −506 |
| Shared tests (config dict, AutoEP MiCS gate, ZeRO) | 0 | 62 | 0 | −19 |
| Helper relocation (`recursive_getattr` → `module_utils`) | 0 | 0 | 0 | +39 |
| **PR total** | | | | **+169 / −8,298** |

### Follow-up sweep (this remove effort, not yet merged)

| Feature | Code − | Tests − | Docs − | Notes |
|---|---:|---:|---:|---|
| 1-bit comm backends | 717 | 676 | 0 | `runtime/comm/{nccl,mpi,compressed,hccl}.py`, `runtime/compression/`, `tests/onebit/`, extras `1bit` / `1bit_mpi` |
[...]

PR #8579, opened 2026-09-18
Deprecate and remove graph harvesting

hwchen2017:
Deprecate and remove graph harvesting as part of #8489 
• Remove the  graph_harvesting  config option; supplying it (any value, including  false / null ) now raises  DeepSpeedConfigError , consistent with other removed features.
• Remove  graph_process / graph_cache  and the  use_graph  argument from  get_global_norm_of_tensors  /  clip_tensors_by_global_norm , collapsing them to the direct (non-graph) path.
• Optimize global-norm computation: previously each tensor's p‑th power was accumulated in-place into a reused, cached scratch buffer ( graph_cache['norm_tensors_compute_buffer'] ) so memory addresses stayed fixed for graph replay. It now computes each tensor's norm into a fresh list, stacks them, raises to the p‑th power, and sums after the model/expert-parallel all-reduce). 
• Drop the  graph_harvesting  plumbing in  BF16_Optimizer ,  DeepSpeedEngine , and  constants.py ; simplify  update_hp_grads  (the graph-only CPU flag-fixup loop was redundant).
• Add a config-rejection test.


pengdurice:
feel free to resolve the conflicts and lgtm.

Issue #8367, opened 2026-08-30
[Feature] Per-Head Muon: per-head Newton-Schulz orthogonalization for attention projections


delock:
## Motivation

DeepSpeed now supports the Muon optimizer across ZeRO stages 1/2/3. In 2026, two frontier labs independently converged on the same refinement for attention projections: **orthogonalizing per attention head rather than over the full projection matrix**.

- **Kimi K3** ([arXiv:2607.24653](https://arxiv.org/abs/2607.24653), §2.5, "Per-Head Muon"): partitions the momentum of Q/K/V projections along the head dimension and orthogonalizes each head's block separately. Their motivation: full-matrix orthogonalization treats all heads as a single coupled block, so heads with larger gradient/momentum scales dominate the shared update direction while smaller-scale heads receive insufficiently normalized updates. Per-head orthogonalization equalizes update scale across heads, improves stability at larger scales, and is *cheaper* (Newton–Schulz on tall per-head blocks costs less than on the full projection).
- **Zhipu GLM-5 "Muon Split"** ([arXiv:2602.15763](https://arxiv.org/abs/2602.15763)): the same idea applied to MLA up-projection matrices (split by attention head, orthogonalize each head independently), closing the performance gap between MLA and GQA under Muon.

Today in DeepSpeed, `muon_update()` (`deepspeed/runtime/zero/muon/original_muon.py`) applies Newton–Schulz to the full 2D parameter matrix — i.e., the "coupled block" behavior both papers describe.

## Proposal

1. **Optimizer kernel**: add a per-head mode to `muon_update` — reshape the update/momentum to `[..., num_heads, head_dim]` and run the existing bmm-based Newton–Schulz on the batched per-head blocks (the NS implementations already support batched inputs, so this is mostly a view/reshape branch). Apply the existing `max(1, m/n)**0.5` scaling per head block.
2. **Parameter metadata pipeline**: tag attention projection parameters with head structure (`num_heads`, `head_dim`, fused-vs-split QKV layout), similar to the existing `use_muon` tagging. The AutoTP model scan / presets already identify attention head structure, so this metadata can be sourced from there.
3. **ZeRO integration**: the engine call site (post-unflatten, per-parameter `muon_update`) stays unchanged — just pass the head shape so the update can be viewed per-head and reshaped back before being applied to the flat buffer.
[...]

PR #8580, opened 2026-09-18
[muon] Per-head Newton-Schulz for fused QKV projections

0z5a:
**Incremental review:** [Files changed](https://github.com/deepspeedai/DeepSpeed/pull/8580/files) against `master`.

Implements the fused-QKV part of #8367, on top of #8384.

## What this PR does

A fused QKV weight holds Q, K and V in one matrix. #8384 left it on the full-matrix path because the three sections do not share a head count under GQA. They do not have to.

**Every head is `head_dim` contiguous rows of dim 0 in both layouts that ship:**

- sectioned (`cat([q_proj, k_proj, v_proj])`, what most training stacks store): `n_q` query heads, then `n_kv` key heads, then `n_kv` value heads, each `head_dim` rows;
- interleaved per KV group (Falcon's `view(..., num_kv_heads, num_heads // num_kv_heads + 2, head_dim)`, GPT-NeoX's per-head `q, k, v`): each group is `n_q/n_kv` query heads, one key head, one value head, each `head_dim` rows.

Both layouts present the same row partition in a different order, and per-head Newton-Schulz is block-wise with the same `max(1, head_dim / in_features)**0.5` scale on every block - so the update does not depend on which layout the weight uses, and the Q/K/V labelling never enters it. GQA changes how many heads a section holds, not where the blocks are. CodeGen is covered rather than mis-split for the same reason: its `qkv_proj` is built by a four-block reshape, which still lands on head boundaries whenever 4 divides `num_heads`.

**The step size stops depending on how the checkpoint stores the projection.** The full-matrix path multiplies the orthogonalized update by `max(1, rows / cols)**0.5`, so a fused `[3h, h]` qkv is scaled by sqrt(3) while separately stored q/k/v are each scaled by 1.0 - the same logical projection gets a Muon step that differs by that factor depending only on its storage layout. Per-head scales every block by `max(1, head_dim / in_features)**0.5`, which is 1.0 in a fused or a split matrix alike, so after this change the two layouts agree. This is the stronger argument for the change; the coverage one is secondary.

So a fused weight needs no section bookkeeping: it is tagged with `rows // head_dim` head blocks, which for 8 query heads over 2 KV heads is 12 rather than 8 or `3 * 8`. The candidate requires the **exact** fused total `(num_attention_heads + 2 * num_kv_heads) * head_dim`, which is also what declines a transposed `Conv1D` weight (GPT-2's `c_attn`, `[hidden, 3*hidden]`, whose dim 0 is the input axis).

The tag rides on `muon_num_heads`, so `muon_update`, the six ZeRO/DDP call sites and the config surface need no new API.

## Not in this PR

- **Falcon's `query_key_value` stays on the full-matrix path**, and a test pins why: its config reports `num_attention_heads` key/value heads while `multi_query=True` makes the module build a single one, so the config's fused total (192 rows) does not describe the weight (80). That is a pre-existing `AutoTPMeta` reading rather than something the per-head split introduces; the exact-total check sends it to the full-matrix path instead of splitting the wrong blocks.
- **A fused leaf under a module listed in `_LINEAR_ATTENTION_OWNERS` stays on the full-matrix path**, and a test pins that too. #8436's whitelist says a listed module's own geometry, not the config's, decides for its projections - and that path answers only for query and KV. A fused leaf under a listed owner therefore has no owner candidate, and the config candidates must not be consulted for it, which is the coincidence the whitelist exists to prevent: Kimi-K3's config describes the model's MLA layers (8 heads x 74), so the fused total it proposes is 1776 rows, and a fused `in_proj_qkv` of exactly that width would otherwise be tagged with 24 config-shaped blocks. Every fused leaf was declined before this tagger existed, so the outcome is unchanged for the one architecture whose geometry the config does not describe. The owner path has no fused candidate to offer either: a fused matrix only splits into uniform blocks when its q, k and v widths agree, which a listed owner does not promise (`KimiDeltaAttention` carries `head_k_dim` and `head_v_dim` separately).
- **AutoTP.** It repacks a fused weight into this rank's piece of each section, or interleaves it, so a rank's layout is not asserted by this change. The data-parallel path is what is trained and measured below.

## Tests

```
$ pytest tests/unit/runtime/zero/test_per_head_muon.py                     # CPU
83 passed

$ pytest tests/unit/v1/ops/muon/test_per_head_muon_accelerator.py          # 2 x 5090, ZeRO 1/2/3
26 passed

$ pre-commit run --files <changed files>
all hooks passed
```

CPU coverage added: fused tagging under GQA (`8` query heads over `2` KV heads is `12` blocks, not `8`), the exact-total guard, the `Conv1D` decline, the real architectures, and sectioned/interleaved equivalence on a real gradient - the test builds one fused weight in each layout and requires the updates to be the same permutation of each other, which is the property the change rests on. It also covers the listed-owner decline: a fused projection that confirms on the config's fused total is still declined, the same shape on an unlisted module is still tagged, and the decline is reported as settled rather than as a name that went unmatched. Accelerated coverage adds a fused training model across ZeRO 1/2/3 at world_size 2.

The shard-equivalence bound is the Newton-Schulz compute dtype's own epsilon rather than bitwise identity. Orthogonalizing 8 heads in one batch and two batches of 4 is the same computation either way, but the batched GEMM rounds differently at different batch sizes: `gram` iterates in fp16 and came out bitwise equal, while `standard` iterates in bf16 and landed 4.2e-3 apart in relative norm (measured on an RTX 5090, torch 2.13). The whole-matrix path is three orders of magnitude further out, which the test right after it pins at more than 0.1 relative.

## End-to-end training

2 x NVIDIA L20 46 GB, ZeRO-2, bf16, torch 2.10.0+cu128, dp=2. A 17,177,088-parameter LLaMA-shaped model whose attention holds Q|K|V in one `qkv_proj` (6 layers, hidden 512, 8 query heads over 2 KV heads, `head_dim` 64), 120 optimizer steps on 983,040 tokens of real text, global batch 16, both arms from an identical initialisation.

| arm | mean loss (first 10) | mean loss (last 10) | tok/s | peak GiB |
| `per_head_muon: false` | 4.0266 | 2.6789 | 85,411 | 1.68 |
| `per_head_muon: true` | 4.0227 | 2.6781 | 82,368 | 1.77 |

- With the flag on, `blocks.*.qkv_proj.weight` carries `muon_num_heads = 12` and is the only tagged parameter; with the flag off every tag is `None`.
- One optimizer step from the shared initialisation moves **34/34** trainable weights differently between the arms (fused-weight L1 difference 0.235 against a weight L1 of 8679.4), so the per-head branch executes in the real training path instead of being tagged and skipped.
- On a real backward-pass gradient for the fused weight, the per-head update norms land within **1.30x** of each other where the full-matrix ones span **2.78x**; the two updates differ by 2.46 in relative norm.

The two loss curves are indistinguishable, and that is expected rather than a result: both arms start from an untrained initialisation whose heads are interchangeable, so the split has nothing to separate - the same outcome #8384 reported for its models. The mechanism measurement is what shows the feature does something; a convergence claim needs trained checkpoints and is not made here.


alanhuangyoo (commented):
**Owner geometry is lost for a fused leaf once this rebases on #8436 (blocking)**

`_owner_candidate` only answers for QUERY and KV, so a module in `_LINEAR_ATTENTION_OWNERS` with a fused projection gets `[]` and falls back to the config candidates — and `in_proj_qkv` is in `_FUSED_QKV_LEAVES`. Those leaves were declined outright before this PR, so this is the coincidence the whitelist was added to prevent. Either extend the owner path to the fused kind, or decline fused leaves under a listed owner. The rebase itself is clean — I ran it locally, 106 tests pass on 2 GPUs.

**Fused and separate layouts now agree (non-blocking, worth putting in the description)**

The full-matrix path scales a fused qkv by `max(1, rows/cols)**0.5`, up to sqrt(3) for `[3h, h]`, while separately stored q/k/v each get 1.0. Today the step size depends on how the checkpoint happens to store the projection; per-head makes both 1.0. That's a correctness argument, stronger than the coverage one.

Also checked CodeGen — its 4-block reshape still lands on head boundaries when 4 divides num_heads, so it's covered rather than mis-split.

0z5a:
Both points are addressed in `7325f2a` (plus `e52efde` for a test bound), on top of a rebase onto current master.

**Blocking: owner geometry for a fused leaf.** You were right that the rebase makes this reachable. A fused leaf under a module in `_LINEAR_ATTENTION_OWNERS` used to be declined outright; after this branch it classified as `FUSED_QKV`, `_owner_candidate` answered `[]` for it because that path only covers QUERY and KV, and it fell through to the config candidates. On Kimi-K3's config (which describes the model's MLA layers, 8 heads x 74) a fused `in_proj_qkv` of 1776 rows was tagged with **24 config-shaped blocks** - measurable before the fix with `_attention_head_count` on that shape.

I took the second option you offered: a fused leaf under a listed owner is now declined. The owner path has no fused candidate to give either - a fused matrix only splits into uniform blocks when its q, k and v widths agree, and a listed owner does not promise that (`KimiDeltaAttention` carries `head_k_dim` and `head_v_dim` separately) - so declining keeps the whitelist's guarantee and leaves that one architecture exactly where it was before any fused tagging existed. The decline is a settled reason, so it is not reported as an attention name that went unmatched. Three tests pin it: the decline with its reason, the same shape on an **unlisted** module still tagged from the config, and the reporting being quiet.

**Non-blocking: the step-size argument.** Added to the description, and it is the reason I would lead with too. The full-matrix path scales by `max(1, rows / cols)**0.5`, so a fused `[3h, h]` qkv gets sqrt(3) while separately stored q/k/v each get 1.0 - the same logical projection takes a different Muon step depending only on how the checkpoint stores it. Per-head scales every block by `max(1, head_dim / in_features)**0.5`, which is 1.0 in both layouts, so they now agree. CodeGen is in the description as covered: its four-block reshape still lands on head boundaries when 4 divides `num_heads`.

**One thing I found while validating, which is why there is a third commit.** `test_per_head_on_a_shard_is_the_shard_of_per_head` required `torch.equal` between orthogonalizing 8 heads in one batch and two batches of 4. That is bitwise on this hardware for `gram` (fp16), but not for `standard` (bf16), where the batched GEMM picks a different kernel at a different batch size: measured 4.2e-3 in relative norm, max absolute 6.3e-3 against a result whose mean magnitude is 4.3e-2, on an RTX 5090 with torch 2.13. The computation is unchanged - each head is orthogonalized against itself either way - so the bound is now the compute dtype's own epsilon (`ns_compute_dtype`), and the whole-matrix path is still pinned at more than 0.1 relative by the test after it. Worth knowing that the L20 run in the description passes bitwise but other hardware does not.

Re-validated on the rebased head: `tests/unit/runtime/zero/test_per_head_muon.py` 83 passed; `tests/unit/v1/ops/muon/test_per_head_muon_accelerator.py` 26 passed on 2 x RTX 5090 across ZeRO 1/2/3; yapf, codespell and the rest of the changed-file hooks clean.

alanhuangyoo (commented):
Pulled 7325f2a, 109 pass on 2 GPUs. Declining is the right call since KimiDeltaAttention's k and v widths don't have to match. LGTM.

PR #8583, opened 2026-09-18
fix(rollout): reject shared prefill with continuous batching

### LINKED ISSUE #8458: [BUG] Do not silently ignore use_shared_prefill with continuous batching
## Context

`HybridEngineRollout` supports both shared prompt prefill and continuous
batching, but their interaction is currently undefined.

When continuous batching is enabled through
`SamplingConfig.continuous_batch_size`, the continuous path is selected before
the shared-prefill logic. As a result, `HybridEngineRolloutConfig.use_shared_prefill`
is silently ignored.

Continuous batching currently requires `n_samples_per_prompt=1`, so shared
prefill does not provide a clear benefit in this mode. However, silently
ignoring the configuration can be confusing and may hide an invalid combination
of options.

## Proposed work

Define and document the behavior of `use_shared_prefill` when continuous
batching is enabled.

Possible choices include:

- explicitly reject the combination with a clear `ValueError`; or
- define and implement compatible shared-prefill semantics for continuous
  batching.

## Acceptance criteria

- The interaction between `use_shared_prefill` and continuous batching is
  explicitly defined.
- The configuration is not silently ignored.
- Tests cover both compatible and incompatible configurations.
- The default non-continuous generation path remains unchanged.

Follow-up from DeepSpeed PR #8368. cc @delock

modelpath-dev:
## Why

`HybridEngineRollout.generate` selects the continuous path whenever `SamplingConfig.continuous_batch_size` is set, before any shared-prefill logic runs. With `HybridEngineRolloutConfig.use_shared_prefill=True`, that flag was therefore ignored with no error. This PR rejects the combination explicitly, matching the approach already taken for CUDA graph capture on the continuous path.

Fixes #8458.

## Scope

- `HybridEngineRollout._validate_continuous_inputs`: raise `ValueError` when `use_shared_prefill` is enabled.
- Unit test for the incompatible pair.
- `docs/code-docs/source/inference-engine.rst`: document that shared prefill and continuous batching cannot be combined.

Out of scope: implementing shared-prefill semantics inside continuous batching.

## Tradeoffs

Rejection over a compatible shared-prefill continuous mode. Continuous batching already requires `n_samples_per_prompt=1`, so shared prefill has no clear benefit there today.

## Blast Radius

Only the experimental continuous path changes, and only when shared prefill is also requested. Default `generate()` and shared-prefill-only configs are unchanged.

## Verification

Before the fix, `_validate_continuous_inputs` accepted `use_shared_prefill=True` with `continuous_batch_size=1` (silent ignore). After the fix it raises `ValueError: continuous batching does not support shared prompt prefill`.

```
TRANSFORMERS_NO_TF=1 PYTHONPATH=. python3 -m pytest \
  tests/unit/runtime/rollout/test_hybrid_engine_rollout.py::test_continuous_generation_rejects_shared_prefill \
  tests/unit/runtime/rollout/test_hybrid_engine_rollout.py::test_continuous_generation_rejects_unsupported_inputs \
  tests/unit/runtime/rollout/test_hybrid_engine_rollout.py::test_shared_prefill_rejects_graph_capture \
  tests/unit/runtime/rollout/test_hybrid_engine_rollout.py::test_generate_uses_shared_prefill_for_multiple_samples \
  -q
# 4 passed
```


deepspeed/runtime/rollout/hybrid_engine_rollout.py
```diff
@@ -510,6 +510,8 @@ def _validate_continuous_inputs(self, requests, sampling, max_batch_size):
             raise ValueError("max_batch_size must be positive")
         if self.use_graph_capture:
             raise ValueError("continuous batching does not yet support CUDA graph capture")
+        if self.use_shared_prefill:
```
nathon-lee: Thanks for addressing this configuration conflict and for adding the regression test. Could we include the exact configuration field in the error message to make the failure more actionable?

For example:

```python
raise ValueError(
    "continuous batching does not support "
    "HybridEngineRolloutConfig.use_shared_prefill"
)
```

The current message is understandable, but naming the option explicitly would help users quickly identify which setting needs to be changed.

delock:
Hi @modelpath-dev can you fix the DCO failure?  Thanks!

tests/unit/runtime/rollout/test_hybrid_engine_rollout.py
```diff
@@ -96,6 +96,21 @@ def test_continuous_generation_rejects_unsupported_inputs():
         rollout.generate(request, SamplingConfig(max_new_tokens=2, temperature=0.5, continuous_batch_size=1))
 
 
+def test_continuous_generation_rejects_shared_prefill():
```
delock, line 99: I don't think we need a test here.   If there is behavior change, then this test will need change again, which is not desired.

Issue #3081, opened 2023-03-22
[BUG] Deepspeed Inference Error For ~3800 Context Length and Higher


benathi:
**Describe the bug**
When the context length exceeds 3778, the code breaks. Using 64 generated tokens. Deepspeed version 0.8.0.

Error: 

[launch.py:318:sigkill_handler] Killing subprocess

According to an early investigation, attn_softmax_v2 seems to be the kernel causing errors

**To Reproduce**

GPT2 model with random weights and long context

**Expected behavior**
The code should run without process being killed

**ds_report output**

--------------------------------------------------
DeepSpeed C++/CUDA extension op report
--------------------------------------------------
NOTE: Ops not installed will be just-in-time (JIT) compiled at
      runtime if needed. Op compatibility means that your system
      meet the required dependencies to JIT install the op.
--------------------------------------------------
JIT compiled ops requires ninja
ninja .................. [OKAY]
--------------------------------------------------
op name ................ installed .. compatible
--------------------------------------------------
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
 [WARNING]  async_io: please install the libaio-devel package with yum
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
async_io ............... [NO] ....... [NO]
cpu_adagrad ............ [NO] ....... [OKAY]
cpu_adam ............... [NO] ....... [OKAY]
fused_adam ............. [NO] ....... [OKAY]
fused_lamb ............. [NO] ....... [OKAY]
quantizer .............. [NO] ....... [OKAY]
random_ltd ............. [NO] ....... [OKAY]
sparse_attn ............ [NO] ....... [OKAY]
spatial_inference ...... [NO] ....... [OKAY]
transformer ............ [NO] ....... [OKAY]
stochastic_transformer . [NO] ....... [OKAY]
transformer_inference .. [NO] ....... [OKAY]
utils .................. [NO] ....... [OKAY]
--------------------------------------------------
DeepSpeed general environment info:
torch install path ............... ['/home/ec2-user/anaconda3/envs/vectorinf3/lib/python3.8/site-packages/torch']
torch version .................... 1.13.1
deepspeed install path ........... ['/home/ec2-user/research/DeepSpeedInt8Upgrade/deepspeed']
deepspeed info ................... 0.8.0+unknown, unknown, unknown
torch cuda version ............... 11.7
torch hip version ................ None
[...]

PR #8592, opened 2026-09-19
Check input + max_new_tokens against max_out_tokens in InferenceEngine.generate

udsy19:
**Impact: crash-on-valid-input**

Entry point / who reaches it: the documented `deepspeed.init_inference()` +
`engine.generate()` workflow (DeepSpeed's own inference tutorial calls
`ds_engine = deepspeed.init_inference(model, ...)` then `ds_engine.generate(...)`).
`InferenceEngine.__init__` patches `self.generate = self._generate` whenever the
wrapped model has a `.generate` method, so this is the public, user-facing entry
point, not an internal helper.

How it's triggered: `max_out_tokens` sizes the fixed KV-cache workspace
(`kv_cache_size = [batch, heads, max_out_tokens, head_size]`,
`deepspeed/ops/transformer/inference/op_binding/workspace.py:45`) for the WHOLE
generation, input plus newly generated tokens. #4349 added a check in
`_generate()` (`deepspeed/inference/engine.py:632-639`) that raises a clear
`RuntimeError` when the input alone already exceeds `max_out_tokens`, but never
adds `max_new_tokens` before comparing. A caller who raises `max_tokens` to fit
a long input, then requests generation that pushes input+new past that same
budget, sails through unguarded into the wrapped module's `generate()` and
writes past the fixed-size buffer instead of the clear error #4349 already
gives for the input-only case — undefined behaviour, exactly what #4349
itself calls "segfaults, garbage output, and generally bad behavior". Related
to #3081 ("~3800 context length" + "64 generated tokens" crashing inference):
same defect shape, in the same guard.

```
$ pytest tests/unit/inference/test_inference_config.py::test_generate_rejects_input_plus_max_new_tokens_over_budget -m inference
# guard reverted to upstream/master:
1 failed
Failed: DID NOT RAISE RuntimeError
# with the fix:
1 passed
```

Fix: `_generate()` now reads `max_new_tokens` from kwargs (or
`generation_config.max_new_tokens`) and adds it to the input length before
comparing against `max_out_tokens`; input-only behaviour is unchanged when
`max_new_tokens` is absent.

Negative control: reverting only `deepspeed/inference/engine.py` to
`upstream/master` (test file untouched) reproduces the failure above;
restoring the fix passes. No other test in this file changed behaviour.

Signed-off-by: Udaya Tejas <udayatejas2004@gmail.com>


ebarkhordar (commented):
I ran this at `21374033` in a clean container (python:3.11-slim, torch CPU, transformers 5.17.0, `pip install .` from the branch), driving your own `GenerateStub` with `max_out_tokens: 100`.

The new check fires for `max_new_tokens`, but `max_length` carries the same budget and is not read, so the identical overflow goes through when a caller writes it the other way:

```
max_new_tokens=20  (the PR's own case)    -> RuntimeError: Input with size 90 and max_new_tokens 20 together exceed maximum length of 100.
max_length=4000    (same overflow, total) -> NO RAISE, returned 'reached-generate'
```

`max_length` is already a total of input plus new tokens, so it wants a wider comparison at `engine.py:641` rather than a second branch: the larger of `tensor_length + max_new_tokens` and `max_length`.

Separately, your `generation_config` fallback at `:637` cannot be reached. Passing any `GenerationConfig` raises `TypeError: '>' not supported between instances of 'NoneType' and 'int'` at `:628` first, because `num_beams` at `:624` comes back `None` on transformers 5.x. The test only passes `max_new_tokens=` directly, so nothing covers that path. `getattr(gen_config, "num_beams", 1) or 1` would reach it, and the pin is `transformers>=4.32.1` with no upper bound.

Also small: the input-only path now reports "and max_new_tokens None together exceed" when no generation kwarg was passed. That is #4349's case and the old wording read better for it.

I did not run a real model, so this is about the guard, not what the workspace does past it.

deepspeed/inference/engine.py
```diff
@@ -630,12 +630,32 @@ def _generate(self, *inputs, **kwargs):
...
+                if max_new_tokens is None:
+                    max_new_tokens = getattr(gen_config, "max_new_tokens", None)
+                if max_length is None:
+                    max_length = getattr(gen_config, "max_length", None)
             for input_tensor in kwargs["input_ids"]:
                 tensor_length = input_tensor.shape[-1]
-                if tensor_length > self._config.max_out_tokens:
+                total_length = tensor_length if max_new_tokens is None else tensor_length + max_new_tokens
+                if max_length is not None:
+                    total_length = max(total_length, max_length)
```
ebarkhordar: Thanks for picking both of these up so quickly. I ran the new commit in a clean container at `faeb85cd` (python:3.11-slim, torch 2.14.0+cpu, transformers 5.17.0), and the `max_length` half over-rejects when a caller sets both budgets.

`max(total_length, max_length)` takes the larger of the two, but transformers gives `max_new_tokens` precedence and ignores `max_length` when both are present, so the larger number is the one that will not be used. A call that fits now fails:

```
input=90  max_out_tokens=100  max_new_tokens=5  max_length=4000
  21374033 (before this commit) -> NO RAISE
  faeb85cd  -> RuntimeError: Input with size 90, max_new_tokens 5, and max_length 4000
               together exceed maximum length of 100
  transformers on the same inputs: "max_new_tokens will take precedence",
               out len 95, so 5 new tokens inside the budget of 100
```

The `num_beams` repair makes this reachable from a `generation_config` too, where it hit the TypeError first: `GenerationConfig(max_length=4096)` with `max_new_tokens=50` against `max_out_tokens=1024` raises, and that generation is really 140 tokens.

Mirroring the precedence instead of taking the max clears both, and your two new tests still pass on it:

```python
if max_new_tokens is not None:
    total_length = tensor_length + max_new_tokens
elif max_length is not None:
    total_length = max_length
else:
    total_length = tensor_length
```

I ran this on CPU against a stub module, so I have not exercised a real GPU generate path.

ebarkhordar: Thanks for turning that around so fast. I ran b718dfdd in a clean container (python:3.11-slim, torch 2.14.0+cpu, transformers 5.17.0) and the precedence part is right, but the snippet I handed you dropped `tensor_length` from the `max_length` branch, and that opens a gap I should have caught. `max_length` is a total, so when it is smaller than the input the guard now compares against a number below the input length and stops firing.

```
input=2000  max_out_tokens=1024  max_length=512
  21374033 -> RuntimeError
  faeb85cd -> RuntimeError
  b718dfdd -> NO RAISE, reaches generate()

same via generation_config=GenerationConfig(max_length=512): NO RAISE at b718dfdd
control, max_length=4000 on that input: still raises at all three
```

It is not a silent overflow on a real model: `_validate_generated_length` raises `ValueError: Input length of input_ids is 2000, but max_length is set to 512` immediately after. What goes missing is the engine's own error, the one that names `max_out_tokens`.

Keeping the max inside the branch restores it:

```python
elif max_length is not None:
    total_length = max(tensor_length, max_length)
```

On that, the two cases above raise again, `max_length=900` over an 800 token input still passes through, and the test file's results are unchanged: 8 passed and 6 failed both with and without it, the 6 being a missing c++ for `deepspeed_shm_comm` plus the CUDA only cuda-graph test on this CPU box.

Sorry for the round trip, that one is on my suggestion.

Issue #8439, opened 2026-09-06
Default gradient_clipping divides every Muon update by its own norm, shrinking the step by a model-sized factor


alanhuangyoo:
## Summary

`gradient_clipping` defaults to `1.0` (`GRADIENT_CLIPPING_DEFAULT`), and under Muon the quantity it clips is not the gradient — it is the orthogonalized update, whose norm is a property of the model's shape rather than of the gradient. Every Muon step is therefore divided by a constant that grows with the model, and the resulting step size is pinned by the clip threshold instead of by the learning rate. Nothing reports it.

## Measurement

Four square Linear layers, one step, ZeRO-1, fp32, `lr=0.02`, same gradient in every row. Only `gradient_clipping` differs: `0` disables it, `1.0` is what a config that does not mention it gets.

| hidden | reported grad_norm | \|Δw\| with clipping off | \|Δw\| at the default 1.0 | attenuation |
| --- | --- | --- | --- | --- |
| 128 | 4.2182 | 0.038865 | 0.009214 | 4.22x |
| 512 | 7.1564 | 0.058267 | 0.008142 | 7.16x |
| 2048 | 22.7588 | 0.174972 | 0.007688 | 22.76x |
| 4096 | 44.2460 | 0.339524 | 0.007674 | 44.25x |
| 8192 | 85.6024 | 0.676925 | 0.007908 | 85.60x |

Three things to note:

1. The attenuation equals the reported norm, to three digits, at every size. The clip is dividing the update by its own norm.
2. Unclipped, the step grows with the model, as Muon intends. Clipped, it is **flat at ~0.0077 across a 64x range of model width** — the step size is set by the clip threshold, not by Muon or by `lr`.
3. The factor grows with the model: 4x on a toy, 86x by hidden 8192, on a 4-layer MLP. A real model has far more matrices.

Identical numbers on ZeRO-1 and ZeRO-2. ZeRO-3 reports the same norm; I did not measure |Δw| there because the parameter is partitioned.

## Why it happens

Muon's update is **scale-invariant in the gradient**. Newton-Schulz normalizes its input and returns a near-orthogonal matrix, so the output depends on the gradient's direction and not its magnitude:

```
gradient scaled by 1      -> update norm 15.023504
gradient scaled by 0.001  -> update norm 15.024741
gradient scaled by 1000   -> update norm 15.029729
relative difference between the x1 and x1000 updates: 0.0055   (half-precision NS noise)
[...]

PR #8600, opened 2026-09-19
Run Muon once per optimizer step under ZeRO-3, not once per micro-batch

### LINKED ISSUE #8443: ZeRO-3 applies Muon's Newton-Schulz once per micro-batch, so gradient accumulation changes the optimizer
## Summary

Under ZeRO-3, Muon's Newton-Schulz runs once per **micro-batch** instead of once per optimizer step. With `gradient_accumulation_steps: n` the momentum buffer is advanced `n` times per step and the orthogonalization is applied to partial gradients rather than to the accumulated one. ZeRO-1 and ZeRO-2 are correct.

## Measurement

Two Linear layers (so two Muon matrices), two optimizer steps, counting calls into the Newton-Schulz kernels in `original_muon`. The correct count is 2 steps x 2 matrices = 4 in every row.

| stage | gas | Newton-Schulz calls | expected |
| --- | --- | --- | --- |
| 1 | 1 | 4 | 4 |
| 1 | 4 | 4 | 4 |
| 2 | 1 | 4 | 4 |
| 2 | 4 | 4 | 4 |
| 3 | 1 | 4 | 4 |
| 3 | 4 | **16** | 4 |

Stage 3 with `gas=4` does `4x` the work, one orthogonalization per micro-batch.

That is not only wasted compute. Same data, same seed, three optimizer steps, `momentum=0.95`, `gradient_clipping: 0`, and the same 8 samples per optimizer step either way — one micro-batch of 8 with `gas=1`, four of 2 with `gas=4`. The mean gradient is identical by construction, so the two runs must agree:

```
stage 2:  gas=1 vs gas=4   relative difference in the weights = 3.3e-04
stage 3:  gas=1 vs gas=4   relative difference in the weights = 1.0e-01
```

3.3e-04 is the half-precision Newton-Schulz noise floor — stage 2 is the same training run either way. Stage 3 is 300x that: a different one.

## Why

`stage3.py`:

```python
    def _apply_distributed_muon_update(self, communication_data_type, buffer_to_reduce):
        if not self.use_muon:
            return
        ...
```

and its only call site is inside the IPG bucket reduce path:

```python
        dist.all_reduce(buffer_to_reduce, group=process_group)
        ...
        self._apply_distributed_muon_update(communication_data_type, buffer_to_reduce)
        for param in params_in_bucket:
            grad = param.grad
```

That path runs on every micro-batch. There is no accumulation-boundary guard, and `use_muon` is the only condition. ZeRO-1/2 does the same work in `get_flat_partition`, which `ipg_epilogue` calls only under `if self.is_gradient_accumulation_boundary():` — which is why those two stages come out right.

The consequences, in order of how much they matter:

1. **The momentum decay is wrong.** `momentum.lerp_(grad, 1 - beta)` runs `n` times per optimizer step, so the effective retention is `beta**n` rather than `beta`. At `beta=0.95` and `gas=4` that is 0.81, and at `gas=16` it is 0.44 — the momentum the user configured is not the momentum they get, and the discrepancy moves with an unrelated knob.
2. **The orthogonalization sees partial gradients.** Newton-Schulz is not linear, so a sum of orthogonalized micro-batch gradients is not the orthogonalization of their sum. Since Muon's update is scale-invariant, each micro-batch contributes a unit-scale update regardless of how few samples it saw, which weights noisy micro-batches equally with the rest.
3. **`n` times the Newton-Schulz cost**, which for large matrices is the expensive part of the step.

ZeRO-3 with gradient accumulation is the standard configuration for a model large enough to need ZeRO-3, so this is the common path rather than a corner.

## What the fix needs

`is_gradient_accumulation_boundary()` already exists on the class, but guarding the call with it is not enough on its own: the current code takes the full-shape gradient out of `buffer_to_reduce` inside the reduce path, and on the boundary micro-step that buffer holds only that micro-step's contribution. Applying the update once, at the boundary, on the accumulated gradient means reading the accumulated partition instead — which is what ZeRO-1/2 does in `get_flat_partition`.

Happy to implement it if the shape of that is agreed. Flagging first because the change is inside the ZeRO-3 reduce path and is more than a guard.

## Reproduction

```python
import torch, deepspeed
import deepspeed.runtime.zero.muon.original_muon as om

NS = {"n": 0}
for name in ("zeropower_via_gram_newtonschulz", "zeropower_via_newtonschulz5"):
    f = getattr(om, name)
    setattr(om, name, (lambda f: (lambda *A, **K: (NS.__setitem__("n", NS["n"] + 1), f(*A, **K))[1]))(f))

model = torch.nn.Sequential(torch.nn.Linear(64, 64, bias=False), torch.nn.Linear(64, 64, bias=False))
GAS = 4
    model=model, model_parameters=model.parameters(),
    config={"train_micro_batch_size_per_gpu": 2, "gradient_accumulation_steps": GAS,
            "gradient_clipping": 0.0,
            "zero_optimization": {"stage": 3, "reduce_scatter": False},
            "optimizer": {"type": "Muon", "params": {"lr": 0.02}}})

for _ in range(2 * GAS):                      # two optimizer steps
    x = torch.randn(2, 64, device=engine.device)
    engine.backward(engine(x).square().sum())
    engine.step()
print("newton_schulz calls:", NS["n"])        # 16 on stage 3, 4 on stages 1 and 2
```

alanhuangyoo:
Fixes #8443.

## The problem

ZeRO-3 applies Muon inside the gradient reduce (`_apply_distributed_muon_update`, called from `__avg_scatter_contiguous_grads`), and that runs every micro-batch. With `gradient_accumulation_steps: n`, the momentum advances `n` times per optimizer step, and Newton-Schulz orthogonalizes each micro-batch's partial gradient instead of the accumulated one. ZeRO-1/2 apply Muon at the accumulation boundary and are correct.

On 2 GPUs, fp32, with the same 8 samples per step either way (one micro-batch of 8 at `gas=1`, four of 2 at `gas=4`), three steps, relative difference in the weights:

| | master | this PR |
| --- | ---: | ---: |
| ZeRO-2, `gas=1` vs `gas=4` | 3.6e-4 | 3.6e-4 |
| ZeRO-3, `gas=1` vs `gas=4` | **1.3e-1** | 3.6e-4 |
| Newton-Schulz calls, ZeRO-3, 2 matrices, 2 steps, `gas=4` | 16 | 4 |

ZeRO-3 now lands on exactly ZeRO-2's figure.

## The change

This is option 1 from the discussion in #8443. It is scoped to ZeRO-3 without optimizer offload.

- The reduce path no longer runs Muon when optimizer offload is off. The partitions accumulate the raw averaged gradient, as they do for every other optimizer.
- `step()` calls `_apply_muon_to_accumulated_grads()` after the overflow check and before the gradient norm. For each Muon sub-group, it:
  - all-gathers each parameter's accumulated gradient partitions, in chunks bounded by `reduce_bucket_size` as the reduce buckets were;
  - runs the existing round-robin Muon update once;
  - writes each rank's slice back into its partition.
- The per-sub-group body of `_apply_distributed_muon_update` is moved into `_muon_update_sub_group`. It takes the full-shape gradients explicitly, because at step time the parameters are partitioned and `param.grad` can't hold them. The reduce path calls it with `param.grad` as before.
- The gradient norm is still taken over the Muon update, as before. Clipping semantics are unchanged (#8439 / #7776 are separate).
- Because the update now runs after the overflow check, a step the loss scaler discards no longer touches the momentum. That is the ZeRO-3 counterpart of #8435.
- Collectives: each Muon parameter's gradient and momentum are gathered once per step instead of once per micro-batch. At `gas=n` that is `n` times fewer.

The optimizer-offload path is unchanged; #8464 is working on it. jinyouzhi added a pointer to this shape in #8464, and the overlap is limited to `_apply_distributed_muon_update`.

## Testing

On 2×H20:

- New file `tests/unit/v1/ops/muon/test_muon_zero3_grad_accum.py`. Both tests fail on master and pass here.
  - `test_newton_schulz_runs_once_per_matrix_per_step`: Newton-Schulz calls summed over ranks come to 2 × steps, not 2 × steps × gas.
  - `test_gradient_accumulation_matches_one_large_micro_batch[2, 3]`: `gas=1` and `gas=4` agree to within half-precision Newton-Schulz noise at both stages.
- `tests/unit/v1/ops/muon/` plus `tests/unit/runtime/zero/test_per_head_muon.py`: 260 passed.


tests/unit/v1/ops/muon/test_muon_zero3_grad_accum.py
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) Microsoft Corporation.
```
delock: should remove Microsoft copyright head.

deepspeed/runtime/zero/stage3.py
delock, line 1712: `param` is no longer defined, will always get a None here.

delock (changes requested):

deepspeed/runtime/zero/stage3.py
alanhuangyoo: Good catch, that was mine: pulling the grad from the list dropped the `param = params[base_i + rank]` line. Restored in 31f5e29. Added a test that takes one step with and without per-head under ZeRO-3 and checks it moves q and k but not the MLP. It fails without the fix.

tests/unit/v1/ops/muon/test_muon_zero3_grad_accum.py
alanhuangyoo: Removed, thanks.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1738,9 +1753,42 @@ def _apply_distributed_muon_update(self, communication_data_type: torch.dtype, b
...
+            partitions = self.averaged_gradients[i]
+            # Bound what is materialized at once, as the reduce path's buckets did.
+            start = 0
+            while start < len(group):
+                end, numel = start, 0
+                while end < len(group) and (end == start or numel + group[end].ds_numel <= self.reduce_bucket_size):
+                    numel += group[end].ds_numel
+                    end += 1
+                params, chunk = group[start:end], partitions[start:end]
+                full_grads = self._partitioned_buffers_all_gather(params, chunk, chunk[0].dtype)
```
delock: This all-gather should use `self.communication_data_type` instead of `chunk[0].dtype` — otherwise a user-configured fp16 comm dtype is silently overridden and the gather's communication volume and buffer footprint double.

deepspeed/runtime/zero/stage3.py
```diff
@@ -1738,9 +1753,42 @@ def _apply_distributed_muon_update(self, communication_data_type: torch.dtype, b
...
+                group_items = [(param, self.grad_position[self.get_param_id(param)][1], full_grad)
+                               for param, full_grad in zip(params, full_grads)]
+                self._muon_update_sub_group(i, group_items, chunk[0].dtype)
```
delock: Same for the momentum gather inside `_muon_update_sub_group`: the momentum buffer itself is stored in `communication_data_type` (`_create_momentum_buffer`), so gathering it in fp32 gains no precision — it only doubles the traffic. Please pass `self.communication_data_type` so both paths match the storage dtype.

delock (changes requested):

delock:
Hi @alanhuangyoo , the updated PR looks overall good for me.  The code review revealed a fix needed for communication data type, can you take a look at the comments?  Thanks!

deepspeed/runtime/zero/stage3.py
alanhuangyoo: Fixed in 952cc44, both gathers now use `self.communication_data_type`. Muon suite still passes (261 on 2 GPUs).

deepspeed/runtime/zero/stage3.py
alanhuangyoo: Same commit.

delock (approved):

alanhuangyoo:
@delock The cpu-torch-latest failure was mine: `test_zero3_autoep_contiguous_grads_average_by_global_dp` builds the optimizer without `__init__` and stubs out `_apply_distributed_muon_update`, and I had added an `offload_optimizer` check at the call site. Moved it inside the function in 3473e2c, so the reduce path is back to calling it unconditionally like on master. That test and the Muon offload/ZeRO-3 tests pass locally. The push turned auto-merge off, sorry, could you re-enable it?

PR #8054, opened 2026-06-08
Add Biren SUPA accelerator support

frozenleaves:
# Add Biren SUPA Accelerator Support

## Summary

This PR adds accelerator backend support for the **Biren SUPA GPU** (the Biren Technology GPU, software stack SUPA) to DeepSpeed, enabling DeepSpeed to automatically detect the device, run training and inference on Biren GPUs, and reuse DeepSpeed's existing operator invocation framework (fused optimizer, transformer inference, quantizer, async-io, etc.).

SUPA is onboarded as the 9th supported accelerator, following `cuda / cpu / xpu / npu / mps / hpu / mlu / sdaa`. It adheres to DeepSpeed's existing `DeepSpeedAccelerator` abstract interface and the `op_builder` plugin mechanism, with **zero intrusion** into existing backends — the only existing file modified is the accelerator auto-detection entry point `accelerator/real_accelerator.py`.

### 1. Accelerator auto-detection and registration — `accelerator/real_accelerator.py` (the only existing file modified)

- Add `'supa'` to `SUPPORTED_ACCELERATOR_LIST`.
- **Explicit specification** (`DS_ACCELERATOR=supa`): attempt `import torch_supa`, and emit a clear error message if it is missing.
- **Auto-detection**: add a SUPA probing branch that determines availability via `import torch_supa` and checking `torch.supa.is_available()`.
  - Critical ordering: because `torch_supa` spoofs `torch.cuda`, the SUPA detection branch **must come before** the CUDA detection, otherwise Biren cards would be misidentified as CUDA devices. This constraint is clearly noted with a comment in the code.
- In the third-step instantiation logic, add the `accelerator_name == 'supa'` → `SUPA_Accelerator()` branch.

### 2. Accelerator implementation — `accelerator/supa_accelerator.py`
Implements all interfaces of the `DeepSpeedAccelerator` abstract base class. The vast majority of APIs delegate directly to `torch.supa.*` (mirroring the semantics of `torch.cuda.*`):

- **Device management**: `device / set_device / current_device / device_count / synchronize`, etc.
- **RNG**: `manual_seed(_all) / get_rng_state / set_rng_state / default_generator`.
- **Stream / Event**: `Stream / Event / current_stream / default_stream`.
- **Memory management**: `empty_cache / memory_allocated / max_memory_allocated / memory_reserved / memory_stats / total_memory / available_memory`, etc. (some use `hasattr` for capability probing, for compatibility across different versions of torch_supa).
[...]

PR #8605, opened 2026-09-20
Implement graph operations for NPU accelerator

### LINKED ISSUE #8604: [REQUEST] NPU: implement graph operations for short kernel sequences (follow-up to #4318)
#4318 introduced accelerator graph operations (`create_graph` / `capture_to_graph` / `replay_graph`) to capture short kernel sequences — gradient norm, gradient clipping, and the BF16 optimizer's hp→lp gradient write-back, enabled via the `graph_harvesting` config and driven by `graph_process` in `deepspeed/runtime/utils.py`. CUDA (#4318), HPU (#4912), MLU (#6472), and SUPA (#8054) have all implemented the trio on top of their runtime graph APIs.

The NPU accelerator still carries the stub that #4318 originally left: the three methods return `None` / a no-op context / nothing. The feature is therefore dead on NPU — `graph_process` caches the graph and "replays" it every step, but the replay does nothing, so the captured computation silently stops running (verified on real hardware: from the second step on, the reported gradient norm stops tracking the actual value, clipping stops being applied, and the hp→lp write-back stops, with no error raised).

Implement the three methods on `NPU_Accelerator` (`accelerator/npu_accelerator.py`) with the graph API torch_npu ships, mirroring the HPU/MLU/SUPA templates:

```python
def create_graph(self):
    return torch.npu.NPUGraph()

def capture_to_graph(self, graph, pool=None, stream=None):
    return torch.npu.graph(graph, pool, stream)

def replay_graph(self, graph):
    graph.replay()
```

Add hardware-independent contract tests (`tests/unit/accelerator/test_npu_accelerator.py`) that stub `torch.npu` and run on CPU-only CI, following the pattern of `tests/unit/accelerator/test_mps_accelerator.py`. Real-device validation (numerical parity and the `graph_harvesting` end-to-end path, with performance data) is reported in the PR, as done in #8532.

- **Leave the stub as-is.** The silent no-op replay trap remains for anyone who enables `graph_harvesting`, and the launch-bound speedup stays unrealized on Ascend.
- **Make `graph_process` raise when the accelerator cannot capture.** Prevents the silent skip but provides none of the performance benefit, and still needs this implementation to ever be removed.
- **Wait for the torch_npu graph API to stabilize.** The interface shape used here (graph object + `graph()` context manager + `replay()`) is exactly what the abstract interface already assumes and what the other vendor runtimes were integrated with at the time.

**Additional context**
- Config chain: `ds_config["graph_harvesting"]: true` → `engine.graph_harvesting` → `BF16_Optimizer(graph_harvesting=True)` → `graph_process`. The chain works on NPU once the trio is implemented, verified end-to-end on Ascend 910B4: the three recorded functions are captured and replay matches eager execution.
- Validation environment: CANN 8.5.0, torch 2.9.0+cpu, torch_npu 2.9.0.post7, Ascend 910B4 (aarch64).

VenusTZZ:
Fixes #8604

Follow-up to #4318 (graph operations for short kernel sequences).

## Summary
- Implement create_graph / capture_to_graph / replay_graph on NPU_Accelerator via the graph API torch_npu ships: torch.npu.NPUGraph() creates the graph object, torch.npu.graph(graph, pool, stream) is the capture context manager, and graph.replay() replays. Previously NPU_Accelerator carried the stub that #4318 originally left (None / a no-op context / nothing), so graph_process cached an empty graph and every subsequent "replay" did nothing: with graph_harvesting enabled, gradient norms silently went stale, clipping silently stopped, and the hp→lp gradient write-back silently stopped — reproduced on hardware (see below).
- Mirrors the HPU (#4912), MLU (#6472), and SUPA (#8054) implementations, which wrap their runtime graph APIs the same way.
- Adds tests/unit/accelerator/test_npu_accelerator.py with three hardware-independent contract tests that stub torch.npu and run on CPU-only CI, following the pattern of tests/unit/accelerator/test_mps_accelerator.py.

## Implementation notes
The torch_npu graph API. NPUGraph wraps the runtime's graph object with capture_begin / capture_end / replay / reset / pool. torch.npu.graph(graph, pool, stream) is the context manager that performs the synchronization, cache cleanup, side-stream switching, and capture begin/end around the with block. pool and stream pass through positionally and default to None; graph_process captures without pool/stream, so the defaults are part of the contract. The graph APIs are core torch_npu APIs available across the torch versions DeepSpeed supports, so the methods call them directly with no existence checks — matching the CUDA/HPU/MLU/SUPA implementations.

What it unblocks. The whole graph_harvesting chain already existed and is accelerator-agnostic: ds_config["graph_harvesting"]: true → engine.graph_harvesting → BF16_Optimizer(graph_harvesting=True) → graph_process (gradient norm, clipping, update_hp_grads). Implementing the trio is the only missing piece on NPU.

Performance on Ascend 910B4
Whole-block timing with torch.npu.synchronize() on both sides of the block, 3 rounds, best round (shared-box methodology), CANN 8.5.0, torch 2.9.0+cpu, torch_npu 2.9.0.post7, Ascend 910B4 (aarch64):

| Workload | eager (us/iter) | graph replay (us/iter) | speedup |
|---|---|---|---|
| 200 independent tiny ops (norm-style) | 6456 | 910 | 7.1x |
| 50-step chained small ops (clip-style) | 3067 | 135 | 22.7x |
| 20 ops on a 1M-element tensor | 1271 | 307 | 4.1x |

End-to-end: with graph_harvesting: true on Ascend 910B4, _norm_tensors, clip_tensors, and _update_hp_grads_func are all captured into graph_cache, and 5 training steps (BF16_Optimizer, bf16 + fp32 gradient accumulation) are bitwise-identical to eager execution (equal losses, torch.equal parameters).

Before/after: the same graph_process scenario run against the stub vs. this implementation — with the stub the reported norm froze at its first-step value, clipping stopped being applied, and the hp accumulation stopped, with no error raised; with this change all three track eager exactly.

## Test plan
- pre-commit run --files on touched paths — yapf, flake8, check-torchdist, check-torchcuda, check-license, codespell all pass
- pytest tests/unit/accelerator/ — 5 passed (3 new NPU contract tests + 2 existing MPS tests)
- pytest tests/unit/v1/pin_memory/ — 45 passed (no regression)
- Real-device on Ascend 910B4: capture → replay parity across two consecutive replays and explicit pool/stream arguments; graph_harvesting end-to-end chain; stub-vs-fix corruption demonstration

## Known limitation
- Pre-existing and unrelated to NPU: engine.step() calls zero_grad() with the torch default set_to_none=True, which trips the graph_harvesting assert in deepspeed/runtime/bf16_optimizer.py on every accelerator. The end-to-end validation above calls optimizer.step() with zero_grad(set_to_none=False); worth a separate issue.


delock (approved):

VenusTZZ:
Thanks for the review, @delock!

  I noticed the cpu-torch-latest CI run was failing on the new NPU graph tests and found the cause: npu_accelerator.py only had import torch.npu inside a try/except, which binds the torch name only when it succeeds. On machines with torch_npu it always succeeds (which is why local runs passed), but on CPU-only CI the ImportError is swallowed and torch stays unbound — so the three graph methods raise NameError.

  Fixed by adding a plain import torch in the same try block (one line, setup-stage behavior unchanged). Reproduced the failure in a CI-like environment (torch 2.10.0+cpu without torch_npu) and confirmed the tests pass there, with no regression on NPU.

  Pushed as 9d9d00fb.

The workflow runs for the new commit are waiting for approval — would appreciate it if you could approve them.

delock:
> Thanks for the review, @delock!
> 
> I noticed the cpu-torch-latest CI run was failing on the new NPU graph tests and found the cause: npu_accelerator.py only had import torch.npu inside a try/except, which binds the torch name only when it succeeds. On machines with torch_npu it always succeeds (which is why local runs passed), but on CPU-only CI the ImportError is swallowed and torch stays unbound — so the three graph methods raise NameError.
> 
> Fixed by adding a plain import torch in the same try block (one line, setup-stage behavior unchanged). Reproduced the failure in a CI-like environment (torch 2.10.0+cpu without torch_npu) and confirmed the tests pass there, with no regression on NPU.
> 
> Pushed as [9d9d00f](https://github.com/deepspeedai/DeepSpeed/commit/9d9d00fb2be712ae0f99fd533f1e6586f45b8067).
> 
> The workflow runs for the new commit are waiting for approval — would appreciate it if you could approve them.

Thanks for the update.  I have a question for the test.  From the content of test it looks like it should be device agnostic.  Why the test didn't use get_accelerator() to get the NPU_accelerator on a NPU system?  We encourage writing device agnostic test instead of test designed for specific device.  Is there a reason for this test to be NPU specific?

VenusTZZ:
> Thanks for the update. I have a question for the test. From the content of test it looks like it should be device agnostic. Why the test didn't use get_accelerator() to get the NPU_accelerator on a NPU system? We encourage writing device agnostic test instead of test designed for specific device. Is there a reason for this test to be NPU specific?

Thanks for the review, @delock. You're right — I should have driven this through the accelerator interface from the start; stubbing torch.npu to run on CPU-only CI  is what made the test NPU-specific. Apologies for the miss.

Reworked in bb038ba7 to be device-agnostic via get_accelerator():
- test_graph_capture_replay_roundtrip — real create/capture/replay roundtrip on whatever device the machine has; asserts replay re-executes the captured work (skips on backends without graph support).
- test_graph_ops_run_eagerly_without_graph_support — the no-graph contract (create_graph → None, capture runs eagerly, replay is a no-op), which is what CPU-only CI executes.

Verified on Ascend 910B4 (real NPUGraph capture/replay) and in a CPU-only CI-replica; pre-commit clean.

VenusTZZ:
Thanks a lot for reviewing and merging this pull request!

PR #8606, opened 2026-09-20
Nightly full-suite run with classified failure triage and automated git-bisect

delock:
## Motivation

The master full suite on `modal-torch-latest` grew from ~52 to ~70 minutes within a month and keeps climbing ~7-8 min/week, repeatedly exhausting its Sandbox/job time budgets (the 3600 -> 4200 -> 5400s raises are chasing the growth). This PR stops paying the full-suite cost on every push and adds automated, noise-free triage for the nightly regression signal.

## Changes

**1. Master pushes select their own tests; nightly runs the full suite**
- `push` to `master` now diffs against the tip it replaced (`github.event.before`) instead of forcing `all`; the fetcher's fail-safe still widens to the full suite whenever the merge-base is unavailable.
- A `schedule` (cron, daily) run always executes the full suite.
- `concurrency` group now includes `github.event_name` so a nightly run and pushes don't cancel each other.

**2. Controller classifies its own failures (`ci/torch_latest.py`)**
- `infra` (exit 75): the Sandbox never got a GPU instance — no test ran, no candidate information.
- `timeout` (exit 124): the Sandbox lifetime budget was exhausted mid-run.
- `test` (exit 1): everything else, raised with a full traceback as before.
- Each class prints a `DS_CI_FAILURE_CLASS=<class>` sentinel line that survives into job logs even when the job is killed before exiting.

**3. Nightly triage (`nightly-bisect.yml`, `ci/nightly_classify.sh`, `ci/bisect_nightly.sh`)**
- **green** → move the `nightly-last-green` git tag to the tested SHA (persistent good endpoint for bisect; operational failure days simply leave it in place and widen the next range).
- **infra** → no-op (per design, no retry).
- **timeout** → open an issue (a timeout is not bisectable; intermediate commits would time out too). A failed run with no sentinel is treated as a timeout (job was killed).
- **test failures** → `git bisect run` between `nightly-last-green` and the failing SHA, dispatching `modal-torch-latest` at each step (each step is a full run, up to ~105 min; the 6-hour hosted-job cap bounds the range). The culprit is reported in an issue. Inconclusive intermediate steps abort the bisect rather than skip, so the searched range is never silently narrowed.

**4. Housekeeping**: refreshed stale self-tests that no longer matched master (sandbox timeout 4200 -> 5400 left behind by #8571, merge-group trusted-ref expression, the baked-image `FakeImage` chain, and the removed `reinstall Torch packages` runtime command). `ci/test_torch_latest.py` is 24/24 green; pre-commit passes on all changed files.

## Known limitations (why this is a draft)

- Bisect steps dispatch the **full** suite (workflow_dispatch has no base ref); a nightly failure spanning many days of merges may exceed the 6-hour job budget and abort inconclusively. A follow-up could pass the failing test list as a subset to shrink steps to ~15 min.
- Revisions that predate this PR's sentinel have ambiguous classification in bisect; the first bisect over a range crossing this merge may abort and need one manual rerun.
- The nightly run doubles as an upstream-drift sentinel (it installs transformers from git `main`); red nightlies caused by upstream drift will bisect to a "culprit" only if the drift genuinely breaks our tests — the issue text keeps the run link for human judgment.


delock:
**Update: bisect steps now run only the failing tests.**

- `workflow_dispatch` on `modal-torch-latest` accepts a `test_targets` input (newline-separated test files), validated through the same `validate-selection` allowlist as a graph-selected list.
- The nightly-bisect job extracts the nightly's `FAILED` test files and dispatches each bisect step with just those targets: a step drops from a ~70-minute full run to install overhead (~8 min) plus the failing tests.
- Targets missing at a step's commit are dropped from that step's dispatch (a test cannot fail where it does not exist — but other failing targets may still condemn the commit, so the step still runs); a step where none survive is good without dispatching.

With ~15-minute steps, the 6-hour job budget now covers a dozen-plus bisect steps instead of three.

delock:
Per maintainer feedback: master receives no direct pushes (the merge queue lands every entry, and each `merge_group` run already tested the merged tree), so a post-push Sandbox run duplicated the queue's work on every merge.

06977afa0 removes the `push` trigger entirely, along with the now-dead `github.event.before` selection branch — cleaner than skipping `deploy` on push, because the whole run (including `collect-tests`) no longer fires. Cross-entry interaction regressions remain the nightly full suite's job. Sandbox execution now happens only for `merge_group`, `schedule`, and `workflow_dispatch`.

delock:
Codex comments has been addressed in lateset commit, thanks!

PR #8612, opened 2026-09-20
Forward the per-head Muon tag at every muon_update call site

vineethsaivs:
Per-head Muon (#8384) is silently ignored on three of the six paths that call `muon_update`, so the attention projections it tags are orthogonalized whole.

`deepspeed.initialize` tags each attention projection with `muon_num_heads` and the call sites read it back off the parameter. Three do not: the round-robin branch of ZeRO 1/2 `get_flat_partition` (its unpadded twin does), `MuonWithAuxAdam.step`, and the new BF16_Optimizer Muon support. Which branch of `get_flat_partition` runs depends on `round_robin_gradients`, so whether the feature takes effect turns on an unrelated setting, and the update stays finite either way. Forward `num_heads` at all three.

Test: `TORCHDYNAMO_DISABLE=1 DS_ACCELERATOR=cpu PYTHONPATH=.:tests python -m pytest tests/unit/runtime/zero/test_per_head_muon.py -q`. Two new cases fail before and 78 pass after: one steps `MuonWithAuxAdam` on a tagged projection and asserts the applied update is the per-head one rather than the whole-matrix one (0.32 apart here), and one names any call site that drops the tag. yapf 0.43.0 and flake8 pass on the changed files. Executed on Apple M2 Pro CPU; no GPU or distributed run.


alanhuangyoo (commented):
Thanks, good catch, I missed these when I added per-head. Small thing: from what I can see the padded branch of `get_flat_partition` is picked by `zero_optimization.parameter_alignment`, not `round_robin_gradients` (round robin alone stays unpadded on an H20 here), so the description might need a tweak.

Also the AST test feels a bit fragile, a positional arg would trip it. Maybe a small run with `parameter_alignment: true` could cover the ZeRO path the way your `MuonWithAuxAdam` test does? 78 pass on an H20 here btw.

vineethsaivs:
You're right, thanks: round robin only reorders the params, and the padding comes from `parameter_alignment` when a param's size is off the alignment. I fixed the description and added a bf16 ZeRO 1/2 engine test with `parameter_alignment: true` that spies on `muon_update`; it fails with the fix reverted. I kept the AST scan since it's the only thing covering BF16_Optimizer, but it now counts positional args.

PR #8618, opened 2026-09-21
Add device-agnostic ZeRO-3 CPU-offload end-to-end benchmark

delock:
## What does this PR do?

Adds `benchmarks/pin_memory/zero3_offload_bench.py`, a companion to the existing `h2d_d2h_bench.py`:

- `h2d_d2h_bench.py` measures isolated H2D/D2H copy bandwidth per pinning arm;
- this new script measures the **end-to-end effect on ZeRO-3 CPU-offload training step time**, comparing `DS_PIN_MEMORY_REGISTER_DEVICE=1` vs `0` on the native pin backend (`DS_PIN_MEMORY_BACKEND=native`) with `offload_optimizer` + `offload_param` (both cpu, pin_memory).

### Design

- Each arm (registered / unregistered) runs in its **own subprocess** with a fresh rendezvous port, so device state, JIT caches, and any lingering registrations from one arm can never contaminate the other.
- Model selection: a real HF architecture via `--model` (instantiated from the config with random weights, which suffice for timing) or a self-contained synthetic stack via `--hidden` / `--layers` (no network access needed).
- Device-agnostic: the device API is resolved via `getattr(torch, device.type)`, so the same script runs on **cuda and xpu**; synchronization and peak-memory reporting adapt automatically.
- Prints a per-arm table (avg/min step time, GPU peak memory), the speedup, per-step saving, throughput in tok/s, and a machine-readable `DRIVERRESULT=` JSON line.

### Usage

```bash
# real model architecture
python benchmarks/pin_memory/zero3_offload_bench.py --model Qwen/Qwen2.5-7B --batch 4 --seq 512

# synthetic stack, no network needed
python benchmarks/pin_memory/zero3_offload_bench.py --hidden 2048 --layers 12 --batch 4 --seq 128
```

Note: the native backend mlocks host memory; multi-GB models need a raised `RLIMIT_MEMLOCK` (`ulimit -l`) or root.

## Notes forReviewers

Follow-up to the pin-memory registration hooks (#8283) and the XPU implementation (#8315): this benchmark is the end-to-end counterpart of the copy-level numbers, runnable on any accelerator that implements the native pin backend.

🦉


benchmarks/pin_memory/zero3_offload_bench.py
```diff
@@ -0,0 +1,253 @@
+# SPDX-License-Identifier: Apache-2.0
```
sfc-gh-truwase, line 1: Can we move to https://github.com/deepspeedai/DeepSpeedExamples/tree/master/benchmarks

delock:
Sounds good — moved it to DeepSpeedExamples in deepspeedai/DeepSpeedExamples#1013 (script unchanged, plus a README). Closing this in favor of that PR.

PR #8619, opened 2026-09-21
fix: preserve CUDA architecture variant suffixes

adenzhou1350:
## Summary
- accept PyTorch-supported CUDA architecture variants such as `9.0a` and `12.0a`
- preserve the variant suffix through sorting, deduplication, `TORCH_CUDA_ARCH_LIST`, and emitted `-gencode` flags
- add regression coverage for JIT and non-JIT paths, including mixed PTX variants

PyTorch 2.11 accepts `9.0a`, `10.0a`, `11.0a`, `12.0a`, and related targets. DeepSpeed currently parses the minor component with `int()`, so these valid values fail before compilation.

## Testing
- `pytest tests/unit/ops/test_op_builder.py -q -k "not forked_child"` — 17 passed
- repository pre-commit hooks for the changed files — all applicable hooks pass
- flake8 under Python 3.11 — pass
- `git diff --check` — pass

The omitted fork/CUDA integration test is platform-specific and cannot run on Windows; it is unrelated to architecture parsing.


ebarkhordar (commented):
I ran this against `5f46f0bf` in a clean `python:3.11-slim` container with `torch 2.14.0+cpu`, as a differential against the merge base `4bd251b6`.

The comment above the dedupe says it exists to match PyTorch's own, so I compared them directly: the non-JIT `-gencode` list this function returns, against `torch.utils.cpp_extension._get_cuda_arch_flags()` fed the `TORCH_CUDA_ARCH_LIST` the same call writes.

```
input          writes         gencode == torch   at base 4bd251b6
9.0a           9.0a           True               ValueError: '0a'
9.0a+PTX       9.0a+PTX       True               ValueError: '0a'
12.0a;9.0a     9.0a;12.0a     True               ValueError: '0a'
9.0;9.0a       9.0;9.0a       True               ValueError: '0a'
10.0a+PTX      10.0a+PTX      True               ValueError: '0a'
8.0;8.0+PTX    8.0+PTX        True               match
9.0+PTX        9.0+PTX        True               match
```

`9.0;9.0a` is the row I expected to be awkward, since the two now key apart and both survive the dedupe. PyTorch splits them the same way, so the emitted pair agrees there too. The last two rows are unchanged from base, so the existing parity is kept.

The asymmetry in how `_parse_cuda_arch` is called (a split list in the dedupe loop, `cc.split('.')` in the JIT sort) is fine: all eight `filter_ccs` implementations, the base one and the seven overrides, return `[cc.split('.') for cc in ccs]`. I enumerated those by parsing `op_builder/*.py` rather than grepping, so that is the complete set.

I did not test whether nvcc accepts the emitted `compute_90a`/`sm_90a` pairs. There is no CUDA toolkit in this container.

PR #8544, opened 2026-09-16
Add affine AutoEP checkpoint placement

jinyouzhi:
Follow up #8385 

This pull request introduces support for flexible expert placement in DeepSpeed's AutoEP (Automatic Expert Placement) system, enabling non-uniform, non-contiguous, and replicated expert layouts. The main changes add a versioned expert placement descriptor, validation logic, and integration into checkpoint consolidation and metadata validation. This lays the groundwork for more advanced expert scheduling and model parallelism strategies.

The most important changes are:

**AutoEP Expert Placement Descriptor and Affine Map Lowering**
- Added a new module `autoep_affine.py` that defines the expert placement descriptor, validation, legacy uniform descriptor synthesis, and lowering to affine maps for sharded tensor reconstruction. This enables flexible, versioned expert placement beyond the legacy uniform contiguous layout.

**Integration into Checkpoint Consolidation and Metadata**
- Updated `autoep_universal.py` to:
  - Accept and validate the new expert placement descriptor in layer metadata, relaxing the requirement that `num_local_experts * ep_size == num_experts` when a placement is provided.
  - Use the placement descriptor and affine map for reconstructing full expert tensors during checkpoint consolidation, supporting arbitrary expert layouts. [[1]](diffhunk://#diff-ee96c370ed05cdc2757c212d5a407fbc3bd4370814512fbfa9041a8e2189a179L291-R345) [[2]](diffhunk://#diff-ee96c370ed05cdc2757c212d5a407fbc3bd4370814512fbfa9041a8e2189a179R360-R377)
  - Validate placement consistency during expert file consolidation.
  - Import and use the new placement logic. [[1]](diffhunk://#diff-ee96c370ed05cdc2757c212d5a407fbc3bd4370814512fbfa9041a8e2189a179R19-R27) [[2]](diffhunk://#diff-ee96c370ed05cdc2757c212d5a407fbc3bd4370814512fbfa9041a8e2189a179R49)

**Metadata Validation Enhancements**
- Updated `autoep_zero3_metadata.py` to:
  - Import and use placement validation and legacy descriptor synthesis.
[...]

PR #8623, opened 2026-09-22
[UC] Plan a scale-1 affine shard-to-shard transfer without rebuilding the parameter

0z5a:
**Incremental review:** [Files changed](https://github.com/deepspeedai/DeepSpeed/pull/8623/files) against `master`.

### Summary

- Compose a source affine map and a target affine map into strided copies between named ranks of each shard, so a parameter can move between two tensor-parallel layouts without any rank holding the full logical tensor.
- Metadata only, scale 1 only. No transport, no CUDA, no checkpoint format change, and no claim about transfer speed -- nothing here moves a byte.
- Two new files. No edits to `affine.py`, `universal_checkpoint.py`, `ds_to_universal.py` or `layers.py`, so it does not collide with #8622, #8609 or #8544.

`affine_ir_spec.md` §8.2 states two rules as normative for a composer -- compress in N dimensions, and group runs by source rank before folding -- and this implements those two rules and nothing beyond them.

### The direction rule, which is the easy thing to get wrong

A piece's `source_*` addresses the full logical parameter and its `dest_*` addresses the shard holding it. So the address to *read* in a source shard is the source piece's `dest_*`, not its `source_*`. Every offset and stride the plan emits points into a shard, both sides are in elements as `torch.as_strided` takes them, and byte conversion is left to whoever moves them. A source rank and a target rank that carry the same number are different namespaces, not the same process, so the plan names both.

### Support set

| | accepted | refused |
|---|---|---|
| piece strides | both vectors are the row-major strides of the tensor they address | a permutation, a broadcast stride, any axis correspondence other than identity |
| scale | 1 | anything else, for the reason `affine.py` already gives: the optimizer's `lr` and `eps` would stay in the source coordinate |
| shard rank | same number of axes as the logical parameter | a flat shard beside a matrix parameter -- no axis correspondence to plan against |
| replicas | several ranks declaring one region, resolved to a fixed lowest holder | two pieces with different location sets over one region; two regions overlapping without agreeing |
| shard addresses | each rank's pieces partition its shard | two pieces of one rank occupying the same shard addresses |
| padding | identical logical shapes only | any truncate/pad, which needs `V_orig` and is not attempted here |
| plan size | within explicit budgets | exhausted budget is an error, never a shortened plan |

### Coverage is argued over pieces

`validate_coverage` walks every element of the parameter by design, so it is never called. In its place:

1. Each rank's pieces must partition its shard **by address**. This does not follow from the logical boxes being disjoint -- three pieces can cover three disjoint runs of the parameter while two land on the same shard addresses, and their element counts sum correctly either way.
2. Each target region must be tiled by the intersections that reach it. Because the source regions were made disjoint in step 1, those intersections are disjoint without further checks, so the volumes deciding it.
3. Volume totals are compared **last**. One gap and one overlap of equal size cancel in any sum, including a sum over the whole plan, which is why `moved_elements` is reported and not trusted.

Merging joins two segments only when every other axis matches in both position and extent and each side's own address continues at its own stride. It is deterministic and it is not a shortest plan.

### Validation

Executed on the 8x RTX 5090 host, CPU-only (`CUDA_VISIBLE_DEVICES=""`; Python 3.12.13, torch 2.13.0+cu130), code at this PR's head. The `tests/unit/checkpoint` suite was run on a pristine `1eb56d29` worktree and on this one, and the failing/erroring test names are **identical** -- 241 either way, none introduced here. Two environmental causes on both trees: the multi-rank cases cannot run without a visible accelerator, and this environment ships pytest 9.1.0 while the repository's dev requirements pin `pytest>=7.2.0,<8.4.0`, so `DistributedFixture`/`DistributedTest` never register their fixtures and every distributed checkpoint test errors at setup before reaching DeepSpeed.

- **89 tests** in `test_affine_transfer_plan.py`: executed plans compared element for element against `rebuild` -> `extract` across seven layouts x nine degree pairs (row, uneven row, column, replicated, sub-params, GPTBigCode segments, Yuan block gather), plus uneven tail blocks, 60 seeded random layouts including a change of split axis, non-contiguous rank numbers, determinism under reordered pieces, one corruption per refusal class, and budget exhaustion.
- The 8x8 row-TP2 -> column-TP4 case is asserted against hand-derived numbers rather than against a computed oracle: 8 blocks of 4x2, source offset `2b` with strides `(8,1)`, target offset `8a` with strides `(2,1)`, 64 elements total.
- The materializing operations (`rebuild`, `extract`, `uncovered_offsets`, `holders`, `validate_coverage`) are bound to raisers on **both** the defining module and the importing module before planning a layout, so the no-materialization claim is tested rather than asserted.

Planning cost against geometry, fixed topology (row TP2 -> column TP4), logical size swept across eight orders of magnitude:

| logical shape | segments | pairs checked | plan bytes | median ms | p95 ms |
|---|---:|---:|---:|---:|---:|
| 10 x 8 | 8 | 8 | 653 | 0.3077 | 0.4641 |
| 100 x 8 | 8 | 8 | 669 | 0.2934 | 0.4391 |
| 1,000 x 8 | 8 | 8 | 685 | 0.2967 | 0.4974 |
| 10,000 x 8 | 8 | 8 | 701 | 0.3095 | 0.4476 |
| 100,000 x 8 | 8 | 8 | 717 | 0.3041 | 0.4373 |
| 1,000,000 x 8 | 8 | 8 | 733 | 0.3015 | 0.4468 |
| 100,000,000 x 8 | 8 | 8 | 765 | 0.3220 | 0.4721 |
| 1,000,000,000 x 8 | 8 | 8 | 781 | 0.3105 | 0.4756 |

Segments, pair checks and peak allocation do not move; `plan_bytes` grows only because the integers get more digits. No parameter of any of these sizes is created.

The same parameter at the same size, varying geometry (48 x 8):

| case (source -> target) | src pieces | tgt pieces | pairs | before merge | segments | plan bytes | median ms |
|---|---:|---:|---:|---:|---:|---:|---:|
| row2->row4 | 2 | 4 | 8 | 4 | 4 | 334 | 0.2295 |
| row4->row2 | 4 | 2 | 8 | 4 | 4 | 334 | 0.2505 |
| row2->col4 | 2 | 4 | 8 | 8 | 8 | 665 | 0.2559 |
| replicated4->row4 | 1 | 4 | 4 | 4 | 4 | 337 | 0.1209 |
| bigcode2->bigcode4 | 3 | 8 | 24 | 8 | 8 | 673 | 0.2630 |
| bigcode4->bigcode2 | 5 | 4 | 20 | 6 | 6 | 501 | 0.2627 |

What the path this replaces has to hold, on CPU, FP32:

| parameter | degrees | tensors held by rebuild->extract | its median ms | planning ms | plan bytes |
|---|---|---:|---:|---:|---:|
| 64x64 | 2->4 | 0.0 MB | 0.173 | 0.2661 | 350 |
| 512x512 | 2->4 | 3.1 MB | 0.44 | 0.281 | 371 |
| 2048x2048 | 2->4 | 50.3 MB | 2.641 | 0.2747 | 389 |
| 4096x4096 | 2->8 | 201.3 MB | 15.051 | 0.4423 | 793 |

The last two columns are **not** a speedup claim. Planning is cheaper here because it moves no data, and a real transfer still has to happen after it. The comparison that survives is the memory one: the materializing path holds the full parameter plus both sides' shards, and this holds neither.

### One question for @Achyuthan-S

§8.2 measures bigcode 2->4 at 7 pieces. This planner emits 8 segments, and the difference is the replica policy rather than the geometry: with GPTBigCode the query rows and the key/value block are separate segments for every target rank, and only target rank 3's two halves can share a source rank -- its query rows and the key/value block both sit in source rank 1's shard, contiguously, so they fold into one copy and the count becomes 7. Choosing the holder per target produces that fold; this version always takes the lowest declared holder, because which replica is cheapest depends on a topology the metadata does not know. Happy to make that policy a parameter if you would rather it live here than in the transport.

### Scope

- Planner only. No NCCL, no executor, no CUDA graph, no chunking, no overlap -- those are the follow-up this is meant to feed.
- Does not decide whether two parameters are the same parameter. Equal shapes are not evidence of that, and fused parameters have been matched by byte count before.
- No padding or vocabulary translation, no non-unit scale, no optimizer coordinate transform, no quantization or pack/unpack difference.
- Replica *value* agreement is an input contract. A metadata-only planner cannot detect a corrupt replica; `rebuild` can, because it reads both.
- Refs #8230, Refs #8252. This is the phase-2 planning step that `affine_ir_spec.md` §8 defers; it does not close either issue.

Test and benchmark sources are retained locally; the validation results below refer to those local files.


Achyuthan-S:
@0z5a Both numbers are reachable — §8.2's 7 is not conditional on choosing the replica per target.

I recomposed it from `segmented_map` at TP2 and TP4 and executed the plans against a real
tensor, checked element-for-element against `extract`: lowest holder gives 7, per-target
choice gives 6, both exact with nothing uncovered. Under lowest holder the fold is on target
rank 1 rather than 3 — source rank 0's query block ends exactly where its kv block begins,
and target rank 1's query rows are the last 8 of it, so `src0[8:32] -> tgt1[0:24]` is one
contiguous copy on both sides.

If you are getting 8, I would guess the merge stops at the query/kv boundary. That rule is
§5 consequence 4, and it is about map pieces: a piece carries `locations`, so it cannot span
a region that is rank-private in one half and replicated in the other. A plan segment carries
no `locations` — the holder is resolved before any copy is emitted — so the boundary is not
load-bearing there. Worth stating in the spec, since §5 reads like it forbids the fold
everywhere.

On the policy: agreed it should be a parameter. Lowest holder is not topology-neutral either
— it points every replica read at rank 0, which is the case a real transport least wants.

0z5a:
> @0z5a Both numbers are reachable — §8.2's 7 is not conditional on choosing the replica per target.
> 
> 
> 
> I recomposed it from `segmented_map` at TP2 and TP4 and executed the plans against a real
> 
> tensor, checked element-for-element against `extract`: lowest holder gives 7, per-target
> 
> choice gives 6, both exact with nothing uncovered. Under lowest holder the fold is on target
> 
> rank 1 rather than 3 — source rank 0's query block ends exactly where its kv block begins,
> 
> and target rank 1's query rows are the last 8 of it, so `src0[8:32] -> tgt1[0:24]` is one
> 
> contiguous copy on both sides.
> 
> 
> 
> If you are getting 8, I would guess the merge stops at the query/kv boundary. That rule is
> 
> §5 consequence 4, and it is about map pieces: a piece carries `locations`, so it cannot span
> 
> a region that is rank-private in one half and replicated in the other. A plan segment carries
> 
> no `locations` — the holder is resolved before any copy is emitted — so the boundary is not
> 
> load-bearing there. Worth stating in the spec, since §5 reads like it forbids the fold
> 
> everywhere.
> 
> 
> 
> On the policy: agreed it should be a parameter. Lowest holder is not topology-neutral either
> 
> — it points every replica read at rank 0, which is the case a real transport least wants.

Thanks, that clears up the distinction.

The current merge is indeed stopping at the source-piece boundary, so the TP2 -> TP4 BigCode case stays at 8. That is carrying the map-level homogeneity constraint from §5 into the resolved transfer plan, where locations is no longer part of the descriptor.

The update will separate those two rules:

* map pieces remain subject to the §5/P5 replication-boundary constraint;
* resolved transfer segments may fold across that boundary when the selected source rank is the same and both source and destination shard addresses are affine-contiguous.

With the existing lowest-holder policy, BigCode TP2 -> TP4 should therefore regress to 7 segments. I’ll also make replica-holder selection an explicit planner policy rather than baking lowest-holder into the geometry. A per-target policy can then exercise the 6-segment case separately, without treating that as the canonical/default plan.

This also seems worth clarifying next to the composer rules in §8.2: P5 constrains the IR pieces, not already-resolved transfer segments.

Achyuthan-S:
Verified against the new head. I rebuilt the maps with segmented_map at TP2 and TP4 rather than using your fixture, planned with plan_transfer, and executed the segments element for element against extract: lowest holder gives 7, a per-target selector gives 6, both write 768 with nothing uncovered. Your committed suite also passes here — 97, CPU.

The spec text is right, and separating "resolved copies are not another homogeneous map" from P5 is the part I'd have got wrong if I'd written it alone. A folded copy having no single logical origin is the consequence I hadn't followed through.

One thing to tighten: "a specified per-target holder choice yields 6" depends on which per-target policy. A selector that takes the highest holder still gives 7 — I tried it. Only a selector that picks the rank holding the target's own private rows reaches 6, because that is what puts the query read adjacent to the kv read in the source shard. Worth saying, since
"per-target" alone does not imply the fold.

The selector hook is the right shape. It is the seam a cost model plugs into, which is the part I took on #8252.

